PR #546, opened 2016-10-05
Payoff some tech debt in tests

dmohs:
[This commit](https://github.com/broadinstitute/rawls/commit/8fccfdce4c03c1ff4837e5918e038b3e44b25f15) fixes a bug in our test data which made it unusable in some circumstances. Likely due to that problem, some tests create test data when they could be using the stock test data. This PR proves that to be true and provides an example of how to clean things up.

Found while working on [951](https://broadinstitute.atlassian.net/browse/GAWB-951).
    and
    DBs in Google Cloud Console
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- [x] Tell  that the PR exists if he wants to look at it **(apply requires_doge label)**
- Review cycle:
  - LR reviews
- [x]  sign off
- [x] **LR** sign off
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [x] **Submitter**: Verify swagger UI on dev environment still works after deployment
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

### DIFF
--- a/src/test/resources/logback-test.xml
+++ b/src/test/resources/logback-test.xml
@@ -15,6 +15,9 @@
         </encoder>
     </appender>
 
+    <!-- Thanks but no thanks on the giant block of log configuration status. -->
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener"/>
+
     <root level="info">
         <appender-ref ref="FILE"/>
         <appender-ref ref="console"/>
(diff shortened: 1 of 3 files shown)


dvoet (commented):
So this now instantiates the entire test data set instead of only what it needs. Why is that better? What was the issue that was fixed?

dmohs:
@dvoet This is better because it requires less code to accomplish the same task without significantly decreasing performance. The issue fixed is "more code than necessary to do something." I trust it is not necessary to enumerate the benefits.

More generally, isn't this the entire point of having test data? If not, why does it make sense in the previous test ("get a users own status") and not in this one?

dvoet (approved):
meh.. I don't really care:
pro - less code - but I have lower standards for test
con - slower test

I think people have been using the default test data because they are lazy (which I am a fan of) at the expense of slower tests and I would like to figure out how to fix that, i.e. how can we continue to be lazy but not have slower tests.

MatthewBemis:
:+1:

dmohs:
> how can we continue to be lazy but not have slower tests

+1

PR #547, opened 2016-10-05
GAWB-1096 Reworked how attribute lists are stored in the database and exported to JSON

helgridly:
This PR covers GAWB-1114 as well. It has a few moving parts, so here's a walkthrough. The links dotted through the doc will jump you around the diff.

Sibling PR in orch: https://github.com/broadinstitute/firecloud-orchestration/pull/271
Sibling PR in UI: <as yet unwritten>
# The Context

https://broadinstitute.atlassian.net/browse/GAWB-1096

TL;DR: Put this.participant into a method config input and you don't get an error. You should.
# The Bugs
## Bug One: We don't throw an error when an attribute expression returns an entity

Here's [the bug](https://github.com/broadinstitute/rawls/pull/547/files#diff-5311caed81a6a46128e4132cfeeedac8L243): "Since we know we are not dealing with a reference here as this is the attribute final func, we can pass in None."

Turns out that's wrong: we've pulled the attribute with the matching name (in this case, "participant"), but because we pass in None for the referenced entity, we end up skipping past [the branch we wanted to go into](https://github.com/broadinstitute/rawls/blob/8114dfef312fe18d555710f774dcb03820a968ab/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala#L348) and end up treating it as a value instead, ultimately [returning AttributeNull](https://github.com/broadinstitute/rawls/blob/8114dfef312fe18d555710f774dcb03820a968ab/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala#L397).

The fix is to [distinguish between reference and value attributes](https://github.com/broadinstitute/rawls/pull/547/files#diff-5311caed81a6a46128e4132cfeeedac8R251) ahead of time, parse the value attributes nicely, and then [handle the entity references](https://github.com/broadinstitute/rawls/pull/547/files#diff-5311caed81a6a46128e4132cfeeedac8R257) by turning them into fake `AttributeEntityReference` objects that we can [catch and throw an exception in later](https://github.com/broadinstitute/rawls/pull/547/files#diff-5311caed81a6a46128e4132cfeeedac8R399).
### Detour: Distinguishing between reference and value attributes

Well, this is easy, right? Here are the fields of an AttributeRecord:

```
trait AttributeRecord[OWNER_ID] {
  val id: Long
  val ownerId: OWNER_ID
  val namespace: String
  val name: String
  val valueString: Option[String]
  val valueNumber: Option[Double]
  val valueBoolean: Option[Boolean]
  val valueEntityRef: Option[Long]
  val listIndex: Option[Int]
  val listLength: Option[Int]
}
```

It's an entity reference if `valueEntityRef` is defined and a value otherwise. For lists, we populate the `listIndex` field to indicate which element of the list the record refers to, pull 'em all, and reassemble the list afterwards.

For empty lists, we just save a dummy record with `listIndex = -1` to indicate that the list is empty, and leave all the `valueFoo` fields blank.

Ah.

Now we can't tell the difference between an empty reference list and an empty value list, because [they're saved to the database in the same way.](https://github.com/broadinstitute/rawls/pull/547/files#diff-37c2cc14b84b1a0ed1d99ba2adb74b18L234)

Guess we have to fix that now.
## Bug Two: Can't tell the difference between an empty value list and an empty reference list

This doesn't _seem_ to be a huge problem: the list is empty, who cares?

Well, maybe you're a user who accidentally writes `this.samples` in your method config expression when you really mean `this.samples.name` (I've seen this happen), and it doesn't error the first time because your sample set was empty. Then you reuse the method config a second time and it suddenly stops working because we fixed Bug One and now throw errors when you end up with a ref at the end of a value expression.

So let's fix that. New rule! For empty value lists, [we jam a -1 in the `valueNumber` field](https://github.com/broadinstitute/rawls/pull/547/files#diff-37c2cc14b84b1a0ed1d99ba2adb74b18R199). For empty ref lists, we leave them all blank (because `valueEntityRef` is enforced with a foreign key constraint, so we can't put a dummy value in there). This also means we get rid of `AttributeEmptyList` in favour of [two new types](https://github.com/broadinstitute/rawls/pull/547/files#diff-37c2cc14b84b1a0ed1d99ba2adb74b18R248), `AttributeEntityReferenceEmptyList` and `AttributeValueEmptyList`.

Then when we go to unmarshal the list from the database, if the `listIndex` is -1 we can [look at the `valueNumber` field and distinguish whether it's an empty ref or value list](https://github.com/broadinstitute/rawls/pull/547/files#diff-37c2cc14b84b1a0ed1d99ba2adb74b18R381).

This is great - we've now added types to empty lists (something even Scala can't do!) and we can distinguish between them.

Unfortunately we're still not done yet.
## Bug Three: JSON representation is still ambiguous

`AttributeEmptyList` was serialized out to JSON as `[]`. Now we've ditched that in favour of two separate empty list types, how do we distinguish between them in JSON? Unless we do something about this, we can't roundtrip our data through JSON, as we'll have no idea how to interpret `[]` and _must_ error.

Welp, there's only one thing we can do. Here's [how lists are represented now](https://github.com/broadinstitute/rawls/pull/547/files#diff-6242a09492d6dabd946ed5cadd93113aR68):

```
{
    "itemsType" : "AttributeValue", //or "EntityReference"
    "items" : [1,2,3]
}
```

Eww.

That's what the changes around [AttributeFormat](https://github.com/broadinstitute/rawls/pull/547/files#diff-6242a09492d6dabd946ed5cadd93113aR39) are about.

But we only want to use this representation when talking to the outside world. Cromwell expects the old behaviour of lists being plain old JSON arrays, so we encapsulate that behaviour [in a new different WDLJsonSupport class](https://github.com/broadinstitute/rawls/pull/547/files#diff-527aaa28a026d49549dfda679d66e494R5).
### Tweaking JsonSupport

The first version of this PR had an icky implementation of providing different AttributeList serialization strategies (object-y version vs. plain old lists). This has since been reworked thanks to a suggestion from Doge.

`JsonSupport.AttributeFormat` now [has a mixin](https://github.com/broadinstitute/rawls/pull/547/files#diff-6242a09492d6dabd946ed5cadd93113aR97) called `AttributeListSerializer`. There are two versions of this mixin: `TypedAttributeListSerializer` and `PlainArrayAttributeListSerializer`, representing the object-y version and the plain old list version. [The default version of the implicit](https://github.com/broadinstitute/rawls/pull/547/files#diff-6242a09492d6dabd946ed5cadd93113aR128) uses the `TypedAttributeListSerializer`, but things that interact with Cromwell provide [their own JsonSupport implicit](https://github.com/broadinstitute/rawls/pull/547/files#diff-527aaa28a026d49549dfda679d66e494R6) that overrides the mixin.
## Bonus time: deserialize fewer things from Cromwell

A bunch of our endpoints just pass through to Cromwell. We were deserializing them to Attributes and then back to JSON again in the HTTP response. [I made that go away, because why?](https://github.com/broadinstitute/rawls/pull/547/files#diff-4c7743f07a412bba7e50f81369371f28R200)
# Checklist time!
    and
    DBs in Google Cloud Console
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- [ ] Tell  that the PR exists if he wants to look at it **(apply requires_doge label)**
- Review cycle:
  - LR reviews
- [ ]  sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161007_empty_attributevalue_lists.xml
+++ b/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161007_empty_attributevalue_lists.xml
@@ -1,21 +1,31 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
-    <changeSet id="add_minusone_to_empty_entityattribute_value_lists" author="hussein" logicalFilePath="dummy">
+    <changeSet id="fixup_empty_entityattribute_value_lists" author="hussein" logicalFilePath="dummy">
         <sql>
             UPDATE ENTITY_ATTRIBUTE eattr
             SET
-            eattr.value_number = -1
+            eattr.value_number = -1, eattr.list_index = NULL
             WHERE
             eattr.list_index = -1 AND eattr.name not in ('participants', 'samples', 'pairs')
         </sql>
     </changeSet>
 
-    <changeSet id="add_minusone_to_empty_wsattribute_value_lists" author="hussein" logicalFilePath="dummy">
+    <changeSet id="fixup_empty_entityattribute_ref_lists" author="hussein" logicalFilePath="dummy">
+        <sql>
+            UPDATE ENTITY_ATTRIBUTE eattr
+            SET
+            eattr.list_index = NULL
+            WHERE
+            eattr.list_index = -1 AND eattr.name in ('participants', 'samples', 'pairs')
+        </sql>
+    </changeSet>
+
+    <changeSet id="fixup_empty_wsattribute_value_lists" author="hussein" logicalFilePath="dummy">
         <sql>
             UPDATE WORKSPACE_ATTRIBUTE wsattr
             SET
-            wsattr.value_number = -1
+            wsattr.value_number = -1, wsattr.list_index = NULL
             WHERE
             wsattr.list_index = -1
         </sql>
@@ -25,7 +35,7 @@
         <sql>
             UPDATE SUBMISSION_ATTRIBUTE subattr
             SET
-            subattr.value_number = -1
+            subattr.value_number = -1, subattr.list_index = NULL
             WHERE
             subattr.list_index = -1
         </sql>
--- a/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161007_empty_attributevalue_lists.xml
+++ b/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161007_empty_attributevalue_lists.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="add_minusone_to_empty_entityattribute_value_lists" author="hussein" logicalFilePath="dummy">
+        <sql>
+            UPDATE ENTITY_ATTRIBUTE eattr
+            SET
+            eattr.value_number = -1
+            WHERE
+            eattr.list_index = -1 AND eattr.name not in ('participants', 'samples', 'pairs')
+        </sql>
+    </changeSet>
+
+    <changeSet id="add_minusone_to_empty_wsattribute_value_lists" author="hussein" logicalFilePath="dummy">
+        <sql>
+            UPDATE WORKSPACE_ATTRIBUTE wsattr
+            SET
+            wsattr.value_number = -1
+            WHERE
+            wsattr.list_index = -1
+        </sql>
+    </changeSet>
+
+    <changeSet id="add_minusone_to_empty_subattribute_value_lists" author="hussein" logicalFilePath="dummy">
+        <sql>
+            UPDATE SUBMISSION_ATTRIBUTE subattr
+            SET
+            subattr.value_number = -1
+            WHERE
+            subattr.list_index = -1
+        </sql>
+    </changeSet>
+
+</databaseChangeLog>
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -98,11 +98,8 @@ trait AttributeComponent {
 
   import driver.api._
 
-  //Magic const we put in listIndex to indicate the dummy row is a list type.
-  val EMPTY_LIST_INDEX_MAGIC = -1
-
   def isEntityRefRecord[T](rec: AttributeRecord[T]): Boolean = {
-    val isEmptyRefListDummyRecord = rec.listIndex.isDefined && rec.listIndex.get == -1 && rec.valueNumber.isEmpty
+    val isEmptyRefListDummyRecord = rec.listLength.isDefined && rec.listLength.get == 0 && rec.valueNumber.isEmpty
     rec.valueEntityRef.isDefined || isEmptyRefListDummyRecord
   }
 
@@ -197,7 +194,7 @@ trait AttributeComponent {
     def insertAttributeRecords(ownerId: OWNER_ID, attributeName: AttributeName, attribute: Attribute, workspaceId: UUID): Seq[ReadWriteAction[Int]] = {
 
       def insertEmptyVal: Seq[ReadWriteAction[Int]] = {
-        Seq(insertAttributeValue(ownerId, attributeName, AttributeNumber(-1), Option(EMPTY_LIST_INDEX_MAGIC), Option(0)))
+        Seq(insertAttributeValue(ownerId, attributeName, AttributeNumber(-1), None, Option(0)))
       }
 
       def insertEmptyRef: Seq[ReadWriteAction[Int]] = {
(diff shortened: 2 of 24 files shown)


src/test/scala/org/broadinstitute/dsde/rawls/expressions/SlickSimpleExpressionParserTest.scala
```diff
@@ -104,6 +104,13 @@ class SlickSimpleExpressionParserTest extends FunSuite with TestDriverComponent
     }
   }
 
+  test("attribute expression that refers to an entity should except") {
```
helgridly, line 107: this is the bug that started it all!

davidangb:
High-level comments, just from reading the PR description:
- could you represent an empty list by setting list_length=0 instead of list_index=-1? It feels more semantically correct.
- I don't like overriding the meaning of value_number (by setting it to -1) to indicate it's an empty value list instead of an empty reference list. What about something like setting list_length to 0 for empty value lists, and list_length to -1 for empty reference lists? (Or, set list_length to -2^31 for empty value lists, and -2^30 for empty reference lists) I realize you're trying to avoid schema changes by reusing columns.

helgridly:
@davidangb 
1. list_index needs to be set to something, and everything >= 0 is taken for nonempty lists. Therefore it needs to be a magic negative number anyway, which is sufficient to indicate emptiness. We could _also_ set list_length=0 and use that instead but that's now two pieces of information representing the same thing.
2. "There's a weird dummy value in the number column" seemed more intuitive as an indicator of empty-val-list than overriding the use of the list_length column. I'd have put a dummy value in the ref column to indicate empty-ref-list too, but it's FK constrained so it had to stay null.

helgridly:
Also, two more notes while we're here:

N1: list_length is never used or checked - we've always used list_index to check for magic-ness.  
N2: At some point in the future the magic-numberiness will continue when we need to disambiguate between sets and lists.

src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161007_empty_attributevalue_lists.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="add_minusone_to_empty_entityattribute_value_lists" author="hussein" logicalFilePath="dummy">
+        <sql>
+            UPDATE ENTITY_ATTRIBUTE eattr
+            SET
+            eattr.value_number = -1
+            WHERE
+            eattr.list_index = -1 AND eattr.name not in ('participants', 'samples', 'pairs')
```
dvoet, line 10: icky but I don't have a less icky answer

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -98,6 +98,14 @@ trait AttributeComponent {
 
   import driver.api._
 
+  //Magic const we put in listIndex to indicate the dummy row is a list type.
+  val EMPTY_LIST_INDEX_MAGIC = -1
```
dvoet: I don't know if we care or not but we can make this -1 go away because we have since added list length. A list length of 0 means what you would expect. A list length of None means it is not a list. So we could just get rid of this magic number now.

src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
```diff
@@ -1,57 +1,145 @@
...
+
+    case _ => throw new DeserializationException("unexpected json type")
+  }
+
+  def writeAttributeList[T <: Attribute](listType: String, list: Seq[T]): JsValue = {
+    JsObject( Map(LIST_ITEMS_TYPE_KEY -> JsString(listType), LIST_ITEMS_KEY -> JsArray(list.map( writeAttribute ).toSeq:_*)) )
+  }
+
+  def readAttributeList(jsMap: Map[String, JsValue]) = {
+    val attrList: Seq[Attribute] = jsMap(LIST_ITEMS_KEY) match {
```
dvoet, line 73: do you want jsMap.get(LIST_ITEMS_KEY) instead? Currently you will get a key exception if LIST_ITEMS_KEY does not exist. Same on line 74.

src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
```diff
@@ -1,57 +1,145 @@
...
-      case _ => throw new DeserializationException("unexpected json type")
-    }
 
-    def getAttributeList(s: Seq[Attribute]) = s match {
-      case e: Seq[_] if e.isEmpty => AttributeEmptyList
-      case v: Seq[AttributeValue @unchecked] if (s.map(_.isInstanceOf[AttributeValue]).reduce(_&&_)) => AttributeValueList(v)
-      case r: Seq[AttributeEntityReference @unchecked] if (s.map(_.isInstanceOf[AttributeEntityReference]).reduce(_&&_)) => AttributeEntityReferenceList(r)
-      case _ => throw new DeserializationException("illegal array type")
+      case JsObject(members) if ENTITY_OBJECT_KEYS subsetOf members.keySet =>
+        AttributeEntityReference(members(ENTITY_TYPE_KEY).asInstanceOf[JsString].value, members(ENTITY_NAME_KEY).asInstanceOf[JsString].value)
```
dvoet: do a match instead of type casting:
(members(ENTITY_TYPE_KEY), members(ENTITY_NAME_KEY)) match {
  case (JsString(x), JsString(y)) =>
  case _ => throw exception
}

dvoet (changes requested):
a couple minor things

src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
helgridly: it's only called from `readListType()` which has already checked the key does exist

src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
helgridly: done in next commit

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
helgridly: david an was sad about it, and i care too, so i'll just eat this one and get it done

src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
```diff
@@ -119,8 +120,11 @@ class JsonSupport extends DefaultJsonProtocol {
       case JsBoolean(b) => AttributeBoolean(b)
       case JsNumber(n) => AttributeNumber(n)
 
-      case JsObject(members) if ENTITY_OBJECT_KEYS subsetOf members.keySet =>
-        AttributeEntityReference(members(ENTITY_TYPE_KEY).asInstanceOf[JsString].value, members(ENTITY_NAME_KEY).asInstanceOf[JsString].value)
+      case JsObject(members) if ENTITY_OBJECT_KEYS subsetOf members.keySet => (members(ENTITY_TYPE_KEY), members(ENTITY_NAME_KEY)) match {
+        case (JsString(typeKey), JsString(nameKey)) => AttributeEntityReference(typeKey, nameKey)
+        case _ => throw new RawlsException()
```
dvoet: add a message to this exception

dvoet (approved):
add message then 👍

PR #552, opened 2016-10-19
GAWB-1059: Endpoints for listing workspaces by attribute

MattPutnam:
The query is doing a full scan of the Workspace Attribute table, which seems unavoidable without adding an index on attribute name at least.

To find the desired attributes:

```
select x2.`namespace`, x2.`name`, x2.`id`, x2.`bucket_name`, x2.`created_date`, x2.`last_modified`, x2.`created_by`, x2.`is_locked`, x2.`realm_group_name`, x2.`record_version` from `WORKSPACE` x2, (select `namespace` as x3, `name` as x4, `value_boolean` as x5, `owner_id` as x6 from `WORKSPACE_ATTRIBUTE` where ((`namespace` = 'library') and (`name` = 'published')) and (`value_boolean` = true)) x7 where x2.`id` = x7.x6
```

Explain:

| id | select_type | table | type | possible_keys | key | key_len | ref | rows | Extra |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | PRIMARY | <derived2> | ALL | NULL | NULL | NULL | NULL | 600 | NULL |
| 1 | PRIMARY | x2 | eq_ref | PRIMARY | PRIMARY | 16 | x7.x6 | 1 | NULL |
| 2 | DERIVED | WORKSPACE_ATTRIBUTE | ALL | NULL | NULL | NULL | NULL | 600 | "Using where" |

Then to get the workspaces:

```
select x2.x3, x2.x4, x2.x5, x2.x6, x2.x7, x2.x8, x2.x9, x2.x10, x2.x11, x2.x12, x13.`id`, x13.`name`, x13.`entity_type`, x13.`workspace_id`, x13.`record_version`, x13.`all_attribute_values` from (select `owner_id` as x3, `namespace` as x5, `value_entity_ref` as x10, `list_index` as x11, `list_length` as x12, `value_boolean` as x9, `id` as x4, `value_string` as x7, `name` as x6, `value_number` as x8 from `WORKSPACE_ATTRIBUTE` where `owner_id` in (select x14.`id` from `WORKSPACE_ATTRIBUTE` x15, `WORKSPACE` x14 where (((x15.`namespace` = 'library') and (x15.`name` = 'published')) and (x15.`value_boolean` = true)) and (x14.`id` = x15.`owner_id`))) x2 left outer join `ENTITY` x13 on x2.x10 = x13.`id`
```

Explain:

| id | select_type | table | type | possible_keys | key | key_len | ref | rows | Extra |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | PRIMARY | <derived2> | ALL | NULL | NULL | NULL | NULL | 450 | NULL |
| 1 | PRIMARY | x13 | eq_ref | PRIMARY | PRIMARY | 8 | x2.x10 | 1 | NULL |
| 2 | DERIVED | x15 | ALL | UNQ_WORKSPACE_ATTRIBUTE | NULL | NULL | NULL | 600 | "Using where; Start temporary" |
| 2 | DERIVED | WORKSPACE_ATTRIBUTE | ref | UNQ_WORKSPACE_ATTRIBUTE | UNQ_WORKSPACE_ATTRIBUTE | 16 | rawls.x15.owner_id | 1 | NULL |
| 2 | DERIVED | x14 | eq_ref | PRIMARY | PRIMARY | 16 | rawls.x15.owner_id | 1 | "Using index; End temporary" |
    and
    DBs in Google Cloud Console
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- [x] Tell  that the PR exists if he wants to look at it **(apply requires_doge label)**
- Review cycle:
  - LR reviews
- [ ]  sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -165,7 +165,7 @@ trait AttributeComponent {
       findByNameQuery(attrName).filter { rec =>
         attrValue match {
           case AttributeString(s) => rec.valueString === s
-          case AttributeNumber(n) => rec.valueNumber === n
+          case AttributeNumber(n) => rec.valueNumber === n.doubleValue
           case AttributeBoolean(b) => rec.valueBoolean === b
           case _ => throw new RawlsException("Unsupported attribute type")
         }
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -161,12 +161,13 @@ trait AttributeComponent {
       filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
     }
 
-    def queryByAttribute(attrName: AttributeName, attrValue: Attribute): WorkspaceAttributeQueryType = {
+    def queryByAttribute(attrName: AttributeName, attrValue: AttributeValue): WorkspaceAttributeQueryType = {
       findByNameQuery(attrName).filter { rec =>
         attrValue match {
-          case s:AttributeString => rec.valueString === s.value
-          case n:AttributeNumber => rec.valueNumber === n.value.toDouble
-          case b:AttributeBoolean => rec.valueBoolean === b.value
+          case AttributeString(s) => rec.valueString === s
+          case AttributeNumber(n) => rec.valueNumber === n
+          case AttributeBoolean(b) => rec.valueBoolean === b
+          case _ => throw new RawlsException("Unsupported attribute type")
         }
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -154,24 +154,7 @@ trait AttributeComponent {
 
   protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord)
 
-  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord) {
-    private type WorkspaceAttributeQueryType = driver.api.Query[WorkspaceAttributeTable, WorkspaceAttributeRecord, Seq]
-
-    def findByNameQuery(attrName: AttributeName): WorkspaceAttributeQueryType = {
-      filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
-    }
-
-    def queryByAttribute(attrName: AttributeName, attrValue: AttributeValue): WorkspaceAttributeQueryType = {
-      findByNameQuery(attrName).filter { rec =>
-        attrValue match {
-          case AttributeString(s) => rec.valueString === s
-          case AttributeNumber(n) => rec.valueNumber === n.doubleValue
-          case AttributeBoolean(b) => rec.valueBoolean === b
-          case _ => throw new RawlsException("Unsupported attribute type")
-        }
-      }
-    }
-  }
+  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord)
 
   protected object submissionAttributeQuery extends AttributeQuery[Long, SubmissionAttributeRecord, SubmissionAttributeTable](new SubmissionAttributeTable(_), SubmissionAttributeRecord)
 
@@ -288,6 +271,21 @@ trait AttributeComponent {
       }
 
       createRecord(0, ownerId, attributeName.namespace, attributeName.name, valueString, valueNumber, valueBoolean, None, listIndex, listLength)
+    }
+
+    def findByNameQuery(attrName: AttributeName) = {
+      filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
+    }
+
+    def queryByAttribute(attrName: AttributeName, attrValue: AttributeValue) = {
+      findByNameQuery(attrName).filter { rec =>
+        attrValue match {
+          case AttributeString(s) => rec.valueString === s
+          case AttributeNumber(n) => rec.valueNumber === n.doubleValue
+          case AttributeBoolean(b) => rec.valueBoolean === b
+          case _ => throw new RawlsException("Unsupported attribute type")
+        }
+      }
     }
 
     def deleteAttributeRecords(attributeRecords: Seq[RECORD]): DBIOAction[Int, NoStream, Write] = {
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -154,7 +154,23 @@ trait AttributeComponent {
 
   protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord)
 
-  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord)
+  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord) {
+    private type WorkspaceAttributeQueryType = driver.api.Query[WorkspaceAttributeTable, WorkspaceAttributeRecord, Seq]
+
+    def findByNameQuery(attrName: AttributeName): WorkspaceAttributeQueryType = {
+      filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
+    }
+
+    def queryByAttribute(attrName: AttributeName, attrValue: Attribute): WorkspaceAttributeQueryType = {
+      findByNameQuery(attrName).filter { rec =>
+        attrValue match {
+          case s:AttributeString => rec.valueString === s.value
+          case n:AttributeNumber => rec.valueNumber === n.value.toDouble
+          case b:AttributeBoolean => rec.valueBoolean === b.value
+        }
+      }
+    }
+  }
 
   protected object submissionAttributeQuery extends AttributeQuery[Long, SubmissionAttributeRecord, SubmissionAttributeTable](new SubmissionAttributeTable(_), SubmissionAttributeRecord)
 
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -231,6 +231,16 @@ trait WorkspaceComponent {
       }
     }
 
+    /**
+      * Lists all workspaces with a particular attribute name/value pair.
+      *
+      * ** Note: This is an inefficient query.  It performs a full scan of the attribute table since
+      *    there is no index on attribute name and/or value.  This method is only being used in one
+      *    place; if you find yourself needing this for other things, consider adding such an index.
+      * @param attrName
+      * @param attrValue
+      * @return
+      */
     def getWorkspacesWithAttribute(attrName: AttributeName, attrValue: AttributeValue) = {
       for {
         attribute <- workspaceAttributeQuery.queryByAttribute(attrName, attrValue)
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -85,6 +85,10 @@ trait WorkspaceComponent {
       loadWorkspaces(getPublishedWorkspaces)
     }
 
+    def listWithAttribute(attributeMap: AttributeMap): ReadAction[Seq[Workspace]] = {
+      loadWorkspaces(getWorkspacesWithAttribute(attributeMap))
+    }
+
     def save(workspace: Workspace): ReadWriteAction[Workspace] = {
       validateUserDefinedString(workspace.namespace)
       validateUserDefinedString(workspace.name)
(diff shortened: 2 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -154,7 +154,23 @@ trait AttributeComponent {
...
+    def findByNameQuery(attrName: AttributeName): WorkspaceAttributeQueryType = {
+      filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
+    }
+
+    def queryByAttribute(attrName: AttributeName, attrValue: Attribute): WorkspaceAttributeQueryType = {
+      findByNameQuery(attrName).filter { rec =>
+        attrValue match {
+          case s:AttributeString => rec.valueString === s.value
+          case n:AttributeNumber => rec.valueNumber === n.value.toDouble
+          case b:AttributeBoolean => rec.valueBoolean === b.value
```
dvoet: since this does not match all cases you should have a case _ => throw exception

also your cases would be better like
case AttributeString(s) => rec.valueString === s
etc.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -154,7 +154,23 @@ trait AttributeComponent {
 
   protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord)
 
-  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord)
+  protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord) {
+    private type WorkspaceAttributeQueryType = driver.api.Query[WorkspaceAttributeTable, WorkspaceAttributeRecord, Seq]
+
+    def findByNameQuery(attrName: AttributeName): WorkspaceAttributeQueryType = {
```
dvoet: why not put these 2 methods in AttributeQuery? Why is it specific to workspaces?

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -227,6 +235,18 @@ trait WorkspaceComponent {
       }
     }
 
+    def getPublishedWorkspaces = {
+      getWorkspacesWithAttribute(Map(AttributeName.libraryAttribute("published") -> AttributeBoolean(true)))
+    }
+
+    def getWorkspacesWithAttribute(attributeMap: AttributeMap) = {
```
dvoet: why does this take a map as an input but ignore all but the head? This should take attrName and attrValue as parameters.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -81,6 +81,14 @@ trait WorkspaceComponent {
       loadWorkspaces(workspaceQuery)
     }
 
+    def listPublished(): ReadAction[Seq[Workspace]] = {
```
dvoet: a) I don't think this belongs in rawls at all
b) it certainly does not belong here in the DAO. This is business logic and should be in WorkspaceService. 

This goes for getPublishedWorkspaces too.

src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -46,6 +46,8 @@ object AttributeName {
 
   def withDefaultNS(name: String) = AttributeName(defaultNamespace, name)
 
+  def libraryAttribute(name: String) = AttributeName(libraryNamespace, name)
```
dvoet: I would call this withLibraryNS to be consistent with withDefaultNS

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -122,6 +124,8 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     case UpdateWorkspace(workspaceName, operations) => pipe(updateWorkspace(workspaceName, operations)) to sender
     case ListWorkspaces => pipe(listWorkspaces()) to sender
     case ListAllWorkspaces => pipe(listAllWorkspaces()) to sender
+    case ListPublishedWorkspaces => pipe(listPublishedWorkspaces()) to sender
+    case ListWorkspacesWithAttribute(attributeMap) => pipe(listWorkspacesWithAttribute(attributeMap)) to sender
```
dvoet: The convention is to prefix admin functions with Admin (e.g. AdminListPublishedWorkspaces). Also the asFCAdmin should be up here:
case AdminListPublishedWorkspaces => pipe(asFCAdmin { listPublishedWorkspaces() }) to sender

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -237,6 +239,22 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
...
+    path("admin" / "publishedWorkspaces") {
+      get {
+        requestContext => perRequest(requestContext,
+          WorkspaceService.props(workspaceServiceConstructor, userInfo),
+          WorkspaceService.ListPublishedWorkspaces)
+      }
+    } ~
+    path("admin" / "workspacesByAttribute") {
+      post {
+        entity(as[AttributeMap]) { attributeMap =>
```
dvoet: since we only handle a single key/value, this should take a single key/value, not a map. Also, this probably should be a Get not a post.

davidangb:
I don't think rawls should have the list-published endpoint; it should have just the query-by-attribute endpoint. Orchestration has the Library knowledge (see https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala#L27) ... orchestration can just call the query-by-attribute endpoint, passing the appropriate criteria.

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -237,6 +239,22 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
             WorkspaceService.ListAllWorkspaces)
       }
     } ~
+    path("admin" / "publishedWorkspaces") {
```
dvoet: I don't think rawls should have this API. Whoever is calling this should use the admin/workspacesByAttribute api instead.

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1306,6 +1310,22 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+    asFCAdmin {
+      dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listPublished.map(RequestComplete(StatusCodes.OK, _))
+      }
+    }
+  }
+
+  def listWorkspacesWithAttribute(attributeMap: AttributeMap): Future[PerRequestMessage] = {
+    asFCAdmin {
+      dataSource.inTransaction { dataAccess =>
```
davidangb: add something along the lines of

```
      if (attributeMap.size != 1)
        throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, "specify one attribute only"))
```

here. I think it's reasonable to support only a single attribute for now - and your code lower down only looks for AttributeString, AttributeNumber, AttributeBoolean, so the lower code only supports a single attribute - therefore we should validate the request.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
dvoet: do we care that this will work on lists as well?

dvoet (changes requested):
I see you checked the "Database checks" box yet you did not include explain plans.

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MattPutnam: I was following the pattern of ListAllWorkspaces

dvoet: older code, older pattern :( if you want to change that one too, be my guest

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: if we are only supporting size=1, it should not be a map at all, just pass in the key and the value.

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
davidangb: I disagree this should be a GET. The request requires a body, so it should be a POST.

It needs a body so we can distinguish string/numeric/boolean values. We may need to specify a body of:

```
{ "myAttribute" : "123"}
```

which is different from:

```
{ "myAttribute" : 123 }
```

we can't distinguish those if they're in the URL:

```
/api/admin/workspacesByAttribute/myAttribute/123
```

in the url, should `123` be a string or a number?

dvoet: I understand the problem. Semantically, though, this feels like it should be a GET. Also I feel like distinguishing the type in such a subtle way is not a great API. I know it is verbose, but what about query parameters:
attributeName
valueString
valueBoolean
valueNumber

it also seems icky but it is clear what is going on.

Alternately, I would love a way that queried independent of type (i.e. "123" would match the number 123 as well as the string).

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -237,6 +239,15 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
             WorkspaceService.ListAllWorkspaces)
       }
     } ~
+    path("admin" / "workspacesByAttribute") {
+      get {
```
dvoet: I should have been more specific. I agree that if there is a request body this should be a post. But this feels like it should be a get with query parameters.

src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -1088,7 +1088,17 @@ AdminApiServiceSpec extends ApiServiceSpec {
       check {
         assertResult(StatusCodes.OK) { status }
         responseAs[Array[Workspace]] should contain
-        theSameElementsAs(Array(testData.workspace, testData.workspaceNoGroups))
+        theSameElementsAs(Array(testData.workspace, testData.workspaceNoGroups, testData.workspacePublished, testData.workspaceNoAttrs))
+      }
+  }
+
+  it should "return 200 when getting workspaces by attribute" in withTestDataApiServices { services =>
+    Get(s"/admin/workspaces?attributeName=number&valueNumber=10", httpJson(Map("number" -> 10))) ~>
```
MatthewBemis: the json payload isn't needed anymore now that you're using query params

MatthewBemis:
One minor comment and an explain plan for WorkspaceComponent.getWorkspacesWithAttribute and you'll have a :+1: from me

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -232,9 +235,21 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
...
+              case Some(attributeName) =>
+                val name = AttributeName.fromDelimitedName(attributeName)
+                (stringOption, numberOption, booleanOption) match {
+                  case (Some(string), None, None) => WorkspaceService.AdminListWorkspacesWithAttribute(name, AttributeString(string))
+                  case (None, Some(number), None) => WorkspaceService.AdminListWorkspacesWithAttribute(name, AttributeNumber(number.toInt))
+                  case (None, None, Some(boolean)) => WorkspaceService.AdminListWorkspacesWithAttribute(name, AttributeBoolean(boolean.toBoolean))
+                  case _ => throw new RawlsException("Specify exactly one of valueString, valueNumber, or valueBoolean")
+                }
+            })
+        }
```
davidangb, line 252: for code style, I'd prefer to break this up so the large pattern-match is outside of the call to perRequest():

```
val msg = nameOption match {
  ...
}

perRequest(requestContext,
  WorkspaceService.props(workspaceServiceConstructor, userInfo),
  msg)
```

src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -1088,7 +1088,17 @@ AdminApiServiceSpec extends ApiServiceSpec {
...
+      sealRoute(services.adminRoutes) ~>
+      check {
+        assertResult(StatusCodes.OK) { status }
+        responseAs[Array[Workspace]] should contain
+        theSameElementsAs(Array(testData.workspace, testData.workspaceNoGroups, testData.workspacePublished))
```
davidangb, line 1101: could you add tests for number and boolean, too?

src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -232,9 +235,21 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
...
           requestContext => perRequest(requestContext,
             WorkspaceService.props(workspaceServiceConstructor, userInfo),
-            WorkspaceService.ListAllWorkspaces)
+            nameOption match {
+              case None => WorkspaceService.ListAllWorkspaces
+              case Some(attributeName) =>
+                val name = AttributeName.fromDelimitedName(attributeName)
+                (stringOption, numberOption, booleanOption) match {
+                  case (Some(string), None, None) => WorkspaceService.AdminListWorkspacesWithAttribute(name, AttributeString(string))
+                  case (None, Some(number), None) => WorkspaceService.AdminListWorkspacesWithAttribute(name, AttributeNumber(number.toInt))
```
davidangb: why `toInt`? I believe AttributeNumber is backed by a BigDecimal ... so converting to int here could lose precision.

davidangb (commented):
this is getting close!

davidangb:
👍 from me on the latest changes. I still see as outstanding a response to Doug's older comment about methods in workspaceAttributeQuery vs. AttributeQuery (as well as the explain plans)

davidangb:
Re: the full table scan: of course we want to avoid those. But I don't think we should add an index for this specific PR/endpoint. This endpoint is admin-only, and it will only be used by developers to reset our ES indices. This should not happen too often - though it will happen relatively often during active development. To that end, I don't want to add a db index that will (slightly) affect runtime writes, just to make a developer- and admin- endpoint lighter.

dvoet:
the reason I vote for an index now is that I suspect we will use this more later and we will forget that there is no index... how will we remember? But I don't care enough to hold back this PR.

dvoet (approved):
concerned about future need for index but proceed

davidangb:
I suggest then adding either 1) an index, or 2) comments in the relevant code section(s) to indicate there is no index. (if you add a comment, I do not feel the need to re-review)

PR #561, opened 2016-10-26
GAWB-1173 Calculate Google Bucket Usage

breilly2:
    and
    DBs in Google Cloud Console
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/rawls.yaml
+++ b/src/main/resources/swagger/rawls.yaml
@@ -3247,7 +3247,7 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
-  '/api/workspaces/{workspaceNamespace}/{workspaceName}/getBucketUsage':
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/bucketUsage':
     get:
       responses:
         '200':
@@ -4381,7 +4381,7 @@ definitions:
     properties:
       usageInBytes:
         type: integer
-        description: The current storage bucked usage in bytes
+        description: The current storage bucket usage in bytes
 
   Attribute:
     description: ''
--- a/src/main/resources/swagger/rawls.yaml
+++ b/src/main/resources/swagger/rawls.yaml
@@ -3247,6 +3247,35 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/getBucketUsage':
+    get:
+      responses:
+        '200':
+          description: Successful Request
+          schema:
+            $ref: '#/definitions/BucketUsageResponse'
+      description: Get the storage bucket usage
+      tags:
+        - workspaces
+      summary: Get bucket usage
+      operationId: getBucketUsage
+      parameters:
+        - in: path
+          description: Workspace Namespace
+          name: workspaceNamespace
+          required: true
+          type: string
+        - in: path
+          description: Workspace Name
+          name: workspaceName
+          required: true
+          type: string
+      security:
+        - authorization:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
   '/register/user':
      post:
        responses:
@@ -4344,6 +4373,15 @@ definitions:
         description: An optional realm for the workspace
       attributes:
         $ref: ''
+
+  BucketUsageResponse:
+    description: ''
+    required:
+      - usageInBytes
+    properties:
+      usageInBytes:
+        type: integer
+        description: The current storage bucked usage in bytes
 
   Attribute:
     description: ''
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -62,6 +62,20 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def getBucketACL(bucketName: String): Future[Option[List[BucketAccessControl]]]
 
+  /**
+    * Queries the objects in a bucket and calculates the total usage (bytes).
+    *
+    * Note: maxResults is used for integration testing of multi-page queries. While it could potentially be used for
+    * performance tuning, it would be better to build that into the service instead of giving the caller a dial to mess
+    * with. For that reason, the maxResults parameter should be removed in favor of extracting Storage building from the
+    * service implementation to enable test doubles to be injected.
+    *
+    * @param bucketName the name of the bucket to query
+    * @param maxResults (optional) the page size to use when fetching objects
+    * @return the size in bytes of the data stored in the bucket
+    */
+  def getBucketUsage(bucketName: String, maxResults: Option[Long] = None): Future[BigInt]
+
   def diagnosticBucketWrite(user: RawlsUser, bucketName: String): Future[Option[ErrorReport]]
 
   def diagnosticBucketRead(userInfo: UserInfo, bucketName: String): Future[Option[ErrorReport]]
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -25,7 +25,7 @@ import com.google.api.services.oauth2.Oauth2.Builder
 import com.google.api.services.plus.PlusScopes
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule.{Action, Condition}
 import com.google.api.services.storage.model.Bucket.{Lifecycle, Logging}
-import com.google.api.services.storage.model.{Bucket, BucketAccessControl, ObjectAccessControl, StorageObject}
+import com.google.api.services.storage.model._
 import com.google.api.services.storage.{Storage, StorageScopes}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.crypto.{Aes256Cbc, EncryptedBytes, SecretKey}
@@ -322,6 +322,25 @@ class HttpGoogleServicesDAO(
     retryWithRecoverWhen500orGoogleError(() => { Option(executeGoogleRequest(aclGetter).getItems.toList) }) {
       case e: HttpResponseException => None
     }
+  }
+
+  override def getBucketUsage(bucketName: String, maxResults: Option[Long]): Future[BigInt] = {
+    def processPage(pageToken: Option[String] = None): Future[BigInt] = {
+      val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
+      if (maxResults.isDefined) fetcher setMaxResults maxResults.get
+      if (pageToken.isDefined) fetcher setPageToken pageToken.get
+      retryWithRecoverWhen500orGoogleError(() => {
+        val result = executeGoogleRequest(fetcher)
+        val total: BigInt = (BigInt(0) /: result.getItems) { (sum, o) => sum + o.getSize }
+        result match {
+          case r: Objects if r.getNextPageToken == null => total
+          case _ => total + Await.result(processPage(Some(result.getNextPageToken)), Duration.Inf)
+        }
+      }) {
+        case e: HttpResponseException => throw new RawlsException(s"Error gathering storage bucket usage: ${e.getMessage}")
+      }
+    }
+    processPage()
   }
 
   override def getBucket(bucketName: String): Future[Option[Bucket]] = {
--- a/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -326,27 +326,25 @@ class HttpGoogleServicesDAO(
 
   override def getBucketUsage(bucketName: String, maxResults: Option[Long]): Future[BigInt] = {
     def sumItemSizes(items: java.util.List[StorageObject]) = {
-      (BigInt(0) /: items) { (sum, o) => sum + o.getSize }
+      items.map(o => BigInt(o.getSize)).sum
     }
 
-    def processPage(pageToken: Option[String] = None): Future[BigInt] = {
+    def fetchUsage(sum: BigInt, pageToken: Option[String] = None): Future[BigInt] = {
       val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
-      if (maxResults.isDefined) fetcher setMaxResults maxResults.get
-      if (pageToken.isDefined) fetcher setPageToken pageToken.get
+      maxResults.foreach(fetcher.setMaxResults(_))
+      pageToken.foreach(fetcher.setPageToken)
 
-      retryWithRecoverWhen500orGoogleError(() => {
+      retryWhen500orGoogleError(() => {
         val result = executeGoogleRequest(fetcher)
-        result match {
-          case r if r.getItems == null => BigInt(0)
-          case r if r.getNextPageToken == null => sumItemSizes(r.getItems)
-          case r => sumItemSizes(r.getItems) + Await.result(processPage(Some(result.getNextPageToken)), Duration.Inf)
-        }
-      }) {
-        case e: HttpResponseException => throw new RawlsException(s"Error gathering storage bucket usage: ${e.getMessage}")
+        val newSum = Option(result.getItems).map(sumItemSizes).getOrElse(BigInt(0)) + sum
+        (newSum, Option(result.getNextPageToken))
+      }) flatMap {
+        case (newSum, None) => Future.successful(newSum)
+        case (newSum, Some(nextPageToken)) => fetchUsage(newSum, Some(nextPageToken))
       }
     }
 
-    processPage()
+    fetchUsage(BigInt(0))
   }
 
   override def getBucket(bucketName: String): Future[Option[Bucket]] = {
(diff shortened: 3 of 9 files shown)


breilly2:
Please comment on anything, including suggestions to better follow Scala idioms. I'm learning here, so not mentioning an improvement is going to mean I keep doing things the wrong way. Thanks.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +324,31 @@ class HttpGoogleServicesDAO(
...
+      val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
+      if (maxResults.isDefined) fetcher setMaxResults maxResults.get
+      if (pageToken.isDefined) fetcher setPageToken pageToken.get
+
+      retryWithRecoverWhen500orGoogleError(() => {
+        val result = executeGoogleRequest(fetcher)
+        result match {
+          case r if r.getItems == null => BigInt(0)
+          case r if r.getNextPageToken == null => sumItemSizes(r.getItems)
+          case r => sumItemSizes(r.getItems) + Await.result(processPage(Some(result.getNextPageToken)), Duration.Inf)
```
MatthewBemis: Await.result is a blocking task so generally we only use it in tests and in production code that is very rarely used by admins. this endpoint could be called every time a user loads a workspace in the UI so we want to avoid that here

src/main/resources/swagger/rawls.yaml
```diff
@@ -3247,6 +3247,35 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/getBucketUsage':
```
breilly2: It just occurred to me that "getBucketUsage" is a horrible name because "get" is redundant with the HTTP method. I'll change this to "bucketUsage" unless someone can suggest a better name. Since this is an exported API, others who have been around longer probably have stronger feelings about this.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: I thought I thought this through and decided it was okay here, but maybe I missed something. Yes, I am blocking on the result for the next page, but that's somewhat necessary in order to do the addition. Also, retryWithRecoverWhen500orGoogleError() is already returning a Future, so the only thing being blocked here fulfilling that Future which necessarily needs to wait.

I thought about finding a way to get rid of the Await.result(), and would be happy to for cleanliness. Still new to this, so suggestions are welcome.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -62,6 +62,20 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+    * Note: maxResults is used for integration testing of multi-page queries. While it could potentially be used for
+    * performance tuning, it would be better to build that into the service instead of giving the caller a dial to mess
+    * with. For that reason, the maxResults parameter should be removed in favor of extracting Storage building from the
+    * service implementation to enable test doubles to be injected.
+    *
+    * @param bucketName the name of the bucket to query
+    * @param maxResults (optional) the page size to use when fetching objects
+    * @return the size in bytes of the data stored in the bucket
+    */
+  def getBucketUsage(bucketName: String, maxResults: Option[Long] = None): Future[BigInt]
```
geoffjentry, line 77: I'm only butting in because I saw you ask about scala idioms, anything I say here might contradict Rawls idioms. I also made no effort to look at surrounding code, potentially compounding the Rawls idiom effect. YMMV.

Generally `get` is frowned upon in scala, it's more typical to just have `bucketUsage` in a case like this.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +324,31 @@ class HttpGoogleServicesDAO(
     }
   }
 
+  override def getBucketUsage(bucketName: String, maxResults: Option[Long]): Future[BigInt] = {
+    def sumItemSizes(items: java.util.List[StorageObject]) = {
+      (BigInt(0) /: items) { (sum, o) => sum + o.getSize }
```
geoffjentry: IMO leave the operator characters in the past as a bad memory of days gone by in scala, stick w/ `fold`

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +324,31 @@ class HttpGoogleServicesDAO(
...
+    }
+
+    def processPage(pageToken: Option[String] = None): Future[BigInt] = {
+      val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
+      if (maxResults.isDefined) fetcher setMaxResults maxResults.get
```
geoffjentry: I'm not 100% what these 2 lines are doing but I think you can roll with a `foreach` 

If you haven't seen it, http://blog.originate.com/blog/2014/06/15/idiomatic-scala-your-options-do-not-match/ has a phenomenal cheatsheet for options

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
geoffjentry: In my neck of the woods, nearly every single time someone has convinced me that `Await.result` is fine in the context it was in has led to a horrible problem. I still end up getting convinced it is fine. I'm not smart.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +324,31 @@ class HttpGoogleServicesDAO(
...
+          case r if r.getItems == null => BigInt(0)
+          case r if r.getNextPageToken == null => sumItemSizes(r.getItems)
+          case r => sumItemSizes(r.getItems) + Await.result(processPage(Some(result.getNextPageToken)), Duration.Inf)
+        }
+      }) {
+        case e: HttpResponseException => throw new RawlsException(s"Error gathering storage bucket usage: ${e.getMessage}")
+      }
+    }
+
+    processPage()
```
geoffjentry: The `()` is typically used to signify side effects which I don't think is the case here

src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -228,6 +228,8 @@ case class WorkspacePermissionsPair(workspaceId: String,
 
 case class WorkspaceStatus(workspaceName: WorkspaceName, statuses: Map[String, String])
 
+case class BucketUsageResponse(usageInBytes: BigInt)
```
geoffjentry, line 231: case classes should always be marked as final

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: I was getting an error without the () because of the one optional parameter. I agree, I wish I could remove the (). I'm also not entirely satisfied with the code. I could remove the default parameter value and explicitly pass in None here, but then I don't like how that reads. I may see if I can make it more clear by splitting out fetchFirstPage() and fetchNextPage(String) functions.

src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
breilly2: @MatthewBemis do you think there would be objection to adding `final` to all of these? It's either that or leave it as is. I'm not going to break the established practice by making only this one final.

MatthewBemis: i personally don't object to this but i'll defer to @dvoet 

we have a bunch of model classes so we'd want to change them there too

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
breilly2: I'm not crazy about the "get" prefix here, but I think I'll leave it for now since most other functions in `GoogleServicesDAO` have it. I also don't want to imply that it's a simple property access. Maybe just being in `GoogleServicesDAO` is enough, or maybe we could lean toward a different prefix verb such as "fetch".

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: Thanks. In this case, I think I need to `foldLeft` or `foldRight`:

```
items.foldLeft(BigInt(0)) { (sum, o) => sum + o.getSize }
```

The reason is that `items` is an array of some object type rather than an array of `BigInt`. Is that okay or should I first `map` it to an array of `BigInt` and then `fold` it?

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
geoffjentry: Like I said, I have no idea on rawls idioms :)

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: agreed, what is /: anyway. You can probably use reduce instead
`items.map(_.getSize).reduce(_ + _)`

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +324,31 @@ class HttpGoogleServicesDAO(
...
+    def sumItemSizes(items: java.util.List[StorageObject]) = {
+      (BigInt(0) /: items) { (sum, o) => sum + o.getSize }
+    }
+
+    def processPage(pageToken: Option[String] = None): Future[BigInt] = {
+      val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
+      if (maxResults.isDefined) fetcher setMaxResults maxResults.get
+      if (pageToken.isDefined) fetcher setPageToken pageToken.get
+
+      retryWithRecoverWhen500orGoogleError(() => {
```
dvoet: you don't have any recovery code so you can just call retryWhen500orGoogleError and remove the error handling block at the end

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: Unfortunately, it is not fine here either. The problem is that Await blocks the thread which can lead to resource contention/starvation.

The solution is to chain futures together with map or flatMap. I would try something like

```
def processPage(sum: BigInt, pageToken: Option[String] = None): Future[(BigInt, Option[String])] = {
  val fetcher = getStorage(getBucketServiceAccountCredential).objects().list(bucketName)
  maxResults.forEach { fetcher setMaxResults maxResults.get }
  pageToken.forEach { fetcher setPageToken pageToken.get }

  retryWhen500orGoogleError(() => {
    val result = executeGoogleRequest(fetcher)
    (Option(result.getItems).map(sumItemSizes(_.getItems)).getOrElse(BigInt(0)), Option(result.getNextPageToken))
  }) flatMap { 
    case (pageSize, None) => Future.successful(sum + pageSize)
    case (pageSize, Some(nextPageToken)) => processPage(sum + pageSize, Option(nextPageToken))
  }
}
```

src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
dvoet: meh, none of our other case classes are final... just don't extend them and we will be fine

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1466,6 +1468,16 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
+    dataSource.inTransaction { dataAccess =>
+      withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { workspaceContext =>
+        DBIO.from(gcsDAO.getBucketUsage(workspaceContext.workspace.bucketName)) map { usage =>
+          RequestComplete(BucketUsageResponse(usage))
```
dvoet: we have learned that it is a bad idea to include external service calls in a transaction. Try

```
def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
      workspaceContext.workspace.bucketName
    }
  } flatMap { bucketName =>
    gcsDAO.getBucketUsage(bucketName))
  } map { usage =>
    RequestComplete(BucketUsageResponse(usage))
  }
}
```

or better yet

```
def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
  for {
    bucketName <- dataSource.inTransaction { dataAccess =>
        workspaceContext.workspace.bucketName
      }
    }

    usage <- gcsDAO.getBucketUsage(bucketName))

  } yield {
    RequestComplete(BucketUsageResponse(usage))
  }
}
```

src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
```diff
@@ -329,6 +329,30 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationT
     }
   }
 
+  it should "calculate data size" in {
+    val random = scala.util.Random
+    val testUser = testCreator.copy(userSubjectId = random.nextLong().toString)
+    val googleWorkspaceInfo = Await.result(gcsDAO.setupWorkspace(testUser, testProject, UUID.randomUUID.toString, testWorkspace, None), Duration.Inf)
```
dvoet: As I said with face-words, you don't need a workspace to test this, just a bucket with some stuff in it.

dvoet (changes requested):
also add a test to WorkspaceApiServiceSpec that everything besides the actual google call works

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: According to the book I was looking at, `/:` is a `foldLeft` operator.

There's some extra complexity that `getSize` returns a `java.math.BigInteger`. By being a little more verbose and aggressively moving to `BigInt`s as soon as possible, `reduce(_ + _)` can actually become `sum`.

```
items.map((s: StorageObject) => BigInt(s.getSize)).sum
```

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
breilly2: I knew this would come up but wasn't sure how to handle it. Thanks for the suggestions.

Initially, I didn't like the "better yet"; `for` implied to me working with a collection. However, the notion of having sequential generators and an eventual yield reads very nicely... better than the `flatMap`/`map` (which actually is no better than `for` in terms of implying working with a collection).

I did have to use `DBIO.from(Future.successful(...)` to get the bucket name out of the transaction context, but that makes sense.

src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1847,5 +1847,39 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
...
-}
+  it should "return 404 when requesting bucket size for a non-existent workspace" in withTestWorkspacesApiServicesAndUser("reader-access") { services =>
+    Get(s"/workspaces/${testWorkspaces.workspace.namespace}/${testWorkspaces.workspace.name}x/bucketUsage") ~>
+      sealRoute(services.workspaceRoutes) ~>
+      check {
+        assertResult(StatusCodes.NotFound) { status }
+      }
+  }
+
+  // TODO: adjust user access for this test once the requirements of GAWB-1145 settle down
```
MatthewBemis, line 1858: what parts of the requirements are still fluid?

breilly2: There is some discussion about which types of users should see this information. No changes have made their way from the comment thread to the description yet, but there is talk of only showing this to owners or owners and writers.

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1466,6 +1468,19 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
+    for {
+      bucketName <- dataSource.inTransaction { dataAccess =>
+        withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { workspaceContext =>
+          DBIO.from(Future.successful(workspaceContext.workspace.bucketName))
```
dvoet: DBIO.successful instead of DBIO.from and Future.successful

breilly2:
The new approach (using Google storage logs) seems to be working well and is much more viable. I'm abandoning these changes, but thank you all for your feedback.

PR #568, opened 2016-11-09
GAWB-1173 Get Google Bucket Usage from Storage Logs

breilly2:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -233,6 +244,26 @@ class HttpGoogleServicesDAO(
...
 
+  def createStorageLogsBucket(billingProject: RawlsBillingProjectName): Future[String] = {
+    val bucketName = getStorageLogsBucketName(billingProject)
+    logger debug s"storage log bucket: $bucketName"
+
+    retryWithRecoverWhen500orGoogleError(() => {
+      val bucket = new Bucket().setName(bucketName)
+      val deleteAfterOneYear = new Lifecycle.Rule()
+        .setAction(new Action().setType("Delete"))
+        .setCondition(new Condition().setAge(365))
```
breilly2: Maybe a year is too much, maybe it's too little. Seems there should be some limit on it, though. Other suggestions welcome.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +355,38 @@ class HttpGoogleServicesDAO(
     }
   }
 
+  override def getBucketUsage(projectName: RawlsBillingProjectName, bucketName: String, maxResults: Option[Long]): Future[BigInt] = {
+
+    def usageFromLogObject(o: StorageObject): Future[BigInt] = {
+      streamObject(o.getBucket, o.getName) { inputStream =>
+        val content = Source.fromInputStream(inputStream).mkString
+        val byteHours = BigInt(content.split('\n')(1).split(',')(1).replace("\"", ""))
```
breilly2: Ugly, but the CSV file is pretty small and it's straightforward to grab the 2nd column of the 2nd row.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -926,7 +1017,9 @@ class HttpGoogleServicesDAO(
       } match {
         case Success(response) =>
           logger.debug(GoogleRequest(request.getRequestMethod, request.buildHttpRequestUrl().toString, payload, System.currentTimeMillis() - start, Option(response.getStatusCode), None).toJson(GoogleRequestFormat).compactPrint)
-          response.parseAs(request.getResponseClass)
+          val result = processUnparsedResponse(response)
+          response.disconnect()
```
breilly2: We weren't previously doing this in `executeGoogleRequest`, but the JavaDoc for `AbstractGoogleClientRequest.executeUnparsed()` says it's up to the caller to call `disconnect()`.

src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
```diff
@@ -376,15 +377,88 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationT
...
+
+      val usage = Await.result(gcsDAO.getBucketUsage(projectName, bucketName, Some(1L)), Duration.Inf)
+
+      // 240 bytes per day, 2 days of logs, 24 hours per day
+      // 240 * 2 / 24 = 20
+      usage should be (20)
+
+    } finally {
+      try {
+        for (i <- 1 to 2) {
```
breilly2: The setup is ugly but this cleanup is even uglier. I'd rather just loop through a collection of object names to delete but I'm not sure of the right Scala idiom to build it while doing the setup. Suggestions welcome.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: a) stick it in configuration
b) this includes access logs right? I bet you FISMA has something to say about how long we should keep them, ask bernick.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +355,38 @@ class HttpGoogleServicesDAO(
...
+    def usageFromLogObject(o: StorageObject): Future[BigInt] = {
+      streamObject(o.getBucket, o.getName) { inputStream =>
+        val content = Source.fromInputStream(inputStream).mkString
+        val byteHours = BigInt(content.split('\n')(1).split(',')(1).replace("\"", ""))
+        // convert byte/hours to byte/days to better match the billing unit of GB/days
+        byteHours / 24
+      }
+    }
+
+    def recurse(pageToken: Option[String] = None): Future[BigInt] = {
```
breilly2: Given a max of one year of storage logs, this should not be necessary (since the default page size is 1000). However, I developed this paging code in the previous implementation of the feature and I don't see any reason to throw it away. Also, if we can do this, then can simply decide some day to keep 5 years worth of logs and not worry about breaking this feature.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +355,38 @@ class HttpGoogleServicesDAO(
...
+      val fetcher = getStorage(getBucketServiceAccountCredential).
+        objects().
+        list(getStorageLogsBucketName(projectName)).
+        setPrefix(bucketName + "_storage_")
+      maxResults.foreach(fetcher.setMaxResults(_))
+      pageToken.foreach(fetcher.setPageToken)
+
+      retryWhen500orGoogleError(() => {
+        val result = executeGoogleRequest(fetcher)
+        (result.getItems, result.getNextPageToken)
```
dvoet: wrap these 2 results in Option so you can match on None instead of null below, that is the scala way

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +355,38 @@ class HttpGoogleServicesDAO(
...
+      }) flatMap {
+        case (null, _) => Future.failed(new RawlsException(s"No storage logs available for $bucketName"))
```
dvoet: probably should make a specific exception class so we can handle this nicely up the stack

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -324,6 +355,38 @@ class HttpGoogleServicesDAO(
...
+        case (items, null) => usageFromLogObject(items.last)
+        case (_, nextPageToken) => recurse(Some(nextPageToken))
```
dvoet: In general you should always do Option(nextPageToken). In this particular case they have the same result. But by using Some directly you run the risk of generating an instance of Some(null).

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -935,7 +1033,7 @@ class HttpGoogleServicesDAO(
           throw regrets
       }
     } else {
-      request.execute()
+      executeRequest()
```
dvoet: We don't actually run in production with this case and passing in 2 different ways to execute the request seems confusing. I think we should remove this execution path.

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1510,6 +1512,19 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      bucketName <- dataSource.inTransaction { dataAccess =>
+        withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { workspaceContext =>
+          DBIO.successful(workspaceContext.workspace.bucketName)
+        }
+      }
+      usage <- gcsDAO.getBucketUsage(RawlsBillingProjectName(workspaceName.namespace), bucketName)
+    } yield {
+      RequestComplete(BucketUsageResponse(usage))
+    }
+  }
```
dvoet: I think we should handle the case of missing usage log separately (not as a 500 internal server error). you probably would handle that by calling recover on the Future resulting from the for comprehension. Maybe 204 no content?

src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1923,5 +1923,39 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
...
 
+  it should "return 404 when requesting bucket size for a non-existent workspace" in withTestWorkspacesApiServicesAndUser("reader-access") { services =>
+    Get(s"/workspaces/${testWorkspaces.workspace.namespace}/${testWorkspaces.workspace.name}x/bucketUsage") ~>
+      sealRoute(services.workspaceRoutes) ~>
+      check {
+        assertResult(StatusCodes.NotFound) { status }
+      }
+  }
+
+  // TODO: adjust user access for this test once the requirements of GAWB-1145 settle down
```
dvoet: can we do this todo now?

src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1923,5 +1923,39 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
...
+      Get(s"/workspaces/${testWorkspaces.workspace.namespace}/${testWorkspaces.workspace.name}/bucketUsage") ~>
+        sealRoute(services.workspaceRoutes) ~>
+        check {
+          assertResult(StatusCodes.OK) { status }
+          assertResult(BucketUsageResponse(BigInt(42))) {
+            responseAs[BucketUsageResponse]
+          }
+        }
+    }
+  }
```
dvoet: add tests for when size is not available

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: It looks like I'll have to modify rawls.conf.ctmpl in firecloud-develop. Developers will need to regenerate their config files or manually add a new parameter. Is this the right approach or is there a better way?

dvoet:
put it in application.conf (which may have been renamed to reference.conf
in the resources dir

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
breilly2: There are 2 cases I can think of when this might happen:
1. The workspace and bucket have just been created and therefore have no logs yet.
2. There's a bug in rawls or someone has manually interfered with the logging config or log objects.

For case #1, a 204 might be okay, as would a 200 with a result of 0.
For case #2, I think a 500 is entirely appropriate. 2xx doesn't seem right because that would indicate success. 4xx doesn't seem right because there isn't anything wrong with the client request. I didn't see any better 5xx codes, so I still think 500 covers this.

The trick then is being able to distinguish between #1 and #2. Since there are potentially no storage log files in both cases, my only thought is to immediately write a fake storage log when a workspace is created. Then, the lack of any log objects represents an unexpected server issue and, therefore, a 500 response.

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: I'm fine with removing one of the execution paths, but I'd like to keep the one that uses `request.executeUnparsed()` so that we keep the option for debug logging. I've found the logging to be extremely helpful during development. I'll make sure we continue to avoid serializing the payload if debug logging is disabled. Is that okay with you @dvoet?

dvoet: that's what I mean. Keep the debug path, drop the other.

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: writing a fake file is probably a good approach.

dvoet (approved):

dvoet:
👍

src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1923,5 +1923,39 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
...
+    Get(s"/workspaces/${testWorkspaces.workspace.namespace}/${testWorkspaces.workspace.name}x/bucketUsage") ~>
+      sealRoute(services.workspaceRoutes) ~>
+      check {
+        assertResult(StatusCodes.NotFound) { status }
+      }
+  }
+
+  // TODO: adjust user access for this test once the requirements of GAWB-1145 settle down
+  for (access <- Seq("no-access")) {
+    it should s"not allow a $access user to request bucket usage" in withTestWorkspacesApiServicesAndUser("no-access") { services =>
```
breilly2: Just found a bug here. `"no-access"` should be `access`.

MatthewBemis:
:+1: pending your test fix and rebase

PR #572, opened 2016-11-16
GAWB-1274: Implement pending workspace access w/ endpoints

MatthewBemis:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [ ] **Product Owner** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161115_pending_workspace_access.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161115_pending_workspace_access.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="pending-ws-access-1">
+        <sql stripComments="true">
+            create table PENDING_WORKSPACE_ACCESS (
+                workspace_id binary(16) not null,
+                user_email varchar(254) not null,
+                origin_subject_id varchar(254) not null,
+                invite_date timestamp not null,
+                access_level varchar(254) not null,
+                primary key PK_PENDING_WORKSPACE_ACCESS (workspace_id, user_email)
+            )
+        </sql>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="pending-ws-access-2">
+        <sql stripComments="true">
+            alter table PENDING_WORKSPACE_ACCESS add constraint FK_PENDING_WS_ACCESS_WORKSPACE foreign key (workspace_id) references WORKSPACE(ID);
+        </sql>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="pending-ws-access-3">
+        <sql stripComments="true">
+            alter table PENDING_WORKSPACE_ACCESS add constraint FK_PENDING_WS_ACCESS_ORIGIN_USER foreign key (origin_subject_id) references USER(SUBJECT_ID);
+        </sql>
+    </changeSet>
+    
+</databaseChangeLog>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -13,6 +13,7 @@ import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
 import org.broadinstitute.dsde.rawls.jobexec.SubmissionSupervisor.SubmissionStarted
 import org.broadinstitute.dsde.rawls.model.ProjectRoles.ProjectRole
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels.{ProjectOwner, WorkspaceAccessLevel}
+import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels._
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.expressions._
 import org.broadinstitute.dsde.rawls.user.UserService
@@ -55,7 +56,7 @@ object WorkspaceService {
   case class AdminListWorkspacesWithAttribute(attributeName: AttributeName, attributeValue: AttributeValue) extends WorkspaceServiceMessage
   case class CloneWorkspace(sourceWorkspace: WorkspaceName, destWorkspace: WorkspaceRequest) extends WorkspaceServiceMessage
   case class GetACL(workspaceName: WorkspaceName) extends WorkspaceServiceMessage
-  case class UpdateACL(workspaceName: WorkspaceName, aclUpdates: Seq[WorkspaceACLUpdate]) extends WorkspaceServiceMessage
+  case class UpdateACL(workspaceName: WorkspaceName, aclUpdates: Seq[WorkspaceACLUpdate], inviteUsersNotFound: Boolean) extends WorkspaceServiceMessage
   case class LockWorkspace(workspaceName: WorkspaceName) extends WorkspaceServiceMessage
   case class UnlockWorkspace(workspaceName: WorkspaceName) extends WorkspaceServiceMessage
   case class CheckBucketReadAccess(workspaceName: WorkspaceName) extends WorkspaceServiceMessage
@@ -129,7 +130,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     case AdminListWorkspacesWithAttribute(attributeName, attributeValue) => asFCAdmin { listWorkspacesWithAttribute(attributeName, attributeValue) } pipeTo sender
     case CloneWorkspace(sourceWorkspace, destWorkspaceRequest) => pipe(cloneWorkspace(sourceWorkspace, destWorkspaceRequest)) to sender
     case GetACL(workspaceName) => pipe(getACL(workspaceName)) to sender
-    case UpdateACL(workspaceName, aclUpdates) => pipe(updateACL(workspaceName, aclUpdates)) to sender
+    case UpdateACL(workspaceName, aclUpdates, inviteUsersNotFound) => pipe(updateACL(workspaceName, aclUpdates, inviteUsersNotFound)) to sender
     case LockWorkspace(workspaceName: WorkspaceName) => pipe(lockWorkspace(workspaceName)) to sender
     case UnlockWorkspace(workspaceName: WorkspaceName) => pipe(unlockWorkspace(workspaceName)) to sender
     case CheckBucketReadAccess(workspaceName: WorkspaceName) => pipe(checkBucketReadAccess(workspaceName)) to sender
@@ -288,6 +289,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
         _ <- DBIO.seq(dataAccess.workspaceQuery.deleteWorkspaceSubmissions(workspaceContext.workspaceId))
         _ <- DBIO.seq(dataAccess.workspaceQuery.deleteWorkspaceMethodConfigs(workspaceContext.workspaceId))
         _ <- dataAccess.workspaceQuery.deleteWorkspaceEntitiesAndAttributes(workspaceContext.workspaceId)
+        _ <- dataAccess.workspaceQuery.deleteWorkspaceInvites(workspaceContext.workspaceId)
 
         // Delete groups
         _ <- DBIO.seq(groupRefsToRemove.map { group => dataAccess.rawlsGroupQuery.delete(group) }.toSeq: _*)
@@ -434,10 +436,15 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     dataSource.inTransaction { dataAccess =>
       withWorkspaceContext(workspaceName, dataAccess) { workspaceContext =>
         requireAccessIgnoreLock(workspaceContext.workspace, WorkspaceAccessLevels.Owner, dataAccess) {
-          dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext).map { emailsAndAccess =>
-            // toMap below will drop duplicate keys, keeping the last entry only
-            // sort by access level to make sure higher access levels remain in the resulting map
-            RequestComplete(StatusCodes.OK, emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.toMap)
+          dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext).flatMap { emailsAndAccess =>
+            dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId).map { invites =>
+              // toMap below will drop duplicate keys, keeping the last entry only
+              // sort by access level to make sure higher access levels remain in the resulting map
+              val granted = emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, false)}
+              val pending = invites.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, true)}
+
+              RequestComplete(StatusCodes.OK, WorkspaceACL((granted ++ pending).toMap))
+            }
           }
         }
       }
@@ -450,7 +457,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
    *                   Seq, use NoAccess to remove an entry, all other preexisting accesses remain unchanged
    * @return
    */
-  def updateACL(workspaceName: WorkspaceName, aclUpdates: Seq[WorkspaceACLUpdate]): Future[PerRequestMessage] = {
+  def updateACL(workspaceName: WorkspaceName, aclUpdates: Seq[WorkspaceACLUpdate], inviteUsersNotFound: Boolean): Future[PerRequestMessage] = {
 
     import org.broadinstitute.dsde.rawls.model.WorkspaceACLJsonSupport._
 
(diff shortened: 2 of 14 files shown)


src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161115_pending_workspace_access.xml
```diff
@@ -0,0 +1,20 @@
...
+        <sql stripComments="true">
+            create table PENDING_WORKSPACE_ACCESS (
+                workspace_id binary(16) not null,
+                user_email varchar(254) not null,
+                origin_subject_id varchar(254) not null,
+                invite_date timestamp not null,
+                access_level varchar(254) not null,
+                primary key PK_PENDING_WORKSPACE_ACCESS (workspace_id, user_email),
+                foreign key FK_PENDING_WS_ACCESS_WORKSPACE(workspace_id) references WORKSPACE(id),
+                foreign key FK_PENDING_WS_ACCESS_ORIGIN_USER(origin_subject_id) references USER(SUBJECT_ID)
```
MatthewBemis: note to self: fix this create statement so mysql doesn't ignore the foreign key names in favor of auto-generating them

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -435,10 +440,16 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-            // sort by access level to make sure higher access levels remain in the resulting map
-            RequestComplete(StatusCodes.OK, emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.toMap)
+          dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext).flatMap { emailsAndAccess =>
+            dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId).map { invites =>
+              // toMap below will drop duplicate keys, keeping the last entry only
+              // sort by access level to make sure higher access levels remain in the resulting map
+              RequestComplete(
+                StatusCodes.OK,
+                WorkspaceACL(emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.toMap,
+                invites.sortBy { case (_, accessLevel) => accessLevel }.toMap)
```
MatthewBemis: this response now looks like:

```
{
  "granted": {
    "realuser1@gmail.com": "WRITER",
    "realuser2@gmail.com": "READER",
    "realuser3gmail.com": "OWNER",
    "realuser4@gmail.com": "READER",
    "realuser5@gmail.com": "OWNER",
    "GROUP_PROJECT_broad-dsde-dev-Owner@dev.test.firecloud.org": "PROJECT_OWNER"
  },
  "pending": {
    "fake-user-foo@example.com": "WRITER",
    "fake-user@example.com": "OWNER"
  }
}
```

src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -470,21 +481,14 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-            //we will process the emails that are valid, and report any others as invalid
-            RequestComplete( ErrorReport(StatusCodes.NotFound, s"Couldn't find some users/groups by email", emailNotFoundReports) )
-          }
+          val changesMade = actualChangesToMake.map { case (member, accessLevel) =>
+            member match {
+              case Left(userRef) => WorkspaceACLUpdateResponse(userRef.userSubjectId.value, accessLevel)
+              case Right(groupRef) => WorkspaceACLUpdateResponse(groupRef.groupName.value, accessLevel)
+            }
+          }.toSeq
+          RequestComplete(StatusCodes.OK, WorkspaceACLUpdateResponseList(changesMade, emailsNotFound))
```
MatthewBemis: this response now looks like:

```
{
  "usersUpdated": [
    {
      "subjectId": "110707025587602912910",
      "accessLevel": "READER"
    },
    {
      "subjectId": "110347448408766049948",
      "accessLevel": "READER"
    }
  ],
  "usersNotFound": [
    {
      "email": "fake-user@example.com",
      "accessLevel": "OWNER"
    }
  ]
}
```

src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -99,6 +99,26 @@ trait WorkspaceApiService extends HttpService with PerRequestCreator with UserIn
             WorkspaceService.UnlockWorkspace(WorkspaceName(workspaceNamespace, workspaceName)))
         }
       }
+    } ~
+    path("workspaces" / Segment / Segment / "invites" ) { (workspaceNamespace, workspaceName) =>
+      post {
+        requireUserInfo() { userInfo =>
+          entity(as[WorkspaceACLUpdate]) { invite =>
```
MatthewBemis: i was a little undecided on what these two endpoints should look like. three options i see are:

1) keep them as is
2) keep them almost as is, but make them take batches of invites to create/delete
3) make it similar to patch ACL where there is one endpoint and invites can be deleted by setting the invite's accessLevel to `NO ACCESS`

i thought of the 3rd option after the fact and think it would be nice to have that consistency but i'm not sure how necessary it is. i also need to add tests to WorkspaceApiServiceSpec but will do so after this question is answered

MatthewBemis: update: i'm going to go with #3 because per the http spec, delete technically shouldn't be allowed to have a json body

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -171,6 +195,25 @@ trait WorkspaceComponent {
     def unlock(workspaceName: WorkspaceName): ReadWriteAction[Int] = {
       findByNameQuery(workspaceName).map(_.isLocked).update(false)
     }
+
+    //use marsal method
```
dvoet: spelling... but what does this comment mean?

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -171,6 +195,25 @@ trait WorkspaceComponent {
...
+    def saveInvite(workspaceId: UUID, userEmail: String, originUser: String, accessLevel: WorkspaceAccessLevel): ReadWriteAction[Boolean] = {
```
dvoet: save usually returns the thing it saved

src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceACL.scala
```diff
@@ -4,10 +4,14 @@ import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels.WorkspaceAccessLevel
 import spray.json._
 
-case class WorkspaceACL(acl: Map[String, WorkspaceAccessLevel])
+case class WorkspaceACL(granted: Map[String, WorkspaceAccessLevel], pending: Map[String, WorkspaceAccessLevel])
```
dvoet: would it be better like

```
case class AccessEntry(level: WorkspaceAccessLevel, pending: Boolean)
case class WorkspaceACL(acl: Map[String, AccessEntry])
```

src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: ignore this. i thought i deleted this comment because i addressed the thing that i was referring to (it was a todo)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -463,31 +469,49 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
     }
 
+    def determineInvites(emailsNotFound: Seq[WorkspaceACLUpdate], inviteUsersNotFound: Boolean): Future[Seq[WorkspaceACLUpdate]] = {
+      dataSource.inTransaction { dataAccess =>
+        withWorkspaceContext(workspaceName, dataAccess) { workspaceContext =>
+          dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId) flatMap { invites =>
+            val (toDelete, toSave) = emailsNotFound.partition(_.accessLevel.equals(WorkspaceAccessLevels.NoAccess))
```
dvoet: just a matter style: I don't really like the partition pattern that I see every now and then. I find it more readable and more compact to do a single map:
```
DBIO.seq(emailsNotFound.map { 
  case (email, WorkspaceAccessLevels.NoAccess) => dataAccess.workspaceQuery.removeInvite(workspaceContext.workspaceId, email)
  case invite => dataAccess.workspaceQuery.saveInvite(workspaceContext.workspaceId, userInfo.userSubjectId, invite)
})
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -463,31 +469,49 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      }
+    }
+
+    def getUsersUpdatedResponse(actualChangesToMake: Map[Either[RawlsUserRef, RawlsGroupRef], WorkspaceAccessLevel]): Seq[WorkspaceACLUpdateResponse] = {
+      actualChangesToMake.map { case (member, accessLevel) =>
+        member match {
+          case Left(userRef) => WorkspaceACLUpdateResponse(userRef.userSubjectId.value, accessLevel)
+          case Right(groupRef) => WorkspaceACLUpdateResponse(groupRef.groupName.value, accessLevel)
+        }
+      }.toSeq
```
dvoet, line 507: you don't need a double nested match
```
actualChangesToMake.map {
  case (Left(userRef), accessLevel)) => WorkspaceACLUpdateResponse(userRef.userSubjectId.value, accessLevel)
  case (Right(groupRef), accessLevel)) => WorkspaceACLUpdateResponse(groupRef.groupName.value, accessLevel)
}.toSeq
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -463,31 +469,49 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+          dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId) flatMap { invites =>
+            val (toDelete, toSave) = emailsNotFound.partition(_.accessLevel.equals(WorkspaceAccessLevels.NoAccess))
+            val dedupedInvites = toSave.filterNot(update => invites.contains((update.email, update.accessLevel)))
+
+            if(inviteUsersNotFound) {
+              DBIO.sequence(dedupedInvites.map { invite =>
+                dataAccess.workspaceQuery.saveInvite(workspaceContext.workspaceId, userInfo.userSubjectId, WorkspaceACLUpdate(invite.email, invite.accessLevel))
+              }) flatMap { _ =>
+                DBIO.sequence(toDelete.map { invite =>
+                  dataAccess.workspaceQuery.removeInvite(workspaceContext.workspaceId, invite.email)
```
dvoet: should't we remove invites regardless of inviteUsersNotFound?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -463,31 +469,49 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-            RequestComplete(StatusCodes.OK, changesMade)
-          } else {
-            //this is a case where we don't want to rollback the transaction in the event of getting an error.
-            //we will process the emails that are valid, and report any others as invalid
-            RequestComplete( ErrorReport(StatusCodes.NotFound, s"Couldn't find some users/groups by email", emailNotFoundReports) )
-          }
-
+          val response = if(inviteUsersNotFound) {
+            WorkspaceACLUpdateResponseList(getUsersUpdatedResponse(actualChangesToMake), emailInvites, Seq.empty)
+          } else WorkspaceACLUpdateResponseList(getUsersUpdatedResponse(actualChangesToMake), Seq.empty, emailInvites)
```
dvoet: I don't like that the logic around inviteUsersNotFound is split up. It is used in determineInvites which actually does the invites then again here which figures out what the return value of determineInvites actually means.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -506,12 +530,12 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
    * @return tuple: messages to send to UserService to overwrite acl groups, email that were not found in the process
    */
-  private def determineCompleteNewAcls(aclUpdates: Seq[WorkspaceACLUpdate], dataAccess: DataAccess, workspaceContext: SlickWorkspaceContext): ReadAction[(Iterable[OverwriteGroupMembers], Seq[String], Map[Either[RawlsUserRef,RawlsGroupRef], WorkspaceAccessLevels.WorkspaceAccessLevel])] = {
+  private def determineCompleteNewAcls(aclUpdates: Seq[WorkspaceACLUpdate], dataAccess: DataAccess, workspaceContext: SlickWorkspaceContext): ReadAction[(Iterable[OverwriteGroupMembers], Seq[WorkspaceACLUpdate], Map[Either[RawlsUserRef,RawlsGroupRef], WorkspaceAccessLevels.WorkspaceAccessLevel])] = {
     for {
       refsToUpdateByEmail <- dataAccess.rawlsGroupQuery.loadRefsFromEmails(aclUpdates.map(_.email))
       existingRefsAndLevels <- dataAccess.workspaceQuery.findWorkspaceUsersByAccessLevel(workspaceContext.workspaceId)
     } yield {
-      val emailsNotFound = aclUpdates.map(_.email).diff(refsToUpdateByEmail.keySet.toSeq)
+      val emailsNotFound = aclUpdates.filterNot(aclChange => refsToUpdateByEmail.keySet.contains(aclChange.email))
```
dvoet, line 552: why this change?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -359,89 +359,145 @@ class WorkspaceServiceSpec extends FlatSpec with ScalatestRouteTest with Matcher
...
+    }
+  }
+
+  it should "invite a user to a workspace" in withTestDataServices { services =>
+    val vComplete = Await.result(services.workspaceService.updateACL(testData.workspace.toWorkspaceName, Seq(WorkspaceACLUpdate("obama@whitehouse.gov", WorkspaceAccessLevels.Owner)), true), Duration.Inf)
+      .asInstanceOf[RequestComplete[(StatusCode, WorkspaceACLUpdateResponseList)]]
+
+    assertResult(StatusCodes.OK, "Invite user shouldn't error") {
+      vComplete.response._1
+    }
```
dvoet, line 522: check return value of WorkspaceACLUpdateResponseList

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -463,31 +469,49 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      actualChangesToMake.map { case (member, accessLevel) =>
+        member match {
+          case Left(userRef) => WorkspaceACLUpdateResponse(userRef.userSubjectId.value, accessLevel)
+          case Right(groupRef) => WorkspaceACLUpdateResponse(groupRef.groupName.value, accessLevel)
+        }
+      }.toSeq
+    }
+
     val userServiceRef = context.actorOf(UserService.props(userServiceConstructor, userInfo))
     for {
```
dvoet, line 515: I think this for comprehension needs a little work. The steps should be

1. overwriteGroupMessagesFuture (no change)
1. Future.traverse(overwriteGroupMessages) (no change)
1. remove appropriate invites
1. maybe invite users which returns the actual list of invites sent

All the while progressively filtering emailsNotFound (i.e. removing invites also removes some unknown emails and inviting will remove any outstanding emails and maybe all the rest). At the end you have a list of emails not found and invites sent and you don't need to do a further check of inviteUsersNotFound.

dvoet (changes requested):

dvoet:
@MatthewBemis look at https://github.com/broadinstitute/rawls/pull/572#discussion_r89335477 first, it supersedes some of the other comments

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: yes that's a good catch, the planned UI flow would break in this case

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: I need to return the accessLevel along with the email. Before it was just the email

dvoet (approved):

dvoet:
👍

ansingh7115:
👍

dvoet:
👍

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161115_pending_workspace_access.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="pending-ws-access-1">
+        <sql stripComments="true">
```
breilly2, line 5: Any reason to write the DDL by hand instead of using a `createTable` tag? http://www.liquibase.org/documentation/changes/create_table.html You should be able to include the foreign key constraints there and reduce this to one `changeSet`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -497,6 +500,37 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  private def getExistingWorkspaceInvites(workspaceName: WorkspaceName) = {
```
breilly2: `getExistingWorkspaceInvites`, `deleteWorkspaceInvites`, and `saveWorkspaceInvites` are all only called from inside `updateACL`. Other cases like this have the functions defined inside `updateACL` instead of being private on `WorkspaceService`. Is there a reason for these to be different or could they be changed to use consistent visibility? (I'm not sure which is preferred.)

MatthewBemis: there is no reason for these to be different. I'll update

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20161115_pending_workspace_access.xml
MatthewBemis: i got into the habit of writing straight SQL when i wrote a few large migrations in the past. the reasons i had were:

1) it's easier to reason about the changes when they're written in the same language that they're going to be executed in
2) a lot of complex operations can't be expressed using the liquibase definitions
3) sometimes the sql generated by liquibase is slightly different than the sql that i want (we have this problem with slick)

now in the case of a simple create table- i agree that none of those things really matter at all. i wrote it this way because that's how i normally do it but that doesn't mean it's the right thing to do here :) i'm happy to change it if you'd like

breilly2: No need to change if it's generally accepted to do it this way. I think the only reason to prefer using the Liquibase model is to minimize the amount of rewriting if we switched to a different database engine. You do still have to be careful, but Liquibase can do some of the translation to different dialects.

In this case, I think the biggest gain (which could probably also be done with straight SQL) would be pulling the foreign key constraints into a the `create table` statement so that it's all part of one atomic change set. May not be worth changing at this point (it's a pain to test) but something to think about for next time.

It might be worth a larger discussion about which style we want to prefer.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -497,6 +500,37 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+          pairs.map(pair => WorkspaceACLUpdate(pair._1, pair._2))
+        }
+      }
+    }
+  }
+
+  private def deleteWorkspaceInvites(invites: Seq[WorkspaceACLUpdate], existingInvites: Seq[WorkspaceACLUpdate], workspaceName: WorkspaceName) = {
+    dataSource.inTransaction { dataAccess =>
+      withWorkspaceContext(workspaceName, dataAccess) { workspaceContext =>
+        val removals = invites.filter(_.accessLevel.equals(WorkspaceAccessLevels.NoAccess))
```
breilly2: IntelliJ is showing a warning for `_.accessLevel.equals(WorkspaceAccessLevels.NoAccess)`:

```
Comparing unrelated types less... (⌘F1) 
Detects comparisons (== and !=) of expressions which cannot be of the same type.
```
Manual inspection looks fine to me, but maybe there's something subtle going on here that I'm missing.

Interestingly, `_.accessLevel == WorkspaceAccessLevels.NoAccess` eliminates the warning. Coming from a Java background, this feels opposite to me, but I think (not 100% positive) that that equality expression will do the right thing in this case.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -434,10 +436,15 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-            // sort by access level to make sure higher access levels remain in the resulting map
-            RequestComplete(StatusCodes.OK, emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.toMap)
+          dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext).flatMap { emailsAndAccess =>
+            dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId).map { invites =>
+              // toMap below will drop duplicate keys, keeping the last entry only
+              // sort by access level to make sure higher access levels remain in the resulting map
+              val granted = emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, false)}
+              val pending = invites.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, true)}
+
+              RequestComplete(StatusCodes.OK, WorkspaceACL((granted ++ pending).toMap))
```
breilly2, line 446: This looks like a change in return type without a corresponding change in rawls.yaml. However, it's not obvious to me that rawls.yaml is incorrect. I'm not positive that anything needs to be done there, but I think it's worth a double-check.

MatthewBemis: Good eye, there was a change to be made here

MatthewBemis:
@breilly2 the latest commit addresses the 3 comments you made.

also note that I merged the other PR (https://github.com/broadinstitute/rawls/pull/607) into this one since it was approved

breilly2 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -506,7 +506,9 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
         case (Right(groupRef), accessLevel) => WorkspaceACLUpdateResponse(groupRef.groupName.value, accessLevel)
       }.toSeq
 
-      WorkspaceACLUpdateResponseList(usersUpdated, invitesUpdated, (emailsNotFound diff invitesUpdated diff existingInvites))
+      val usersNotFound = emailsNotFound.filterNot(aclUpdate => invitesUpdated.map(_.email).contains(aclUpdate.email)).diff(existingInvites)
```
breilly2: This is subtle. It would be great to have some tests for this.

dvoet:
👍

breilly2 (approved):
:+1:

PR #576, opened 2016-11-21
Use scratch tables only for update part of upsert attributes

jmthibault79:
GAWB-958

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [ ] **Product Owner** sign off
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [x] **Submitter**: Verify swagger UI on dev environment still works after deployment
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -5,7 +5,7 @@ import java.util.UUID
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model._
-import slick.ast.TypedType
+import slick.ast.{BaseTypedType, TypedType}
 import slick.dbio.Effect.Write
 import slick.driver.JdbcDriver
 import spray.http.StatusCodes
@@ -53,8 +53,7 @@ case class EntityAttributeScratchRecord(id: Long,
                                      valueEntityRef: Option[Long],
                                      listIndex: Option[Int],
                                      listLength: Option[Int],
-                                     transactionId: String) extends AttributeScratchRecord[Long] {
-}
+                                     transactionId: String) extends AttributeScratchRecord[Long]
 
 case class WorkspaceAttributeRecord(id: Long,
                                     ownerId: UUID, // workspace id
@@ -158,9 +157,7 @@ trait AttributeComponent {
   }
 
   protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord)
-
   protected object workspaceAttributeQuery extends AttributeQuery[UUID, WorkspaceAttributeRecord, WorkspaceAttributeTable](new WorkspaceAttributeTable(_), WorkspaceAttributeRecord)
-
   protected object submissionAttributeQuery extends AttributeQuery[Long, SubmissionAttributeRecord, SubmissionAttributeTable](new SubmissionAttributeTable(_), SubmissionAttributeRecord)
 
   protected abstract class AttributeScratchQuery[OWNER_ID: TypeTag, RECORD <: AttributeRecord[OWNER_ID], TEMP_RECORD <: AttributeScratchRecord[OWNER_ID], T <: AttributeScratchTable[OWNER_ID, TEMP_RECORD]](cons: Tag => T, createRecord: (Long, OWNER_ID, String, String, Option[String], Option[Double], Option[Boolean], Option[Long], Option[Int], Option[Int], String) => TEMP_RECORD) extends TableQuery[T](cons) {
@@ -179,7 +176,7 @@ trait AttributeComponent {
    * @tparam OWNER_ID the type of the ownerId field
    * @tparam RECORD the record class
    */
-  protected abstract class AttributeQuery[OWNER_ID: TypeTag, RECORD <: AttributeRecord[OWNER_ID], T <: AttributeTable[OWNER_ID, RECORD]](cons: Tag => T, createRecord: (Long, OWNER_ID, String, String, Option[String], Option[Double], Option[Boolean], Option[Long], Option[Int], Option[Int]) => RECORD) extends TableQuery[T](cons)  {
+  protected abstract class AttributeQuery[OWNER_ID: TypeTag: BaseTypedType, RECORD <: AttributeRecord[OWNER_ID], T <: AttributeTable[OWNER_ID, RECORD]](cons: Tag => T, createRecord: (Long, OWNER_ID, String, String, Option[String], Option[Double], Option[Boolean], Option[Long], Option[Int], Option[Int]) => RECORD) extends TableQuery[T](cons)  {
 
     def marshalAttribute(ownerId: OWNER_ID, attributeName: AttributeName, attribute: Attribute, entityIdsByRef: Map[AttributeEntityReference, Long]): Seq[T#TableElementType] = {
 
@@ -245,6 +242,10 @@ trait AttributeComponent {
       filter(rec => rec.namespace === attrName.namespace && rec.name === attrName.name)
     }
 
+    def findByOwnerQuery(ownerIds: Seq[OWNER_ID]) = {
+      filter(_.ownerId inSetBind ownerIds)
+    }
+
     def queryByAttribute(attrName: AttributeName, attrValue: AttributeValue) = {
       findByNameQuery(attrName).filter { rec =>
         attrValue match {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -26,10 +26,29 @@ trait AttributeRecord[OWNER_ID] {
   val valueEntityRef: Option[Long]
   val listIndex: Option[Int]
   val listLength: Option[Int]
+
+  def toComparable: ComparableAttributeRecord[OWNER_ID]
 }
 
 trait AttributeScratchRecord[OWNER_ID] extends AttributeRecord[OWNER_ID] {
   val transactionId: String
+  override def toComparable = throw new RawlsException("There is no equivalent comparable type for AttributeScratchRecords")
```
jmthibault79: Alternative: a trait which extends from only AttributeRecords that we will want to compare, so we don't have to do this

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -157,9 +185,19 @@ trait AttributeComponent {
     def * = (id, ownerId, namespace, name, valueString, valueNumber, valueBoolean, valueEntityRef, listIndex, listLength, transactionId) <>(WorkspaceAttributeScratchRecord.tupled, WorkspaceAttributeScratchRecord.unapply)
   }
 
-  protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord)
+  protected object entityAttributeQuery extends AttributeQuery[Long, EntityAttributeRecord, EntityAttributeTable](new EntityAttributeTable(_), EntityAttributeRecord) {
+    // TODO make generic
```
jmthibault79: It's not easy to push this up to `AttributeQuery` because `inSetBind` uses magic implicits.  I wasn't able to figure it out.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -312,20 +350,6 @@ trait AttributeComponent {
 
     object AlterAttributesUsingScratchTableQueries extends RawSqlQuery {
       val driver: JdbcDriver = AttributeComponent.this.driver
-      /*!CAUTION! - do not drop the ownerIds clause at the end or it will delete every attribute in the table!
```
jmthibault79: No more owner IDs!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -312,20 +350,6 @@ trait AttributeComponent {
...
-
-        deleteMasterFromAction: deletes any row in *_ATTRIBUTE that doesn't
-          exist in *_ATTRIBUTE_TEMP but exists in *_ATTRIBUTE (bound by ownerIds)
-        updateInMasterAction: updates any row in *_ATTRIBUTE that also
-          exists in *_ATTRIBUTE_TEMP (bound by ownerIds)
-        insertIntoMasterAction: inserts any row into *_ATTRIBUTE that doesn't
-          exist in *_ATTRIBUTE but exists in *_ATTRIBUTE_TEMP (bound by ownerIds)
-
-        These left joins give us the rows related to the attributes that are in the temp table.
```
jmthibault79: No more left joins!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -272,17 +272,18 @@ trait EntityComponent {
       entities.foreach(validateEntity)
 
       for {
-        preExistingEntityRecs <- lookupEntitiesByNames(workspaceContext.workspaceId, entities.map(_.toReference)).map(updateEntityRecords(_, entities))
```
jmthibault79, line 275: several small refactorings with individual commits

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -337,38 +361,42 @@ trait AttributeComponent {
...
     }
 
+    def toComparableMap(recs: Traversable[RECORD]) =
+      recs.map { rec => (rec.toComparable, rec) }.toMap
+
+    def upsertAction(attributesToSave: Traversable[RECORD], existingAttributes: Traversable[RECORD], insertFunction: Seq[RECORD] => String => ReadWriteAction[Unit]) = {
+      val toSaveAttrMap = toComparableMap(attributesToSave)
+      val existingAttrMap = toComparableMap(existingAttributes)
+
+      val (attrsToUpdateMap, attrsToInsertMap) = toSaveAttrMap.partition { case (k, v) => existingAttrMap.keySet.contains(k) }
```
helgridly: as discussed offline, i think you can sidestep the whole comparable business:
```
val (attrsToUpdate, attrsToInsert) = attributesToSave partition { attr =>
    existingAttributes.exists( exAttr => exAttr.id == attr.id )
}
```
with the guts of that `exists` pulled out into a function if necessary.

helgridly:
fuckin' A. team. 👍

jmthibault79:
~I have not yet confirmed that the non-set version is performant.~
The non-set version is too slow, so I reverted it.

jmthibault79:
The cleaner version "Remove the Comparable structures" consistently times out (after 1 min) while the Comparable version takes ~ 45 sec.  Not great, but tolerable.  I'll attempt to optimize.

jmthibault79:
New version with better-contained Comparables, now called PrimaryKeys

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -337,38 +332,66 @@ trait AttributeComponent {
...
-          deleteFromMasterAction(ownerIds, transactionId) andThen
-          updateInMasterAction(ownerIds, transactionId) andThen
-          insertIntoMasterAction(ownerIds, transactionId) andFinally
+        insertIntoScratchFunction(transactionId) andThen
+          updateInMasterAction(transactionId) andThen
           clearAttributeScratchTableAction(transactionId)
       }
     }
 
+    // enables set operations on attribute primary key only
```
helgridly: can you add a bit more commentary with context for the actual problem here? i (sadly) agree that this is necessary but it's esoteric and difficult to wrap your head around.

e.g. by "set operations on primary key" you mean you want to determine whether to upsert or delete an attribute based on its db index, whereas the normal behaviour of `filter` uses all the fields, including values that might be modified. these PK classes are a cut-down definition of the records that contain only the fields needed to identify an attribute record that needs updating.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -337,38 +332,66 @@ trait AttributeComponent {
...
+
+    trait AttributeRecordPrimaryKey[ID_TYPE] {    // can't call it OWNER_ID because that is already in scope
+      val ownerId: ID_TYPE
+      val namespace: String
+      val name: String
+      val listIndex: Option[Int]
```
helgridly: what about listLength? (haven't thought about it, maybe you have. just thinking about the weird empty-list edge case)

helgridly:
👍 pending adding more comments + doublechecking empty-lists.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
jmthibault79: I'm using the primary key which does not include listLength.  Whether it should, I don't know.

ansingh7115:
👍 with tests

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
dvoet: try adding BaseTypedType as a type class of OWNER_ID in the definition of AttributeQuery:
```protected abstract class AttributeQuery[OWNER_ID: TypeTag : **BaseTypedType**, RECORD <: AttributeRecord[OWNER_ID], T <: AttributeTable[OWNER_ID, RECORD]](cons: Tag => T, createRecord: (Long, OWNER_ID, String, String, Option[String], Option[Double], Option[Boolean], Option[Long], Option[Int], Option[Int]) => RECORD) extends TableQuery[T](cons) {```

Comment (helgridly) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala]: what the heck is BaseTypedType? google isn't turning up much

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala [hunk 9]
```diff
@@ -310,22 +319,47 @@ trait AttributeComponent {
...
+    // Delete and Insert use standard Slick actions, but Update uses a scratch table
+    // (AlterAttributesUsingScratchTableQueries) for DB performance reasons.
+
+    // we use set operations (filter, contains, partition) to split the attributes,
+    // which uses the element type's equals method.  Ideally we would use AttributeRecord.equals,
+    // but we only want to use a subset of the fields: the primary key.
+    // AttributeRecordPrimaryKey encapsulates only those fields.
+
+    case class AttributeRecordPrimaryKey(ownerId: OWNER_ID, namespace: String, name: String, listIndex: Option[Int]) {
+      override def equals(o: Any) = o match {
```
dvoet: don't case classes get equals and hashCode for free? Also, just summing the constituent hash codes does not create a good hash code as a result. That probably does not really matter here but is good to know. Intellij can generate a good hashcode function for your (Code -> Generate...) or see http://stackoverflow.com/questions/2738886/what-is-a-best-practice-of-writing-hash-function-in-java

dvoet: I take that "does not really matter" comment back. You are using these as keys in a Map and there likely to be a lot of them so a good hashCode function probably matters quite a bit. Back in my java days I would say you should also be intelligently sizing your map... don't know if that matters in scala or not. Perhaps profiling this for a large upload is a good idea.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -294,21 +295,20 @@ trait EntityComponent {
...
-          attributeRec <- entityAttributeQuery.marshalAttribute(entityIdsByName(entity.toReference), attributeName, attribute, entityIdsByName)
-        } yield attributeRec
-
-        entityAttributeScratchQuery.batchInsertAttributes(attributeRecsToEntityId.toSeq, transactionId)
+      def insertScratchAttributes(attributeRecs: Seq[EntityAttributeRecord])(transactionId: String): ReadWriteAction[Unit] = {
+        entityAttributeScratchQuery.batchInsertAttributes(attributeRecs, transactionId)
       }
 
-      entityAttributeQuery.AlterAttributesUsingScratchTableQueries.upsertAction(entityIds, insertScratchAttributes)
+      entityAttributeQuery.findByOwnerQuery(entityIds) flatMap { existingAttributes =>
```
dvoet: this seems like a spot where repeatable read isolation level might be important. We have already read these attributes and are reading them again here. If they are different the second time the upsert will be incorrect. However, does our optimistic locking protect against this? The upsert may be wrong but the transaction will roll back. But then again, looking at our code, the first time we read the data (WorkspaceService.scala:653) we are not grabbing the entity version so the optimistic lock is not comprehensive enough.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
dvoet: TypedType is the thing slick uses to understand how to convert scala types into database speak. But there is a further extension called BaseTypedType which is more special (and required) in some way.

I figured this out by following the implicits (inSetBind is an implicit function, I found where it is and tried calling it directly). I eventually came up with the compile error "could not find implicit value for evidence parameter of type slick.ast.BaseTypedType[OWNER_ID]"

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
jmthibault79: Removed my implementation in favor of the Scala built-in.  I'll profile hash code and map sizing but not as part of this PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
jmthibault79: Excellent.  That worked!

jmthibault79:
@ansingh7115 added a test

ansingh7115:
👍

PR #577, opened 2016-11-21
Fix sbt reStart

dmohs:


jmthibault79:
This solved my Dev deploy problem.  💯

MatthewBemis:
:+1: from me, which is based on joel's testimonial and your explanation from earlier. otherwise, i don't know enough about sbt to understand why this actually works

dmohs:
Doug gave me a verbal thumb.

PR #586, opened 2016-11-29
GAWB-1259 Added ability to publish rawlsModel to Artifactory

helgridly:
Assuming that `ARTIFACTORY_USERNAME` and `ARTIFACTORY_PASSWORD` are in your env, this command line will publish a rawlsModel jar to artifactory:

```
sbt -mem 4096 assembly rawlsModel:publish -Dproject.isSnapshot=true
```

Jenkins should run this every build (will sync up with devops to do this). When we want to publish an "official" release of rawlsModel we can run this manually, flipping `isSnapshot` to `false`.

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off


helgridly:
pinging @kshakir and @geoffjentry for a once-over too

project/Publishing.scala
```diff
@@ -0,0 +1,35 @@
...
+  private val artifactoryCredentials: Credentials = {
+    val username = sys.env.getOrElse("ARTIFACTORY_USERNAME", "")
+    val password = sys.env.getOrElse("ARTIFACTORY_PASSWORD", "")
+    Credentials("Artifactory Realm", "artifactory.broadinstitute.org", username, password)
+  }
+
+  val modelPublishSettings: Seq[Setting[_]] =
+    //we only publish to libs-release-local because of a bug in sbt that makes snapshots take
+    //priority over the local package cache. see here: https://github.com/sbt/sbt/issues/2687#issuecomment-236586241
+    Seq(publishTo := Option(artifactoryResolver(false)), credentials += artifactoryCredentials)
```
helgridly, line 28: slight difference to cromwell here in passing `false` rather than the sbt `isSnapshot` key, which doesn't work right

project/Publishing.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  val noPublishSettings: Seq[Setting[_]] =
+    Seq(
+      publish := {},
+      publishLocal := {}
+    )
+}
```
helgridly, line 35: we only want to publish rawlsModel, so i added this to the other projects for safekeeping in case someone does sbt publish (which they shouldn't)

project/Version.scala
```diff
@@ -0,0 +1,20 @@
+import sbt.Keys._
+import sbt._
+
+object Version {
+  val baseModelVersion = "0.1"
+  val lastModelCommit = ("git log -n 1 --pretty=format:%h model" !!).trim
```
helgridly, line 6: this gets the last commit in the model directory so we don't publish a new identical rawlsModel jar just because there are changes to rawlsCore. the trim is necessary because the string return value includes the newline, which is cute

project/Version.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  val modelVersionSettings: Seq[Setting[_]] =
+    Seq( version := versionString )
+
+  val versionString = {
+    val version = baseModelVersion + "-" + lastModelCommit
+
+    // The project isSnapshot string passed in via command line settings, if desired.
+    val isSnapshot = sys.props.get("project.isSnapshot").forall(_.toBoolean)
```
helgridly: i couldn't wrap my head around this until i looked it up, so i'll save you the trouble: `forall` returns true for empty iterables.

jmthibault79: That's too cute for me.  `getOrElse("project.isSnapshot", "true").toBoolean` tells us explicitly what it's doing.

project/Settings.scala
```diff
@@ -43,23 +45,25 @@ object Settings {
...
+  ) ++ modelVersionSettings ++ modelPublishSettings
 
-  val rawlsSettings = List(
+  //commonSettings -> defaultSettings overrides name so we put it first
+  val rawlsSettings = commonSettings ++ List(
     name := "rawls-core",
     version := "0.1",
     libraryDependencies ++= rawlsDependencies
-  ) ++ commonSettings ++ rawlsAssemblySettings ++ rawlsCompileSettings
+  ) ++ rawlsAssemblySettings ++ noPublishSettings ++ rawlsCompileSettings ++ noPublishSettings
```
jmthibault79: `noPublishSettings` is here twice.  Is that correct?  `rawlsCompileSettings` is no longer last, so contradicts the comment below.

jmthibault79:
I want to see more comments in Settings.scala and renaming each `val rawlsFoo` to something more explicit.  It's not clear what you mean by Rawls vs. not-Rawls.

Also, the call to Defaults.defaultSettings on line 20 is deprecated.

project/Settings.scala
helgridly: Good catch, thanks

helgridly:
You mean the difference between rawlsModel and rawlsCore? rawlsModel contains the model and serialisation classes used by orch (and potentially other services). rawlsCore is everything else. I can comment saying that, and if you have suggestions for alternate names I'm all ears.

jmthibault79:
No, I mean assemblySettings vs. rawlsAssemblySettings and rawlsSettings vs. commonSettings vs. rootSettings.

rawlsModel and rawlsCore make sense to me.

helgridly:
Gotcha. Ta.

jmthibault79:
I like the new comments and renames.  Thanks.  Please also update the "commonSettings -> coreDefaultSettings" comments to something like "here we override the name in commonSettings so we need to include it first"

jmthibault79:
As we facetalked, `assembly` is not required in the command line, and removing it is a big speed improvement.

jmthibault79:
👍

PR #601, opened 2016-12-19
Add comment about billing scope

dmohs:
Someone could mistakenly believe that changing this would automatically cause a client to request the correct scopes.


dvoet (changes requested):
I don't think this is true. The error message details which scopes are required.

dmohs:
Discussed with Doug offline. Maybe the UI isn't the right place to know about all of the scopes we need, but it is currently the place we place this knowledge, so okay for now.

PR #603, opened 2017-01-03
GAWB-1439 handle missing storage logs

breilly2:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -116,7 +116,6 @@ class HttpGoogleServicesDAO(
   private def getBucketName(workspaceId: String) = s"${groupsPrefix}-${workspaceId}"
 
   override def setupWorkspace(userInfo: UserInfo, project: RawlsBillingProject, workspaceId: String, workspaceName: WorkspaceName, realm: Option[RawlsGroupRef], realmProjectOwnerIntersection: Option[Set[RawlsUserRef]]): Future[GoogleWorkspaceInfo] = {
-    val bucketName = getBucketName(workspaceId)
 
     // we do not make a special access group for project owners because the only member would be a single group
     // we will just use that group directly and avoid potential google problems with a group being in too many groups
@@ -160,7 +159,8 @@ class HttpGoogleServicesDAO(
       }
     }
 
-    def insertBucket: (Map[WorkspaceAccessLevel, RawlsGroup], Option[Map[WorkspaceAccessLevel, RawlsGroup]]) => Future[GoogleWorkspaceInfo] = { (accessGroupsByLevel, intersectionGroupsByLevel) =>
+    def insertBucket: (Map[WorkspaceAccessLevel, RawlsGroup], Option[Map[WorkspaceAccessLevel, RawlsGroup]]) => Future[String] = { (accessGroupsByLevel, intersectionGroupsByLevel) =>
+      val bucketName = getBucketName(workspaceId)
       retryWhen500orGoogleError {
         () => {
 
@@ -169,7 +169,7 @@ class HttpGoogleServicesDAO(
           val groupsByAccess = intersectionGroupsByLevel getOrElse accessGroupsByLevel
 
           // bucket ACLs should be:
-          //   project owner - bucker writer
+          //   project owner - bucket writer
           //   workspace owner - bucket writer
           //   workspace writer - bucket writer
           //   workspace reader - bucket reader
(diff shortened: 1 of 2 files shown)


dvoet:
should we stop writing an initial storage log?

breilly2:
I considered that, but decided that the initial storage log has value. At the time that the workspace bucket is created, we know that there is nothing in it so we can declare the size to be zero. In my mind, the extra work is worth the initial experience of seeing "$0.00" instead of "Not Available".

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -273,10 +273,10 @@ class HttpGoogleServicesDAO(
 
     retryWithRecoverWhen500orGoogleError(() => {
       val bucket = new Bucket().setName(bucketName)
-      val deleteAfterOneYear = new Lifecycle.Rule()
+      val storageLogExpiration = new Lifecycle.Rule()
```
dmohs, line 283: Why this rename? I judge that "delete after one year" is a slightly better name for a rule than "storage log expiration."

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
```diff
@@ -461,12 +463,48 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationT
...
-    intercept[GoogleStorageLogException] {
-      Await.result(gcsDAO.getBucketUsage(RawlsBillingProjectName(testProject), "no-bucket-here-" + UUID.randomUUID.toString), Duration.Inf)
+  it should "return 0 usage when there are no storage logs and the bucket is empty" in {
+    val storage = gcsDAO.getStorage(gcsDAO.getBucketServiceAccountCredential)
+
+    val projectName = RawlsBillingProjectName(testProject)
+    val logBucketName = gcsDAO.getStorageLogsBucketName(projectName)
+    val bucketName = "services-dao-spec-" + UUID.randomUUID.toString
+    val bucket = new Bucket().setName(bucketName)
+    val bucketInserter = storage.buckets.insert(testProject, bucket)
```
dmohs, line 477: I don't love the idea of creating and deleting a real Google bucket on each run of our unit tests, but I think this is something we need to discuss in the context of testing philosophy. Consider this just a note to myself.

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
```diff
@@ -461,12 +463,48 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationT
...
+
+    try {
+      val usage = Await.result(gcsDAO.getBucketUsage(projectName, bucketName), Duration.Inf)
+      usage should be (0)
+    } finally {
+      Await.result(gcsDAO.deleteBucket(bucketName, bucketDeletionMonitor), Duration.Inf)
     }
   }
 
+  it should "return 'Not Available' when there are no storage logs and the bucket is NOT empty" in {
```
dmohs, line 497: This test could use a comment as to why there would be no storage logs (I'm guessing some time has to pass after object insertion for the logs to show up).

dmohs (approved):
Looks good to me with a couple notes.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
breilly2: I agreed, until "one year" became "180 days" at which point the variable name became misleading. Since the time period is in config, it can't be encoded into the variable name.

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
breilly2: Good point. I had written it in the JIRA ticket, but I agree that it's worth repeating here.

breilly2:
Added requested comments. However, I am holding this PR open while I investigate whether or not Google writes a 0 storage log when a bucket goes from non-empty to empty. (I'm fearing that they don't. If that's the case this feature may be far from done.)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dmohs: Ah! So misleading is bad, but I might rename it to `deleteExpired` or `deleteAfterExpiration`. Totally subjective (no need to change), but to me a rule feels more like a verb than a noun.

dmohs (approved):

breilly2:
Good news! Google actually writes a 0 storage log once for an empty bucket that previously had something in it! I think this is good to go.

dvoet (approved):
👍

breilly2:
One more commit to isolate inserting the initial storage log from creating the bucket. Please review the most recent commit. Thanks.

dmohs (approved):
Still looks good to me.

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
dvoet: this is not part of our unit tests, it is integration testing which happens for each develop build

dvoet (approved):
👍

PR #611, opened 2017-01-24
GAWB-1473: Allow owners to grant share permissions to others

MatthewBemis:
firecloud-ui: https://github.com/broadinstitute/firecloud-ui/pull/643
firecloud-orch: broadinstitute/firecloud-orchestration#352

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3042,7 +3042,7 @@ paths:
           schema:
             type: array
             items:
-              $ref: '#/definitions/WorkspaceListResponse'
+              $ref: '#/definitions/WorkspaceResponse'
         '500':
           description: Rawls Internal Error
           schema:
@@ -4493,6 +4493,21 @@ definitions:
         type: date-time
         description: When the event finished
 
+  WorkspaceResponse:
+    description: ''
+    properties:
+      accessLevel:
+        $ref: '#/definitions/WorkspaceAccessLevel'
+      canShare:
+        type: boolean
+      workspace:
+        $ref: '#/definitions/Workspace'
+      workspaceSubmissionStats:
+        $ref: '#/definitions/WorkspaceSubmissionStats'
+      owners:
+        type: array
+        items:
+          type: string
   WorkspaceListResponse:
     description: ''
     properties:
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -2938,6 +2938,8 @@ paths:
       responses:
         '200':
           description: Successful Request
+          schema:
+            $ref: '#/definitions/WorkspaceACLUpdateResponseList'
         '400':
           description: Attempting to change access controls you can't (your own, or public-access group)
           schema:
@@ -4288,6 +4290,35 @@ definitions:
       canShare:
         type: boolean
         description: Set to true if you want this user to be able to share the workspace with other users
+
+  WorkspaceACLUpdateResponseList:
+    description: ''
+    required:
+      - usersUpdated
+      - invitesSent
+      - invitesUpdated
+      - usersNotFound
+    properties:
+      usersUpdated:
+        type: array
+        description: the users or groups who were updated
+        items:
+          $ref: '#/definitions/WorkspaceACLUpdateResponse'
+      invitesSent:
+        type: array
+        description: the list of invites that were sent
+        items:
+          $ref: '#/definitions/WorkspaceACLUpdate'
+      invitesUpdated:
+        type: array
+        description: the list of invites that were updated
+        items:
+          $ref: '#/definitions/WorkspaceACLUpdate'
+      usersNotFound:
+        type: array
+        description: the users or groups who were not found
+        items:
+          $ref: '#/definitions/WorkspaceACLUpdate'
 
   WorkspaceAccessEntry:
     description: ''
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -4277,6 +4277,7 @@ definitions:
     required:
       - email
       - accessLevel
+      - canShare
     properties:
       email:
         type: string
@@ -4284,12 +4285,16 @@ definitions:
       accessLevel:
         type: string
         description: The access level to grant to this user or group (OWNER, READER, WRITER, NO ACCESS)
+      canShare:
+        type: boolean
+        description: Set to true if you want this user to be able to share the workspace with other users
 
   WorkspaceAccessEntry:
     description: ''
     required:
       - accessLevel
       - pending
+      - canShare
     properties:
       accessLevel:
         type: string
@@ -4297,6 +4302,9 @@ definitions:
       pending:
         type: boolean
         description: The status of the users access
+      canShare:
+        type: boolean
+        description: True if the user can share the workspace with others, false otherwise
 
   Workspace:
     description: ''
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -106,8 +106,8 @@ trait WorkspaceComponent {
     def workspaceId = column[UUID]("workspace_id")
     def userSubjectId = column[String]("user_subject_id")
 
-    def workspace = foreignKey("FK_WS_USER_SHARE_WS", workspaceId, workspaceQuery)(_.id)
-    def user = foreignKey("FK_WS_USER_SHARE_GROUP", userSubjectId, rawlsUserQuery)(_.userSubjectId)
+    def workspace = foreignKey("FK_USER_SHARE_PERMS_WS", workspaceId, workspaceQuery)(_.id)
+    def user = foreignKey("FK_USER_SHARE_PERMS_USER", userSubjectId, rawlsUserQuery)(_.userSubjectId)
 
     def * = (workspaceId, userSubjectId) <> (WorkspaceUserShareRecord.tupled, WorkspaceUserShareRecord.unapply)
   }
@@ -116,8 +116,8 @@ trait WorkspaceComponent {
     def workspaceId = column[UUID]("workspace_id")
     def groupName = column[String]("group_name")
 
-    def workspace = foreignKey("FK_WS_GROUP_SHARE_WS", workspaceId, workspaceQuery)(_.id)
-    def group = foreignKey("FK_WS_GROUP_SHARE_GROUP", groupName, rawlsGroupQuery)(_.groupName)
+    def workspace = foreignKey("FK_GROUP_SHARE_PERMS_WS", workspaceId, workspaceQuery)(_.id)
+    def group = foreignKey("FK_GROUP_SHARE_PERMS_GROUP", groupName, rawlsGroupQuery)(_.groupName)
 
     def * = (workspaceId, groupName) <> (WorkspaceGroupShareRecord.tupled, WorkspaceGroupShareRecord.unapply)
   }
@@ -268,8 +268,18 @@ trait WorkspaceComponent {
       findWorkspaceInvitesQuery(workspaceId).delete
     }
 
-    def getUserSharePerms(subjectId: RawlsUserSubjectId, workspaceContext: SlickWorkspaceContext): ReadAction[Int] = {
-      workspaceUserShareQuery.filter(rec => (rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId)).countDistinct.result
+    def getUserSharePerms(subjectId: RawlsUserSubjectId, workspaceContext: SlickWorkspaceContext): ReadAction[Boolean] = {
+      //here i need to check the groups for each user to see if those groups have share permissions
+      rawlsGroupQuery.listGroupsForUser(RawlsUserRef(subjectId)).flatMap { x =>
+        println(x)
+        val groupNames = x.map(_.groupName.value)
+        println(groupNames)
+        workspaceGroupShareQuery.filter(rec => (rec.groupName) inSet(groupNames)).countDistinct.result.map(rows => rows > 0) flatMap { y =>
+          println(y)
+          if(y) DBIO.successful(y)
+          else workspaceUserShareQuery.filter(rec => (rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId)).countDistinct.result.map(rows => rows > 0)
+        }
+      }
     }
 
     def listEmailsAndAccessLevel(workspaceContext: SlickWorkspaceContext): ReadAction[(Seq[(String, WorkspaceAccessLevel, Boolean)], Seq[(String, WorkspaceAccessLevel, Boolean)])] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -327,13 +327,11 @@ trait WorkspaceComponent {
 
       userShareQuery.result.map(_.map { case (access, email, canShare) =>
         val accessLevel = WorkspaceAccessLevels.withName(access)
-        // Sharing is implied for owners and project owners, so the permission may not be in the DB. Rectify that here
-        (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
+        (email, accessLevel, canShare)
       }).flatMap { userResults =>
         groupShareQuery.result.map(_.map { case (access, email, canShare) =>
           val accessLevel = WorkspaceAccessLevels.withName(access)
-          // Sharing is implied for owners and project owners, so the permission may not be in the DB. Rectify that here
-          (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
+          (email, accessLevel, canShare)
         }).map { groupResults =>
           userResults ++ groupResults
         }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -325,15 +325,13 @@ trait WorkspaceComponent {
       val userShareQuery = accessAndUserEmail.joinLeft(workspaceUserShareQuery).on(_._3 === _.userSubjectId).map { case ((accessLevel, userEmail, _), shareRecord) => (accessLevel, userEmail, shareRecord.isDefined) }
       val groupShareQuery = accessAndSubGroupEmail.joinLeft(workspaceGroupShareQuery).on(_._3 === _.groupName).map { case ((accessLevel, groupEmail, _), shareRecord) => (accessLevel, groupEmail, shareRecord.isDefined) }
 
-      userShareQuery.result.map(_.map { case (access, email, hasSharePermission) =>
-        val accessLevel = WorkspaceAccessLevels.withName(access)
-        (email, accessLevel, hasSharePermission)
-      }).flatMap { userResults =>
-        groupShareQuery.result.map(_.map { case (access, email, hasSharePermission) =>
+      for {
+        userResults <- userShareQuery.result
+        groupResults <- groupShareQuery.result
+      } yield {
+        (userResults ++ groupResults).map { case (access, email, hasSharePermission) =>
           val accessLevel = WorkspaceAccessLevels.withName(access)
           (email, accessLevel, hasSharePermission)
-        }).map { groupResults =>
-          userResults ++ groupResults
         }
       }
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -32,6 +32,9 @@ case class WorkspaceAccessRecord(workspaceId: UUID, groupName: String, accessLev
 /** result class for workspaceQuery.findAssociatedGroupsToIntersect, target = group1 intersect group2 */
 case class GroupsToIntersect(target: RawlsGroupRef, group1: RawlsGroupRef, group2: RawlsGroupRef)
 
+case class WorkspaceUserShareRecord(workspaceId: UUID, subjectId: String)
+case class WorkspaceGroupShareRecord(workspaceId: UUID, groupName: String)
+
 case class PendingWorkspaceAccessRecord(
   workspaceId: UUID,
   userEmail: String,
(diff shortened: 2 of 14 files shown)


core/src/main/resources/swagger/rawls.yaml
```diff
@@ -4277,26 +4279,63 @@ definitions:
...
         type: string
         description: email address of the user or group whose permissions will be changed
       accessLevel:
         type: string
         description: The access level to grant to this user or group (OWNER, READER, WRITER, NO ACCESS)
+      canShare:
+        type: boolean
+        description: Set to true if you want this user to be able to share the workspace with other users
+
+  WorkspaceACLUpdateResponseList:
```
MatthewBemis, line 4294: this particular addition was forgotten in the firecloud invites PR. i did make one small change to this response- i differentiate invitesSent from invitesUpdated so orchestration doesn't potentially send out too many invites. small issue so i fixed it here. orch has been updated accordingly

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
+
+      userPermissionQuery.result.map(_.map { case (access, email, canShare) =>
+        val accessLevel = WorkspaceAccessLevels.withName(access)
+        // Sharing is implied for owners and project owners, so the permission may not be in the DB. Rectify that here
+        (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
+      }).flatMap { userResults =>
+        groupPermissionQuery.result.map(_.map { case (access, email, canShare) =>
+          val accessLevel = WorkspaceAccessLevels.withName(access)
+          // Sharing is implied for owners and project owners, so the permission may not be in the DB. Rectify that here
+          (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
```
MatthewBemis: to expand on the comment, i deemed it too much overhead to maintain share records in the DB for access levels that implicitly had the ability to share. this would be particularly annoying for project owners, who are altered outside of the context of a workspace (and doing so may impact many workspaces)

this also means we don't need to write any sort of migration

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -195,14 +195,16 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-              RequestComplete(StatusCodes.OK, WorkspaceListResponse(accessLevel, workspaceContext.workspace, stats, owners))
+          getUserSharePermissions(workspaceContext, accessLevel, dataAccess) flatMap { canShare =>
+            if (accessLevel < WorkspaceAccessLevels.Read)
+              DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound, noSuchWorkspaceMessage(workspaceName))))
+            else {
+              for {
+                stats <- getWorkspaceSubmissionStats(workspaceContext, dataAccess)
+                owners <- getWorkspaceOwners(workspaceContext.workspace, dataAccess)
+              } yield {
+                RequestComplete(StatusCodes.OK, WorkspaceResponse(accessLevel, canShare, workspaceContext.workspace, stats, owners))
```
MatthewBemis, line 206: I changed this response type because it was also used in the listWorkspaces API call. I needed to return the canShare value but calculating it for the entire list of workspaces was useless and would add a lot of extra queries. So I broke them apart

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
     def deleteWorkspaceInvites(workspaceId: UUID) = {
       findWorkspaceInvitesQuery(workspaceId).delete
     }
-    
-    def listEmailsAndAccessLevel(workspaceContext: SlickWorkspaceContext): ReadAction[Seq[(String, WorkspaceAccessLevel)]] = {
+
+    def updateUserSharePermissions(workspaceId: UUID, users: Map[RawlsUserRef, Option[Boolean]]) = {
```
dvoet: these update functions are ugly. I would just have an insert and a delete since there isn't actually an update.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
+        workspaceGroupShareQuery.filter(_.workspaceId === workspaceId).delete
+    }
+
+    def getUserSharePermissions(subjectId: RawlsUserSubjectId, workspaceContext: SlickWorkspaceContext): ReadAction[Boolean] = {
+      workspaceUserShareQuery.filter(rec => (rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId)).countDistinct.result.map(rows => rows > 0) flatMap { canShare =>
+        if(canShare) DBIO.successful(canShare)
+        else rawlsGroupQuery.listGroupsForUser(RawlsUserRef(subjectId)).flatMap { userGroups =>
+          val groupNames = userGroups.map(_.groupName.value)
+          workspaceGroupShareQuery.filter(rec => (rec.groupName) inSet(groupNames)).countDistinct.result.map(rows => rows > 0) flatMap { canShareViaGroup =>
+            DBIO.successful(canShareViaGroup)
```
dvoet: do you need the last flatmap?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
-        (email, WorkspaceAccessLevels.withName(access))
-      })
+      } yield (access, subGroup)).map { case (access, subGroup) => (access.accessLevel, subGroup.groupEmail, subGroup.groupName) }
+
+      /*  The left join here is important. Since we are only going to store share-permission records for users and groups that have been
+          explicitly granted that ability, we want to be able to return all users and groups even if they don't have share permissions.
+          After the join, a null column equates to false, and a non-null column equates to true. This conversion is done with permission.isDefined
+       */
+      val userPermissionQuery = accessAndUserEmail.joinLeft(workspaceUserShareQuery).on(_._3 === _.userSubjectId).map { case (userInfo, permission) => (userInfo._1, userInfo._2, permission.isDefined) }
+      val groupPermissionQuery = accessAndSubGroupEmail.joinLeft(workspaceGroupShareQuery).on(_._3 === _.groupName).map { case (groupInfo, permission) => (groupInfo._1, groupInfo._2, permission.isDefined) }
```
dvoet: use the term Share instead of Permission

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
dvoet: why do we have this logic here instead of in requireSharePermission? This makes it feel like there is an actual row in the db which is maybe changeable. I think I would prefer requireSharePermission have the logic that allows access if the user is >= Owner or canShare

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
+        else rawlsGroupQuery.listGroupsForUser(RawlsUserRef(subjectId)).flatMap { userGroups =>
+          val groupNames = userGroups.map(_.groupName.value)
+          workspaceGroupShareQuery.filter(rec => (rec.groupName) inSet(groupNames)).countDistinct.result.map(rows => rows > 0) flatMap { canShareViaGroup =>
+            DBIO.successful(canShareViaGroup)
+          }
+        }
+      }
+    }
+
+    def listEmailsAndAccessLevel(workspaceContext: SlickWorkspaceContext): ReadAction[(Seq[(String, WorkspaceAccessLevel, Boolean)], Seq[(String, WorkspaceAccessLevel, Boolean)])] = {
```
dvoet: why does this return a tuple now?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -435,15 +438,17 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
           dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext).flatMap { emailsAndAccess =>
             dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId).map { invites =>
               // toMap below will drop duplicate keys, keeping the last entry only
               // sort by access level to make sure higher access levels remain in the resulting map
-              val granted = emailsAndAccess.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, false)}
-              val pending = invites.sortBy { case (_, accessLevel) => accessLevel }.map { case (email, accessLevel) => email -> AccessEntry(accessLevel, true)}
+              val grantedUsers = emailsAndAccess._1.sortBy { case (_, accessLevel, _) => accessLevel }.map { case (email, accessLevel, canShare) => email -> AccessEntry(accessLevel, false, canShare) }
+              val grantedGroups = emailsAndAccess._2.sortBy { case (_, accessLevel, _) => accessLevel }.map { case (email, accessLevel, canShare) => email -> AccessEntry(accessLevel, false, canShare) }
```
dvoet: in general I am seeing too much tuple access by position (i.e. x._2). This makes the code less readable. Can you review this PR for these cases and change them to case statements where possible?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -481,7 +554,7 @@ trait WorkspaceComponent {
       }
     }
 
-    def findWorkspaceUsersAndAccessLevel(workspaceId: UUID): ReadAction[Set[(Either[RawlsUserRef, RawlsGroupRef], WorkspaceAccessLevel)]] = {
+    def findWorkspaceUsersAndAccessLevel(workspaceId: UUID): ReadAction[Set[(Either[RawlsUserRef, RawlsGroupRef], (WorkspaceAccessLevel, Option[Boolean]))]] = {
```
dvoet: why is the Boolean an Option?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -544,20 +562,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      val refsToUpdate = aclUpdates.map { aclUpdate => (refsToUpdateByEmail.get(aclUpdate.email), aclUpdate.accessLevel, aclUpdate.canShare) }.collect {
+        case (Some(ref), WorkspaceAccessLevels.NoAccess, _) => ref -> (WorkspaceAccessLevels.NoAccess, Option(false))
+        case (Some(ref), accessLevel, canShare) => ref -> (accessLevel, canShare)
       }.toSet
 
+      val refsToUpdateAndSharePermission = refsToUpdate.map(ref => ref._1 -> ref._2._2)
+      val refsToUpdateAndAccessLevel = refsToUpdate.map(ref => ref._1 -> ref._2._1)
+
+      val existingRefsAndSharePermission = existingRefsAndLevels.map(ref => ref._1 -> ref._2._2)
+      val existingRefsAndAccessLevel = existingRefsAndLevels.map(ref => ref._1 -> ref._2._1)
```
dvoet: :nauseated_face:

dvoet (changes requested):
Maybe adding a new case class that contains both WorkspaceAccessLevel and canShare can help clean things up.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: So it would be easier to diff this with the user-supplied aclUpdates in determineCompleteNewAcls, but I suppose that would be better handled in determineCompleteNewAcls and not in the database code

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: this is used in getACL, which now returns acls in the form:

```
[
   {
      "email": "foo@bar.baz",
      "canShare": true,
      "accessLevel": "OWNER"
   }
]
```

it does feel dishonest to be pretending that there are rows in the DB but it felt cleaner code-wise to deal with that here than in getACL

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,77 @@ trait WorkspaceComponent {
...
+        (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
+      }).flatMap { userResults =>
+        groupShareQuery.result.map(_.map { case (access, email, canShare) =>
+          val accessLevel = WorkspaceAccessLevels.withName(access)
+          // Sharing is implied for owners and project owners, so the permission may not be in the DB. Rectify that here
+          (email, accessLevel, ((accessLevel >= WorkspaceAccessLevels.Owner) || canShare))
+        }).map { groupResults =>
+          userResults ++ groupResults
+        }
+      }
```
dvoet, line 336: rewrite this last block as a for comprehension

```
      for {
        userResults <- userShareQuery.result
        groupResults <- groupShareQuery.result
      } yield {
        (userResults ++ groupResults).map { case (access, email, hasSharePermission) =>
          (email, WorkspaceAccessLevels.withName(access), hasSharePermission)
        }
      }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
+      } yield (access, subGroup)).map { case (access, subGroup) => (access.accessLevel, subGroup.groupEmail, subGroup.groupName) }
 
-      (accessAndUserEmail union accessAndSubGroupEmail).result.map(_.map { case (access, email) => 
-        (email, WorkspaceAccessLevels.withName(access))
-      })
+      /*  The left join here is important. Since we are only going to store share-permission records for users and groups that have been
+          explicitly granted that ability, we want to be able to return all users and groups even if they don't have share permissions.
+          After the join, a null column equates to false, and a non-null column equates to true. This conversion is done with permission.isDefined
+       */
+      val userShareQuery = accessAndUserEmail.joinLeft(workspaceUserShareQuery).on(_._3 === _.userSubjectId).map { case ((accessLevel, userEmail, _), shareRecord) => (accessLevel, userEmail, shareRecord.isDefined) }
```
MatthewBemis: The SQL query that slick actually generates:

```
SELECT 
    x2.`access_level`, x3.`EMAIL`, x4.`workspace_id` IS NOT NULL
FROM
    `WORKSPACE_ACCESS` x2
        INNER JOIN
    (`GROUP` x5
    INNER JOIN (`GROUP_USERS` x6
    INNER JOIN `USER` x3 ON x3.`SUBJECT_ID` = x6.`USER_SUBJECT_ID`) ON x5.`NAME` = x6.`GROUP_NAME`) ON ((x2.`workspace_id` = 0xE5D5CA221F0146DF9F92551963A9A1)
        AND (x2.`is_realm_acl` = FALSE))
        AND (x2.`group_name` = x5.`NAME`)
        LEFT OUTER JOIN
    `WORKSPACE_USER_SHARE` x4 ON x3.`SUBJECT_ID` = x4.`user_subject_id`;
```

explain plan:

screen shot 2017-01-25 at 9 31 03 am

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -496,12 +573,23 @@ trait WorkspaceComponent {
...
+        (subjectId, accessLevel, permission.isDefined)
+      }
+      val groupsWithSharePermission = subGroupQuery.joinLeft(workspaceGroupShareQuery).on(_._1 === _.groupName).map { case ((groupName, accessLevel), permission) =>
+        (groupName, accessLevel, permission.isDefined)
+      }
+
       for {
-        users <- userQuery.result.map {
-          _.map { case (subjectId, accessLevel) => (Left(RawlsUserRef(RawlsUserSubjectId(subjectId))), WorkspaceAccessLevels.withName(accessLevel)) }
+        users <- usersWithSharePermission.result.map {
```
dvoet, line 580: do this map and the one for groups a single time in the yield

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -242,25 +267,73 @@ trait WorkspaceComponent {
...
 
-      (accessAndUserEmail union accessAndSubGroupEmail).result.map(_.map { case (access, email) => 
-        (email, WorkspaceAccessLevels.withName(access))
-      })
+      /*  The left join here is important. Since we are only going to store share-permission records for users and groups that have been
+          explicitly granted that ability, we want to be able to return all users and groups even if they don't have share permissions.
+          After the join, a null column equates to false, and a non-null column equates to true. This conversion is done with permission.isDefined
+       */
+      val userShareQuery = accessAndUserEmail.joinLeft(workspaceUserShareQuery).on(_._3 === _.userSubjectId).map { case ((accessLevel, userEmail, _), shareRecord) => (accessLevel, userEmail, shareRecord.isDefined) }
+      val groupShareQuery = accessAndSubGroupEmail.joinLeft(workspaceGroupShareQuery).on(_._3 === _.groupName).map { case ((accessLevel, groupEmail, _), shareRecord) => (accessLevel, groupEmail, shareRecord.isDefined) }
```
MatthewBemis: The SQL query that slick actually generates:

```
SELECT 
    x2.`access_level`, x3.`EMAIL`, x4.`workspace_id` IS NOT NULL
FROM
    `WORKSPACE_ACCESS` x2
        INNER JOIN
    (`GROUP` x5
    INNER JOIN (`GROUP_SUBGROUPS` x6
    INNER JOIN `GROUP` x3 ON x3.`NAME` = x6.`CHILD_NAME`) ON x5.`NAME` = x6.`PARENT_NAME`) ON ((x2.`workspace_id` = 0xE5D5CA221F0146DF9F92551963A9A1)
        AND (x2.`is_realm_acl` = FALSE))
        AND (x2.`group_name` = x5.`NAME`)
        LEFT OUTER JOIN
    `WORKSPACE_GROUP_SHARE` x4 ON x3.`NAME` = x4.`group_name`;
```

explain plan:

screen shot 2017-01-25 at 9 32 22 am

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -507,15 +513,33 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
       val usersNotFound = emailsNotFound.filterNot(aclUpdate => invitesUpdated.map(_.email).contains(aclUpdate.email)).filterNot(aclUpdate => existingInvites.map(_.email).contains(aclUpdate.email))
+      val invitesSent = invitesUpdated.filterNot(aclUpdate => existingInvites.map(_.email).contains(aclUpdate.email))
 
-      WorkspaceACLUpdateResponseList(usersUpdated, invitesUpdated, usersNotFound)
+      WorkspaceACLUpdateResponseList(usersUpdated, invitesSent, invitesUpdated.diff(invitesSent), usersNotFound)
+    }
+
+    def updateWorkspaceSharePermissions(actualChangesToMake: Map[Either[RawlsUserRef, RawlsGroupRef], Option[Boolean]]) = {
+      val (usersToAdd, usersToRemove) = actualChangesToMake.collect{ case (Left(userRef), canShare) => userRef -> canShare }
+        .filter { case (_, canShare) => canShare.isDefined }.map{ case (user, canShare) => user -> canShare.get }.toSeq.partition { case (_, canShare) => canShare }
```
dvoet: you can get rid of the inner filter and map
``` 
     val (usersToAdd, usersToRemove) = actualChangesToMake.collect{ case (Left(userRef), Some(canShare)) => userRef -> canShare }.toSeq.partition { case (_, canShare) => canShare }
```

dvoet (commented):
last change

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -473,7 +479,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       dataSource.inTransaction { dataAccess =>
         withWorkspaceContext(workspaceName, dataAccess) { workspaceContext =>
           dataAccess.workspaceQuery.getInvites(workspaceContext.workspaceId).map { pairs =>
-            pairs.map(pair => WorkspaceACLUpdate(pair._1, pair._2))
+            pairs.map(pair => WorkspaceACLUpdate(pair._1, pair._2, None))
```
breilly2: Can `WorkspaceACLUpdate` default the 3rd parameter to `None`? That would make this easier to read by not having to wonder what `None` here means.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -32,6 +32,9 @@ case class WorkspaceAccessRecord(workspaceId: UUID, groupName: String, accessLev
 /** result class for workspaceQuery.findAssociatedGroupsToIntersect, target = group1 intersect group2 */
 case class GroupsToIntersect(target: RawlsGroupRef, group1: RawlsGroupRef, group2: RawlsGroupRef)
 
+case class WorkspaceUserShareRecord(workspaceId: UUID, subjectId: String)
+case class WorkspaceGroupShareRecord(workspaceId: UUID, groupName: String)
```
breilly2, line 36: It bothers me that there are separate tables here. I guess it makes sense that `RawlsUser` and `RawlsGroup` are different entities. However, there is definitely an area of functional overlap that is leading to a lot of code duplication. This duplication is now extending to the database for a feature that is truly shared between the two.

I don't have any immediate suggestions for changing this, and maybe there's something forcing it to be this way. However, this does seem to be the root of quite a bit of code duplication in these changes and I wanted to point it out so we could keep it in mind going forward.

breilly2 (changes requested):
Also awaiting implementations for some tests, but wanted to get a batch of comments in.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1694,23 +1695,78 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
...
+        assertResult(StatusCodes.OK, response.entity.asString) { status }
+        responseAs[WorkspaceACL].acl should contain (testData.userOwner.userEmail.value -> AccessEntry(WorkspaceAccessLevels.ProjectOwner, false, true))
+      }
   }
 
-  it should "allow a user in a group with share permissions to share equal to and below their access level" in withTestDataApiServicesAndUser("writer-access") { services =>
-    assert(true)
+  it should "allow a user in a group with share permissions to share equal to and below their access level" in withTestDataApiServicesAndUser("reader-access") { services =>
+    import WorkspaceACLJsonSupport._
+    Patch(s"/workspaces/${testData.workspaceToTestGrant.namespace}/${testData.workspaceToTestGrant.name}/acl", httpJson(Seq(WorkspaceACLUpdate(testData.userReader2.userEmail.value, WorkspaceAccessLevels.Read, None)))) ~>
```
breilly2: I'm not 100% sure this is a different test from "allow a writer with share permissions to share equal to and below their access level". Isn't `testData.userReader2` the target of the change, not the user making the change? I would expect there to be a difference with `in withTestDataApiServiceAndUser("reader-access")` to have the change being requested by a user in a group.

breilly2 (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
MatthewBemis: This test is a little tricky because I had to set part of it up in TestDriverComponent. I wanted to test that a user in dbGapAuthorizedUsers (user "reader-access") could share if they were in that group and the group had share permission on the workspace.

Relevant lines are 209 of TestDriverComponent where I add "reader-access" to the dbGap group and line 544 where I manipulate the database to give that group the share permission on the test workspace.

The reason that I had to do it this way is that I can only run a test as one user. I'd need to use the user "owner-access" in order to add dbGapAuthorizedUsers to the workspace with the share permissions, but then I can't test anything as the user "reader-access". 

Basically, the assumption for this test is that dbGapAuthorizedUsers has already been granted share permissions by the owner and that "reader-access" is a member of that group.

breilly2: Okay, I understand that now.

The problem I see is that the group membership and permissions are not obvious when looking at this test. You have to trace through the all-in-one test fixture to see it, including understanding that there is a database lookup linking `"reader-access"` to `testData.userReader` because there is no way to navigate that statically. The risk with having these facts so far apart is that it's not obvious how changes in `TestDriverComponent` might affect which tests. For example, I committed tests related to bucket usage that use `"reader-access"` on the same day that you added that user to a group. So, while I was doing development, the user wasn't in a group but, by the time both of our branches made it to develop, the user was. Thankfully, it doesn't make a difference for my tests, but I think it illustrates the danger that I talked about at the cross-team tech meeting a few weeks ago.

Can the username be changed to make the user specifics obvious, such as `"reader-access-group-can-share"`? Or maybe even introduce a new user with that name, leaving `"reader-access"` as it was.

breilly2 (approved):
:+1: Awesome!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
dvoet: agreed

dvoet (approved):

PR #615, opened 2017-01-30
GAWB-1494/1495/1498: Realms MVP for admins

MatthewBemis:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
@@ -299,8 +299,8 @@ trait RawlsGroupComponent {
       }
     }
 
-    def deleteRealmRecord(groupRef: RawlsGroupRef): ReadWriteAction[Int] = {
-      (realmQuery.filter(_.groupName === groupRef.groupName.value)).delete
+    def deleteRealmRecord(groupRef: RawlsRealmRef): ReadWriteAction[Int] = {
+      (realmQuery.filter(_.groupName === groupRef.realmName.value)).delete
     }
 
     private def marshalRawlsGroup(group: RawlsGroup): RawlsGroupRecord = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
@@ -8,6 +8,7 @@ import org.broadinstitute.dsde.rawls.model._
 case class RawlsGroupRecord(groupName: String, groupEmail: String, updatedDate: Option[Timestamp])
 case class GroupUsersRecord(userSubjectId: String, groupName: String)
 case class GroupSubgroupsRecord(parentGroupName: String, childGroupName: String)
+case class RealmRecord(groupName: String)
 
 trait RawlsGroupComponent {
   this: DriverComponent
@@ -48,8 +49,17 @@ trait RawlsGroupComponent {
     def pk = primaryKey("PK_GROUP_SUBGROUPS", (parentGroupName, childGroupName))
   }
 
+  class RealmsTable(tag: Tag) extends Table[RealmRecord](tag, "REALM") {
+    def groupName = column[String]("GROUP_NAME", O.Length(254))
+
+    def * = (groupName) <> (RealmRecord.apply _, RealmRecord.unapply)
+
+    def group = foreignKey("FK_REALM_GROUP_NAME", groupName, rawlsGroupQuery)(_.groupName)
+  }
+
   protected val groupUsersQuery = TableQuery[GroupUsersTable]
   protected val groupSubgroupsQuery = TableQuery[GroupSubgroupsTable]
+  protected val realmsQuery = TableQuery[RealmsTable]
   type GroupQuery = Query[RawlsGroupTable, RawlsGroupRecord, Seq]
   private type GroupUsersQuery = Query[GroupUsersTable, GroupUsersRecord, Seq]
   private type GroupSubgroupsQuery = Query[GroupSubgroupsTable, GroupSubgroupsRecord, Seq]
@@ -279,6 +289,18 @@ trait RawlsGroupComponent {
             subgroups <- findSubgroupsByGroupName(groupRec.groupName).result
           } yield Option(unmarshalRawlsGroup(groupRec, users, subgroups))
       }
+    }
+
+    def setGroupAsRealm(groupRef: RawlsGroupRef): ReadWriteAction[Int] = {
+      uniqueResult[RawlsGroupRecord](rawlsGroupQuery.filter(_.groupName === groupRef.groupName.value)).flatMap {
+        case None => DBIO.successful(0)
+        case Some(groupRec) =>
+          realmsQuery += RealmRecord(groupRef.groupName.value)
+      }
+    }
+
+    def deleteRealmRecord(groupRef: RawlsGroupRef): ReadWriteAction[Int] = {
+      (realmsQuery.filter(_.groupName === groupRef.groupName.value)).delete
     }
 
     private def marshalRawlsGroup(group: RawlsGroup): RawlsGroupRecord = {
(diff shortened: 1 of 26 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -551,6 +556,32 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+      dataAccess.workspaceQuery.listWorkspacesInRealm(groupRef) flatMap {
+        case Seq() =>
+          dataAccess.rawlsGroupQuery.deleteRealmRecord(groupRef) flatMap { _ =>
+            withGroup(groupRef, dataAccess) { group =>
+              dataAccess.rawlsGroupQuery.delete(groupRef) andThen
+                DBIO.from(gcsDAO.deleteGoogleGroup(group)) map { _ => RequestComplete(StatusCodes.OK) }
+            }
+          }
+        case workspaceNames =>
+          DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict, s"Unable to delete realm ${groupRef.groupName}. You must remove the following workspaces before you can remove the realm: ${workspaceNames}")))
```
MatthewBemis: I haven't yet thought about the security considerations for revealing all of the workspaces in a realm (since it's admin/owner only I expect it to be fine), nor have I really considered if this was the correct behavior in this case

dvoet: don't list all the workspaces, just say that some exist, an admin can always look in the db

breilly2:
I would expect the types of `WorkspaceModel.Workspace.realm` and `WorkspaceModel.WorkspaceRequest.realm` to change to `Option[RawlsRealmRef]`.

dvoet (approved):
👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
```diff
@@ -281,6 +291,18 @@ trait RawlsGroupComponent {
       }
     }
 
+    def setGroupAsRealm(groupRef: RawlsGroupRef): ReadWriteAction[Int] = {
```
breilly2: This doesn't sit right with me because it seems valid to promote any group to a realm at any time when this should only be done when creating the realm. Right now, realm creation is in `UserService`. Maybe some of that code should move down into `RawlsGroupComponent`?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -25,6 +25,7 @@ case class RawlsGroup(groupName: RawlsGroupName, groupEmail: RawlsGroupEmail, us
 
 object RawlsGroup {
   implicit def toRef(g: RawlsGroup) = RawlsGroupRef(g.groupName)
+  implicit def toRealmRef(g: RawlsGroup) = RawlsRealmRef(g.groupName)
```
breilly2: IntelliJ flags the lack of an explicit return type here (same for `RawlsGroup.toRef` and `RawlsUser.toRef`) as something that can slow down compilation:

> Absence of explicit return type for implicit functions can make compilation slower. Also it can negatively affect performance of the scala-plugin.

Also, this doesn't appear to be used anywhere... or maybe it is and IntelliJ isn't noticing the implicit usage?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -551,6 +557,32 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     }
   }
 
+  def createRealm(realmRef: RawlsRealmRef) = {
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.rawlsGroupQuery.load(realmRef) flatMap {
+        case Some(_) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict, s"Group [${realmRef.realmName.value}] already exists")))
+        case None =>
+          createGroupInternal(realmRef, dataAccess) andThen dataAccess.rawlsGroupQuery.setGroupAsRealm(realmRef) map { _ => RequestComplete(StatusCodes.Created) }
```
breilly2: This is the specific combination of actions that seem like they should be pulled out of here and into somewhere else (like `RawlsGroupComponent`). As it is, `createGroupInternal` creates the Google group and inserts the `GROUP` row, then `setGroupAsRealm` inserts the `REALM` row.

Ultimately, I think there should be a function that queues both inserts (cohesive creation of a realm) without declaring its own transaction. Then something above it can set up the `Future` to create the Google group. Then this service call can establish the transaction, invoke the thing to do the actual work, and handle the service response.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -87,6 +87,22 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
           WorkspaceService.AdminAbortSubmission(WorkspaceName(workspaceNamespace, workspaceName), submissionId))
       }
     } ~
+    path("admin" / "realms") {
+      post { //create a realm
```
breilly2: Do we prefer `PUT` or `POST` for create services? I took a quick look at some other services and it currently looks inconsistent.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/UserModel.scala
```diff
@@ -5,6 +5,15 @@ import spray.json._
 sealed trait UserAuthRef
 case class RawlsUserRef(userSubjectId: RawlsUserSubjectId) extends UserAuthRef
 case class RawlsGroupRef(groupName: RawlsGroupName) extends UserAuthRef
+case class RawlsRealmRef(realmName: RawlsGroupName) extends UserAuthRef
+
+object RawlsGroupRef {
+  implicit def toRealmRef(ref: RawlsGroupRef) = RawlsRealmRef(ref.groupName)
```
breilly2: More instances of implicits that IntelliJ warns will slow compilation.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
MatthewBemis: POST for creates. I think most of the inconsistent endpoints are back from before we tried to follow the more typical REST conventions

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
MatthewBemis: Good call with the explicit return types. We should probably add a follow-on ticket to update all of our implicits to have return types. I'm sure most of us would appreciate a little bit more speed from IntelliJ at this point...

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsGroupComponent.scala
MatthewBemis: I was imagining a scenario where admins could promote groups to be realms. I'll check with PO to see if that's something that we actually want. If not, then you're right that this is a potentially misleading/dangerous function to have

Regardless, I'll organize the whole idea of realm creation a little bit better between `UserService` and `RawlsGroupComponent`

breilly2:
I was thinking about it this morning and nothing that I commented on is a show-stopper. I would be okay with an initial merge to develop to unblock the remaining stories in the sprint before making any further changes to this one.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
jmthibault79: +1

model/src/main/scala/org/broadinstitute/dsde/rawls/model/UserModel.scala
```diff
@@ -5,6 +5,15 @@ import spray.json._
 sealed trait UserAuthRef
 case class RawlsUserRef(userSubjectId: RawlsUserSubjectId) extends UserAuthRef
 case class RawlsGroupRef(groupName: RawlsGroupName) extends UserAuthRef
+case class RawlsRealmRef(realmName: RawlsGroupName) extends UserAuthRef
+
+object RawlsGroupRef {
+  implicit def toRealmRef(ref: RawlsGroupRef): RawlsRealmRef = RawlsRealmRef(ref.groupName)
```
jmthibault79: Some conversions that this method allows will be invalid: a non-Realm Group to a Realm.  I am concerned that this will cause hard-to-debug problems.  Is it necessary?

PR #618, opened 2017-02-01
Fix create project polling

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-1579

The problem with the initial implementation is that we were not polling google operations "correctly". Sometime they fail. The new code polls correctly and handles failures. The solution is more complex. We need to track the operations in the database and do the right thing when they complete.

also see https://github.com/broadinstitute/firecloud-orchestration/pull/363

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -749,7 +749,7 @@ class HttpGoogleServicesDAO(
       case t: HttpResponseException if StatusCode.int2StatusCode(t.getStatusCode) == StatusCodes.Conflict =>
         throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"A google project by the name $projectName already exists"))
     } map ( googleOperation => {
-      if (toScalaBool(googleOperation.getDone) && Option(googleOperation.getError).map(_.getCode == GoogleRpcErrorCodes.ALREADY_EXISTS).getOrElse(false)) {
+      if (toScalaBool(googleOperation.getDone) && Option(googleOperation.getError).exists(_.getCode == GoogleRpcErrorCodes.ALREADY_EXISTS)) {
         throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"A google project by the name $projectName already exists"))
       }
       RawlsBillingProjectOperationRecord(projectName.value, CREATE_PROJECT_OPERATION, googleOperation.getName, toScalaBool(googleOperation.getDone), Option(googleOperation.getError).map(error => toErrorMessage(error.getMessage, error.getCode)), API_CLOUD_RESOURCE_MANAGER)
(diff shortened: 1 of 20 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -743,17 +748,58 @@ class HttpGoogleServicesDAO(
...
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"A google project by the name $projectName already exists"))
+      }
+      RawlsBillingProjectOperationRecord(projectName.value, CREATE_PROJECT_OPERATION, googleOperation.getName, toScalaBool(googleOperation.getDone), Option(googleOperation.getError).map(error => toErrorMessage(error.getMessage, error.getCode)), API_CLOUD_RESOURCE_MANAGER)
+    })
+  }
+
+  override def pollOperation(rawlsBillingProjectOperation: RawlsBillingProjectOperationRecord): Future[RawlsBillingProjectOperationRecord] = {
+    val credential = getBillingServiceAccountCredential
+
+    // this code is a colossal DRY violation but because the operations collection is different
```
breilly2, line 762: You might be able to be more DRY by using higher-order functions but I'm not sure it would be any more readable, especially with this comment being here. This is well contained and explained with comments. :+1:

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -743,17 +748,58 @@ class HttpGoogleServicesDAO(
     }).recover {
       case t: HttpResponseException if StatusCode.int2StatusCode(t.getStatusCode) == StatusCodes.Conflict =>
         throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"A google project by the name $projectName already exists"))
-    } map ( _ => ())
+    } map ( googleOperation => {
+      if (toScalaBool(googleOperation.getDone) && Option(googleOperation.getError).map(_.getCode == GoogleRpcErrorCodes.ALREADY_EXISTS).getOrElse(false)) {
```
breilly2: IntelliJ says, "Replace map and getOrElse(false) with exists" which yields:

`Option(googleOperation.getError).exists(_.getCode == GoogleRpcErrorCodes.ALREADY_EXISTS)`

Other the the overloading of "exists" (which is only a little confusing to me due to still getting used to `Option`), I think this reads nicely.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -793,10 +839,25 @@ class HttpGoogleServicesDAO(
...
+      }) map { googleOperation =>
+        RawlsBillingProjectOperationRecord(projectName.value, service, googleOperation.getName, toScalaBool(googleOperation.getDone), Option(googleOperation.getError).map(error => toErrorMessage(error.getMessage, error.getCode)), API_SERVICE_MANAGEMENT)
+      }})
 
+    } yield {
+      operations
+    })
+  }
+
+  override def completeProjectSetup(project: RawlsBillingProject): Future[Try[Unit]] = {
```
breilly2, line 853: It's unclear to me what the division is between `beginProjectSetup` and `completeProjectSetup` such that, if I were to add something, I wouldn't be sure where it should happen. Could you add some docs to explain the difference?

breilly2 (changes requested):
Overall, this looks like what I would expect based on my knowledge of the problem. I'm guessing that there might be other Google operations beyond billing projects that we might want some of this. I could even imagine this generalized to a system-wide event bus that could be used to orchestrate long-running processes with hooks for providing feedback in the UI, but that can be for a later time.

dvoet:
@breilly2 ready for re-review

breilly2 (approved):
Added comments are perfect, and I appreciate the added test for creating workspaces projects that aren't ready. Suggestion for the future: should the UI not allow a user to choose a project that isn't ready?

dvoet:
seems like the UI already does not allow selection of Creating project, need it to filter out Error as well.

PR #624, opened 2017-02-03
[FEEDBACK PLZ] Attempts at generifying ErrorReport

helgridly:
ErrorReport exists in both Rawls and Orchestration, and they are very similar. I took a stab at updating RawlsModel so that each project could define their own ErrorReport object (and thus specify a different "source" string).

I wanted to avoid having to update every instance of ErrorReport(...) to RawlsErrorReport(...). Distinguishing between the case class and the per-project singleton was difficult to impossible during imports, so I chose to rename the case class to something else. Hence CErrorReport.

TLDR anyone who wants our error reporting mechanisms can import rawlsModel and then define:

```
object ErrorReport extends ErrorReportBase {
  def SOURCE = "myproj"
}
```

and it should all work. But if they're passing around the type they have to refer to it as CErrorReport.

I'm not sure how much I like this. Thoughts please?

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ErrorReportable.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ErrorReportable.scala
@@ -1,23 +1,22 @@
 package org.broadinstitute.dsde.rawls.dataaccess
 
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
-import org.broadinstitute.dsde.rawls.model._
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
 import spray.http.StatusCodes
 import spray.httpx.UnsuccessfulResponseException
 
 trait ErrorReportable {
-  val errorReportSource: String
+  def errorReportSource: ErrorReportSource
 
   def toErrorReport(throwable: Throwable) = {
     throwable match {
       case gjre: GoogleJsonResponseException =>
         val statusCode = StatusCodes.getForKey(gjre.getStatusCode)
-        ErrorReport(errorReportSource, ErrorReport.message(gjre), statusCode, ErrorReport.causes(gjre), Seq.empty, Option(gjre.getClass))
+        ErrorReport(ErrorReport.message(gjre), statusCode, ErrorReport.causes(gjre), Seq.empty, Option(gjre.getClass))(errorReportSource)
       case ure: UnsuccessfulResponseException =>
-        ErrorReport(errorReportSource, ErrorReport.message(ure), Option(ure.response.status), ErrorReport.causes(throwable), Seq.empty, Option(ure.getClass))
+        ErrorReport(ErrorReport.message(ure), Option(ure.response.status), ErrorReport.causes(throwable), Seq.empty, Option(ure.getClass))(errorReportSource)
       case _ =>
-        ErrorReport(errorReportSource, ErrorReport.message(throwable), None, ErrorReport.causes(throwable), throwable.getStackTrace, Option(throwable.getClass))
+        ErrorReport(ErrorReport.message(throwable), None, ErrorReport.causes(throwable), throwable.getStackTrace, Option(throwable.getClass))(errorReportSource)
     }
   }
-
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ErrorReportable.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ErrorReportable.scala
@@ -1,7 +1,7 @@
 package org.broadinstitute.dsde.rawls.dataaccess
 
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
-import org.broadinstitute.dsde.rawls.model.ErrorReport
+import org.broadinstitute.dsde.rawls.model._
 import spray.http.StatusCodes
 import spray.httpx.UnsuccessfulResponseException
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
@@ -1,0 +1,5 @@
+package org.broadinstitute.dsde.rawls.model
+
+object ErrorReport extends ErrorReportBase {
+  def SOURCE = "rawls"
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
@@ -1,5 +1,0 @@
-package org.broadinstitute.dsde.rawls.model
-
-object ErrorReport extends ErrorReportBase {
-  def SOURCE = "rawls"
-}
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
@@ -54,6 +54,7 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system) with FlatSpe
   var subTerminalWorkflow = UUID.randomUUID().toString
   var subOneMissingWorkflow = UUID.randomUUID().toString
   var subTwoGoodWorkflows = UUID.randomUUID().toString
+  var subCromwellBadWorkflows = UUID.randomUUID().toString
 
   val subTestData = new SubmissionTestData()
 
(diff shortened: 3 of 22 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
```diff
@@ -0,0 +1,5 @@
+package org.broadinstitute.dsde.rawls.model
+
+object ErrorReport extends ErrorReportBase {
+  def SOURCE = "rawls"
+}
```
helgridly: this gives Rawls the `ErrorReport(...)` singleton

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -251,34 +251,45 @@ case class WorkspaceStatus(workspaceName: WorkspaceName, statuses: Map[String, S
 
 case class BucketUsageResponse(usageInBytes: BigInt)
 
-case class ErrorReport(source: String, message: String, statusCode: Option[StatusCode], causes: Seq[ErrorReport], stackTrace: Seq[StackTraceElement], exceptionClass: Option[Class[_]])
+case class CErrorReport(source: String, message: String, statusCode: Option[StatusCode], causes: Seq[CErrorReport], stackTrace: Seq[StackTraceElement], exceptionClass: Option[Class[_]])
```
helgridly: Also calling this `ErrorReport` made the compiler very unhappy, so I renamed it.

breilly2: I don't like this name, especially since you used a suffix when choosing `ErrorReportBase` (presumably to make it easy to find when looking for things that start with "ErrorReport"). If changing `ErrorReport` to `RawlsErrorReport` is the right thing to do and avoids having to rename this, then I think it should be done. To make the review easier, you could do that rename as a separate PR ahead of the rest of these changes.

helgridly:
I also don't like CErrorReport. But I strongly dislike RawlsErrorReport: we
are in the Rawls codebase, so of course it's a Rawls error report. It's
unnecessarily wordy and redundant.

The philosophically ideal answer IMO is to call them both ErrorReport and
figure out the imports. But I think that makes it a pain for devs and leads
to weird syntax things to disambiguate, though I confess I haven't checked.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
dvoet: I think this is the thing should be called RawlsErrorReport, it is the part that is rawls specific.

helgridly:
@dvoet yes, that's what I was referring to. But I'm still not convinced it
needs to be.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
helgridly: I also don't like CErrorReport. But I strongly dislike RawlsErrorReport: we
are in the Rawls codebase, so of course it's a Rawls error report. It's
unnecessarily wordy and redundant.

The philosophically ideal answer IMO is to call them both ErrorReport and
figure out the imports. But I think that makes it a pain for devs and leads
to weird syntax things to disambiguate, though I confess I haven't checked.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReport.scala
helgridly: @dvoet yes, that's what I was referring to. But I'm still not convinced it
needs to be.

davidangb: +1 to doug's comment. This is the service-specific class, so it can/should have a service-specific name.

davidangb:
1. Is there a way to set the `SOURCE` string as an implicit somewhere, instead of creating a new class/object/whatever just to hold the source string?

2. Have you tried using this in orchestration? This refactors the model and proves it serves Rawls; does it work for orch?

helgridly:
1. Is a good idea. I really like that.
2. Not yet. This is part of a larger WIP; you'll notice this PR is into a Rawls feature branch, not dev.

helgridly:
@dvoet @davidangb @breilly2 I like this a LOT more. All services using ErrorReport (i.e. rawls, orch) simply define an implicit in `package.scala` with an `ErrorReportSource` object in it and it all Just Works™️.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ErrorReportable.scala
```diff
@@ -1,23 +1,22 @@
...
 
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
-import org.broadinstitute.dsde.rawls.model.ErrorReport
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
 import spray.http.StatusCodes
 import spray.httpx.UnsuccessfulResponseException
 
 trait ErrorReportable {
-  val errorReportSource: String
+  def errorReportSource: ErrorReportSource
```
helgridly, line 9: I had extenders of this trait define their own implicit errorReportSource, but the implicit lookup-ness was finding the one in dsde.rawls.model's package.scala. Making it be abstract in this class and then used explicitly solved that problem.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -253,32 +253,43 @@ case class BucketUsageResponse(usageInBytes: BigInt)
...
+    ErrorReport(source.source,message,None,Seq(cause),Seq.empty, None)
 
-  def apply(statusCode: StatusCode, message: String, cause: ErrorReport): ErrorReport =
-    new ErrorReport(SOURCE,message,Option(statusCode),Seq(cause),Seq.empty, None)
+  def apply(message: String, causes: Seq[ErrorReport])(implicit source: ErrorReportSource): ErrorReport =
+    ErrorReport(source.source,message,None,causes,Seq.empty, None)
 
-  def apply(statusCode: StatusCode, message: String, causes: Seq[ErrorReport]): ErrorReport =
-    new ErrorReport(SOURCE,message,Option(statusCode),causes,Seq.empty, None)
+  def apply(statusCode: StatusCode, throwable: Throwable)(implicit source: ErrorReportSource): ErrorReport =
```
breilly2, line 265: This doesn't appear to be used and effectively duplicates `apply(throwable: Throwable, statusCode: StatusCode)`. I think I prefer the `apply(statusCode: StatusCode, throwable: Throwable)` signature here, but please either remove this one or migrate calls to this one and remove the other.

`apply(statusCode: StatusCode, message: String, throwable: Throwable)` (on like 274 below) also appears to be new but is not redundant with any others so I have no objections to it.

breilly2:
Are the changes in SubmissionSpec and RemoteServicesMockServer related to the refactor? It's not clear to my what that's about so I'm wondering if I need to review those changes.

helgridly:
@breilly2 Yes. They check that my changes to ErrorReportable work correctly - i.e. that other services set their source. This was never tested before but now it is.

(Mainly I was testing the pattern, which is why it's done for Cromwell but not Agora.)

helgridly:
Also FWIW I'd like to remind folks that this isn't a PR into dev. It's a PR into another WIP feature branch I have; I was looking for "does this look like the right structure" feedback. You can save the nitpicks for later ;)

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/RemoteServicesMockServer.scala
```diff
@@ -534,5 +534,25 @@ class RemoteServicesMockServer(port:Int) extends RawlsTestUtils {
       )
   }
 
+  mockServer.when(
+    request()
+      .withMethod("GET")
+      .withPath("/workflows/v1/29b2e816-ecaf-11e6-b006-92361f002671/outputs")
```
breilly2, line 540: I didn't notice that this was the same as `badLogsAndMetadataWorkflowId`. Using string interpolation would help clarify that.

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -526,6 +533,19 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system) with FlatSpe
...
+    val rqComplete = workspaceService.workflowOutputs(
+      subTestData.wsName,
+      subTestData.submissionTestCromwellBadWorkflows.submissionId,
+      subTestData.badLogsAndMetadataWorkflowId.get)
+    val errorReport = intercept[RawlsExceptionWithErrorReport] {
+      Await.result(rqComplete, Duration.Inf).asInstanceOf[RequestComplete[ErrorReport]].response
+    }
+
+    assertResult(StatusCodes.BadGateway) {errorReport.errorReport.statusCode.get}
+    assertResult("cromwell"){errorReport.errorReport.causes.head.source}
```
breilly2, line 546: Thank you for testing this!

breilly2 (approved):
Better than the first round and a good improvement. Only two minor requests so I'll give this a :+1:.

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/RemoteServicesMockServer.scala
helgridly: hmm. this'd require some refactoring to Do It Right. i'll put in a tech debt ticket

PR #630, opened 2017-02-20
GAWB-1206: MethodConfiguration Provenance

ansingh7115:
# Method Configuration Provenance

**Why are we doing this?**
Currently, method configs in FireCloud are overwritten anytime they are edited, making it impossible to track changes. For example, if we have submission on a particular method config and the method config is edited, the submission is now pointing to a method config that is different from the one it actually ran on. **(1)** The use case we want to support is one where the user can look at a submission run on an method config that was edited and can click method config from the submission page and see all the relevant information about the method config. **(2)** The second use case that we may (or may not) support in the future will allow users to go to a submission page and revert it's method config so that more submissions may be run on it.

**What's the implementation?**
- Liquibase change: Adding a METHOD_CONFIG_VERSION column, which defaults as 1 for new method configs.
- Changes to case classes MethodConfiguration and MethodConfigurationRecord to add a methodConfigVersion field
- Change to save(...) method in MethodConfigurationComponent so that, instead of overwriting a method config on edit, we flip the old config to deleted and append the date and time to the name. Then we add a new row for the new method config (with the version incremented) and add the inputs and outputs. 
- Change to the rename process for method Configs was changed so that it too creates a new row in the db for renamed method configs and deletes/hides the old method config. The overlapping parts of the save and rename processes were broken out into their own methods

**Why is this the implementation?**
It took some time to decide on this implementation because others were more attractive in their thoroughness. In the end, more complicated implementations were cast aside because the 2 use cases did not require this added complexity. If we are only considered the 2 uses cases described above, this implementation is sufficient. There is no reason to differentiate a user deleted vs. archived method config. 

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="method_config_version_column" author="ansingh" logicalFilePath="dummy">
+        <addColumn tableName="METHOD_CONFIG">
+            <column name="METHOD_CONFIG_VERSION" type="INT(11)" defaultValue="1" />
+        </addColumn>
+    </changeSet>
+
+</databaseChangeLog>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
@@ -3,7 +3,7 @@
 
     <changeSet id="method_config_version_column" author="ansingh" logicalFilePath="dummy">
         <addColumn tableName="METHOD_CONFIG">
-            <column name="METHOD_CONFIG_VERSION" type="INT(11)" defaultValue="1" />
+            <column name="METHOD_CONFIG_VERSION" type="INT" defaultValue="1" />
         </addColumn>
     </changeSet>
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
@@ -97,6 +97,9 @@ trait MethodConfigurationComponent {
     /*
       the core methods
      */
+
+    // the save method is used for both creating a new method config and updating an edited method config
+    // the save method is not used for renaming a method config, see the rename method for that
     def save(workspaceContext: SlickWorkspaceContext, methodConfig: MethodConfiguration): ReadWriteAction[MethodConfiguration] = {
 
       uniqueResult[MethodConfigurationRecord](findByName(workspaceContext.workspaceId, methodConfig.namespace, methodConfig.name)) flatMap {
@@ -120,7 +123,7 @@ trait MethodConfigurationComponent {
         (methodConfigurationOutputQuery ++= outputs)
     }
 
-    def saveNewVersion(workspaceContext: SlickWorkspaceContext, oldId: Long, oldName: String, oldConfigVersion: Int, newMethodConfig: MethodConfiguration) = {
+    private def saveNewVersion(workspaceContext: SlickWorkspaceContext, oldId: Long, oldName: String, oldConfigVersion: Int, newMethodConfig: MethodConfiguration) = {
       workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
         hideMethodConfigurationAction(oldId, oldName) andThen
         (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) += marshalMethodConfigNewVersion(workspaceContext.workspaceId, newMethodConfig, oldConfigVersion + 1)) flatMap { configId =>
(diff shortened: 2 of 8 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="method_config_version_column" author="ansingh" logicalFilePath="dummy">
+        <addColumn tableName="METHOD_CONFIG">
+            <column name="METHOD_CONFIG_VERSION" type="INT(11)" defaultValue="1" />
```
helgridly: i don't think we size our INTs.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -96,35 +97,44 @@ trait MethodConfigurationComponent {
...
+      val prerequisites = methodConfig.prerequisites.map { case (key, value) => marshalConfigPrereq(configId, key, value) }
+      val inputs = methodConfig.inputs.map { case (key, value) => marshalConfigInput(configId, key, value) }
+      val outputs = methodConfig.outputs.map{ case (key, value) => marshalConfigOutput(configId, key, value) }
+
+      (methodConfigurationPrereqQuery ++= prerequisites) andThen
+        (methodConfigurationInputQuery ++= inputs) andThen
+        (methodConfigurationOutputQuery ++= outputs)
+    }
+
+    private def saveNewVersion(workspaceContext: SlickWorkspaceContext, oldId: Long, oldName: String, oldConfigVersion: Int, newMethodConfig: MethodConfiguration) = {
```
helgridly: I find this function confusing -- I think you need to make it clearer what "old" and "new" mean here. I think "old" means "current", right? You might want to pass in the entire `methodConfigRec` (above) representing the current-version to save on parameters and make it a bit clearer. I think this method could do with a comment to explain what it does: "hides the current record and make a new record, incrementing the version from the current". You should also make it clear that this function expects them to share a name and namespace, since the function makes absolutely no sense if this isn't true.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -96,35 +97,44 @@ trait MethodConfigurationComponent {
...
+      workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
+        hideMethodConfigurationAction(oldId, oldName) andThen
+        (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) += marshalMethodConfigNewVersion(workspaceContext.workspaceId, newMethodConfig, oldConfigVersion + 1)) flatMap { configId =>
```
helgridly: Can you instead do this:

```marshalMethodConfig(workspaceContext.workspaceId, newMethodConfig.copy(version=oldConfigVersion + 1) )```

That'd let you drop `marshalMethodConfigNewVersion` entirely.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala [hunk 4]
```diff
@@ -133,9 +143,14 @@ trait MethodConfigurationComponent {
...
     }
 
-    def rename(workspaceContext: SlickWorkspaceContext, methodConfigurationNamespace: String, methodConfigurationName: String, newName: String): ReadWriteAction[Int] = {
-      workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
-        findByName(workspaceContext.workspaceId, methodConfigurationNamespace, methodConfigurationName).map(_.name).update(newName)
+    // To rename a method config, we "delete" (or rather hide) the old version of the method config and create a new row for the new method config, with the
+    //  config version # incremented
+    def rename(workspaceContext: SlickWorkspaceContext, methodConfigurationNamespace: String, methodConfigurationName: String, newMethodConfig: MethodConfiguration) = {
+      uniqueResult[MethodConfigurationRecord](findByName(workspaceContext.workspaceId, methodConfigurationNamespace, methodConfigurationName)) flatMap {
+        case None => DBIO.successful(false)
```
helgridly: Does this turn into a 404 up the chain?

helgridly: And why does this return `false` here? What's the return type of `saveNewVersion`? Are they consistent?

helgridly: Yeah, so we've already looked up this method configuration [here](https://github.com/broadinstitute/rawls/pull/630/files#diff-91fe8efb10dfe91234a26e4f80039fc6R1083). If we can't find the record for it now, having just found it, then something has gone severely screwy.

Also it's a bit poo that we're looking up the method configuration from the database twice. You could change the call to `withMethodConfig` to a new `withMethodConfigRec` so you can just pass it all the way down.

helgridly (changes requested):

helgridly:
I think you've got the right approach here -- just needs a little streamlining.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -191,7 +206,7 @@ trait MethodConfigurationComponent {
     }
 
     private def findActiveByWorkspace(workspaceId: UUID): MethodConfigurationQueryType = {
-      findByWorkspace(workspaceId).filter(rec => rec.deleted === false)
+      findByWorkspace(workspaceId).filter(rec => !rec.deleted)
```
jmthibault79: `filterNot(_.deleted)`

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170219_method_config_versions.xml
ansingh7115: Actually we do. In the rawls db, there is either INT(11) or BIGINT(20). For example in entity attribute:
screen shot 2017-02-23 at 11 10 24 am

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
ansingh7115: The issue I had  with getting the method config record in WorkspaceService is that I'm assuming don't use MethodConfigurationRecord outside of MethodConfigurationComponent, in which case maybe it makes more sense to  find the record here and change the method that calls rename in WorkspaceService so it doesn't have withMethodConfig?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
helgridly: Nice.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -96,34 +97,43 @@ trait MethodConfigurationComponent {
...
-      uniqueResult[MethodConfigurationRecord](findByName(workspaceContext.workspaceId, methodConfig.namespace, methodConfig.name)) flatMap {
+      uniqueResult[MethodConfigurationRecord](findByName(workspaceContext.workspaceId, newMethodConfig.namespace, newMethodConfig.name)) flatMap {
         case None =>
-          val configInsert = (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) +=  marshalMethodConfig(workspaceContext.workspaceId, methodConfig))
+          val configInsert = (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) +=  marshalMethodConfig(workspaceContext.workspaceId, newMethodConfig))
           configInsert flatMap { configId =>
-            saveMaps(configId)
+            saveMaps(newMethodConfig, configId)
           }
         case Some(methodConfigRec) =>
```
helgridly: Nitpick: can you rename this to "current" or "existing" methodConfigRec? To contrast with "new"?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -96,34 +97,43 @@ trait MethodConfigurationComponent {
...
+      val prerequisites = methodConfig.prerequisites.map { case (key, value) => marshalConfigPrereq(configId, key, value) }
+      val inputs = methodConfig.inputs.map { case (key, value) => marshalConfigInput(configId, key, value) }
+      val outputs = methodConfig.outputs.map{ case (key, value) => marshalConfigOutput(configId, key, value) }
+
+      (methodConfigurationPrereqQuery ++= prerequisites) andThen
+        (methodConfigurationInputQuery ++= inputs) andThen
+        (methodConfigurationOutputQuery ++= outputs)
+    }
+
+    private def saveNewVersion(workspaceContext: SlickWorkspaceContext, methodConfigRec: MethodConfigurationRecord, newMethodConfig: MethodConfiguration) = {
```
helgridly: ditto here

helgridly:
A couple of variable-rename requests but otherwise 👍

helgridly (approved):

PR #631, opened 2017-02-23
extract google pubsub

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-1653

The primary purpose of this PR is to pull the google pubsub code into a place that can then be used by thurloe. This is done by following the pattern set for the model library. Two new libraries have been created: google and util. Next step is to either publish these to artifactory or pull them into new repos and use google submodules. I am going to explore the former first.

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


model/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReportable.scala
```diff
@@ -1,7 +1,6 @@
-package org.broadinstitute.dsde.rawls.dataaccess
+package org.broadinstitute.dsde.rawls.model
```
dvoet, line 1: I really feel like all the ErrorReport stuff goes in util but not ready to do that yet. That will come when PerRequest is moved to util.

helgridly: Agreed.

build.sbt
```diff
@@ -2,17 +2,30 @@ import Settings._
 import Testing._
 
 
+lazy val rawlsUtil = project.in(file("util"))
```
helgridly: do we want to think about not calling these rawlsFoo in favour of workbenchFoo or something?

build.sbt
```diff
@@ -2,17 +2,30 @@ import Settings._
...
+  .settings(utilSettings:_*)
+  .withTestSettings
+
 lazy val rawlsModel = project.in(file("model"))
   .settings(modelSettings:_*)
   .withTestSettings
 
+lazy val rawlsGoogle = project.in(file("google"))
+  .settings(googleSettings:_*)
+  .dependsOn(rawlsModel)
```
helgridly: Really? That's a shame.

build.sbt
```diff
@@ -2,17 +2,30 @@ import Settings._
...
 lazy val rawlsCore = project.in(file("core"))
   .settings(rawlsCoreSettings:_*)
   .dependsOn(rawlsModel)
+  .dependsOn(rawlsGoogle)
+  .dependsOn(rawlsUtil)
   .withTestSettings
 
 lazy val rawls = project.in(file("."))
   .settings(rootSettings:_*)
+  .aggregate(rawlsGoogle)
```
helgridly: Definitely run and test assembly here. The difference between aggregate and dependsOn tripped me up a lot.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -94,8 +94,8 @@ trait GoogleUtilities extends LazyLogging with Retry {
   }
 
   private def logGoogleRequest[A](request: AbstractGoogleClientRequest[A], startTime: Long, statusCode: Option[Int], errorReport: Option[ErrorReport]): Unit = {
-    import spray.json._
     import GoogleRequestJsonSupport._
+    import spray.json._
```
helgridly, line 98: In-function imports shouldn't really be necessary any more.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReportable.scala
```diff
@@ -12,11 +11,11 @@ trait ErrorReportable {
     throwable match {
       case gjre: GoogleJsonResponseException =>
         val statusCode = StatusCodes.getForKey(gjre.getStatusCode)
-        ErrorReport(ErrorReport.message(gjre), statusCode, ErrorReport.causes(gjre), Seq.empty, Option(gjre.getClass))(errorReportSource)
+        ErrorReport(ErrorReport.message(gjre), statusCode, ErrorReport.causes(gjre)(errorReportSource), Seq.empty, Option(gjre.getClass))(errorReportSource)
```
helgridly, line 14: Shouldn't this be implicit-y?

helgridly (approved):
no blocking changes, but some questions and nitpicks

build.sbt
dvoet: because of ErrorReport stuff which will get moved to util soon I hope

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
dvoet: since GoogleRequestJsonSupport is a protected member, it can't go at the top so I think I will let this ride... in this particular case it is only in the function because it is only used here

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ErrorReportable.scala
dvoet: I thought the whole point of this class was to define a source other than the one defined implicit-y. Before the code was moved it was using the local errorReportSource for the top ErrorReport but the implicit source for the causes. So, it might look like the top is from "google" and the causes are from "rawls." This change makes the whole stack come from "google" (or whatever errorReportSource is set to be).

helgridly: Hmm. That being the case, can you update [this test](https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala#L546) with whatever's appropriate?

dvoet: I would need a special test for this, that test case does not have an exception that lists causes

dvoet:
@helgridly I added a test to cover an error report with a cause that has a root cause. I also changed where we publish to org.broadinstitute.dsde (I added dsde) and I changed the rawls-util and rawls-google artifacts to be workbench-util and workbench-google. And I fiddled with the publish scripts.

helgridly:
👍 please update orch to refer to the relocated rawlsModel once you've merged this

PR #633, opened 2017-02-24
Make rawls build quicker and use lighter docker image

jacmrob:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM broadinstitute/scala-baseimage
+FROM openjdk:8
 
 # Rawls' default port
 EXPOSE 8080
@@ -6,12 +6,10 @@ EXPOSE 5050
 
 ENV GIT_MODEL_HASH $GIT_MODEL_HASH
 
-ADD . /rawls
+RUN mkdir /rawls
+COPY ./rawls.jar /rawls
 
-# catch sbt issues separately
-RUN cd /rawls && sbt update && echo "sbt updated successfully."
-
-RUN ["/bin/bash", "-c", "/rawls/docker/install.sh /rawls"]
+#RUN ["/bin/bash", "-c", "/rawls/docker/install.sh /rawls"]
 
 # Add Rawls as a service (it will start when the container starts)
 RUN mkdir /etc/service/rawls
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM openjdk:8
 
+# To run, must build the jar using ./docker/build.sh
+
 # Rawls' default port
 EXPOSE 8080
 EXPOSE 5050
@@ -10,7 +12,7 @@ RUN mkdir /rawls
 COPY ./rawls-assembly-0.1.jar /rawls
 
 # Add Rawls as a service (it will start when the container starts)
-RUN mkdir /etc/service/rawls
+RUN mkdir -p /etc/service/rawls
 ADD docker/run.sh /etc/service/rawls/run
 
 # These next 4 commands are for enabling SSH to the container.
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,10 @@ EXPOSE 5050
 ENV GIT_MODEL_HASH $GIT_MODEL_HASH
 
 RUN mkdir /rawls
-COPY ./rawls-assembly-0.1.jar /rawls
+COPY ./rawls*.jar /rawls
 
 # Add Rawls as a service (it will start when the container starts)
-RUN mkdir -p /etc/service/rawls
-ADD docker/run.sh /etc/service/rawls/run
+CMD java $JAVA_OPTS -jar $(find /rawls -name 'rawls*.jar')
 
 # These next 4 commands are for enabling SSH to the container.
 # id_rsa.pub is referenced below, but this should be any public key
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,7 @@ EXPOSE 5050
 ENV GIT_MODEL_HASH $GIT_MODEL_HASH
 
 RUN mkdir /rawls
-COPY ./rawls.jar /rawls
-
-#RUN ["/bin/bash", "-c", "/rawls/docker/install.sh /rawls"]
+COPY ./rawls-assembly-0.1.jar /rawls
 
 # Add Rawls as a service (it will start when the container starts)
 RUN mkdir /etc/service/rawls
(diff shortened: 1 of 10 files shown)


docker/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Run tests and make jar
+# Uses docker.
+
+set -eux
+
+# check if mysql is running
+mysql_running=$(docker inspect -f {{.State.Running}} mysql)
+if ! $mysql_running ; then
```
jmthibault79: The check for `Running` is not sufficient.  An "exited" mysql container on my Mac caused this to fail.

jmthibault79: After removing the `mysql` container, `docker inspect` also failed because it was missing.

docker/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    # start up mysql
+    docker pull mysql/mysql-server:5.7.15
+    docker run --name mysql -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d -p 3310:3306 mysql/mysql-server:5.7.15
+
+    # validate mysql
+    sudo yum -y install mysql
```
jmthibault79: Will this run on Macs?

docker/install.sh
```diff
@@ -10,7 +10,7 @@ if [ -e "jenkins_env.sh" ]; then
 fi
 
 # Tests are run in jenkins which has a custom mysql instance just for testing
-sbt -J-Xms4g -J-Xmx4g test -Dmysql.host=172.17.42.1 -Dmysql.port=3310
+sbt -J-Xms4g -J-Xmx4g test -Dmysql.host=mysql -Dmysql.port=3306
```
helgridly, line 13: how does `host=mysql` work?

docker/build.sh
jacmrob: good call!  fixing to make this more robust.

docker/build.sh
jacmrob: nope. i think the safest thing to do is run this validation through a docker container that has mysql

docker/install.sh
jacmrob: `mysql` is the name of the docker container.  because we're also running these tests through docker and the containers are linked, we don't have to use a static IP here anymore

docker/build.sh
jmthibault79: Would this mean we no longer have a way to build docker images locally?  What would be the best way for devs to build images for local/experimental code?

docker/build.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "attempting to remove old mysql container..."
+    docker rm -f $CONTAINER || echo "docker rm failed. nothing to rm."
+
+    # start up mysql
+    echo "starting up mysql container..."
+    docker run --name $CONTAINER -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d -p 3310:3306 mysql/mysql-server:5.7.15
+
+    # validate mysql
+    echo "running mysql validation..."
+    docker run --rm --link mysql:mysql -v $PWD/docker/sql_validate.sh:/working/sql_validate.sh broadinstitute/dsde-toolbox /working/sql_validate.sh
```
dmohs, line 24: Pulling this validation out into a separate script was an excellent choice.

dmohs (commented):
The improvements to efficiency and robustness here are substantial. My only concern is that the container will not run in the same way as it had before. We would need a `CMD` or `ENTRYPOINT` in the Dockerfile or a corresponding compose YAML change.

Dockerfile
```diff
@@ -1,20 +1,18 @@
...
-RUN cd /rawls && sbt update && echo "sbt updated successfully."
-
-RUN ["/bin/bash", "-c", "/rawls/docker/install.sh /rawls"]
+RUN mkdir /rawls
+COPY ./rawls-assembly-0.1.jar /rawls
 
 # Add Rawls as a service (it will start when the container starts)
-RUN mkdir /etc/service/rawls
+RUN mkdir -p /etc/service/rawls
 ADD docker/run.sh /etc/service/rawls/run
```
dmohs: This and the previous line are really artifacts of how the previous base image worked and don't make sense here. I would remove them, along with `docker/run.sh`, and add a `CMD` directive. I think `CMD` is more appropriate than `ENTRYPOINT` for this case.

jacmrob: @dmohs do you mean to get rid of the `docker/run.sh` script altogether, and put its contents in the `CMD`, or to copy it into the docker image and run it as the `CMD`?

dmohs: That is a judgement call, but my guess is that it would be more idiomatic docker practice to add the entire thing as the `CMD` directive.

dmohs: (and remove the script file)

Dockerfile
```diff
@@ -1,21 +1,18 @@
...
-RUN cd /rawls && sbt update && echo "sbt updated successfully."
-
-RUN ["/bin/bash", "-c", "/rawls/docker/install.sh /rawls"]
+RUN mkdir /rawls
+COPY ./rawls*.jar /rawls
 
 # Add Rawls as a service (it will start when the container starts)
-RUN mkdir /etc/service/rawls
-ADD docker/run.sh /etc/service/rawls/run
+CMD java $JAVA_OPTS -jar $(find /rawls -name 'rawls*.jar')
```
dmohs, line 15: This looks perfect. Have you tested? Just to make sure Docker's translation from Dockerfile command to shell is correct?

dmohs (approved):
One question but this looks good to go to me.

Dockerfile
```diff
@@ -1,21 +1,18 @@
-FROM broadinstitute/scala-baseimage
+FROM openjdk:8
```
jmthibault79, line 1: This is not working for me.

```
$ docker build .
Sending build context to Docker daemon 124.7 MB
Step 1/7 : FROM openjdk:8
manifest for openjdk:8 not found
```

jmthibault79 (changes requested):

Dockerfile
jmthibault79: I'm hearing that Dockerhub may be down, which would explain this...

jmthibault79: Works!  👍

jmthibault79 (approved):

PR #636, opened 2017-03-01
GAWB-1206: Absorb Rename method config into Update

ansingh7115:
Currently, when editing a method configuration, renaming is a separate path from other updates to the method config. Here, rename is absorbed into the general update process.

- The APIs are left alone
- WorkspaceService has renameMethodConfig point to update 
- MethodConfigurationComponent - create and update method both point to save

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet id="method_config_drop_index" author="ansingh" logicalFilePath="dummy">
+        <sql stripComments="true">
+            DROP INDEX IDX_CONFIG ON METHOD_CONFIG;
+        </sql>
+    </changeSet>
+
+    <changeSet id="method_config_create_index" author="ansingh" logicalFilePath="dummy">
+        <sql stripComments="true">
+            CREATE INDEX IDX_CONFIG ON METHOD_CONFIG (NAMESPACE, NAME, WORKSPACE_ID, METHOD_CONFIG_VERSION);
+        </sql>
+    </changeSet>
+
+
+</databaseChangeLog>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
@@ -8,7 +8,7 @@
 
     <changeSet id="method_config_create_index" author="ansingh" logicalFilePath="dummy">
         <sql stripComments="true">
-            CREATE INDEX IDX_CONFIG ON METHOD_CONFIG (NAMESPACE, NAME, WORKSPACE_ID, METHOD_CONFIG_VERSION);
+            CREATE UNIQUE INDEX IDX_CONFIG ON METHOD_CONFIG (WORKSPACE_ID, NAMESPACE, NAME, METHOD_CONFIG_VERSION);
         </sql>
     </changeSet>
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
@@ -24,7 +24,7 @@ case class MethodConfigurationOutputRecord(methodConfigId: Long, id: Long, key: 
 
 case class MethodConfigurationPrereqRecord(methodConfigId: Long, id: Long, key: String, value: String)
 
-trait MethodConfigurationComponent {
+trait MethodConfigurationComponent  {
   this: DriverComponent with WorkspaceComponent =>
 
   import driver.api._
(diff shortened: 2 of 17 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -223,6 +228,10 @@ trait MethodConfigurationComponent {
     }
 
     def findByName(workspaceId: UUID, methodNamespace: String, methodName: String): MethodConfigurationQueryType = {
+      filter(rec => rec.namespace === methodNamespace && rec.name === methodName && rec.workspaceId === workspaceId)
```
MatthewBemis: it doesn't look like `findByName` is used anymore now that you've added `findActiveByName`. remove it?

MatthewBemis (approved):
one small comment. I'll let you decide whether or not you want to address it

:+1:

helgridly:
^ please do address matt's comment :)

I think if we do two updates inside a second we'll still get a constraint violation exception. We talked about adding the method config version to the unique key, which in turn means we don't have to add the timestamp when archiving a config. Can we do this?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
ansingh7115: I'll remove it!

ansingh7115:
Yes, I forgot about that. I will change the unique index

helgridly:
Since you're making db changes, also remember to take backups of the dev/alpha databases. And put a note in the release candidate ticket suggesting folks do so for production too.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -631,7 +631,8 @@ class WorkspaceServiceSpec extends FlatSpec with ScalatestRouteTest with Matcher
     }
 
     //Check method configs to be deleted exist
-    assertResult(Vector(MethodConfigurationShort("testConfig2","Sample",MethodRepoMethod("myNamespace","method-a",1),"dsde"), MethodConfigurationShort("testConfig1","Sample",MethodRepoMethod("ns-config","meth1",1),"ns"))) {
+    assertResult(Vector(MethodConfigurationShort("testConfig1","Sample",MethodRepoMethod("ns-config","meth1",1),"ns"),
```
helgridly: you can use [assertSameElements](https://github.com/broadinstitute/rawls/blob/b49619dc8400dbdf67153ea5260eea79d6d4133d/core/src/test/scala/org/broadinstitute/dsde/rawls/RawlsTestUtils.scala#L63) to sidestep ordering annoyances if they don't actually matter

ansingh7115: Thanks!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
```diff
@@ -0,0 +1,16 @@
...
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet id="method_config_drop_index" author="ansingh" logicalFilePath="dummy">
+        <sql stripComments="true">
+            DROP INDEX IDX_CONFIG ON METHOD_CONFIG;
+        </sql>
+    </changeSet>
+
+    <changeSet id="method_config_create_index" author="ansingh" logicalFilePath="dummy">
+        <sql stripComments="true">
+            CREATE INDEX IDX_CONFIG ON METHOD_CONFIG (NAMESPACE, NAME, WORKSPACE_ID, METHOD_CONFIG_VERSION);
```
MatthewBemis: a few things:

- you'll also want to update the index that's defined in `MethodConfigurationTable` in `MethodConfigurationComponent.scala`

- this new index is no longer unique. `CREATE UNIQUE INDEX ...` should work

- ordering the columns in the index as `WORKSPACE_ID, NAMESPACE, NAME, METHOD_CONFIG_VERSION` would allow for the index to be used by more queries. the specific one that i'm thinking about is listing all method configs for a workspace- which wouldn't be able to use the index in it's current order because `WORKSPACE_ID` is not a prefix of the index. also, the change in column order is what likely caused the results of your tests to change

MatthewBemis (changes requested):
a few comments about the index

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20170302_add_method_config_version_to_unique_index.xml
ansingh7115: All good suggestions

MatthewBemis (approved):
👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -167,8 +171,9 @@ trait MethodConfigurationComponent {
         }
     }
 
-    def hideMethodConfigurationAction(id: Long, name: String): ReadWriteAction[Int] = {
-      findById(id).map(rec => (rec.deleted, rec.name)).update(true, name + "_" + DateTime.now().toString("yyyy-MM-dd_HH:mm:ss"))
+    def hideMethodConfigurationAction(id: Long, methodConfigName: String): ReadWriteAction[Int] = {
+      findById(id).map(rec => (rec.deleted, rec.name))
+        .update(true, methodConfigName + "_" + DateTime.now().toString("yyyy-MM-dd_HH:mm:ss"))
```
helgridly, line 176: Is there any need for the timestamp now?

helgridly: And does trashing the timestamp mean having to migrate existing data? I don't think so...

helgridly: After much in-person conversation, yes the timestamps are needed. The submission record references method config by database ID, but the case class uses the MC name, so trashing the timestamp makes us lose information (and potentially repoint submissions to the wrong method configurations on resave).

PR #645, opened 2017-03-14
[DNM] branch feedback: working on queries

davidangb:
Sarah - take a look at this for your db queries.

I haven't done a comprehensive review of your branch - just focused on the queries.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -269,6 +269,14 @@ trait AttributeComponent {
       }
     }
 
+    def findUniqueStringsByNameQuery(attrName: AttributeName) = {
+      filter(rec =>
+        rec.namespace === attrName.namespace &&
+        rec.name === attrName.name &&
+        rec.valueString.isDefined
+      ).distinctOn(_.valueString.get)
+    }
+
     def deleteAttributeRecords(attributeRecords: Seq[RECORD]): DBIOAction[Int, NoStream, Write] = {
       filter(_.id inSetBind attributeRecords.map(_.id)).delete
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -274,7 +274,7 @@ trait AttributeComponent {
         rec.namespace === attrName.namespace &&
         rec.name === attrName.name &&
         rec.valueString.isDefined
-      ).distinctOn(_.valueString.get)
+      ).map(_.valueString).distinct
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -269,6 +269,14 @@ trait AttributeComponent {
       }
     }
 
+    def findUniqueStringsByNameQuery(attrName: AttributeName) = {
+      filter(rec =>
+        rec.namespace === attrName.namespace &&
+        rec.name === attrName.name &&
+        rec.valueString.isDefined
+      ).map(_.valueString).distinct
+    }
```
davidangb, line 278: after trying a few approaches, I really wasn't happy with the size of the SQL query, coupled with the amount of work we had to do in Scala to get the results we wanted. Here's a more optimized query.

This generates SQL of:
```
select distinct `value_string` from `WORKSPACE_ATTRIBUTE` where ((`namespace` = 'tag') and (`name` = 'tags')) and (`value_string` is not null)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -134,8 +134,10 @@ trait WorkspaceComponent {
       loadWorkspaces(workspaceQuery)
     }
 
-    def getAllTags() = { //: ReadAction[Seq[Attribute]] = {
-      workspaceAttributeQuery.findByNameQuery(AttributeName("tag", "tags")).result
+    def getAllTags(): ReadAction[Seq[String]] = {
+      workspaceAttributeQuery.findUniqueStringsByNameQuery(AttributeName("tag", "tags")).result map {recs =>
+        recs map (_.get)
+      }
```
davidangb, line 140: using `result` isn't necessarily wrong - just be conscious of where you use it!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -356,18 +356,9 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-          }
-
-          // fruitless type test: a value of type org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceAttributeRecord cannot also be a org.broadinstitute.dsde.rawls.model.AttributeValueList
-//          case ss: AttributeValueList => ss.list collect {
-//            case s: AttributeString => s.value // if (s.value.contains(query)) => s.value
-        }).flatten
-        RequestComplete(StatusCodes.OK, tagValues.toSet) // making it a set so there's no duplicates
+      dataAccess.workspaceQuery.getAllTags().map { tags =>
+        val filteredTags = tags.filter(_.toLowerCase.contains(query.toLowerCase))
+        RequestComplete(StatusCodes.OK, filteredTags.toSet)
```
davidangb, line 361: It may make more sense to push the `contains` here all the way down to the SQL, instead of returning a bunch of rows and then throwing them away in Scala.

PR #649, opened 2017-03-16
GAWB-1405: tag autocomplete (searches all tags in the db)

skwalker:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3390,7 +3390,7 @@ paths:
               - email
               - profile
               - https://www.googleapis.com/auth/devstorage.full_control
-  '/api/workspaces/tags':
+  '/api/workspaces/tags/autocomplete/{query}':
       get:
         responses:
           '200':
@@ -3409,6 +3409,12 @@ paths:
           - workspaces
         summary: Get tags
         operationId: getTags
+        parameters:
+          - in: path
+            description: Query
+            name: query
+            required: true
+            type: string
         security:
         - authorization:
           - openid
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3390,6 +3390,31 @@ paths:
               - email
               - profile
               - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/workspaces/tags':
+      get:
+        responses:
+          '200':
+            description: Successful Request
+            schema:
+              type: array
+              description: list of all tags used in FireCloud
+              items:
+                type: string
+          '500':
+            description: Rawls Internal Error
+            schema:
+              $ref: '#/definitions/ErrorReport'
+        description: Get all tags used in FireCloud (for autocomplete)
+        tags:
+          - workspaces
+        summary: Get tags
+        operationId: getTags
+        security:
+        - authorization:
+          - openid
+          - email
+          - profile
+          - https://www.googleapis.com/auth/devstorage.full_control
   '/register/user':
      post:
        responses:
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3390,7 +3390,7 @@ paths:
               - email
               - profile
               - https://www.googleapis.com/auth/devstorage.full_control
-  '/api/workspaces/tags/autocomplete/{query}':
+  '/api/workspaces/tags':
       get:
         responses:
           '200':
@@ -3410,17 +3410,16 @@ paths:
         summary: Get tags
         operationId: getTags
         parameters:
-          - in: path
-            description: Query
-            name: query
-            required: true
+          - in: query
+            description: Query string
+            name: q
+            required: false
             type: string
         security:
         - authorization:
           - openid
           - email
           - profile
-          - https://www.googleapis.com/auth/devstorage.full_control
   '/register/user':
      post:
        responses:
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3397,9 +3397,9 @@ paths:
             description: Successful Request
             schema:
               type: array
-              description: list of all tags used in FireCloud
+              description: list of all tags used in FireCloud and their associated frequencies
               items:
-                type: string
+                $ref: '#/definitions/WorkspaceTag'
           '500':
             description: Rawls Internal Error
             schema:
@@ -4990,3 +4990,15 @@ definitions:
       role:
         type: string
         description: the role of the user in the project
+  WorkspaceTag:
+    description: tag with count
+    required:
+      - tag
+      - count
+    properties:
+      tag:
+        type: string
+        description: tag string
+      count:
+        type: number
+        description: number of usages of the tag across FireCloud
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -269,11 +269,12 @@ trait AttributeComponent {
       }
     }
 
-    def findUniqueStringsByNameQuery(attrName: AttributeName) = {
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: String) = {
       filter(rec =>
         rec.namespace === attrName.namespace &&
         rec.name === attrName.name &&
-        rec.valueString.isDefined
+        rec.valueString.isDefined &&
+        rec.valueString.like('%' + queryString + '%')
       ).map(_.valueString).distinct
     }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -269,13 +269,17 @@ trait AttributeComponent {
       }
     }
 
-    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: String) = {
-      filter(rec =>
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String]) = {
+      // ToDo: figure out what the generated query is
+      val basicFilter = filter(rec =>
         rec.namespace === attrName.namespace &&
-        rec.name === attrName.name &&
-        rec.valueString.isDefined &&
-        rec.valueString.like(s"%$queryString%")
-      ).map(_.valueString).distinct
+          rec.name === attrName.name &&
+          rec.valueString.isDefined)
+
+      (queryString match {
+        case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
+        case None => basicFilter
+      }).map(_.valueString).distinct
     }
 
     def deleteAttributeRecords(attributeRecords: Seq[RECORD]): DBIOAction[Int, NoStream, Write] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -270,18 +270,19 @@ trait AttributeComponent {
     }
 
     def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String]) = {
-      // ToDo: figure out what the generated query is
 
       val basicFilter = filter(rec =>
         rec.namespace === attrName.namespace &&
           rec.name === attrName.name &&
           rec.valueString.isDefined)
 
-      (queryString match {
+      val res = (queryString match {
         case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
         case None => basicFilter
       }).groupBy(_.valueString).map(queryThing =>
-        (queryThing._1.get, queryThing._2.length))
+        (queryThing._1, queryThing._2.length))
+
+      res.sortBy(r => (r._2.desc, r._1))
     }
 
     def deleteAttributeRecords(attributeRecords: Seq[RECORD]): DBIOAction[Int, NoStream, Write] = {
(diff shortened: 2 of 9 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -853,4 +853,44 @@ class WorkspaceServiceSpec extends FlatSpec with ScalatestRouteTest with Matcher
     }
 
   }
+
```
skwalker, line 856: I wasn't sure where to do the tests; should I have some in workspaceApiSpec and the AttributeComponent for the query? Or is that redundant?

core/src/main/resources/swagger/rawls.yaml
```diff
@@ -3390,6 +3390,37 @@ paths:
...
+            description: Query
+            name: query
+            required: true
+            type: string
+        security:
+        - authorization:
+          - openid
+          - email
+          - profile
+          - https://www.googleapis.com/auth/devstorage.full_control
```
rushtong: Is this right? Do tag queries need dev storage access?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
rushtong: I think the only thing you need in `AttributeComponentSpec` is something that tests `findUniqueStringsByNameQuery`. It could use a check that makes sure you really are only getting unique strings from a collection that has duplicates as well as non-tag attributes.

rushtong (approved):
👍 pending one more test for `findUniqueStringsByNameQuery`

core/src/main/resources/swagger/rawls.yaml
```diff
@@ -3390,6 +3390,37 @@ paths:
               - email
               - profile
               - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/workspaces/tags/autocomplete/{query}':
```
davidangb: In terms of making good REST APIs, I'd prefer that:
* the endpoint is a noun, not a verb: "tags" not "autocomplete"
* filters are query params

_(I like this page as a reference: http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api)_

So I suggest you move the endpoint to `/api/workspaces/tags?q={userinput}`

Thinking ahead to future stories, when we need an endpoint that returns only those tags **associated with workspaces to which the user has access**, this would be easily achievable through another query param, e.g. `all=(true|false)`. We could also, in the future, make the `q` param optional, as a means of returning all tags.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -269,6 +269,15 @@ trait AttributeComponent {
       }
     }
 
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: String) = {
+      filter(rec =>
+        rec.namespace === attrName.namespace &&
+        rec.name === attrName.name &&
+        rec.valueString.isDefined &&
+        rec.valueString.like(s"%$queryString%")
```
davidangb: Could you show, using the Slick logger, what this query generates now that the `like` is present? I want to make sure we aren't opening ourselves up to sql injection.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -134,6 +134,12 @@ trait WorkspaceComponent {
       loadWorkspaces(workspaceQuery)
     }
 
+    def getTags(queryString: String): ReadAction[Seq[String]] = {
+      workspaceAttributeQuery.findUniqueStringsByNameQuery(AttributeName("tag", "tags"), queryString).result map {recs =>
```
davidangb: this should use `AttributeName.withTagsNS` instead of hardcoding.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -107,6 +107,11 @@ trait WorkspaceApiService extends HttpService with PerRequestCreator with UserIn
         requestContext => perRequest(requestContext, WorkspaceService.props(workspaceServiceConstructor, userInfo),
           WorkspaceService.GetBucketUsage(WorkspaceName(workspaceNamespace, workspaceName)))
       }
+    } ~
+    path("workspaces" / "tags" / "autocomplete" / Segment ) { (query) =>
+      get {
+        requestContext => perRequest(requestContext, WorkspaceService.props(workspaceServiceConstructor, userInfo), WorkspaceService.GetTags(query))
+      }
```
davidangb: see previous comment re: changing endpoint path

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
davidangb: Also please add test(s) that return more than one tag.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -853,4 +853,44 @@ class WorkspaceServiceSpec extends FlatSpec with ScalatestRouteTest with Matcher
...
+  it should "return the correct tags from autocomplete" in withTestDataServices { services =>
+
+    // when no tags, return empty set
+    val res1 = Await.result(services.workspaceService.getTags("notag"), Duration.Inf)
+      .asInstanceOf[RequestComplete[(StatusCode, Set[String])]]
+    assertResult(Set()) {
```
davidangb: syntax nit: `Set.empty` or `Set.empty[String]` instead of `Set()`

davidangb (changes requested):
comments inline!

The ticket GAWB-1405 talks about **frequency** of tags - that seems to be missing from this PR.

The ticket also talks about "similar spelling" - is PO satisfied with the functionality of the `like` clause for "similar spelling"?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
dvoet: I would not trust the slick logger, I would try to capture it on the mysql side.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -281,7 +281,7 @@ trait AttributeComponent {
         case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
         case None => basicFilter
       }).groupBy(_.valueString).map(queryThing =>
-        (queryThing._1.get, queryThing._2.length)) // not sure if we should be concerned about the .get erroring out?
+        (queryThing._1.get, queryThing._2.length))
```
skwalker: I think I should be able to do orderBy here but I wasn't able to get it to working so I gave up

skwalker:
@davidangb  
from the mysql logs:

```select `value_string`, count(1) from `WORKSPACE_ATTRIBUTE` where (((`namespace` = 'tag') and (`name` = 'tags')) and (`value_string` is not null)) and (`value_string` like '%can%') group by `value_string` ```

is that ok?

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala [hunk 9]
```diff
@@ -269,13 +269,19 @@ trait AttributeComponent {
...
         rec.namespace === attrName.namespace &&
-        rec.name === attrName.name &&
-        rec.valueString.isDefined &&
-        rec.valueString.like(s"%$queryString%")
-      ).map(_.valueString).distinct
+          rec.name === attrName.name &&
+          rec.valueString.isDefined)
+
+      (queryString match {
+        case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
```
rushtong: Do we care about case sensitivity or is that handled through the filter?

rushtong (approved):
One comment inline that is likely not valid since the tests handle mixed case queries and results.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
rushtong: Does this not work? `(queryThing._1.get, queryThing._2.length)).sortBy(_._2)` ... or _._1, etc.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
rushtong: Ignore ... tests cover the mixed case scenario.

skwalker: yeah the like means we are case-insensitive, which is what Moran wanted so that we don't have a `Cancer` tag and a `cancer` tag

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -269,6 +269,21 @@ trait AttributeComponent {
...
+      val basicFilter = filter(rec =>
+        rec.namespace === attrName.namespace &&
+          rec.name === attrName.name &&
+          rec.valueString.isDefined)
+
+      (queryString match {
+        case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
+        case None => basicFilter
+      }).groupBy(_.valueString).map(queryThing =>
+        (queryThing._1.get, queryThing._2.length))
```
davidangb: would adding `.sortBy(_._2.desc)` work here?

```
      }).groupBy(_.valueString).map(queryThing =>
        (queryThing._1.get, queryThing._2.length)).sortBy(_._2.desc)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -352,6 +354,13 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
     }
 
+  def getTags(query: Option[String]): Future[PerRequestMessage] =
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.workspaceQuery.getTags(query).map { result =>
+        RequestComplete(StatusCodes.OK, result.sortWith(_.count > _.count))
```
davidangb: I'd prefer to move the sorting down into the SQL (see previous comment), but if that's a problem I'm ok with it here.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
```diff
@@ -411,4 +411,30 @@ class AttributeComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+    runAndWait(insertWorkspaceAttributeRecords(workspace2ID, AttributeName.withTagsNS, AttributeString("cancer")))
+    runAndWait(insertWorkspaceAttributeRecords(workspace2ID, AttributeName.withTagsNS, AttributeString("buffalo")))
+
+    assertResult(Vector(("cancer", 2), ("cantaloupe", 1))) {
+      runAndWait(workspaceAttributeQuery.findUniqueStringsByNameQuery(AttributeName.withTagsNS, Some("can")).result)
+    }
+
+    assertResult(Vector(("buffalo", 1), ("cancer", 2), ("cantaloupe", 1))) {
+      runAndWait(workspaceAttributeQuery.findUniqueStringsByNameQuery(AttributeName.withTagsNS, None).result)
+    }
```
davidangb: in the case where multiple tags have the same count, sort order is deterministic, but ... defined by database row order? It's possible this test will fail in the future if some underlying technology such as mysql version changes. No action necessary now - just commenting for the future.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
davidangb: case sensitivity is defined by the db collation, which is `utf8_general_ci`, at least in dev. The "ci" in the collation signifies case-insensitive. So if we ever change our db implementation or db collation, this could be different ... but I'm sure we'd have a ton of other repercussions as well.

davidangb (commented):
minor comments inline, looking good! If you can resolve sorting in SQL, this should be 👍

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
skwalker: sorting by count and then by alphabetical order on tag, so this shouldn't be an issue anymore

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -269,6 +269,22 @@ trait AttributeComponent {
...
+          rec.name === attrName.name &&
+          rec.valueString.isDefined)
+
+      val res = (queryString match {
+        case Some(query) => basicFilter.filter(_.valueString.like(s"%${query}%"))
+        case None => basicFilter
+      }).groupBy(_.valueString).map(queryThing =>
+        (queryThing._1, queryThing._2.length))
+
+      res.sortBy(r => (r._2.desc, r._1))
```
davidangb: one last tweak? You can put the `.get` back in after the sortBy, e.g.:
```
res.sortBy(r => (r._2.desc, r._1)).map(x => (x._1.get, x._2))
```
and then you're returning a String, not an Option[String], and calling code is simplified. This means you'd have to revert the `Some()`s you added in the most recent commit.

davidangb (approved):
looks good, 👍 once tests pass!

PR #650, opened 2017-03-17
Feature groups 2.0

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-1740

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -649,6 +575,150 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+      }.recover {
+        // assume any sql exception is a FK violation, to be more specific catch MySQLIntegrityConstraintViolationException
+        // but it is good to keep the code free of mysql specific code
+        case sqle: SQLException => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, "Cannot delete group because it is in use", sqle))
+      }
+
+      _ <- Future.traverse(groupEmailsToDelete) { group =>
+        gcsDAO.deleteGoogleGroup(group).recover {
+          // log any exception but ignore
+          case t: Throwable => logger.error(s"error deleting google group $group", t)
```
dmohs, line 714: Doesn't group creation fail if the Google group already exists? If failing to delete the group isn't fatal, we should probably allow group creation to pass if the group exists.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
```diff
@@ -86,4 +86,20 @@ trait UserWiths {
...
   }
 
+  def withManagedGroupOwnerAccess[T](managedGroupRef: ManagedGroupRef, userRef: RawlsUserRef, dataAccess: DataAccess)(op: ManagedGroup => ReadWriteAction[T])(implicit executionContext: ExecutionContext): ReadWriteAction[T] = {
+    dataAccess.managedGroupQuery.load(managedGroupRef) flatMap {
+      case None => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound, s"group [${managedGroupRef.usersGroupName.value}] not found")))
+      case Some(managedGroup) =>
+        dataAccess.rawlsGroupQuery.isGroupMember(managedGroup.ownersGroup, userRef) flatMap {
+          case true => op(managedGroup)
+          case false =>
+            // figure out what error to show - if they are a user show 403 otherwise 404
```
dmohs, line 96: +1

dmohs:
This is how I read the architecture from this PR: We still have the concept of a realm, which a workspace can be associated with. A realm is implemented using a new abstraction, a "managed group," which is a Google group that can be managed through our API. I like this.

Chet has convinced us that having a new concept called an "Authorization Domain" (AD) makes sense. The UI will have a separate page for managing ADs and a workspace can be associated with zero or more ADs (zero or one is fine for now).

In this version of the API, it exposes managed groups. It would be much better for our users if the API used the exact same nomenclature as the UI. So, instead of:
```
/api/groups/{groupName}
```
I would prefer:
```
/api/authdomains/{name}
```
etc.

We'd also want workspace creation to use the term `authdomain` instead of `realm`. For code readability, a full replacement of "realm" with "authdomain" would be helpful.

Finally, the term `email` is going to be controversial, since Chet believes users are more likely to think of that as their contact email. My latest suggestion to him is `signInEmail`, but my personal favorite so far is `accountEmail`.

In summary, authorization domains are a thing and we would do well by both ourselves and our users to embrace the term in both our API and our code.

dmohs:
In light of our meeting, it seems that the `realm` -> `authDomain` rename is my only suggestion.

I also heard back from Chet. He suggested `registrationEmail` which is fine by me.

dvoet:
re registrationEmail: I am going to leave it just plain email because it can also be a group email. I also find it hard to believe that users will be confused because the email we care about is shown when they login and at the top right of our app.

re realm rename: that should be another PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
dvoet: We can't do that because an existing group may have members and then it will be out of sync. This particular bit of code is a best effort cleanup. It is difficult because the database entries are already gone at this point so there is not much else to do but move on. I could include the google deletes in the transaction, but that is bad practice, plus what if 1 group is deleted and another fails... also a bad state. We can't delete the google groups first because it might be in use.

This seems like the least evil/complicated code.

dmohs: I think the non-fatal deletion is a good choice. I'm more concerned about creation because it smells like the cause of an annoying scenario. From memory, I think it looks something like this:

I set up Rawls to use my own database, then try to create a user. Since everything is hooked to dev.test.firecloud.org, this fails because the user already exists. Not sure how to recover. Thoughts?

dmohs (approved):
One more question about group creation/deletion but overall I'm good with with as-is now.

PR #660, opened 2017-03-24
ah-gawb-1709: add update library attribute endpoint which will check permissions and update attributes 

ahaessly:
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


rushtong (approved):
👍

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -335,22 +337,32 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
-  def updateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]): Future[PerRequestMessage] =
+  def adminUpdateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]): Future[PerRequestMessage] = asFCAdmin {
```
davidangb: the `asFCAdmin` should be on *either* the pipeTo at line 191 or here, but not both. This class is inconsistent in its pattern - you'll see it done both ways (but not both ways at the same time). We should prefer it at the pipeTo level.

core/src/bin/README.md
```diff
@@ -1,5 +1,9 @@
...
 
 Jenkins runs `publishSnapshot.sh` on every dev build, but that makes "unofficial" `-SNAP` versions.
 
-Note that you need `ARTIFACTORY_USERNAME` and `ARTIFACTORY_PASSWORD` in your env for either of these to work.
\ No newline at end of file
+Note that you need `ARTIFACTORY_USERNAME` and `ARTIFACTORY_PASSWORD` in your env for either of these to work (you can get this info from jenkins config for rawls).
+
+cmd$ ARTIFACTORY_USERNAME=<name> ARTIFACTORY_PASSWORD=<pw> core/src/bin/publishRelease.sh
+
+You can view what is in the artifactory here: http://artifactory.broadinstitute.org/artifactory/libs-release-local
```
davidangb: I don't feel strongly about this, so take it or leave it. That said:
* I'd prefer that doc about orchestration lives in orch's readme, not rawls
* I'd prefer that internal info, such as calling out Jenkins configs, lives in a wiki or elsewhere, instead of in this public file.

davidangb (commented):
two comments inline (one optional). This should really have a test, too - can you crib off the existing update-workspace tests in WorkspaceApiServiceSpec and ensure an admin can update a workspace that s/he otherwise wouldn't have access to?

dvoet:
Why do we need this? Going down this path makes me feel like we have the wrong security model.

Put another way, currently rawls is in control of its authorization model and the programatic use of this api changes that. We are going from an easy model to reason about to one that is not which is a bad thing for security.

ahaessly:
For data delivery, PMs only have read access to the workspace. But they will need to catalog and publish in library. The new permission catalog combined with read access should allow these users to modify library attributes. Therefore, we need a way to bypass the standard acls for updating attributes on the workspace. One alternative would be to have rawls know about the library permissions model and to do the check (which i don't think anyone wants). Does it help knowing that once library moves to being its own service we won't need this admin call?

dvoet:
no, it does not make me feel any better. It feels like we are repeatedly incurring significant tech debt.

I think I would be more comfortable with building into rawls that if you have both read+ access and the new permission in a workspace you are able to edit attributes in the library namespace on that workspace. That way all authorization enforcement is kept in the same place. Rawls already knows about library namespace attributes and the new role so we are not adding more knowledge.

I think if you have to use the word bypass with respect to security you should do it a different way.

ahaessly:
I think it's tech debt either way until we get library managing its own data. It actually seemed more desirable to keep the library specific logic in orchestration until it is moved. Towards this goal, I was able to remove rawls' knowledge of library attributes. Here's a link to the tech doc that talks about the permissions we are now enforcing - and we are not foregoing permissions checking, just implementing a model tailored to library. 
https://docs.google.com/a/broadinstitute.org/document/d/1P7nPgblmsNixFH18DC7_B4OfoW3O4aUcXw70p0V7c44/edit?usp=sharing

davidangb:
I agree with the separation of concerns here - I'd rather have Library understand/manage its own permissions, instead of needing to coordinate across multiple artifacts to effect a logic change. I also agree it's really messy right now - once Library metadata is not backed by rawls attributes, this will make a lot more sense.

core/src/test/scala/org/broadinstitute/dsde/rawls/openam/MockUserInfoDirectives.scala
```diff
@@ -15,3 +15,10 @@ trait MockUserInfoDirectives extends UserInfoDirectives {
     provide(userInfo)
   }
 }
+
+trait MockAdminUserInfoDirectives extends UserInfoDirectives {
+  def requireUserInfo(magnet: ImplicitMagnet[ExecutionContext]): Directive1[UserInfo] = {
+    // just return the cookie text as the common name
+    provide(UserInfo("admin", OAuth2BearerToken("token"), 123, "123456789876543212345"))
+  }
+}
```
davidangb: nit: `MockAdminUserInfoDirectives`code is differently structured than `MockUserInfoDirectives` (including the comment); would be nice to have them exactly parallel. I don't feel strongly about this.

davidangb (commented):
One very minor code comment; otherwise :+1: for code/syntax. Still need to resolve the higher-level architectural questions.

core/src/test/scala/org/broadinstitute/dsde/rawls/openam/MockUserInfoDirectives.scala
ahaessly: no problem - it actually had changed since i made my code changes

core/src/bin/README.md
```diff
@@ -1,5 +1,9 @@
-Running the `publishRelease.sh` script publishes a release of rawls-model, workbench-util and workbench-google to Artifactory. You should do this manually when you change something in `model/src`, `util/src` or `google/src` so clients such as Orchestration can bump their version and stay in sync.
+Running the `publishRelease.sh` script publishes a release of rawls-model, workbench-util and workbench-google to Artifactory. You should do this manually from the base directory of the repo when you change something in `model/src`, `util/src` or `google/src` so clients such as Orchestration can bump their version and stay in sync. In Orchestration, the file that contains the reference is project/Dependencies.scala.
```
davidangb: I still don't think the rawls readme should talk about orchestration code, but also won't hold the PR for this issue.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+      else if (names.head == discoverableWSAttribute)
+        withDiscoverabilityModifyPermissions(canShare, hasCatalog, userLevel) _
+      else
+        withModifyPermissions(hasCatalog, userLevel) _
+    } else {
+      assert(!names.contains(publishedFlag))
+      if (names.contains(discoverableWSAttribute))
+        withModifyAndDiscoverabilityModifyPermissions(canShare, hasCatalog, userLevel) _
+      else
+        withModifyPermissions(hasCatalog, WorkspaceAccessLevels.Write) _
```
davidangb: shouldn't this be:
```
withModifyPermissions(hasCatalog, userLevel) _
```

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+    }
```
davidangb: caveat: I haven't actually tried to run this to see if it works ... but a more Scala-idiomatic way to write this would be:
```
    names match {
      case `publishedFlag` :: Nil => withChangePublishedPermissions(isCurator, hasCatalog, userLevel) _
      case `discoverableWSAttribute` :: Nil => withDiscoverabilityModifyPermissions(canShare, hasCatalog, userLevel) _
      case x if x.contains(publishedFlag) => // throw; this is the assert condition in your original code
      case x if x.contains(discoverableWSAttribute) => withModifyAndDiscoverabilityModifyPermissions(canShare, hasCatalog, userLevel) _
      case _ => withModifyPermissions(hasCatalog, userLevel) _
    }
```
(I'm a little worried about the Seq construction using `::`. You might need `:+`,  or might need to do ```case Seq(`publishedFlag`)```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -338,25 +340,39 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      } match {
+        case Success(result) => result
+        case Failure(e: AttributeUpdateOperationException) =>
+          DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, s"Unable to update ${workspaceName}", ErrorReport(e))))
+        case Failure(regrets) => DBIO.failed(regrets)
+      }
+      updateAction.map { savedWorkspace =>
+        RequestComplete(StatusCodes.OK, savedWorkspace)
+      }
+    }
```
davidangb: Hmm ... in the `updateLibraryAttributes`/`updateWorkspace` refactor - which I like from a modularity perspective - the db transaction boundaries changed. Now, it's theoretically possible that the workspace changed between the time we get it and the time we update it. Consider moving the transaction boundary back out of the private method and into the two calling methods?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -338,25 +340,39 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
     }
+  }
+
+  def updateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]): Future[PerRequestMessage] = {
+    withAttributeNamespaceCheck(operations.map(_.name)) {
+    getWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Write) flatMap { ctx =>
+        updateWorkspace(workspaceName, operations, ctx)
+      }
+    }
+  }
```
davidangb: nothing prevents a user from calling this with library attributes and bypassing the library security checks! Maybe update `attributeNamespaceCheck` to reject library namespace?

core/src/test/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+
+    "LibraryPermissionsSupport" - {
+      "with catalog and read permissions" - {
+        "should be allowed to modify workspace" in {
+          val mod = getPermissionFunction(modifyNames, isCurator=false, canShare=false, hasCatalogOnly=true, userLevel=WorkspaceAccessLevels.Read)
+          assertResult("org.broadinstitute.dsde.rawls.util.LibraryPermissionsSupport$$anonfun$getPermissionFunction$3")(mod.getClass.getName)
+          assertResult(RequestComplete(StatusCodes.OK))(Await.result(mod({Future.successful(RequestComplete(StatusCodes.OK))}), Duration.Inf))
+        }
+        "should be allowed to modify discoverability" in {
+          val mod = getPermissionFunction(discoverabilityAndModNames, isCurator=false, canShare=false, hasCatalogOnly=true, userLevel=WorkspaceAccessLevels.Read)
```
davidangb: optional feedback: instead of defining a new `val discoverabilityAndModNames = Seq(AttributeName("library","test1"), discoverableWSAttribute)` and using it here, you could instead do:
```
val mod = getPermissionFunction(discoverabilityOnly ++ modifyNames, isCurator=false
```
which is a bit more self-documenting. I'm fine either way.

core/src/test/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+          assertResult("org.broadinstitute.dsde.rawls.util.LibraryPermissionsSupport$$anonfun$getPermissionFunction$2")(mod.getClass.getName)
+          assertResult(RequestComplete(StatusCodes.OK))(Await.result(mod({Future.successful(RequestComplete(StatusCodes.OK))}), Duration.Inf))
+        }
+        "should not be allowed to modify publish" in {
+          val mod = getPermissionFunction(publishNames, isCurator=false, canShare=true, hasCatalogOnly=false, userLevel=WorkspaceAccessLevels.Read)
+          assertResult("org.broadinstitute.dsde.rawls.util.LibraryPermissionsSupport$$anonfun$getPermissionFunction$1")(mod.getClass.getName)
+          intercept[RawlsExceptionWithErrorReport](Await.result(mod({Future.successful(RequestComplete(StatusCodes.OK))}), Duration.Inf))
+        }
+      }
+    }
```
davidangb: I see why you were wondering about reflection. Checking the `$$anonfun$getPermissionFunction$...` name here does feel fragile and difficult to read.

Best I can think of is to create a true class hierarchy, where you have a base\trait `LibraryPermChecker` and multiple subclasses, e.g. `ChangePublishedChecker`, `ChangeDiscoverabilityChecker`, `ChangeMetadataChecker`, etc. - each of which would implement a `withPermissions` method. Have `getPermissionFunction` return an instance of one of the subclasses, and have `withLibraryPermissions` use that instance. So you'd have something like:
```
          val checkerToInvoke = getPermissionFunction(names, isCurator, canShare, hasCatalogOnly, userLevel)
          checkerToInvoke.withPermissions(op)
```
then, in the tests, you can have the asserts look at the actual class name (ChangePublishedChecker vs ChangeDiscoverabilityChecker), instead of looking at the anonymous function.

davidangb (commented):
back to you!

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+          userLevel <- gm(RawlsUser(userInfo), ctx, dataAccess)
+        } yield {
+          val functionToInvoke = getPermissionFunction(names, isCurator, canShare, hasCatalogOnly, userLevel)
+            functionToInvoke(op)
+        }
+      } flatMap(identity)
+    }
+  }
+
+  protected def getPermissionFunction(names: Seq[AttributeName], isCurator: Boolean, canShare: Boolean, hasCatalogOnly: Boolean, userLevel:WorkspaceAccessLevel) = {
```
rushtong: By adding an explicit return, (always my preference in scala code)  
```(Future[PerRequestMessage]) => Future[PerRequestMessage]```
you can then avoid the need to include the `_`s.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala (hunk 5 above)
Comment (rushtong) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala]: +1

File: core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala (hunk 6 above)
Comment (rushtong) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala]: +1. Relying on `.head` seems brittle.

**Review by rushtong — APPROVED** [review]
👍  Pending addressing feedback in other reviews.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala [hunk 12]
```diff
@@ -0,0 +1,109 @@
...
+import org.broadinstitute.dsde.rawls.webservice.PerRequest.PerRequestMessage
+import spray.http.StatusCodes
+
+import scala.concurrent.Future
+
+/**
+  * Created by ahaessly on 3/31/17.
+  */
+trait LibraryPermissionsSupport extends RoleSupport {
+  final val publishedFlag = AttributeName("library","published")
```
dvoet: use AttributeName.withLibraryNS

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+  final val discoverableWSAttribute = AttributeName("library","discoverableByGroups")
+
+  def withLibraryPermissions(workspace: Workspace,
+                             operations: Seq[AttributeUpdateOperation],
+                             dataSource: SlickDataSource,
+                             userInfo: UserInfo,
+                             gm: (RawlsUser, SlickWorkspaceContext, DataAccess) => ReadAction[WorkspaceAccessLevel])
```
dvoet: what is gm?

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+    val hasCatalog = hasCatalogOnly && userLevel >= WorkspaceAccessLevels.Read
+    if (names.size == 1) {
+      if (names.head == publishedFlag)
+        withChangePublishedPermissions(isCurator, hasCatalog, userLevel) _
+      else if (names.head == discoverableWSAttribute)
+        withDiscoverabilityModifyPermissions(canShare, hasCatalog, userLevel) _
+      else
+        withModifyPermissions(hasCatalog, userLevel) _
+    } else {
+      assert(!names.contains(publishedFlag))
```
dvoet: this should not be an assert as it will probably lead to a 500 error. Asserts should not be used to indicate bad parameters anyway. Instead throw a RawlsExceptionWithErrorReport

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
dvoet: +1 for using match but the :: Nil stuff may only work for List. Instead you can use Seq(`publishedFlag`).

dvoet (commented):
Stopping for now

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,110 @@
...
+      hasCatalogOnly <- dataAccess.workspaceQuery.getUserCatalogPermissions(RawlsUserSubjectId(userInfo.userSubjectId), ctx)
+      userLevel <- gm(RawlsUser(userInfo), ctx, dataAccess)
+      result <- getPermissionChecker(names, isCurator, canShare, hasCatalogOnly, userLevel).withPermissions(op)
+    } yield result
+  }
+
+  def getPermissionChecker(names: Seq[AttributeName], isCurator: Boolean, canShare: Boolean, hasCatalogOnly: Boolean, userLevel: WorkspaceAccessLevel) = {
+    val hasCatalog = hasCatalogOnly && userLevel >= WorkspaceAccessLevels.Read
+    // need to conver to seq to reduce del and add ops when changing discoverable attribute, and back to Seq to match in case
+    names.toSet.toSeq match {
```
davidangb: use `names.distinct`

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,110 @@
...
+      case Seq(`publishedFlag`) => ChangePublishedChecker(isCurator, hasCatalog, userLevel)
+      case Seq(`discoverableWSAttribute`) => ChangeDiscoverabilityChecker(canShare, hasCatalog, userLevel)
+      case x if x.contains(publishedFlag) => throw new RawlsException("Unsupported parameter - can't modify published with other attribtes")
```
davidangb: nit: typo on "attributes"

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,110 @@
...
+  def withPermissions(op: => ReadWriteAction[Workspace]): ReadWriteAction[Workspace] = {
+    if (ChangeMetadataChecker(hasCatalog: Boolean, maxLevel: WorkspaceAccessLevels.WorkspaceAccessLevel).canModify &&
+      ChangeDiscoverabilityChecker(canShare: Boolean, hasCatalog: Boolean, maxLevel: WorkspaceAccessLevels.WorkspaceAccessLevel).canModify)
+      op
+    else
+      throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, s"You must be an owner or have catalog with read permissions."))
+  }
+  protected def canModify(canShare: Boolean, hasCatalog: Boolean, maxLevel: WorkspaceAccessLevels.WorkspaceAccessLevel): Boolean = {
+    maxLevel >= WorkspaceAccessLevels.Owner || canShare || hasCatalog
+  }
```
davidangb: this `canModify` is never used. Remove?

davidangb (approved):
Some very minor comments inline, otherwise 👍 !!!!!

For follow-on work/stories, I'd love to see:

* unit tests against `withLibraryAttributeNamespaceCheck` and `withAttributeNamespaceCheck`
* a refactor that introduces a class along the lines of:
```
case class WorkspaceUserPermissions(
  userInfo: UserInfo,
  workspace: Workspace,
  isCurator: Boolean,
  canShare: Boolean,
  hasCatalogOnly: Boolean,
  maxLevel: WorkspaceAccessLevels.WorkspaceAccessLevel) {
  
  def hasCatalog = hasCatalogOnly && maxLevel >= WorkspaceAccessLevels.Read
  
}
```
which you could then use to pass around as arguments to many of the methods in this PR, instead of having lots of methods that accept 3 or 4 or more of those arguments individually.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,106 @@
...
+trait LibraryPermissionsSupport extends RoleSupport {
+  final val publishedFlag = AttributeName.withLibraryNS("published")
+  final val discoverableWSAttribute = AttributeName.withLibraryNS("discoverableByGroups")
+
+  def withLibraryPermissions(ctx: SlickWorkspaceContext,
+                             operations: Seq[AttributeUpdateOperation],
+                             dataAccess: DataAccess,
+                             userInfo: UserInfo,
+                             isCurator: Boolean,
+                             getMaxAccess: (RawlsUser, SlickWorkspaceContext, DataAccess) => ReadAction[WorkspaceAccessLevel])
```
dvoet: why is this a function?

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,106 @@
...
+    } yield result
+  }
+
+  def getPermissionChecker(names: Seq[AttributeName], isCurator: Boolean, canShare: Boolean, hasCatalogOnly: Boolean, userLevel: WorkspaceAccessLevel) = {
+    val hasCatalog = hasCatalogOnly && userLevel >= WorkspaceAccessLevels.Read
+    // need to multiple delete and add ops when changing discoverable attribute
+    names.distinct match {
+      case Seq(`publishedFlag`) => ChangePublishedChecker(isCurator, hasCatalog, userLevel)
+      case Seq(`discoverableWSAttribute`) => ChangeDiscoverabilityChecker(canShare, hasCatalog, userLevel)
+      case x if x.contains(publishedFlag) => throw new RawlsException("Unsupported parameter - can't modify published with other attributes")
```
dvoet: this will probably bubble up as a 500 error but probably should be a 400 error, use RawlsExceptionWithErrorReport to specify the status code

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,106 @@
...
+    val names = operations.map(attribute => attribute.name)
+    for {
+      canShare <- dataAccess.workspaceQuery.getUserSharePermissions(RawlsUserSubjectId(userInfo.userSubjectId), ctx)
+      hasCatalogOnly <- dataAccess.workspaceQuery.getUserCatalogPermissions(RawlsUserSubjectId(userInfo.userSubjectId), ctx)
+      userLevel <- getMaxAccess(RawlsUser(userInfo), ctx, dataAccess)
+      result <- getPermissionChecker(names, isCurator, canShare, hasCatalogOnly, userLevel).withPermissions(op)
+    } yield result
+  }
+
+  def getPermissionChecker(names: Seq[AttributeName], isCurator: Boolean, canShare: Boolean, hasCatalogOnly: Boolean, userLevel: WorkspaceAccessLevel) = {
```
dvoet: specify the return type please

davidangb:
👍

core/src/test/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSpec.scala
```diff
@@ -0,0 +1,194 @@
...
+    createdBy="createdBy",
+    createdDate=DateTime.now(),
+    lastModified=DateTime.now(),
+    attributes=Map.empty,
+    bucketName="bucketName",
+    accessLevels=Map.empty,
+    realmACLs=Map())
+
+
+  "LibraryPermissionsSupport" - {
```
dvoet: I don't think these tests are sufficient because they all rely on direct access of getPermissionChecker but none test if getPermissionChecker is called correctly. I can't really tell from the rest of the tests whether or not all cases are tested. I would rather see either test data constructed that exercises all cases or mock out the calls to dataAccess.workspaceQuery inside the for comprehension of withLibraryPermissions.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,106 @@
...
+    val invalidNamespaces = namespaces -- Set(AttributeName.libraryNamespace)
+    if (invalidNamespaces.isEmpty) op
+    else {
+      val err = ErrorReport(statusCode = StatusCodes.Forbidden, message = s"All attributes must be in the library namespace")
+      throw new RawlsExceptionWithErrorReport(errorReport = err)
+    }
+  }
+}
+
+trait LibraryPermissionChecker {
```
dvoet: I think this would be better expressed [with functions](https://gist.github.com/dvoet/c750a1d7ebe70d8f706752afb59effb4) instead of command objects.

dvoet (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSpec.scala
dvoet: thinking a bit more, I really want the test data approach. That way we have confidence that the queries are right too.

dvoet:
@davidangb I only see 1 of my comments addressed

dvoet (approved):

dvoet:
👍 thanks

PR #663, opened 2017-03-29
GAWB-1836 Poll for workflows smarter

helgridly:
**@Reviewers:** If you could also thumb the firecloud-develop config prs linked below that'd be much appreciated.

A few changes in here:

1. A new `processInterval` config value that replaces manually switching between `LookForWorkflows` and `ScheduleNextWorkflowQuery` with a knob we can turn.
2. Added `TransactionIsolation.ReadCommitted` to the db query for list workspaces. When a lot of workflows are being submitted in a workspace, the workflow last changed date is continually being written to. The default isolation of `RepeatableRead` meant this query was waiting for table locks and thus took a long time if you had access to a workspace that was submitting lots of workflows.
3. JProfiler to version 9.2 and a fix to the jprofiler dockerfile so it runs the JProfiler-enabled Rawls when it's deployed.

`parallelSubmitters = 50, processInterval = 0` pegged the Rawls database at 100% and made the rest of the app slow, according to 🐶 . Lowering `parallelSubmitters = 10` and throwing a ton of workflows at it makes the db hit 8/16CPUs, which is plenty headroom for other things. With `parallelSubmitters = 10, processInterval = 10ms`, we get a theoretical max throughput of 1000wf/s compared to our previous max of 50wf/s. This assumes the Cromwell batch size is still 1; I haven't fiddled with that at all.

This PR and its pals all need to travel together:
https://github.com/broadinstitute/rawls/pull/663
https://github.com/broadinstitute/firecloud-develop/pull/374
https://github.com/broadinstitute/firecloud-develop/pull/375
https://github.com/broadinstitute/firecloud-develop/pull/376

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -152,6 +152,7 @@ object Boot extends App with LazyLogging {
         shardedExecutionServiceCluster,
         conf.getInt("executionservice.batchSize"),
         gcsDAO.getBucketServiceAccountCredential,
+        util.toScalaDuration(conf.getDuration("executionservice.processInterval")),
         util.toScalaDuration(conf.getDuration("executionservice.pollInterval")),
         maxActiveWorkflowsTotal,
         maxActiveWorkflowsPerUser,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
@@ -27,15 +27,17 @@ object WorkflowSubmissionActor {
             executionServiceCluster: ExecutionServiceCluster,
             batchSize: Int,
             credential: Credential,
+            processInterval: FiniteDuration,
             pollInterval: FiniteDuration,
             maxActiveWorkflowsTotal: Int,
             maxActiveWorkflowsPerUser: Int,
             runtimeOptions: Option[JsValue]): Props = {
-    Props(new WorkflowSubmissionActor(dataSource, methodRepoDAO, googleServicesDAO, executionServiceCluster, batchSize, credential, pollInterval, maxActiveWorkflowsTotal, maxActiveWorkflowsPerUser, runtimeOptions))
+    Props(new WorkflowSubmissionActor(dataSource, methodRepoDAO, googleServicesDAO, executionServiceCluster, batchSize, credential, processInterval, pollInterval, maxActiveWorkflowsTotal, maxActiveWorkflowsPerUser, runtimeOptions))
   }
 
   sealed trait WorkflowSubmissionMessage
-  case object ScheduleNextWorkflowQuery extends WorkflowSubmissionMessage
+  case object ScheduleNextWorkflow extends WorkflowSubmissionMessage
+  case object ProcessNextWorkflow extends WorkflowSubmissionMessage
   case object LookForWorkflows extends WorkflowSubmissionMessage
   case class SubmitWorkflowBatch(workflowIds: Seq[Long]) extends WorkflowSubmissionMessage
 
@@ -47,6 +49,7 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
                               val executionServiceCluster: ExecutionServiceCluster,
                               val batchSize: Int,
                               val credential: Credential,
+                              val processInterval: FiniteDuration,
                               val pollInterval: FiniteDuration,
                               val maxActiveWorkflowsTotal: Int,
                               val maxActiveWorkflowsPerUser: Int,
@@ -54,11 +57,14 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
 
   import context._
 
-  self ! ScheduleNextWorkflowQuery
+  self ! ScheduleNextWorkflow
 
   override def receive = {
-    case ScheduleNextWorkflowQuery =>
-      scheduleNextWorkflowQuery
+    case ScheduleNextWorkflow =>
+      scheduleNextWorkflowQuery(pollInterval + (scala.util.Random.nextInt(1000) milliseconds))
+
+    case ProcessNextWorkflow =>
+      scheduleNextWorkflowQuery(processInterval)
 
     case LookForWorkflows =>
       getUnlaunchedWorkflowBatch() pipeTo self
@@ -68,11 +74,11 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
 
     case Status.Failure(t) =>
       logger.error(t.getMessage)
-      scheduleNextWorkflowQuery
+      self ! ScheduleNextWorkflow
   }
 
-  def scheduleNextWorkflowQuery: Cancellable = {
-    system.scheduler.scheduleOnce(pollInterval + (scala.util.Random.nextInt(1000) milliseconds), self, LookForWorkflows)
+  def scheduleNextWorkflowQuery(delay: FiniteDuration): Cancellable = {
+    system.scheduler.scheduleOnce(delay, self, LookForWorkflows)
   }
 }
 
@@ -83,7 +89,6 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
   val executionServiceCluster: ExecutionServiceCluster
   val batchSize: Int
   val credential: Credential
-  val pollInterval: FiniteDuration
   val maxActiveWorkflowsTotal: Int
   val maxActiveWorkflowsPerUser: Int
   val runtimeOptions: Option[JsValue]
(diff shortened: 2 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -152,6 +152,7 @@ object Boot extends App with LazyLogging {
         shardedExecutionServiceCluster,
         conf.getInt("executionservice.batchSize"),
         gcsDAO.getBucketServiceAccountCredential,
+        util.toScalaDuration(conf.getDuration("executionservice.processInterval")),
```
helgridly, line 155: process interval: interval between processing batches of workflows
poll interval: interval between polling when you've not found any and need to check back later

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -68,11 +74,11 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
 
     case Status.Failure(t) =>
       logger.error(t.getMessage)
-      scheduleNextWorkflowQuery
+      scheduleNextWorkflowQuery(pollInterval + (scala.util.Random.nextInt(1000) milliseconds))
```
helgridly: this makes me sad. i think `Future.succesful(ScheduleNextWorkflow)` might work too

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -392,7 +393,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
 
       results.map { responses => RequestComplete(StatusCodes.OK, responses) }
-    }
+    }, TransactionIsolation.ReadCommitted)
```
helgridly, line 396: be more okay with reading possibly-stale data from the workflows table when other people are writing to it. not super important that we have the TOTALLY LATEST last changed date, especially if workflows are being launched constantly

docker/jprofiler/Dockerfile
```diff
@@ -1,5 +1,7 @@
-FROM broadinstitute/rawls:dev-base
+FROM broadinstitute/rawls:dev
 
-RUN wget http://download-aws.ej-technologies.com/jprofiler/jprofiler_linux_9_1.tar.gz && tar -xvf jprofiler_linux_9_1.tar.gz && rm jprofiler_linux_9_1.tar.gz
+RUN wget http://download-aws.ej-technologies.com/jprofiler/jprofiler_linux_9_2.tar.gz && tar -xvf jprofiler_linux_9_2.tar.gz && rm jprofiler_linux_9_2.tar.gz
```
helgridly, line 3: upgrade to the version of jprofiler that i have installed on my computer

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -47,18 +49,22 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
...
   import context._
 
-  self ! ScheduleNextWorkflowQuery
+  self ! ScheduleNextWorkflow
 
   override def receive = {
-    case ScheduleNextWorkflowQuery =>
-      scheduleNextWorkflowQuery
+    case ScheduleNextWorkflow =>
+      scheduleNextWorkflowQuery(pollInterval + (scala.util.Random.nextInt(1000) milliseconds))
```
jmthibault79, line 64: I prefer jitter as a function of the timespan e.g. `rand(90% to 110%) * timespan` so we don't need to keep track of the appropriate scale for jitter separately.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -27,15 +27,17 @@ object WorkflowSubmissionActor {
...
             pollInterval: FiniteDuration,
             maxActiveWorkflowsTotal: Int,
             maxActiveWorkflowsPerUser: Int,
             runtimeOptions: Option[JsValue]): Props = {
-    Props(new WorkflowSubmissionActor(dataSource, methodRepoDAO, googleServicesDAO, executionServiceCluster, batchSize, credential, pollInterval, maxActiveWorkflowsTotal, maxActiveWorkflowsPerUser, runtimeOptions))
+    Props(new WorkflowSubmissionActor(dataSource, methodRepoDAO, googleServicesDAO, executionServiceCluster, batchSize, credential, processInterval, pollInterval, maxActiveWorkflowsTotal, maxActiveWorkflowsPerUser, runtimeOptions))
   }
 
   sealed trait WorkflowSubmissionMessage
-  case object ScheduleNextWorkflowQuery extends WorkflowSubmissionMessage
```
jmthibault79, line 38: Since it's scheduling the next `LookForWorkflows` and may not necessarily process any workflows if it doesn't find any, some descriptor of that in the name is appropriate.  `-Query` may not be the best suffix but it's better than nothing.  Maybe `-Search` ?

`ProcessNextWorkflow` works because we know there are workflows to process.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
jmthibault79: `self ! ScheduleNextWorkflow` ?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: hmm. might be worth throwing jitter into utils, since the concept is used across rawls. i didn't write this code originally, just moved it.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: my goal here was to make ScheduleNextWorkflow and ProccessNextWorkflow look similar

docker/jprofiler/Dockerfile
jmthibault79: Please keep this commit separate when you squash

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
jmthibault79: Yes please :)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
jmthibault79: I like this and I also want this to remain a separate commit

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: i'll do this as a separate pr, then

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: oh right. yes! thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
dvoet: why do you need 2 different messages, just make ScheduleNextWorkflow a case class with the duration to wait the parameter.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
dvoet: moving it is a good idea, but I disagree that it should be a function of the timespan. For long intervals it would be too long and for short, too short.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: imo these are two different commands and how long to wait is down to the receiver of the message. also the bike shed should be painted yellow

dvoet: if the only difference is how long to wait, the code is more clear if you use the same message.

PR #664, opened 2017-03-30
GAWB-1569: Return conflict subtrees, allow users to link all existing entities in those subtrees to new imports

MatthewBemis:
Got stuck down a horrible path of trying to prematurely optimize the code, which made it an absolute horror to read through and understand. I completely re-wrote everything yesterday with a different mindset and this is what I came up with.

-When there are conflicts with the entities that you directly specify, these are called "hard conflicts".
-When there are conflicts within the subtrees of the entities that you specify, these are called "soft conflicts". A tree that shows the sources of the conflicts will be returned. These entities aren't linked unless you call the endpoint with the `linkExistingEntities` parameter set to `true`. 

Both cases return a 409.

--------------------------------------------------------------------------------------------------------

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ test-output/
 *.log
 config/
 rawlsdb/
+*.pid
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ test-output/
 *.log
 config/
 rawlsdb/
-*.pid
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -2072,9 +2072,7 @@ paths:
         '201':
           description: Successful Request
           schema:
-            type: array
-            items:
-              $ref: '#/definitions/Entity'
+            $ref: '#/definitions/EntityCopyResponse'
         '404':
           description: Source workspace not found
           schema:
@@ -2082,7 +2080,7 @@ paths:
         '409':
           description: There are conflicts in your copy operation. Retry with the paramater linkExistingEntities=true to link all soft conflicts to the entities that already exist in your workspace.
           schema:
-            $ref: '#/definitions/ErrorReport'
+            $ref: '#/definitions/EntityCopyResponse'
         '422':
           description: Destination workspace must match source's realm, if the source has a realm
           schema:
@@ -3931,6 +3929,61 @@ definitions:
         description: stack trace
         items:
           $ref: '#/definitions/StackTraceElement'
+
+  EntityCopyResponse:
+    description: ''
+    required:
+      - entitiesCopied
+      - softConflicts
+      - hardConflicts
+    properties:
+      entitiesCopied:
+        type: array
+        description: the entities that were successfully copied into the workspace
+        items:
+          $ref: '#/definitions/AttributeEntityReference'
+      hardConflicts:
+        type: array
+        description: ''
+        items:
+          $ref: '#/definitions/EntityHardConflict'
+      softConflicts:
+        type: array
+        description: ''
+        items:
+          $ref: '#/definitions/EntitySoftConflict'
+
+  EntityHardConflict:
+    description: ''
+    required:
+      - entityType
+      - entityName
+    properties:
+      entityType:
+        type: string
+        description: 'the type of the entity'
+      entityName:
+        type: string
+        description: 'the name of the entity'
+
+  EntitySoftConflict:
+    description: ''
+    required:
+      - entityType
+      - entityName
+      - conflicts
+    properties:
+      entityType:
+        type: string
+        description: 'the type of the entity'
+      entityName:
+        type: string
+        description: 'the name of the entity'
+      conflicts:
+        type: array
+        description: 'the next entity in the path to the conflict'
+        items:
+          $ref: '#/definitions/EntitySoftConflict'
 
   StackTraceElement:
     description: ''
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -2080,7 +2080,7 @@ paths:
           schema:
             $ref: '#/definitions/ErrorReport'
         '409':
-          description: One or more entities with the destination name/type already exist
+          description: There are conflicts in your copy operation. Retry with the paramater linkExistingEntities=true to link all soft conflicts to the entities that already exist in your workspace.
           schema:
             $ref: '#/definitions/ErrorReport'
         '422':
@@ -2097,6 +2097,12 @@ paths:
       summary: copy entities from one workspace into another workspace
       operationId: copy_entities
       parameters:
+        - in: query
+          description: true to link new entities to existing entities, false to fail instead of link
+          name: linkExistingEntities
+          required: false
+          default: false
+          type: boolean
         - in: body
           description: Source and destination for entities
           name: entityCopyDefinition
--- a/core/src/main/resources/swagger/rawls.yaml
+++ b/core/src/main/resources/swagger/rawls.yaml
@@ -3939,17 +3939,17 @@ definitions:
     properties:
       entitiesCopied:
         type: array
-        description: the entities that were successfully copied into the workspace
+        description: 'the entities that were successfully copied into the workspace'
         items:
           $ref: '#/definitions/AttributeEntityReference'
       hardConflicts:
         type: array
-        description: ''
+        description: 'these entities already exist in your workspace'
         items:
           $ref: '#/definitions/EntityHardConflict'
       softConflicts:
         type: array
-        description: ''
+        description: 'the conflicts within your entity subtrees. note that this will be empty if there are hard conflicts'
         items:
           $ref: '#/definitions/EntitySoftConflict'
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -523,33 +523,39 @@ trait EntityComponent {
      *                       that if there is a cycle some of entityIds may be in the result anyway
      * @return the ids of all the entities referred to by entityIds
      */
-    private def recursiveGetEntityReferenceIds(direction: RecursionDirection, entityIds: Set[Long], accumulatedIds: Set[Long]): ReadAction[Set[Long]] = {
-      def oneLevelDown(idBatch: Set[Long]): ReadAction[Set[Long]] = {
-        val query = filter(_.id inSetBind idBatch) join
-          entityAttributeQuery on { (ent, attr) => ent.id === attr.ownerId && ! attr.deleted } filter (_._2.valueEntityRef.isDefined) map (_._2.valueEntityRef.get)
+    private def recursiveGetEntityReferenceIds(direction: RecursionDirection, entityIds: Map[Seq[AttributeEntityReference], Long], accumulatedIds: Map[Seq[AttributeEntityReference], Long]): ReadAction[Map[Seq[AttributeEntityReference], Long]] = {
+      def oneLevelDown(idBatch: Set[Long]): ReadAction[Set[EntityRecord]] = {
+        val query = entityAttributeQuery filter (_.ownerId inSetBind idBatch) join
+          this on { (attr, ent) => attr.valueEntityRef === ent.id && ! attr.deleted } map (_._2)
         query.result.map(_.toSet)
       }
 
-      def oneLevelUp(idBatch: Set[Long]): ReadAction[Set[Long]] = {
-        val query = entityAttributeQuery filter (_.valueEntityRef.isDefined) filter (_.valueEntityRef inSetBind idBatch) join
-          this on { (attr, ent) => attr.ownerId === ent.id && ! ent.deleted } map (_._2.id)
+      def oneLevelUp(idBatch: Set[Long]): ReadAction[Set[EntityRecord]] = {
+        val query = entityAttributeQuery filter (_.valueEntityRef inSetBind idBatch) join
+          this on { (attr, ent) => attr.ownerId === ent.id && ! ent.deleted } map (_._2)
         query.result.map(_.toSet)
       }
 
       // need to batch because some RDBMSes have a limit on the length of an in clause
-      val batchedEntityIds: Iterable[Set[Long]] = createBatches(entityIds)
+      val batchedEntityIds: Iterable[Set[Long]] = createBatches(entityIds.values.toSet)
 
-      val batchActions: Iterable[ReadAction[Set[Long]]] = direction match {
+      val batchActions: Iterable[ReadAction[Set[EntityRecord]]] = direction match {
         case Down => batchedEntityIds map oneLevelDown
         case Up => batchedEntityIds map oneLevelUp
       }
 
-      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refIds =>
-        val untraversedIds = refIds -- accumulatedIds
+      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refRecords =>
+        val refRecordsById = refRecords.map(x => x.id -> x).toMap
+
+        val pathWithLeaf = entityIds.map { case (path, id) =>
+          (path :+ AttributeEntityReference(refRecordsById(id).entityType, refRecordsById(id).name), id)
+        }
+
+        val untraversedIds = pathWithLeaf -- accumulatedIds.keys
         if (untraversedIds.isEmpty) {
           DBIO.successful(accumulatedIds)
         } else {
-          recursiveGetEntityReferenceIds(direction, untraversedIds, accumulatedIds ++ refIds)
+          recursiveGetEntityReferenceIds(direction, untraversedIds, accumulatedIds ++ pathWithLeaf)
         }
       }
     }
@@ -557,14 +563,14 @@ trait EntityComponent {
     /**
      * used in copyEntities load all the entities to copy
      */
-    def getEntitySubtrees(workspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadAction[TraversableOnce[Entity]] = {
-      val startingEntityIdsAction = filter(rec => rec.workspaceId === workspaceContext.workspaceId && rec.entityType === entityType && rec.name.inSetBind(entityNames)).map(_.id)
+    def getEntitySubtrees(workspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadAction[Seq[Seq[AttributeEntityReference]]] = {
+      val startingEntityIdsAction = filter(rec => rec.workspaceId === workspaceContext.workspaceId && rec.entityType === entityType && rec.name.inSetBind(entityNames))
 
       startingEntityIdsAction.result.flatMap { startingEntityIds =>
-        val idSet = startingEntityIds.toSet
+        val idSet = startingEntityIds.map(x => Seq(AttributeEntityReference(x.entityType, x.name)) -> x.id).toMap
         recursiveGetEntityReferenceIds(Down, idSet, idSet)
-      } flatMap { ids =>
-        unmarshalEntities(EntityAndAttributesRawSqlQuery.actionForIds(ids))
+      } map { ids =>
+        ids.keys.toSeq
       }
     }
 
@@ -583,62 +589,71 @@ trait EntityComponent {
         }
       }
 
-      def getSoftConflicts(entityType: String, entityNames: Seq[String]) = {
-        DBIO.sequence(entityNames.map { entityName =>
-          getEntitySubtrees(sourceWorkspaceContext, entityType, Seq(entityName)).flatMap { entities =>
-            getCopyConflicts(destWorkspaceContext, entities).flatMap { conflicts =>
-              for {
-                rootEntity <- get(sourceWorkspaceContext, entityType, entityName)
-              } yield ConflictSubtree(rootEntity.get, conflicts.toSeq, entities.toSeq)
-            }
-          }
-        })
-      }
-
-      def buildConflictSubtree(topLevelEntityRefs: Map[Entity, Seq[AttributeEntityReference]], subtree: ConflictSubtree, result: Seq[EntitySoftConflict]): Seq[EntitySoftConflict] = {
-        if(topLevelEntityRefs.isEmpty) result
+      def buildConflictSubtreeNew(pathsToConflicts: Map[AttributeEntityReference, Seq[Seq[AttributeEntityReference]]], results: Seq[EntitySoftConflict]): Seq[EntitySoftConflict] = {
+        //entityPaths contains the paths to all entities in the subtree, but not all of those are necessarily conflicts
+        //for each key in this map, call the function again
+        if(pathsToConflicts.tail.isEmpty) results
         else {
-          topLevelEntityRefs.map { case (entity, refs) =>
-            val refsToCareAbout = subtree.conflicts.filter(c => refs.contains(AttributeEntityReference(c.entityType, c.name)))
-            val entitiesAndRefsToCareAbout = refsToCareAbout.map(e => e -> getAllAttrRefs(e.attributes)).toMap
-
-            EntitySoftConflict(entity.entityType, entity.name, buildConflictSubtree(entitiesAndRefsToCareAbout, subtree, result))
-          }.toSeq
+          val nextLevels = pathsToConflicts.map(x => x._2.groupBy(y => y.head).map(z => (z._1, z._2.tail)))
+          pathsToConflicts.map { case (topEntity, paths) => EntitySoftConflict(topEntity.entityType, topEntity.entityName, nextLevels.flatMap(nextLevel => buildConflictSubtreeNew(nextLevel, results)).toSeq) }.toSeq
         }
       }
+
+      def getSoftConflicts(entityType: String, entityNames: Seq[String]) = {
+        getEntitySubtrees(sourceWorkspaceContext, entityType, entityNames).flatMap { entities =>
+          getCopyConflicts(destWorkspaceContext, entities.map(x => x.last)).map { conflicts =>
+            val conflictsAsRefs = conflicts.map(_.toReference).toSeq
+            val pathsToConflicts = entities.filter(x => conflictsAsRefs.contains(x.last))
+            val topLevel = pathsToConflicts.groupBy(x => x.head).map(y => y._1 -> y._2.tail)
+
+            buildConflictSubtreeNew(topLevel, Seq.empty)
+          }
+        }
+      }
+
+//      def buildConflictSubtree(topLevelEntityRefs: Map[Entity, Seq[AttributeEntityReference]], subtree: ConflictSubtree, result: Seq[EntitySoftConflict]): Seq[EntitySoftConflict] = {
+//        if(topLevelEntityRefs.isEmpty) result
+//        else {
+//          topLevelEntityRefs.map { case (entity, refs) =>
+//            val refsToCareAbout = subtree.conflicts.filter(c => refs.contains(AttributeEntityReference(c.entityType, c.name)))
+//            val entitiesAndRefsToCareAbout = refsToCareAbout.map(e => e -> getAllAttrRefs(e.attributes)).toMap
+//
+//            EntitySoftConflict(entity.entityType, entity.name, buildConflictSubtree(entitiesAndRefsToCareAbout, subtree, result))
+//          }.toSeq
+//        }
+//      }
 
       val entitiesToCopyRefs = entityNames.map(name => AttributeEntityReference(entityType, name))
 
       getHardConflicts(destWorkspaceContext.workspaceId, entitiesToCopyRefs).flatMap {
         case Seq() => getSoftConflicts(entityType, entityNames).flatMap { softConflicts =>
-          val allEntities = softConflicts.flatMap(_.allChildren)
-          val allConflicts = softConflicts.flatMap(_.conflicts)
+          //val allConflicts = softConflicts.flatMap(_.conflicts)
 
-          if(allConflicts.isEmpty || linkExistingEntities) {
-            val entitiesToClone = allEntities diff allConflicts
-            cloneEntities(destWorkspaceContext, entitiesToClone.toSet, allConflicts.map(_.toReference)).map { _ =>
-              EntityCopyResponse(entitiesToClone.map(_.toReference), Seq.empty, Seq.empty)
+          if(softConflicts.isEmpty || linkExistingEntities) {
+            //val entitiesToClone = allEntities diff allConflicts
+            cloneEntities(destWorkspaceContext, Set.empty, Seq.empty).map { _ =>
+              EntityCopyResponse(Seq.empty[Entity].map(_.toReference), Seq.empty, Seq.empty)
             }
           } else {
-            DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, softConflicts.flatMap(c => buildConflictSubtree(Map(c.entity -> getAllAttrRefs(c.entity.attributes)), c, Seq.empty))))
+            DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, softConflicts))
           }
         }
         case hardConflicts => DBIO.successful(EntityCopyResponse(Seq.empty, hardConflicts.map(c => EntityHardConflict(c.entityType, c.entityName)).toSeq, Seq.empty))
       }
     }
 
-    def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[Entity]): ReadAction[TraversableOnce[Entity]] = {
-      list(destWorkspaceContext, entitiesToCopy.map(_.toReference).toSeq)
+    def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[AttributeEntityReference]): ReadAction[TraversableOnce[EntityRecord]] = {
+      lookupEntitiesByNames(destWorkspaceContext.workspaceId, entitiesToCopy.toSeq)
     }
 
     // the opposite of getEntitySubtrees: traverse the graph to retrieve all entities which ultimately refer to these
 
     def getAllReferringEntities(context: SlickWorkspaceContext, entities: Set[AttributeEntityReference]): ReadAction[Set[AttributeEntityReference]] = {
       lookupEntitiesByNames(context.workspaceId, entities) flatMap { entityRecs =>
-        val idSet = entityRecs.map(_.id).toSet
+        val idSet = entityRecs.map(x => Seq(AttributeEntityReference(x.entityType, x.name)) -> x.id).toMap
         recursiveGetEntityReferenceIds(Up, idSet, idSet)
       } flatMap { ids =>
-        val entityAction = unmarshalEntities(EntityAndAttributesRawSqlQuery.actionForIds(ids))
+        val entityAction = unmarshalEntities(EntityAndAttributesRawSqlQuery.actionForIds(ids.values))
         entityAction map { _.toSet map { e: Entity => e.toReference } }
       }
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -487,6 +487,8 @@ trait EntityComponent {
     }
 
     def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entitiesToClone: TraversableOnce[Entity], entitiesToReference: Seq[Entity]): ReadWriteAction[Int] = {
+      println(s"EntitiesToClone: ${entitiesToClone}")
+      println(s"entitiesToReference: ${entitiesToReference}")
       batchInsertEntities(destWorkspaceContext, entitiesToClone.toSeq.map(marshalNewEntity(_, destWorkspaceContext.workspaceId))) flatMap { ids =>
         val entityIdByEntity = ids.map(record => record.id -> entitiesToClone.filter(p => p.entityType == record.entityType && p.name == record.name).toSeq.head)
         val entityIdsByRef = entityIdByEntity.map { case (entityId, entity) => entity.toReference -> entityId }.toMap
@@ -567,11 +569,11 @@ trait EntityComponent {
       }
     }
 
-    case class ConflictSubtree(entity: Entity, children: Seq[Entity]) {
-      def hasConflicts = children.nonEmpty
+    case class ConflictSubtree(entity: Entity, conflicts: Seq[Entity], allChildren: Seq[Entity]) {
+      def hasConflicts = conflicts.nonEmpty
     }
 
-    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadWriteAction[EntityCopyResponse] = {
+    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean): ReadWriteAction[EntityCopyResponse] = {
 
       def getAllAttrRefs(attrs: AttributeMap): Seq[AttributeEntityReference] = {
         attrs.value.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq
@@ -586,10 +588,10 @@ trait EntityComponent {
       def getSoftConflicts(entityType: String, entityNames: Seq[String]) = {
         DBIO.sequence(entityNames.map { entityName =>
           getEntitySubtrees(sourceWorkspaceContext, entityType, Seq(entityName)).flatMap { entities =>
-            getCopyConflicts(destWorkspaceContext, entities).flatMap { entities =>
+            getCopyConflicts(destWorkspaceContext, entities).flatMap { conflicts =>
               for {
                 rootEntity <- get(sourceWorkspaceContext, entityType, entityName)
-              } yield ConflictSubtree(rootEntity.get, entities.toSeq)
+              } yield ConflictSubtree(rootEntity.get, conflicts.toSeq, entities.toSeq)
             }
           }
         })
@@ -599,8 +601,8 @@ trait EntityComponent {
         if(topLevelEntityRefs.isEmpty) result
         else {
           topLevelEntityRefs.map { case (entity, refs) =>
-            val refsToCareAbout = subtree.children.filter(x => refs.contains(AttributeEntityReference(x.entityType, x.name)))
-            val entitiesAndRefsToCareAbout = refsToCareAbout.map(x => x -> getAllAttrRefs(x.attributes)).toMap
+            val refsToCareAbout = subtree.conflicts.filter(x => refs.contains(AttributeEntityReference(x.entityType, x.name)))
+            val entitiesAndRefsToCareAbout = refsToCareAbout.map(x => x -> x.attributes.values.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq).toMap
 
             EntitySoftConflict(entity.entityType, entity.name, buildConflictSubtree(entitiesAndRefsToCareAbout, subtree, result))
           }.toSeq
@@ -609,17 +611,18 @@ trait EntityComponent {
 
       getHardConflicts(entityType, entityNames).flatMap {
         case Seq() => getSoftConflicts(entityType, entityNames).flatMap { softConflicts =>
-          val nonEmptySoftConflicts = softConflicts.filter(_.hasConflicts)
 
-          if(nonEmptySoftConflicts.isEmpty) {
+          val allEntities = softConflicts.flatMap(_.allChildren)
+          val allConflicts = softConflicts.flatMap(_.conflicts)
+
+          if(allConflicts.isEmpty || linkExistingEntities) {
               //make sure to clone the entities and provide the ones to reference!
-            cloneEntities(destWorkspaceContext, Seq.empty, Seq.empty).map { _ =>
+            cloneEntities(destWorkspaceContext, allEntities diff allConflicts, allConflicts).map { _ =>
               EntityCopyResponse(Seq.empty, Seq.empty, Seq.empty)
             }
           } else {
-            DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, softConflicts.flatMap(c => buildConflictSubtree(Map(c.entity -> getAllAttrRefs(c.entity.attributes)), c, Seq.empty))))
+            DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, softConflicts.flatMap(c => buildConflictSubtree(Map(c.entity -> c.entity.attributes.values.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq), c, Seq.empty))))
           }
-
         }
         case hardConflicts => DBIO.successful(EntityCopyResponse(Seq.empty, hardConflicts.map(c => EntityHardConflict(c.entityType, c.name)), Seq.empty))
       }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -601,29 +601,14 @@ trait EntityComponent {
 
           if(nonEmptySoftConflicts.isEmpty) {
               //make sure to clone the entities and provide the ones to reference!
-            cloneEntities()
             EntityCopyResponse(Seq.empty, Seq.empty, Seq.empty)
           } else {
-            EntityCopyResponse(Seq.empty, Seq.empty, softConflicts.flatMap(x => buildConflictSubtree(Map(x.entity -> x.entity.attributes.values.map(_.asInstanceOf[AttributeEntityReference]).toSeq), x, Seq.empty)))//recursiveBuildConflictTree(nonEmptySoftConflicts.filter(x => x.entity.entityType.equalsIgnoreCase(entityType) && entityNames.contains(x.entity.name)), Seq.empty, Seq.empty))
+            EntityCopyResponse(Seq.empty, Seq.empty, softConflicts.flatMap(c => buildConflictSubtree(Map(c.entity -> c.entity.attributes.values.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq), c, Seq.empty)))
           }
 
         }
         case hardConflicts => DBIO.successful(EntityCopyResponse(Seq.empty, hardConflicts.map(c => EntityHardConflict(c.entityType, c.name)), Seq.empty))
       }
-
-
-//      entityNames.map { name =>
-//        getConflictSubtrees(entityType, entityNames).flatMap { conflictSubtrees =>
-//          conflictSubtrees.filter(_.hasConflicts)
-//
-
-//          if (conflicts.isEmpty) {
-//            cloneEntities(destWorkspaceContext, entities).map(_ => Seq.empty[Entity])
-//          } else {
-//            DBIO.successful(conflicts)
-//          }
-//        }
-//      }
     }
 
     def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[Entity]): ReadAction[TraversableOnce[Entity]] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -557,16 +557,56 @@ trait EntityComponent {
       }
     }
 
-    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadWriteAction[TraversableOnce[Entity]] = {
-      getEntitySubtrees(sourceWorkspaceContext, entityType, entityNames).flatMap { entities =>
-        getCopyConflicts(destWorkspaceContext, entities).flatMap { conflicts =>
-          if (conflicts.isEmpty) {
-            cloneEntities(destWorkspaceContext, entities).map(_ => Seq.empty[Entity])
-          } else {
-            DBIO.successful(conflicts)
-          }
+    case class ConflictSubtree(entity: Entity, children: Seq[Entity]) {
+      def hasConflicts = children.nonEmpty
+    }
+
+    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadWriteAction[EntityCopyResponse] = {
+
+      def getHardConflicts(entityType: String, entityNames: Seq[String]) = {
+        DBIO.sequence(entityNames.map(name => get(destWorkspaceContext, entityType, name))).map { entityOption =>
+          entityOption.collect { case Some(e) => e }
         }
       }
+
+      def getSoftConflicts(entityType: String, entityNames: Seq[String]) = {
+        DBIO.sequence(entityNames.map { entityName =>
+          getEntitySubtrees(sourceWorkspaceContext, entityType, Seq(entityName)).flatMap { entities =>
+            getCopyConflicts(destWorkspaceContext, entities).flatMap { entities =>
+              for {
+                rootEntity <- get(sourceWorkspaceContext, entityType, entityName)
+              } yield ConflictSubtree(rootEntity.get, entities.toSeq)
+            }
+          }
+        })
+      }
+
+      getHardConflicts(entityType, entityNames).flatMap {
+        case Seq() => getSoftConflicts(entityType, entityNames).map { softConflicts =>
+          val nonEmptySoftConflicts = softConflicts.filter(_.hasConflicts)
+
+          if(nonEmptySoftConflicts.isEmpty) {
+              //make sure to clone the entities and provide the ones to reference!
+            EntityCopyResponse(Seq.empty, Seq.empty, Seq.empty)
+          } else EntityCopyResponse(Seq.empty, Seq.empty, nonEmptySoftConflicts.map(_.entity))
+
+        }
+        case hardConflicts => DBIO.successful(EntityCopyResponse(Seq.empty, hardConflicts, Seq.empty))
+      }
+
+
+//      entityNames.map { name =>
+//        getConflictSubtrees(entityType, entityNames).flatMap { conflictSubtrees =>
+//          conflictSubtrees.filter(_.hasConflicts)
+//
+
+//          if (conflicts.isEmpty) {
+//            cloneEntities(destWorkspaceContext, entities).map(_ => Seq.empty[Entity])
+//          } else {
+//            DBIO.successful(conflicts)
+//          }
+//        }
+//      }
     }
 
     def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[Entity]): ReadAction[TraversableOnce[Entity]] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -577,7 +577,7 @@ trait EntityComponent {
 
       def getHardConflicts(workspaceId: UUID, entityRefs: Seq[AttributeEntityReference]) = {
         val batchActions = createBatches(entityRefs.toSet).map(batch => lookupEntitiesByNames(workspaceId, batch))
-        DBIO.sequence(batchActions).map(_.flatten.toSet).map { recs =>
+        DBIO.sequence(batchActions).map(_.flatten.toSeq).map { recs =>
             recs.map(rec => AttributeEntityReference(rec.entityType, rec.name))
         }
       }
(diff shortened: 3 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,15 +565,60 @@ trait EntityComponent {
...
-          if (conflicts.isEmpty) {
-            cloneEntities(destWorkspaceContext, entities).map(_ => Seq.empty[Entity])
+    case class ConflictSubtree(entity: Entity, conflicts: Seq[Entity], allChildren: Seq[Entity]) {
+      def hasConflicts = conflicts.nonEmpty
+    }
+
+    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean): ReadWriteAction[EntityCopyResponse] = {
+
+      def getAllAttrRefs(attrs: AttributeMap): Seq[AttributeEntityReference] = {
+        attrs.values.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq
```
MatthewBemis: suggestions to make this not be horrible?

jmthibault79: I haven't looked closely yet but `filter` + `map` can often be `collect` instead

.gitignore
```diff
@@ -14,4 +14,5 @@ test-output/
 /bin/
 *.log
 config/
-rawlsdb/
\ No newline at end of file
+rawlsdb/
+*.pid
```
jmthibault79: why is this here?

MatthewBemis: Isaac committed this so I defer to @zarsky-broad for this one

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -125,9 +125,13 @@ trait EntityApiService extends HttpService with PerRequestCreator with UserInfoD
     } ~
     path("workspaces" / "entities" / "copy" ) {
       post {
-        entity(as[EntityCopyDefinition]) { copyDefinition =>
-          requestContext => perRequest(requestContext, WorkspaceService.props(workspaceServiceConstructor, userInfo),
-            WorkspaceService.CopyEntities(copyDefinition, requestContext.request.uri))
+        parameters('linkExistingEntities.?) { (linkExistingEntities) =>
+          val linkExistingEntitiesBool = Try(linkExistingEntities.getOrElse("false").toBoolean).getOrElse(false)
```
jmthibault79, line 129: I want a prettier way to express this but so far my mental attempts are no better

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -160,6 +160,14 @@ case class EntityQueryResultMetadata(unfilteredCount: Int, filteredCount: Int, f
 
 case class EntityQueryResponse(parameters: EntityQuery, resultMetadata: EntityQueryResultMetadata, results: Seq[Entity])
 
+case class EntityCopyResponse(entitiesCopied: Seq[EntityShort], hardConflicts: Seq[EntityHardConflict], softConflicts: Seq[EntitySoftConflict])
+
+case class EntitySoftConflict(entityType: String, name: String, conflicts: Seq[EntitySoftConflict])
+
+case class EntityHardConflict(entityType: String, name: String)
+
+case class EntityShort(entityType: String, name: String)
```
jmthibault79: same as `AttributeEntityReference` so we can use that

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1618,6 +1618,45 @@ class EntityApiServiceSpec extends ApiServiceSpec {
       }
   }
 
+  it should "return 409 for copying entities into a workspace with subtree conflicts, but successfully copy when asked to" in withTestDataApiServices { services =>
```
jmthibault79: Please add asserts for the results returned in the existing copy-conflict tests (around lines 209 and 1609)

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1618,6 +1618,45 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+    val sourceWorkspace = WorkspaceName(testData.workspace.namespace, testData.workspace.name)
+    val entityCopyDefinition1 = EntityCopyDefinition(sourceWorkspace, testData.controlledWorkspace.toWorkspaceName, testData.sample1.entityType, Seq(testData.sample1.name))
+    val entityCopyDefinition2 = EntityCopyDefinition(sourceWorkspace, testData.controlledWorkspace.toWorkspaceName, testData.sample3.entityType, Seq(testData.sample3.name))
+
+    Post("/workspaces/entities/copy", httpJson(entityCopyDefinition1)) ~>
+      sealRoute(services.entityRoutes) ~>
+      check {
+        assertResult(StatusCodes.Created, response.entity.asString) {
```
jmthibault79, line 1686: Add asserts for expected results for all of these

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1618,6 +1618,45 @@ class EntityApiServiceSpec extends ApiServiceSpec {
       }
   }
 
+  it should "return 409 for copying entities into a workspace with subtree conflicts, but successfully copy when asked to" in withTestDataApiServices { services =>
+    val sourceWorkspace = WorkspaceName(testData.workspace.namespace, testData.workspace.name)
+    val entityCopyDefinition1 = EntityCopyDefinition(sourceWorkspace, testData.controlledWorkspace.toWorkspaceName, testData.sample1.entityType, Seq(testData.sample1.name))
+    val entityCopyDefinition2 = EntityCopyDefinition(sourceWorkspace, testData.controlledWorkspace.toWorkspaceName, testData.sample3.entityType, Seq(testData.sample3.name))
```
jmthibault79: Add a comment explaining why this causes a soft conflict (reference to sample1)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,47 @@ trait EntityComponent {
...
 
-      allEntitiesAction.flatMap(cloneEntities(destWorkspaceContext, _))
+      allEntitiesAction.flatMap(cloneEntities(destWorkspaceContext, _, Seq.empty))
     }
 
     def batchInsertEntities(workspaceContext: SlickWorkspaceContext, entities: Seq[EntityRecord]): ReadWriteAction[Seq[EntityRecord]] = {
       if(!entities.isEmpty) {
         workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
-          insertInBatches(entityQuery, entities) andThen selectEntityIds(workspaceContext, entities)
+          insertInBatches(entityQuery, entities) andThen selectEntityIds(workspaceContext, entities.map(e => (e.entityType, e.name)))
```
jmthibault79: `entities.map(_.toReference)`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,47 @@ trait EntityComponent {
...
       }
     }
 
-    def selectEntityIds(workspaceContext: SlickWorkspaceContext, entities: Seq[EntityRecord]): ReadAction[Seq[EntityRecord]] = {
+    def selectEntityIds(workspaceContext: SlickWorkspaceContext, entities: Seq[(String, String)]): ReadAction[Seq[EntityRecord]] = {
       val entitiesGrouped = entities.grouped(batchSize).toSeq
 
       DBIO.sequence(entitiesGrouped map { batch =>
-        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map(_.toReference))
+        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map{ case (entityType, name) => AttributeEntityReference(entityType, name) })
```
jmthibault79: I think this will just be `batch` if you change the `entities` param to `Seq[AttributeEntityReference]`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,47 @@ trait EntityComponent {
...
-            entityAttributeQuery.marshalAttribute(entityId, attributeName, attr, entityIdsByRef)
+        val entitiesToReferenceById = selectEntityIds(destWorkspaceContext, entitiesToReference.map(e => (e.entityType, e.name)))
+
+        val attributeRecords = entitiesToReferenceById map { entityRecsToReference =>
+          val entityIdsToReferenceById = entityRecsToReference.map(record => record.id -> entitiesToReference.filter(p => p.entityType == record.entityType && p.name == record.name).head)
+          val entityIdsToReferenceByRef = entityIdsToReferenceById.map { case (entityId, entity) => entity.toReference -> entityId }.toMap
+          entityIdByEntity flatMap { case (entityId, entity) =>
+            entity.attributes.flatMap { case (attributeName, attr) =>
+              entityAttributeQuery.marshalAttribute(entityId, attributeName, attr, (entityIdsByRef ++ entityIdsToReferenceByRef))
+            }
```
dvoet: I am finding this function very confusing. First I think changing entitiesToReference to be Seq[AttributeEntityReference] will be helpful. Then there are a lot of lines of code to get a map of entity ref to database id and it is mixed with marshaling attributes. I would have 1 function call to take entitiesToReference and return Map[AttributeEntityReference, Long]. Last, the nested flatMaps are a pain, try using a for comprehension instead.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
dvoet: attrs.values.collect { case ref: AttributeEntityReference => ref }.toSeq

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,15 +565,60 @@ trait EntityComponent {
...
+
+    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean): ReadWriteAction[EntityCopyResponse] = {
+
+      def getAllAttrRefs(attrs: AttributeMap): Seq[AttributeEntityReference] = {
+        attrs.values.filter(_.isInstanceOf[AttributeEntityReference]).map(_.asInstanceOf[AttributeEntityReference]).toSeq
+      }
+
+      def getHardConflicts(entityType: String, entityNames: Seq[String]) = {
+        DBIO.sequence(entityNames.map(name => get(destWorkspaceContext, entityType, name))).map { entityOption =>
+          entityOption.collect { case Some(e) => e }
```
dvoet: this is doing a query per entity, DBIO.sequence is usually the wrong thing to use. Use inSetBind possibly in batches (see DriverComponent.createBatches) and then it is ok to use DBIO.sequence and put the batches back together.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,15 +565,60 @@ trait EntityComponent {
...
+        }
+      }
+
+      def getSoftConflicts(entityType: String, entityNames: Seq[String]) = {
+        DBIO.sequence(entityNames.map { entityName =>
+          getEntitySubtrees(sourceWorkspaceContext, entityType, Seq(entityName)).flatMap { entities =>
```
dvoet: this is even worse because getEntitySubtrees is recursive so it is a recursive query per entity. But look, getEntitySubtrees was written to deal with a batch but you pass in Seq(entityName) instead.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,15 +565,60 @@ trait EntityComponent {
...
+      }
+
+      getHardConflicts(entityType, entityNames).flatMap {
+        case Seq() => getSoftConflicts(entityType, entityNames).flatMap { softConflicts =>
+          val allEntities = softConflicts.flatMap(_.allChildren)
+          val allConflicts = softConflicts.flatMap(_.conflicts)
+
+          if(allConflicts.isEmpty || linkExistingEntities) {
+            cloneEntities(destWorkspaceContext, allEntities diff allConflicts, allConflicts).map { _ =>
+              EntityCopyResponse(Seq.empty, Seq.empty, Seq.empty)
```
dvoet: the first param to EntityCopyResponse should not be empty

dvoet (changes requested):
I am going to stop reviewing. Please go back over the code with the thought of copying 1000 entities. How many db queries will that be?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,47 @@ trait EntityComponent {
...
       DBIO.sequence(entitiesGrouped map { batch =>
-        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map(_.toReference))
+        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map{ case (entityType, name) => AttributeEntityReference(entityType, name) })
       }).map(_.flatten)
     }
 
-    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entities: TraversableOnce[Entity]): ReadWriteAction[Int] = {
-      batchInsertEntities(destWorkspaceContext, entities.toSeq.map(marshalNewEntity(_, destWorkspaceContext.workspaceId))) flatMap { ids =>
-        val entityIdByEntity = ids.map(record => record.id -> entities.filter(p => p.entityType == record.entityType && p.name == record.name).toSeq.head)
+    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entitiesToClone: TraversableOnce[Entity], entitiesToReference: Seq[Entity]): ReadWriteAction[Int] = {
```
jmthibault79: I'm guessing you have tried refactoring this already but since the method boils down to Insert Entities and Insert Attributes -- can this be simplified?  It is difficult to understand what's going on here.

.gitignore
zarsky-broad: @jmthibault79 The docker run script creates a `.pid` file while running, to facilitate cleanup in case it's interrupted. This is already ignored in the other repos.

jmthibault79: Seems a fine thing to exclude but not in this PR

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jmthibault79: OK seeing Doug's comments I will pause my review of EntityComponent

.gitignore
zarsky-broad: As you prefer, we haven't given it its own PR in the other repos but we certainly could here.

MatthewBemis: I'll kick it out in my next commit

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -784,27 +784,23 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
         withWorkspaceContextAndPermissions(entityCopyDef.sourceWorkspace, WorkspaceAccessLevels.Read, dataAccess) { sourceWorkspaceContext =>
           realmCheck(sourceWorkspaceContext, destWorkspaceContext) flatMap { _ =>
             val entityNames = entityCopyDef.entityNames
             val entityType = entityCopyDef.entityType
-            val copyResults = dataAccess.entityQuery.copyEntities(sourceWorkspaceContext, destWorkspaceContext, entityType, entityNames)
-            copyResults.flatMap(conflicts => conflicts.size match {
-              case 0 => {
+            val copyResults = dataAccess.entityQuery.copyEntities(sourceWorkspaceContext, destWorkspaceContext, entityType, entityNames, linkExistingEntities)
+            copyResults.flatMap(response => (response.hardConflicts.isEmpty, response.softConflicts.isEmpty || linkExistingEntities) match {
+              case (true, true) => {
```
jmthibault79: this is really just an `if` so you can do that :)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -784,27 +784,23 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
                 // get the entities that were copied into the destination workspace
                 dataAccess.entityQuery.list(destWorkspaceContext, entityType).map { allEntities =>
                   val entityCopies = allEntities.filter((e: Entity) => entityNames.contains(e.name)).toList
-                  RequestComplete(StatusCodes.Created, entityCopies)
+                  RequestComplete(StatusCodes.Created, response.copy(entitiesCopied = entityCopies.map(_.toReference).toSeq))
```
jmthibault79: Why can't we return the response as-is?  I don't like this.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -608,8 +622,8 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
           runAndWait(entityQuery.getCopyConflicts(context, Seq(testData.sample1))).toSet
         }
 
-        assertResult(Set(testData.sample1, testData.aliquot1)) {
```
jmthibault79: If `aliquoit1` is now a Soft Conflict, we should assert that too.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: This basically just preserves the behavior that already exists but it looks like this response has always been wrong because it doesn't return any entities that were copied from within the subtree, it only returns top level entities. So I'm already planning to re-work this for that reason.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -618,6 +632,41 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
 
   }
 
+  it should "copy entities with a conflict in the entity subtrees and properly link already existing entities" in withDefaultTestDatabase {
+
+    runAndWait(workspaceQuery.save(workspace2))
+    runAndWait(workspaceQuery.save(workspace3))
+    withWorkspaceContext(workspace2) { context1 =>
```
jmthibault79: oh no, rename these travesties

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -618,6 +632,41 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+
+        assertResult(List()){
+          runAndWait(entityQuery.list(context1, "sample")).toList
+        }
+
+        assertResult(List(participant1)){
+          runAndWait(entityQuery.list(context1, "participant")).toList
+        }
+
+        runAndWait(entityQuery.copyEntities(context2, context1, "sample", Seq("sample1"), true))
```
jmthibault79: assert for the expected result too

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
jmthibault79: ok thanks

jmthibault79:
Done this round of review

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
MatthewBemis: line ~209 is a create conflict test, but I fixed everywhere else

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
MatthewBemis: soft conflicts are not checked if a hard conflict is found

core/src/main/resources/swagger/rawls.yaml
```diff
@@ -2080,7 +2080,7 @@ paths:
           schema:
             $ref: '#/definitions/ErrorReport'
         '409':
-          description: One or more entities with the destination name/type already exist
+          description: There are conflicts in your copy operation. Retry with the paramater linkExistingEntities=true to link all soft conflicts to the entities that already exist in your workspace.
           schema:
             $ref: '#/definitions/ErrorReport'
```
dvoet: conflict tree model goes here

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,50 @@ trait EntityComponent {
...
+        val entityIdByEntity = ids.map(record => record.id -> entitiesToClone.filter(p => p.entityType == record.entityType && p.name == record.name).toSeq.head)
         val entityIdsByRef = entityIdByEntity.map { case (entityId, entity) => entity.toReference -> entityId }.toMap
 
-        val attributeRecords = entityIdByEntity flatMap { case (entityId, entity) =>
-          entity.attributes.flatMap { case (attributeName, attr) =>
-            entityAttributeQuery.marshalAttribute(entityId, attributeName, attr, entityIdsByRef)
-          }
+        val entitiesToReferenceById = selectEntityIds(destWorkspaceContext, entitiesToReference)
+
+        def recsToRefById(entityRecsToReference: Seq[EntityRecord]) = {
```
dvoet: rewrite
`entityRecsToReference.map(record => entitiesToReference.find(_ == record.toReference).get -> record.id).toMap`

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
jmthibault79: Thanks for the variable rename :) but I am still waiting on this one

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,50 @@ trait EntityComponent {
...
+          val entityRefById = entityRecsToReference.map(record => record.id -> entitiesToReference.filter(p => p.entityType == record.entityType && p.entityName == record.name).head)
+          entityRefById.map { case (entityId, entity) => entity -> entityId }.toMap
+        }
+
+        val attributeRecords = entitiesToReferenceById map { entityRecsToReference =>
+          for {
+            (entityId, entity) <- entityIdByEntity
+            (attributeName, attr) <- entity.attributes
+            rec <- entityAttributeQuery.marshalAttribute(entityId, attributeName, attr, (entityIdsByRef ++ recsToRefById(entityRecsToReference)))
```
dvoet: recsToRefById is a def that is called for every single attribute to calculate the exact same map!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+     * @param accumulatedPaths the ids accumulated from the prior call. If you wish entityIds to be in the overall
      *                       results, start with entityIds == accumulatedIds, otherwise start with Seq.empty but note
      *                       that if there is a cycle some of entityIds may be in the result anyway
      * @return the ids of all the entities referred to by entityIds
      */
-    private def recursiveGetEntityReferenceIds(direction: RecursionDirection, entityIds: Set[Long], accumulatedIds: Set[Long]): ReadAction[Set[Long]] = {
-      def oneLevelDown(idBatch: Set[Long]): ReadAction[Set[Long]] = {
-        val query = filter(_.id inSetBind idBatch) join
-          entityAttributeQuery on { (ent, attr) => ent.id === attr.ownerId && ! attr.deleted } filter (_._2.valueEntityRef.isDefined) map (_._2.valueEntityRef.get)
+    private def recursiveGetEntityReferenceIds(direction: RecursionDirection, entityIds: Set[Long], accumulatedEntities: Map[Long, AttributeEntityReference], accumulatedPaths: Map[Seq[AttributeEntityReference], Long]): ReadAction[Seq[Seq[AttributeEntityReference]]] = {
```
jmthibault79: Rename: now returns references instead of IDs

Can the return Action type be flattened from a `Seq[Seq]`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
       }
     }
 
     /**
      * used in copyEntities load all the entities to copy
      */
-    def getEntitySubtrees(workspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadAction[TraversableOnce[Entity]] = {
-      val startingEntityIdsAction = filter(rec => rec.workspaceId === workspaceContext.workspaceId && rec.entityType === entityType && rec.name.inSetBind(entityNames)).map(_.id)
+    def getEntitySubtrees(workspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String]): ReadAction[Seq[Seq[AttributeEntityReference]]] = {
+      val startingEntityIdsAction = filter(rec => rec.workspaceId === workspaceContext.workspaceId && rec.entityType === entityType && rec.name.inSetBind(entityNames))
```
jmthibault79: rename: entity records now

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
 
       startingEntityIdsAction.result.flatMap { startingEntityIds =>
-        val idSet = startingEntityIds.toSet
-        recursiveGetEntityReferenceIds(Down, idSet, idSet)
-      } flatMap { ids =>
-        unmarshalEntities(EntityAndAttributesRawSqlQuery.actionForIds(ids))
+        val idSet = startingEntityIds.map(rec => Seq(AttributeEntityReference(rec.entityType, rec.name)) -> rec.id).toMap
```
jmthibault79: Rename (not a set, also not IDs), use `rec.toReference`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+        recursiveGetEntityReferenceIds(Down, startingEntityIds.map(_.id).toSet, startingEntityIds.map(rec => rec.id -> AttributeEntityReference(rec.entityType, rec.name)).toMap, idSet)
```
jmthibault79: Break out the third param into a `val` above this line.
Also, the third param and the fourth param seem to be the same map in a different order.  That may be fine but I just want to double check that this makes sense.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+              val grouped = unmergedSoftConflicts.groupBy(c => (c.entityType, c.entityName))
+              DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, grouped.map{ case ((entityType, entityName), conflicts) => EntitySoftConflict(entityType, entityName, conflicts.flatMap(_.conflicts))}.toSeq))
+            }
+          }
+        }
+        case hardConflicts => DBIO.successful(EntityCopyResponse(Seq.empty, hardConflicts.map(c => EntityHardConflict(c.entityType, c.entityName)), Seq.empty))
+      }
+    }
+
+    def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[AttributeEntityReference]): ReadAction[TraversableOnce[EntityRecord]] = {
```
jmthibault79, line 615: What is the purpose of this method?  It does not match its name.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jmthibault79: My comments here also apply to `getAllReferringEntities()`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -784,28 +784,18 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-                  RequestComplete(StatusCodes.Created, entityCopies)
-                }
-              }
-              case _ => {
-                val basePath = s"/${destWorkspaceContext.workspace.namespace}/${destWorkspaceContext.workspace.name}/entities/"
-                val conflictingUris = conflicts.map(conflict => ErrorReport(uri.copy(path = Uri.Path(basePath + s"${conflict.entityType}/${conflict.name}")).toString(), Seq.empty))
-                DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict, "Unable to copy entities. Some entities already exist.", conflictingUris.toSeq)))
-              }
-            })
+            val copyResults = dataAccess.entityQuery.copyEntities(sourceWorkspaceContext, destWorkspaceContext, entityType, entityNames, linkExistingEntities)
```
jmthibault79, line 794: Glorious!

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -544,12 +572,10 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
         // note: we're copying FROM workspace2 INTO workspace
         assertResult(Seq.empty) {
-          runAndWait(entityQuery.getCopyConflicts(context1, Seq(x1, x2_updated)))
+          runAndWait(entityQuery.getCopyConflicts(context1, Seq(x1, x2_updated).map(_.toReference)))
         }
 
-        assertResult(Seq.empty) {
-          runAndWait(entityQuery.copyEntities(context2, context1, "SampleSet", Seq("x2")))
-        }
+        Seq(AttributeEntityReference(x1.entityType, x1.name), AttributeEntityReference(x2.entityType, x2.name)) should contain theSameElementsAs runAndWait(entityQuery.copyEntities(context2, context1, "SampleSet", Seq("x2"), false)).entitiesCopied
```
jmthibault79: `x1.toReference`, `x2.toReference`

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -585,17 +612,13 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
         // note: we're copying FROM workspace2 INTO workspace
         assertResult(Seq.empty) {
-          runAndWait(entityQuery.getCopyConflicts(context1, Seq(a1, a2, a3, a4, a5, a6, a)))
+          runAndWait(entityQuery.getCopyConflicts(context1, allEntities.map(_.toReference)))
         }
 
-        assertResult(Seq.empty) {
-          runAndWait(entityQuery.copyEntities(context2, context1, "test", Seq("a1")))
-        }
+        allEntities.map(_.toReference) should contain theSameElementsAs runAndWait(entityQuery.copyEntities(context2, context1, "test", Seq("a1"), false)).entitiesCopied
```
jmthibault79: This looks like safe usage but please convert these to `assertSameElements` instead

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: I think that the name is accurate. It returns the subset of `entitiesToCopy` that exist in `destWorkspaceContext`. 

The changes to this are just a performance improvement on what was already there. Previously it would look up each entity in full one by one. The only information we need from this method is which entities in `entitiesToCopy` exist and we can do that in one lightweight query instead of `n` full attribute joins

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
      * Starting with entities specified by entityIds, recursively walk references accumulating all the ids
      * @param direction whether to walk Down (my object references others) or Up (my object is referenced by others) the graph
      * @param entityIds the ids to start with
-     * @param accumulatedIds the ids accumulated from the prior call. If you wish entityIds to be in the overall
+     * @param accumulatedPaths the ids accumulated from the prior call. If you wish entityIds to be in the overall
```
jmthibault79: add `accumulatedEntities` and check that the param comments for both are still accurate

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
-      val batchActions: Iterable[ReadAction[Set[Long]]] = direction match {
+      val batchActions: Iterable[ReadAction[Set[(Long, EntityRecord)]]] = direction match {
         case Down => batchedEntityIds map oneLevelDown
         case Up => batchedEntityIds map oneLevelUp
       }
 
-      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refIds =>
-        val untraversedIds = refIds -- accumulatedIds
+      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refRecords =>
+        val newAccumulatedEntities = accumulatedEntities ++ refRecords.map(rec => rec._2.id -> rec._2.toReference)
```
jmthibault79: Add a comment about why we want the child entity's id here instead of the parent's

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
 
-      val batchActions: Iterable[ReadAction[Set[Long]]] = direction match {
+      val batchActions: Iterable[ReadAction[Set[(Long, EntityRecord)]]] = direction match {
         case Down => batchedEntityIds map oneLevelDown
         case Up => batchedEntityIds map oneLevelUp
       }
 
-      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refIds =>
-        val untraversedIds = refIds -- accumulatedIds
+      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { refRecords =>
```
jmthibault79: I would prefer a name like "pair" because it's not just a record

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+        val newAccumulatedEntities = accumulatedEntities ++ refRecords.map(rec => rec._2.id -> rec._2.toReference)
+
+        val newAccumulatedPaths = refRecords.flatMap { rec =>
+          val keysThatEndWithParent = accumulatedPaths.keys.filter(_.last.equals(accumulatedEntities(rec._1)))
+          val pathsToExtend = accumulatedPaths.filter(p => keysThatEndWithParent.toSeq.contains(p._1))
```
jmthibault79: I don't understand what this line is doing.  Is it just to associate the map's values with the `keysThatEndWithParent` ?  If that's true, we can probably combine the two lines and only filter once.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jmthibault79: Oh right, it's the destination!  Thanks.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+          val extendedPaths = pathsToExtend.map(p => (p._1 :+ rec._2.toReference, rec._2.id))
+
+          accumulatedPaths ++ extendedPaths
+        }.toMap
+
+        val untraversedIds = refRecords.map(_._2.id) -- accumulatedEntities.keys
         if (untraversedIds.isEmpty) {
-          DBIO.successful(accumulatedIds)
+          DBIO.successful(accumulatedPaths.keys.toSeq)
```
jmthibault79: Should this be `newAccumulatedPaths` ?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
       }
-    }
 
-    def getCopyConflicts(destWorkspaceContext: SlickWorkspaceContext, entitiesToCopy: TraversableOnce[Entity]): ReadAction[TraversableOnce[Entity]] = {
-      val entityQueries = entitiesToCopy.map { entity =>
-        findEntityByName(destWorkspaceContext.workspaceId, entity.entityType, entity.name).result.map {
-          case Seq() => None
-          case _ => Option(entity)
+      def getSoftConflicts(paths: Seq[Seq[AttributeEntityReference]]) = {
+        getCopyConflicts(destWorkspaceContext, paths.map(_.last)).map { conflicts =>
```
jmthibault79: Why do we only look at `_.last` here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
-          if (conflicts.isEmpty) {
-            cloneEntities(destWorkspaceContext, entities).map(_ => Seq.empty[Entity])
-          } else {
-            DBIO.successful(conflicts)
-          }
+    def copyEntities(sourceWorkspaceContext: SlickWorkspaceContext, destWorkspaceContext: SlickWorkspaceContext, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean): ReadWriteAction[EntityCopyResponse] = {
+      def getHardConflicts(workspaceId: UUID, entityRefs: Seq[AttributeEntityReference]) = {
+        val batchActions = createBatches(entityRefs.toSet).map(batch => lookupEntitiesByNames(workspaceId, batch))
+        DBIO.sequence(batchActions).map(_.flatten.toSeq).map { recs =>
+          recs.map(rec => AttributeEntityReference(rec.entityType, rec.name))
```
jmthibault79: `_.toReference`

core/src/main/resources/swagger/rawls.yaml
jmthibault79: `EntityCopyResponse`

jmthibault79: Also for the 201/Created case

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -784,28 +784,18 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
     }
 
-  def copyEntities(entityCopyDef: EntityCopyDefinition, uri: Uri): Future[PerRequestMessage] =
+  def copyEntities(entityCopyDef: EntityCopyDefinition, uri: Uri, linkExistingEntities: Boolean): Future[PerRequestMessage] =
     dataSource.inTransaction { dataAccess =>
       withWorkspaceContextAndPermissions(entityCopyDef.destinationWorkspace, WorkspaceAccessLevels.Write, dataAccess) { destWorkspaceContext =>
         withWorkspaceContextAndPermissions(entityCopyDef.sourceWorkspace, WorkspaceAccessLevels.Read, dataAccess) { sourceWorkspaceContext =>
           realmCheck(sourceWorkspaceContext, destWorkspaceContext) flatMap { _ =>
```
jmthibault79, line 791: This could be `andThen {` instead, if you like

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+              val allConflictRefs = softConflicts.flatten
+              val entitiesToClone = allEntityRefs diff allConflictRefs
+
+              for {
+                entities <- list(sourceWorkspaceContext, entitiesToClone)
+                _ <- cloneEntities(destWorkspaceContext, entities.toSet, allConflictRefs)
+              } yield EntityCopyResponse(entities.map(_.toReference).toSeq, Seq.empty, Seq.empty)
+            } else {
+              val unmergedSoftConflicts = softConflicts.flatMap(c => buildSoftConflictTree(c, Seq.empty))
+              val grouped = unmergedSoftConflicts.groupBy(c => (c.entityType, c.entityName))
```
jmthibault79: `groupBy(_.toReference)`

jmthibault79: wait I got carried away.  SoftConflict doesn't have a toReference :)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+              DBIO.successful(EntityCopyResponse(Seq.empty, Seq.empty, grouped.map{ case ((entityType, entityName), conflicts) => EntitySoftConflict(entityType, entityName, conflicts.flatMap(_.conflicts))}.toSeq))
```
jmthibault79: Why is it necessary to group after generating the tree?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,50 @@ trait EntityComponent {
...
-        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map(_.toReference))
+        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch)
       }).map(_.flatten)
     }
 
-    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entities: TraversableOnce[Entity]): ReadWriteAction[Int] = {
-      batchInsertEntities(destWorkspaceContext, entities.toSeq.map(marshalNewEntity(_, destWorkspaceContext.workspaceId))) flatMap { ids =>
-        val entityIdByEntity = ids.map(record => record.id -> entities.filter(p => p.entityType == record.entityType && p.name == record.name).toSeq.head)
+    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entitiesToClone: TraversableOnce[Entity], entitiesToReference: Seq[AttributeEntityReference]): ReadWriteAction[Int] = {
+      batchInsertEntities(destWorkspaceContext, entitiesToClone.toSeq.map(marshalNewEntity(_, destWorkspaceContext.workspaceId))) flatMap { ids =>
```
jmthibault79: these are records, not IDs

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: re: rename- will do
re: changing return type: no it must be a seq of seqs otherwise we'll lose important information about certain intermediate leaves in the path

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -463,39 +463,50 @@ trait EntityComponent {
...
       DBIO.sequence(entitiesGrouped map { batch =>
-        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch.map(_.toReference))
+        EntityRecordRawSqlQuery.action(workspaceContext.workspaceId, batch)
       }).map(_.flatten)
     }
 
-    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entities: TraversableOnce[Entity]): ReadWriteAction[Int] = {
-      batchInsertEntities(destWorkspaceContext, entities.toSeq.map(marshalNewEntity(_, destWorkspaceContext.workspaceId))) flatMap { ids =>
-        val entityIdByEntity = ids.map(record => record.id -> entities.filter(p => p.entityType == record.entityType && p.name == record.name).toSeq.head)
+    def cloneEntities(destWorkspaceContext: SlickWorkspaceContext, entitiesToClone: TraversableOnce[Entity], entitiesToReference: Seq[AttributeEntityReference]): ReadWriteAction[Int] = {
```
jmthibault79: I had some ideas about refactoring this and then I couldn't help myself: https://github.com/broadinstitute/rawls/tree/jt_loop_refactor

You're welcome to add this to the PR (ffwd) or ignore it.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: if we have the entity tree:

```
sample_set
   ^- sampleA
      ^- participant1
   ^- sampleB
      ^- participant2
```

we should have something like the following for the `paths`:

```
Seq(
Seq(sample_set), 
Seq(sample_set, sampleA), 
Seq(sample_set, sampleA, participant1), 
Seq(sample_set, sampleB), 
Seq(sample_set, sampleB, participant2)
)
```

So we'll pull off sample_set, sampleA, participant1, sampleB, and participant2. I suppose this is equivalent to a flattened and de-duped `paths` but this is just how my brain spit it out as I was imagining the whole thing... but that doesn't mean it's the most clear way for me to write it :)

Off the top of my head I can't think of a case where these two ways of writing could be different

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: Thinking back on it, the method name `buildSoftConflictTree` is kind of a lie. It's more like `buildSoftConflictPaths`. But to illustrate what's going on, I'll build off of the same example from my other comment:

given the entity tree: 
```
sample_set
   ^- sampleA
      ^- participant1
   ^- sampleB
      ^- participant2
```

we should have something like the following for the paths:

```
Seq(
Seq(sample_set), 
Seq(sample_set, sampleA), 
Seq(sample_set, sampleA, participant1), 
Seq(sample_set, sampleB), 
Seq(sample_set, sampleB, participant2)
)
```

Assume `participant1` and `participant2` are soft conflicts. `softConflicts` would end up being:

```
Seq(
Seq(sample_set, sampleA, participant1), 
Seq(sample_set, sampleB, participant2)
)
```

We need to merge those paths into a tree based on their root, otherwise we'll end up with something like:

```
Seq(
EntitySoftConflict(sample_set, sample_set, Seq(EntitySoftConflict(sample, sampleA, EntitySoftConflict(participant, participant1, Seq.empty))))
EntitySoftConflict(sample_set, sample_set, Seq(EntitySoftConflict(sample, sampleB, EntitySoftConflict(participant, participant2, Seq.empty))))
)
```

when really we want something like this:

```
Seq(
EntitySoftConflict(sample_set, sample_set, Seq(EntitySoftConflict(sample, sampleA, EntitySoftConflict(participant, participant1, Seq.empty)), EntitySoftConflict(sample, sampleB, EntitySoftConflict(participant, participant2, Seq.empty))))
)
```

I will rename some variables/functions related to building the conflict tree as well as leave some comments.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
      * Starting with entities specified by entityIds, recursively walk references accumulating all the ids
      * @param direction whether to walk Down (my object references others) or Up (my object is referenced by others) the graph
      * @param entityIds the ids to start with
-     * @param accumulatedIds the ids accumulated from the prior call. If you wish entityIds to be in the overall
+     * @param accumulatedPaths the ids accumulated from the prior call. If you wish entityIds to be in the overall
      *                       results, start with entityIds == accumulatedIds, otherwise start with Seq.empty but note
      *                       that if there is a cycle some of entityIds may be in the result anyway
      * @return the ids of all the entities referred to by entityIds
```
dvoet: update return value to include information about paths

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
dvoet: get rid of all the _1s and _2s with case statements

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
dvoet: indeed and it is concerning that the tests don't pick that up

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
+          accumulatedPaths ++ extendedPaths
+        }.toMap
+
+        val untraversedIds = refRecords.map(_._2.id) -- accumulatedEntities.keys
         if (untraversedIds.isEmpty) {
-          DBIO.successful(accumulatedIds)
+          DBIO.successful(accumulatedPaths.keys.toSeq)
         } else {
-          recursiveGetEntityReferenceIds(direction, untraversedIds, accumulatedIds ++ refIds)
+          recursiveGetEntityReferenceIds(direction, untraversedIds, newAccumulatedEntities, newAccumulatedPaths)
```
dvoet: I took a stab at rewriting this because the block computing newAccumulatedPaths is not right. Also you should not need the accumulatedEntities parameter.

```
    private def recursiveGetEntityReferenceIds(direction: RecursionDirection, entityIds: Set[Long], accumulatedPaths: Map[Seq[AttributeEntityReference], Long]): ReadAction[Seq[Seq[AttributeEntityReference]]] = {
      def oneLevelDown(idBatch: Set[Long]): ReadAction[Set[(Long, EntityRecord)]] = {
        val query = entityAttributeQuery filter (_.ownerId inSetBind idBatch) join
          this on { (attr, ent) => attr.valueEntityRef === ent.id && ! attr.deleted } map (rec => (rec._1.ownerId, rec._2))
        query.result.map(_.toSet)
      }

      def oneLevelUp(idBatch: Set[Long]): ReadAction[Set[(Long, EntityRecord)]] = {
        val query = entityAttributeQuery filter (_.valueEntityRef inSetBind idBatch) join
          this on { (attr, ent) => attr.ownerId === ent.id && ! ent.deleted } map (rec => (rec._1.valueEntityRef.get, rec._2))
        query.result.map(_.toSet)
      }

      // need to batch because some RDBMSes have a limit on the length of an in clause
      val batchedEntityIds: Iterable[Set[Long]] = createBatches(entityIds)

      val batchActions: Iterable[ReadAction[Set[(Long, EntityRecord)]]] = direction match {
        case Down => batchedEntityIds map oneLevelDown
        case Up => batchedEntityIds map oneLevelUp
      }

      val accumulatedPathsById = accumulatedPaths.map(e => e._2 -> e._1)
      DBIO.sequence(batchActions).map(_.flatten.toSet).flatMap { parentIdsAndChildRecs =>
        val childPaths = parentIdsAndChildRecs.map { case (parentId, childRec) =>
          (accumulatedPathsById(parentId) :+ childRec.toReference) -> childRec.id
        }.toMap

        val newAccumulatedPaths = accumulatedPaths ++ childPaths
        val untraversedIds = parentIdsAndChildRecs.map { case (_, childRec) => childRec.id } -- accumulatedPathsById.keySet

        if (untraversedIds.isEmpty) {
          DBIO.successful(newAccumulatedPaths.keys.toSeq)
        } else {
          recursiveGetEntityReferenceIds(direction, untraversedIds, newAccumulatedPaths)
        }
      }
    }
```

the "get entity subtrees from a list of entities" test passes... but I am not to confident in that test

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -507,86 +518,126 @@ trait EntityComponent {
...
-          case _ => Option(entity)
+      def getSoftConflicts(paths: Seq[Seq[AttributeEntityReference]]) = {
+        getCopyConflicts(destWorkspaceContext, paths.map(_.last)).map { conflicts =>
+          val conflictsAsRefs = conflicts.toSeq.map(_.toReference)
+          paths.filter(p => conflictsAsRefs.contains(p.last))
         }
       }
-      DBIO.sequence(entityQueries).map(_.toStream.collect { case Some(e) => e })
+
+      def buildSoftConflictTree(pathsRemaining: Seq[AttributeEntityReference], result: Seq[EntitySoftConflict]): Seq[EntitySoftConflict] = {
```
dvoet: I don't think result is required... it is always Seq.empty is it not?

dvoet (changes requested):
I still have more to go but I though I would get what I have so far out to you

dvoet (approved):

core/src/main/resources/swagger/rawls.yaml
```diff
@@ -3932,6 +3930,61 @@ definitions:
...
+        items:
+          $ref: '#/definitions/AttributeEntityReference'
+      hardConflicts:
+        type: array
+        description: ''
+        items:
+          $ref: '#/definitions/EntityHardConflict'
+      softConflicts:
+        type: array
+        description: ''
```
jmthibault79: Add note that this will be empty if any hard conflicts were detected

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -517,13 +517,13 @@ trait EntityComponent {
     private def recursiveGetEntityReferences(direction: RecursionDirection, entityIds: Set[Long], accumulatedEntities: Map[Long, AttributeEntityReference], accumulatedPaths: Map[Seq[AttributeEntityReference], Long]): ReadAction[Seq[Seq[AttributeEntityReference]]] = {
       def oneLevelDown(idBatch: Set[Long]): ReadAction[Set[(Long, EntityRecord)]] = {
         val query = entityAttributeQuery filter (_.ownerId inSetBind idBatch) join
-          this on { (attr, ent) => attr.valueEntityRef === ent.id && ! attr.deleted } map (rec => (rec._1.ownerId, rec._2))
+          this on { (attr, ent) => attr.valueEntityRef === ent.id && ! attr.deleted } map { case (attr, entity) => (attr.ownerId, entity)}
```
jmthibault79: yay

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -481,24 +481,24 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
-                                Seq(testData.sset2.toReference, testData.sample2.toReference))
-      val sampleSet3Paths = Seq(Seq(testData.sset3.toReference),
-                                Seq(testData.sset3.toReference, testData.sample5.toReference),
-                                Seq(testData.sset3.toReference, testData.sample6.toReference))
+      val sampleSet1Paths = Seq(EntityPath(Seq(testData.sset1.toReference)),
+                                EntityPath(Seq(testData.sset1.toReference, testData.sample1.toReference)),
+                                EntityPath(Seq(testData.sset1.toReference, testData.sample1.toReference, testData.aliquot1.toReference)),
+                                EntityPath(Seq(testData.sset1.toReference, testData.sample2.toReference)),
+                                EntityPath(Seq(testData.sset1.toReference, testData.sample3.toReference)),
+                                EntityPath(Seq(testData.sset1.toReference, testData.sample3.toReference, testData.sample1.toReference)))
```
jmthibault79, line 489: why doesn't this path include `aliquiot1` ?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1615,6 +1615,65 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+      }
+
+    //test the default case of no parameter set
+    Post("/workspaces/entities/copy", httpJson(entityCopyDefinition2)) ~>
+      sealRoute(services.entityRoutes) ~>
+      check {
+        assertResult(StatusCodes.Conflict, response.entity.asString) {
+          status
+        }
+        assertResult(EntityCopyResponse(Seq.empty, Seq.empty, Seq(EntitySoftConflict(testData.sample3.entityType, testData.sample3.name, Seq(EntitySoftConflict(testData.sample1.entityType, testData.sample1.name, Seq.empty), EntitySoftConflict(testData.sample1.entityType, testData.sample1.name, Seq(EntitySoftConflict(testData.aliquot1.entityType, testData.aliquot1.name, Seq.empty)))))))) {
```
jmthibault79: Please format this line (and 1663 - hm, maybe make this a val?) in a way that shows the hierarchy.  Something like:

```
assertResult(EntityCopyResponse(Seq.empty, Seq.empty, Seq(
          EntitySoftConflict(testData.sample3.entityType, testData.sample3.name, Seq(
            EntitySoftConflict(testData.sample1.entityType, testData.sample1.name, Seq.empty),
            EntitySoftConflict(testData.sample1.entityType, testData.sample1.name, Seq(EntitySoftConflict(testData.aliquot1.entityType, testData.aliquot1.name, Seq.empty)))))))) {
```

jmthibault79:
Done this round.  A few more questions.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1615,6 +1615,65 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+
+        assertResult(EntityCopyResponse(Seq.empty, Seq(EntityHardConflict("Sample", "sample1")), Seq.empty)) {
+          responseAs[EntityCopyResponse]
+        }
+      }
+  }
+
+  it should "return 409 for copying entities into a workspace with subtree conflicts, but successfully copy when asked to" in withTestDataApiServices { services =>
+    val sourceWorkspace = WorkspaceName(testData.workspace.namespace, testData.workspace.name)
+    val entityCopyDefinition1 = EntityCopyDefinition(sourceWorkspace, testData.controlledWorkspace.toWorkspaceName, testData.sample1.entityType, Seq(testData.sample1.name))
```
jmthibault79, line 1679: Add another test which copies `aliquot1` into an empty workspace and verify that also copying `sample3` causes a Soft Conflict

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
MatthewBemis: because the path to `aliquot1` from `sample1` was already explored in `EntityPath(Seq(testData.sset1.toReference, testData.sample1.toReference, testData.aliquot1.toReference)),`. i need to test this to make sure it's 100% true but this behavior is what allows us to deal with cycles without getting stuck in a loop

jmthibault79: hm.  How does a Soft Conflict actually get used?  Maybe as long as you have one path to a soft-conflicting entity, you don't need all paths to it.

jmthibault79:
Yeah!  👍

PR #680, opened 2017-04-21
Gawb 796 adding api for send changed workspace notifications

UrsaStutsman:
*WORK IN PROGRESS*

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/NotificationsApiService.scala
```diff
@@ -1,15 +1,30 @@
...
  * Created by dvoet on 3/28/17.
  */
-trait NotificationsApiService extends HttpService {
+trait NotificationsApiService extends HttpService with PerRequestCreator with UserInfoDirectives {
+  implicit val executionContext: ExecutionContext
+
+  val workspaceServiceConstructor: UserInfo => WorkspaceService
+  //this route requires UserInfo
+  val sendChangeNotificationRoute = requireUserInfo() { userInfo =>
+    path("notifications"/ "workspace" / Segment / Segment / "sendChangeNotification") { (namespace, name) =>
```
dvoet: I would put this in the notificationRoutes below

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,24 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+        withWorkspaceContext(workspaceName, dataAccess) {workspaceContext =>
+          dataAccess.workspaceQuery.listEmailsAndAccessLevel(workspaceContext)
+        }
+      }
+    }
+
+    for {
+      emails <- getEmails
+    } yield {
+      val notificationMessages = emails.map(_._1).map(email => Notifications.WorkspaceChangedNotification(email, workspaceName))
```
dvoet: as you are probably finding, email is not what you want. You want the subject id for each user. WorkspaceChangedNotification needs to have a recipientUserId member and that should be the subject id

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/NotificationsApiService.scala
breilly2: Actually, does this belong in the notifications service or the workspace service? I think the notifications service is for managing notifications overall whereas this is for sending a specific notification related to a specific workspace. I'm inclined to say this belongs in the workspace service, which would match what was outlined in the design document.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
 
+  def sendChangeNotifications(workspaceName: WorkspaceName) = {
+
+    def getUserEmails = {
+      dataSource.inTransaction{ dataAccess =>
+        withWorkspaceContext(workspaceName, dataAccess) {workspaceContext =>
+          requireAccessIgnoreLock(workspaceContext.workspace, WorkspaceAccessLevels.Write, dataAccess) {
+            DBIO.sequence(workspaceContext.workspace.accessLevels.values.map { group =>
+              dataAccess.rawlsGroupQuery.flattenGroupMembership(group)
+            }.map(_.map(user => userInfo.userEmail)))
```
dvoet: there may be duplicate emails across groups, `_.map` should probably be `_.toSet.map`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  def sendChangeNotifications(workspaceName: WorkspaceName) = {
```
dvoet: put in the return type for this

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+          }
+        }
+      }
+    }
+    for {
+      emails <- getUserEmails
+    } yield {
+      val notificationMessages = emails.map(email => Notifications.WorkspaceChangedNotification(email, workspaceName))
+      notificationDAO.fireAndForgetNotifications(notificationMessages)
+    }
```
dvoet: this for comprehension could be rewritten as

```
    getUserEmails.map { emails =>
      val notificationMessages = emails.map(email => Notifications.WorkspaceChangedNotification(email, workspaceName))
      notificationDAO.fireAndForgetNotifications(notificationMessages)
    }
```

They are equivalent. Usually when there is only 1 element in the for I just go with map. For comprehensions are really handy when you start nesting flatMaps and maps.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
UrsaStutsman: So I maybe misunderstanding something here. But I thought what was getting returned here is a set?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
UrsaStutsman: yeah - I think I was just so excited by finally grasping that syntax I had to use it... but cleaner in this case not to, so making the change...

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
UrsaStutsman: And while commuting home I realized that this is actually probably a set of sets or something like that... So will dig in here...

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  def sendChangeNotifications(workspaceName: WorkspaceName) = {
+
+    def getUserEmails = {
+      dataSource.inTransaction{ dataAccess =>
+        withWorkspaceContext(workspaceName, dataAccess) {workspaceContext =>
+          requireAccessIgnoreLock(workspaceContext.workspace, WorkspaceAccessLevels.Write, dataAccess) {
```
breilly2: This is the same pattern as everywhere else so I'm not requesting a change here. However, it occurs to me that this authorization code is pushed further into the code than I'd like. I think it would be easier to read if it were more declarative, or at least closer to the top/left of the service function. Just planting the thought in case anyone is inspired to improve this.

breilly2:
Looking good. Happy to review or help with any test code that you're working on.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -641,6 +643,28 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+            })
+          }
+        }
+      }
+    }
+    getUsers.map { groups =>
+      val userIds = groups.flatten.map(user => user.userSubjectId.value).toSet
+      val notificationMessages = userIds.map { userId => Notifications.WorkspaceChangedNotification(userId, workspaceName) }
+      notificationDAO.fireAndForgetNotifications(notificationMessages)
+      RequestComplete(StatusCodes.OK, userIds)
```
dvoet: I think reporting back the user ids notified is a) not useful because we don't generally expose user ids and b) a security leak as it allows writers to so who is in the workspaces where generally only owners or people with "can share" access can do this. If you want to return something, perhaps the count. Otherwise return No Content status instead of OK

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1430,6 +1430,47 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
       }
   }
 
+  // Send Change Notification for Workspace require WRITE access. Accept if OWNER or WRITE; Reject if READ or NO ACCESS
```
dvoet: I would like to keep all notification tests in the same place. Please move these to org.broadinstitute.dsde.rawls.webservice.NotificationSpec.

dvoet: Also, look in those tests to see how they check that a notification was actually sent.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
UrsaStutsman: Hmmm. Ok I had added that in to make testing (thoroughly) easier... A count could work. I can refactor towards that...

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
UrsaStutsman: Just making sure - even though we're putting the api for this under Workspaces, the tests should be moved to the notification.spec? Just want to make sure I'm understanding.

dvoet: yes, all the notification testing will be too scattered otherwise

dvoet:
at the end of line 21 of NotificationServiceSpec use MockUserInfoDirectivesWithUser instead of MockUserInfoDirectives, that will fix the security related tests

core/src/test/scala/org/broadinstitute/dsde/rawls/openam/MocklUserInfoDirectivesWithUser.scala
```diff
@@ -0,0 +1,20 @@
...
+import spray.http.OAuth2BearerToken
+import spray.routing.Directive1
+import spray.routing.Directives._
+
+import scala.concurrent.ExecutionContext
+
+/**
+  * Created by ursas on 5/3/17.
+  */
+trait MockUserInfoDirectivesWithUser extends UserInfoDirectives {
```
MatthewBemis: nit: typo in file name?

MatthewBemis (approved):
:+1:

PR #681, opened 2017-04-24
GAWB-1796 renameForHiding using UUIDs, not timestamps

helgridly:
OK, so here's the deal. There's a magic formula that, given a number of records you want to store, tells you how many possible hash outputs your hash function needs to generate if you want to guarantee your probability of a hash collision stays below a certain number.

This code:  
* adds said formula
* asks "how many bits of entropy represent this space of possible hash outputs?"
* figures out how many characters of a string are required to represent this much entropy
* returns that many characters of a randomly generated UUID

In practise this drops the length of the string we append from 23 characters `yyyy-MM-dd_HH.mm.ss.SSS` to 12 characters `DUJV72Gr3KBc` if we have 10 million records.

**Also**: I added a deleted_date column to everything affected by this because I felt bad that we were losing the information encoded in the timestamp.

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
-import java.util.UUID
+import java.sql.Timestamp
+import java.util.{Date, UUID}
 
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
 import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport, model}
@@ -14,7 +15,14 @@ import spray.http.StatusCodes
 /**
  * Created by dvoet on 2/4/16.
  */
-case class EntityRecord(id: Long, name: String, entityType: String, workspaceId: UUID, recordVersion: Long, allAttributeValues: Option[String], deleted: Boolean) {
+case class EntityRecord(id: Long,
+                        name: String,
+                        entityType: String,
+                        workspaceId: UUID,
+                        recordVersion: Long,
+                        allAttributeValues: Option[String],
+                        deleted: Boolean,
+                        deletedDate: Option[Timestamp]) {
   def toReference = AttributeEntityReference(entityType, name)
 }
 
@@ -33,11 +41,12 @@ trait EntityComponent {
     def version = column[Long]("record_version")
     def allAttributeValues = column[Option[String]]("all_attribute_values")
     def deleted = column[Boolean]("deleted")
+    def deletedDate = column[Option[Timestamp]]("deleted_date")
 
     def workspace = foreignKey("FK_ENTITY_WORKSPACE", workspaceId, workspaceQuery)(_.id)
     def uniqueTypeName = index("idx_entity_type_name", (workspaceId, entityType, name), unique = true)
 
-    def * = (id, name, entityType, workspaceId, version, allAttributeValues, deleted) <> (EntityRecord.tupled, EntityRecord.unapply)
+    def * = (id, name, entityType, workspaceId, version, allAttributeValues, deleted, deletedDate) <> (EntityRecord.tupled, EntityRecord.unapply)
   }
 
   object entityQuery extends TableQuery(new EntityTable(_)) {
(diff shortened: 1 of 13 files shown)


jmthibault79 (approved):
Ok but I would prefer "suffix"

helgridly:
@jmthibault79 I made this more complicated because a) it was arithmetically overflowing b) the strings were getting long.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponentSpec.scala
```diff
@@ -77,12 +77,22 @@ class DriverComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
-    assert( seventeenBitPostfix.length == 5 )
+    assert(getNumberOfBitsForSufficientRandomness(65, 1.0/32.0) == 17)
+
+    //check we don't overflow when we have a ton of records:
+    //2^34 records! 17179869184
+    //2^30 is close to 1 in a billion: 1073741824
+    assert(getNumberOfBitsForSufficientRandomness(17179869184L, 1.0/1073741824) == 97)
+
+    //fact: base64 represents every six bits with one character. but we round up to the nearest byte, ergo:
+    // (ceil(ceil(bits/8)*8/6), essentially
```
jmthibault79: Can you make your maths an explicit function?  So the tests look like:

`assert(getRandomStringWithThisManyBitsOfEntropy(59).length == expectedStringLength(59))`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -67,26 +76,26 @@ trait EntityComponent {
...
       }
 
       // the where clause for this query is filled in specific to the use case
       val baseEntityAndAttributeSql =
-        s"""select e.id, e.name, e.entity_type, e.workspace_id, e.record_version, e.deleted,
-          a.id, a.namespace, a.name, a.value_string, a.value_number, a.value_boolean, a.value_json, a.value_entity_ref, a.list_index, a.list_length, a.deleted,
-          e_ref.id, e_ref.name, e_ref.entity_type, e_ref.workspace_id, e_ref.record_version, e_ref.deleted
+        s"""select e.id, e.name, e.entity_type, e.workspace_id, e.record_version, e.deleted, e.deleted_date,
+          a.id, a.namespace, a.name, a.value_string, a.value_number, a.value_boolean, a.value_json, a.value_entity_ref, a.list_index, a.list_length, a.deleted, a.deleted_date,
+          e_ref.id, e_ref.name, e_ref.entity_type, e_ref.workspace_id, e_ref.record_version, e_ref.deleted, e_ref.deleted_date
```
helgridly, line 98: it's an absolute nightmare to find what's gone wrong if you don't do this right

helgridly:
@jmthibault79 are you re-reviewing this still or is this ready to go?

jmthibault79:
re-reviewing

jmthibault79:
Do we still need a `deleted` column if deleted_date is nullable?

helgridly:
I'm assuming a bool is quicker to index in the db than a timestamp.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -14,7 +15,14 @@ import spray.http.StatusCodes
 /**
  * Created by dvoet on 2/4/16.
  */
-case class EntityRecord(id: Long, name: String, entityType: String, workspaceId: UUID, recordVersion: Long, allAttributeValues: Option[String], deleted: Boolean) {
+case class EntityRecord(id: Long,
```
jmthibault79, line 18: +1 to formatting

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jmthibault79: accurate

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -206,7 +206,8 @@ case class MethodConfiguration(
                    outputs: Map[String, AttributeString],
                    methodRepoMethod: MethodRepoMethod,
                    methodConfigVersion: Int = 1,
-                   deleted: Boolean = false
+                   deleted: Boolean = false,
```
jmthibault79, line 209: So ... we expose the fact that the method config has been deleted in the API?  We don't do this for Entities / Attributes.  Hm.  This is something we should think about some more.  (not in this PR)

helgridly: Yes - there's a possible future world where we might a) allow people to click through to deleted MCs from past submissions; b) show the edit history for an MC. These are both very maybe and for now the UI is safe to assume that anything they see is not deleted.

jmthibault79:
Yeah - saving a boolean probably isn't worth that optimization.

Good to go.

PR #689, opened 2017-05-05
[GAWB-2021] Rawls admin endpoint: what's on the workflow queue?

rtitle:
See https://broadinstitute.atlassian.net/browse/GAWB-2021

Added `/api/admin/submissions/queueStatusByUser` endpoint for listing workflow counts by status and by user.

Unit tests pass. I tested locally against the dev DB, and got the following results:
```
{
  "statuses": {
    "Aborting": 937,
    "Submitted": 96
  },
  "users": {
    "joel.broad.dev@gmail.com": {
      "statuses": {
        "Submitted": 1
      }
    },
    "abaumann.firecloud@gmail.com": {
      "statuses": {
        "Aborting": 937
      }
    },
    "test.firec@gmail.com": {
      "statuses": {
        "Submitted": 21
      }
    },
    "obamacloud@gmail.com": {
      "statuses": {
        "Submitted": 74
      }
    }
  },
  "maxActiveWorkflowsTotal": 20000,
  "maxActiveWorkflowsPerUser": 2000
}
```

Here is the new database query:
```
rawls [DEBUG] [18:41:46.278] [slick.db-86] slick.jdbc.JdbcBackend.statement - Preparing statement: 
select x2.`SUBJECT_ID`, x2.`EMAIL`, x3.`ID`, x3.`EXTERNAL_ID`, x3.`SUBMISSION_ID`, x3.`STATUS`, x3.`STATUS_LAST_CHANGED`, x3.`ENTITY_ID`, x3.`record_version`, x3.`EXEC_SERVICE_KEY` from `WORKFLOW` x3, `SUBMISSION` x4, `USER` x2 where (x4.`SUBMITTER` = x2.`SUBJECT_ID`) and ((x3.`STATUS` in (?, ?, ?, ?, ?)) and (x3.`SUBMISSION_ID` = x4.`ID`))
rawls [DEBUG] [18:41:46.561] [slick.db-86] slick.jdbc.JdbcBackend.benchmark - Execution of prepared statement took 283ms
```
And its explain plan (it seems reasonable):
image

  - _RT: see https://github.com/broadinstitute/firecloud-admin-cli/pull/37_
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -275,6 +275,13 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
           )
         }
       }
+    } ~
+    path("admin" / "submissions" / "queueStatusByUser") {
```
rtitle: I wasn't quite sure about the endpoint name -- if anyone has other suggestions, let me know.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -305,6 +313,34 @@ class ExecutionJsonSupport extends JsonSupport {
   implicit val ActiveSubmissionFormat = jsonFormat3(ActiveSubmission)
 
   implicit val WorkflowQueueStatusResponseFormat = jsonFormat3(WorkflowQueueStatusResponse)
+
+  implicit object WorkflowQueueStatusByUserResponseFormat extends RootJsonFormat[WorkflowQueueStatusByUserResponse] {
```
rtitle, line 322: Originally this was a one-liner:
```
implicit val WorkflowQueueStatusByUserResponseFormat = jsonFormat4(WorkflowQueueStatusByUserResponse)
```
...which worked, but I needed a custom formatter to conform to the JSON schema in [GAWB-2021](https://broadinstitute.atlassian.net/browse/GAWB-2021). I think this JSON format is more clear anyway.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -285,6 +287,20 @@ trait WorkflowComponent {
...
 
+    def countWorkflowsByQueueStatusByUser: ReadAction[Map[String, Map[String, Int]]] = {
+      // Query for a Seq of (User, Workflow) tuples.
+      val userWorkflows = (for {
+        workflow <- findQueuedAndRunningWorkflows
+        submission <- submissionQuery if workflow.submissionId === submission.id
+        user <- rawlsUserQuery if submission.submitterId === user.userSubjectId
+      } yield (user, workflow)).result
+
+      // Combine results into a nested map using Scalaz semigroups for aggregation.
```
jmthibault79: Since we don't currently have any other uses of semigroups in Workbench, I'm concerned that other developers (including me) won't be able to reason about what's going on here, from lack of familiarity.

Could you accomplish this in another way, or add a detailed comment about how this works?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
jmthibault79: Seems reasonable to me.  Please move this up to group with the other `submissions` endpoints.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
jmthibault79: Could you add a test for the JSON formatting?  Could be as simple as a round-trip JSON string comparison.

jmthibault79:
Looks good!  Could you add a few more tests which start with adding workflows, and also add some numbers for expected counts to show that we're not just comparing 0 to 0.

A nice way to check Swagger syntax is by pasting it in here: http://bigstickcarpet.com/swagger-parser/www/index.html

I like the Explain Plan diagram - could you show me how you made it sometime?  :)

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
rtitle: Will do.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
rtitle: Heh, I was trying to gauge people's reactions by importing scalaz (I noticed it's already a dependency!).

I feel like a lot of FP ideas have confusing names, but are actually relatively simple concepts. In particular a semigroup abstracts _combining_ things of the same type. Scalaz provides implicit semigroups for several different types, which can all be invoked with the `|+|` operator. 

For example, the semigroup for `Int` is addition:
```
scala> import scalaz._, Scalaz._
import scalaz._
import Scalaz._

scala> 1 |+| 2
res0: Int = 3
```
The semigroup for `List` is concatenation:
```
scala> List(1,2,3) |+| List(4,5,6)
res1: List[Int] = List(1, 2, 3, 4, 5, 6)
```
The semigroup for `Map` groups elements of the same key, and applies the semigroup for the value type:
```
scala> Map('a' -> 1, 'b' -> 2) |+| Map('a' -> 42, 'c' -> 3)
res2: scala.collection.immutable.Map[Char,Int] = Map(a -> 43, c -> 3, b -> 2)
```

You get the idea. So in this case we're using the semigroups for `Map` and `Int` to reduce a list of tuples into a `Map[String, Map[String, Int]]`. 

I could write it without scalaz using standard library functions, but I would argue it would make the code _more_ complicated. :) That said, I can give it a shot.

Also, as I was discussing with @helgridly, I'd be happy to give workbench folks an overview of some FP concepts, like semigroups, monoids, functors, monads, etc if people are interested. :)

rtitle:
Thanks @jmthibault79! I'll add some tests.

Thanks for the swagger tip, it took me a while to find the syntax issue. :(

Sure, I just ran the query in MySQL Workbench and scrolled down to "Execution Plan".

jmthibault79:
Oh cool I should really try out MySQL Workbench!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
jmthibault79: Thanks for the explanation.  There are definitely other places in Rawls where we do this, and it might be good to convert them to semigroups once we're more comfortable.  But for now, I would like a comment along the lines of "this operator combines maps' keys and sums their values"

A general overview of these concepts would be excellent.  We're pretty comfortable with monads (or at least map/flatMap) but not the rest of the family.  Red Team (Cromwell) does more Scalaz.  As we grow closer organizationally maybe they will have more of an influence on us...

helgridly: Jeff has been banging the "we can afford to FP a little harder" drum for a while and I think this might be the thing that pushes us over the edge. I'm running it by the other TLs now; watch this space, but my money's on:

1) general consensus that this is worth doing;
2) a preference for cats over scalaz; and
3) a tech talk or two to properly introduce a couple of these concepts to the team.

rtitle: Yay! I can port this to cats (I think it's the same syntax) and add more explanatory comments if people are ok with it.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -285,6 +287,20 @@ trait WorkflowComponent {
       groupedSeq.map(_.toMap)
     }
 
+    def countWorkflowsByQueueStatusByUser: ReadAction[Map[String, Map[String, Int]]] = {
+      // Query for a Seq of (User, Workflow) tuples.
+      val userWorkflows = (for {
+        workflow <- findQueuedAndRunningWorkflows
+        submission <- submissionQuery if workflow.submissionId === submission.id
+        user <- rawlsUserQuery if submission.submitterId === user.userSubjectId
+      } yield (user, workflow)).result
```
helgridly: So this returns all the workflows from the database, and then you count them in Scala. That's a _lot_ of rows. Why not group by and count in the database?

helgridly: In half-assed pseudo-SQL: `count(*) join workflows, submission, users, group by (workflow.status, user.id)`.

rtitle: That's a good point.. it might be more efficient to do that in the database. Let me try that out next week.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -285,6 +287,20 @@ trait WorkflowComponent {
...
+
+      // Combine results into a nested map using Scalaz semigroups for aggregation.
+      userWorkflows.map(_.foldLeft(Map.empty[String, Map[String, Int]]) { case (result, (user, workflow)) =>
+        result |+| Map(user.userEmail -> Map(workflow.status -> 1))
+      })
+    }
```
danbills: a foldMap would clean this up a little.

```scala
      userWorkflows.map(_.toList.foldMap { case (user, workflow) =>
        Map(user.userEmail -> Map(workflow.status -> 1))
      })
```

rtitle: Cool, thanks for the tip. I'm actually refactoring this to do the aggregation in the database via slick instead of using scalaz. Foldable/foldMap is neat though. :)

rtitle: Btw I came across this, seemed kinda interesting re. slick and cats: https://github.com/RMSone/slick-cats

geoffjentry: Not being published is kind of a pita though :( but yeah

rtitle:
@helgridly @jmthibault79 pushed commit with code review feedback:

1. Added test cases which create some workflows/users and validate expected counts.
2. Added `ExecutionModelSpec` for testing JSON serialization/deserialization.
3. Updated `WorkflowComponent` to do the grouping in the database instead of in Scala. The new query is something like:
```
select x2.`EMAIL`, x3.`STATUS`, count(1) 
from `WORKFLOW` x3, `SUBMISSION` x4, `USER` x2 
where (x4.`SUBMITTER` = x2.`SUBJECT_ID`) 
and ((x3.`STATUS` in ('Queued', 'Launching', 'Submitted', 'Running', 'Aborting')) 
and (x3.`SUBMISSION_ID` = x4.`ID`)) 
group by x2.`EMAIL`, x3.`STATUS`
```
And the explain plan shows the group:
image

model/src/main/scala/org/broadinstitute/dsde/rawls/model/RawlsEnumeration.scala
```diff
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.rawls.model
 
-trait RawlsEnumeration[T <: RawlsEnumeration[T]] { self: T =>
+trait RawlsEnumeration[T <: RawlsEnumeration[T]] extends Product with Serializable { self: T =>
```
rtitle, line 3: Btw, https://github.com/lloydmeta/enumeratum is a nice Scala enum library.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -1153,13 +1173,19 @@ class AdminApiServiceSpec extends ApiServiceSpec {
         val groupedWorkflowRecs = workflowRecs.groupBy(_.status)
           .filterKeys((WorkflowStatuses.queuedStatuses ++ WorkflowStatuses.runningStatuses).map(_.toString).contains)
           .mapValues(_.size)
-        val expected = WorkflowQueueStatusByUserResponse(
+        val userOwnerWorkflows = (testData.userOwner.userEmail.value ->
```
jmthibault79: Some nitpicks: please put the `testUserWorkflows` first and add a comment for the `userOwnerWorkflows` describing that that these are equal to all workflows in the systems besides those of the test user.  Renaming `statusCounts` to something like `testUserStatusCounts` would help to clarify that too.

Since this test would now break if we ever added a third user's workflows to the Default Test Database, this is a good candidate for using `withConstantTestDataApiServices` instead.  We try never to change that one.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/RawlsEnumeration.scala
jmthibault79: That does look nice.

Why was this change needed?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
rtitle: Makes sense, will fix. Good tip about withConstantTestDataApiServices.

jmthibault79:
The new tests are great and the DB logic makes sense.  I'll have to take your word on the `cats` stuff until I understand it.  👍

model/src/main/scala/org/broadinstitute/dsde/rawls/model/RawlsEnumeration.scala
rtitle: Good question, type inference! Basically all case classes automatically extend Product and Serializable, so to avoid it showing up in inferred types (i.e. when using case classes with parent traits in collections) you have to add it to the base trait.

I could have also added it to the WorkflowStatus trait, but I'm kind of assuming that all concrete implementations of RawlsEnumeration will be case classes/objects.

This SO article helps explain it a little better: http://stackoverflow.com/questions/36525804/scala-case-class-extending-product-with-serializable

geoffjentry: Another explanation, including why one should use final as well: http://underscore.io/blog/posts/2015/06/04/more-on-sealed.html

rtitle:
Thanks @jmthibault79! I pushed the unit test improvements.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -214,6 +214,14 @@ case class WorkflowQueueStatusResponse
   workflowCountsByStatus: Map[String, Int]
 )
 
+case class WorkflowQueueStatusByUserResponse
+(
+  statuses: Map[String, Int],
```
helgridly: Can we type alias this as `StatusCounts`, and _possibly_ do something similar for `users`? `users.convertTo[Map[String, Map[String, Map[String, Int]]]]` on 336 below is a bit of a mindbender.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponentSpec.scala
```diff
@@ -251,6 +255,69 @@ class SubmissionComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+      result(st.toString) should be >= count
+    }
+  }
+
+  it should "count workflows by queue status by user" in withDefaultTestDatabase {
+    // Create a new test user and some test submissions
+    val testUserEmail = "testUser"
+    val testUserStatusCounts = Map(WorkflowStatuses.Submitted -> 1, WorkflowStatuses.Running -> 10, WorkflowStatuses.Aborting -> 100)
+    withWorkspaceContext(testData.workspace) { ctx =>
+      val testUser = RawlsUser(UserInfo(testUserEmail, OAuth2BearerToken("token"), 123, "123456789876543212346"))
```
helgridly: Isn't this already `testData.userWriter`? I think this user has already been created and saved.

core/src/test/scala/org/broadinstitute/dsde/rawls/model/ExecutionModelSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+        |  },
+        |  "maxActiveWorkflowsTotal": 2000,
+        |  "maxActiveWorkflowsPerUser": 200
+        |}
+      """.stripMargin.parseJson
+
+    // Verify round trip JSON serialization/deserialization
+    val testJson = testResponse.toJson
+    testJson should equal (expectedJson)
+    testJson.convertTo[WorkflowQueueStatusByUserResponse] should equal (testResponse)
```
helgridly, line 66: Nice.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -1140,4 +1140,56 @@ class AdminApiServiceSpec extends ApiServiceSpec {
...
   }
+
+  it should "get queue status by user" in withConstantTestDataApiServices { services =>
+    import driver.api._
+
+    // Create a new test user and some new submissions
+    val testUserEmail = "testUser"
+    val testUserStatusCounts = Map(WorkflowStatuses.Submitted -> 1, WorkflowStatuses.Running -> 10, WorkflowStatuses.Aborting -> 100)
+    withWorkspaceContext(constantData.workspace) { ctx =>
+      val testUser = RawlsUser(UserInfo(testUserEmail, OAuth2BearerToken("token"), 123, "123456789876543212346"))
```
helgridly: ditto above

helgridly:
A couple more nitpicks (sorry) but after those you're good to go. 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
rtitle: Sure, can add type aliases for `statuses` and `users`.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponentSpec.scala
rtitle: Whoops yeah, copy-paste. Will make the id unique.

helgridly:
Do you need an entirely new user? Or can you reuse one?

rtitle:
I wanted a new user so I could guarantee their expected workflow counts. An existing user might have workflows not controlled by this test.

rtitle:
@helgridly pushed 1 more commit ^, let me know if approved.

helgridly:
do it

PR #692, opened 2017-05-07
GAWB-1697: unmarshal numeric attributes as ints vs. doubles

davidangb:
When we retrieve a numeric attribute from the database, we represent it as a `Double`. Then, when we create an `AttributeNumber` object, we construct the attribute's `BigDecimal` from that `Double` ... which tell the `BigDecimal` that it has a mantissa, even if the mantissa is `.0`. 

Finally, when we serialize the attribute to a json string in the APIs, spray's `JsNumber` is sensitive to the scale of that `BigDecimal`, and will output the `.0`. Thus, if the user supplies an integer for an attribute (say, 12), we'll round-trip and serialize it as a double ("12.0").

This has been a quirk of the API for a while, but hasn't caused any problems. Now, as part of GAWB-1697, this *is* causing problems. We're trying to validate an `AttributeMap` against the Library schema ... and the Library schema defines certain attributes as integers. When we serialize a workspace's pre-existing Library attributes, then validate them, we fail validation because we are supplying doubles instead of integers.

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -433,7 +433,8 @@ trait AttributeComponent {
       if (attributeRec.valueBoolean.isDefined) {
         AttributeBoolean(attributeRec.valueBoolean.get)
       } else if (attributeRec.valueNumber.isDefined) {
-        AttributeNumber(attributeRec.valueNumber.get)
+        val n = attributeRec.valueNumber.get
+        if (n.isValidInt) AttributeNumber(n.toInt) else AttributeNumber(n)
       } else if (attributeRec.valueString.isDefined) {
         AttributeString(attributeRec.valueString.get)
       } else if (attributeRec.valueJson.isDefined) {
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
@@ -234,7 +234,7 @@ class AttributeComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers 
     assertResult(0) { runAndWait(workspaceAttributeQuery.filter(_.ownerId === workspaceId).result).size }
   }
 
-  it should "unmarshall attribute records" in withEmptyTestDatabase {
+  it should "unmarshal attribute records" in withEmptyTestDatabase {
     val attributeRecs = Seq(
       ((1, WorkspaceAttributeRecord(dummyId2, workspaceId, AttributeName.defaultNamespace, "string", Some("value"), None, None, None, None, None, None, false, None)), None),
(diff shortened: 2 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
```diff
@@ -269,6 +269,46 @@ class AttributeComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+      }
+    }
+
+    assertScale("int", 0)
+    assertScale("negInt", 0)
+    assertScale("doubleRounded", 0)
+    assertScale("negDoubleRounded", 0)
+    assertScale("double", 2)
+    assertScale("negDouble", 2)
+  }
```
davidangb, line 310: I considered testing the json serialization, but it felt weird to include jsonformat stuff here, so I tested the lower-level detail (the scale of the BigDecimal).

helgridly:
Doesn't this just invert your problem, so that things defined as doubles get turned into integers if they're .0?

davidangb:
Yes, we have the inverse problem if someone enters n.0, we will deserialize as an integer. However we don't have a use case for that, and it seems a matter of user intention. A REAL solution would be to have an AttributeDouble and AttributeInteger instead of AttributeNumber - as well as different columns in the db - but I didn't want to go that route. If you really object, I will write up a tech doc with all the solution approaches I considered.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -433,7 +433,8 @@ trait AttributeComponent {
       if (attributeRec.valueBoolean.isDefined) {
         AttributeBoolean(attributeRec.valueBoolean.get)
       } else if (attributeRec.valueNumber.isDefined) {
-        AttributeNumber(attributeRec.valueNumber.get)
+        val n = attributeRec.valueNumber.get
```
danbills, line 436: perhaps consider pattern matching on attributeRec.valueBoolean, valueString,etc.

davidangb: Agreed! Was just looking to keep this PR as small as possible.

danbills: The eternal conflict :)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
rtitle: I'm adding an `ExecutionModelSpec` in a different PR (https://github.com/broadinstitute/rawls/pull/689) for testing JSON serializers, maybe we could add some JSON test cases there in the future.

helgridly:
:+1:

PR #694, opened 2017-05-11
[GAWB-1826] Give users access to JES metadata for calls inside a workflow

rtitle:
JIRA:
https://broadinstitute.atlassian.net/browse/GAWB-1826

Related PRs:
https://github.com/broadinstitute/firecloud-orchestration/pull/445
https://github.com/broadinstitute/firecloud-ui/pull/785

Removed `'/api/admin/genomics/operations/{jobId}` and added `/api/workspaces/genomics/operations/{jobId}` (without admin privileges).

Cleaned up some imports using the IDE.

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.rawls
 
-import java.io.{File, StringReader}
-import javax.naming.NameAlreadyBoundException
+import java.io.StringReader
 import javax.naming.directory.AttributeInUseException
 
 import _root_.slick.backend.DatabaseConfig
@@ -12,12 +11,12 @@ import akka.pattern.ask
 import akka.util.Timeout
 import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets
 import com.google.api.client.json.jackson2.JacksonFactory
-import com.typesafe.config.{ConfigRenderOptions, ConfigFactory}
+import com.typesafe.config.{ConfigFactory, ConfigRenderOptions}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.genomics.GenomicsService
 import org.broadinstitute.dsde.rawls.google.HttpGooglePubSubDAO
-import org.broadinstitute.dsde.rawls.jobexec.{WorkflowSubmissionActor, SubmissionSupervisor}
+import org.broadinstitute.dsde.rawls.jobexec.{SubmissionSupervisor, WorkflowSubmissionActor}
 import org.broadinstitute.dsde.rawls.model.{ApplicationVersion, UserInfo}
 import org.broadinstitute.dsde.rawls.monitor._
 import org.broadinstitute.dsde.rawls.statistics.StatisticsService
@@ -28,11 +27,10 @@ import spray.can.Http
 import spray.http.StatusCodes
 import spray.json._
 
-import scala.concurrent.duration._
-import scala.util.{Try, Failure, Success}
-
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.collection.JavaConversions._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
 
 object Boot extends App with LazyLogging {
   private def startup(): Unit = {
@@ -172,6 +170,7 @@ object Boot extends App with LazyLogging {
         submissionSupervisor,
         bucketDeletionMonitor,
         userServiceConstructor,
+        genomicsServiceConstructor,
         maxActiveWorkflowsTotal,
         maxActiveWorkflowsPerUser),
       userServiceConstructor,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -128,7 +128,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def deleteToken(rawlsUserRef: RawlsUserRef): Future[Unit]
   def revokeToken(rawlsUserRef: RawlsUserRef): Future[Unit]
 
-  def getGenomicsOperation(jobId: String): Future[JsObject]
+  def getGenomicsOperation(jobId: String): Future[Option[JsObject]]
 
   def toProxyFromUser(userSubjectId: RawlsUserSubjectId): String
   def toUserFromProxy(proxy: String): String
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -127,7 +127,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def deleteToken(rawlsUserRef: RawlsUserRef): Future[Unit]
   def revokeToken(rawlsUserRef: RawlsUserRef): Future[Unit]
 
-  def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[Option[JsObject]]
+  def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[JsObject]
 
   def toProxyFromUser(userSubjectId: RawlsUserSubjectId): String
   def toUserFromProxy(proxy: String): String
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -127,7 +127,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def deleteToken(rawlsUserRef: RawlsUserRef): Future[Unit]
   def revokeToken(rawlsUserRef: RawlsUserRef): Future[Unit]
 
-  def getGenomicsOperation(jobId: String): Future[Option[JsObject]]
+  def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[Option[JsObject]]
 
   def toProxyFromUser(userSubjectId: RawlsUserSubjectId): String
   def toUserFromProxy(proxy: String): String
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -3,12 +3,13 @@ package org.broadinstitute.dsde.rawls.dataaccess
 import akka.actor.ActorRef
 import com.google.api.client.auth.oauth2.Credential
 import com.google.api.services.admin.directory.model.Group
-import com.google.api.services.storage.model.{BucketAccessControl, Bucket}
+import com.google.api.services.storage.model.{Bucket, BucketAccessControl}
 import org.broadinstitute.dsde.rawls.dataaccess.slick.RawlsBillingProjectOperationRecord
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels._
 import org.broadinstitute.dsde.rawls.model._
 import org.joda.time.DateTime
 import spray.json.JsObject
+
 import scala.concurrent.Future
 import scala.util.Try
 
@@ -127,7 +128,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def deleteToken(rawlsUserRef: RawlsUserRef): Future[Unit]
   def revokeToken(rawlsUserRef: RawlsUserRef): Future[Unit]
 
-  def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[JsObject]
+  def getGenomicsOperation(jobId: String): Future[JsObject]
 
   def toProxyFromUser(userSubjectId: RawlsUserSubjectId): String
   def toUserFromProxy(proxy: String): String
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -753,18 +753,19 @@ class HttpGoogleServicesDAO(
     } )
   }
 
-  override def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[Option[JsObject]] = {
+  override def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[JsObject] = {
     val opId = s"operations/$jobId"
 
     val genomicsApi = new Genomics.Builder(httpTransport, jsonFactory, getUserCredential(userInfo)).setApplicationName(appName).build()
     val operationRequest = genomicsApi.operations().get(opId)
 
-    retryWithRecoverWhen500orGoogleError[Option[JsObject]](() => {
+    retryWithRecoverWhen500orGoogleError[JsObject](() => {
       // Google library returns a Map[String,AnyRef], but we don't care about understanding the response
       // So, use Google's functionality to get the json string, then parse it back into a generic json object
-      Some(executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject)
-    }) {
-      case t: HttpResponseException if t.getStatusCode == StatusCodes.NotFound.intValue => None
+      executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject
+    }) { case e: GoogleJsonResponseException =>
+      val statusCode = StatusCodes.getForKey(e.getStatusCode).getOrElse(StatusCodes.InternalServerError)
+      throw new RawlsExceptionWithErrorReport(ErrorReport(statusCode, e.getDetails.getMessage))
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -8,8 +8,6 @@ import com.google.api.client.auth.oauth2.{Credential, TokenResponse}
 import com.google.api.client.googleapis.auth.oauth2.{GoogleClientSecrets, GoogleCredential}
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
-import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
-import com.google.api.client.http.json.JsonHttpContent
 import com.google.api.client.http.{HttpResponseException, InputStreamContent}
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.admin.directory.model._
@@ -23,7 +21,7 @@ import com.google.api.services.compute.{Compute, ComputeScopes}
 import com.google.api.services.genomics.{Genomics, GenomicsScopes}
 import com.google.api.services.oauth2.Oauth2.Builder
 import com.google.api.services.plus.PlusScopes
-import com.google.api.services.servicemanagement.{model, ServiceManagement}
+import com.google.api.services.servicemanagement.ServiceManagement
 import com.google.api.services.servicemanagement.model.EnableServiceRequest
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule.{Action, Condition}
 import com.google.api.services.storage.model.Bucket.{Lifecycle, Logging}
@@ -37,20 +35,18 @@ import org.broadinstitute.dsde.rawls.model.UserAuthJsonSupport._
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels._
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.monitor.BucketDeletionMonitor.{BucketDeleted, DeleteBucket}
-import org.broadinstitute.dsde.rawls.util.{Retry, FutureSupport}
+import org.broadinstitute.dsde.rawls.util.{FutureSupport, Retry}
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.joda.time
 import spray.client.pipelining._
-import spray.http.HttpHeaders.Authorization
-import spray.http.{HttpResponse, OAuth2BearerToken, StatusCode, StatusCodes}
+import spray.http.{OAuth2BearerToken, StatusCode, StatusCodes}
 import spray.json._
-import spray.json.DefaultJsonProtocol._
 
 import scala.collection.JavaConversions._
-import scala.concurrent.{Future, _}
 import scala.concurrent.duration._
+import scala.concurrent.{Future, _}
 import scala.io.Source
-import scala.util.{Failure, Success, Try}
+import scala.util.{Success, Try}
 
 class HttpGoogleServicesDAO(
   useServiceAccountForBuckets: Boolean,
@@ -753,10 +749,10 @@ class HttpGoogleServicesDAO(
     } )
   }
 
-  override def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[JsObject] = {
+  override def getGenomicsOperation(jobId: String): Future[JsObject] = {
     val opId = s"operations/$jobId"
 
-    val genomicsApi = new Genomics.Builder(httpTransport, jsonFactory, getUserCredential(userInfo)).setApplicationName(appName).build()
+    val genomicsApi = new Genomics.Builder(httpTransport, jsonFactory, getGenomicsServiceAccountCredential).setApplicationName(appName).build()
     val operationRequest = genomicsApi.operations().get(opId)
 
     retryWithRecoverWhen500orGoogleError[JsObject](() => {
@@ -1003,7 +999,6 @@ class HttpGoogleServicesDAO(
       .build()
   }
 
-  // Note: not currently used as we now use the user account for genomics operations
   def getGenomicsServiceAccountCredential: Credential = {
     new GoogleCredential.Builder()
       .setTransport(httpTransport)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -753,13 +753,10 @@ class HttpGoogleServicesDAO(
     } )
   }
 
-  override def getGenomicsOperation(jobId: String): Future[Option[JsObject]] = {
-
-    import spray.json._
-
+  override def getGenomicsOperation(userInfo: UserInfo, jobId: String): Future[Option[JsObject]] = {
     val opId = s"operations/$jobId"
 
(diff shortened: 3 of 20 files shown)


helgridly:
This allows you to read metadata for operation IDs that aren't yours. Should it?

rtitle:
Whoops did I miss some permissions stuff? Should it be doing some workspace access checks (maybe `WorkspaceAccessLevels.Read` or higher)?

I guess that would also mean passing the workspace info in the API call..

helgridly:
a) I misread the PR initially, so that's not true (but there might be a wrinkle regardless); and
b) I should have done a better job explaining the background here -- sorry.

So operation IDs are the JES IDs. They're not associated with workspaces, because JES has no concept of workspaces - it just runs jobs. opIDs _are_ associated with Google projects, so if you ask for the opID metadata as a user, JES will check you're in the relevant project and 40x as appropriate. The wrinkle is that I'm not sure which scopes are required for this operation, so we should find that out, since we'll need to ask the user for them if they haven't already authorized us.

jmthibault79:
I am also unsure about whether we should care if the user can access Operation ID Metadata outside of the Workspace in question.  Let's ask the PO.

If we don't care: 👍 for this PR as-is.

If we do, then requiring `Read`+ access sounds right to me, and I'd make the URL `/api/workspaces/{ns}/{name}/genomics/operations/{jobId}`

helgridly:
Actually, I lied _again_ - `GoogleDAO.getGenomicsOperation` uses rawls' genomics-credential for the query. So this is broken -- we should be using the user's.

helgridly:
Which means we need to:

a) find out where we define the list of scopes the user needs, and make sure `GenomicsScopes.GENOMICS` is in that list;
b) pass in the user's credential (like in [`listBillingAccounts`](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L659)) and execute the Google request as them.

helgridly:
[Here](https://github.com/broadinstitute/firecloud-orchestration/blob/fa61350f915debf800519bc6c7bd80a85530853c/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala#L58) is where we define the scopes we want from the user. Thanks Joel.

rtitle:
Ah wow, thanks guys. Will work on those changes.

rtitle:
Pushed commits to make GoogleServicesDAO.getGenomicsOperation use the user's credential (with the appropriate scopes) instead of a service account. I now can't access my test operation with my user id. :)

```
Response Body
{
  "statusCode": 403,
  "source": "rawls",
  "causes": [],
  "stackTrace": [],
  "message": "Permission denied; need genomics.operations.get permission for operation EO_hv-6-KxibjYjL0OCrmCQgw7vetLsXKg9wcm9kdWN0aW9uUXVldWU"
}
Response Code
403
```

I posted the question about whether we also need to care about workspace permissions to the JIRA: https://broadinstitute.atlassian.net/browse/GAWB-1826?focusedCommentId=91968&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-91968

helgridly:
Ugh. I had assumed that if you have access to a workspace, you have access to the Google Project. But we only give you access to the bucket!

So: the API should live at `/api/workspaces/<namespace>/<name>/operationID/<opID>`, and should check your workspace permissions. Then it should use the Rawls service account (which has genomics scopes) to actually query JES.

Lesson learned: stop assuming things. Mea culpa.

rtitle:
Ok

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -28,11 +27,10 @@ import spray.can.Http
 import spray.http.StatusCodes
 import spray.json._
 
-import scala.concurrent.duration._
-import scala.util.{Try, Failure, Success}
-
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.collection.JavaConversions._
+import scala.concurrent.ExecutionContext.Implicits.global
```
rtitle, line 31: While cleaning up imports I noticed that we're using the global execution context. I was worried that it was also being used for slick blocking calls, but thankfully I see we're using a separate thread pool for slick.

This is probably fine for our case, but just fyi another way to do it is to define dispatchers within akka and reference them using the `ActorSystem`. This allows you to define & tune all execution contexts from within akka.

http://doc.akka.io/docs/akka/current/scala/dispatchers.html

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -127,7 +128,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def deleteToken(rawlsUserRef: RawlsUserRef): Future[Unit]
   def revokeToken(rawlsUserRef: RawlsUserRef): Future[Unit]
 
-  def getGenomicsOperation(jobId: String): Future[Option[JsObject]]
+  def getGenomicsOperation(jobId: String): Future[JsObject]
```
rtitle: Just using Future.failed to represent the exception case instead of Option.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -753,21 +749,19 @@ class HttpGoogleServicesDAO(
...
-    retryWithRecoverWhen500orGoogleError[Option[JsObject]](() => {
+    retryWithRecoverWhen500orGoogleError[JsObject](() => {
       // Google library returns a Map[String,AnyRef], but we don't care about understanding the response
       // So, use Google's functionality to get the json string, then parse it back into a generic json object
-      Some(executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject)
-    }) {
-      case t: HttpResponseException if t.getStatusCode == StatusCodes.NotFound.intValue => None
+      executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject
+    }) { case e: GoogleJsonResponseException =>
+      val statusCode = StatusCodes.getForKey(e.getStatusCode).getOrElse(StatusCodes.InternalServerError)
```
rtitle: This propagates the status code + message from the Google HTTP call to the end user.

rtitle:
@jmthibault79 @helgridly re-review when you have a sec? I think it's correct now :)

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
geoffjentry: Also IIRC Rawls uses the `blocking` construct so if you ever *do* do what @rtitle suggests note that it only does anyting on the global ec

helgridly: We don't use `blocking` often, I don't think.

rtitle: I see use of `blocking` on Google HTTP calls, and a separate thread pool for slick, which is good. :)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: What if it's not a `GoogleJsonResponseException`?

rtitle: This is a `PartialFunction`, so the exception will complete the Future and the API will return a 500 (I think).

helgridly: This is a significant change to the previous behaviour. Before, non-404s would potentially get retried because they'd fall through and get caught in `GoogleUtilities.when500orGoogleError`. Now we're just passing the Google error straight up the chain, wrapped in a `RawlsExceptionWithErrorReport`, so it bypasses our retry mechanism.

I'd guess this is probably why the `Option` was there -- to distinguish between a 404-failure (which just means we couldn't find your job, turns into `None`) and a retry-able failure (might succeed next time, or otherwise return a properly-failed `Future` if it's legit buggered).

rtitle: I think the retries still work as before - the `recover` partial function only gets invoked if the operation still fails after all retries are done.

W.r.t. to the response status, I think it's better behavior. :) The Rawls response now gives information about what went wrong in Google.

### Old behavior:
- Case 1
  - Google returns 404
  - This method "catches" the 404 and returns None
  - Rawls endpoint returns 404 anyway: 
```case None => RequestComplete(StatusCodes.NotFound, s"jobId $jobId not found.")```
- Case 2
  - Google returns 403 (or anything else)
  - Exception is not "caught" by this method
  - Rawls endpoint returns 500

### New behavior:
- Google returns an error
- Google status + message is wrapped in `RawlsExceptionWithErrorReport`
- Rawls API returns the Google status + message

Comment (rtitle) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala]: Crap, I understood `retryWithRecoverWhen500orGoogleError` wrong.  Looking more...

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala [hunk 4]
```diff
@@ -1804,6 +1798,15 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  def getGenomicsOperation(workspaceName: WorkspaceName, jobId: String): Future[PerRequestMessage] = {
+    dataSource.inTransaction { dataAccess =>
+      withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { workspaceContext =>
+        val genomicsServiceRef = context.actorOf(GenomicsService.props(genomicsServiceConstructor, userInfo))
+        DBIO.from((genomicsServiceRef ? GenomicsService.GetOperation(jobId)).mapTo[PerRequestMessage])
```
jmthibault79: Since this is an external call, we should move this outside the transaction.

rtitle: Makes sense.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -753,21 +751,25 @@ class HttpGoogleServicesDAO(
...
-      // So, use Google's functionality to get the json string, then parse it back into a generic json object
-      Some(executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject)
-    }) {
-      case t: HttpResponseException if t.getStatusCode == StatusCodes.NotFound.intValue => None
+    // This predicate is true when it's a 500 or Google error but NOT a Google 404.
+    //
+    // This uses the "applicative builder" syntax from cats to combine 2 predicates
+    // (when500orGoogleError and whenGoogle404) using a binary operator (e.g.
+    // (Boolean, Boolean) => Boolean), and return a resulting predicate.
+    val predicate = (when500orGoogleError |@| whenGoogle404).map((a, b) => a && !b)
```
rtitle: I hope you guys aren't too annoyed at me using cats again. :)

In this case I think it's cleaner: combining predicates using `|@|` and passing the combined predicate to the retry loop; rather than hard-coding the predicate, jamming a `recover` on the Future, and using `Option` to represent a "special" recoverable value to sort of override the predicate.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -36,11 +36,22 @@ trait GoogleUtilities extends LazyLogging with Retry {
...
 
-  protected def retryWhen500orGoogleError[T](op: () => T) = {
+  protected def whenGoogle404: Predicate[Throwable] = throwable => {
+    throwable match {
+      case t: GoogleJsonResponseException if t.getStatusCode == 404 => true
+      case _ => false
+    }
+  }
+
+  protected def retryWhen[T](pred: Predicate[Throwable])(op: () => T): Future[T] = {
```
rtitle: I feel like all these `op: () => T` (which is a Function1[Unit, T]) could be written using by-name parameters instead: `op: => T`. But I'm not sure if there is a good reason for this, so I left it as is.

rtitle: I'm also thinking of adding a spec for the `Retry` logic, since the logic tripped me up and I don't see any existing tests for it.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -753,21 +751,25 @@ class HttpGoogleServicesDAO(
...
+
+    retryWhen(predicate) { () =>
+      executeGoogleRequest(operationRequest).toPrettyString.parseJson.asJsObject
```
helgridly: You've lost the helpful comment that was above this line.

rtitle: whoops, thanks

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: I don't have a problem with the `|@|`, but I'd rename `predicate` to something a bit more descriptive.

I still think this is less clear to users. You're treating 404 as a failure when it's an expected return status for the use case. The user will get whatever Google responds with, and it'll be caught by `PerRequest.logResponse` and logged as a warning. I don't think either of these are desirable; there's a lot to be said for transforming expected errors into something user-readable.

rtitle: I'll rename `predicate`.

Not sure I understand entirely. The previous logic caught Google 404's and returned Rawls 404's (with a similar error message). See https://github.com/broadinstitute/rawls/pull/694/files#diff-16a4ec11905001338e194fb84e2fad72L43. The new behavior for the 404 case is essentially the same:
```
https://local.broadinstitute.org:20443/api/workspaces/broad-dsde-dev/rtitle-1826-test/genomics/operations/foo

{
  "statusCode": 404,
  "source": "rawls",
  "causes": [],
  "stackTrace": [],
  "message": "Job not found: \"foo\""
}
```
The difference with the new code is, if Google returns an error besides 404, before the Rawls API would return 500 with the actual Google error buried somewhere in the stack trace. Now, the Google response status and message are used in the Rawls response. The status + message are displayed to the user in a modal dialog, so I thought it would be more useful to propagate the Google error instead of just showing a generic 500.

But if you feel this is not ideal I can change it.

rtitle:
@helgridly @jmthibault79 I think this is ready for another pass.  Changes since the last review:
1. Reverted previous behavior of `HttpGoogleServicesDAO.getGenomicsOperation` w.r.t. retries and response codes
2. Run Google genomics call outside of the DB transaction
3. Added a `RetrySpec` to increase code coverage by 0.0000001% :)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1804,6 +1798,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+    // We don't need any DB information beyond that, so just return Unit on success.
+    dataSource.inTransaction { dataAccess =>
+      withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { _ =>
+        DBIO.successful(())
+      }
+    }
+    // Next call GenomicsService, which actually makes the Google Genomics API call.
+    .flatMap { _ =>
+      val genomicsServiceRef = context.actorOf(GenomicsService.props(genomicsServiceConstructor, userInfo))
+      (genomicsServiceRef ? GenomicsService.GetOperation(jobId)).mapTo[PerRequestMessage]
```
helgridly, line 1812: I had to look this up: `mapTo[PerRequestMessage]` is a less sucky version of `asInstanceOf[Future[PerRequestMessage]]`.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
```diff
@@ -0,0 +1,179 @@
...
+import akka.testkit.TestKit
+import com.typesafe.scalalogging.{LazyLogging, Logger}
+import org.mockito.ArgumentCaptor
+import org.mockito.ArgumentMatchers._
+import org.mockito.Mockito._
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.time.{Minutes, Seconds, Span}
+import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
+import org.slf4j.{Logger => Underlying}
```
helgridly: Why did you rename this? I'm confused every time I see it.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
```diff
@@ -0,0 +1,179 @@
...
+import scala.concurrent.duration._
+import scala.reflect._
+
+/**
+  * Created by rtitle on 5/16/17.
+  */
+class RetrySpec extends TestKit(ActorSystem("MySpec")) with FlatSpecLike with BeforeAndAfterAll with Matchers with MockitoSugar with ScalaFutures {
+  import system.dispatcher
+
+  implicit override val patienceConfig = PatienceConfig(timeout = scaled(Span(10, Seconds)))
```
helgridly: I'm assuming this is some Mockito magic. Could you comment it?

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
```diff
@@ -0,0 +1,179 @@
...
+    val result = testable.retry()(() => testable.failure)
+
+    // result should be a failure
+    // invocationCount should be 4 (1 initial run plus 3 retries)
+    // log count should be 4 (3 retries plus 1 failure message)
+    whenReady(result.failed) { ex =>
+      ex shouldBe an [Exception]
+      ex should have message "test exception"
+      testable.invocationCount should equal(4)
+      verify(testable.underlying, times(4)).info(anyString, any[Throwable])
```
helgridly: This is `|+|` all over again :) Could you slap a comment on here explaining what's going on? It took me a while to figure out the meaning, and I _still_ find `verify().info()` bizarre phrasing (shouldn't the thing you're testing against be inside the verify function?).

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
```diff
@@ -0,0 +1,179 @@
...
+    when(mockLogger.isInfoEnabled).thenReturn(true)
+    mockLogger
+  }
+
+  private def captor[T: ClassTag]: ArgumentCaptor[T] =
+    ArgumentCaptor.forClass(classTag[T].runtimeClass).asInstanceOf[ArgumentCaptor[T]]
+
+}
+
+class TestRetry(val system: ActorSystem, val underlying: Underlying) extends Retry with LazyLogging {
```
helgridly: Please also rename `underlying` here.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
```diff
@@ -0,0 +1,179 @@
...
+      ex should have message "test exception"
+      testable.invocationCount should equal(7)
+      verify(testable.underlying, times(7)).info(anyString, any[Throwable])
+    }
+  }
+
+  it should "retry until a success" in {
+    val testable = new TestRetry(system, setUpMockLogger)
+
+    val result = testable.retryUntilSuccessOrTimeout()(100 milliseconds, 1 day)(() => testable.failureNTimes(10))
```
helgridly: Can we make this a _bit_ less than a day? Surely a minute is enough.

helgridly (approved):
Some nitpicks and comment-requests, but go ahead and merge this after they're done. 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: Yes! Always use `mapTo` instead of `asInstanceOf` to safely cast a Future to an expected type.
(It's often needed with Akka because actors are untyped 😢 )

geoffjentry: [Not for long](http://blog.akka.io/typed/2017/05/05/typed-intro)

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
rtitle: To deal with name conflicts between `org.slf4j.Logger` and `com.typesafe.scalalogging.Logger`. `Retry` mixes in `LazyLogging` which uses `com.typesafe.scalalogging.Logger`. But it turns out `com.typesafe.scalalogging.Logger` is not mockable because it's a `final` class, so I'm mocking the underlying `org.slf4j.Logger` instead and passing it to the `com.typesafe.scalalogging.Logger` constructor:
https://github.com/typesafehub/scalalogging/blob/master/scalalogging-slf4j/src/main/scala/com/typesafe/scalalogging/slf4j/Logger.scala#L38

I agree it's confusing, maybe I'll rename the type to `SLF4JLogger` for clarity.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
rtitle: Actually not Mockito, it's ScalaTest Futures:
http://doc.scalatest.org/2.2.4/index.html#org.scalatest.concurrent.Futures

Basically it configures how long the calls to `whenReady` will wait. I'll comment it.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
rtitle: Yeah Mockito takes some getting used to. It's a pretty powerful testing tool though. `verify` takes in a mock and an expected number of times, then you can call methods on the returned mock with argument matchers to verify that the mock was called with the right arguments.

I'll add some comments. :)

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
rtitle: Sure, 1 day is excessive. Though the idea here is to _not_ hit the timeout -- `whenReady` will give up anyway after 10 seconds.

util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala
rtitle: 👍

rtitle:
Thanks! Fixing nitpicks, then will merge.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: What! How have I not seen this

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1804,6 +1798,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+  def getGenomicsOperation(workspaceName: WorkspaceName, jobId: String): Future[PerRequestMessage] = {
+    // First check the workspace context and permissions in a DB transaction.
+    // We don't need any DB information beyond that, so just return Unit on success.
+    dataSource.inTransaction { dataAccess =>
+      withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { _ =>
+        DBIO.successful(())
+      }
+    }
+    // Next call GenomicsService, which actually makes the Google Genomics API call.
+    .flatMap { _ =>
```
jmthibault79, line 1810: I usually use `andThen` in this case.  Is that just a style difference or is this way better?

rtitle: Actually `andThen` wouldn't work in this case because it returns the result of the _first_ future. Here I want to return the result of the second future, but only if the first future is successful. So I think `flatMap` is appropriate.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -36,11 +36,15 @@ trait GoogleUtilities extends LazyLogging with Retry {
     }
   }
 
-  protected def retryWhen500orGoogleError[T](op: () => T) = {
+  protected def retryWhen[T](pred: Predicate[Throwable])(op: () => T): Future[T] = {
```
jmthibault79: This doesn't seem to be referenced anywhere.  Did you intend to compose the other `retryWhen`s using this?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
geoffjentry: Well this blog post is only a week or two old, althogh they've been pushing out sequels at a good clip. I've been taking an "I'll believe it when I see it" approach as typed akka has been Coming Soon (including here and gone already) for so many years, but it looks like they're serious this time.

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
```diff
@@ -15,13 +15,15 @@ trait Retry {
   this: LazyLogging =>
   val system: ActorSystem
 
+  type Predicate[A] = A => Boolean
```
jmthibault79, line 18: I like this.  Would you mind bringing `def always` up here and defining it in terms of `Predicate`?

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
rtitle: Ah good catch, it's not being used anywhere. I was using it earlier but refactored code such that it's no longer needed. I can just remove it.

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
rtitle: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
jmthibault79: hmmmmmmm I hope I have not been using `andThen` incorrectly.

Thanks.

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
rtitle: Done

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: It's probably fine, or else it wouldn't compile :)

jmthibault79:
👍 but I think we are waiting on QA's approval for merges to dev at the moment

PR #703, opened 2017-05-22
[GAWB-769] Create a /status endpoint for Rawls

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-769
Sorry for the semi-large PR.

Features:
- `HealthMonitor` actor which periodically tests the following subsystems (period set to 1 minute), and stores the current state. Note all the status checks are asynchronous and don't depend on one another. I also defined a separate thread pool to use for status checks since it's not user facing.
  - Agora
  - Cromwell
  - DB
  - Google billing
  - Google buckets
  - Google genomics
  - Google groups
  - Google pubsub
  - LDAP
- Unauthenticated /status endpoint which asks `HealthMonitor` for the current status and returns it to the user.
- Unit tests for both of the above.

I tried to pick reasonable checks for each subsystem, but let me know if something looks off (especially for the Google ones).

Here's an example "ok" status (running locally):
```
rtitle@rawls[rt-gawb-769]$ curl "https://local.broadinstitute.org:20443/status"
{
  "ok": true,
  "systems": {
    "GooglePubSub": {
      "ok": true
    },
    "GoogleGenomics": {
      "ok": true
    },
    "LDAP": {
      "ok": true
    },
    "Database": {
      "ok": true
    },
    "Agora": {
      "ok": true
    },
    "GoogleGroups": {
      "ok": true
    },
    "GoogleBilling": {
      "ok": true
    },
    "Cromwell": {
      "ok": true
    },
    "GoogleBuckets": {
      "ok": true
    }
  }
}
```

Here's an example non-ok status (I hacked the Agora URL). Note the overall status is not-ok:
```
rtitle@rawls[rt-gawb-769]$ curl "https://local.broadinstitute.org:20443/status"
{
  "ok": false,
  "systems": {
    "GooglePubSub": {
      "ok": true
    },
    "GoogleGenomics": {
      "ok": true
    },
    "LDAP": {
      "ok": true
    },
    "Database": {
      "ok": true
    },
    "Agora": {
      "ok": false,
      "messages": ["Connection attempt to agora.fake.broadinstitute.org:443 failed"]
    },
    "GoogleGroups": {
      "ok": true
    },
    "GoogleBilling": {
      "ok": true
    },
    "Cromwell": {
      "ok": true
    },
    "GoogleBuckets": {
      "ok": true
    }
  }
}
```

  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,15 @@ submission-monitor-dispatcher {
   throughput = 1
 }
 
+health-monitor-dispatcher {
+  type = Dispatcher
+  executor = "thread-pool-executor"
+  thread-pool-executor {
+    fixed-pool-size = 8
```
rtitle, line 25: 8 threads seems reasonable, as there are 8 subsystems.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/JndiUserDirectoryDAO.scala
```diff
@@ -48,6 +48,15 @@ class JndiUserDirectoryDAO(providerUrl: String, user: String, password: String,
     members.contains(new Person(user).name)
   }
 
+  override def listUsers(implicit executionContext: ExecutionContext): Future[List[RawlsUserSubjectId]] = withContext { ctx =>
+    val name = userDnFormat.replaceAll("cn=.*?,", "")
+    val filter = s"(&${userObjectClasses.map(oc => s"(objectclass=$oc)").mkString})"
```
rtitle: LDAP craziness. Just looking at this myself, maybe it could use a comment.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -143,7 +144,8 @@ object Boot extends App with LazyLogging {
 
     val genomicsServiceConstructor: (UserInfo) => GenomicsService = GenomicsService.constructor(slickDataSource, gcsDAO, userDirDAO)
     val statisticsServiceConstructor: (UserInfo) => StatisticsService = StatisticsService.constructor(slickDataSource, gcsDAO, userDirDAO)
-    val methodRepoDAO = new HttpMethodRepoDAO(conf.getConfig("methodrepo").getString("server"))
+    val agoraConfig = conf.getConfig("methodrepo")
+    val methodRepoDAO = new HttpMethodRepoDAO(agoraConfig.getString("server"), agoraConfig.getString("path"))
```
rtitle, line 148: I split the Agora URL config into 2 components:
```
methodrepo {
  server = "https://agora.dsde-dev.broadinstitute.org:443"
  path = "/api/v1"
}
```
This is because the Agora status endpoint just lives at /status, so I need to be able to get the base URL. I think this means an update to https://github.com/broadinstitute/firecloud-develop, I'll open a PR for that shortly.

rtitle: Done https://github.com/broadinstitute/firecloud-develop/pull/456

rtitle:
@helgridly I think you're going to be the lucky reviewer on this. Let me know what you think.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -227,6 +228,16 @@ case class SubmissionWorkflowStatusResponse(
...
+case class SubsystemStatus(
+  ok: Boolean,
+  messages: Option[List[String]]
+)
+
+case class StatusCheckResponse(
+  ok: Boolean,
+  systems: Map[Subsystem, SubsystemStatus]
+)
+
```
helgridly: Why are these in `ExecutionModel`? If you put them somewhere in the `model/` folder you can reuse them in Orchestration by republishing the rawlsModel library, too.

rtitle: No particular reason. Good tip, will do.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/JndiUserDirectoryDAO.scala
helgridly: yes please!

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -238,21 +249,19 @@ class ExecutionJsonSupport extends JsonSupport {
 
   implicit override val attributeFormat = new AttributeFormat with PlainArrayAttributeListSerializer
 
-  implicit object WorkflowStatusFormat extends RootJsonFormat[WorkflowStatus] {
-    override def write(obj: WorkflowStatus): JsValue = JsString(obj.toString)
-    override def read(json: JsValue): WorkflowStatus = json match {
-      case JsString(name) => WorkflowStatuses.withName(name)
-      case x => throw new DeserializationException("invalid value: " + x)
+  def rawlsEnumerationFormat[T <: RawlsEnumeration[T]](construct: String => T): RootJsonFormat[T] = {
```
helgridly: cool, thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -0,0 +1,238 @@
...
+  val DefaultStaleThreshold = 15 minutes
+
+  val OkStatus = SubsystemStatus(true, None)
+  val UnknownStatus = SubsystemStatus(false, Some(List("Unknown status")))
+  def failedStatus(message: String) = SubsystemStatus(false, Some(List(message)))
+
+  // Actor API:
+
+  /** Triggers subsystem checking */
+  case object CheckAll
```
helgridly: We usually make the case classes/objects we use for Actor message extend a common base class per service... but having looked into it, I have no idea why. We don't use the base class anywhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -0,0 +1,238 @@
...
+      (Agora, checkAgora),
+      (Cromwell, checkCromwell),
+      (Database, checkDB),
+      (GoogleBilling, checkGoogleBilling),
+      (GoogleBuckets, checkGoogleBuckets),
+      (GoogleGenomics, checkGoogleGenomics),
+      (GoogleGroups, checkGoogleGroups),
+      (GooglePubSub, checkGooglePubsub),
+      (LDAP, checkLDAP)
+    ).foreach(processSubsystemResult _ tupled)
```
helgridly: I'm assuming from the context that `processSubsystemResult _ tupled` unpacks the tuple and passes it to the function `processSubsystemResult`, but I'm struggling to parse it in my mind. `processSubsystemResult( _.tupled )` ?

rtitle: Actually `tupled` is a method on `scala.Function2` which returns a function which takes the parameters as a tuple (instead of individual args).
See: http://www.scala-lang.org/api/2.9.2/scala/Function2.html

For example:
```
scala> def foo(a: Int, b: Int) = a + b
foo: (a: Int, b: Int)Int

scala> :t foo _
(Int, Int) => Int

scala> :t (foo _).tupled
((Int, Int)) => Int

// not relevant here, just noting :)
scala> :t (foo _).curried
Int => (Int => Int)
```

So with parens this would be written as 
```
.foreach((processSubsystemResult _).tupled)
```
Maybe that is more clear...

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
rtitle: Yeah I don't think it's necessary for Akka but I could do that to follow the convention.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: Isn't that more like `(processSubsystemResult.tupled) _`, then?

helgridly: Oh, I'm getting confused about what _ means. `processSubsystemResult _` refers to the function itself; _ isn't a stand-in for the mapped-thing here.

rtitle: Scala makes you append `_` to denote a partially applied function. So that ^ doesn't compile, but this does:
```
.foreach((processSubsystemResult _).tupled(_))
```
Note there are different meanings of `_` in that, because Scala.

I could also just scrap the whole thing and do:
```
.foreach { case (subsystem, res) => processSubsystemResult(subsystem,res) }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -0,0 +1,238 @@
...
+  // Use the execution context for this actor's dispatcher for all asynchronous operations.
+  // We define a separate execution context (a fixed thread pool) for health checking to
+  // not interfere with user facing operations.
+  import context.dispatcher
+
+  /**
+    * Contains each subsystem status along with a timestamp of when the entry was made.
+    * Initialized with unknown status.
+    */
+  var data: Map[Subsystem, (SubsystemStatus, Long)] = {
```
helgridly: gasp! a `var`!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  private def store(subsystem: Subsystem, status: SubsystemStatus): Unit = {
+    data = data + (subsystem -> (status, System.currentTimeMillis))
+    logger.debug(s"New health monitor state: $data")
+  }
+
+  private def getCurrentStatus: StatusCheckResponse = {
+    val now = System.currentTimeMillis()
+    // Convert any expired statuses to unknown
```
helgridly, line 248: When would this actually happen in practice? The scheduler re-checks minutely. If Rawls returns `Unknown`, is that because it's broken? :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
rtitle: ...but it's ok because it's akka and I'm not exposing the state to any asynchronous operations (futures). I've also seen context.become used to evolve actor state which lets you use immutable data structures, but that strikes me as overly complex and not really necessary (no strong preference though).

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
rtitle: Hm, I was thinking about the case where a subsystem check hangs and doesn't return, but I'm enforcing a timeout on the futures so the future should fail if not complete after 1 minute. I guess this _shouldn't_ happen in practice; but it might be good as a defensive check in case our status checks stop running for any reason. (Maybe the akka scheduler dies? Not sure if that can happen.)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: Why not just make `processSubsystemResult` take a tuple?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: Let's be consistent for now, and decide if we want to be terse later.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
rtitle: Yeah that sounds good :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: The FP ancestors spin in their graves!

helgridly: (it's fine)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: `Unknown` turns into an HTTP 500 anyway, so as far as the things pinging `/status` are concerned, it's still a failure. Fine by me, I guess.

helgridly:
A few comments to address but overall a 👍

helgridly:
Oh, and please use squash-and-merge when you do hit it. I think you missed it in firecloud-ui recently.

rtitle:
Thanks! Whoops sorry about firecloud-ui. Will address the comments.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
geoffjentry: You should make it private though even though technically no one should be able to get at it since it's all actorrefs. Over in cromwell we generally prefer FSM to context.become but it's all basically the same.

rtitle:
Applied code review comments.

dvoet:
I was just passing by... the asynchronous stuff is great and all, but it seems overkill to me. Why don't we just go get status fresh every time we are asked?

rtitle:
A couple reasons:

1. The /status endpoint is unauthenticated - I didn't want each call to /status to trigger DB queries, Google calls, etc. It opens us up to DDoS.
2. I want the /status endpoint to be reliable. If there _is_ a problem with the DB (for example) that is causing queries to hang and time out after a minute, I don't want to tie that behavior to /status. Instead, the call to /status will return quickly and report that there is a problem with the DB (but not other subsystems).

I think a little bit of asynchronicity is a good thing in this case because it decouples the checks themselves (which might fail) with the request for status (which should be reliable). For example https://dev.twitter.com/overview/status doesn't check its subsystems in real-time; it monitors them asynchronously and reports the results when asked.

dvoet:
excellent, can you stick those points in a comment in the code? maybe in the scaladoc for HealthMonitor?

rtitle:
Will do.

PR #709, opened 2017-06-09
[GAWB-2056] The status of the last run is showing green when there are errors in them

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-2056 (see latest comments)

### Problem: 
In "get workspace" calls, Rawls was returning the timestamps of the most recent successful and failed workflows. This is used in the UI to display either green or red in the workspace summary and list pages (red if latest failed workflow timestamp > latest successful workflow timestamp; green otherwise).

However, these workflows might be in the same submission. In the UI, a submission is red if _any_ of its workflows failed. This led to confusing behavior: a workspace might be green in the Summary tab, but have only failed submissions in the Monitor tab.

### Solution:
Change semantics of `WorkspaceSubmissionStats` to return the timestamps of the most recent successful and failed _submission_, taking into account the logic that a submission is a failure if any of its workflows are a failure.

Tested via unit tests and manually using a UI. Also confirmed the generated slick query matches the query in the [comments](https://github.com/broadinstitute/rawls/compare/rt-gawb-2056?expand=1#diff-ce257bad14de3e732d04d31146839b87R511).

Also, query explain plan: 
image

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
@@ -5,8 +5,6 @@ import java.sql.Timestamp
 import java.util.UUID
 
 import akka.util.ByteString
-import cats._
-import cats.implicits._
 import org.apache.commons.codec.binary.Base64
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
@@ -119,48 +117,6 @@ trait DriverComponent {
 
   def renameForHiding(recordCount: Long, name: String): String = {
     name + "_" + getSufficientlyRandomSuffix(recordCount)
-  }
-
-  /**
-    * Converts a `Seq[(A, B)]` into a `Map[A, B]`, combining the values with a `Monoid[B]` in case of key conflicts.
-    *
-    * For example:
-    * {{{
-    * scala> groupPairs(Seq(("a", 1), ("b", 2), ("a", 3)))
-    * res0: Map[String,Int] = Map(b -> 2, a -> 4)
-    * }}}
-    * */
-  def groupPairs[A, B: Monoid](pairs: Seq[(A, B)]): Map[A, B] =
-    pairs.toList.foldMap { case (a, b) => Map(a -> b) }
-
-  /**
-    * Converts a `Seq[(A, F[B])]` into a `Map[A, F[B]]`, combining the values with the _universal_ monoid for type F.
-    * This can be useful for when using groupBy with slick aggregate functions such as max, min, avg, etc which return
-    * an Option.
-    *
-    * For example:
-    * {{{
-    * scala> case class Foo(i: Int)
-    * defined class Foo
-    *
-    * scala> groupPairs(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
-    * << does not compile as there is no Monoid instance for Foo >>
-    *
-    * scala> groupPairsK(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
-    * res9: Map[String,Option[Foo]] = Map(b -> Some(Foo(2)), a -> Some(Foo(1)), c -> Some(Foo(3)))
-    * }}}
-    */
-  def groupPairsK[F[_], A, B](pairs: Seq[(A, F[B])])(implicit M: MonoidK[F]): Map[A, F[B]] =
-    groupPairs(pairs)(M.algebra[B])
-
-  // Same as above but with triples.
-  // Note: if we used shapeless we could generalize these functions for any arity.
-  // But I'm not going to add shapeless just for this. :)
-  def groupTriples[A, B, C: Monoid](trips: Seq[(A, B, C)]): Map[A, Map[B, C]] =
-    trips.toList.foldMap { case (a, b, c) => Map(a -> Map(b -> c)) }
-
-  def groupTriplesK[F[_], A, B, C](trips: Seq[(A, B, F[C])])(implicit M: MonoidK[F]): Map[A, Map[B, F[C]]] = {
-    groupTriples(trips)(M.algebra[C])
   }
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
@@ -121,9 +121,6 @@ trait DriverComponent {
     name + "_" + getSufficientlyRandomSuffix(recordCount)
   }
 
-  // Note: if we used shapeless we could abstract the following functions further by abstracting over the arity.
-  // But I'm not going to add shapeless just for this. :)
-
   /**
     * Converts a `Seq[(A, B)]` into a `Map[A, B]`, combining the values with a `Monoid[B]` in case of key conflicts.
     *
@@ -145,14 +142,20 @@ trait DriverComponent {
     * {{{
     * scala> case class Foo(i: Int)
     * defined class Foo
+    *
+    * scala> groupPairs(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
+    * << does not compile as there is no Monoid instance for Foo >>
+    *
     * scala> groupPairsK(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
     * res9: Map[String,Option[Foo]] = Map(b -> Some(Foo(2)), a -> Some(Foo(1)), c -> Some(Foo(3)))
     * }}}
-    *
     */
   def groupPairsK[F[_], A, B](pairs: Seq[(A, F[B])])(implicit M: MonoidK[F]): Map[A, F[B]] =
     groupPairs(pairs)(M.algebra[B])
 
+  // Same as above but with triples.
+  // Note: if we used shapeless we could generalize these functions for any arity.
+  // But I'm not going to add shapeless just for this. :)
   def groupTriples[A, B, C: Monoid](trips: Seq[(A, B, C)]): Map[A, Map[B, C]] =
     trips.toList.foldMap { case (a, b, c) => Map(a -> Map(b -> c)) }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
@@ -4,14 +4,15 @@ import java.nio.ByteOrder
 import java.sql.Timestamp
 import java.util.UUID
 
-import akka.util.{ByteString, ByteStringBuilder}
+import akka.util.ByteString
+import cats._
+import cats.implicits._
+import org.apache.commons.codec.binary.Base64
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
-import org.joda.time.DateTime
 import slick.driver.JdbcDriver
 import slick.jdbc.{GetResult, PositionedParameters, SQLActionBuilder, SetParameter}
 import spray.http.StatusCodes
-import org.apache.commons.codec.binary.Base64
 
 import scala.concurrent.ExecutionContext
 
@@ -118,6 +119,45 @@ trait DriverComponent {
 
   def renameForHiding(recordCount: Long, name: String): String = {
     name + "_" + getSufficientlyRandomSuffix(recordCount)
+  }
+
+  // Note: if we used shapeless we could abstract the following functions further by abstracting over the arity.
+  // But I'm not going to add shapeless just for this. :)
+
+  /**
+    * Converts a `Seq[(A, B)]` into a `Map[A, B]`, combining the values with a `Monoid[B]` in case of key conflicts.
+    *
+    * For example:
+    * {{{
+    * scala> groupPairs(Seq(("a", 1), ("b", 2), ("a", 3)))
+    * res0: Map[String,Int] = Map(b -> 2, a -> 4)
+    * }}}
+    * */
+  def groupPairs[A, B: Monoid](pairs: Seq[(A, B)]): Map[A, B] =
+    pairs.toList.foldMap { case (a, b) => Map(a -> b) }
+
+  /**
+    * Converts a `Seq[(A, F[B])]` into a `Map[A, F[B]]`, combining the values with the _universal_ monoid for type F.
+    * This can be useful for when using groupBy with slick aggregate functions such as max, min, avg, etc which return
+    * an Option.
+    *
+    * For example:
+    * {{{
+    * scala> case class Foo(i: Int)
+    * defined class Foo
+    * scala> groupPairsK(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
+    * res9: Map[String,Option[Foo]] = Map(b -> Some(Foo(2)), a -> Some(Foo(1)), c -> Some(Foo(3)))
+    * }}}
+    *
+    */
+  def groupPairsK[F[_], A, B](pairs: Seq[(A, F[B])])(implicit M: MonoidK[F]): Map[A, F[B]] =
+    groupPairs(pairs)(M.algebra[B])
+
+  def groupTriples[A, B, C: Monoid](trips: Seq[(A, B, C)]): Map[A, Map[B, C]] =
+    trips.toList.foldMap { case (a, b, c) => Map(a -> Map(b -> c)) }
+
+  def groupTriplesK[F[_], A, B, C](trips: Seq[(A, B, F[C])])(implicit M: MonoidK[F]): Map[A, Map[B, F[C]]] = {
+    groupTriples(trips)(M.algebra[C])
   }
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import java.sql.Timestamp
 import java.util.{Date, UUID}
 
+import cats.{Monoid, MonoidK}
 import cats.instances.int._
 import cats.instances.option._
 import org.broadinstitute.dsde.rawls.RawlsException
@@ -10,6 +11,7 @@ import org.broadinstitute.dsde.rawls.dataaccess.SlickWorkspaceContext
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels.WorkspaceAccessLevel
 import org.broadinstitute.dsde.rawls.model._
+import org.broadinstitute.dsde.rawls.util.CollectionUtils
 import org.joda.time.DateTime
 /**
  * Created by dvoet on 2/4/16.
@@ -545,8 +547,8 @@ trait WorkspaceComponent {
         submissionDates <- filteredSubmissionMaxDateQuery.result
         runningSubmissions <- runningSubmissionsQuery.result
       } yield {
-        val submissionDatesByWorkspaceByStatus = groupTriplesK(submissionDates)
-        val runningSubmissionCountByWorkspace = groupPairs(runningSubmissions)
+        val submissionDatesByWorkspaceByStatus: Map[UUID, Map[String, Option[Timestamp]]] = groupByWorkspaceIdThenStatus(submissionDates)
+        val runningSubmissionCountByWorkspace: Map[UUID, Int] = groupByWorkspaceId(runningSubmissions)
 
         workspaceIds.map { wsId =>
           val (lastFailedDate, lastSuccessDate) = submissionDatesByWorkspaceByStatus.get(wsId) match {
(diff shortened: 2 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -731,14 +759,6 @@ trait WorkspaceComponent {
       WorkspaceGroups(toGroupMap(realmAclRecs), toGroupMap(accessGroupRecs))
     }
   }
-
-  private def groupByWorkspaceId(runningSubmissions: Seq[(UUID, Int)]): Map[UUID, Int] = {
```
rtitle: These methods seemed useful so I cats-ified and generalized them and moved to `DriverComponent`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
```diff
@@ -119,6 +120,48 @@ trait DriverComponent {
...
+    *
+    * scala> groupPairs(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
+    * << does not compile as there is no Monoid instance for Foo >>
+    *
+    * scala> groupPairsK(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("c", Foo(3).some)))
+    * res9: Map[String,Option[Foo]] = Map(b -> Some(Foo(2)), a -> Some(Foo(1)), c -> Some(Foo(3)))
+    * }}}
+    */
+  def groupPairsK[F[_], A, B](pairs: Seq[(A, F[B])])(implicit M: MonoidK[F]): Map[A, F[B]] =
+    groupPairs(pairs)(M.algebra[B])
```
helgridly: something tells me it'll be a while before this PR gets merged! (i have to wrap my head around this)

rtitle: Yeah sorry if this got a little fancy.. it's not really too bad. Check out:
https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/semigroupk.md

helgridly: A bunch of feedback here.

0. Check out `CollectionUtils` - seems like a good place for these to go.
1. I don't understand this.
2. Link to the [HTML version](http://typelevel.org/cats/typeclasses/semigroupk.html) of the documentation: it contains hugely important comment lines that aren't rendered inside GitHub!
3. sbt console gives me this when I try to define groupPairsK:

```
<pastie>:18: warning: higher-kinded type should be enabled
by making the implicit value scala.language.higherKinds visible.
This can be achieved by adding the import clause 'import scala.language.higherKinds'
or by setting the compiler option -language:higherKinds.
See the Scaladoc for value scala.language.higherKinds for a discussion
why the feature should be explicitly enabled.
```

4. `groupPairs` returns a map. You're then dereferencing the key defined by `M.algebra[B]`? I don't know what that does and I'd have thought that this would give you a result type of `F[B]`.
5. Your `groupPairsK` example is equivalent to `toMap`, which doesn't help me understand. Perhaps

```
scala> groupPairsK(Seq(("a", Foo(1).some), ("b", Foo(2).some), ("b", Foo(3).some)))
res0: Map[String,Option[Foo]] = Map(b -> Some(Foo(2)), a -> Some(Foo(1)))
```

would be more helpful, assuming I've read the Cats documentation (and therefore know that in the absence of a quantified `F` (= `Option`) means it'll just do `orElse`).

TLDR: I am :+1: on `groupPairs` and `groupTriples`, though they should go into `CollectionUtils`. I am :-1: on the K-versions; they rely on non-obvious behaviour (e.g. `orElse` for `Option`) that make them hard to understand.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -499,36 +500,56 @@ trait WorkspaceComponent {
...
+        (workspaceId, status)
+      }.map { case ((workspaceId, status), recs) =>
+        (workspaceId, status, recs.map(_._3).max, recs.length)
+      }
+
+      // Note: a submission is successful if it contains _only_ successful workflows.
+      // A submission is a failure if it contains _any_ failed workflows.
+      val filteredSubmissionMaxDateQuery = submissionMaxDateQuery.filter { case (_, status, _, count) =>
+        status === WorkflowStatuses.Failed.toString || (status === WorkflowStatuses.Succeeded.toString && count === 1)
+      }.map { case (workspaceId, status, max, _) => (workspaceId, status, max)}
```
helgridly: I haven't gone through this yet.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -499,36 +500,56 @@ trait WorkspaceComponent {
...
 
       for {
-        workflowDates <- workflowDatesGroupedQuery.result
+        submissionDates <- filteredSubmissionMaxDateQuery.result
         runningSubmissions <- runningSubmissionsQuery.result
       } yield {
-        val workflowDatesByWorkspaceByStatus: Map[UUID, Map[String, Option[Timestamp]]] = groupByWorkspaceIdThenStatus(workflowDates)
-        val runningSubmissionCountByWorkspace: Map[UUID, Int] = groupByWorkspaceId(runningSubmissions)
+        val submissionDatesByWorkspaceByStatus = groupTriplesK(submissionDates)
+        val runningSubmissionCountByWorkspace = groupPairs(runningSubmissions)
```
helgridly: I think the rename from `groupByWorkspaceIdThenStatus` -> `groupTriplesK` and `groupByWorkspaceId` -> `groupPairs` loses a lot of readability here. Can you keep the function definitions for readability, even if they just point to other ones?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
rtitle: Thanks for the feedback here. After reflecting a bit, maybe it would have been better to leave `groupByWorkspaceIdThenStatus` and `groupByWorkspaceId` as is -- I didn't really need to change them for this bug. I'm still learning how to scala in a large team, and readability is important.

Digging in to your points:

0. +1
1. see below
2. ok
3. Huh, I haven't seen that warning before. I had to enable the `-feature` compiler flag to see it. I guess it's complaining about the `F[_]` type parameter. The [scaladoc](http://www.scala-lang.org/api/2.10.3/#scala.language$) is kind of funny:
> Why control it? Higher kinded types in Scala lead to a Turing-complete type system, where compiler termination is no longer guaranteed. They tend to be useful mostly for type-level computation and for highly generic design patterns. The level of abstraction implied by these design patterns is often a barrier to understanding for newcomers to a Scala codebase. Some syntactic aspects of higher-kinded types are hard to understand for the uninitiated and type inference is less effective for them than for normal types. Because we are not completely happy with them yet, it is possible that some aspects of higher-kinded types will change in future versions of Scala. So an explicit enabling also serves as a warning that code involving higher-kinded types might have to be slightly revised in the future.

4,5. So here's why I added the K version. In this case I wanted to group triples where the map value is an `Option[java.sql.Timestamp]`. Therefore I need a Monoid instance for `java.sql.Timestamp`, which doesn't exist. It wouldn't make much sense to define a Monoid for timestamps -- what would you do, add them together? However, in this case I'm guaranteed to have no key conflicts because of the SQL structure (`group by`, etc). So I can just use a Monoid for Option which just takes one or the other, regardless of the value inside the "box". That's exactly what `MonoidK[Option]` does.

Furthermore, I'm not sure how I would implement this in terms of `groupTriples` without using `MonoidK`. Perhaps this would be more clear?
```
private def groupByWorkspaceIdThenStatus(workflowDates: Seq[(UUID, String, Option[Timestamp])]): Map[UUID, Map[String, Option[Timestamp]]] = {
  // bla bla comment about bringing a monoid into scope with Option.orElse behavior
  implicit val optionUniversalMonoid: Monoid[Option[Timestamp]] = MonoidK[Option].algebra[Timestamp]
  CollectionUtils.groupTriples(workflowDates)
}
```
Then at least the `MonoidK` stuff is localized in a private method with some explanation specific to the use case.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
rtitle: Sure

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
rtitle: I went ahead and pushed a commit with ^ those changes, please take a look.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/CollectionUtils.scala
```diff
@@ -10,4 +15,20 @@ object CollectionUtils {
...
+    * Converts a `Seq[(A, B)]` into a `Map[A, B]`, combining the values with a `Monoid[B]` in case of key conflicts.
+    *
+    * For example:
+    * {{{
+    * scala> groupPairs(Seq(("a", 1), ("b", 2), ("a", 3)))
+    * res0: Map[String,Int] = Map(b -> 2, a -> 4)
+    * }}}
+    * */
+  def groupPairs[A, B: Monoid](pairs: List[(A, B)]): Map[A, B] =
+    pairs.foldMap { case (a, b) => Map(a -> b) }
```
helgridly: `groupByTuplesFlatten` above could call this, right?

helgridly: also these should probably prefer `Seq` to `List`

rtitle: Re Seq: sure

I guess this would work:
```
  //A saner group by than Scala's.
  def groupByTuples[A, B]( tupleSeq: Seq[(A,B)] ): Map[A, Seq[B]] = {
    tupleSeq.toList.foldMap { case (a, b) => Map(a -> Seq(b)) }
  }

  def groupByTuplesFlatten[A, B]( tupleSeq: Seq[(A, Seq[B])] ): Map[A, Seq[B]] = {
    groupPairs(tupleSeq)
  }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -733,11 +756,22 @@ trait WorkspaceComponent {
...
-    workflowDates.groupBy { case (wsId, _, _) => wsId }.mapValues(_.groupBy { case (_, status, _) => status }.mapValues { case Seq((_, _, timestamp)) => timestamp })
+    // There is no Monoid instance for Option[Timestamp] so we need to bring one into scope.
+    // However a Monoid for Timestamp doesn't really make sense -- what would it do, add them together?
+    // We can take advantage of the _universal_ monoid for Option which combines Option values using
+    // Option.orElse. It's called universal because it works no matter the type inside the Option.
+    // This is fine in this case because there are guaranteed no key conflicts due to the SQL query
+    // structure (group by, etc).
+    //
+    // TL/DR: The following line brings into scope a Monoid[Option[Timestamp]] which combines values
+    // using Option.orElse.
```
helgridly: I think part of my confusion here is that you're introducing Monoid, which is used to combine things, using a non-combining operation (`orElse`), in a situation where you never need to combine two of them anyway because the UUID/String combo is guaranteed to be unique. This is a pretty head-bendy way to achieve the desired result, even if it does work.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/CollectionUtils.scala
geoffjentry: Why would it prefer `Seq`? Do you expect it to be taking arbitrary `Seq`s?

helgridly: Seq is less specific and we use it all over the place. We'd have to start jamming `.toList` everywhere were we to use List.

rtitle: Er I'd still need to do some MonoidK magic to get a monoid for B.

rtitle: Unfortunately cats doesn't provide instances for Seq :(
https://github.com/typelevel/cats/issues/1222

geoffjentry: Maybe people shouldn't have overused `Seq` and then that wouldn't be an issue 

We have the same problem too and it's annoying. We occasionally run into issues where the thing really only works properly on List but things compile due to Seq and someone jammed something bad in there.

helgridly: I meant just `groupByTuplesFlatten` anyway, but oh well

geoffjentry: @rtitle Cuz the Cats folk understand the value in saying what you mean ;)

helgridly: Thanks for your input, Jeff! When we wholesale switch from Seq to List I'll let you know so you can say I told you so.

rtitle: I'll change these to take `Seq` (and just call `toList` on them).

I won't touch groupByTuples/groupByTuplesFlatten because that would require introducing monoids and I don't want to break any code.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
rtitle: At least the `orElse` is explicit now. Before, in this code:
```
...mapValues { case Seq((_, _, timestamp)) => timestamp })
```
it's still expecting a unique UUID/String combo, and if that were not the case, there would be a runtime `MatchError`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -499,36 +502,56 @@ trait WorkspaceComponent {
...
+      // submission date query:
+      //
+      // select workspaceId, status, max(submissionDate)
+      // from (
+      //   select distinct submission.workspaceId, workflow.status, submission.submissionDate
+      //   from submission
+      //   join workflow on workflow.submissionId = submission.id
+      //   where submission.workspaceId in (:workspaceIds)) v
+      // group by 1, 2
+      // having (status = 'Failure' or (status = 'Succeeded' and count(v.*) = 1))
```
rtitle: @helgridly based on our conversation how does this revised query look to you?
```
      select workspaceId, status, max(subEndDate)
      from (
        select submission.id, submission.workspaceId, workflow.status, max(workflow.statusLastChangedDate) as subEndDate
        from submission
        join workflow on workflow.submissionId = submission.id
        where submission.workspaceId in (:workspaceIds)
        group by 1, 2, 3) v
      group by 1, 2
      having (status = 'Failure' or (status = 'Succeeded' and count(v.id) = 1))
```
Explanation: 
- inner query returns the most recent workflow status change date, per workflow status and submission
- outer query returns the most recent workflow status change date where:
  - the status is Failure; or
  - the status is Succeeded and that is the _only_ status in the submission

I can code this up and try to break it with tests too, just thought I'd post the SQL first.

helgridly: I think that's right, though SQL isn't my native language!

rtitle: Actually after some testing I think it's not right. SQL is not my native language either (or maybe it's my slick mapping). :( Still working through it, expect another iteration...

rtitle:
@helgridly I think this is ready for another look, thanks

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -499,36 +502,70 @@ trait WorkspaceComponent {
...
+          workflows <- workflowQuery if submissions.id === workflows.submissionId
+        } yield (submissions.id, submissions.workspaceId, workflows.status, workflows.statusLastChangedDate)
+      ).filter { case (_, _, status, _) =>
+        status === WorkflowStatuses.Failed.toString || status === WorkflowStatuses.Succeeded.toString
+      }.map { case (submissionId, workspaceId, status, statusLastChangedDate) =>
+        (submissionId, workspaceId, Case If (status === WorkflowStatuses.Failed.toString) Then 1 Else 0, statusLastChangedDate)
+      }.groupBy { case (submissionId, workspaceId, _, _) =>
+        (submissionId, workspaceId)
+      }.map { case ((submissionId, workspaceId), recs) =>
+        (submissionId, workspaceId, recs.map(_._3).max, recs.map(_._4).max)
```
helgridly: The third element is supposed to be a count, right? `recs.length`?

rtitle: Actually I meant `max`. `sum` would work as well. This is here so the value will be >0 if the submission contains any failed workflows, and it will be 0 if it contains all successful workflows.

`count` would _not_ work since it would just count all the rows, regardless of the workflow status (in SQL count(0) == count(1) == count(*) since 0 and 1 are both non-null values).

helgridly: Yeah, I meant `sum` -- sorry. This jumped out at me because the `outerSubmissionDateQuery` line calls this value `numFailures`. If it's just a 1/0 then maybe just call `exists` here so the later line can be `hasFailures`.

rtitle: Right, +1

MatthewBemis:
Could you explain the explain plan diagram? :p 

Looking at it I see the words "full table scan" twice which makes me think maybe we're missing an index somewhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
rtitle: I'm not sure `.exists` would work in this context -- I just changed it to `.count` to better reflect the variable name/type.

rtitle:
Sure:
1. First, the diagram in the description is out of date since I changed the query - below is the current one.
2. The bottom "Full table scan" was just due to how I was testing it:
```
// causes full table scan since no index on workspace.name
... where s.workspace_id in (select id from workspace where name = 'my_test_workspace') ...
```
The below diagram is how it is executed in code:
```
// no full table scan
... where s.workspace_id in (:workspaceIds) ... 
```
image

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
helgridly: I still think this is less readable than doing it explicitly but I'm going on vacation tomorrow and thus can't afford to spend more time arguing about it :) Instead, if you replace your comment with the following one, I'll shut up and thumb this. In the long-term, we need a centralised place for these explanations, because throwing them in comments on their first use won't help new hires understand them if they happen upon the uncommented second or third use first. But we can worry about that (a little) later.

---

The function groupTriples, called below, transforms a Seq((T1, T2, T3)) to a Map(T1 -> Map(T2 -> T3)). It does this by calling foldMap, which in turn requires a monoid for T3. In our case, T3 is an Option[Timestamp], so we need to provide an implicit monoid for Option[Timestamp].

There isn't really a sane monoid implementation for Timestamp (what would you do, add them?). Thankfully it turns out that the UUID/String pairs in workflowDates are always unique, so it doesn't matter what the monoid does because it'll never be used to combine two Option[Timestamp]s. It just needs to be provided in order to make the compiler happy.

To do this, we use the universal monoid for Option, MonoidK[Option]. Note that the inner Option takes no type parameter: MonoidK doesn't care about the type inside Option, it just calls orElse on the Option for its "combine" operator. Finally, the call to algebra[Timestamp] turns a MonoidK[Option] into a Monoid[Option[Timestamp]] by leaving the monoid implementation alone (so it still calls orElse) and poking the Timestamp type into the Option.
```

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala [hunk 9]
```diff
@@ -733,11 +770,22 @@ trait WorkspaceComponent {
...
+    // We can take advantage of the _universal_ monoid for Option which combines Option values using
+    // Option.orElse. It's called universal because it works no matter the type inside the Option.
+    // This is fine in this case because there are guaranteed no key conflicts due to the SQL query
+    // structure (group by, etc).
+    //
+    // TL/DR: The following line brings into scope a Monoid[Option[Timestamp]] which combines values
+    // using Option.orElse.
+
+    implicit val optionTimestampMonoid: Monoid[Option[Timestamp]] = MonoidK[Option].algebra[Timestamp]
+    CollectionUtils.groupTriples(workflowDates.toList)
```
helgridly: you don't need the `toList` here any more, now you've swapped them to use Seq

helgridly:
👍 pending two final nitpicks

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
rtitle: ok :)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
rtitle: thanks

PR #718, opened 2017-06-26
[GAWB-2243] Database deadlock on creating submission using Mutect pipeline

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-2243

There are 2 main fixes here:
1. In `WorkflowSubmissionActor`, don't process workflows whose submission has status `Aborting`.
2. In `SubmissionMonitorActor`, when setting workflow statuses to `Aborted` for a submission, do it in batches of 10 instead of updating _all_ workflows for a submission in a single query.

This seems to fix the problem in my test environment -- before the fix I can consistently reproduce the deadlock in `SubmissionApiServiceSpec` by creating and aborting large submissions; with the fixes the deadlock is gone.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -33,5 +33,4 @@ gcs {
 
 submissionmonitor {
   submissionPollInterval = 1m
-  workflowAbortBatchSize = 10
 }
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -30,3 +30,8 @@ health-monitor-dispatcher {
 gcs {
   bucketLogsMaxAge = "180"
 }
+
+submissionmonitor {
+  submissionPollInterval = 1m
+  workflowAbortBatchSize = 10
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -430,16 +430,17 @@ trait WorkflowComponent {
       val query = if (excludedSubmitters.isEmpty && excludedSubmissionStatuses.isEmpty) {
         queuedWorkflows
       } else {
-        val filteredSubmissions = if (excludedSubmitters.isEmpty) {
-          submissionQuery.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
-        } else if (excludedSubmissionStatuses.isEmpty) {
+        val excludedSubmittersQuery = if (excludedSubmitters.nonEmpty) {
           submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters))
-        } else {
-          submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters)).filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
-        }
+        } else submissionQuery
+
+        val filteredSubmissionsQuery = if (excludedSubmissionStatuses.nonEmpty) {
+          excludedSubmittersQuery.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
+        } else excludedSubmittersQuery
+
         for {
           workflows <- queuedWorkflows
-          submission <- filteredSubmissions if submission.id === workflows.submissionId
+          submission <- filteredSubmissionsQuery if submission.id === workflows.submissionId
         } yield workflows
       }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -3,15 +3,18 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import java.sql.Timestamp
 import java.util.UUID
 
+import cats.instances.int._
+import cats.instances.list._
+import cats.instances.map._
+import cats.syntax.foldable._
 import org.broadinstitute.dsde.rawls.RawlsException
-import org.broadinstitute.dsde.rawls.dataaccess.{ExecutionServiceCluster, ExecutionServiceId, SlickWorkspaceContext}
-import org.broadinstitute.dsde.rawls.model._
+import org.broadinstitute.dsde.rawls.dataaccess.{ExecutionServiceId, SlickWorkspaceContext}
+import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
+import org.broadinstitute.dsde.rawls.model._
 import org.joda.time.DateTime
-import slick.dbio.Effect.{Read, Write}
+import slick.dbio.Effect.Write
 import slick.driver.JdbcDriver
-import cats._
-import cats.implicits._
 
 /**
  * Created by mbemis on 2/18/16.
@@ -328,7 +331,7 @@ trait WorkflowComponent {
       getFirstQueuedWorkflow(userInfo.userSubjectId.value) flatMap { optRec =>
         val query = optRec match {
           case Some(workflow) => findWorkflowsQueuedBefore(workflow.statusLastChangedDate)
-          case _ => findQueuedWorkflows(Seq.empty)
+          case _ => findQueuedWorkflows(Seq.empty, Seq.empty)
         }
         query.length.result
       }
@@ -422,14 +425,14 @@ trait WorkflowComponent {
       } yield workflow
     }
 
-    def findQueuedWorkflows(excludedSubmitters: Seq[String]): WorkflowQueryType = {
+    def findQueuedWorkflows(excludedSubmitters: Seq[String], excludedSubmissionStatuses: Seq[SubmissionStatus]): WorkflowQueryType = {
       val queuedWorkflows = filter(_.status === WorkflowStatuses.Queued.toString)
-      val filteredSubmissions = (
-        if (excludedSubmitters.nonEmpty) {
-          submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters))
-        } else submissionQuery
-      ).filterNot(_.status === "Aborting")
-
+      val filteredSubmissions = (excludedSubmitters.nonEmpty, excludedSubmissionStatuses.nonEmpty) match {
+        case (false, false) => submissionQuery
+        case (true, false) => submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters))
+        case (false, true) => submissionQuery.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
+        case _ => submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters)).filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
+      }
       val query =
         for {
           workflows <- queuedWorkflows
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -424,14 +424,18 @@ trait WorkflowComponent {
 
     def findQueuedWorkflows(excludedSubmitters: Seq[String]): WorkflowQueryType = {
       val queuedWorkflows = filter(_.status === WorkflowStatuses.Queued.toString)
-      val query = if (excludedSubmitters.isEmpty) {
-        queuedWorkflows
-      } else {
+      val filteredSubmissions = (
+        if (excludedSubmitters.nonEmpty) {
+          submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters))
+        } else submissionQuery
+      ).filterNot(_.status === "Aborting")
+
+      val query =
         for {
           workflows <- queuedWorkflows
-          submission <- submissionQuery if workflows.submissionId === submission.id && (!submission.submitterId.inSetBind(excludedSubmitters))
+          submission <- filteredSubmissions if submission.id === workflows.submissionId
         } yield workflows
-      }
+
       query.sortBy(_.id)
     }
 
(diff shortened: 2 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -3,15 +3,18 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import java.sql.Timestamp
 import java.util.UUID
 
+import cats.instances.int._
```
rtitle, line 6: Just cleaning up imports; there's actually no new cats stuff in here.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -422,16 +425,24 @@ trait WorkflowComponent {
...
+      val query = if (excludedSubmitters.isEmpty && excludedSubmissionStatuses.isEmpty) {
         queuedWorkflows
       } else {
+        val filteredSubmissions = if (excludedSubmitters.isEmpty) {
+          submissionQuery.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
+        } else if (excludedSubmissionStatuses.isEmpty) {
+          submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters))
+        } else {
+          submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters)).filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString)))
+        }
```
helgridly: This is clearer IMO:

```
val excludedSubmitters = if (excludedSubmitters.nonEmpty) { submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters)) } else { submissionQuery }
val filteredSubmissions = if (excludedSubmissionStatuses.nonEmpty) { submissionQuery.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString))) } else { excludedSubmitters }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -110,7 +110,14 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging {
 
     def abortQueuedWorkflows(submissionId: UUID) = {
       datasource.inTransaction { dataAccess =>
-        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted)
+        // Update workflows in batches of 10 to limit database deadlocks between
+        // this actor and WorkflowSubmissionActor. See GAWB-2243.
+        dataAccess.workflowQuery.listWorkflowRecsForSubmissionAndStatuses(submissionId, WorkflowStatuses.Queued).flatMap { workflows =>
+          val grouped = workflows.grouped(10).map { wfs =>
```
helgridly: config-ify this please

helgridly: (its default should live in reference.conf, iirc)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -131,7 +131,7 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
       for {
         // we exclude workflows submitted by users that have exceeded the max workflows per user
         excludedUsers <- dataAccess.workflowQuery.listSubmittersWithMoreWorkflowsThan(maxActiveWorkflowsPerUser, WorkflowStatuses.runningStatuses)
-        workflowRecs <- dataAccess.workflowQuery.findQueuedWorkflows(excludedUsers.map { case (submitter, count) => submitter }).take(batchSize).result
+        workflowRecs <- dataAccess.workflowQuery.findQueuedWorkflows(excludedUsers.map { case (submitter, count) => submitter }, Seq(SubmissionStatuses.Aborting)).take(batchSize).result
```
helgridly: `SubmissionStatuses.terminalStatuses + SubmissionStatuses.Aborting` ?

jmthibault79: Are there other Submission statuses we should exclude, e.g. `Aborted` or `Failed` ?

helgridly (changes requested):
Couple of minor comments in the code, plus a philosophical question to follow.

How did you test this? In a FiaB? Locally, running a docker MySQL? I suppose there's no way to tell for sure without running it on a real-ish environment.

Batching feels like treating the symptom but not the disease. It's a pragmatic solution, but did you see this? https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks.html Plus the following three sections might have some advice on how to avoid this and/or why it was kneecapping the db in the first place. I'd like to at least understand what's going on in the guts of innodb before turning down the batch size and crossing fingers. Bernick knows a lot about the vagaries of MySQL, if you haven't spoken to him already.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
```diff
@@ -142,6 +225,62 @@ class SubmissionApiServiceSpec extends ApiServiceSpec {
...
+    val submission = createAndMonitorSubmission(wsName, methodConf, testData.largeSset, Option("this.hasSamples"), services)
+
+    assertResult(10000) {
+      submission.workflows.size
+    }
+
+    abortSubmission(services, wsName, submission.submissionId)
+  }
+
+  // Ignored as this test is prone to intermittent timeout-related failures.
```
jmthibault79: Should this still be ignored?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
rtitle: Yeah maybe `SubmissionStatuses.terminalStatuses + SubmissionStatuses.Aborting` as @helgridly suggested below.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
rtitle: Probably would need to be (note difference in second line):
```
val excludedSubmitters = if (excludedSubmitters.nonEmpty) { submissionQuery.filterNot(_.submitterId.inSetBind(excludedSubmitters)) } else { submissionQuery }
val filteredSubmissions = if (excludedSubmissionStatuses.nonEmpty) { excludedSubmitters.filterNot(_.status.inSetBind(excludedSubmissionStatuses.map(_.toString))) } else { excludedSubmitters }
```
But yeah, I can make that change.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: Ok

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
rtitle: +1

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
rtitle: Hm... I'll check again. I was seeing timeout errors (the test API timeout is set to 5 sec) due to the sheer number of workflows, but there are no database deadlocks. Maybe I could increase the timeout for this test only to get it to pass consistently.

rtitle:
@helgridly I tested locally using a docker MySQL, while logged into the CLI to watch for deadlocks (I thought my laptop was going to melt). I agree we don't know _for sure_ this is what happened in prod, but the symptoms/failure modes seem to match. It would be good to test this in a more real-ish environment as well.

Thanks for the link - I did see it but I'll dig more into the innards of innodb (InnardsDB?) to see if anything else can be done.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: I'm wondering if it should go in `rawls.conf`, alongside:
```
  batchSize = 10 // used by WorkflowSubmissionActor
  processInterval = 10ms // used by WorkflowSubmissionActor
  pollInterval = 1m // used by SubmissionMonitorActor
  parallelSubmitters = 10 // used by WorkflowSubmissionActor
  maxActiveWorkflowsPerServer = 10000 // used by WorkflowSubmissionActor
```
?

Is there a rule of thumb about when to add things to rawls.conf vs reference.conf? Does rawls.conf have things that might change between environments, while reference.conf is relatively static?

rtitle: Er actually all the above are used by `WorkflowSubmissionActor`. SubmissionSupervisor takes a `pollInterval` parameter but it's not in a config file:

https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L24
https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala#L148-L151

So... maybe I'll add a new config section to reference.conf containing `submissionPollInterval` and `workflowAbortBatchSize`, and pass both to SubmissionSupervisor/MonitorActor.

dvoet: I don't think we have a hard policy here. Everything in rawls.conf can override anything in ref.conf but that is a relatively recent fact which means almost all of our config is in rawls.conf. But now it does not really matter anymore. I would be happy if we had a move to put all relatively static things in ref.conf and remove them from rawls.conf.

helgridly: IIRC ref.conf is for things we don't expect users to change, and Rawls.conf is for things we do. This is reaching back into the mists of my mind somewhat, though.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -110,7 +110,14 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging {
...
+            dataAccess.workflowQuery.batchUpdateStatus(wfs, WorkflowStatuses.Aborted)
+          }.toSeq
+          DBIO.fold(grouped, 0)(_ + _)
+        }
```
dvoet: I am curious why this would have any effect since it all happens in the same transaction anyway. Perhaps because a large update causes mysql to try and get too big a lock?

core/src/main/resources/reference.conf
```diff
@@ -29,4 +29,9 @@ health-monitor-dispatcher {
 
 gcs {
   bucketLogsMaxAge = "180"
+}
+
+submissionmonitor {
+  submissionPollInterval = 1m
```
ansingh7115, line 35: Just curious - why did you set this submissionPollInterval to 1m? Is it arbitrary? Or is there a reasoning behind it?

helgridly: Same as the [workflow poll interval](https://github.com/broadinstitute/firecloud-develop/blob/dev/configs/rawls/rawls.conf.ctmpl#L53).

rtitle: That's what it was using before, it was just set via a default parameter in code rather than a config file:
https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L24

ansingh7115: Ah, thanks

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: Ooh, I might have an explanation. Before, the deadlock was between

transaction 1:
```
update WORKFLOW set status = 'Aborted', ... where status = 'Queued' and submission_id = ...
```
transaction 2:
```
update WORKFLOW set status = 'Launching', ... where id = ...
```
So T1 was doing a full scan on the `idx_workflow_status` index, locking any rows it finds. Then T2 comes in and locks the primary key index for a row already processed by the scan in T1, and boom, deadlock.

Now, the new query is a little different. T1 now looks like:
```
select id from WORKFLOW where status = 'Queued' and submissionId = ...
update WORKFLOW set status = 'Aborted' where id in (<ids returned from query above>)
```
The select is no longer locking the `idx_workflow_status` index, so there is no more deadlock (it would be locked if it were a "select for update"). So it's not really the batching that is fixing it, it's the different query structure. To test this, I re-ran my test with batchSize = Integer.MAX_VALUE (effectively disabling batching), and the deadlock still does not occur.

References:
https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks.html
https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks-handling.html
https://www.xaprb.com/blog/2006/08/03/a-little-known-way-to-cause-a-database-deadlock/

I'm still reading, not sure if the best solution yet, just posting my findings along the way.

rtitle:
@helgridly after thinking about this more, I reverted the "batching" change from this PR. It wasn't really needed to prevent the deadlock, and it wasn't efficient. So the only change here is #1 above: don't submit workflows whose submission is Aborting/Aborted/Done.

Based on testing this actually prevents the deadlock by itself. And the change makes sense anyway.

I also un-ignored the deadlock unit test - it passes after increasing the timeout.

Can you take another look?

helgridly (approved):

helgridly:
This is great -- thank you for the deeper look. Makes a lot of sense not to try submitting workflows that are going to be aborted anyway. 👍

PR #719, opened 2017-06-27
GAWB-1289: Fixing method config page bugs

ansingh7115:
There are a bunch of bugs on the method config details page. 

This rawls PR fixes the following bug:

Currently, if you import a method config with any kind of inputs (required or optional), a submission will be created but will immediately fail because there are no values for the inputs. This should not happen. In the case of required inputs, when you hit the launch button, an error message should tell you that you didn't add give your inputs values. It should not let you create a submission. In the case of optional inputs, you should be able to create a submission, but it should not immediately fail because you don't have values assigned for your inputs (since they're optional).

Interestingly, if you edit and then save the imported config (without actually changing anything), and then attempt to launch, the method configs behave correctly: required inputs with no values do not launch and show an error on the launch modal, optional inputs with no values submit and then run correctly.

This means that there was some kind of issue with the inputs upon import, but not after it had been saved the first time. 

I looked at the way the inputs are saved. Upon import, the METHOD_CONFIG_INPUT table saves the inputs and their values (this column is blank initially). If you edit the method config and save it, a new row is created for the saved method config in the METHOD_CONFIG table. However, new rows were only being created in the METHOD_CONFIG_INPUT table if there were actual values associated with them. If no values were present, no rows were added to the input table.

I'm attempting to fix this in tandem with a ui PR: https://github.com/broadinstitute/firecloud-ui/pull/840

This PR extends what a "missing input" is. Previously a missing input was any required input that was null. However, because the input value was being created as an AttributeString("..."), this meant it was never null, or never missing. (an AttributeString("") should be a missing input. an AttributeString("\"\"") is an empty input). I made it so that any AttributeString("") is missing. This causes any empty input fields to prevent the creation of a submission.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
@@ -136,9 +136,9 @@ trait MethodConfigurationComponent {
       workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
         hideMethodConfigurationAction(currentMethodConfigRec.id, currentMethodConfigRec.name) andThen
         (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) += marshalMethodConfig(workspaceContext.workspaceId,
-          newMethodConfig.copy(methodConfigVersion=currentMethodConfigRec.methodConfigVersion + 1))) flatMap { configId => {
-        saveMaps(newMethodConfig, configId)}
-      }
+          newMethodConfig.copy(methodConfigVersion=currentMethodConfigRec.methodConfigVersion + 1))) flatMap { configId =>
+            saveMaps(newMethodConfig, configId)
+          }
     }
 
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import java.sql.Timestamp
 import java.util.{Date, UUID}
 
-import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.dataaccess.SlickWorkspaceContext
 import org.broadinstitute.dsde.rawls.model._
 import org.joda.time.DateTime
@@ -91,7 +90,7 @@ trait MethodConfigurationComponent {
   protected val methodConfigurationOutputQuery = TableQuery[MethodConfigurationOutputTable]
   protected val methodConfigurationPrereqQuery = TableQuery[MethodConfigurationPrereqTable]
 
-  object methodConfigurationQuery extends TableQuery(new MethodConfigurationTable(_)) with LazyLogging {
+  object methodConfigurationQuery extends TableQuery(new MethodConfigurationTable(_)) {
 
     private type MethodConfigurationQueryType = driver.api.Query[MethodConfigurationTable, MethodConfigurationRecord, Seq]
     private type MethodConfigurationInputQueryType = driver.api.Query[MethodConfigurationInputTable, MethodConfigurationInputRecord, Seq]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -135,8 +135,9 @@ trait MethodConfigurationComponent {
     private def save(workspaceContext: SlickWorkspaceContext, currentMethodConfigRec: MethodConfigurationRecord, newMethodConfig: MethodConfiguration) = {
       workspaceQuery.updateLastModified(workspaceContext.workspaceId) andThen
         hideMethodConfigurationAction(currentMethodConfigRec.id, currentMethodConfigRec.name) andThen
-        (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) += marshalMethodConfig(workspaceContext.workspaceId, newMethodConfig.copy(methodConfigVersion=currentMethodConfigRec.methodConfigVersion + 1))) flatMap { configId =>
-        saveMaps(newMethodConfig, configId)
+        (methodConfigurationQuery returning methodConfigurationQuery.map(_.id) += marshalMethodConfig(workspaceContext.workspaceId,
+          newMethodConfig.copy(methodConfigVersion=currentMethodConfigRec.methodConfigVersion + 1))) flatMap { configId => {
+        saveMaps(newMethodConfig, configId)}
```
helgridly: nitpick: indentation / bracketing

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigResolver.scala
```diff
@@ -53,8 +51,18 @@ object MethodConfigResolver {
   case class MethodInput(workflowInput: WorkflowInput, expression: String)
 
   def gatherInputs(methodConfig: MethodConfiguration, wdl: String): Try[Seq[MethodInput]] = parseWDL(wdl) map { workflow =>
+
+    def isAttributeEmpty(fqn: FullyQualifiedName): Boolean = {
+      methodConfig.inputs.get(fqn) match {
+          // The reason this is specifically an AttributeString rather than any other kind of attribute is because
+          //   MethodConfiguration also only uses AttributeString. Unsure if that's okay?
+        case Some(AttributeString(value)) => value == null || value.isEmpty
```
helgridly: null not necessary here because it can't be

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigResolver.scala
```diff
@@ -53,8 +51,18 @@ object MethodConfigResolver {
...
+        case _ => true
```
helgridly: this case = something reeeally bad has happened

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigResolver.scala
```diff
@@ -53,8 +51,18 @@ object MethodConfigResolver {
...
+      }
+    }
+
     val agoraInputs = workflow.inputs
-    val missingInputs = agoraInputs.filter { case (fqn, workflowInput) => !methodConfig.inputs.contains(fqn) && !workflowInput.optional }.keys
+    val missingInputs = agoraInputs.filter { case (fqn, workflowInput) => (!methodConfig.inputs.contains(fqn) || isAttributeEmpty(fqn)) && !workflowInput.optional }.keys
```
helgridly: yes!

helgridly (commented):
nitpicks but 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigResolver.scala
```diff
@@ -53,8 +51,14 @@ object MethodConfigResolver {
   case class MethodInput(workflowInput: WorkflowInput, expression: String)
 
   def gatherInputs(methodConfig: MethodConfiguration, wdl: String): Try[Seq[MethodInput]] = parseWDL(wdl) map { workflow =>
+    def isAttributeEmpty(fqn: FullyQualifiedName): Boolean = {
+      methodConfig.inputs.get(fqn) match {
+        case Some(AttributeString(value)) => value.isEmpty
+        case _ => throw new AssertionError(s"MethodConfiguration ${methodConfig.namespace}/${methodConfig.name} input ${fqn} value is unavailable")
```
rtitle: You might want to throw `RawlsException` here rather than `AssertionError`. AssertionError is an `Error` not an `Exception`, so it might bubble up where you don't expect it to. `RawlsException` should be handled by the application.

helgridly: yes, this. sorry for lying to you anu

rtitle:
LGTM 👍

PR #721, opened 2017-06-29
[GAWB-2142] Implement metrics for workflow and submission statuses

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-2142

Not quite ready yet as I'm still working on testing -- including end-to-end testing with statsd and stackdriver. I thought I'd open this PR early for review/feedback though.

The following metrics were added:
- Lifetime counts of submission status, broken out by workspace
  - Answers the question:
    -  _How many submissions were submitted/aborted/etc over the last N days [for a given workspace]?_
- Lifetime counts of workflow status, broken out by (workspace, submission)
  - Answers the question:
    - _How many workflows were submitted/queued/launched/aborted/etc over the last N days [for a given workspace and submission]?_
- Current counts of workflow statuses, broken out by (workspace, submission)
  - Answers the question:
    - _How many workflows are queued/running/aborting/etc **right now** [for a given workspace and submission]?_

The last one is more of a gauge of current system status, while the first 2 are more for looking at trends over time.

Note: reviewing tests with `?w=1` is recommended as there are a lot of whitespace changes with the addition of the pattern:
```
withStatsD {
  // test code
}
```

----------

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -438,7 +438,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }.recover { case NonFatal(e) =>
       // Recover on errors since this just affects metrics and we don't want it to blow up the whole actor if it fails
       logger.error("Error occurred checking current workflow status counts", e)
-      Map.empty
+      Map.empty[WorkflowStatus, Int]
     }.map(SaveCurrentWorkflowStatusCounts.apply)
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -6,16 +6,25 @@ import akka.actor._
 import akka.pattern._
 import com.google.api.client.auth.oauth2.Credential
 import com.typesafe.scalalogging.LazyLogging
+import nl.grons.metrics.scala.Counter
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction, ReadWriteAction, WorkflowRecord}
 import org.broadinstitute.dsde.rawls.jobexec.SubmissionMonitorActor._
+import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
+import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
+import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.FutureSupport
 
 import scala.concurrent.duration.{Duration, FiniteDuration}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
+import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction, ReadWriteAction, WorkflowRecord}
+
+import scala.concurrent.{ExecutionContext, Future}
+import akka.pattern._
+import java.util.UUID
 
 import nl.grons.metrics.scala.{Counter, MetricName}
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
@@ -132,9 +141,10 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     def abortQueuedWorkflows(submissionId: UUID) = {
       datasource.inTransaction { dataAccess =>
-        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted)
-      }.andThen { case _ =>
-        workflowStatusCounter(WorkflowStatuses.Aborted) += 1
+        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted).map { n =>
+          workflowStatusCounter(WorkflowStatuses.Aborted) += n
+          n
+        }
       }
     }
 
@@ -237,10 +247,13 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       // to minimize database updates do 1 update per status
       DBIO.seq(updatedRecs.groupBy(_.status).map { case (status, recs) =>
         val workflowStatus = WorkflowStatuses.withName(status)
-        dataAccess.workflowQuery.batchUpdateStatus(recs, workflowStatus).map(n => workflowStatusCounter(workflowStatus) += n)
+        dataAccess.workflowQuery.batchUpdateStatus(recs, workflowStatus).map { n =>
+          workflowStatusCounter(workflowStatus) += n
+          n
+        }
       }.toSeq: _*) andThen
         DBIO.successful(workflowsWithOutputs)
-      }
+    }
 
     //Then in a new transaction, update the submission status.
     //If this transaction throws an exception, it'll roll back, but the workflows will still be in their terminal state.
@@ -278,7 +291,10 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           }
         } flatMap { newStatus =>
           logger.debug(s"submission $submissionId terminating to status $newStatus")
-          dataAccess.submissionQuery.updateStatus(submissionId, newStatus).map(n => submissionStatusCounter(newStatus) += n)
+          dataAccess.submissionQuery.updateStatus(submissionId, newStatus).map { n =>
+            submissionStatusCounter(newStatus) += n
+            n
+          }
         } map(_ => true)
       } else {
         DBIO.successful(false)
@@ -375,7 +391,10 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
   def saveErrors(errors: Seq[(WorkflowRecord, Seq[AttributeString])], dataAccess: DataAccess)(implicit executionContext: ExecutionContext) = {
     DBIO.sequence(errors.map { case (workflowRecord, errorMessages) =>
-      dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed).map(n => workflowStatusCounter(WorkflowStatuses.Failed) += n) andThen
+      dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed).map { n =>
+        workflowStatusCounter(WorkflowStatuses.Failed) += n
+        n
+      } andThen
         dataAccess.workflowQuery.saveMessages(errorMessages, workflowRecord.id)
     })
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -74,7 +74,8 @@ class SubmissionMonitorActor(val workspaceName: WorkspaceName,
                              override val rawlsMetricBaseName: String) extends Actor with SubmissionMonitor with LazyLogging {
   import context._
 
-  // Marked volatile because it is accessed by statsd in a separate thread
+  // Marked volatile because it is read by a separate statsd thread.
+  // It is only written by the actor.
   @volatile
   private var currentWorkflowStatusCounts: Map[WorkflowStatus, Int] = Map.empty
 
@@ -417,9 +418,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
   def saveErrors(errors: Seq[(WorkflowRecord, Seq[AttributeString])], dataAccess: DataAccess)(implicit executionContext: ExecutionContext) = {
     DBIO.sequence(errors.map { case (workflowRecord, errorMessages) =>
-      dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed).map { n =>
-        workflowStatusCounter(WorkflowStatuses.Failed) += n
-        n
+      dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed).map { count =>
+        workflowStatusCounter(WorkflowStatuses.Failed) += count
+        count
       } andThen
         dataAccess.workflowQuery.saveMessages(errorMessages, workflowRecord.id)
     })
@@ -428,10 +429,10 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
   def checkCurrentWorkflowStatusCounts(implicit executionContext: ExecutionContext): Future[SaveCurrentWorkflowStatusCounts] = {
     datasource.inTransaction { dataAccess =>
       dataAccess.workflowQuery.countWorkflowsForSubmissionByQueueStatus(submissionId)
-    }.map { m =>
-      SaveCurrentWorkflowStatusCounts(m.map { case (k, v) => WorkflowStatuses.withName(k) -> v })
+    }.map { workflowStatusCounts =>
+      SaveCurrentWorkflowStatusCounts(workflowStatusCounts.map { case (k, v) => WorkflowStatuses.withName(k) -> v })
     }.recover { case NonFatal(e) =>
-      // Recover on errors since this just affects metrics and we don't want it to blow up the whole actor
+      // Recover on errors since this just affects metrics and we don't want it to blow up the whole actor if it fails
       logger.error("Error occurred checking current workflow status counts", e)
       SaveCurrentWorkflowStatusCounts(Map.empty)
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -6,7 +6,7 @@ import akka.actor._
 import akka.pattern._
 import com.google.api.client.auth.oauth2.Credential
 import com.typesafe.scalalogging.LazyLogging
-import nl.grons.metrics.scala.{Counter, Gauge}
+import nl.grons.metrics.scala.Counter
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction, ReadWriteAction, WorkflowRecord}
@@ -18,20 +18,8 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.FutureSupport
 
 import scala.concurrent.duration._
-import scala.concurrent.{Await, ExecutionContext, Future}
-import scala.util.{Failure, Success, Try}
-import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction, ReadWriteAction, WorkflowRecord}
-
 import scala.concurrent.{ExecutionContext, Future}
-import akka.pattern._
-import java.util.UUID
-
-import nl.grons.metrics.scala.{Counter, MetricName}
-import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
-import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
-import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
-
-import scala.collection.immutable
+import scala.util.{Failure, Success, Try}
 
 /**
  * Created by dvoet on 6/26/15.
@@ -112,19 +100,6 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
   val credential: Credential
   val submissionPollInterval: Duration
 
-  def workflowStatusCounter(status: WorkflowStatus): Counter =
-    ExpandedMetricBuilder
-      .expand(WorkspaceMetric, workspaceName)
-      .expand(SubmissionMetric, submissionId)
-      .expand(WorkflowStatusMetric, status)
-      .asCounter()
-
-  def submissionStatusCounter(status: SubmissionStatus): Counter =
-    ExpandedMetricBuilder
-      .expand(WorkspaceMetric, workspaceName)
-      .expand(SubmissionStatusMetric, status)
-      .asCounter()
-
   import datasource.dataAccess.driver.api._
 
   /**
@@ -140,9 +115,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     def abortQueuedWorkflows(submissionId: UUID) = {
       datasource.inTransaction { dataAccess =>
-        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted).map { n =>
-          workflowStatusCounter(WorkflowStatuses.Aborted) += n
-          n
+        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted).map { count =>
+          workflowStatusCounter(WorkflowStatuses.Aborted) += count
+          count
         }
       }
     }
@@ -246,9 +221,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       // to minimize database updates do 1 update per status
       DBIO.seq(updatedRecs.groupBy(_.status).map { case (status, recs) =>
         val workflowStatus = WorkflowStatuses.withName(status)
-        dataAccess.workflowQuery.batchUpdateStatus(recs, workflowStatus).map { n =>
-          workflowStatusCounter(workflowStatus) += n
-          n
+        dataAccess.workflowQuery.batchUpdateStatus(recs, workflowStatus).map { count =>
+          workflowStatusCounter(workflowStatus) += count
+          count
         }
       }.toSeq: _*) andThen
         DBIO.successful(workflowsWithOutputs)
@@ -290,9 +265,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           }
         } flatMap { newStatus =>
           logger.debug(s"submission $submissionId terminating to status $newStatus")
-          dataAccess.submissionQuery.updateStatus(submissionId, newStatus).map { n =>
-            submissionStatusCounter(newStatus) += n
-            n
+          dataAccess.submissionQuery.updateStatus(submissionId, newStatus).map { count =>
+            submissionStatusCounter(newStatus) += count
+            count
           }
         } map(_ => true)
       } else {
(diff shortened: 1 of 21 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+    def expand[A: Expansion](key: String, a: A) = {
+      new ExpandedMetricBuilder().expand(key, a)
+    }
+  }
+
+  /**
+    * Adds a .countDBResult method to Counter which counts the result of a numeric DBIOAction.
+    */
+  protected implicit class CounterDBIOActionSupport(counter: Counter) {
+    def countDBResult[R <% Long, S <: NoStream, E <: Effect](action: DBIOAction[R, S, E])(implicit executionContext: ExecutionContext): DBIOAction[R, NoStream, E] =
```
jmthibault79: IntelliJ gave me a deprecation warning here: https://issues.scala-lang.org/browse/SI-7629

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -110,7 +172,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging {
 
     def abortQueuedWorkflows(submissionId: UUID) = {
       datasource.inTransaction { dataAccess =>
-        dataAccess.workflowQuery.batchUpdateWorkflowsOfStatus(submissionId, WorkflowStatuses.Queued, WorkflowStatuses.Aborted)
+        workflowStatusCounter(WorkflowStatuses.Aborted).countDBResult {
```
jmthibault79: What do you think about updating counters at the `WorkflowComponent` level?  That way every call to `batchUpdateWorkflowsOfStatus()` etc. would be instrumented without higher levels having to worry about it.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -123,20 +133,25 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
...
-            DBIO.successful(workflowRecs)
+            DBIO.successful((Seq.empty, None, None))
           } else {
-            //they should also all have the same submission ID
-            DBIO.successful(workflowRecs.filter(_.submissionId == workflowRecs.head.submissionId))
+            // they should also all have the same submission ID
+            val submissionId = workflowRecs.head.submissionId
+            val filteredWorkflowRecs = workflowRecs.filter(_.submissionId == submissionId)
+            for {
+              submissionRecOpt <- dataAccess.submissionQuery.findById(submissionId).result.map(_.headOption)
```
jmthibault79: I like `headOption` for safety generally, but these are foreign key relations so Rawls should probably complain more loudly when this assumption fails

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: View bounds are deprecated? News to me. There's probably another way to do this.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
rtitle: True - I saw `_.head` in other places. Maybe I'll change this to be consistent.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+    */
+  protected val rawlsMetricBaseName: String
+  override lazy val metricBaseName = MetricName(rawlsMetricBaseName)
+
+  /**
+    * Typeclass for something that can be converted into a metric name fragment with a given key.
+    * Metric name fragments are combined via ExpandedMetricBuilder to generate an "expanded" metric name.
+    * By default this just calls toString on the object of type A, but this can be overridden.
+    */
+  @implicitNotFound(msg = "Cannot expand instances of type ${A}")
```
jmthibault79, line 38: oh this is cool

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1428,7 +1435,11 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
           workflowFailureMode = workflowFailureMode
         )
 
-        dataAccess.submissionQuery.create(workspaceContext, submission) map { _ =>
+        dataAccess.submissionQuery.create(workspaceContext, submission) map { sub =>
+          submissionStatusCounter(workspaceName, SubmissionStatuses.Submitted) += 1
+          workflowStatusCounter(workspaceName, UUID.fromString(submissionId), WorkflowStatuses.Queued) += workflows.size
+          sub
+        } map { _ =>
```
jmthibault79: can we remove this line and also stop holding on to the `sub` ?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: My thinking was that `WorkflowComponent` should only be concerned with updating the database, and that metric updates should be at the level above it (in the actual business logic). Plus, these metrics usually need a `Workspace` and a `Submission` which would need to be passed into `WorkflowComponent` solely for the purpose of updating metrics, which doesn't seem right.

Maybe instrumentation relating specifically to the database could live at the `*Component` layer (such as query timing, etc). But I felt these metrics are more closely tied to the actor logic.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: Yeah it gives nicer compiler messages :)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: Yeah no need for the double-`map` I guess :)

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+    SharedHealthCheckRegistries.clear()
+    val reporter = StatsDReporter.forRegistry(SharedMetricRegistries.getOrCreate("default"))
+      .convertRatesTo(TimeUnit.SECONDS)
+      .convertDurationsTo(TimeUnit.MILLISECONDS)
+      .build(statsD)
+    reporter.start(1, TimeUnit.SECONDS)
+    try {
+      val result = testCode
+      eventually(timeout(10 seconds)) {
+        val order = mockitoInOrder(statsD)
```
jmthibault79, line 35: I haven't used Mockito before.  Let me see if I have interpreted this correctly.

* record the calls to `statsD` methods in order
* verify that the calls were made in the order `connect()`, `send()` at least once, and then `close()`
* enable callers to access the arguments given to `send()` for further test verification

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+      }
+      result
+    } finally {
+      reporter.stop()
+      SharedMetricRegistries.clear()
+      SharedHealthCheckRegistries.clear()
+    }
+  }
+
+  protected def expectedMetric(workspace: Workspace, submission: Submission, workflowStatus: WorkflowStatus, name: Option[String], expectedTimes: Int): (String, String) =
```
jmthibault79: expectedWFStatusMetric

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/RemoteServicesMockServer.scala
```diff
@@ -331,7 +331,7 @@ class RemoteServicesMockServer(port:Int) extends RawlsTestUtils {
       response()
         .withHeaders(jsonHeader)
         .withBody {
-          (1 to numWorkflows).map(_ => ExecutionServiceStatus(UUID.randomUUID().toString, "Submitted")).toList.toJson.toString
+          (1 to numWorkflows).map(n => ExecutionServiceStatus(UUID.randomUUID().toString, if (n == numWorkflows) "Failed" else "Submitted")).toList.toJson.toString
```
jmthibault79: what is this change for?

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
rtitle: Yes that's exactly right. Furthermore it checks that `connect()` and `close()` were called exactly once, and it waits up to 10 seconds (polling every x milliseconds, whatever the default is) before giving up.

rtitle: Yes that's exactly right. Furthermore it checks that `connect()` and `close()` were called exactly once, and it waits up to 10 seconds (polling every x milliseconds, whatever the default is) before giving up.

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
rtitle: Yeah thanks

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
rtitle: Park street strikes again

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/RemoteServicesMockServer.scala
rtitle: Long story. :) 

This used to be hard coded to return 2 successes and 1 failure. In https://github.com/broadinstitute/rawls/pull/718 I changed it to be dynamic, but lost the failure. The tests still passed.

However when working on the instrumentation tests I found the test whose _intent_ it was to check for a failure -- the assertion was just weak (it was checking `status != Queued` or something). But while verifying instrumentation the test actually failed due to lack of a wf failure status. So I changed this logic to make the last workflow a failure in Cromwell.

jmthibault79: hoo boy.  :)

Could you add `// last one fails` or something to make the intention clear?

rtitle: Yeah. 

BTW this is one of the reasons I like Mockito: it makes every test responsible for its own mocked data so you're not guessing which test uses which piece of a global test data set.

http://www.scalacheck.org/ is another nice tool for test data generation & property based testing -- I could give the team an overview sometime if interested.

geoffjentry: Id love to see scalacheck start to get pulled in more

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -77,17 +106,50 @@ class SubmissionMonitorActor(val workspaceName: WorkspaceName,
       logger.debug(s"done checking status for submission $submissionId, terminateActor = $terminateActor")
       if (terminateActor) stop(self)
       else scheduleNextMonitorPass
+    case CheckCurrentWorkflowStatusCounts =>
+      logger.debug(s"check current workflow status counts for submission $submissionId")
+      checkCurrentWorkflowStatusCounts pipeTo self
+    case SaveCurrentWorkflowStatusCounts(statusCounts) =>
```
helgridly, line 112: Does this need to be a separate message at all? It's not doing much. Why not put it inside `checkCurrentWorkflowStatusCounts`?

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+    *       .asCounter("count")
+    *   // counter has name:
+    *   // <baseName>.workspace.<workspaceNamespace>.<workspaceName>.submission.<submissionId>.workflowStatus.<workflowStatus>.count
+    *   counter += 1000
+    * }}}
+    *
+    * Note the above will only compile if there are [[Expansion]] instances for the types passed to the expand method.
+    */
+  protected class ExpandedMetricBuilder[A] private (m: String = "") {
+    def expand[A: Expansion](key: String, a: A) = {
```
helgridly: The reuse of `A` here is giving a "suspicious shadowing by a Type Parameter" warning for me in IntelliJ.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+    * By default this just calls toString on the object of type A, but this can be overridden.
+    */
+  @implicitNotFound(msg = "Cannot expand instances of type ${A}")
+  protected sealed trait Expansion[A] {
+    def makeName(key: String, a: A) = s"$key.${a.toString}"
+  }
+
+  // Typeclass instances:
+
+  protected implicit object WorkspaceNameExpansion extends Expansion[WorkspaceName] {
```
helgridly, line 45: comment here: overriding `makeName` to override default `toString` based implementation

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+    override def makeName(key: String, n: WorkspaceName): String = s"$key.${n.toString.replace('/', '.')}"
+  }
+
+  protected implicit object UUIDExpansion extends Expansion[UUID]
```
helgridly: comment here: providing an implicit for UUIDs but can use the default `makeName`

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  protected implicit def WorkflowStatusExpansion[A <: WorkflowStatus] = new Expansion[A] {}
```
helgridly: why can't this be `implicit object WorkflowStatusExpansion extends Expansion[WorkflowStatus]`? that compiles for me

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -77,17 +106,50 @@ class SubmissionMonitorActor(val workspaceName: WorkspaceName,
...
+  private def scheduleNextCheckCurrentWorkflowStatus: Cancellable = {
+    system.scheduler.scheduleOnce(addJitter(submissionPollInterval), self, CheckCurrentWorkflowStatusCounts)
+  }
+
+  private def initGauges(): Unit = {
+    try {
+      WorkflowStatuses.allStatuses.foreach { status =>
+        ExpandedMetricBuilder
+          .expand(WorkspaceMetric, workspaceName)
+          .expand(SubmissionMetric, submissionId)
```
helgridly: Can we pre-create and reuse as far as:

```
ExpandedMetricBuilder
    .expand(WorkspaceMetric, workspaceName)
    .expand(SubmissionMetric, submissionId)
```

since we're re-building this all over the place (not just here but every time `workflowStatusCounter` and `submissionStatusCounter` are called)?

Or am I prematurely optimising?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -77,17 +106,50 @@ class SubmissionMonitorActor(val workspaceName: WorkspaceName,
...
+      logger.debug(s"saving current workflow status counts for submission $submissionId")
+      currentWorkflowStatusCounts = statusCounts
+      scheduleNextCheckCurrentWorkflowStatus
 
     case Status.Failure(t) => throw t // an error happened in some future, let the supervisor handle it
   }
 
-  def scheduleNextMonitorPass: Cancellable = {
+  private def scheduleNextMonitorPass: Cancellable = {
     system.scheduler.scheduleOnce(submissionPollInterval, self, StartMonitorPass)
```
helgridly: since you're nearby: this probably wants jitter too

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -107,36 +114,41 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
...
+      case _ => Future.successful(ScheduleNextWorkflow)
     } recover {
       // if we found some but another actor reserved the first look again immediately
-      case t: RawlsConcurrentModificationException => ProcessNextWorkflow
+      case _: RawlsConcurrentModificationException => ProcessNextWorkflow
     }
   }
 
-  def reserveWorkflowBatch(dataAccess: DataAccess, activeCount: Int)(implicit executionContext: ExecutionContext): ReadWriteAction[Seq[WorkflowRecord]] = {
+  def reserveWorkflowBatch(dataAccess: DataAccess, activeCount: Int)(implicit executionContext: ExecutionContext): ReadWriteAction[Option[(Seq[WorkflowRecord], SubmissionRecord, WorkspaceRecord)]] = {
```
helgridly, line 131: Blergh. I see why you need this - you need the submission and workspace details for `getUnlaunchedWorkflowBatch` - but this feels like a super heavyweight solution. I can't see another way around it though (sure, we could pass just the `submissionId` and `WorkspaceName` but it's still an option-of-triple).

rtitle:
Successfully tested pushing these metrics to stackdriver. See:

https://app.google.stackdriver.com/monitoring/1052499/rawls-metrics?project=broad-dsp-stackdriver
screen shot 2017-07-05 at 6 01 14 pm

They are appearing as expected as "flat" metrics in the SD UI. The next step will be to auto-create labels so we can do interesting aggregation in SD. (That will be done as part of the statsd/collectd plugin, not Rawls.)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2146,6 +2157,19 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
     }
   }
+
+  private def submissionStatusCounter(workspaceName: WorkspaceName, status: SubmissionStatus): Counter =
+    ExpandedMetricBuilder
+      .expand(WorkspaceMetric, workspaceName)
+      .expand(SubmissionStatusMetric, status)
```
helgridly: submission id here?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: Yeah, I see your point. I guess I was just hesitant to add side effects to what is now pure database-affecting code. But you're right it might make the code easier to work with. Let me take a stab at moving the counters to `WorkflowComponent` (there are still multiple methods that need updating) and see how it looks.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: Good question! Because `checkCurrentWorkflowStatusCounts` is running in a `Future` (using a thread from the `SlickDataSource` thread pool) and it's only safe to update actor state from the actor thread itself. So I pipe the database Future back to the actor and update the actor state upon receiving a `SaveCurrentWorkflowStatusCounts` message.

helgridly: Oh, this is terrifying. Gonna look into this to wrap my head around it.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
rtitle: Yeah probably could reuse that. Though I don't think building the metric names is too expensive. :)

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: Ah, that works because we currently always pass the supertype `WorkflowStatus` to `ExpandedMetricBuilder`. E.g.:
```
// this compiles
private def workflowStatusCounter(status: WorkflowStatus): Counter =
  ExpandedMetricBuilder
    .expand(WorkspaceMetric, workspaceName)
    .expand(SubmissionMetric, submissionId)
    .expand(WorkflowStatusMetric, status)
    .asCounter()
```
But it doesn't work if you try to pass a subtype of `WorkflowStatus`:
```
// this does not compile: Cannot expand instances of type org.broadinstitute.dsde.rawls.model.WorkflowStatuses.Launching.type
private def foo: Counter = 
  ExpandedMetricBuilder
    .expand(WorkflowStatusMetric, WorkflowStatuses.Launching)
    .asCounter()
```
So that's why I added the upper type bound `A <: WorkflowStatus`. Originally I thought I could solve this by making the type parameter covariant (e.g. `sealed trait Expansion[+A]`) but that opened up a whole new can of worms.

So I think this is the cleaner solution, although maybe it could use a comment.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: Wow good catch! The type parameter on the class is totally unneeded and left over from an earlier incarnation of this.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
rtitle: +1 on Blergh. On the plus side it saves some database queries because we no longer need to re-lookup the `Workspace` and `Submission` in `submitWorkflowBatch` (since we remember them from workflow reservation time).

Maybe it could be made a bit cleaner with a container case class:
```
case class WorkflowBatch(workflowIds: Seq[Long], submissionRec: SubmissionRecord, workspaceRec: WorkspaceRecord)
```
and we pass that around instead of a `Tuple3`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: Actually I was intending to count submission statuses per workspace, so it is not scoped to a single submission.

rtitle:
@helgridly I pushed code review feedback, including moving incrementing the counters to the `WorkflowComponent`/`SubmissionComponent` classes. Tests are still passing - I plan to repeat my FiaB test as well. Can you take another look? Thanks.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: It's such a random glom of objects that making a case class just feels like denial. Might as well leave it as-is.

helgridly: Oh, you already changed it. Never mind then.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -95,10 +101,27 @@ trait RawlsInstrumented extends DefaultInstrumented {
...
 
+  // Handy counter definitions which can be used by implementing classes:
+
+  protected def submissionStatusCounterProvider(workspaceName: WorkspaceName) =
+    (status: SubmissionStatus) => ExpandedMetricBuilder
+      .expand(WorkspaceMetric, workspaceName)
+      .expand(SubmissionStatusMetric, status)
+      .asCounter()
+
+  protected def workflowStatusCounterProvider(workspaceName: WorkspaceName, submissionId: UUID) =
```
helgridly: but this is `SubmissionMonitorActor.workflowStatusCounter` by another name, no? Can't we move that implicit somewhere shared and reuse it?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -202,7 +200,11 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
   }
 
   //submit the batch of workflows with the given ids
-  def submitWorkflowBatch(workflowIds: Seq[Long], submissionRec: SubmissionRecord, workspaceRec: WorkspaceRecord)(implicit executionContext: ExecutionContext): Future[WorkflowSubmissionMessage] = {
+  def submitWorkflowBatch(batch: WorkflowBatch)(implicit executionContext: ExecutionContext): Future[WorkflowSubmissionMessage] = {
+
+    val WorkflowBatch(workflowIds, submissionRec, workspaceRec) = batch
```
helgridly: TIL this offensively ugly syntax to do a useful thing.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
helgridly: Great explanation, thanks!

helgridly:
Can we collapse `SubmissionMonitorActor.workflowStatusCounter` and `RawlsInstrumented.workflowStatusCounterProvider` down to the same object, presumably a shared implicit defined somewhere sensible? Same for `submissionStatusCounter` and any other similar friends.

👍 otherwise (and feel free to merge after the above change if my request is uncontroversial)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
jmthibault79: IMO random-glom-of-objects is a primary use case for case classes

PR #730, opened 2017-07-14
GAWB-1859 fix for attach outputs issue

helgridly:
The test attached fails to bind all outputs on develop. It passes here!

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -222,7 +222,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
             _ <- abortActiveWorkflows(submissionId)
           } yield {}
         } else {
-          Future.successful()
+          Future.successful(Unit)
         }
         abortFuture flatMap( _ => queryForWorkflowStatuses() )
       case None => throw new RawlsException(s"Submission ${submissionId} could not be found")
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -271,29 +271,31 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     // Update statuses for workflows and submission in a separate txn.
     val statusUpdateTxn = datasource.inTransaction { dataAccess =>
+
       // Refetch workflows as some may have been marked as Failed by handleOutputs.
-      val workflowUpdates = dataAccess.workflowQuery.findWorkflowByIds( workflowsWithOutputs.map( _._1.id ) ).result map { updatedRecs =>
+      val workflowUpdatesDB = dataAccess.workflowQuery.findWorkflowByIds( workflowsWithOutputs.map( _._1.id ) ).result map { updatedRecs =>
+
         //New statuses according to the execution service.
-        val workflowIdToNewStatus = workflowsWithOutputs.map({ case (workflowRec, outputs) => workflowRec.id -> workflowRec.status }).toMap
+        val workflowIdToNewStatus = workflowsWithOutputs.map({ case (workflowRec, _) => workflowRec.id -> workflowRec.status }).toMap
 
         // No need to update statuses for any workflows that are in terminal statuses.
         // Doing so would potentially overwrite them with the execution service status if they'd been marked as failed by attachOutputs.
         val workflowsToUpdate = updatedRecs.filter(rec => !WorkflowStatuses.terminalStatuses.contains(WorkflowStatuses.withName(rec.status)))
-        val updatedWorkflows = workflowsToUpdate.map(rec => rec.copy(status = workflowIdToNewStatus(rec.id)))
+        val workflowsWithNewStatuses = workflowsToUpdate.map(rec => rec.copy(status = workflowIdToNewStatus(rec.id)))
 
         // to minimize database updates batch 1 update per workflow status
-        DBIO.seq(updatedWorkflows.groupBy(_.status).map { case (status, recs) =>
+        DBIO.seq(workflowsWithNewStatuses.groupBy(_.status).map { case (status, recs) =>
           dataAccess.workflowQuery.batchUpdateStatus(recs, WorkflowStatuses.withName(status))
         }.toSeq: _*)
       }
 
       // update submission after workflows are updated
-      val submissionUpdates = checkOverallStatus(dataAccess) map { shouldStop: Boolean =>
+      val submissionUpdatesDB = checkOverallStatus(dataAccess) map { shouldStop: Boolean =>
         //return a message about whether our submission is done entirely
         StatusCheckComplete(shouldStop)
       }
 
-      workflowUpdates andThen submissionUpdates
+      workflowUpdatesDB andThen submissionUpdatesDB
     }
 
     attachOutputsTxn.flatMap( _ => statusUpdateTxn )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -269,19 +269,18 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     //just the workflow records in this response list which have outputs
     val workflowsWithOutputs = response.statusResponse.collect {
       case Success(Some((workflowRec, Some(outputs)))) =>
-        //println(s"workflow ${workflowRec.submissionId.toString.split("-").head} has outputs")
         (workflowRec, outputs)
     }
 
     // Attach the outputs in a txn of their own. If attaching outputs fails, it will mark the workflow as failed. This is fine.
-    /*val attachOutputsTxn =*/ datasource.inTransaction { dataAccess =>
+    datasource.inTransaction { dataAccess =>
       handleOutputs(workflowsWithOutputs, dataAccess)
     } flatMap { _ =>
-      // Update statuses for workflows and submission in a separate txn.
+      // NEW TXN! Update statuses for workflows and submission.
       datasource.inTransaction { dataAccess =>
 
         // Refetch workflows as some may have been marked as Failed by handleOutputs.
-        /*val workflowUpdatesDB = */ dataAccess.workflowQuery.findWorkflowByIds(workflowsWithStatuses.map(_.id)).result flatMap { updatedRecs =>
+        dataAccess.workflowQuery.findWorkflowByIds(workflowsWithStatuses.map(_.id)).result flatMap { updatedRecs =>
 
           //New statuses according to the execution service.
           val workflowIdToNewStatus = workflowsWithStatuses.map({ workflowRec => workflowRec.id -> workflowRec.status }).toMap
@@ -292,33 +291,19 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           val workflowsWithNewStatuses = workflowsToUpdate.map(rec => rec.copy(status = workflowIdToNewStatus(rec.id)))
 
           // to minimize database updates batch 1 update per workflow status
-          val zzz: Iterable[ReadWriteAction[Int]] = workflowsWithNewStatuses.groupBy(_.status).map { case (status, recs) =>
-            val x: ReadWriteAction[Int] = dataAccess.workflowQuery.batchUpdateStatus(recs, WorkflowStatuses.withName(status))
-            x
-          }
-          val fff = DBIO.seq(zzz.toSeq: _*)
+          DBIO.seq( workflowsWithNewStatuses.groupBy(_.status).map { case (status, recs) =>
+              dataAccess.workflowQuery.batchUpdateStatus(recs, WorkflowStatuses.withName(status))
+          }.toSeq: _*)
 
-
-          fff.asTry map {
-            case Success(aaa) => DBIO.successful(aaa)
-            case Failure(regret) =>
-              println(s"$submissionId failure: $regret")
-              DBIO.failed(regret)
-          }
         } flatMap { _ =>
-
-        // update submission after workflows are updated
-        /* val submissionUpdatesDB = */ updateSubmissionStatus(dataAccess) map { shouldStop: Boolean =>
-          //return a message about whether our submission is done entirely
-          StatusCheckComplete(shouldStop)
+          // update submission after workflows are updated
+          updateSubmissionStatus(dataAccess) map { shouldStop: Boolean =>
+            //return a message about whether our submission is done entirely
+            StatusCheckComplete(shouldStop)
+          }
         }
       }
-
-        //workflowUpdatesDB flatMap { _ => submissionUpdatesDB }
-      }
     }
-
-    //attachOutputsTxn.flatMap( _ => statusUpdateTxn )
   }
 
   /**
@@ -337,7 +322,6 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
             case _ => SubmissionStatuses.Done
           }
         } flatMap { newStatus =>
-          println(s"submission $submissionId terminating to status $newStatus")
           dataAccess.submissionQuery.updateStatus(submissionId, newStatus)
         } map(_ => true)
       } else {
@@ -350,8 +334,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     if (workflowsWithOutputs.isEmpty) {
       DBIO.successful(Unit)
     } else {
-      val xoxo = workflowsWithOutputs.map( _._2.outputs.keys).flatten
-      //println(s"${workflowsWithOutputs.head._1.submissionId.toString.split("-").head} handleOutputs")
+      workflowsWithOutputs.map( _._2.outputs.keys).flatten
       for {
         // load all the starting data
         entitiesById <-      listWorkflowEntitiesById(workflowsWithOutputs, dataAccess)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -216,16 +216,15 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     submissionFuture flatMap {
       case Some(submission) =>
-        if(submission.status == SubmissionStatuses.Aborting) {
+        val abortFuture = if(submission.status == SubmissionStatuses.Aborting) {
           for {
             _ <- abortQueuedWorkflows(submissionId)
             _ <- abortActiveWorkflows(submissionId)
-            getStatuses <- queryForWorkflowStatuses()
-          } yield getStatuses
+          } yield {}
+        } else {
+          Future.successful()
         }
-        else {
-          queryForWorkflowStatuses()
-        }
+        abortFuture flatMap( _ => queryForWorkflowStatuses() )
       case None => throw new RawlsException(s"Submission ${submissionId} could not be found")
     }
   }
@@ -273,7 +272,12 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         (workflowRec, outputs)
     }
 
-    // Attach the outputs in a txn of their own. If attaching outputs fails, it will mark the workflow as failed. This is fine.
+    // Attach the outputs in a txn of their own.
+    // If attaching outputs fails for legit reasons (e.g. they're missing), it will mark the workflow as failed. This is correct.
+    // If attaching outputs throws an exception (because e.g. deadlock or ConcurrentModificationException), the status will remain un-updated
+    // and will be re-processed next time we call queryForWorkflowStatus().
+    // This is why it's important to attach the outputs before updating the status -- if you update the status to Successful first, and the attach
+    // outputs fails, we'll stop querying for the workflow status and never attach the outputs.
     datasource.inTransaction { dataAccess =>
       handleOutputs(workflowsWithOutputs, dataAccess)
     } flatMap { _ =>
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -274,40 +260,43 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+    val attachOutputsTxn = datasource.inTransaction { dataAccess =>
+      handleOutputs(workflowsWithOutputs, dataAccess)
+    }
+
+    // Update statuses for workflows and submission in a separate txn.
+    val statusUpdateTxn = datasource.inTransaction { dataAccess =>
+      // Refetch workflows as some may have been marked as Failed by handleOutputs.
+      val workflowUpdates = dataAccess.workflowQuery.findWorkflowByIds( workflowsWithOutputs.map( _._1.id ) ).result map { updatedRecs =>
+        //New statuses according to the execution service.
+        val workflowIdToNewStatus = workflowsWithOutputs.map({ case (workflowRec, outputs) => workflowRec.id -> workflowRec.status }).toMap
```
jmthibault79: outputs are not used here, so you can do `case (workflowRec, _)` or even just get the `keys()` first

helgridly: tru dat

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -188,7 +188,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       }
     }
 
-    def abortActiveWorkflows(submissionId: UUID) = {
+    def abortActiveWorkflows(submissionId: UUID): Future[Seq[Future[(Option[String], Try[ExecutionServiceStatus])]]] = {
```
rtitle: I know you didn't change this function, but that type signature is a bit of a red flag to me. Furthermore it doesn't look like we're mapping over the inner `Seq[Future[...]]` (just the outer Future) so it seems like `executionServiceCluster.abort()` could be executing concurrently with `executionServiceCluster.outputs()`, which doesn't seem right. Will think about how to clean that up...

helgridly: I saw that and had the same thoughts. It's probably part of why we send a _lot_ of /abort requests to Cromwell. We could traverse and flatten, but that'd have us blocking on the HTTP responses here. I don't actually think it's super terrible that these execute async - we might send a few more of them than we need to but it'll clear up eventually.

rtitle: Well I'd still advocate fixing this method to have a sane type signature, and if we don't want to wait for the results before checking for status then just don't call it in a for comprehension below.

My attempt at fixing the method:
```
    def abortActiveWorkflows(submissionId: UUID): Future[Seq[Try[ExecutionServiceStatus]]] = {
        // look up abortable WorkflowRecs for this submission
        dataAccess.workflowQuery.findWorkflowsForAbort(submissionId).result
      }.flatMap { workflows =>
        Future.traverse(workflows) { workflow =>
          executionServiceCluster.abort(workflow, UserInfo.buildFromTokens(credential))
        }
      }
    }
```

rtitle: And then:
```
    submissionFuture flatMap {
      case Some(submission) =>
        if (submission.status == SubmissionStatuses.Aborting) {
          // kick off asynchronous aborts but don't wait for results 
          abortQueuedWorkflows(submissionId)
          abortActiveWorkflows(submissionId)
        }
        queryForWorkflowStatuses()
      case None => throw new RawlsException(s"Submission ${submissionId} could not be found")
    }
```

dvoet:
so the gist of this is that we keep the 2 transaction but reverse the order to that if the attach fails and triggers a retry the workflow won't be ignored later. Correct?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
helgridly: hah. i was about to say "but then you lose the `externalId`s in the return type!", and then realised we're throwing away the result anyway. will incorporate, thanks.

helgridly: though i'm still gonna zip in the ids.

helgridly: also we should wait for `abortQueued` because that's a db. also barf, why are these futures anyway

helgridly:
@rtitle @jmthibault79 @dvoet this is done for reals now, please re-review

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -317,7 +314,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
    * @param executionContext
    * @return true if the submission is done/aborted
    */
-  def checkOverallStatus(dataAccess: DataAccess)(implicit executionContext: ExecutionContext): ReadWriteAction[Boolean] = {
+  def updateSubmissionStatus(dataAccess: DataAccess)(implicit executionContext: ExecutionContext): ReadWriteAction[Boolean] = {
```
helgridly, line 321: I renamed this because `check` sounds like a read-only operation to me.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -50,7 +50,7 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
   }
 
   override val supervisorStrategy =
-    OneForOneStrategy(maxNrOfRetries = 3) {
+    OneForOneStrategy(maxNrOfRetries = -1) {
```
helgridly: are we okay with this?

dvoet: would be nice if we had a way to monitor for out of control submission monitor actors

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -274,36 +262,45 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-    //Then in a new transaction, update the submission status.
-    //If this transaction throws an exception, it'll roll back, but the workflows will still be in their terminal state.
-    //The submission supervisor will restart the submission monitor, which will find this submission (again) in a non-terminal state,
-    //query Cromwell again, and restart from the top of this function with the results.
-    workflowsWithOutputsF flatMap { workflowsWithOutputs =>
+    // Attach the outputs in a txn of their own. If attaching outputs fails, it will mark the workflow as failed. This is fine.
+    datasource.inTransaction { dataAccess =>
+      handleOutputs(workflowsWithOutputs, dataAccess)
+    } flatMap { _ =>
+      // NEW TXN! Update statuses for workflows and submission.
```
dvoet: add a comment as to why the transaction order is important

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: i guess i could try counting the number of retries and start spamming errors if it gets too high. sentry alert?

helgridly: (monitoring would help with this too, though only after the fact)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
helgridly: k

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
jmthibault79: :+1:

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
jmthibault79: 💯

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -216,8 +218,8 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       case Some(submission) =>
         if(submission.status == SubmissionStatuses.Aborting) {
           for {
-            abortQueued <- abortQueuedWorkflows(submissionId)
-            abortActive <- abortActiveWorkflows(submissionId)
+            _ <- abortQueuedWorkflows(submissionId)
+            _ <- abortActiveWorkflows(submissionId)
             getStatuses <- queryForWorkflowStatuses()
```
jmthibault79: Can move this outside the `if`

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -317,7 +314,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
    * @param executionContext
    * @return true if the submission is done/aborted
    */
-  def checkOverallStatus(dataAccess: DataAccess)(implicit executionContext: ExecutionContext): ReadWriteAction[Boolean] = {
+  def updateSubmissionStatus(dataAccess: DataAccess)(implicit executionContext: ExecutionContext): ReadWriteAction[Boolean] = {
     dataAccess.workflowQuery.listWorkflowRecsForSubmissionAndStatuses(submissionId, (WorkflowStatuses.queuedStatuses ++ WorkflowStatuses.runningStatuses):_*) flatMap { workflowRecs =>
       if (workflowRecs.isEmpty) {
         dataAccess.submissionQuery.findById(submissionId).map(_.status).result.head.map { status =>
```
jmthibault79, line 324: not your code but I'd prefer `uniqueResult` to `result.head` here

jmthibault79 (changes requested):
The test fails locally for me (timeout) even after increasing the value to 20 sec

jmthibault79 (approved):

jmthibault79:
It works when I set timeout to 30 sec 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
helgridly: if the submission record has vanished we are in deep doo-doo

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: i think this should just be a case of subclassing `OneForOneStrategy` and overriding `processFailure` to also log an error when it's above a given number of retries.

jmthibault79:
Added a few commits relating to `SubmissionSupervisor` and `ThresholdOneForOneStrategy`

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -222,7 +222,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
             _ <- abortActiveWorkflows(submissionId)
           } yield {}
         } else {
-          Future.successful()
+          Future.successful(Unit)
```
helgridly: this wants to be `Future.successful(())`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -335,7 +335,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       // Remove Google Groups
       _ <- Future.traverse(groupsToRemove) {
         case Some(group) => gcsDAO.deleteGoogleGroup(group)
-        case None => Future.successful(())
+        case None => Future.successful(Unit)
```
helgridly: but these should be left alone

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
```diff
@@ -1,16 +1,405 @@
...
+  }
+
+  it should "execute the threshold function when threshold < retries" in {
+    implicit val timeout = Timeout(5.seconds)
+
+    val varInit = 1000
+    var testVar: Int = varInit
+
+    // threshold function
+    def setVar(cause: Throwable, v: Int): Unit = {
```
helgridly, line 133: i see. cool test

helgridly (commented):
👍 delta annoying fixes to `Unit`. Thank you so much for picking this up.

jmthibault79:
@dvoet and @rtitle please review my additions

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -214,16 +216,15 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     submissionFuture flatMap {
       case Some(submission) =>
-        if(submission.status == SubmissionStatuses.Aborting) {
+        val abortFuture = if(submission.status == SubmissionStatuses.Aborting) {
           for {
```
rtitle, line 220: So, the current code will call `abortQueuedWorkflows` and `abortActiveWorkflows` in sequence (chained via callback).

However since these two methods are operating on mutually exclusive data (i.e. Queued workflows and Running workflows), I question whether they could be run in parallel instead. See:
```
        val abortFuture = if(submission.status == SubmissionStatuses.Aborting) {
          Future.sequence(Seq(abortQueuedWorkflows(submissionId), abortActiveWorkflows(submissionId))).mapTo[Unit]
        } else {
          Future.successful(())
        }
        abortFuture flatMap( _ => queryForWorkflowStatuses() )
```

Thoughts?

helgridly: hah, cute. sure, why not!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -49,11 +50,17 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
-      }
+  // restart the actor on failure (e.g. a DB deadlock or failed transaction)
+  // if this actor has failed more than 3 times, log each new failure
+  override val supervisorStrategy = {
+    val alwaysRestart: SupervisorStrategy.Decider = {
+      case _ => Restart
     }
+
+    def thresholdFunc(cause: Throwable, count: Int): Unit = {
+      system.log.error(cause, s"error monitoring submission: SubmissionMonitorActor has been restarted $count times")
```
rtitle: Is `system.log` correct here? Elsewhere we mix in `LazyLogging` and use slf4j.

rtitle: Also maybe log the exception too

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -477,6 +478,43 @@ class SubmissionMonitorSpec(_system: ActorSystem) extends TestKit(_system) with
     }
   }
 
+  it should "attach outputs and not deadlock with multiple submissions all updating the same entity at once" in withDefaultTestDatabase { dataSource: SlickDataSource =>
+    //create a bunch of submissions all running on the same two entities
+    val numSubmissions = 50
```
rtitle: I used `withCustomTestDatabase` recently to do something similar:

https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala#L57

Might be cleaner?

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -477,6 +478,43 @@ class SubmissionMonitorSpec(_system: ActorSystem) extends TestKit(_system) with
...
+        runAndWait(submissionQuery.create(ctx, testSub))
+        runAndWait(updateWorkflowExecutionServiceKey("unittestdefault"))
+
+        testSub
+      }
+
+      submissions.foreach ( sub => createSubmissionMonitorActor(dataSource, sub, new SubmissionTestExecutionServiceDAO(WorkflowStatuses.Succeeded.toString)) )
+
+      //they're all being monitored. they should all complete just fine, without deadlocking or otherwise barfing
+      awaitCond({
```
rtitle: (sorry for the nitpick) `awaitCond({ ... })` can just be `awaitCond { ... }`

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
```diff
@@ -0,0 +1,405 @@
...
+  // This configures how long the calls to `whenReady(Future)` and `eventually` will wait
+  // before giving up and failing the test.
+  // See: http://doc.scalatest.org/2.2.4/index.html#org.scalatest.concurrent.Futures
+  implicit override val patienceConfig = PatienceConfig(timeout = scaled(Span(10, Seconds)))
+
+  val alwaysRestart: PartialFunction[Throwable, Directive] = {
+    case _ => Restart
+  }
+
+  object TestException extends Exception("boom")
```
rtitle, line 26: 💥

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
```diff
@@ -0,0 +1,405 @@
...
+    def setVar(cause: Throwable, v: Int): Unit = {
+      assertResult(TestException)(cause)
+      testVar = v
+    }
+
+    val strategy = new ThresholdOneForOneStrategy(thresholdLimit = 3, maxNrOfRetries = Option(1))(alwaysRestart)(setVar)
+    val supervisor: ActorRef = system.actorOf(Props(new SupervisorTestActor(strategy)))
+
+    // retrieve a new supervised child
+    val childRef: ActorRef = whenReady(supervisor ? "newchild") {
```
rtitle: Maybe
```
val childRef: ActorRef = (supervisor ? "newchild").mapTo[ActorRef].futureValue
```

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
```diff
@@ -0,0 +1,405 @@
...
+      _.asInstanceOf[ActorRef]
+    }
+
+    // start count = 0
+    // increment a few times and retrieve new counts
+
+    whenReady(childRef ? "increment") { resp =>
```
rtitle, line 77: Just a note: akka TestKit has ways of testing actors in a synchronous manner, so you don't need to deal with futures in tests:

http://doc.akka.io/docs/akka/2.5.3/scala/testing.html#synchronous-testing-testactorref

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
```diff
@@ -0,0 +1,405 @@
...
+      resp.asInstanceOf[Int] shouldEqual 1
```
rtitle: Maybe add a `resp shouldBe an [Int]` before blindly casting :)

http://www.scalatest.org/user_guide/using_matchers#checkingAnObjectsClass

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
rtitle: +1

rtitle:
A few comments above - overall looks good! The new logic in `SubmissionMonitorActor` makes sense to me.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: the exception is logged as part of the log call (but will change this to LazyLogging)

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
helgridly: good call. also i just discovered it's way terser to extend `EmptyWorkspace` instead of `TestData` when you're doing this

util/src/test/scala/org/broadinstitute/dsde/rawls/util/ThresholdOneForOneStrategySpec.scala
jmthibault79: Nice.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
helgridly: this change works on my machine but travis hates it

PR #732, opened 2017-07-14
GAWB-2143 and GAWB-2144: instrument all Spray requests

jmthibault79:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
@@ -1,0 +1,16 @@
+package org.broadinstitute.dsde.rawls.metrics
+
+import spray.routing.Directive0
+import spray.routing.directives.BasicDirectives.mapHttpResponse
+import spray.routing.directives.MiscDirectives.requestInstance
+
+trait InstrumentationDirectives extends RawlsInstrumented {
+  def instrumentRequest: Directive0 = requestInstance flatMap { request =>
+    val timeStamp = System.currentTimeMillis
+    mapHttpResponse { response =>
+      httpRequestCounter(ExpandedMetricBuilder.empty)(request, response).inc()
+      httpRequestTimer(ExpandedMetricBuilder.empty)(request, response).time(System.currentTimeMillis - timeStamp)
+      response
+    }
+  }
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
@@ -2,11 +2,13 @@ package org.broadinstitute.dsde.rawls.metrics
 
 import java.util.UUID
 
+import nl.grons.metrics.scala.{Counter, Timer}
 import nl.grons.metrics.scala.Counter
 import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
 import org.broadinstitute.dsde.rawls.model.WorkspaceName
 import slick.dbio.{DBIOAction, Effect, NoStream}
+import spray.http.{HttpRequest, HttpResponse}
 
 import scala.concurrent.ExecutionContext
 
@@ -19,6 +21,9 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
   final val SubmissionMetricKey = "submission"
   final val SubmissionStatusMetricKey = "submissionStatus"
   final val WorkflowStatusMetricKey = "workflowStatus"
+  final val HttpRequestMethodMetricKey = "httpRequestMethod"
+  final val HttpRequestUriMetricKey = "httpRequestUri"
+  final val HttpResponseStatusCodeMetricKey = "httpResponseStatusCode"
 
   /**
     * An ExpandedMetricBuilder for a WorkspaceName and a submission ID.
@@ -45,6 +50,19 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
     status => builder
       .expand(WorkflowStatusMetricKey, status)
       .asCounter("count")
+
+  protected def httpRequestMetric(builder: ExpandedMetricBuilder): (HttpRequest, HttpResponse) => ExpandedMetricBuilder = {
+     (httpRequest, httpResponse) => builder
+      .expand(HttpRequestMethodMetricKey, httpRequest.method)
+      .expand(HttpRequestUriMetricKey, httpRequest.uri)
+      .expand(HttpResponseStatusCodeMetricKey, httpResponse.status)
+  }
+
+  protected def httpRequestCounter(builder: ExpandedMetricBuilder): (HttpRequest, HttpResponse) => Counter =
+    httpRequestMetric(builder)(_, _).asCounter("request")
+
+  protected def httpRequestTimer(builder: ExpandedMetricBuilder): (HttpRequest, HttpResponse) => Timer =
+    httpRequestMetric(builder)(_, _).asTimer("latency")
 }
 
 object RawlsInstrumented {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.rawls.webservice
 
 import akka.actor.{Actor, Props}
 import org.broadinstitute.dsde.rawls.genomics.GenomicsService
+import org.broadinstitute.dsde.rawls.metrics.InstrumentationDirectives
 import org.broadinstitute.dsde.rawls.model.{ApplicationVersion, UserInfo}
 import org.broadinstitute.dsde.rawls.openam.StandardUserInfoDirectives
 import org.broadinstitute.dsde.rawls.statistics.StatisticsService
@@ -20,26 +21,28 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 
 object RawlsApiServiceActor {
-  def props(workspaceServiceConstructor: UserInfo => WorkspaceService, userServiceConstructor: UserInfo => UserService, genomicsServiceConstructor: UserInfo => GenomicsService, statisticsServiceConstructor: UserInfo => StatisticsService, statusServiceConstructor: () => StatusService, appVersion: ApplicationVersion, googleClientId: String, submissionTimeout: FiniteDuration)(implicit executionContext: ExecutionContext): Props = {
-    Props(new RawlsApiServiceActor(workspaceServiceConstructor, userServiceConstructor, genomicsServiceConstructor, statisticsServiceConstructor, statusServiceConstructor, appVersion, googleClientId, submissionTimeout))
+  def props(workspaceServiceConstructor: UserInfo => WorkspaceService, userServiceConstructor: UserInfo => UserService, genomicsServiceConstructor: UserInfo => GenomicsService, statisticsServiceConstructor: UserInfo => StatisticsService, statusServiceConstructor: () => StatusService, appVersion: ApplicationVersion, googleClientId: String, submissionTimeout: FiniteDuration, rawlsMetricBaseName: String)(implicit executionContext: ExecutionContext): Props = {
+    Props(new RawlsApiServiceActor(workspaceServiceConstructor, userServiceConstructor, genomicsServiceConstructor, statisticsServiceConstructor, statusServiceConstructor, appVersion, googleClientId, submissionTimeout, rawlsMetricBaseName))
   }
 }
 
-class RawlsApiServiceActor(val workspaceServiceConstructor: UserInfo => WorkspaceService, val userServiceConstructor: UserInfo => UserService, val genomicsServiceConstructor: UserInfo => GenomicsService, val statisticsServiceConstructor: UserInfo => StatisticsService, val statusServiceConstructor: () => StatusService, val appVersion: ApplicationVersion, val googleClientId: String, val submissionTimeout: FiniteDuration)(implicit val executionContext: ExecutionContext) extends Actor
+class RawlsApiServiceActor(val workspaceServiceConstructor: UserInfo => WorkspaceService, val userServiceConstructor: UserInfo => UserService, val genomicsServiceConstructor: UserInfo => GenomicsService, val statisticsServiceConstructor: UserInfo => StatisticsService, val statusServiceConstructor: () => StatusService, val appVersion: ApplicationVersion, val googleClientId: String, val submissionTimeout: FiniteDuration, override val workbenchMetricBaseName: String)(implicit val executionContext: ExecutionContext) extends Actor
   with RootRawlsApiService with WorkspaceApiService with EntityApiService with MethodConfigApiService with SubmissionApiService
   with AdminApiService with UserApiService with StandardUserInfoDirectives with BillingApiService with NotificationsApiService
-  with StatusApiService {
+  with StatusApiService with InstrumentationDirectives {
 
   def actorRefFactory = context
   def apiRoutes = options{ complete(OK) } ~ baseRoute ~ workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutes ~ notificationsRoutes
   def registerRoutes = options{ complete(OK) } ~ createUserRoute ~ getUserStatusRoute
 
   def receive = runRoute(
-    swaggerRoute ~
-    versionRoute ~
-    statusRoute ~
-    pathPrefix("api") { apiRoutes } ~
-    pathPrefix("register") { registerRoutes }
+    instrumentRequest {
+      swaggerRoute ~
+      versionRoute ~
+      statusRoute ~
+      pathPrefix("api") { apiRoutes } ~
+      pathPrefix("register") { registerRoutes }
+    }
   )
 }
 
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
@@ -16,6 +16,7 @@ import org.broadinstitute.dsde.rawls.user.UserService
 import spray.http._
 import spray.json.DefaultJsonProtocol._
 import spray.json._
+import spray.routing.Route
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, ExecutionContext}
@@ -92,12 +93,21 @@ class AdminApiServiceSpec extends ApiServiceSpec {
       ActiveSubmission(constantData.workspace.namespace, constantData.workspace.name, constantData.submission1),
       ActiveSubmission(constantData.workspace.namespace, constantData.workspace.name, constantData.submission2))
 
-    Get(s"/admin/submissions") ~>
-      sealRoute(services.adminRoutes) ~>
-      check {
-        assertResult(StatusCodes.OK) { status }
-        assertSameElements(expected, responseAs[Seq[ActiveSubmission]])
+    withStatsD {
+      Get("/admin/submissions") ~>
+        sealRoute(instrumentRequest { services.adminRoutes }) ~>
+        check {
+          assertResult(StatusCodes.OK) {
+            status
+          }
+          assertSameElements(expected, responseAs[Seq[ActiveSubmission]])
+        }
+    } { capturedMetrics =>
+      val expected = expectedHttpRequestMetrics("get", "admin.submissions", StatusCodes.OK.intValue, 1)
+      assert {
+        expected subsetOf capturedMetrics.toSet
       }
+    }
   }
 
   it should "return 200 when listing active submissions on deleted entities" in withConstantTestDataApiServices { services =>
@@ -282,21 +292,27 @@ class AdminApiServiceSpec extends ApiServiceSpec {
 
     import UserAuthJsonSupport.CreateRawlsBillingProjectFullRequestFormat
 
-    Post(s"/billing", httpJson(createRequest)) ~>
-      sealRoute(services.billingRoutes) ~>
-      check {
-        assertResult(StatusCodes.Created) {
-          status
+    withStatsD {
+      Post(s"/billing", httpJson(createRequest)) ~> services.sealedInstrumentedRoutes ~>
+        check {
+          assertResult(StatusCodes.Created) {
+            status
+          }
         }
-      }
 
-    Delete(s"/admin/billing/${projectName.value}/user/nobody") ~>
-      sealRoute(services.adminRoutes) ~>
-      check {
-        assertResult(StatusCodes.NotFound) {
-          status
+      Delete(s"/admin/billing/${projectName.value}/user/nobody") ~> services.sealedInstrumentedRoutes ~>
+        check {
+          assertResult(StatusCodes.NotFound) {
+            status
+          }
         }
+    } { capturedMetrics =>
+      val expected = expectedHttpRequestMetrics("post", "billing", StatusCodes.Created.intValue, 1) ++
+        expectedHttpRequestMetrics("delete", s"admin.billing.${projectName.value}.user.nobody", StatusCodes.NotFound.intValue, 1)
+      assert {
+        expected subsetOf capturedMetrics.toSet
       }
+    }
   }
 
   it should "return 404 when removing a user from a nonexistent billing project" in withTestDataApiServices { services =>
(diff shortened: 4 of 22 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsde.rawls.metrics
+
+import spray.routing.Directive0
+import spray.routing.directives.MiscDirectives.requestInstance
+import spray.routing.directives.BasicDirectives.pass
+
+trait InstrumentationDirectives extends RawlsInstrumented {
+
+  // TODO can I add timing here (Anu's story) - google "scala spray request timing"
```
jmthibault79: Anu has now done this on her branch.  not sure if we should merge

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  // TODO: "expand nothing" is ugly but the base name will get added regardless so I don't want to do it twice
+  lazy val builder: ExpandedMetricBuilder = ExpandedMetricBuilder.expand("")
+
+  def instrumentRequest: Directive0 = requestInstance flatMap { request =>
+    httpRequestCounter(builder)(request).inc()
```
jmthibault79: my pref. would be a version of `httpRequestCounter` that doesn't require an existing `ExpandedMetricBuilder `

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -3,10 +3,15 @@ package org.broadinstitute.dsde.rawls.metrics
 import java.util.UUID
 
 import nl.grons.metrics.scala._
+import org.apache.http.HttpStatus
```
jmthibault79: This file is copied from Rob's branch (which is coming later) so comments here may be moot.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -272,17 +281,25 @@ class AdminApiServiceSpec extends ApiServiceSpec {
...
-      check {
-        Delete(s"/admin/billing/${project.projectName.value}/nobody") ~>
-          sealRoute(services.adminRoutes) ~>
-          check {
-            assertResult(StatusCodes.NotFound) {
-              status
+    val putUri: Uri = s"/admin/billing/register/${project.projectName.value}"
+    val delUri: Uri = s"/admin/billing/${project.projectName.value}/nobody"
+
+    val sealedInstrumentedAdminRoutes: Route = sealRoute(instrumentRequest { services.adminRoutes })
```
jmthibault79: I'm thinking about adding something like this, maybe to `ApiServices`:

```
val allRoutesInstrumented = sealRoute(instrumentRequest {
  services.adminRoutes ~ services.billingRoutes   // etc
}
```

This would be similar to what we're doing in `RawlsApiServiceActor`.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -92,12 +93,20 @@ class AdminApiServiceSpec extends ApiServiceSpec {
...
+      Get(uri) ~>
+        sealRoute(instrumentRequest { services.adminRoutes }) ~>
+        check {
+          assertResult(StatusCodes.OK) {
+            status
+          }
+          assertSameElements(expected, responseAs[Seq[ActiveSubmission]])
+        }
+    } { capturedMetrics =>
+      capturedMetrics should contain (expectedHttpRequestCounterMetric(HttpMethods.GET, uri, 1))
```
jmthibault79: Adding metrics checks to a few tests per file, arbitrarily.

If you like these, I'll add a few more per file to the PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
jmthibault79: ... and now I have added one

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -134,6 +179,22 @@ trait RawlsInstrumented extends DefaultInstrumented {
     status => builder
       .expand(WorkflowStatusMetric, status)
       .asCounter()
+
+  protected def httpRequestCounter(currentBuilder: Option[ExpandedMetricBuilder] = None): HttpRequest => Counter = {
```
jmthibault79: I modified this in order to call it without an existing builder.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -37,26 +43,48 @@ trait RawlsInstrumented extends DefaultInstrumented {
     */
   @implicitNotFound(msg = "Cannot expand instances of type ${A}")
   protected sealed trait Expansion[A] {
-    def makeName(key: String, a: A) = s"$key.${a.toString}"
+    def makeName(key: Option[String], a: A) = s"${getKey(key)}${a.toString}"
```
rtitle: I know this is copy-pasted code, but recently I'm thinking we might want to make `key` non-optional. The reason is that we need to specify a key-value pair when generating stackdriver labels, so we should probably enforce a key in scala code.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -37,26 +43,48 @@ trait RawlsInstrumented extends DefaultInstrumented {
...
-    override def makeName(key: String, n: WorkspaceName): String = s"$key.${n.toString.replace('/', '.')}"
+    override def makeName(key: Option[String], n: WorkspaceName): String = s"${getKey(key)}${n.toString.replace('/', '.')}"
   }
 
   // Provide an implicit for UUIDs but can use the default makeName
   protected implicit object UUIDExpansion extends Expansion[UUID]
 
-  // Provide implicits for WorkflowStatus and SubmissionStatus using the default makeName.
+  // Provide implicits for various RawlsEnumerations using the default makeName.
+  // TODO: should we provide an expansion for RawlsEnumeration in general?
```
rtitle: I went ahead and made an `Expansion` for `T <: RawlsEnumeration` in my branch.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -37,26 +43,48 @@ trait RawlsInstrumented extends DefaultInstrumented {
...
+  protected implicit object HttpMethodExpansion extends Expansion[HttpMethod] {
+    override def makeName(key: Option[String], m: HttpMethod): String = s"${getKey(key)}${m.toString.toLowerCase}"
+  }
+
+  protected implicit object UriExpansion extends Expansion[Uri] {
+    override def makeName(key: Option[String], uri: Uri): String = s"${getKey(key)}${uri2String(uri)}"
+
+    private def uri2String(uri: Uri): String = {
+      val path = if (uri.path.startsWithSlash) uri.path.tail.toString else uri.path
+      path.toString.replace('/', '_')
```
rtitle: I also changed the `_` to `.`

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -20,26 +21,28 @@ import scala.concurrent.ExecutionContext
...
   }
 }
 
-class RawlsApiServiceActor(val workspaceServiceConstructor: UserInfo => WorkspaceService, val userServiceConstructor: UserInfo => UserService, val genomicsServiceConstructor: UserInfo => GenomicsService, val statisticsServiceConstructor: UserInfo => StatisticsService, val statusServiceConstructor: () => StatusService, val appVersion: ApplicationVersion, val googleClientId: String, val submissionTimeout: FiniteDuration)(implicit val executionContext: ExecutionContext) extends Actor
+class RawlsApiServiceActor(val workspaceServiceConstructor: UserInfo => WorkspaceService, val userServiceConstructor: UserInfo => UserService, val genomicsServiceConstructor: UserInfo => GenomicsService, val statisticsServiceConstructor: UserInfo => StatisticsService, val statusServiceConstructor: () => StatusService, val appVersion: ApplicationVersion, val googleClientId: String, val submissionTimeout: FiniteDuration, val rawlsMetricBaseName: String)(implicit val executionContext: ExecutionContext) extends Actor
```
rtitle: Minor, `override val rawlsMetricBaseName: String`?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -20,26 +21,28 @@ import scala.concurrent.ExecutionContext
...
 
   def receive = runRoute(
-    swaggerRoute ~
-    versionRoute ~
-    statusRoute ~
-    pathPrefix("api") { apiRoutes } ~
-    pathPrefix("register") { registerRoutes }
+    instrumentRequest {
+      swaggerRoute ~
+        versionRoute ~
```
rtitle: Indentation looks off here

rtitle: Also should we instrument serving the swagger index and status/version checks? Those are likely to be pretty uninteresting metrics but I guess there is not much downside...

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
jmthibault79: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: OK, makes sense. For my subsystem metrics I am passing in `httpRequestCounter(ExpandedMetricBuilder.expand(SubsystemMetricKey, Subsystems.Cromwell))` (for example) but it makese sense for it to be optional for your use case.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
jmthibault79: I can add this but I don't understand what it does

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
jmthibault79: I figure it doesn't hurt to capture everything.

IntelliJ made this ugly so I'll force-indent.

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
```diff
@@ -59,4 +59,15 @@ trait StatsDTestUtils { this: MockitoSugar with Eventually with RawlsTestUtils =
 
   protected def expectedSubmissionStatusMetric(workspaceName: WorkspaceName, submissionStatus: SubmissionStatus, expectedTimes: Int): (String, String) =
     (s"test.workspace.${workspaceName.toString.replace('/', '.')}.submissionStatus.${submissionStatus.toString}", expectedTimes.toString)
+
+  // TODO don't repeat the code from HttpMethodExpansion.makeName() / UriExpansion.makeName()
```
rtitle: Yeah. :) Maybe `expectedHttpRequestCounterMetric` could just take the expected URI path as a `String`? Then at least we're comparing things directly.

Also I'm wondering whether this passes since above I saw `path.toString.replace('/', '_')`..

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
helgridly: so i understand: is this because stackdriver goes looking for the text `thing.thing_id` in the metric?

jmthibault79: ok - I was just ignoring keys because I didn't realize they were important

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
rtitle: Looks reasonable - though I'm wondering about just passing Strings directly like
```
expectedHttpRequestCounterMetric("get.admin.submissions.request", 1)
```
to make it a bit more clear the metric format we're verifying.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
helgridly: Is changing `_` to `.` going to be confusing? Treating `/` as a separate metric seems like a good idea but I'm less convinced by `_`.

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
jmthibault79: But I have to do the `uri2str()` logic somewhere.  Might as well centralize it here.

It passes because my `UriExpansion` converts `.` to `_` but now I will replace both.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
rtitle: So then would the API tests use `allRoutesInstrumented`? Currently the tests only use the routes for the actual API service they're testing. I wonder if adding all the routes for every test would slow down the tests. :)

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -301,37 +318,44 @@ class AdminApiServiceSpec extends ApiServiceSpec {
...
-        }
-        assertResult(Set.empty) {
-          responseAs[Seq[RawlsBillingProjectName]].toSet
-        }
-      }
+    val getUri: Uri = s"/admin/billing/list/${testUser.userEmail.value}"
+    val putUri: Uri = s"/admin/billing/${project1.projectName.value}/user/${testUser.userEmail.value}"
 
-    createBillingProject(project1)
+    val sealedInstrumentedAdminRoutes: Route = sealRoute(instrumentRequest { services.adminRoutes })
```
rtitle: Could at least share `sealedInstrumentedAdminRoutes` between tests :)

jmthibault79: Yes.  I have been considering adding a master instrumented route to be shared by *all* tests.

```
val allRoutesInstrumented = sealRoute(instrumentRequest {
  services.adminRoutes ~ services.billingRoutes   // etc
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: So when we create stackdriver labels we need to specify a `"label:key" "value"`. This gets translated as a little drop-down in the SD UI (here "instance_id" is the key):

image

What I'm doing in my script is take a list of "known" keys (specified in a config in firecloud-develop) and parse out the values following those keys. So `workspace.<ws-id>.submission.<sub-id>` would be translated as 
```
label:workspace ws-id
label:submission sub-id
```
and we get the nice drop-down menus in SD.

I never said it was the most elegant thing in the world but that's how it's currently working...

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
helgridly: +1 for readable strings in tests

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: It's confusing in that it gives you several sub-metrics for the "uri-path" key. It doesn't look like SD supports sub-labels. What I'm trying is just enumerating them. E.g. given a path `/foo/bar/baz` I'm generating
```
label:uri-path-0 "foo"
label:uri-path-1 "bar"
label:uri-path-2 "baz"
```
Then users will see a drop-down for each path component where they can select "Any" or a specific value. I can see how that might be useful functionality, just need to make sure it's not too confusing.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
rtitle: It just indicates you're overriding the val from a base class, like Java `@Override`. I think it's a compiler warning otherwise?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
rtitle: git indent --force

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
rtitle: I guess I'm advocating hard-coding strings in tests to verify against the actual `uri2String()` logic.

rtitle:
In general this looks good, nitpicky comments aside. 

I have done quite a bit of refactoring in my dev branch, most notably:
- Moved `RawlsInstrumented` to a new `rawls/metrics` subproject
- Renamed `RawlsInstrumented` to `WorkbenchInstrumented` (since it may be used outside of Rawls now that it's in a common location)
- Split up `WorkbenchInstrumented` and the `Expansion` typeclass into separate files
- Corresponding unit test changes

This doesn't affect your work per se, but it will cause some conflicts. Perhaps the best thing to do still is:
1. Merge this PR to dev once ready & tested
2. I'll rebase my branch from dev and resolve conflicts
3. I'll open a PR from my branch once ready

Let me know if you disagree.

core/src/test/scala/org/broadinstitute/dsde/rawls/StatsDTestUtils.scala
jmthibault79: OK I can do that

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
jmthibault79: hm ok.  So for requests we could do something like
```
label:requestMethod get
label:requestUri workspace.<ws-id>.submission.<sub-id>
```

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
jmthibault79: done

jmthibault79:
Thanks for being flexible.  Yes I think this plan is best.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -0,0 +1,18 @@
...
+import spray.routing.directives.BasicDirectives._
+import spray.routing.directives.MiscDirectives.requestInstance
+
+trait InstrumentationDirectives extends RawlsInstrumented {
+  def instrumentRequest: Directive0 = requestInstance flatMap { request =>
+    httpRequestCounter(request).inc()
+    mapRequestContext { context =>
+      val timeStamp = System.currentTimeMillis
+      context.withHttpResponseMapped { response =>
+        val rTime = httpRequestTimer(request).time(System.currentTimeMillis - timeStamp)
```
rtitle: It would be good to count the response (i.e. status code) here as well. Could we add a:
```
httpResponseCounter(response).inc()
```
?

Futuremore, maybe `httpResponseCounter` should take in the `HttpRequest` as well, so we can count (method, uri, response code) triples. We can modify `httpResponseCounter` to take in a `HttpRequest` like so:
```
httpResponseCounter(request, response).inc()
```
This will require a change in `RawlsInstrumented`.

rtitle: Also (minor), the `val rTime` is unnecessary.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: 👍

jmthibault79:
Added response Status Codes and keys to metrics, and made Counter/Timer more generalized

TODO:
* more tests

jmthibault79:
This PR has been rebased and updated for Rob's refactoring.  Please re-review @rtitle @helgridly

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -92,12 +93,21 @@ class AdminApiServiceSpec extends ApiServiceSpec {
...
+        sealRoute(instrumentRequest { services.adminRoutes }) ~>
+        check {
+          assertResult(StatusCodes.OK) {
+            status
+          }
+          assertSameElements(expected, responseAs[Seq[ActiveSubmission]])
+        }
+    } { capturedMetrics =>
+      val expected = expectedHttpRequestMetrics("get", "admin.submissions", StatusCodes.OK.intValue, 1)
+      assert {
```
rtitle: In general I like using Scalatest matchers instead of `assert(boolean)` because it gives nicer error messages. Unfortunately it looks like our current version of scalatest doesn't have a great way of verifying subsets (it was added in 3.0, see https://groups.google.com/forum/#!topic/scalatest-users/SyyW8RC1pfA).

That said, I would probably still prefer something like this:
```
val capturedSet = capturedMetrics.toSet
expectedHttpRequestMetrics("get", "admin.submissions", StatusCodes.OK.intValue, 1).foreach { m =>
  capturedSet should contain (m)
}
```
Then at least you would see _which_ expected metric was not contained in the expected set in the error message.

metrics/src/main/scala/org/broadinstitute/dsde/rawls/metrics/Expansion.scala
```diff
@@ -44,6 +45,34 @@ object Expansion {
     */
   implicit def RawlsEnumerationExpansion[A <: RawlsEnumeration[_]] = new Expansion[A] {}
 
+  /**
+    * Implicit expansion for HttpMethod.
+    */
+  implicit object HttpMethodExpansion extends Expansion[HttpMethod] {
```
rtitle, line 51: Can you add cases to `ExpansionSpec` for the new expansions for `HttpMethod`, `Uri`, and `HttpStatus`?

rtitle:
Looks good! Just a couple requested changes, and just in test code.

rtitle (changes requested):
Looks good! Just a couple requested changes, and just in test code.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
jmthibault79: OK, I added `RawlsTestUtils.assertSubsetOf()` which uses `should contain` internally

metrics/src/main/scala/org/broadinstitute/dsde/rawls/metrics/Expansion.scala
jmthibault79: done

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -45,6 +50,19 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
     status => builder
       .expand(WorkflowStatusMetricKey, status)
       .asCounter("count")
+
+  protected def httpRequestMetric(builder: ExpandedMetricBuilder): (HttpRequest, HttpResponse) => ExpandedMetricBuilder = {
```
helgridly, line 54: functions that RETURN functions. look how far we've come

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -20,26 +21,28 @@ import scala.concurrent.ExecutionContext
...
   def apiRoutes = options{ complete(OK) } ~ baseRoute ~ workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutes ~ notificationsRoutes
   def registerRoutes = options{ complete(OK) } ~ createUserRoute ~ getUserStatusRoute
 
   def receive = runRoute(
-    swaggerRoute ~
-    versionRoute ~
-    statusRoute ~
-    pathPrefix("api") { apiRoutes } ~
-    pathPrefix("register") { registerRoutes }
+    instrumentRequest {
```
helgridly, line 39: this is shit hot

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -282,21 +290,25 @@ class AdminApiServiceSpec extends ApiServiceSpec {
 
     import UserAuthJsonSupport.CreateRawlsBillingProjectFullRequestFormat
 
-    Post(s"/billing", httpJson(createRequest)) ~>
-      sealRoute(services.billingRoutes) ~>
-      check {
-        assertResult(StatusCodes.Created) {
-          status
+    withStatsD {
+      Post(s"/billing", httpJson(createRequest)) ~> services.sealedInstrumentedRoutes ~>
```
helgridly, line 294: presumably nothing is lost here by posting this to all the routes combined rather than just the billing routes

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
rtitle: this is going to be convenient to use with `spray.client.pipelining` as well :)

rtitle (approved):
:shipit:

helgridly (approved):
First of all: this is excellent work. I really like how it just becomes a call to `instrumentRequest` where the routes are concatenated. Very elegant! 

I am skeptical of the need to check that every request we make during tests is instrumented. The `instrumentRequest` directive covers that for everything so checking on every request we do seems like overkill. If this functionality breaks, a zillion tests will fail at once. We only need one test failure to be alerted of the issue; the rest is unnecessary noise.

I'm open to hearing arguments to the contrary but I don't think we need to make "check that the count of your requests is 1" is a requirement for all future `*ApiServiceSpec` tests.

(But I'm not going to say "go delete it all!".)

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
jmthibault79: It's only for the instrumentation tests, and also it's closer to how Rawls actually processes routes than isolation.

jmthibault79:
The idea behind making so many tests was to have coverage of different kinds of routes, so we chose a few arbitrarily from each `ApiSpec`.  The majority of tests remain uninstrumented.

PR #740, opened 2017-07-26
GAWB-2367 Moving workbench-util to broadinstitute/workbench-libs

helgridly:
This PR moves the `workbench-util` package from Rawls to the new repo [broadinstitute/workbench-libs](github.com/broadinstitute/workbench-libs).

I've moved this to PR against a snapshot branch of develop at PR time called `with_workbenchlib`, so that if we return to this project in The Mysterious Future we'll be able to see the changes before a giant rebase / merge conflict.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -94,6 +95,18 @@ class HealthMonitor private (val slickDataSource: SlickDataSource, val googleSer
...
+    * Zero is an ok status with no messages.
+    * Append uses && on the ok flag, and ++ on the messages.
+    */
+  private implicit val SubsystemStatusMonoid = new Monoid[SubsystemStatus] {
+    def combine(a: SubsystemStatus, b: SubsystemStatus): SubsystemStatus = {
+      SubsystemStatus(a.ok && b.ok, a.messages |+| b.messages)
+    }
+    def empty: SubsystemStatus = OkStatus
+  }
+
```
helgridly: I moved this up because scalac was complaining that it couldn't find the implicit later on.

project/Dependencies.scala
```diff
@@ -56,10 +56,16 @@ object Dependencies {
...
-  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.7.22" % "test"
+  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.0.1" % "test"
+  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.8.47" % "test"
   val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "3.9.2" % "test"
 
+  def excludeAkkaActor(m: ModuleID): ModuleID = m.exclude("com.typesafe.akka", "akka-actor")
+
+  //decouple akka dependency from workbench-libs
+  val workbenchUtil: ModuleID = excludeAkkaActor("org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.1-02148b1-SNAP")
+  val workbenchUtilTest: ModuleID = excludeAkkaActor((workbenchUtil % Test).classifier("tests"))
```
helgridly: This is really unintuitive but is required to make available classes in the `test` director of `workbench-util`. We could avoid this by moving `MockitoTestUtils` out from `test` to `src` in `workbench-util`; I think that would be a better solution.

project/Dependencies.scala
```diff
@@ -56,10 +56,16 @@ object Dependencies {
   val mysqlConnector: ModuleID =  "mysql"                         % "mysql-connector-java"  % "5.1.38"
   val liquibaseCore: ModuleID =   "org.liquibase"                 % "liquibase-core"        % "3.5.3"
   val logbackClassic: ModuleID =  "ch.qos.logback"                % "logback-classic"       % "1.1.6"
-  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "2.2.4" % "test"
-  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.7.22" % "test"
+  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.0.1" % "test"
+  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.8.47" % "test"
```
helgridly, line 60: bumped these to be in line with workbench-libs as `MockitoTestUtils` was complaining about an incompatibility with `MockitoSugar` (they moved it between versions).

project/Dependencies.scala
```diff
@@ -56,10 +56,16 @@ object Dependencies {
...
   val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "3.9.2" % "test"
 
+  def excludeAkkaActor(m: ModuleID): ModuleID = m.exclude("com.typesafe.akka", "akka-actor")
+
+  //decouple akka dependency from workbench-libs
+  val workbenchUtil: ModuleID = excludeAkkaActor("org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.1-02148b1-SNAP")
```
helgridly: This needs un`-SNAP`-ing once https://github.com/broadinstitute/workbench-libs/pull/6 has been merged.

project/Dependencies.scala
```diff
@@ -56,10 +56,16 @@ object Dependencies {
...
-  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.7.22" % "test"
+  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.0.1" % "test"
+  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.8.47" % "test"
   val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "3.9.2" % "test"
 
+  def excludeAkkaActor(m: ModuleID): ModuleID = m.exclude("com.typesafe.akka", "akka-actor")
+
+  //decouple akka dependency from workbench-libs
+  val workbenchUtil: ModuleID = excludeAkkaActor("org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.1-e8bdfd0")
+  val workbenchUtilTest: ModuleID = excludeAkkaActor((workbenchUtil % Test).classifier("tests"))
```
helgridly, line 67: This is really unintuitive but is required to make available classes in the `test` director of `workbench-util`. We could avoid this by moving `MockitoTestUtils` out from `test` to `src` in `workbench-util`; I think that would be a better solution.

metrics/src/test/scala/org/broadinstitute/dsde/rawls/metrics/StatsDTestUtils.scala
```diff
@@ -4,7 +4,8 @@ import java.util.concurrent.TimeUnit
 
 import com.codahale.metrics.{MetricFilter, SharedMetricRegistries}
 import com.readytalk.metrics.{StatsD, StatsDReporter}
-import org.broadinstitute.dsde.rawls.util.MockitoTestUtils
+import org.broadinstitute.dsde.rawls.model.Subsystems.Subsystem
```
jmthibault79: this line fails `test:compile`

helgridly: crap. thanks

build.sbt
```diff
@@ -3,37 +3,30 @@ import Testing._
...
-  .withTestSettings
-
 lazy val rawlsModel = project.in(file("model"))
   .settings(modelSettings:_*)
   .withTestSettings
 
 lazy val workbenchMetrics = project.in(file("metrics"))
   .settings(metricsSettings:_*)
-  .dependsOn(workbenchUtil % compileAndTest)
+  .dependsOn(rawlsModel)
```
rtitle: Merge/rebase issue? I purposefully removed the `rawlsModel` dependency from `workbenchMetrics`.

helgridly: Possibly. I found a dependency on subsystem in StatsDTestUtils, but that might be an unused import.

rtitle: Yeah I think I removed those imports - `StatsDTestUtils` shouldn't need `rawlsModel`.

https://github.com/broadinstitute/rawls/blob/develop/metrics/src/test/scala/org/broadinstitute/dsde/rawls/metrics/StatsDTestUtils.scala#L3-L12

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitorSpec.scala
```diff
@@ -13,7 +13,7 @@ import org.broadinstitute.dsde.rawls.monitor.HealthMonitor._
 import org.mockito.ArgumentMatchers._
 import org.mockito.Mockito._
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
-import org.scalatest.mock.MockitoSugar
+import org.scalatest.mockito.MockitoSugar
```
rtitle, line 16: Nice. Running `test:compile`, I see a warning for `MetricsSpec` -- looks like it needs this change too.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
rtitle: I was curious so I tried moving this back to the bottom, and it compiles for me.

See:
https://github.com/broadinstitute/rawls/compare/remove_workbenchlibs...rt-monoid-at-bottom
https://travis-ci.org/broadinstitute/rawls/builds/257961117

Not that I necessarily object to putting it at the top. :)

project/Dependencies.scala
rtitle: Yay! I feel like there were some places where I was wishing for scalatest 3.x recently..

geoffjentry: Weren't you trying to use the async matchers recently? That could be why

project/Dependencies.scala
```diff
@@ -56,10 +56,16 @@ object Dependencies {
   val mysqlConnector: ModuleID =  "mysql"                         % "mysql-connector-java"  % "5.1.38"
   val liquibaseCore: ModuleID =   "org.liquibase"                 % "liquibase-core"        % "3.5.3"
   val logbackClassic: ModuleID =  "ch.qos.logback"                % "logback-classic"       % "1.1.6"
-  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "2.2.4" % "test"
-  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.7.22" % "test"
+  val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.0.1" % "test"
+  val mockito: ModuleID =         "org.mockito"                   % "mockito-core"          % "2.8.47" % "test"
   val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "3.9.2" % "test"
 
+  def excludeAkkaActor(m: ModuleID): ModuleID = m.exclude("com.typesafe.akka", "akka-actor")
```
rtitle, line 63: Should we exclude `akka-http-spray-json` as well? No idea if that conflicts with `spray-json`.

rtitle: Oh, `akka-http-spray-json` is in `workbenchModel`, not `workbenchUtil`. If/when we depend on `workbenchModel`, we might want to exclude it.

project/Dependencies.scala
rtitle: I don't think we can/should move `MockitoTestUtils` to `src/main/scala` -- it requires scalatest and having scalatest as a non-test dependency would be bad. I think this solution is fine.

project/Dependencies.scala
```diff
@@ -137,6 +141,8 @@ object Dependencies {
     akkaTestkit,
     sprayTestkit,
     mockserverNetty,
-    mockito
+    mockito,
+    workbenchUtil,
```
rtitle, line 145: Could maybe omit these from `rawlsCoreDependencies` because it's already including `modelDependencies ++ googleDependencies ++ metricsDependencies`, all of which contain `workbenchUtil[Test]`

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitorSpec.scala
rtitle: In fact, `MetricsSpec` could just extend `MockitoTestUtils` rather than use `MockitoSugar` directly.

project/Dependencies.scala
geoffjentry: I think they can exist next to each other, you just run the risk of importing the wrong one

rtitle (changes requested):
Looks good - nitpicks aside, my main request is to remove the `rawlsModel` dependency from `workbenchMetrics`. Also it looks like Travis failed, haven't checked why.

rtitle:
Test failures with:
```
 java.lang.NoSuchMethodError: akka.actor.ActorSystem.shutdown()
```
Maybe some test code needs to be changed to use `terminate()`. That's unfortunate that it fails at runtime, not compile time.

project/Dependencies.scala
rtitle: That could be it!

Also this: https://groups.google.com/forum/?pli=1#!topic/scalatest-users/SyyW8RC1pfA, I really wanted to do `should contain allElementsOf` for some reason :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
helgridly: Hmm. I'll move it back to minimise changes.

project/Dependencies.scala
helgridly: I'll leave it to documentation in `workbench-lib` then.

build.sbt
helgridly: Bad merge for me then. It was a bit messy. Will fix.

project/Dependencies.scala
helgridly: My plan is to not have rawls depend on `workbenchModel` until we upgrade from spray to akka-http.

project/Dependencies.scala
helgridly: I don't buy that that definition of `rawlsCoreDependencies` is right rather than just lazy.

helgridly:
> Maybe some test code needs to be changed to use `terminate()`. That's unfortunate that it fails at runtime, not compile time.

???!!!!! I thought the whole point of static typing was to avoid this madness! This is _extremely_ uncool.

rtitle:
```
java.lang.NoSuchMethodError: akka.actor.ActorSystem.shutdown()V
	at spray.testkit.RouteTest$class.cleanUp(RouteTest.scala:47)
	at org.broadinstitute.dsde.rawls.webservice.NotificationSpec.cleanUp(NotificationSpec.scala:20)
	at spray.testkit.ScalatestInterface$class.afterAll(ScalatestInterface.scala:28)
	at org.broadinstitute.dsde.rawls.webservice.NotificationSpec.org$broadinstitute$dsde$rawls$webservice$ApiServiceSpec$$super$afterAll(NotificationSpec.scala:20)
	at org.broadinstitute.dsde.rawls.webservice.ApiServiceSpec$class.afterAll(ApiServiceSpec.scala:55)
	at org.broadinstitute.dsde.rawls.webservice.NotificationSpec.afterAll(NotificationSpec.scala:20)
```

It's coming from `spray.testkit` -- I wonder if spray.testkit is tied to an older version of akka. :( I dunno, haven't researched it too much.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitorSpec.scala
helgridly: Might as well leave it if it's not using anything in `MockitoTestUtils` (but will fix `mock` -> `mockito`).

helgridly:
uh oh. can we resolve this with some judicious `exclude`s?

helgridly:
Spray docs say it explicitly relies on akka-actor-2.2, and you have to provide it yourself: http://spray.io/documentation/1.2.4/spray-testkit/

project/Dependencies.scala
breilly2: The other option would be to have a separate artifact/source tree whose purpose is providing support for testing. Then, from that artifact's point of view, `MockitoTestUtils` belongs in `src/main` and it's up to modules that depend on it to import it in test scope. (I may be using more maven-like terminology and thinking here but I think it's mostly applicable.)

project/Dependencies.scala
```diff
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val akkaV = "2.3.6"
+  val akkaV = "2.5.3"
```
jmthibault79, line 4: ~TestKit is now failing with `java.lang.NoSuchMethodError: akka.actor.ActorSystem.shutdown()` when running tests.  I'm guessing it's due to this.~
OK I see Rob already caught this

jmthibault79:
~This PR breaks `sbt test-only` which is something I use very frequently.~
Nope this was PEBKAC, please carry on.

helgridly:
Looks like we need Rawls to go to akka-http for this to go through. Closing but we might come back to it one day.

PR #748, opened 2017-08-01
[GAWB-2146] Instrument Google API calls

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-2146

See inline comments. Let me know if anything is hard to understand. Unfortunately there are not really any specs for the Google http stuff. There are integration tests which I haven't really tackled yet. I've also been testing in fiab.

Still todo:
* Move relevant code to https://github.com/broadinstitute/workbench-libs
* Run the integration tests, perhaps add test cases to them

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -28,49 +29,47 @@ class HttpExecutionServiceDAO( executionServiceURL: String, submissionTimeout: F
...
     val formData = Map("workflowSource" -> BodyPart(wdl), "workflowInputs" -> BodyPart(inputs.mkString("[", ",", "]"))) ++ options.map("workflowOptions" -> BodyPart(_))
-    pipeline(Post(url, MultipartFormData(formData)))
+    pipelineWithTimeout[Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]]](userInfo, timeout) apply (Post(url, MultipartFormData(formData)))
   }
 
   override def status(id: String, userInfo: UserInfo): Future[ExecutionServiceStatus] = {
     val url = executionServiceURL + s"/workflows/v1/${id}/status"
-    val pipeline = addAuthHeader(userInfo) ~> instrumentedGzSendReceive ~> unmarshal[ExecutionServiceStatus]
-    retry(when500) { () => pipeline(Get(url)) }
+    retry(when500) { count => pipeline[ExecutionServiceStatus](userInfo, count) apply Get(url) }
```
rtitle: You might want to scroll down and check out `Retry.scala`. Basically I changed it so each retry iteration passes the current retry count as an `Int`, instead of just passing `Unit` every time. The retryCount is then passed to `instrumentedGzSendReceive` and is instrumented with a different suffix (`.retry` instead of `.request`). This should let us see how often we're retrying requests to which services.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -171,7 +174,7 @@ class HttpGoogleServicesDAO(
     def insertBucket: (Map[WorkspaceAccessLevel, RawlsGroup], Option[Map[WorkspaceAccessLevel, RawlsGroup]]) => Future[String] = { (accessGroupsByLevel, intersectionGroupsByLevel) =>
       val bucketName = getBucketName(workspaceId)
       retryWhen500orGoogleError {
-        () => {
+        implicit count => {
```
rtitle: See `GoogleUtilities.scala`. The retry count and counters are passed implicitly to `executeGoogleRequest`, so I luckily didn't have to change too much in this class.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -313,8 +316,8 @@ class HttpGoogleServicesDAO(
   override def deleteBucket(bucketName: String, monitorRef: ActorRef): Future[Unit] = {
     val buckets = getStorage(getBucketServiceAccountCredential).buckets
     val deleter = buckets.delete(bucketName)
-    retryWithRecoverWhen500orGoogleError(() => {
-      executeGoogleRequest(deleter)
+    retryWithRecoverWhen500orGoogleError(count => {
+      executeGoogleRequest(deleter)(GoogleInstrumentedService.Buckets, count)
```
rtitle: See `GoogleInstrumented.scala` and `GoogleInstrumentedServiceMapper.scala`. Basically I try to infer the `GoogleInstrumentedService` enum based on the type parameter of `AbstractGoogleClientRequest[T]` (it's like they want us to make a typeclass). Most of the time this is possible because there's a one-to-one mapping between the type parameter and service -- e.g. `AbstractGoogleClientRequest[com.google.api.services.storage.model.Bucket]` is always a GoogleBuckets request. However sometimes the type is ambiguous -- for example most delete requests are a `AbstractGoogleClientRequest[Void]`. So I also made it possible to explicitly pass the service enum here.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -10,7 +10,7 @@ trait InstrumentationDirectives extends RawlsInstrumented {
   def instrumentRequest: Directive0 = requestInstance flatMap { request =>
     val timeStamp = System.currentTimeMillis
     mapHttpResponse { response =>
-      httpRequestCounter(ExpandedMetricBuilder.empty)(request, response).inc()
+      httpRequestCounter(ExpandedMetricBuilder.empty)(request, response, false).inc()
```
rtitle: `httpRequestCounter` now takes a boolean `isRetry` parameter.

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
```diff
@@ -1,41 +1,36 @@
...
-import org.broadinstitute.dsde.rawls.dataaccess.slick.{RawlsBillingProjectOperationRecord, TestDriverComponent}
 
 import scala.collection.JavaConversions._
-import scala.concurrent.duration._
+import scala.concurrent.duration.{Duration, _}
+import scala.concurrent.{Await, Future}
 import scala.util.Try
 
-class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationTestConfig with Retry with TestDriverComponent with BeforeAndAfterAll with LazyLogging {
+class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers with IntegrationTestConfig with Retry with TestDriverComponent with BeforeAndAfterAll with LazyLogging with Eventually with MockitoTestUtils with StatsDTestUtils {
```
rtitle, line 34: I haven't actually run these integration tests -- I should probably figure out how.

google/src/main/scala/org/broadinstitute/dsde/rawls/metrics/GoogleInstrumented.scala
```diff
@@ -0,0 +1,33 @@
...
+/**
+  * Mixin trait for Google instrumentation.
+  */
+trait GoogleInstrumented extends WorkbenchInstrumented {
+  final val GoogleServiceMetricKey = "googleService"
+
+  protected implicit def googleCounters[A: GoogleInstrumentedServiceMapper]: GoogleCounters[A] =
+    (request, responseOrException, isRetry) => {
+      val base = ExpandedMetricBuilder
+        .expand(GoogleServiceMetricKey, implicitly[GoogleInstrumentedServiceMapper[A]].service)
```
rtitle: This is where we implicitly look up the `GoogleInstrumentedServiceMapper` typeclass.

google/src/main/scala/org/broadinstitute/dsde/rawls/metrics/GoogleInstrumented.scala
```diff
@@ -0,0 +1,33 @@
...
+        .expand(HttpRequestMethodMetricKey, request.getRequestMethod.toLowerCase)
+        .expand(HttpResponseStatusCodeMetricKey, responseOrException.fold(_.getStatusCode, _.getStatusCode))
+      val counter = base.asCounter(if (isRetry) "retry" else "request")
+      val timer = base.asTimer("latency")
+      (counter, timer)
+    }
+
+  protected implicit def googleCountersWithService[A](service: GoogleInstrumentedService): GoogleCounters[A] = {
```
rtitle: This lets us explicitly pass in a `GoogleInstrumentedService` to `GoogleUtilities.executeGoogleRequest` if we don't have a typeclass instance.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -111,7 +114,7 @@ class HttpGoogleServicesDAO(
     }
   }
 
-  def allowGoogleCloudStorageWrite(bucketName: String): Unit = {
+  def allowGoogleCloudStorageWrite(bucketName: String)(implicit retryCount: Int = 0): Unit = {
```
helgridly: doesn't this mean that _any_ `Int` implicit in scope can be dropped in here? this seems terrifying.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -115,6 +119,12 @@ trait GoogleUtilities extends LazyLogging with Retry {
     logger.debug(GoogleRequest(request.getRequestMethod, request.buildHttpRequestUrl().toString, payload, System.currentTimeMillis() - startTime, statusCode, errorReport).toJson(GoogleRequestFormat).compactPrint)
   }
 
+  private def instrumentGoogleRequest[A](request: AbstractGoogleClientRequest[A], startTime: Long, responseOrException: Either[HttpResponseException, com.google.api.client.http.HttpResponse], isRetry: Boolean)(implicit counters: GoogleCounters[A]): Unit = {
+    val (counter, timer) = counters(request, responseOrException, isRetry)
+    counter += 1
```
helgridly: so the `retryCount` maintained in `Retry.scala` and passed around everywhere is the actual incrementing number of times you've attempted this call; and this counter is the statsd metric counter that you log?

google/src/main/scala/org/broadinstitute/dsde/rawls/metrics/GoogleInstrumentedServiceMapper.scala
```diff
@@ -0,0 +1,111 @@
...
+  // Google Genomics
+
+  implicit object GenomicsOperationMapper extends GoogleInstrumentedServiceMapper[Operation] {
+    override def service: GoogleInstrumentedService = GoogleInstrumentedService.Genomics
+  }
+
+  implicit object ListOperationsResponseMapper extends GoogleInstrumentedServiceMapper[ListOperationsResponse] {
+    override def service: GoogleInstrumentedService = GoogleInstrumentedService.Genomics
+  }
+}
```
helgridly: and they say there's less boilerplate in scala!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
helgridly: This seems off to me. Why does the call being retried care how many times it's been retried? If it did, it would have implemented its own `recover`. Maybe I'm stuck in OO-land here but this seems to violate encapsulation or something. Not so much here but this is especially noticeable in the Google classes where we have to pass around the `implicit count =>` in every call we want to retry.

For the Google case -- what if there was a GoogleUtilities/Retry mashup mixin called `GoogleRetry` with its own Google-aware definitions of the retry functions? Could we then keep the retry count isolated in the retry functions, rather than having to pass them through everything retry-able so they can pop out in `executeGoogleCall`? I'm okay tracking retries as a separate metric to the HTTP requests themselves (i.e. not appending `.retry` or `request` on the end).

helgridly (commented):
Does this get any simpler if we track retries separately to the requests?

core/src/test/scala/org/broadinstitute/dsde/rawls/integrationtest/HttpGoogleServicesDAOSpec.scala
helgridly: `sbt it:test`

helgridly: At one point they used to run on Jenkins builds. I _think_ they still do.

rtitle: Thanks

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
rtitle: Essentially, yes. For example the `counter` is either (depending on the value of `isRetry`):

`dev.firecloud.rawls.googleService.Buckets.httpRequestMethod.get.httpResponseStatusCode.200.request`
  - in this case we're counting a successful Buckets request

`dev.firecloud.rawls.googleService.Buckets.httpRequestMethod.get.httpResponseStatusCode.500.retry`
  - In this case we're counting a retry to a Buckets request. 

I think it's valuable to know _which_ services we're retrying most often, which is why I modeled it this way.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
rtitle: Yeah, I suppose it is. I made it implicit mainly to avoid editing too much code, but I guess developer laziness is not a great reason for using implicits. I'm tinkering with encapsulating this type of logic in an `InstrumentedRetry` trait, so this might change.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
rtitle: I see your point. This is kind of forcing an `IndexedSeq` approach on `Retry` -- where each iteration carries around its own retry index. But that might not be desirable in all cases. I'm thinking of encapsulating this in an `InstrumentedRetry` trait, which publishes retry counts to counters it is aware of. Still hand-wavy for now as I'm still figuring it out.

rtitle:
Verified on fiab, it seems to be working. Assigning reviewers.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -61,20 +63,22 @@ trait GoogleUtilities extends LazyLogging with Retry {
...
           processResponse(response)
         } finally {
           response.disconnect()
         }
       case Failure(httpRegrets: HttpResponseException) =>
         logGoogleRequest(request, start, httpRegrets)
+        instrumentGoogleRequest(request, start, Left(httpRegrets))
         throw httpRegrets
       case Failure(regrets) =>
         logGoogleRequest(request, start, regrets)
```
jmthibault79, line 84: instrument here

metrics/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentedRetry.scala
```diff
@@ -0,0 +1,30 @@
...
+  * Created by rtitle on 8/3/17.
+  */
+trait InstrumentedRetry extends Retry {
+  this: WorkbenchInstrumented with LazyLogging =>
+
+  /**
+    * Converts an AccumulatingFuture[A] to a Future[A].
+    * Given an implicit Histogram, instruments the number of failures in the histogram.
+    */
+  protected implicit def accumulatingFutureToFuture[A](af: AccumulatingFuture[A])(implicit histo: Histogram, executionContext: ExecutionContext): Future[A] = {
```
jmthibault79: Would it be possible for `Retry`'s version to be called by an implementing class instead?  Would be nice to `override` this but it doesn't look possible

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
```diff
@@ -41,26 +53,35 @@ trait Retry {
...
-        }
-
-        Future.failed(t)
+  private def retryInternal[T](backoffIntervals: Seq[FiniteDuration],
+                               pred: Predicate[Throwable],
+                               failureLogMessage: String)
+                              (op: () => Future[T])
+                              (implicit executionContext: ExecutionContext): AccumulatingFuture[T] = {
+
+    def inner(remainingBackoffIntervales: Seq[FiniteDuration], errors: => List[Throwable]): AccumulatingFuture[T] = {
```
jmthibault79: `¿intervales?`

google/src/main/scala/org/broadinstitute/dsde/rawls/metrics/GoogleInstrumentedService.scala
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsde.rawls.metrics
+
+/**
+  * An instrumented Google service.
+  */
+object GoogleInstrumentedService extends Enumeration {
+  type GoogleInstrumentedService = Value
+  val Billing, Buckets, Genomics, Groups, PubSub = Value
```
jmthibault79: Should probably be `Storage` instead of `Buckets` - and can we reuse `Subsystem` here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -785,10 +800,10 @@ class HttpGoogleServicesDAO(
     val genomicsApi = new Genomics.Builder(httpTransport, jsonFactory, getGenomicsServiceAccountCredential).setApplicationName(appName).build()
     val operationRequest = genomicsApi.operations().get(opId)
 
-    retryWithRecoverWhen500orGoogleError(() => {
+    retryWithRecoverWhen500orGoogleError[Option[JsObject]](() => {
```
helgridly: do these type params have to be specified explicitly now? because if so i guarantee you a developer is going to forget. if that makes a compile error that's fine, but if it does some weird undetectable cast-to-unit nonsense then we're in trouble

if you're just being conscientious and it works fine without the explicit type param then please discard this comment

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
```diff
@@ -16,21 +16,29 @@ trait Retry {
...
+    * A Future that has potentially been retried, with accumulated errors.
+    * There are 3 cases:
+    * 1. The future failed 1 or more times, and the final result is an error.
+    *   - This is represented as {{{Left(NonEmptyList(errors))}}}
+    * 2. The future failed 1 or more times, but eventually succeeded.
+    *   - This is represented as {{{Right(List(errors), A)}}}
+    * 3. The future succeeded the first time.
+    *   - This is represented as {{{Right(List.empty, A)}}}
+    */
+  type AccumulatingFuture[A] = Future[Either[NonEmptyList[Throwable], (List[Throwable], A)]]
```
helgridly: Can we call this something else? Even `RetriedFuture` or something. It's not obvious from the name what it's accumulating.

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
helgridly: this makes me happy

helgridly: though the fact that this function is `retryInternal.inner` does not

helgridly (approved):
minor nitpicks but this is tons better than passing around the count everywhere. much 👍

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
rtitle: ok. Only thing is we won't have a response status code (since this is _not_ a `HttpResponseException`) but we could instrument with an unknown status.

metrics/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentedRetry.scala
rtitle: Yeah it's not overridable because of the implicit parameter.

This might work:
```
    val instrumentedAccumulatingFuture = af.andThen {
      case Success(Left(errors)) =>
        histo += errors.toList.size
      case Success(Right((errors, a))) =>
        histo += errors.size
    }
    super.accumulatingFutureToFuture(instrumentedAccumulatingFuture)
```

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
rtitle: ¡intervales!

Maybe `def loop` instead? I made an inner function by habit to make the error accumulation tail-recursive. But it still might not be tail-recursive because of the `after(remainingBackoffIntervales.head, system.scheduler) { ... }` call. Will think about this more.

helgridly: `loop` is slightly better, yes.

google/src/main/scala/org/broadinstitute/dsde/rawls/metrics/GoogleInstrumentedService.scala
rtitle: Storage instead of Buckets makes sense. I think I didn't reuse `Subsystem` here because workbenchGoogle doesn't depend on rawlsModel. (Which reminds me - I still need to port these changes to https://github.com/broadinstitute/workbench-libs).

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
rtitle: Apparently it's not needed -- it compiles w/o it and other calls to `retryWithRecoverWhen500orGoogleError` work without the type param.

util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala
rtitle: Sure maybe `RetryableFuture`? This behavior is very specific to retries, so it makes sense to name it as such

helgridly: wfm!

rtitle:
Made PR comments. I think this is about ready to go.

I also duplicated the common changes here: https://github.com/broadinstitute/workbench-libs/pull/14

helgridly:
👍 here.

PR #750, opened 2017-08-01
use opendj in rawls for users and groups

UrsaStutsman:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


.travis.yml
```diff
@@ -15,11 +15,13 @@ before_install:
   - docker rm -f mysql || true          # in case the after_success step does not run
   - docker pull mysql/mysql-server:5.6
   # port 3306 is not available because Travis runs its own MySQL after 2006-12-01
-  - docker run -p 3307:3306 --name mysql -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d mysql/mysql-server:5.6
-
+  - docker run -p 3307:3306 --name mysql -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d mysql/mysql-server:5.7
```
breilly2: Be careful to make sure you have the right MySQL version here. It's been downgraded recently to match the version in production so I suspect 5.6 might be the correct one now (please double-check).

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/DirectorySubjectNameSupport.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.model.{RawlsGroupName, RawlsUserSubjectId}
+
+/**
+  * Created by dvoet on 6/6/17.
+  */
+trait DirectorySubjectNameSupport {
+  private val conf = ConfigFactory.parseResources("version.conf").withFallback(ConfigFactory.load())
+
+  val directoryConfig = DirectoryConfig(
```
breilly2, line 13: It bothers me a little that this is duplicated here when it's already in Boot.scala. Is there any way that this can make use of what was already initialized?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -104,13 +103,23 @@ trait RawlsBillingProjectComponent {
...
-      val subgroupsQuery = findDirectProjectSubgroups(rawlsProjectName)
-
-      val membersQuery = usersQuery union subgroupsQuery
-
-      membersQuery.result.map(_.map{ case(email, role) => RawlsBillingProjectMember(RawlsUserEmail(email), ProjectRoles.withName(role))})
+    def loadDirectProjectMembersWithEmail(rawlsProjectName: RawlsBillingProjectName): ReadWriteAction[Seq[RawlsBillingProjectMember]] = {
+      for {
+        roleGroupRecs <- rawlsBillingProjectGroupQuery.filter(_.projectName === rawlsProjectName.value).result
+        roleGroups <- DBIO.sequence(roleGroupRecs.map { roleRec => rawlsGroupQuery.load(RawlsGroupRef(RawlsGroupName(roleRec.groupName))).map(roleRec -> _)})
+        users <- rawlsUserQuery.load(roleGroups.flatMap(_._2.get.users))
```
breilly2: I'm struggling a little with being able to read this. It might be helpful to find a more expressive way to write `_._2`. Also, is it guaranteed that `get` won't be called on `None` here? Looking at `rawlsGroupQuery.load`, I think `None` is very possible. It would probably represent an inconsistency between the Rawls DB and LDAP which ideally won't happen, but given that we don't have distributed transactions between the two I suspect it's very possible.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -145,13 +145,6 @@ trait AdminApiService extends HttpService with PerRequestCreator with UserInfoDi
           UserService.AdminSynchronizeGroupMembers(RawlsGroupRef(RawlsGroupName(groupName))))
       }
     } ~
-    path("admin" / "users") {
```
breilly2, line 148: So we're dropping this from the web service API? It's admin so I guess we should be the only ones depending on it anyway, and I guess it probably belongs in Sam instead of Rawls. I guess I'm okay with this. Need to remove it from `rawls.yaml`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsUserComponent.scala
```diff
@@ -21,23 +21,23 @@ trait RawlsUserComponent {
   private type RawlsUserQuery = Query[RawlsUserTable, RawlsUserRecord, Seq]
 
   object rawlsUserQuery extends TableQuery(new RawlsUserTable(_)) {
-
+//replicate
```
breilly2: Success! It replicated 3 times! :)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiDirectoryDAO.scala
```diff
@@ -0,0 +1,389 @@
...
+
+        Option(unmarshallGroup(attributes))
+
+      }.recover {
+        case e: NameNotFoundException => None
+
+      }.get
+    }
+
+    def load(groupRefs: TraversableOnce[RawlsGroupRef]): ReadWriteAction[TraversableOnce[RawlsGroup]] = withContext { ctx =>
```
breilly2: I'm not sure what this function should do if any of the groups are not found. Based on current callers, this is only called to fetch subgroups that were returned from a previous load. There is a very very small window during which a reference could be broken. It's very unlikely that it will happen but it could happen someday and would be nearly impossible to track down.

Can we detect groups that aren't found and warn-log them? I'd also like to see some scaladoc on this describing the behavior if a group is not found. (This isn't a problem for the singular `load` above which returns `Option[RawlsGroup]`.)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponentSpec.scala
```diff
@@ -307,7 +308,7 @@ class SubmissionComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
 
     // Validate testUser counts
     val result = runAndWait(workflowQuery.countWorkflowsByQueueStatusByUser)
-    result should contain key (testUserEmail)
+    result should contain key testUserId
```
breilly2: This looks like an API change to me. Granted it's admin endpoint but it was touched by A-Team in May so they probably care. Have they been consulted?

breilly2:
Pretty good test coverage. Nice that the group tests didn't have to change! One test that I couldn't find that I'd like to see is removing a user from a sub-group and verifying that they are no longer in the parent group.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
dvoet: Alex OKed its removal, it is not useful anymore

README.md
```diff
@@ -26,16 +26,15 @@ See the wiki for detailed documentation.
 Ensure that docker is up to date and initialized.
 Spin up mysql locally and validate that it is working:
 ```
-./docker/run-mysql.sh start
+./docker/run-mysql.sh
```
dvoet: the README in develop is the correct one. And we should as the same steps for opendj

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
```diff
@@ -51,5 +51,5 @@
     <include file="changesets/20170616_submission_workflow_failure_mode_column.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170626_multiple_auth_domains.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170811_index_scratch_tables.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20170816_save_rawls_instance_for_workflow.xml" relativeToChangelogFile="true"/>
+    <include file="changesets/20170821_removing_foreign_key_constraints.xml" relativeToChangelogFile="true"/>
```
dvoet: clobbered changes in develop!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/DirectorySubjectNameSupport.scala
dvoet: this is all transition code anyway... let's not get too hung up on details

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiDirectoryDAO.scala
```diff
@@ -238,24 +282,24 @@ trait JndiDirectoryDAO extends DirectorySubjectNameSupport with JndiSupport {
     }
 
     def updateSynchronizedDate(rawlsGroupRef: RawlsGroupRef): ReadWriteAction[Unit] = withContext { ctx =>
-      ctx.modifyAttributes(groupDn(rawlsGroupRef.groupName), DirContext.REPLACE_ATTRIBUTE, new BasicAttributes(Attr.groupSynchronizedTimestamp, new Timestamp(System.currentTimeMillis()), true))
+      ctx.modifyAttributes(groupDn(rawlsGroupRef.groupName), DirContext.REPLACE_ATTRIBUTE, new BasicAttributes(Attr.groupSynchronizedTimestamp, dateFormat.format(new Date()), true))
```
breilly2: `java.text.SimpleDateFormat` is not thread-safe. I suspect that this usage is problematic but I'm not 100% sure.

PR #752, opened 2017-08-08
GAWB-2439: Validate Output Expressions and prevent SubmissionMonitorActor crashes

jmthibault79:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
@@ -7,17 +7,18 @@ import java.util.UUID
 import akka.util.ByteString
 import org.apache.commons.codec.binary.Base64
 import org.broadinstitute.dsde.rawls.model._
-import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport, StringValidationUtils}
 import slick.driver.JdbcDriver
 import slick.jdbc.{GetResult, PositionedParameters, SQLActionBuilder, SetParameter}
 import spray.http.StatusCodes
 
 import scala.concurrent.ExecutionContext
 
-trait DriverComponent {
+trait DriverComponent extends StringValidationUtils {
   val driver: JdbcDriver
   val batchSize: Int
   implicit val executionContext: ExecutionContext
+  override implicit val errorReportSource = ErrorReportSource("rawls")
 
   // needed by MySQL but not actually used; we will always overwrite
   val defaultTimeStamp = Timestamp.valueOf("2001-01-01 01:01:01.0")
@@ -37,23 +38,6 @@ trait DriverComponent {
       case Seq() => None
       case Seq(one) => Option(one)
       case tooMany => throw new RawlsException(s"Expected 0 or 1 result but found all of these: $tooMany")
-    }
-  }
-
-  //in general, we only support alphanumeric, spaces, _, and - for user-input
-  def validateUserDefinedString(s: String) = {
-    if(!s.matches("[A-z0-9_-]+")) throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = s"Invalid input: $s. Input may only contain alphanumeric characters, underscores, and dashes.", statusCode = StatusCodes.BadRequest))
-  }
-
-  def validateMaxStringLength(s: String, maxLength: Int) = {
-    if(s.length > maxLength) throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = s"Invalid input: $s. Input may be a max of $maxLength characters.", statusCode = StatusCodes.BadRequest))
-  }
-
-  def validateAttributeName(an: AttributeName, entityType: String) = {
-    if (Attributable.reservedAttributeNames.exists(_.equalsIgnoreCase(an.name)) ||
-      AttributeName.withDefaultNS(entityType + Attributable.entityIdAttributeSuffix).equalsIgnoreCase(an)) {
-
-      throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = s"Attribute name ${an.name} is reserved", statusCode = StatusCodes.BadRequest))
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
@@ -1,0 +1,31 @@
+package org.broadinstitute.dsde.rawls.expressions
+
+import org.broadinstitute.dsde.rawls.{RawlsException, StringValidationUtils}
+import org.broadinstitute.dsde.rawls.model.{AttributeName, ErrorReportSource}
+
+sealed trait OutputExpressionTarget {
+  val prefix: String
+  def remainder(s: String): String = s.stripPrefix(prefix)
+}
+case object ThisEntityTarget extends OutputExpressionTarget { override val prefix = "this." }
+case object WorkspaceTarget extends OutputExpressionTarget { override val prefix = "workspace." }
+
+case class OutputExpression(target: OutputExpressionTarget, attributeName: AttributeName)
+
+object OutputExpression extends StringValidationUtils {
+  override implicit val errorReportSource = ErrorReportSource("rawls")
+
+  def apply(s: String): OutputExpression = {
+    val target = if(s.startsWith(ThisEntityTarget.prefix))
+      ThisEntityTarget
+    else if(s.startsWith(WorkspaceTarget.prefix))
+      WorkspaceTarget
+    else
+      throw new RawlsException(s"Invalid output expression: $s")
+
+    val attributeName = AttributeName.fromDelimitedName(target.remainder(s))
+    validateUserDefinedString(attributeName.name)
+
+    OutputExpression(target, attributeName)
+  }
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -10,6 +10,7 @@ import nl.grons.metrics.scala.Counter
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction, ReadWriteAction, WorkflowRecord}
+import org.broadinstitute.dsde.rawls.expressions.{OutputExpression, ThisEntityTarget, WorkspaceTarget}
 import org.broadinstitute.dsde.rawls.jobexec.SubmissionMonitorActor._
 import org.broadinstitute.dsde.rawls.jobexec.SubmissionSupervisor.{CheckCurrentWorkflowStatusCounts, SaveCurrentWorkflowStatusCounts}
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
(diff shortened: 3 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
```diff
@@ -40,23 +41,6 @@ trait DriverComponent {
     }
   }
 
-  //in general, we only support alphanumeric, spaces, _, and - for user-input
```
jmthibault79, line 43: Moved to `StringUtils` because I want to use `validateUserDefinedString` outside of a DB context

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
```diff
@@ -0,0 +1,31 @@
...
+  val prefix: String
+  def remainder(s: String): String = s.stripPrefix(prefix)
+}
+case object ThisEntityTarget extends OutputExpressionTarget { override val prefix = "this." }
+case object WorkspaceTarget extends OutputExpressionTarget { override val prefix = "workspace." }
+
+case class OutputExpression(target: OutputExpressionTarget, attributeName: AttributeName)
+
+object OutputExpression extends StringUtils {
+  override implicit val errorReportSource = ErrorReportSource("rawls")
```
jmthibault79: for `StringUtils`.  I could have `StringUtils` set this itself instead.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def apply(s: String): OutputExpression = {
+    val target = if(s.startsWith(ThisEntityTarget.prefix))
```
jmthibault79: I bet there's a fancier way to do this

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
```diff
@@ -0,0 +1,31 @@
...
+      ThisEntityTarget
+    else if(s.startsWith(WorkspaceTarget.prefix))
+      WorkspaceTarget
+    else
+      throw new RawlsException(s"Invalid output expression: $s")
+
+    val attributeName = AttributeName.fromDelimitedName(target.remainder(s))
+    validateUserDefinedString(attributeName.name)
```
jmthibault79, line 27: This is the same validation done later when we try to save attributes back to the data model

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -299,12 +300,12 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     } else {
       for {
         // load all the starting data
-        entitiesById <-      listWorkflowEntitiesById(workflowsWithOutputs, dataAccess)
-        outputExpressions <- listMethodConfigOutputsForSubmission(dataAccess)
-        workspace <-         getWorkspace(dataAccess).map(_.getOrElse(throw new RawlsException(s"workspace for submission $submissionId not found")))
+        entitiesById <-         listWorkflowEntitiesById(workflowsWithOutputs, dataAccess)
+        outputExpressionMap <-  listMethodConfigOutputsForSubmission(dataAccess)
```
jmthibault79, line 304: It's not just Output Expressions.  It's Execution Outputs mapped to them.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -339,14 +340,16 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
       val outputs = outputsResponse.outputs
       logger.debug(s"attaching outputs for ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}: ${outputs}")
-      logger.debug(s"output expressions for ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}: ${outputExpressions}")
+      logger.debug(s"output expressions for ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}: ${outputExpressionMap}")
 
-      val attributes = outputExpressions.map { case (outputName, outputExpr) =>
+      val attributes = outputExpressionMap.map { case (outputName, outputExprStr) =>
         Try {
+          // parse and validate the output expression strings
+          val outputExpr = OutputExpression(outputExprStr)
```
jmthibault79, line 352: Validating here since it's a place we are already handling exceptions nicely.

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpressionSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+    // does not enforce Attribute Namespace constraints
+    OutputExpression("this.arbitrary:whatever") shouldEqual OutputExpression(ThisEntityTarget, AttributeName("arbitrary", "whatever"))
+    OutputExpression("workspace.arbitrary:whatever") shouldEqual OutputExpression(WorkspaceTarget, AttributeName("arbitrary", "whatever"))
+  }
+
+  it should "reject invalid output expressions" in {
+    intercept[RawlsException] { OutputExpression("this.") }
+    intercept[RawlsException] { OutputExpression("this.bad|character") }
+    intercept[RawlsException] { OutputExpression("this.case_sample.attribute") }
```
jmthibault79, line 23: This one is specifically call out in the old JIRA DSDEEPB-1603

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -412,6 +409,61 @@ class SubmissionMonitorSpec(_system: ActorSystem) extends TestKit(_system) with
...
+
+      withWorkspaceContext(testData.workspace) { context =>
+        runAndWait(methodConfigurationQuery.create(context, mcBadExprs))
+        runAndWait(submissionQuery.create(context, subBadExprs))
+      }
+
+      val monitor = createSubmissionMonitor(dataSource, subBadExprs, testData.wsName, new SubmissionTestExecutionServiceDAO(WorkflowStatuses.Succeeded.toString))
+      val workflowRecs = runAndWait(workflowQuery.listWorkflowRecsForSubmission(UUID.fromString(subBadExprs.submissionId)))
+
+      runAndWait(monitor.handleOutputs(workflowRecs.map(r => (r, ExecutionServiceOutputs(r.externalId.get, Map("bad1" -> Left(AttributeString("result")))))), this))
```
jmthibault79, line 440: Here is where it would crash before

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.dsde.rawls
+
+import org.broadinstitute.dsde.rawls.model.{Attributable, AttributeName, ErrorReport, ErrorReportSource}
+import spray.http.StatusCodes
+
+trait StringUtils {
+  implicit val errorReportSource: ErrorReportSource
```
jmthibault79: for `RawlsExceptionWithErrorReport`

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.dsde.rawls
+
+import org.broadinstitute.dsde.rawls.model.{Attributable, AttributeName, ErrorReport, ErrorReportSource}
+import spray.http.StatusCodes
+
+trait StringUtils {
```
rtitle: Instead of making this a trait with an `implicit val`, what about making `StringUtils` an object and pass the implicit `ErrorReportSource` to each method? E.g.:
```
def validateUserDefinedString(s: String)(implicit errorReportSource: ErrorReportSource): Unit = // etc
```
IMO that's slightly more idiomatic Scala (maybe I'm biased against OO).

Also `StringUtils` sounds a little general, maybe name this `StringValidationUtils`?

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+  implicit val errorReportSource: ErrorReportSource
+
+  //in general, we only support alphanumeric, spaces, _, and - for user-input
+  def validateUserDefinedString(s: String): Unit = {
+    if(!s.matches("[A-z0-9_-]+")) throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = s"Invalid input: $s. Input may only contain alphanumeric characters, underscores, and dashes.", statusCode = StatusCodes.BadRequest))
```
rtitle: This will compile the regex for each call to this method -- it would be more efficient to compile it ahead of time:
```
val AlphaNum = "[A-z0-9_-]+".r

def validateUserDefinedString(s: String): Unit = {
  if (AlphaNum.pattern.matcher(s).matches) {
    // etc
  }
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpression.scala
rtitle: You could make a companion object to `OutputExpressionTarget` and implement a `def fromPrefix(s: String): OutputExpressionTarget` method. But I think this is fine :)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -367,12 +370,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }
   }
 
-  def updateEntityAndWorkspace(entity: Entity, workspace: Workspace, workflowOutputs: Map[String, Attribute]): (Option[Entity], Option[Workspace]) = {
-    //Partition outputs by whether their attributes are entity attributes (begin with "this.") or workspace ones (implicitly; begin with "workspace.")
```
rtitle, line 371: Nice to move this logic to an ADT!

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/OutputExpressionSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+    intercept[RawlsException] { OutputExpression("workspace.") }
+    intercept[RawlsException] { OutputExpression("workspace........") }
+    intercept[RawlsException] { OutputExpression("workspace.nope.nope.nope") }
```
rtitle, line 26: https://media.giphy.com/media/6h4z4b3v6XWxO/giphy.gif

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
rtitle: I don't see `RawlsExceptionWithErrorReport` as taking any implicit parameters. Am I missing something?

rtitle: Oh nvm, it's for `ErrorReport`

rtitle:
A couple nitpicky comments, but overall looks good to me. 👍

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
jmthibault79: yes indeed

model/src/main/scala/org/broadinstitute/dsde/rawls/StringUtils.scala
jmthibault79: Open to arguments but I think I prefer the trait way.  Agree with you on the name.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -367,12 +370,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-    //This assumption (that it's either "this." or "workspace.") will be guaranteed by checking of the method config when it's imported; see DSDEEPB-1603.
```
helgridly, line 372: So this assumption was the root of the issue?

helgridly (commented):
IIRC the case we saw was when a user left an output expression empty, right? With my user hat on I can imagine some cases where I might not care about an output and don't need to write it back to the data model. Should we allow empty strings in outputs? Do the checks we make here sync up with what the `/validate` endpoint for method configs does?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
jmthibault79: The problem was caused by the correct prefix `workspace.` followed by nothing

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
jmthibault79: Thanks!  I'll have a few more for part 2 of this story (not this PR)

helgridly (approved):
we talked about this with face so 👍

PR #765, opened 2017-08-18
[GAWB-2438] Reduce instrumentation granularity

rtitle:
https://broadinstitute.atlassian.net/browse/GAWB-2438

Strips out workflow IDs and entity names from metric names to reduce the amount of data we're sending to Hosted Graphite.

Implemented by providing a `redactedUriExpansion()` method which uses a Spray `PathMatcher` to match and redact pieces of a `Uri.Path`.

Tests are passing, will confirm in a fiab.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
@@ -4,7 +4,7 @@ import akka.actor.ActorSystem
 import akka.util.Timeout
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.metrics.RawlsExpansion._
-import org.broadinstitute.dsde.rawls.metrics.{Expansion, InstrumentedRetry, RawlsInstrumented}
+import org.broadinstitute.dsde.rawls.metrics.{Expansion, InstrumentedRetry, RawlsExpansion, RawlsInstrumented}
 import org.broadinstitute.dsde.rawls.model.ExecutionJsonSupport._
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.FutureSupport
@@ -15,6 +15,7 @@ import spray.httpx.SprayJsonSupport._
 import spray.httpx.unmarshalling.FromResponseUnmarshaller
 import spray.json.DefaultJsonProtocol._
 import spray.json.JsObject
+import spray.routing.directives.PathDirectives._
 
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
@@ -30,8 +31,9 @@ class HttpExecutionServiceDAO( executionServiceURL: String, submissionTimeout: F
     ExpandedMetricBuilder.expand(SubsystemMetricKey, Subsystems.Cromwell)
 
   // Strip out workflow IDs from metrics by providing a redactedUriExpansion
-  private val WorkflowIdRegex = """^.*workflows\.v1\.([^.]*)\..*$""".r
-  override protected val UriExpansion: Expansion[Uri] = Expansion.redactedUriExpansion(WorkflowIdRegex)
+  override protected val UriExpansion: Expansion[Uri] = RawlsExpansion.redactedUriExpansion(
+    (Slash ~ "workflows") / "v1" / Segment / Neutral
+  )
 
   private def pipeline[A: FromResponseUnmarshaller](userInfo: UserInfo) =
     addAuthHeader(userInfo) ~> instrumentedGzSendReceive ~> unmarshal[A]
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -29,6 +30,11 @@ class HttpExecutionServiceDAO( executionServiceURL: String, submissionTimeout: F
   private implicit lazy val baseMetricBuilder =
     ExpandedMetricBuilder.expand(SubsystemMetricKey, Subsystems.Cromwell)
 
+  // Strip out workflow IDs from metrics by providing a redactedUriExpansion
+  override protected val UriExpansion: Expansion[Uri] = RawlsExpansion.redactedUriExpansion(
+    (Slash ~ "workflows") / "v1" / Segment / Neutral
```
rtitle, line 35: Matches `/workflows/v1/{id}/anything` and redacts the id.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -2,11 +2,27 @@ package org.broadinstitute.dsde.rawls.metrics
...
+import spray.http.Uri
+import spray.routing._
 import spray.routing.directives.BasicDirectives.mapHttpResponse
 import spray.routing.directives.MiscDirectives.requestInstance
+import spray.routing.directives.PathDirectives._
 
 trait InstrumentationDirectives extends RawlsInstrumented {
+
+  // Like Segment in that it matches and consumes any path segment, but does not extract a value.
+  private val SegmentIgnore: PathMatcher0 = Segment.hmap(_ => HNil)
```
rtitle, line 15: Oddly Spray doesn't seem to provide something like this. See http://spray.io/documentation/1.2.4/spray-routing/path-directives/pathmatcher-dsl/.

`Neutral` is not quite right because it doesn't consume the input -- so `Neutral / Segment` does not match `foo/bar`. Proof:

```
scala> val neutralPm = Neutral / Segment
scala> neutralPm(Uri.Path("foo/bar"))
res2 = Unmatched
scala> val explicitPm = "foo" / Segment
scala> explicitPm(Uri.Path("foo/bar"))
res3 = Matched(,bar :: HNil)
```

Weird. TIL some useless Spray knowledge.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -2,11 +2,27 @@ package org.broadinstitute.dsde.rawls.metrics
...
+
+  private val redactWorkflowIds: PathMatcher1[String] =
+    (Slash ~ "workspaces") / SegmentIgnore / SegmentIgnore / "submissions" / SegmentIgnore / "workflows" / (Segment ~ SegmentIgnore.repeat(separator = Slash))
```
helgridly, line 18: whoever invented spray dsl should be shot

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -2,11 +2,27 @@ package org.broadinstitute.dsde.rawls.metrics
...
+
+  private val redactEntityIds: PathMatcher1[String] =
+    (Slash ~ "workspaces") / SegmentIgnore / SegmentIgnore / "entities" / SegmentIgnore / (Segment ~ SegmentIgnore.repeat(separator = Slash))
```
helgridly, line 21: I wouldn't have thought to redact entity names too. Good catch.

helgridly (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
rtitle: Agreed! Originally I did this with regexes but then decided to use the spray pathmatcher instead. I still feel it's a little nicer than regexes in this case.

helgridly: Is it better in akka-http?

rtitle: I think the DSL is similar but it doesn't use (a very old version of) shapeless under the hood which is a ➕

PR #767, opened 2017-08-21
Reduce statsd timer metrics

rtitle:
No JIRA.

For the record, I don't love having to make this change, but I think it should be fine.

Metrics has been enabled on Rawls dev for 1 day. It's working great, but I'm concerned about hitting metric limits. We're currently using 4600 metrics out of 10000. If we enable it on all environments (dev, alpha, staging, prod) we could well exceed our 10k limit.

This change will eliminate 2590 of the metrics currently on dev, meaning dev Rawls will only use about 2k metrics. This should give us some more breathing room. These detailed timer metrics don't seem all that useful anyway.

If we _still_ bump up against quotas after deploying to all environments, my preferred next steps would be (in order):
1) only enable metrics in prod
2) upgrade our graphite plan
3) look at more metrics to filter

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/metrics/src/main/scala/com/readytalk/metrics/WorkbenchStatsD.scala
+++ b/metrics/src/main/scala/com/readytalk/metrics/WorkbenchStatsD.scala
@@ -1,12 +1,11 @@
-package org.broadinstitute.dsde.rawls.metrics
+package com.readytalk.metrics
 
-import com.readytalk.metrics.StatsD
 import com.typesafe.scalalogging.LazyLogging
 
 /**
   * Created by rtitle on 8/21/17.
   */
-class RawlsStatsD(host: String, port: Int) extends StatsD(host, port) with LazyLogging {
+class WorkbenchStatsD(host: String, port: Int) extends StatsD(host, port) with LazyLogging {
   // Filter out detailed timer metrics to prevent hitting Hosted Graphite quotas.
   // Keep: mean, p95, stddev
   val MetricSuffixesToFilter = Set("max", "min", "p50", "p75", "p98", "p99", "p999", "samples", "m1_rate", "m5_rate", "m15_rate", "mean_rate")
@@ -20,6 +19,6 @@ class RawlsStatsD(host: String, port: Int) extends StatsD(host, port) with LazyL
   }
 }
 
-object RawlsStatsD {
-  def apply(host: String, port: Int): RawlsStatsD = new RawlsStatsD(host, port)
+object WorkbenchStatsD {
+  def apply(host: String, port: Int): WorkbenchStatsD = new WorkbenchStatsD(host, port)
 }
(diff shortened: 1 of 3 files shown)


metrics/src/main/scala/com/readytalk/metrics/WorkbenchStatsD.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import com.typesafe.scalalogging.LazyLogging
+
+/**
+  * Created by rtitle on 8/21/17.
+  */
+class WorkbenchStatsD(host: String, port: Int) extends StatsD(host, port) with LazyLogging {
+  // Filter out detailed timer metrics to prevent hitting Hosted Graphite quotas.
+  // Keep: mean, p95, stddev
+  val MetricSuffixesToFilter = Set("max", "min", "p50", "p75", "p98", "p99", "p999", "samples", "m1_rate", "m5_rate", "m15_rate", "mean_rate")
```
rtitle, line 11: Unfortunately these are all hard-coded into metrics-statsd, so this seems like the best/only way to filter them.

https://github.com/ReadyTalk/metrics-statsd/blob/master/metrics3-statsd/src/main/java/com/readytalk/metrics/StatsDReporter.java#L204-L219

helgridly:
This seems legit. The very long list of intervals always seemed a little ridiculous to me.

Did we ever hear back from them about age-out?

rtitle:
Thanks - going to merge this before an alpha snapshot is taken. :)

They did -- the retention/aggregation not configurable by us, and they pointed at this blog post: https://blog.hostedgraphite.com/2017/08/02/advanced-data-views-better-observability-more-control/

I didn't realize they're not using [whisper](http://graphite.readthedocs.io/en/latest/whisper.html), they're using something called Riak instead. 

However they mentioned something interesting: you can set a retention on a per-metric basis by setting a "transient" flag in the metric prefix. For example, we could specify that submission UUID-level metrics expire in 7 days (or whatever). I have yet to experiment with this -- but it might be a useful thing to do in our app since we have a lot of very granular metrics.

geoffjentry:
Riak just died, wonder how long they'll stick with it :)

davidbernick:
(30k metrics, fyi)

rtitle:
@davidbernick do you mean that our quota should be 30k metrics? The current dashboard says 4k uses out of 10k. 30k would be a much nicer number :)

davidbernick:
i thought we had 30k? let me ask...

PR #775, opened 2017-09-05
GAWB-2481: Make authorization domains optional

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -460,7 +460,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
               // add to or replace current attributes, on an individual basis
               val newAttrs = sourceWorkspaceContext.workspace.attributes ++ destWorkspaceRequest.attributes
 
-              withNewWorkspaceContext(destWorkspaceRequest.copy(authorizationDomain = newAuthDomain, attributes = newAttrs), dataAccess) { destWorkspaceContext =>
+              withNewWorkspaceContext(destWorkspaceRequest.copy(authorizationDomain = Option(newAuthDomain), attributes = newAttrs), dataAccess) { destWorkspaceContext =>
                 dataAccess.entityQuery.copyAllEntities(sourceWorkspaceContext, destWorkspaceContext) andThen
                   dataAccess.methodConfigurationQuery.listActive(sourceWorkspaceContext).flatMap { methodConfigShorts =>
                     val inserts = methodConfigShorts.map { methodConfigShort =>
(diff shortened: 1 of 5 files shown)


model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -76,8 +76,8 @@ object AttributeName {
 case class WorkspaceRequest (
                               namespace: String,
                               name: String,
-                              authorizationDomain: Set[ManagedGroupRef],
-                              attributes: AttributeMap
+                              attributes: AttributeMap,
+                              authorizationDomain: Option[Set[ManagedGroupRef]] = Option(Set.empty)
```
dvoet, line 80: is None functionally any different that Option(Set.empty)?

dvoet: I guess what I am really asking is should we treat Option(Set.empty) as an error condition, i.e. if the user specifies the field they probably intended to put something in there so we should tell them they didn't?

breilly2: For a person, perhaps. For a program, I'm inclined to allow an empty set. Also, rejecting an empty list would now be a breaking API change.

abaumann: So if None means no auth domain, what would an empty set be?  Hopefully not a workspace nobody is authorized to access.  If that _is_ the case though I'd rather not let people make a workspace unaccessible like that.  If Option(Set.empty) is the same as None that's perhaps confusing, so I'd probably lean toward treating Option(Set.empty) as an error along the lines of "If an authorization domain is defined it needs to have at least one member"

MatthewBemis: @abaumann as this is implemented, None and Set.empty both mean that no authorization domain is set, so you couldn't create an unaccessible workspace. 

I think I agree that for human users of the endpoint, erroring on an empty set is probably the more clear thing to do- but @breilly2 is right- doing it that way would be a new breaking API change since people have updated their scripts to feed in an empty set

abaumann: yeah, unfortunately - so I guess having them equal seems OK, just odd - we should document this well

UrsaStutsman (approved):
LGTM

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1933,7 +1933,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       val workspace = Workspace(
         namespace = workspaceRequest.namespace,
         name = workspaceRequest.name,
-        authorizationDomain = workspaceRequest.authorizationDomain,
+        authorizationDomain = workspaceRequest.authorizationDomain.getOrElse(Set.empty),
```
breilly2: For readability, conciseness, (and future flexibility in case we want to change how we handle `None`), I think it makes sense to extract `getOrElse(Set.empty)` where possible.

dvoet (approved):

PR #781, opened 2017-09-08
Fix dodgy metrics for submission and workflow gauges

helgridly:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -170,6 +170,7 @@ object Boot extends App with LazyLogging {
       slickDataSource,
       gcsDAO.getBucketServiceAccountCredential,
       util.toScalaDuration(submissionMonitorConfig.getDuration("submissionPollInterval")),
+      submissionMonitorConfig.getBoolean("trackDetailedSubmissionMetrics"),
       workbenchMetricBaseName = metricsPrefix
     ).withDispatcher("submission-monitor-dispatcher"), "rawls-submission-supervisor")
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -168,6 +168,7 @@ object Boot extends App with LazyLogging {
     val submissionSupervisor = system.actorOf(SubmissionSupervisor.props(
       shardedExecutionServiceCluster,
       slickDataSource,
+      gcsDAO.getBucketServiceAccountCredential,
       util.toScalaDuration(submissionMonitorConfig.getDuration("submissionPollInterval")),
       workbenchMetricBaseName = metricsPrefix
     ).withDispatcher("submission-monitor-dispatcher"), "rawls-submission-supervisor")
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
@@ -40,8 +40,9 @@ object SubmissionSupervisor {
             datasource: SlickDataSource,
             bucketCredential: Credential,
             submissionPollInterval: FiniteDuration = 1 minutes,
+            trackDetailedSubmissionMetrics: Boolean = true,
             workbenchMetricBaseName: String): Props = {
-    Props(new SubmissionSupervisor(executionServiceCluster, datasource, bucketCredential, submissionPollInterval, workbenchMetricBaseName))
+    Props(new SubmissionSupervisor(executionServiceCluster, datasource, bucketCredential, submissionPollInterval, trackDetailedSubmissionMetrics, workbenchMetricBaseName))
   }
 }
 
@@ -57,6 +58,7 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
                            datasource: SlickDataSource,
                            bucketCredential: Credential,
                            submissionPollInterval: FiniteDuration,
+                           trackDetailedSubmissionMetrics: Boolean,
                            override val workbenchMetricBaseName: String) extends Actor with LazyLogging with RawlsInstrumented {
   import context._
 
@@ -88,7 +90,7 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
   override def preStart(): Unit = {
     super.preStart()
 
-    initGlobalJobExecGauges()
+    registerGlobalJobExecGauges()
     system.scheduler.schedule(0 seconds, submissionPollInterval, self, UpdateGlobalJobExecGauges)
   }
 
@@ -96,7 +98,7 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
     case SubmissionStarted(workspaceName, submissionId) =>
       val child = startSubmissionMonitor(workspaceName, submissionId, bucketCredential)
       scheduleNextCheckCurrentWorkflowStatus(child)
-      initGauge(workspaceName, submissionId)
+      registerDetailedJobExecGauges(workspaceName, submissionId)
 
     case SaveCurrentWorkflowStatusCounts(workspaceName, submissionId, workflowStatusCounts, submissionStatusCounts, reschedule) =>
       this.activeWorkflowStatusCounts += submissionId -> workflowStatusCounts
@@ -157,7 +159,7 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
     new ThresholdOneForOneStrategy(thresholdLimit = 3)(alwaysRestart)(thresholdFunc)
   }
 
-  private def initGlobalJobExecGauges(): Unit = {
+  private def registerGlobalJobExecGauges(): Unit = {
     try {
       SubmissionStatuses.allStatuses.foreach { status =>
         ExpandedMetricBuilder.expand(SubmissionStatusMetricKey, status).asGaugeIfAbsent("current") {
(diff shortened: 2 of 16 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -108,16 +147,33 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
       WorkflowStatuses.allStatuses.foreach { status =>
         workspaceSubmissionMetricBuilder(workspaceName, submissionId).expand(WorkflowStatusMetricKey, status).asGaugeIfAbsent("current") {
-          currentWorkflowStatusCounts.getOrElse(status, 0)
+          activeWorkflowStatusCounts.get(submissionId).getOrElse(status, 0)
         }
       }
       SubmissionStatuses.allStatuses.foreach { status =>
         workspaceMetricBuilder(workspaceName).expand(SubmissionStatusMetricKey, status).asGaugeIfAbsent("current") {
-          currentSubmissionStatusCounts.getOrElse(status, 0)
+          activeSubmissionStatusCounts.get(workspaceName).getOrElse(status, 0)
```
helgridly: these never get updated after the submission finishes. can we unregister the gauge then?

helgridly: i spent some time looking into "one last update and then unregister" but it was an annoying pain. it doesn't matter so much now that we have the globally-aggregated metrics; if we look at the workflow counts for a submission and they suddenly just stop, it's because the last one succeeded.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -168,6 +168,7 @@ object Boot extends App with LazyLogging {
     val submissionSupervisor = system.actorOf(SubmissionSupervisor.props(
       shardedExecutionServiceCluster,
       slickDataSource,
+      gcsDAO.getBucketServiceAccountCredential,
```
helgridly, line 171: (more on this later)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -26,56 +26,95 @@ import scala.util.control.NonFatal
 object SubmissionSupervisor {
   sealed trait SubmissionSupervisorMessage
 
-  case class SubmissionStarted(workspaceName: WorkspaceName, submissionId: UUID, credential: Credential)
+  case class SubmissionStarted(workspaceName: WorkspaceName, submissionId: UUID)
```
helgridly, line 31: It seemed stupid to pass in this credential every time a submission started, so I made it a constructor param instead.
@dvoet -- the credential itself doesn't expire, right?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -26,56 +26,95 @@ import scala.util.control.NonFatal
...
+   *
+   *  The fields below are marked volatile because they are read by a separate statsd thread.
+   *  They are only written by this actor.
+   */
   @volatile
-  private var currentWorkflowStatusCounts: Map[WorkflowStatus, Int] = Map.empty
+  private var activeSubmissionStatusCounts: Map[WorkspaceName, Map[SubmissionStatus, Int]] = Map.empty
   @volatile
-  private var currentSubmissionStatusCounts: Map[SubmissionStatus, Int] = Map.empty
+  private var activeWorkflowStatusCounts: Map[UUID, Map[WorkflowStatus, Int]] = Map.empty
```
rtitle, line 87: Proposal: maybe we should disable these metrics at the submission level. We're over our limit in Hosted Graphite and it looks like submission-level metrics account for a lot of it:

https://www.hostedgraphite.com/accounts/metrics/delete/preview/?pattern=*

We can keep the code that collects the data; maybe just disable creating the gauges in `initGauge`. WDYT?

Also it looks like there are some issues with the URI-redaction logic; I'll look at addressing those in a separate PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -26,56 +26,95 @@ import scala.util.control.NonFatal
...
+      this.activeWorkflowStatusCounts += submissionId -> workflowStatusCounts
+      this.activeSubmissionStatusCounts += workspaceName -> submissionStatusCounts
       if (reschedule) {
         scheduleNextCheckCurrentWorkflowStatus(sender)
+      } else {
+        //this submission is complete; save some metrics and some memory by removing its workflow status counter gauge
+        unregisterWorkflowGauges(workspaceName, submissionId)
+
+        //if all the submissions in this workspace are in terminal statuses, we can unregister the gauge for its submission count too
+        if( submissionStatusCounts.values.sum == submissionStatusCounts.filterKeys( SubmissionStatuses.terminalStatuses contains _ ).values.sum ) {
```
rtitle: Would this suffice?
```
if (submissionStatusCounts.keySet == SubmissionStatuses.terminalStatuses.toSet)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -86,7 +125,17 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
   }
 
   private def scheduleNextCheckCurrentWorkflowStatus(actor: ActorRef): Cancellable = {
-    system.scheduler.scheduleOnce(addJitter(submissionPollInterval), actor, CheckCurrentWorkflowStatusCounts)
+    system.scheduler.scheduleOnce(submissionPollInterval, actor, CheckCurrentWorkflowStatusCounts)
```
rtitle, line 136: 👍  on removing jitter, a relatively constant check period is probably better.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -86,7 +125,17 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
+  }
+
+  private def updateGlobalJobExecGauges = {
+    datasource.inTransaction { dataAccess =>
+      dataAccess.submissionQuery.countAllStatuses map { subStatuses =>
+        this.globalSubmissionStatusCounts ++= subStatuses.map{ case (k, v) => SubmissionStatuses.withName(k) -> v }
```
rtitle: This is doing a "get-and-set" from a separate thread, which is not safe even with volatile variables. I would suggest either:
a. Piping this future back to the actor similar to `SaveCurrentWorkflowStatusCounts`, and doing the update from the Actor thread (which is safe).
b. Making the maps ConcurrentMaps instead of volatile: https://www.scala-lang.org/api/current/scala/collection/concurrent/TrieMap.html

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -108,20 +157,51 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
+        ExpandedMetricBuilder.expand(WorkflowStatusMetricKey, status).asGaugeIfAbsent("current") {
+          globalWorkflowStatusCounts.getOrElse(status, 0)
+        }
+      }
+    } catch {
+      case NonFatal(e) => logger.warn(s"Could not initialize gauge metrics for jobexec", e)
+    }
+  }
+
   private def initGauge(workspaceName: WorkspaceName, submissionId: UUID): Unit = {
```
rtitle: Minor, maybe rename this `initDetailedJobExecGauges()`

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisorSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
+import org.scalatest.concurrent.Eventually
+
+import scala.concurrent.duration._
+
+class SubmissionSupervisorSpec(_system: ActorSystem) extends TestKit(_system) with FlatSpecLike with Matchers with TestDriverComponent with BeforeAndAfterAll with Eventually with RawlsTestUtils with MockitoTestUtils with RawlsStatsDTestUtils {
+
+  import driver.api._
+
+  def this() = this(ActorSystem("SubmissionSupervisorSpec"))
```
rtitle: Not a big deal, but I wonder why we define this extra constructor instead of:
```
class SubmissionSupervisorSpec extends TestKit(ActorSystem("SubmissionSupervisorSpec")) with Foo with Bar {
  ... 
}
```

rtitle (changes requested):
See comments above. I like the idea of unregistering gauges when the submission is done (i.e., only instrumenting _active_ workspaces & submissions).

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: I can add a config flag for creating these metrics, but I'd like to look into aging them out first.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: yes! thanks, i was thinking this was ridiculous as i wrote it

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: For a), you're specifically saying do the db call here but then dump the results into a message that we pipe back to self, right?

rtitle: yes

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
rtitle: ok, I can look into more the aging-out feature of hosted graphite.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
dvoet: I think that is correct, probably bears some testing though.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: Actually it turns out that terminal statuses has multiple statuses in it (Done and Aborted) so this doesn't work. But `(submissionStatusCounts.keySet -- SubmissionStatuses.terminalStatuses.toSet).isEmpty ` does.

helgridly:
@rtitle re-review please

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -86,7 +133,24 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
+        wfStatuses <- dataAccess.workflowQuery.countAllStatuses
+      } yield {
+        SaveGlobalJobExecCounts( subStatuses.map { case (k, v) => SubmissionStatuses.withName(k) -> v },
+                                 wfStatuses.map  { case (k, v) => WorkflowStatuses.withName(k) -> v } )
+      }
+    }
+  }
+
+  private def saveGlobalJobExecCounts(submissionStatuses: Map[SubmissionStatus, Int], workflowStatus: Map[WorkflowStatus, Int]) = {
+    this.globalSubmissionStatusCounts ++= submissionStatuses
```
rtitle: Wouldn't we want to replace this map instead of adding to it? I.e.:
```
this.globalSubmissionStatusCounts = submissionStatuses
this.globalWorkflowStatusCounts = workflowStatus
```
For example say there is 1 submission which transitions from Running to Done. With this code, we could end up with: `Map(Running -> 1, Done -> 1)` when what we want is `Map(Done -> 1)`.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -26,56 +28,101 @@ import scala.util.control.NonFatal
...
+      val child = startSubmissionMonitor(workspaceName, submissionId, bucketCredential)
       scheduleNextCheckCurrentWorkflowStatus(child)
-      initGauge(workspaceContext, submissionId)
+      registerDetailedJobExecGauges(workspaceName, submissionId)
 
-    case SaveCurrentWorkflowStatusCounts(workflowStatusCounts, submissionStatusCounts, reschedule) =>
-      this.currentWorkflowStatusCounts = workflowStatusCounts
-      this.currentSubmissionStatusCounts = submissionStatusCounts
+    case SaveCurrentWorkflowStatusCounts(workspaceName, submissionId, workflowStatusCounts, submissionStatusCounts, reschedule) =>
+      this.activeWorkflowStatusCounts += submissionId -> workflowStatusCounts
```
rtitle, line 107: `+=` is correct here (I think) -- we want to _replace_ the entry for the given workspace/submission with the current counts.

rtitle:
One more comment about replacing the global maps with the current values, instead of appending to them. Otherwise all looks good.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
helgridly: good catch (again). so much copypasta

rtitle (approved):
:shipit:

PR #784, opened 2017-09-13
GAWB-2570 Cromwell /logs endpoint response changed from under us

helgridly:
Reported by Jon B. :(


rtitle (approved):

rtitle:
Actually tests need fixing

helgridly:
(that last test failure failed on only one of the Travises because of a transient failure to download a package, not because i'm a bad person)

PR #791, opened 2017-09-21
GAWB-2450: Don't launch Submissions with invalid Expressions

jmthibault79:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionParser.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionParser.scala
@@ -1,34 +1,34 @@
 package org.broadinstitute.dsde.rawls.expressions
 
-import org.broadinstitute.dsde.rawls.model.{AttributeString, MethodConfiguration, ValidatedMethodConfiguration}
+import org.broadinstitute.dsde.rawls.model.{AttributeString, ParsedMCExpressions}
 
 import scala.util.{Failure, Success, Try}
-
 import spray.json._
+
+import cats.syntax.functor._
+import cats.instances.try_._
 
 // a thin abstraction layer over SlickExpressionParser
 
 object ExpressionParser {
-  def parseMCExpressions(methodConfiguration: MethodConfiguration, parser: SlickExpressionParser): ValidatedMethodConfiguration = {
+  def parseMCExpressions(inputs: Map[String, AttributeString], outputs: Map[String, AttributeString], parser: SlickExpressionParser): ParsedMCExpressions = {
     def parseAndPartition(m: Map[String, AttributeString], parseFunc:String => Try[Unit] ) = {
       val parsed = m map { case (key, attr) => (key, parseFunc(attr.value)) }
       ( parsed collect { case (key, Success(_)) => key } toSeq,
         parsed collect { case (key, Failure(regret)) => (key, regret.getMessage) } )
     }
-    val (successInputs, failedInputs)   = parseAndPartition(methodConfiguration.inputs, parseAttributeExpr(parser) )
-    val (successOutputs, failedOutputs) = parseAndPartition(methodConfiguration.outputs, parseOutputExpr(parser) )
+    val (successInputs, failedInputs)   = parseAndPartition(inputs, parseInputExpr(parser) )
+    val (successOutputs, failedOutputs) = parseAndPartition(outputs, parseOutputExpr(parser) )
 
-    ValidatedMethodConfiguration(methodConfiguration, successInputs, failedInputs, successOutputs, failedOutputs)
+    ParsedMCExpressions(successInputs, failedInputs, successOutputs, failedOutputs)
   }
 
-  private def parseAttributeExpr(parser: SlickExpressionParser)(expression: String): Try[Unit] = {
-    Try(expression.parseJson) match {
-      case Success(parsed) => Success(true)
-      case Failure(regret) => parser.parseAttributeExpr(expression) map (_ => ())
-    }
+  private def parseInputExpr(parser: SlickExpressionParser)(expression: String): Try[Unit] = {
+    // JSON expressions are valid inputs and do not need to be parsed
+    Try(expression.parseJson).recoverWith { case _ => parser.parseAttributeExpr(expression) }.void
   }
 
   private def parseOutputExpr(parser: SlickExpressionParser)(expression: String): Try[Unit] = {
-    parser.parseOutputExpr(expression) map (_ => ())
+    parser.parseOutputAttributeExpr(expression).void
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
@@ -1,0 +1,50 @@
+package org.broadinstitute.dsde.rawls.expressions
+
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
+import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
+import org.broadinstitute.dsde.rawls.model.{AttributeString, ErrorReport, MethodConfiguration, ValidatedMethodConfiguration}
+import spray.http.StatusCodes
+
+import scala.util.Try
+
+object ExpressionValidator {
+  // inputsToParse is parameterized in order to validate in the presence or absence of the associated Method
+  // presence: inputs which are both empty and optional are pre-validated, so they are skipped here
+  // absence: validate all inputs normally
+
+  private[expressions] def validateAndParse(methodConfiguration: MethodConfiguration, inputsToParse: Map[String, AttributeString], parser: SlickExpressionParser): ValidatedMethodConfiguration = {
+    val (emptyOutputs, outputsToParse) = methodConfiguration.outputs.partition { case (_, expr) => expr.value.isEmpty }
+
+    val parsed = ExpressionParser.parseMCExpressions(inputsToParse, outputsToParse, parser)
+
+    // empty output expressions are also valid
+    val validatedOutputs = emptyOutputs.keys.toSeq ++ parsed.validOutputs
+
+    ValidatedMethodConfiguration(methodConfiguration, parsed.validInputs, parsed.invalidInputs, validatedOutputs, parsed.invalidOutputs)
+  }
+
+  // validate a MC without retrieving the Method from the Repo: assume all inputs are required
+  def validateAndParseMCExpressions(methodConfiguration: MethodConfiguration, parser: SlickExpressionParser): ValidatedMethodConfiguration =
+    validateAndParse(methodConfiguration, methodConfiguration.inputs, parser)
+
+  // validate a MC, skipping optional empty inputs, and return failure when any inputs/outputs are invalid
+  def validateExpressionsForSubmission(methodConfiguration: MethodConfiguration, methodInputsToParse: Seq[MethodConfigResolver.MethodInput], emptyOptionalMethodInputs: Seq[MethodConfigResolver.MethodInput], parser: SlickExpressionParser): Try[ValidatedMethodConfiguration] = {
+    val inputsToParse = methodInputsToParse map { mi => (mi.workflowInput.fqn, AttributeString(mi.expression)) }
+
+    val validated = validateAndParse(methodConfiguration, inputsToParse.toMap, parser)
+
+    Try {
+      if (validated.invalidInputs.nonEmpty || validated.invalidOutputs.nonEmpty) {
+        val inputMsg = if (validated.invalidInputs.isEmpty) Seq() else Seq(s"Invalid inputs: ${validated.invalidInputs.mkString(",")}")
+        val outputMsg = if (validated.invalidOutputs.isEmpty) Seq() else Seq(s"Invalid outputs: ${validated.invalidOutputs.mkString(",")}")
+        val errorStr = (inputMsg ++ outputMsg) mkString " ; "
+        throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, s"Validation errors: $errorStr"))
+      }
+
+      // a MethodInput which is both optional and empty is already valid
+      val emptyOptionalInputs = emptyOptionalMethodInputs map { _.workflowInput.fqn }
+      validated.copy(validInputs = validated.validInputs ++ emptyOptionalInputs)
+    }
+  }
+
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
@@ -3,14 +3,15 @@ package org.broadinstitute.dsde.rawls.util
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.dataaccess._
+import org.broadinstitute.dsde.rawls.expressions.{ExpressionValidator, SlickExpressionParser}
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.MethodInput
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.webservice.PerRequest.PerRequestMessage
 import spray.http.StatusCodes
 
-import scala.concurrent.{Future, ExecutionContext}
-import scala.util.{Try, Failure, Success}
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 //Well, this is a joke.
 trait MethodWiths {
@@ -27,11 +28,11 @@ trait MethodWiths {
   }
 
   def withMethod[T](methodNamespace: String, methodName: String, methodVersion: Int, userInfo: UserInfo)(op: (AgoraEntity) => ReadWriteAction[T])(implicit executionContext: ExecutionContext): ReadWriteAction[T] = {
-    DBIO.from(methodRepoDAO.getMethod(methodNamespace, methodName, methodVersion, userInfo)).asTry.flatMap { agoraEntityOption => agoraEntityOption match {
+    DBIO.from(methodRepoDAO.getMethod(methodNamespace, methodName, methodVersion, userInfo)).asTry.flatMap {
       case Success(None) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound, s"Cannot get ${methodNamespace}/${methodName}/${methodVersion} from method repo.")))
       case Success(Some(agoraEntity)) => op(agoraEntity)
-      case Failure(throwable) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadGateway, s"Unable to query the method repo.",methodRepoDAO.toErrorReport(throwable))))
-    }}
+      case Failure(throwable) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadGateway, s"Unable to query the method repo.", methodRepoDAO.toErrorReport(throwable))))
+    }
   }
 
   def withWdl[T](method: AgoraEntity)(op: (String) => ReadWriteAction[T]): ReadWriteAction[T] = {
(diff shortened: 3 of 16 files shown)


jmthibault79:
ExpressionParser, ExpressionValidator, and ExpressionFixture updates
- withValidatedMCExpressions
- update withMethodInputs to partition out empty optional method inputs
- update SubmissionSpec
- update MethodConfigApiServiceSpec

jmthibault79:
TODO: FIAB/Local Stack testing

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
```diff
@@ -41,16 +42,28 @@ trait MethodWiths {
...
-      withWdl(method) { wdl => MethodConfigResolver.gatherInputs(methodConfig,wdl) match {
-        case Failure(exception) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, exception)))
-        case Success(methodInputs) => op(wdl,methodInputs)
-      }}
+      withWdl(method) { wdl =>
+        MethodConfigResolver.gatherInputs(methodConfig, wdl) match {
+          case Failure(exception) => DBIO.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, exception)))
+          case Success(methodInputs) =>
+            //Remove inputs that are both empty and optional
+            val (emptyOptionalInputs, inputsToProcess) = methodInputs.partition(input => input.workflowInput.optional && input.expression.isEmpty)
```
jmthibault79, line 54: This partition was in `withSubmissionParameters()` but I felt it made more sense here

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2200,25 +2200,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+    (op: (DataAccess, SlickWorkspaceContext, String, SubmissionValidationHeader, Seq[SubmissionValidationEntityInputs], Seq[SubmissionValidationEntityInputs], Option[WorkflowFailureMode]) => ReadWriteAction[PerRequestMessage]): Future[PerRequestMessage] = {
     dataSource.inTransaction { dataAccess =>
       withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Write, dataAccess) { workspaceContext =>
         withMethodConfig(workspaceContext, submissionRequest.methodConfigurationNamespace, submissionRequest.methodConfigurationName, dataAccess) { methodConfig =>
-          withMethodInputs(methodConfig, userInfo) { (wdl, methodInputs) =>
-            withSubmissionEntityRecs(submissionRequest, workspaceContext, methodConfig.rootEntityType, dataAccess) { jobEntityRecs =>
-              withWorkflowFailureMode(submissionRequest) { workflowFailureMode =>
-                //Remove inputs that are both empty and optional
-                val inputsToEvaluate = methodInputs.filter(input => !(input.workflowInput.optional && input.expression.isEmpty))
```
jmthibault79, line 2212: Moved this inside `withMethodInputs`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2200,25 +2200,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-                MethodConfigResolver.evaluateInputExpressions(workspaceContext, inputsToEvaluate, jobEntityRecs, dataAccess) flatMap { valuesByEntity =>
-                  valuesByEntity
-                    .map({ case (entityName, values) => SubmissionValidationEntityInputs(entityName, values) })
-                    .partition({ entityInputs => entityInputs.inputResolutions.forall(_.error.isEmpty) }) match {
-                    case (succeeded, failed) =>
-                      val methodConfigInputs = inputsToEvaluate.map { methodInput => SubmissionValidationInput(methodInput.workflowInput.fqn, methodInput.expression) }
-                      val header = SubmissionValidationHeader(methodConfig.rootEntityType, methodConfigInputs)
-                      op(dataAccess, workspaceContext, wdl, header, succeeded.toSeq, failed.toSeq, workflowFailureMode)
+          withMethodInputs(methodConfig, userInfo) { (wdl, inputsToProcess, emptyOptionalInputs) =>
+            withValidatedMCExpressions(methodConfig, inputsToProcess, emptyOptionalInputs, dataAccess) { _ =>
```
jmthibault79, line 2210: Here's the relevant new line

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -410,32 +412,6 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
       outputs=Map("aggregate_data_workflow.aggregate_data.output_array" -> AttributeString("this.output_array")),
       MethodRepoMethod("dsde", "array_task", 1))
 
-    val methodConfigValidExprs = MethodConfiguration("dsde", "GoodMethodConfig", "Sample", prerequisites=Map.empty,
```
jmthibault79, line 413: Tests that use these have been removed

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -401,7 +401,9 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
     val methodConfigMissingOutputs = MethodConfiguration("ns", "testConfigMissingOutputs", "Sample", Map(), Map(), Map("some.workflow.output" -> AttributeString("this.might_not_be_here")), MethodRepoMethod("ns-config", "meth1", 1))
 
     val methodConfigValid = MethodConfiguration("dsde", "GoodMethodConfig", "Sample", prerequisites=Map.empty, inputs=Map("three_step.cgrep.pattern" -> AttributeString("this.name")), outputs=Map.empty, MethodRepoMethod("dsde", "three_step", 1))
-    val methodConfigUnparseable = MethodConfiguration("dsde", "UnparseableMethodConfig", "Sample", prerequisites=Map.empty, inputs=Map("three_step.cgrep.pattern" -> AttributeString("this..wont.parse")), outputs=Map.empty, MethodRepoMethod("dsde", "three_step", 1))
+    val methodConfigUnparseableInputs = MethodConfiguration("dsde", "UnparseableInputsMethodConfig", "Sample", prerequisites=Map.empty, inputs=Map("three_step.cgrep.pattern" -> AttributeString("this..wont.parse")), outputs=Map.empty, MethodRepoMethod("dsde", "three_step", 1))
```
jmthibault79, line 404: Used in `SubmissionSpec`

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/MethodConfigApiServiceSpec.scala
```diff
@@ -102,9 +102,13 @@ class MethodConfigApiServiceSpec extends ApiServiceSpec {
         assertResult(StatusCodes.Created) {
           status
         }
-        assertResult(ValidatedMethodConfiguration(newMethodConfig, expectedSuccessInputs, expectedFailureInputs, expectedSuccessOutputs, expectedFailureOutputs)) {
```
jmthibault79, line 105: Can't do this because successful outputs may be in a different order

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionParser.scala
```diff
@@ -1,29 +1,28 @@
...
 import scala.util.{Failure, Success, Try}
-
 import spray.json._
 
 // a thin abstraction layer over SlickExpressionParser
 
 object ExpressionParser {
-  def parseMCExpressions(methodConfiguration: MethodConfiguration, parser: SlickExpressionParser): ValidatedMethodConfiguration = {
+  def parseMCExpressions(inputs: Map[String, AttributeString], outputs: Map[String, AttributeString], parser: SlickExpressionParser): ParsedMCExpressions = {
     def parseAndPartition(m: Map[String, AttributeString], parseFunc:String => Try[Unit] ) = {
```
rtitle, line 15: I feel like this could be achieved with cats' `Validated`. I might post some code showing how that would look, feel free to use or ignore.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionParser.scala
```diff
@@ -1,29 +1,28 @@
...
 
-    ValidatedMethodConfiguration(methodConfiguration, successInputs, failedInputs, successOutputs, failedOutputs)
+    ParsedMCExpressions(successInputs, failedInputs, successOutputs, failedOutputs)
   }
 
   private def parseAttributeExpr(parser: SlickExpressionParser)(expression: String): Try[Unit] = {
     Try(expression.parseJson) match {
-      case Success(parsed) => Success(true)
+      case Success(parsed) => Success(())
       case Failure(regret) => parser.parseAttributeExpr(expression) map (_ => ())
```
rtitle: A shorthand for `map(_ => ())` is `.void`. E.g.:
```
parser.parseAttributeExpr(expression).void
```
It needs the following imports:
```
import cats.syntax.functor._   // adds the .void syntax
import cats.instances.try_._   // the thing you're trying to void (could be future, option, etc) 
```

jmthibault79: Aha!  I wanted that but I couldn't figure out the right imports.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
```diff
@@ -0,0 +1,48 @@
...
+    val parsed = ExpressionParser.parseMCExpressions(inputsToParse, outputsToParse, parser)
+
+    // empty output expressions are also valid
+    val validatedOutputs = emptyOutputs.keys.toSeq ++ parsed.validOutputs
+
+    ValidatedMethodConfiguration(methodConfiguration, parsed.validInputs, parsed.invalidInputs, validatedOutputs, parsed.invalidOutputs)
+  }
+
+  // validate a MC without retrieving the Method from the Repo: assume all inputs are required
+  def validateAndParseMCExpressions(methodConfiguration: MethodConfiguration, parser: SlickExpressionParser): ValidatedMethodConfiguration =
```
rtitle, line 27: Maybe I don't know all the context here - but why are there 2 versions, one that throws an exception and one that doesn't? I understand why we wouldn't want to launch submissions for invalid expressions -- what's the use case for _not_ throwing an exception for invalid expressions?

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidatorSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+    inputs = toExpressionMap(parseableInputExpressions) + emptyExpr,
+    outputs = toExpressionMap(parseableOutputExpressions),
+    MethodRepoMethod("dsde", "three_step", 1))
+
+  "ExpressionValidator" should "validateAndParse" in {
+
+    val actualValid = ExpressionValidator.validateAndParse(allValid, allValid.inputs, this)
+    assertSameElements(parseableInputExpressions, actualValid.validInputs)
+    assertSameElements(parseableOutputExpressions, actualValid.validOutputs)
+    assert { actualValid.invalidInputs.isEmpty }
```
rtitle: `actualValid.invalidInputs shouldBe 'empty`?

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -382,15 +382,45 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system) with FlatSpe
...
+  it should "return a successful Submission when given a method configuration with empty outputs" in withWorkspaceService { workspaceService =>
+    val submissionRq = SubmissionRequest("dsde", "EmptyOutputsMethodConfig", "Individual", "indiv1", Some("this.sset.samples"), false)
     val rqComplete = Await.result(workspaceService.createSubmission( testData.wsName, submissionRq ), Duration.Inf).asInstanceOf[RequestComplete[(StatusCode, SubmissionReport)]]
     val (status, newSubmissionReport) = rqComplete.response
     assertResult(StatusCodes.Created) {
       status
     }
 
-    assert( newSubmissionReport.workflows.size == 0 )
+    assert( newSubmissionReport.workflows.size == 3 )
```
rtitle, line 439: `newSubmissionReport.workflows should have size 3`

rtitle:
LGTM

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
jmthibault79: In the general case, it is OK for a Method Config to exist with invalid expressions.  (I didn't realize that to be true until recently, and that false assumption is what drove GAWB-2444.)

It's about saving partial work, essentially.  If the MC has 40 inputs, we want the user to be able to fill in a few and come back later.  This no-fail validation will still inform them about what's missing but it won't block them.  If they later try to run something using the unfinished MC, then it will call the other validation and fail.

rtitle: I see, makes sense. Thanks

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
```diff
@@ -0,0 +1,48 @@
...
+import spray.http.StatusCodes
+
+import scala.util.Try
+
+object ExpressionValidator {
+  // inputsToParse is parameterized in order to validate in the presence or absence of the associated Method
+  // presence: inputs which are both empty and optional are pre-validated, so they are skipped here
+  // absence: validate all inputs normally
+
+  private[expressions] def validateAndParse(methodConfiguration: MethodConfiguration, inputsToParse: Map[String, AttributeString], parser: SlickExpressionParser) = {
```
ansingh7115: Should this return a ValidatedMethodConfiguration?

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
jmthibault79: I'm going to leave this one since it's the more common style for this file

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
jmthibault79: Yes.  Are you saying you want a return type signature?

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  // validate a MC, skipping optional empty inputs, and return failure when any inputs/outputs are invalid
+  def validateExpressionsForSubmission(methodConfiguration: MethodConfiguration, methodInputsToParse: Seq[MethodConfigResolver.MethodInput], emptyOptionalMethodInputs: Seq[MethodConfigResolver.MethodInput], parser: SlickExpressionParser): Try[ValidatedMethodConfiguration] = {
+    val inputsToParse = methodInputsToParse map { mi => (mi.workflowInput.fqn, AttributeString(mi.expression)) }
+
+    val validated = validateAndParse(methodConfiguration, inputsToParse.toMap, parser)
+
+    Try {
+      if (validated.invalidInputs.nonEmpty || validated.invalidOutputs.nonEmpty) {
+        val errors = s"Invalid inputs: ${validated.invalidInputs.mkString(",")} ; Invalid outputs: ${validated.invalidOutputs.mkString(",")}"
```
ansingh7115: Does this mean if I have 1 invalid input and 0 invalid outputs, my error string will look sort of like this:

`Invalid inputs: myinput ; Invalid outputs: `

Could we make it so that, in this sort of a scenario, it just reads like:

`Invalid inputs: myinput`

But maybe this is a tiny thing that doesn't super matter

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
ansingh7115: yes

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
jmthibault79: oh ok I suppose :)

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionParser.scala
rtitle: A rough sketch - was going to PR it but it touches a lot of other code. :) My idea was to model expression validation as a `ValidatedNel[MCValidationError, ParsedMCExpressions]`, where the RHS contains successful inputs/outputs, and the LHS is a list of validation errors.

Again, you don't need to do this for the PR, this was mostly for fun. :)

```
package org.broadinstitute.dsde.rawls.expressions

import cats.data.{Validated, ValidatedNel}
import cats.data.Validated.{Invalid, Valid}
import cats.implicits._
import org.broadinstitute.dsde.rawls.model.{AttributeString, InvalidInput, InvalidOutput, MCValidationError, ParsedMCExpressions}
import spray.json._

import scala.util.Try

// a thin abstraction layer over SlickExpressionParser

object ExpressionParser {

  // Would go in WorkspaceModel
  sealed trait MCValidationError {
    def key: String
    def message: String
  }
  case class InvalidInput(key: String, message: String) extends MCValidationError
  case class InvalidOutput(key: String, message: String) extends MCValidationError

  // Returns a ValidatedNel[MCValidationError, ParsedMCExpressions]
  // The right-hand-side contains only the successes, the left is a list of validation errors
  def parseMCExpressions(inputs: Map[String, AttributeString], outputs: Map[String, AttributeString], parser: SlickExpressionParser): ValidatedNel[MCValidationError, ParsedMCExpressions] = {
    val validatedInputs = parse(inputs, parseAttributeExpr(parser), InvalidInput.apply)
    val validatedOutputs = parse(outputs, parseOutputExpr(parser), InvalidOutput.apply)
    (validatedInputs |@| validatedOutputs).map((validInputs, validOutputs) => ParsedMCExpressions(validInputs, validOutputs))
  }

  private def parse(m: Map[String, AttributeString],
                    parseFunc: String => Validated[Throwable, Unit],
                    errorFunc: (String, String) => MCValidationError): ValidatedNel[MCValidationError, List[String]] = {
    // traverseU is the same oepration as traverse but works around some compiler bugs.
    // Might be removed in Scala 2.12/cats 1.0.
    m.toList.traverseU { case (key, attr) =>
      parseFunc(attr.value) match {
        case Valid(_) =>   Valid(key).toValidatedNel
        case Invalid(e) => Invalid(errorFunc(key, e.getMessage)).toValidatedNel
      }
    }
  }

  private def parseAttributeExpr(parser: SlickExpressionParser)(expression: String): Validated[Throwable, Unit] = {
    Validated.fromTry {
      Try(expression.parseJson).recoverWith { case _ =>
        parser.parseAttributeExpr(expression)
      }.void
    }
  }

  private def parseOutputExpr(parser: SlickExpressionParser)(expression: String): Validated[Throwable, Unit] = {
    Validated.fromTry(parser.parseOutputExpr(expression).void)
  }

}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    val parsed = ExpressionParser.parseMCExpressions(inputsToParse, outputsToParse, parser)
+
+    // empty output expressions are also valid
+    val validatedOutputs = emptyOutputs.keys.toSeq ++ parsed.validOutputs
+
+    ValidatedMethodConfiguration(methodConfiguration, parsed.validInputs, parsed.invalidInputs, validatedOutputs, parsed.invalidOutputs)
+  }
+
+  // validate a MC without retrieving the Method from the Repo: assume all inputs are required
```
helgridly, line 26: I'm getting a bit stuck on "assume all inputs are required" here, but maybe it's just phrasing. A syntactically invalid expression in an optional input is still a bad MC which shouldn't be launched...

jmthibault79:
~TODO~
- ~FIAB/Local Stack testing~
- ~Swagger~

Tested on FIAB.  I was able to save (with red validation error text) but unable to launch an analysis using a MC with invalid expressions.  Empty output expressions are also no longer incorrectly marked as invalid.

Swagger may not need any changes.  `POST` for  `/api/workspaces/{workspaceNamespace}/{workspaceName}/submissions` and `/api/workspaces/{workspaceNamespace}/{workspaceName}/submissions/validate` currently have the error description `400: Expression evaluation failed`.  This remains true!  However we may want to add something indicating that it's not just submission entity expressions as it was before, but now also MC input/output expressions.

Also: does this change warrant a message to the API list?

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionFixture.scala
```diff
@@ -0,0 +1,82 @@
+package org.broadinstitute.dsde.rawls.expressions
+
+// test data for Expression Validation and Parsing
+
+trait ExpressionFixture {
```
helgridly, line 5: this is a fantastic resource to use as a description of "what is an isn't valid". thank you for putting this together!

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionFixture.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  val parseableInputExpressions = Seq(
+    "this.gvcf",
+    "workspace.gvcf",
+    "workspace.library:cohort",
+    "this.library:cohort",
+    "this.arbitrary:whatever",
+    "workspace.arbitrary:whatever",
+    "this.case_sample.foo:ref.bar:attribute",
+    "workspace.yes.we.can",
```
helgridly, line 16: (for instance, i was about to say "wait no, workspace expressions can only evaluate direct attributes, not chains through other entities"... but then i looked up the parser and turns out i was wrong!)

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionFixture.scala
```diff
@@ -0,0 +1,82 @@
+package org.broadinstitute.dsde.rawls.expressions
+
+// test data for Expression Validation and Parsing
+
+trait ExpressionFixture {
+  // Inputs which will not validate, so they don't need to be handled by the parser
+  // ""
```
helgridly: i thought this was a comment relating to `parseableInputExpressions` at first. maybe move this closer to `unparseableInputExpressions` and rephrase to say something like "NOTE: empty strings will be rejected in the validation stage, so we don't test them here".

same for the empty-string outputs case below

helgridly (approved):
Thank you so much for your thoroughness digging into this, unpacking it, and putting it all back together in a more sensible shape. This is awesome.

My only philosophical question: `ValidatedMethodConfiguration` is very similar to `ParsedMCExpressions`. Would it make sense for VMC to contain an instance of PMCE? Or extend it or something?

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionValidator.scala
jmthibault79: Rob got tripped up on this too so maybe some rewording is in order.

There are two different validators here.

`validateAndParseMCExpressions` for MC /validate, create, save, get, import, etc.
It does not know about the Method because it did not query Agora, so it cannot know that any inputs are optional.  It doesn't fail anything, only reports which are invalid.

`validateExpressionsForSubmission` is for Create Submission and Validate Submission.  It does know about optional inputs because it has the Agora Method.  It does validate optional inputs if they are non-empty, and it does fail when any inputs/outputs are invalid.

jmthibault79:
Re: `ValidatedMethodConfiguration` and `ParsedMCExpressions`. I've never had good results making case classes inherit from each other.  Maybe there's a specific use case that it would improve, but I can't think of it so I'd rather leave that alone.

PR #800, opened 2017-10-13
GAWB-2690: Database triggered submission monitoring (Prep for horizontally scaling Rawls)

MatthewBemis:
This PR is the first step in making rawls horizontally scalable.

Presently, when rawls receives a request to create a new submission, rawls adds it to the database and spawns an actor that is in charge of monitoring the status of that submission. It polls every `submissionPollInterval` milliseconds, and when the submission is finally complete, the actor is destroyed. 

In the new world, when rawls receives a request, it only adds it to the database. The submission supervisor will query the database every `submissionPollInterval` milliseconds for any active submissions. The supervisor will then create a child actor for any submission that does not yet have an actor monitoring it. The rest of the story is the same.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
@@ -16,7 +16,7 @@ import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
 import org.broadinstitute.dsde.rawls.model.{SubmissionStatuses, WorkflowStatuses, WorkspaceName}
-import org.broadinstitute.dsde.rawls.util.ThresholdOneForOneStrategy
+import org.broadinstitute.dsde.rawls.util.{ThresholdOneForOneStrategy, addJitter}
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
@@ -29,6 +29,8 @@ object SubmissionSupervisor {
   sealed trait SubmissionSupervisorMessage
 
   case class SubmissionStarted(workspaceName: WorkspaceName, submissionId: UUID)
+  case object StartMonitorPass
+  case object SubmissionMonitorPassComplete
 
   /**
     * A periodic collection of all current workflow and submission status counts.
@@ -93,15 +95,21 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
   override def preStart(): Unit = {
     super.preStart()
 
+    scheduleInitialMonitorPass
     registerGlobalJobExecGauges()
     system.scheduler.schedule(0 seconds, submissionPollInterval, self, RefreshGlobalJobExecGauges)
   }
 
   override def receive = {
+    case StartMonitorPass =>
+      startMonitoringNewSubmissions()
     case SubmissionStarted(workspaceName, submissionId) =>
       val child = startSubmissionMonitor(workspaceName, submissionId, bucketCredential)
       scheduleNextCheckCurrentWorkflowStatus(child)
       registerDetailedJobExecGauges(workspaceName, submissionId)
+
+    case SubmissionMonitorPassComplete =>
+      scheduleNextMonitorPass
 
     case SaveCurrentWorkflowStatusCounts(workspaceName, submissionId, workflowStatusCounts, submissionStatusCounts, reschedule) =>
       this.activeWorkflowStatusCounts += submissionId -> workflowStatusCounts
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -151,6 +168,25 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
+          val wsName = WorkspaceName(activeSub.workspaceNamespace, activeSub.workspaceName)
+          val subId = activeSub.submission.submissionId
+
+          self ! SubmissionStarted(wsName, UUID.fromString(subId))
+        }
+      }
+    } onFailure {
+      case t: Throwable => logger.error("Error starting submission monitor actors for new submissions", t)
+    }
+    self ! SubmissionMonitorPassComplete
```
dvoet: this last line executes outside the inTransaction future, so this is actually a lie. The monitor pass is not complete.

A better pattern is for this function to return `Future[SubmissionMonitorPassComplete]` and then the receive function would be like ```case StartMonitorPass => startMonitoringNewSubmissions() pipe self```

dvoet (changes requested):
you can also remove the code that looks for submissions on boot.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -125,6 +133,15 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
       saveGlobalJobExecCounts(submissionStatuses, workflowStatuses)
   }
 
+  private def scheduleInitialMonitorPass: Cancellable = {
+    //Wait anything _up to_ the poll interval for a much wider distribution of submission monitor start times when Rawls starts up
+    system.scheduler.scheduleOnce(addJitter(0 seconds, submissionPollInterval), self, StartMonitorPass)
```
rtitle: I'm not sure this needs jitter. There's only 1 `SubmissionSupervisor` actor, so this is not happening concurrently. Might as well make it run at startup and then at regular intervals.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -151,6 +168,28 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
     }
   }
 
+  private def startMonitoringNewSubmissions: Future[SubmissionMonitorPassComplete.type] = {
+    datasource.inTransaction { dataAccess =>
+      dataAccess.submissionQuery.listAllActiveSubmissions() map { activeSubs =>
+        val monitoredSubmissions = context.children.map(_.path.name).toSet
```
rtitle: I don't think you should be closing over `context` inside a future. It's mutable state within the actor, so it should only be accessed by the actor itself. The akka docs warn against this:
```   
* WARNING: Only valid within the Actor itself, so do not close over it and
* publish it to other threads!
```

It would be better to compute `monitoredSubmissions` outside of the database transaction (it doesn't need the DB results anyway).

(Note: `self` is immutable so doing `self ! msg` inside a future is fine.)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -151,6 +168,28 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
...
+        val unmonitoredSubmissions = activeSubs.filterNot(sub => monitoredSubmissions.contains(sub.submission.submissionId))
+
+        unmonitoredSubmissions.map { activeSub =>
```
rtitle: Minor: could make this `foreach` because you don't need the result.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -125,6 +133,15 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
       saveGlobalJobExecCounts(submissionStatuses, workflowStatuses)
   }
 
+  private def scheduleInitialMonitorPass: Cancellable = {
+    //Wait anything _up to_ the poll interval for a much wider distribution of submission monitor start times when Rawls starts up
+    system.scheduler.scheduleOnce(addJitter(0 seconds, submissionPollInterval), self, StartMonitorPass)
+  }
+
+  private def scheduleNextMonitorPass: Cancellable = {
+    system.scheduler.scheduleOnce(addJitter(submissionPollInterval), self, StartMonitorPass)
```
rtitle: Just wondering if we want a separate config for this. `submissionPollInterval` defines how often we poll for a single submission, and is currently set to 1 minute. Should we be calling `listAllActiveSubmissions` less frequently? (I'm not sure how expensive that query is.)

rtitle (changes requested):
A few comments above. I like the idea in general. :)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
MatthewBemis: That's a good point. It looks like this `listAllActiveSubmissions` query returns way more info than we actually need. For all running submissions it will load: the method config, all of the workflows, the root entity, and the workspace. The only info that the monitor even uses are submission.id, workspace.name, and workspace.namespace. It doesn't look like it even needs the name and namespace (except for logging errors), so maybe workspace.id would suffice for logging and we can get this down to a simple `select ID, WORKSPACE_ID from SUBMISSION where STATUS in [active_statuses]`

MatthewBemis: Actually it looks like metrics rely on workspace name/namespace so I won't screw with that bit at the moment, but it'll still be an improvement to not load all of the workflows.

dvoet (approved):

rtitle (approved):
Looks good!

PR #806, opened 2017-10-19
close jndi search results

dvoet:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiSupport.scala
@@ -10,6 +10,7 @@ import slick.dbio.DBIO
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
+import scala.collection.JavaConverters._
 
 trait JndiSupport extends LazyLogging {
   private val batchSize = 1000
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiSupport.scala
```diff
@@ -34,8 +36,31 @@ trait JndiSupport extends LazyLogging {
...
+      handleSearchResults(op, ctx.search(name, matchingAttributes, attributesToReturn))
+    }
+  }
+
+  protected def withSearchResults[T](ctx: DirContext, name: String, filter: String, cons: SearchControls)(op: Iterator[SearchResult] => T): T = {
+    handleSearchResults(op, ctx.search(name, filter, cons))
+  }
+
+  private def handleSearchResults[T](op: (Iterator[SearchResult]) => T, results: NamingEnumeration[SearchResult]) = {
+    // need to copy the results out of the jndi enum otherwise scala is fancy and continues to use the
```
MatthewBemis, line 52: 😑

MatthewBemis (approved):
at least the Jndi grossness is contained in one place. hopefully unboundid continues to look promising

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/jndi/JndiSupport.scala
```diff
@@ -34,8 +36,31 @@ trait JndiSupport extends LazyLogging {
...
+    // enum behind the scenes which is a problem once the enum is closed :(
+    val b = new mutable.ListBuffer[SearchResult]
```
breilly2, line 54: Thank you for explaining the need to copy the results.

Why use `mutable.ListBuffer`? Then why pass `.toIterator` to `op`? Many ops then `.toSeq` the results and then match on `Seq()` and `Seq(something)`. Does it make sense to shortcut some of the conversions (even just for readability) and then match on `Nil` and `something :: Nil` thereby hopefully only caring about the specific collection types when it truly matters?

I'm mainly interested in the readability of callers of `withSearchResults`. As it is now, I look at `.toSeq` and wonder what the type is that it needs to be changed which then leads me to this function and few answers. If callers could be written more generally, I wouldn't even have to wonder and end up here.

(This is as much of a question as it is a suggestion because I'm not sure what I'm proposing would actually work.)

dvoet: generally I was going for a drop-in replacement. The asScala in the old code resulted in an Iterator so that informed the current type. Using the ListBuffer came later when I ran into problems accessing the collection after the underlying ldap enum was closed. ListBuffer seemed better than the other option-ArrayBuffer. I have been told that Seqs are better than Lists. Pattern matching on Seq is just as valid as Lists.

PR #807, opened 2017-10-23
GAWB-2692: Add monitor "switch" to distinguish between front and back rawls instances

MatthewBemis:
#800 was the first step, and now this allows us to selectively boot up that process.

Config changes:

https://github.com/broadinstitute/firecloud-develop/pull/740 (dev)
https://github.com/broadinstitute/firecloud-develop/pull/743 (local)
https://github.com/broadinstitute/firecloud-develop/pull/744 (fiab)

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -184,7 +184,10 @@ object Boot extends App with LazyLogging {
       gcsConfig.getInt("groupMonitor.workerCount"),
       userServiceConstructor))
 
-    BootMonitors.restartMonitors(slickDataSource, gcsDAO, submissionSupervisor, bucketDeletionMonitor)
+    if(sys.env.getOrElse("BACK_RAWLS", "false").toBoolean) {
```
MatthewBemis: my assumption here is that when jenkins is deploying multiple instances of rawls it will pick one to be back-rawls and set an env var `BACK_RAWLS = true`

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala
```diff
@@ -95,7 +95,6 @@ class SubmissionSupervisor(executionServiceCluster: ExecutionServiceCluster,
   override def preStart(): Unit = {
     super.preStart()
 
-    scheduleInitialMonitorPass
```
MatthewBemis: instead of starting monitoring here, i moved it back to BootMonitors so we could easily control it from Boot.scala

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -184,7 +184,10 @@ object Boot extends App with LazyLogging {
...
+      logger.info("This instance has been marked as BACK_RAWLS. Booting monitors...")
+      BootMonitors.restartMonitors(slickDataSource, gcsDAO, submissionSupervisor, bucketDeletionMonitor)
```
MatthewBemis: quirk: we still create the supervisor for submission monitoring but it will be unused for front-rawls. 

it felt cleaner and less quirky than having the supervisor missing entirely from front-rawlses

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
dvoet: you should read from config, if we decide to use env vars or system properties later we can still feed them in through config... let's keep all our config in the same spot

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
dvoet: I am not sure I agree... why is it cleaner and less quirky?

MatthewBemis: i think it's cleaner because there's less logic separating the two types of rawls. every rawls ends up being constructed the same, it's just a matter of flipping the switch on the monitors or not

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -1,22 +1,78 @@
...
 import scala.concurrent.Await
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
+import scala.util.Try
 
 // handles monitors which need to be started at boot time
 object BootMonitors extends LazyLogging {
 
-  def restartMonitors(dataSource: SlickDataSource, gcsDAO: GoogleServicesDAO, submissionSupervisor: ActorRef, bucketDeletionMonitor: ActorRef): Unit = {
+  def bootMonitors(system: ActorSystem, conf: Config, slickDataSource: SlickDataSource, gcsDAO: HttpGoogleServicesDAO,
```
MatthewBemis, line 23: code contained in `bootMonitors` is the new way of doing things, code in `restartMonitors` is the old way of doing things. soon all of the code in `restartMonitors` will be migrated into `bootMonitors`

jmthibault79:
Seems legit.

One question about ordering: Back-Rawls will now `resetLaunchingWorkflows` before starting `SubmissionSupervisor`, reversing the previous order.  Does that matter?

MatthewBemis:
@jmthibault79 good catch. flipping the order was unintentional. i looked deeper into the code and i think it will still work fine. even though we create the workflow submission actors after resetting the launching workflows, those actors will still pick up the reset workflows in their next pass.

i will test further and switch back if it's wrong but i think it's okay

jmthibault79:
The new order seems to make more sense than the previous to me.

dvoet (approved):

PR #818, opened 2017-11-08
GAWB 2647- Switcheroo

vkumra-broad:
GAWB 2647 - Switch from pet SA to user account

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -42,7 +42,7 @@ class HttpSamDAO(baseSamServiceURL: String)(implicit val system: ActorSystem) ex
     val url = samServiceURL + "/register/user"
     retry(when500) { () =>
       pipeline[Option[UserStatus]](userInfo) apply Get(url) recover {
-        case notOK: UnsuccessfulResponseException if StatusCodes.Conflict == notOK.response.status => None
+        case notOK: UnsuccessfulResponseException if StatusCodes.NotFound == notOK.response.status => None
       }
     }
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -37,6 +37,17 @@ class HttpSamDAO(baseSamServiceURL: String)(implicit val system: ActorSystem) ex
     }
   }
 
+
+  override def getUserStatus(userInfo: UserInfo): Future[Option[UserStatus]] = {
+    val url = samServiceURL + "/register/user"
+    retry(when500) { () =>
+      pipeline[Option[UserStatus]](userInfo) apply Get(url) recover {
+        case notOK: UnsuccessfulResponseException if StatusCodes.Conflict == notOK.response.status => None
+      }
+    }
+  }
+
+
   private def when500( throwable: Throwable ): Boolean = {
     throwable match {
       case ure: spray.client.UnsuccessfulResponseException => ure.responseStatus.intValue / 100 == 5
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
@@ -1,0 +1,24 @@
+package org.broadinstitute.dsde.rawls.dataaccess.slick
+
+import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
+import org.broadinstitute.dsde.rawls.model.{RawlsUser, RawlsUserEmail, RawlsUserSubjectId, SubsystemStatus, UserInfo, UserStatus}
+
+import scala.concurrent.Future
+
+class MockSamDAO(ok: Boolean = true) extends SamDAO {
+  val serviceAccount = RawlsUserEmail("pet-1234567890@test-project.iam.gserviceaccount.com")
+  val userAccount = UserStatus(RawlsUser(RawlsUserSubjectId("1234567890"),RawlsUserEmail("valid-user")), enabled = Map[String, Boolean]())
+
+  override def getStatus(): Future[SubsystemStatus] = {
+    if (ok) Future.successful(SubsystemStatus(true, None))
+    else Future.successful(SubsystemStatus(false, None))
+  }
+
+  override def getUserStatus(userInfo: UserInfo):Future[Option[UserStatus]] = {
+    if(ok) Future.successful(Some(userAccount))
+    else Future.successful(None)
+  }
+
+
+  override def registerUser(userInfo: UserInfo): Future[Option[UserStatus]] = ???
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
@@ -2,12 +2,11 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
 import org.broadinstitute.dsde.rawls.model.{RawlsUser, RawlsUserEmail, RawlsUserSubjectId, SubsystemStatus, UserInfo, UserStatus}
-
 import scala.concurrent.Future
 
 class MockSamDAO(ok: Boolean = true) extends SamDAO {
-  val serviceAccount = RawlsUserEmail("pet-1234567890@test-project.iam.gserviceaccount.com")
-  val userAccount = UserStatus(RawlsUser(RawlsUserSubjectId("1234567890"),RawlsUserEmail("valid-user")), enabled = Map[String, Boolean]())
+  val userAccount = UserStatus(RawlsUser(RawlsUserSubjectId("123456789876543210101"),RawlsUserEmail("project-owner-access")), enabled = Map[String, Boolean]())
+  val notPetServiceAccount = UserStatus(RawlsUser(RawlsUserSubjectId("123456789876543210202"),RawlsUserEmail("project-owner-access-sa")), enabled = Map[String, Boolean]())
 
   override def getStatus(): Future[SubsystemStatus] = {
     if (ok) Future.successful(SubsystemStatus(true, None))
@@ -15,7 +14,14 @@ class MockSamDAO(ok: Boolean = true) extends SamDAO {
   }
 
   override def getUserStatus(userInfo: UserInfo):Future[Option[UserStatus]] = {
-    if(ok) Future.successful(Some(userAccount))
+    if(ok) {
+      if(userInfo.userEmail.value.startsWith("pet-")) {
+        Future.successful(Some(userAccount))
+      }
+      else {
+        Future.successful(Some(notPetServiceAccount))
+      }
+    }
     else Future.successful(None)
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
@@ -20,5 +20,8 @@ class MockSamDAO(ok: Boolean = true) extends SamDAO {
   }
 
 
-  override def registerUser(userInfo: UserInfo): Future[Option[UserStatus]] = ???
+  override def registerUser(userInfo: UserInfo): Future[Option[UserStatus]] = {
+    if(ok) Future.successful(Some(userAccount))
+    else Future.successful(None)
+  }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
@@ -1,23 +1,82 @@
 package org.broadinstitute.dsde.rawls.openam
 
-import org.broadinstitute.dsde.rawls.model.{RawlsUserEmail, RawlsUserSubjectId, UserInfo}
+import akka.actor.ActorSystem
+import com.typesafe.config.ConfigFactory
+import org.broadinstitute.dsde.rawls.dataaccess.DataSource
+import org.broadinstitute.dsde.rawls.dataaccess.jndi.{DirectoryConfig, JndiDirectoryDAO}
+import org.broadinstitute.dsde.rawls.dataaccess.slick.ReadWriteAction
+import org.broadinstitute.dsde.rawls.model.{RawlsUser, RawlsUserEmail, RawlsUserRef, RawlsUserSubjectId, UserInfo}
+import org.broadinstitute.dsde.rawls.user.UserService
+import shapeless.HNil
+import slick.backend.DatabaseConfig
+import slick.driver.JdbcDriver
 import spray.http.{HttpHeader, OAuth2BearerToken}
+import spray.routing
 import spray.routing.Directive1
 import spray.routing.Directives._
 
-trait StandardUserInfoDirectives extends UserInfoDirectives {
+import scala.concurrent.{ExecutionContext, Future}
 
-  def requireUserInfo(): Directive1[UserInfo] = {
-    for(accessToken <- accessTokenHeaderDirective;
-        userEmail <- emailHeaderDirective;
-        accessTokenExpiresIn <- accessTokenExpiresInHeaderDirective;
-        userSubjectId <- userSubjectIdDirective
+trait StandardUserInfoDirectives extends UserInfoDirectives with JndiDirectoryDAO {
+
+  val userServiceConstructor: UserInfo => UserService
+
+  //  def requireUserInfo(): Directive1[UserInfo] = {
+  //    val userInfo= for (accessToken <- accessTokenHeaderDirective;
+  //        userEmail <- emailHeaderDirective;
+  //        accessTokenExpiresIn <- accessTokenExpiresInHeaderDirective;
+  //        userSubjectId <- userSubjectIdDirective;
+  //    ) yield UserInfo(RawlsUserEmail(userEmail), OAuth2BearerToken(accessToken), accessTokenExpiresIn.toLong, RawlsUserSubjectId(userSubjectId))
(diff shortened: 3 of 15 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,52 @@
...
+import org.broadinstitute.dsde.rawls.dataaccess.jndi.JndiDirectoryDAO
+import org.broadinstitute.dsde.rawls.model.{RawlsUser, RawlsUserEmail, RawlsUserRef, RawlsUserSubjectId, UserInfo, UserStatus}
 import spray.http.{HttpHeader, OAuth2BearerToken}
 import spray.routing.Directive1
 import spray.routing.Directives._
 
-trait StandardUserInfoDirectives extends UserInfoDirectives {
+import scala.concurrent.Future
 
+trait StandardUserInfoDirectives extends UserInfoDirectives with JndiDirectoryDAO {
```
rtitle: I don't think you should extend `JndiDirectoryDAO` since this is calling Sam.

It seems the only thing `JndiDirectoryDAO` is used for is the implicit `ExecutionContext`. Removing the `with JndiDirectoryDAO` and adding `import scala.concurrent.ExecutionContext.Implicits.global` works for me.

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,52 @@
...
+
+  val httpSamDAO: SamDAO
+
+  //TODO: project should be regex
```
rtitle: Remove this comment?

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,52 @@
...
+  val petSAdomain = "\\S+@\\S+.iam.gserviceaccount.com"
```
rtitle: Same regex comments as in https://github.com/broadinstitute/sam/pull/48

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,52 @@
...
+  val petEmailPrefix = "pet-"
```
rtitle: `petEmailPrefix` is not used

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,52 @@
...
   def requireUserInfo(): Directive1[UserInfo] = {
-    for(accessToken <- accessTokenHeaderDirective;
-        userEmail <- emailHeaderDirective;
-        accessTokenExpiresIn <- accessTokenExpiresInHeaderDirective;
-        userSubjectId <- userSubjectIdDirective
+
+    val userInfo = for(
```
rtitle: I know this was like this before, but can we change this for comprehension to:
```
    val userInfo = for {
      accessToken <- accessTokenHeaderDirective
      userEmail <- emailHeaderDirective
      accessTokenExpiresIn <- accessTokenExpiresInHeaderDirective
      userSubjectId <- userSubjectIdDirective
    } yield UserInfo(RawlsUserEmail(userEmail), OAuth2BearerToken(accessToken), accessTokenExpiresIn.toLong, RawlsUserSubjectId(userSubjectId))
```
?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
```diff
@@ -21,34 +21,40 @@ trait BillingApiService extends HttpService with PerRequestCreator with UserInfo
 
   val userServiceConstructor: UserInfo => UserService
 
-  val billingRoutes = requireUserInfo() { userInfo =>
-    path("billing" / Segment / "members") { projectId =>
-      get {
-        requestContext => perRequest(requestContext,
-          UserService.props(userServiceConstructor, userInfo),
-          UserService.GetBillingProjectMembers(RawlsBillingProjectName(projectId)))
+  val billingRoutes = detach() {
```
rtitle: What is `detach()`, and why does it need to go on every route?

rtitle (changes requested):
See above

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
vkumra-broad: detach is used to execute the route in a future, I added it to every route that uses requireUserInfo since that makes a call out to sam.

http://spray.io/documentation/1.2.4/spray-routing/execution-directives/detach/

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
vkumra-broad: Is one better than the other?

helgridly: Semicolons aren't really idiomatic Scala. It's just a style thing.

rtitle: yeah ^

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/UserApiService.scala
```diff
@@ -23,126 +23,147 @@ trait UserApiService extends HttpService with PerRequestCreator with UserInfoDir
...
-          UserService.CreateUser)
+  val createUserRoute = detach() {
+    requireUserInfo() { userInfo =>
+      path("user") {
+        detach() {
+          post {
+            requestContext =>
+              perRequest(requestContext,
+                UserService.props(userServiceConstructor, userInfo),
+                UserService.CreateUser)
```
helgridly: You should probably note here that we need to switcheroo here not because pets will ever be in a situation where they'll _need_ to create their owners, but that this endpoint needs to behave correctly if a pet _does_ call this endpoint. If you didn't do a switcheroo here the pet would get created as a new user of its own.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/UserApiService.scala
```diff
@@ -23,126 +23,147 @@ trait UserApiService extends HttpService with PerRequestCreator with UserInfoDir
...
+  val userRoutes = detach() {
+    requireUserInfo() { userInfo =>
+      path("user" / "refreshToken") {
+        put {
+          entity(as[UserRefreshToken]) { token =>
+            requestContext =>
+              perRequest(requestContext,
+                UserService.props(userServiceConstructor, userInfo),
+                UserService.SetRefreshToken(token))
+          }
```
helgridly: note to other reviewers: this giant blob of diff for the rest of this file is because GitHub gets confused by whitespace changes. There's no code changes here; appending ?w=1 to the end of your URL will confirm.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -22,126 +22,142 @@ trait WorkspaceApiService extends HttpService with PerRequestCreator with UserIn
...
+        post {
+          entity(as[WorkspaceRequest]) { workspace =>
+            requestContext =>
+              perRequest(requestContext,
+                WorkspaceService.props(workspaceServiceConstructor, userInfo),
+                WorkspaceService.CreateWorkspace(workspace))
+          }
+        } ~
+          get {
+            requestContext => perRequest(requestContext, WorkspaceService.props(workspaceServiceConstructor, userInfo), WorkspaceService.ListWorkspaces)
```
helgridly: more whitespace-only changes

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/PetSASpec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def withTestDataApiServices[T](testCode: TestApiService => T): T = {
+    withDefaultTestDatabase { dataSource: SlickDataSource =>
+      withApiServices(dataSource)(testCode)
+    }
+  }
+
+  it should "switch to User Account when Accessed from pet SA" in withMinimalTestDatabase { dataSource: SlickDataSource =>
+    withApiServices(dataSource) { services =>
+      val petSA = UserInfo(RawlsUserEmail("pet-123456789876543212345@gserviceaccount.com"), OAuth2BearerToken("token"), 123, RawlsUserSubjectId("123456789876"))
```
helgridly: @gserviceaccount.com isn't the right shape and the regex should barf here, no?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/PetSASpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      runAndWait(dataSource.dataAccess.rawlsUserQuery.createUser(testData.userOwner))
+      Post("/user") ~> addHeader("OIDC_access_token", petSA.accessToken.value) ~> addHeader("OIDC_CLAIM_expires_in", petSA.accessTokenExpiresIn.toString) ~> addHeader("OIDC_CLAIM_email", petSA.userEmail.value) ~> addHeader("OIDC_CLAIM_user_id", petSA.userSubjectId.value) ~> services.sealedInstrumentedRoutes ~>
+        check {
+          assertResult(StatusCodes.Created) {
+            status
+          }
+        }
+    }
+  }
+
```
helgridly: another test: it should behave correctly when accessed by a SA who isn't a pet but a real, grown-up FC user

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
rtitle: f***ing spray. Ok thanks

helgridly (changes requested):
at least one more test, plus some nitpicks

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/PetSASpec.scala
rtitle: Yeah.. it seems like the regex would not match here and the switch would not happen. So this test is probably _creating_ the petSA as a real user? I wonder if there is a different API we can use to better test the switch. Maybe:

1. Create a workspace as a user
2. Get the workspace as the pet
3. Verify the workspace in #1 is returned

rtitle:
My original comments were addressed; my next request is to improve the testing in `PetSASpec` to verify the switch actually occurs.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
jmthibault79: Could it be built in to `requireUserInfo` since it will always be required by it?

vkumra-broad: I tried doing it that way but couldn't find a way to do so.

rtitle (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
jmthibault79: try changing the beginning of the definition to this: `def requireUserInfo(): Directive1[UserInfo] = detach() & {`

vkumra-broad: @jmthibault79  I am getting a warning when I make the above change.

```Users/vkumra/Documents/firecloud/rawls/core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala:25: Adaptation of argument list by inserting () has been deprecated: this is unlikely to be what you want.
[warn]         signature: ExecutionDirectives.detach(dm: spray.routing.directives.DetachMagnet): spray.routing.Directive0
[warn]   given arguments: <none>
[warn]  after adaptation: ExecutionDirectives.detach((): Unit)
[warn]   def requireUserInfo(): Directive1[UserInfo] = detach() & {
[warn]                                                       ^
[warn] there were 28 feature warnings; re-run with -feature for details
[warn] two warnings found
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MockSamDAO.scala
```diff
@@ -0,0 +1,33 @@
+package org.broadinstitute.dsde.rawls.dataaccess.slick
```
dvoet: why isn't this in core/src/test?

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,57 @@
...
+import scala.concurrent.Future
 
 trait StandardUserInfoDirectives extends UserInfoDirectives {
 
+  val httpSamDAO: SamDAO
+
+
+  val petSAdomain = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
+
+  private def isPetSA(email: String) = {
```
dvoet: technically, this is just isSA, don't know if it is a pet or not

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,57 @@
...
+    userInfo flatMap { ui =>
+      onSuccess(getWorkbenchUserEmailId(ui)).map {
+        case Some(resourceType) => UserInfo(resourceType.userInfo.userEmail, ui.accessToken, ui.accessTokenExpiresIn, resourceType.userInfo.userSubjectId)
+        case None => UserInfo(ui.userEmail, ui.accessToken, ui.accessTokenExpiresIn, ui.userSubjectId)
+      }
+    }
+  }
+
+  private def getWorkbenchUserEmailId(userInfo:UserInfo):Future[Option[UserStatus]] = {
+    if (isPetSA(userInfo.userEmail.value))
```
dvoet: use curly braces in your if then else

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,57 @@
...
+      httpSamDAO.getUserStatus(userInfo)
+    else
+      Future(None)
```
dvoet: Future.successful

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,57 @@
...
+    val userInfo = for {
+      accessToken <- accessTokenHeaderDirective
+      userEmail <- emailHeaderDirective
+      accessTokenExpiresIn <- accessTokenExpiresInHeaderDirective
+      userSubjectId <- userSubjectIdDirective
+    } yield UserInfo(RawlsUserEmail(userEmail), OAuth2BearerToken(accessToken), accessTokenExpiresIn.toLong, RawlsUserSubjectId(userSubjectId))
+
+    userInfo flatMap { ui =>
+      onSuccess(getWorkbenchUserEmailId(ui)).map {
+        case Some(resourceType) => UserInfo(resourceType.userInfo.userEmail, ui.accessToken, ui.accessTokenExpiresIn, resourceType.userInfo.userSubjectId)
```
dvoet: the variable name resourceType does not make sense.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/UserApiService.scala
```diff
@@ -25,10 +25,13 @@ trait UserApiService extends HttpService with PerRequestCreator with UserInfoDir
 
   val createUserRoute = requireUserInfo() { userInfo =>
     path("user") {
-      post {
-        requestContext => perRequest(requestContext,
-          UserService.props(userServiceConstructor, userInfo),
-          UserService.CreateUser)
+      detach() {
```
dvoet: why do you need detach here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -37,6 +37,17 @@ class HttpSamDAO(baseSamServiceURL: String)(implicit val system: ActorSystem) ex
     }
   }
 
+
+  override def getUserStatus(userInfo: UserInfo): Future[Option[UserStatus]] = {
+    val url = samServiceURL + "/register/user"
+    retry(when500) { () =>
+      pipeline[Option[UserStatus]](userInfo) apply Get(url) recover {
+        case notOK: UnsuccessfulResponseException if StatusCodes.Conflict == notOK.response.status => None
```
dvoet: should this handle Not_Found instead of Conflict? I don't think /register/user ever returns conflict

core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala
```diff
@@ -1,23 +1,56 @@
...
+
+
+  val petSAdomain = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
+
+  private def isPetSA(email: String) = {
+    petSAdomain.pattern.matcher(email).matches
+  }
+
+
+  def requireUserInfo(): Directive1[UserInfo] = detach(()) & {
```
dvoet: we should have integration tests that ensure that non-pet SAs still work.

dvoet (changes requested):

dvoet (approved):

PR #846, opened 2018-01-31
Rawls supports multiple method repos: DB migration [GAWB-3135]

aednichols:
~~Exists to run tests.~~

Things I've done:

- Migrate the `METHOD_CONFIG` table to represent a config's method as a `METHOD_URI` instead of a tuple of `METHOD_NAMESPACE`, `METHOD_NAME`, `METHOD_VERSION`
- Update the marshaling/unmarshaling code for `MethodRepoMethod` to understand the new data format
- My migrated Rawls DB clone is `anichols-rawls-clone`/`35.224.56.211` in case you want to inspect/run against it

Things I'm thinking about doing:

- Tweak the Liquibase changeset to include rollback code
- Enhanced handling of unexpected data (maybe there are null columns in the prod DB? Can we do better than throw an exception?)

----

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
@@ -9,7 +9,6 @@ import com.netaporter.uri.Uri.parse
 import spray.http.StatusCode
 import spray.json._
 import UserModelJsonSupport.ManagedGroupRefFormat
-import com.netaporter.uri.{PathPart, Uri}
 
 import scala.util.Try
 
@@ -201,9 +200,12 @@ case class MethodRepoMethod(
 
   // Next phase: this method goes away
   def asMethodUrlForRepo(repository: String): String = {
-    (repository      != null && repository.nonEmpty,
-     methodNamespace != null && methodNamespace.nonEmpty,
-     methodName      != null && methodName.nonEmpty,
+
+    def existsAndNonEmpty(subject: String) = subject != null && subject.nonEmpty
+
+    (existsAndNonEmpty(repository),
+     existsAndNonEmpty(methodNamespace),
+     existsAndNonEmpty(methodName),
      methodVersion > 0) match {
       case (true, true, true, true) => s"$repository://$methodNamespace/$methodName/$methodVersion"
       case _ => throw new RawlsException(
@@ -217,11 +219,11 @@ object MethodRepoMethod {
   def apply(uri: String): MethodRepoMethod = {
 
     (for {
-      parsedUri: Uri <- Try(parse(uri)).toOption
-      namespace: String <- parsedUri.host
-      parts: Seq[PathPart] <- Option(parsedUri.pathParts)
-      name: String <- Option(parts.head.part)
-      version: Int <- Try(parts.last.part.toInt).toOption
+      parsedUri <- Try(parse(uri)).toOption
+      namespace <- parsedUri.host
+      parts     <- Option(parsedUri.pathParts)
+      name      <- Option(parts.head.part)
+      version   <- Try(parts.last.part.toInt).toOption
     } yield {
       MethodRepoMethod(namespace, name, version)
     }).getOrElse(throw new RawlsException(s"Could not create a MethodRepoMethod from URI \'$uri\'"))
--- a/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
@@ -201,7 +201,7 @@ case class MethodRepoMethod(
   // Next phase: this method goes away
   def asMethodUrlForRepo(repository: String): String = {
 
-    if (existsAndNonEmpty(repository) && MethodRepos.withName(repository).isDefined && this.validate.isDefined)
+    if (repository.nonEmpty && MethodRepos.withName(repository).isDefined && this.validate.isDefined)
       s"$repository://$methodNamespace/$methodName/$methodVersion"
     else
       throw new RawlsException(
@@ -210,13 +210,11 @@ case class MethodRepoMethod(
   }
 
   def validate: Option[MethodRepoMethod] = {
-    if (existsAndNonEmpty(methodNamespace) && existsAndNonEmpty(methodName) && methodVersion > 0)
+    if (methodNamespace.nonEmpty && methodName.nonEmpty && methodVersion > 0)
       Some(this)
     else
       None
   }
-
-  private def existsAndNonEmpty(subject: String) = subject != null && subject.nonEmpty
 
 }
 
--- a/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
@@ -11,6 +11,8 @@ import spray.json._
 import UserModelJsonSupport.ManagedGroupRefFormat
 import com.netaporter.uri.{PathPart, Uri}
 
+import scala.util.Try
+
 object Attributable {
   // if updating these, also update their use in SlickExpressionParsing
   val entityIdAttributeSuffix = "_id"
@@ -189,6 +191,7 @@ case class EntityCopyDefinition(
                    entityNames: Seq[String]
                    )
 
+// In the next phase (GAWB-3100), this class's members will be swapped out for a singular methodUri
 case class MethodRepoMethod(
                    methodNamespace: String,
                    methodName: String,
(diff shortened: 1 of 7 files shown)


aednichols:
jenkins retest

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +190,45 @@ case class EntityCopyDefinition(
...
+
+object MethodRepoMethod {
+  def apply(uri: String): MethodRepoMethod = {
+
+    (for {
+      parsedUri <- Try(parse(uri)).toOption
+      namespace <- parsedUri.host
+      parts     <- Option(parsedUri.pathParts)
+      name      <- Option(parts.head.part)
+      version   <- Try(parts.last.part.toInt).toOption
```
aednichols: Won't catch `agora://namespace/1` (`parts.head == parts.last`)

aednichols:
Test failure due to fiab-start failing to connect to vault

https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-start/5034/console

jenkins retest

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -5,10 +5,13 @@ import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.SortDirections.SortDirection
 import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels.WorkspaceAccessLevel
 import org.joda.time.DateTime
+import com.netaporter.uri.Uri.parse
```
davidangb, line 11: is there a specific reason you chose this instead of `spray.http.Uri`? (even though we know Spray will be going away soonish)

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +190,46 @@ case class EntityCopyDefinition(
...
                    )
 
+// In the next phase (GAWB-3100), this class's members will be swapped out for a singular methodUri
 case class MethodRepoMethod(
                    methodNamespace: String,
                    methodName: String,
                    methodVersion: Int
-                   )
+                   ) {
+  def asAgoraMethodUrl: String = asMethodUrlForRepo("agora")
```
davidangb: not needed for this PR, but we probably want to make the strings "agora" and "dockstore" (etc) constants or an enum.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +190,46 @@ case class EntityCopyDefinition(
...
+      parsedUri <- Try(parse(uri)).toOption
+      namespace <- parsedUri.host
+      parts     <- Option(parsedUri.pathParts)
+      name      <- Option(parts.head.part)
+      version   <- Try(parts(1).part.toInt).toOption
+      result    <- if (parts.size == 2) Some(MethodRepoMethod(namespace, name, version)) else None
+    } yield {
+      result
+    }).getOrElse(throw new RawlsException(s"Could not create a MethodRepoMethod from URI \'$uri\'"))
+  }
```
davidangb: I think it would be better to throw an exception here in the `apply` instead of possibly creating a `MethodRepoMethod` with None/nulls in it - then, you can avoid the null check above.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+          assertResult("agora://test-namespace/test-name/555") {
+            method.asAgoraMethodUrl
+          }
+        }
+
+        "for some other method repo" in {
+          assertResult("marks-methods-mart://test-namespace/test-name/555") {
+            method.asMethodUrlForRepo("marks-methods-mart")
+          }
+        }
```
davidangb: I think we should have a whitelist (via constants/enums) of the repos we support, and not allow arbitrary repos - since we won't support those arbitrary repos for things like retrieving WDL.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      val badUri1 = null
+      val badUri2 = ""
+      val badUri3 = "agoraphobia"
+      val badUri4 = "agora://"
+      val badUri5 = "agora://test-namespace/test-name"
+      val badUri6 = "agora://test-namespace/1"
+      val badUri7 = "agora://test-namespace/test-name/bad/path"
```
davidangb, line 98: I'm interested what would happen if we include url-encoded slashes or other "weird" characters. Maybe this is actually just testing the Uri parser ... but I know that Dockstore encodes some of their identifiers.

davidangb (commented):
some initial comments, did not test!

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: Good call, made this tweak.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: The Spray API seemed a bit clunkier: for example, I do not see a way to index to an arbitrary position in the path, only head/tail. And to get host, it's `parsedUri.authority.host` instead of `parsedUri.host`.  But it's certainly functional if you'd rather not introduce a new dependency.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +190,70 @@ case class EntityCopyDefinition(
...
+  def validate: Option[MethodRepoMethod] = {
+    (existsAndNonEmpty(methodNamespace),
+     existsAndNonEmpty(methodName),
+     methodVersion > 0) match {
+      case (true, true, true) => Some(this)
+      case _ => None
+    }
+  }
+
+  private def existsAndNonEmpty(subject: String) = subject != null && subject.nonEmpty
```
davidangb: I still don't think you can get `null` here, so it's not really worth checking for. (I hate nulls in Scala, so try to avoid them wherever possible)

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +190,70 @@ case class EntityCopyDefinition(
...
+    }
+  }
+
+  def validate: Option[MethodRepoMethod] = {
+    (existsAndNonEmpty(methodNamespace),
+     existsAndNonEmpty(methodName),
+     methodVersion > 0) match {
+      case (true, true, true) => Some(this)
+      case _ => None
+    }
```
davidangb: just style and non blocking for this PR: I would have written this as
```
    if (existsAndNonEmpty(methodNamespace) && existsAndNonEmpty(methodName) && methodVersion > 0)
      Some(this)
    else
      None
```
while if/else is less Scala-y than a match/case, the match/case is overkill here ... and more importantly, the match/case won't short-circuit evaluation in the same way the if/else will. (micro-optimization, I know)

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.scalatest.{FreeSpec, Matchers}
+
+class WorkspaceModelSpec extends FreeSpec with Matchers {
+
+  "MethodRepoMethod" - {
+
+    val goodMethod = MethodRepoMethod("test-namespace", "test-name", 555)
+    val badMethod1 = MethodRepoMethod("a", "", 1)
+    val badMethod2 = MethodRepoMethod(null, "b", 1)
```
davidangb: I don't think you need the `null` test because it's not possible to achieve given the code.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
davidangb: still think we need some tests for url-encoded uris, but that can wait until we add dockstore support.

davidangb (commented):
a few more comments, very close

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: Agreed, this construct started life as something else iirc

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -201,25 +203,21 @@ case class MethodRepoMethod(
   // Next phase: this method goes away
   def asMethodUrlForRepo(repository: String): String = {
 
-    (existsAndNonEmpty(repository), MethodRepos.withName(repository), this.validate) match {
-      case (true, Some(_), Some(_)) => s"$repository://$methodNamespace/$methodName/$methodVersion"
-      case _ => throw new RawlsException(
+    if (repository.nonEmpty && MethodRepos.withName(repository).isDefined && this.validate.isDefined)
+      s"$repository://${URLEncoder.encode(methodNamespace, "UTF-8")}/${URLEncoder.encode(methodName, "UTF-8")}/$methodVersion"
```
davidangb: up to you/optional...

use a constant for "UTF-8":
```
import java.nio.charset.StandardCharsets.UTF_8
...
${URLEncoder.encode(methodNamespace, UTF_8.name)}
```

davidangb (approved):
:+1: lgtm, one minor optional comment inline.

Please make sure to add a note to whatever release ticket this ends up in that the release will include a (minor) rawls db migration!

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -187,11 +192,66 @@ case class EntityCopyDefinition(
...
+  }
+
+  def withName(name: String): Option[MethodRepository] = name match {
+    case "agora" => Some(Agora)
+    case _ => None
+  }
+
+  case object Agora extends MethodRepository
+
+  val all: Set[MethodRepository] = Set(Agora)
```
kyuksel, line 254: Consider removing unused `val`

kyuksel (approved):
👍 with a minor comment

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: Generally I'd drop it, but I use it immediately in the follow-on ticket

aednichols:
Failures are 8 "well-known" cases in `DataSpec`, Nicole OK'd merge OTS

PR #850, opened 2018-02-02
GAWB-3149 fix for all attributes string overflowing when user has tons of data

helgridly:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -687,7 +687,7 @@ trait EntityComponent {
     // Utility methods
 
     private def createAllAttributesString(entity: Entity): Option[String] = {
-      Option(s"${entity.name} ${entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]).map(AttributeStringifier(_)).mkString(" ")}".toLowerCase)
+      Option(s"${entity.name} ${entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]).map(AttributeStringifier(_)).mkString(" ")}".toLowerCase.take(65534))
     }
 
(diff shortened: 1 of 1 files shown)


breilly2:
If this is what I remember being talked about in the past, this string is used for searching/filtering on the data. If that's the case, then this goes from being a loud failure to silently not considering all fields when searching. Is this the desired behavior? I mean, obviously what's currently happening is wrong, but I worry that this will lead to quiet head-scratching and something that we may never get back to doing the right way. Is there a little more we can do, even just to inform the user that the size of the data might hinder searching/filtering?

helgridly:
It's currently manifesting as a silent failure that's preventing workflows from completing. I don't know which approach we want to take here in terms of communicating with users; I think sorting that bit is some tech debt + discussion with POs.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -687,7 +687,7 @@ trait EntityComponent {
     // Utility methods
 
     private def createAllAttributesString(entity: Entity): Option[String] = {
-      Option(s"${entity.name} ${entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]).map(AttributeStringifier(_)).mkString(" ")}".toLowerCase)
+      Option(s"${entity.name} ${entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]).map(AttributeStringifier(_)).mkString(" ")}".toLowerCase.take(65534))
```
rtitle: Why was this number chosen? Can we make it a constant and add an explanatory comment?

helgridly: it's the size of `TEXT` minus one because i'm scared. yes.

rtitle (approved):

PR #873, opened 2018-03-06
Retrieve and run WDLs from Dockstore [GAWB-3101]

aednichols:
Requires https://github.com/broadinstitute/firecloud-develop/pull/1000, https://github.com/broadinstitute/firecloud-develop/pull/987


aednichols:
Two existing unit tests broke, both with the pattern `Expected 201 Created, but got 409 Conflict`

I'm willing to bet this is because I messed with the test data, fixing...

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/SlickExpressionEvaluator.scala
```diff
@@ -35,7 +35,7 @@ private[expressions] object SlickExpressionEvaluator {
     //Sanity check that we've only got one, and then pass upwards
     dbRootEntityRec flatMap { rootEntityRec =>
       if(rootEntityRec.size != 1) {
-        DBIO.failed(new RawlsException(s"Found != 1 root entity when searching for ${rootType}/$rootName"))
+        DBIO.failed(new RawlsException(s"Expected 1 root entity type, found ${rootEntityRec.size} when searching for $rootType/$rootName"))
```
aednichols, line 38: I encountered this error IRL while testing this feature and thought it was confusing

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -161,8 +161,15 @@ object Boot extends App with LazyLogging {
...
-    val methodRepoDAO = new HttpMethodRepoDAO(agoraConfig.getString("server"), agoraConfig.getString("path"), metricsPrefix)
+    val agoraConfig = conf.getConfig("agora")
+    val dockstoreConfig = conf.getConfig("dockstore")
+    val methodRepoDAO = new HttpMethodRepoDAO(
+      agoraConfig.getString("server"),
+      agoraConfig.getString("path"),
+      dockstoreConfig.getString("server"),
+      dockstoreConfig.getString("path"),
+      metricsPrefix
+    )
```
davidangb: on the topic of strong typing ... I worry about five `String`s in a row in the constructor. Is it worthwhile creating strongly-typed `AgoraConfig` and `DockstoreConfig` classes and using those in the constructor?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
```diff
@@ -59,18 +60,25 @@ class HttpMethodRepoDAO(baseMethodRepoServiceURL: String, apiPath: String = "",
...
+        getDockstoreEntity(dockstoreMethod) flatMap { response: GA4GHToolDescriptor =>
+          // TODO: re-using AgoraEntity feels sketchy to me. It seems to work without any changes, but should we create a DockstoreEntity?
+          Future(Some(AgoraEntity(payload = Some(response.descriptor), entityType = Some(AgoraEntityType.Workflow))))
+        }
     }
   }
 
+  private def getDockstoreEntity(method: DockstoreMethod): Future[GA4GHToolDescriptor] = {
+    httpClientUtils.executeRequestUnmarshalResponse[GA4GHToolDescriptor](http, Get(method.ga4ghDescriptorUrl(dockstoreServiceURL)))
+  }
```
davidangb: I'm open to pushback on this: I find it weird that `getAgoraEntity` and `getDockstoreEntity` are so different.

`getAgoraEntity` accepts a url, and contains a retry + pipeline.
`getDockstoreEntity` accepts an object, and calls out to httpClientUtils.

I realize we might need the extra error-handling in `getAgoraEntity` ... but could these be made more parallel?

davidangb (commented):
first round review ...

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
aednichols: `getAgoraEntity` is called what it is because the URL may specify either `/methods/` or `/configurations/`.

In light of the above observation, I renamed `getDockstoreEntity` to `getDockstoreMethod` for clarity.

Because we are calling public Dockstore APIs, I created an unauthenticated version of `pipeline` so that we can have `getDockstoreMethod` and ` getAgoraEntity` make their HTTP calls at the same level of abstraction.

davidangb (approved):
:+1: lgtm, did not test.

ahaessly (approved):
👍

geoffjentry:
This is awesome!

core/src/main/scala/org/broadinstitute/dsde/rawls/config/MethodRepoConfig.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsde.rawls.config
+
+import com.typesafe.config.Config
+import org.broadinstitute.dsde.rawls.model.MethodRepository
+
+case class MethodRepoConfig[T <: MethodRepository](baseUrl: String, path: String) {
```
kyuksel, line 6: I couldn't see how the type information (`T`) is used. Would we lose anything if it were
```scala
case class MethodRepoConfig(baseUrl: String, path: String)
```
?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
```diff
@@ -46,28 +48,41 @@ class HttpMethodRepoDAO(baseMethodRepoServiceURL: String, apiPath: String = "",
...
       case agoraMethod: AgoraMethod =>
-        getAgoraEntity(s"${methodRepoServiceURL}/methods/${agoraMethod.methodNamespace}/${agoraMethod.methodName}/${agoraMethod.methodVersion}", userInfo)
-      case _ =>
-        throw new RawlsException(s"Method repo '${method.repo.scheme}' not yet supported")
+        getAgoraEntity(s"$agoraServiceURL/methods/${agoraMethod.methodNamespace}/${agoraMethod.methodName}/${agoraMethod.methodVersion}", userInfo)
+      case dockstoreMethod: DockstoreMethod =>
+        getDockstoreMethod(dockstoreMethod) flatMap { response: Option[GA4GHTool] =>
+          response match {
+            case Some(tool) =>
+              // TODO: re-using AgoraEntity feels sketchy to me. It seems to work without any changes, but should we create a DockstoreEntity?
```
kyuksel, line 67: Using `DockstoreEntity` would be clearer to me.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
```diff
@@ -46,28 +48,41 @@ class HttpMethodRepoDAO(baseMethodRepoServiceURL: String, apiPath: String = "",
...
+    retry(when500) { () =>
+      pipeline[Option[AgoraEntity]](userInfo) apply Get(url) recover {
+        case notOK: RawlsExceptionWithErrorReport if notOK.errorReport.statusCode.contains(StatusCodes.NotFound) => None
+      }
+    }
+  }
+
+  private def getDockstoreMethod(method: DockstoreMethod): Future[Option[GA4GHTool]] = {
+    pipeline[Option[GA4GHTool]] apply Get(method.ga4ghDescriptorUrl(dockstoreServiceURL)) recover {
+      case notOK: RawlsExceptionWithErrorReport if notOK.errorReport.statusCode.contains(StatusCodes.NotFound) => None
```
kyuksel, line 85: Would it be better to do something about the other `case`s here?

kyuksel (approved):
LGTM 👍 Minor comments/questions inline.

core/src/main/scala/org/broadinstitute/dsde/rawls/config/MethodRepoConfig.scala
aednichols:
```
val methodRepoDAO = new HttpMethodRepoDAO(
  MethodRepoConfig.apply[Agora.type](conf.getConfig("agora")),
  MethodRepoConfig.apply[Dockstore.type](conf.getConfig("dockstore")),
  metricsPrefix
)
```
in `Boot.scala` - basically, keeping the order of parameters straight without having a separate `DockstoreConfig` and `AgoraConfig` with identical fields

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
aednichols: Making a 404 return `None` is possibly attributable to stale/bad data, but I think anything else is programmer error or a system problem and should throw an exception.

I just copied this handling from `getAgoraEntity`, it makes sense to me though.

core/src/main/scala/org/broadinstitute/dsde/rawls/config/MethodRepoConfig.scala
kyuksel: ah I see

rushtong (approved):
👍

aednichols:
jenkins retest

aednichols:
jenkins retest

davidbernick:
jenkins retest

aednichols:
jenkins retest

PR #894, opened 2018-04-06
streamline SQL query for input resolution records

davidangb:
The SQL query in question uses `entityQuery`, which is a `select *` on the ENTITY table. By using `select *`, we are retrieving the `all_attribute_values` column from the database - which we immediately discard when we turn the EntityRecord into an AttributeEntityReference.

Against the reproduction data we identified from prod, retrieving the all_attribute_values column adds **_480MB_** to the size of the result set - none of which is used - and all of which needs to move over the wire. Changing the SQL query to ignore this column makes the result set much smaller.

In informal testing from my laptop, ignoring this column reduces the query time (including network transfer time) from 8-10 seconds to ~0.2 seconds.

Old SQL query:
```
select x2.`id`, x2.`name`, x2.`entity_type`, x2.`workspace_id`, x2.`record_version`, x2.`all_attribute_values`, x2.`deleted`, x2.`deleted_date`, x3.`id`, x3.`WORKFLOW_ID`, x3.`ERROR_TEXT`, x3.`INPUT_NAME`
from `WORKFLOW` x4, `ENTITY` x2, `SUBMISSION_VALIDATION` x3
where ((x4.`SUBMISSION_ID` = x'my-submission-id') and (x2.`id` = x4.`ENTITY_ID`)) and (x3.`WORKFLOW_ID` = x4.`ID`);
```

New SQL query:
```
select e.entity_type, e.name,
              sv.id, sv.WORKFLOW_ID, sv.ERROR_TEXT, sv.INPUT_NAME
        from ENTITY e, SUBMISSION_VALIDATION sv, WORKFLOW w
        where w.SUBMISSION_ID = x'my-submission-id'
          and e.id = w.ENTITY_ID
          and sv.WORKFLOW_ID = w.ID
```
where clauses and target tables haven't changed, so the explain plan hasn't either.

I don't know if this is **THE** cause of lock issues we've had in prod. However, this query is in the codepath identified to be troublesome, and it is a worthwhile fix even if this isn't the only cause.

We do need to audit everywhere else `entityQuery` is used. I didn't want to hold this PR waiting on that.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -139,10 +139,8 @@ trait WorkflowComponent {
           marshalInputResolution(inputResolution, workflowRecsByEntity(workflow.workflowEntity).id)
         }
 
-        val insertedRecQuery = submissionQuery.loadSubmissionValidationsAndReferences(submissionId)
-
         insertInBatches(submissionValidationQuery, inputResolutionRecs) andThen
-        insertedRecQuery.map(_.map { case (ref, insertedInputResolutionRec) =>
+        submissionQuery.loadSubmissionValidationsAndReferences(submissionId).map(_.map { case (ref, insertedInputResolutionRec) =>
           val name = insertedInputResolutionRec.inputName
           (ref, name) -> insertedInputResolutionRec
         }.toMap)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -139,15 +139,10 @@ trait WorkflowComponent {
           marshalInputResolution(inputResolution, workflowRecsByEntity(workflow.workflowEntity).id)
         }
 
-        val insertedRecQuery = for {
-          workflowRec <- findWorkflowsBySubmissionId(submissionId)
-          workflowEntityRec <- entityQuery if workflowEntityRec.id === workflowRec.workflowEntityId
-          insertedInputResolutionRec <- submissionValidationQuery if insertedInputResolutionRec.workflowId === workflowRec.id
-        } yield (workflowEntityRec, insertedInputResolutionRec)
+        val insertedRecQuery = submissionQuery.loadSubmissionValidationsAndReferences(submissionId)
 
         insertInBatches(submissionValidationQuery, inputResolutionRecs) andThen
-        insertedRecQuery.result.map(_.map { case (workflowEntityRec, insertedInputResolutionRec) =>
-          val ref = workflowEntityRec.toReference
+        insertedRecQuery.map(_.map { case (ref, insertedInputResolutionRec) =>
           val name = insertedInputResolutionRec.inputName
           (ref, name) -> insertedInputResolutionRec
         }.toMap)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
@@ -342,6 +342,10 @@ trait SubmissionComponent {
       idOp.map(id => id.getOrElse(throw new RawlsException(s"$entityRef not found")))
     }
 
+    def loadSubmissionValidationsAndReferences(submissionId: UUID): ReadAction[Seq[(AttributeEntityReference, SubmissionValidationRecord)]] = {
+      EntityRefAndInputResolutionRawSqlQuery.action(submissionId) map ( _ map { p => (p.entityReference, p.submissionValidationRec) } )
+    }
+
     def getMethodConfigOutputExpressions(submissionId: UUID): ReadAction[Map[String, String]] = {
       val query = for {
         submission <- submissionQuery if submission.id === submissionId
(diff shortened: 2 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -139,15 +139,10 @@ trait WorkflowComponent {
...
 
-        val insertedRecQuery = for {
-          workflowRec <- findWorkflowsBySubmissionId(submissionId)
-          workflowEntityRec <- entityQuery if workflowEntityRec.id === workflowRec.workflowEntityId
-          insertedInputResolutionRec <- submissionValidationQuery if insertedInputResolutionRec.workflowId === workflowRec.id
-        } yield (workflowEntityRec, insertedInputResolutionRec)
+        val insertedRecQuery = submissionQuery.loadSubmissionValidationsAndReferences(submissionId)
 
         insertInBatches(submissionValidationQuery, inputResolutionRecs) andThen
-        insertedRecQuery.result.map(_.map { case (workflowEntityRec, insertedInputResolutionRec) =>
```
jmthibault79: Since we're going from Query to Action here, it makes sense that Slick would return a giant pile of data before we trimmed it on the app side.  I wonder if Slick would be smart enough to do the data reduction for us if we moved this logic inside the Query.

Great find!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -139,15 +139,10 @@ trait WorkflowComponent {
           marshalInputResolution(inputResolution, workflowRecsByEntity(workflow.workflowEntity).id)
         }
 
-        val insertedRecQuery = for {
-          workflowRec <- findWorkflowsBySubmissionId(submissionId)
-          workflowEntityRec <- entityQuery if workflowEntityRec.id === workflowRec.workflowEntityId
-          insertedInputResolutionRec <- submissionValidationQuery if insertedInputResolutionRec.workflowId === workflowRec.id
-        } yield (workflowEntityRec, insertedInputResolutionRec)
+        val insertedRecQuery = submissionQuery.loadSubmissionValidationsAndReferences(submissionId)
```
helgridly: nit: this isn't a query any more, it's an action, and the variable name should say so

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -139,15 +139,10 @@ trait WorkflowComponent {
...
 
         insertInBatches(submissionValidationQuery, inputResolutionRecs) andThen
-        insertedRecQuery.result.map(_.map { case (workflowEntityRec, insertedInputResolutionRec) =>
-          val ref = workflowEntityRec.toReference
+        insertedRecQuery.map(_.map { case (ref, insertedInputResolutionRec) =>
```
helgridly: if DBIOs behave like Futures, the `insertedRecQuery` will have started running already, so you won't have the ordering you want

helgridly (changes requested):
:+1: but I am a little suspicious that `andThen` doesn't do what you want

helgridly (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -139,15 +139,8 @@ trait WorkflowComponent {
           marshalInputResolution(inputResolution, workflowRecsByEntity(workflow.workflowEntity).id)
         }
 
-        val insertedRecQuery = for {
-          workflowRec <- findWorkflowsBySubmissionId(submissionId)
-          workflowEntityRec <- entityQuery if workflowEntityRec.id === workflowRec.workflowEntityId
-          insertedInputResolutionRec <- submissionValidationQuery if insertedInputResolutionRec.workflowId === workflowRec.id
-        } yield (workflowEntityRec, insertedInputResolutionRec)
```
dvoet, line 146: I don't really care that much, but why not just change what is yielded here? Did you need to go down the route of raw sql?

davidangb: I believe that doing it in the yield is too late - it still selects all columns in the underlying SQL. I did think of making a lighter drop-in replacement just for `entityQuery`, but then I saw that there were a handful of other columns we were selecting that weren't necessary, and decided to drop to raw SQL.

jmthibault79: Because at yield time it is still a Query and not yet an Action, Slick should reduce the data transfer to only what we need - as long as we do this before calling `result()` instead of after.

davidangb: alright, so @jmthibault79 @dvoet give me some guidance here ... would you prefer to see the PR refactored to try out the yield? Or leave as-is?

jmthibault79: My general preference is to keep it Slick-style, switching to Raw only if that's necessary.  So something like `yield (workflowEntityRec.entityType, workflowEntityRec.entityName, insertedInputResolutionRec)`

davidangb:
closing in favor of #895.

PR #913, opened 2018-05-07
Enable access to subworkflow metadata

### LINKED ISSUE #9: GAWB-3378: Always requesting subworkflow metadata from Cromwell produces too much load
Description: 
After  GAWB-3293 QA APPROVED  we now set expandSubWorkflows to true in Rawls -> Cromwell metadata GET calls. Because Rawls makes many of these calls, and each requires a meaningful amount of CPU/memory/DB work and data transfer on the Cromwell side, this has been causing problems.

We still want subworkflow metadata (and sub-sub-sub-workflow metadata, etc.), but we should only retrieve what we need, rather than always requesting the entire tree. This can span multiple API calls, for simplicity.

Create a design doc with a plan and coordinate with Cromwell and Field Ops.

Comment: 
Implementation plan, based on the Design Doc discussion:

Set up or access an environment with a past submission using a substantial tree of subworkflows, or run a new submission with this.

Use curl to call Cromwell GET Metadata for the top level workflow with expandSubWorkflows=true and record the payload size. Run it multiple times, attempting to overload Cromwell's memory (as observed by 503 responses and memory graphs in Google Console).

After re-deploying Cromwell, make curl calls with expandSubWorkflows=false and record the payload size. Try to overload again (should not be able to do this).

Use curl again to call with expandSubWorkflows=true and explicit selection of desired keys. Record payload size and try to overload. If both of these results are good, update Rawls to set expandSubWorkflows=true and take key parameters. Update FC-UI to pass the set of desired keys as parameters.

Link to GAWB Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3378

Unblocks UI support for nested workflows: https://broadinstitute.atlassian.net/browse/GAWB-3199

jmthibault79:
Requires https://github.com/broadinstitute/workbench-libs/pull/117
Resolves https://github.com/DataBiosphere/firecloud-app/issues/9

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


jmthibault79:
Requires https://github.com/broadinstitute/workbench-libs/pull/117
Resolves https://github.com/DataBiosphere/firecloud-app/issues/9

automation/src/test/scala/org/broadinstitute/dsde/test/api/TestMethods.scala
```diff
@@ -1,213 +0,0 @@
-package org.broadinstitute.dsde.test.api
```
jmthibault79, line 1: Workbench Libs `SubWorkflowFixtures` is an updated version of this

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
```diff
@@ -54,22 +53,108 @@ class ShardedHttpExecutionServiceCluster (readMembers: Map[ExecutionServiceId,Ex
...
+      id <- shard.asJsObject().getFields("subWorkflowId")
+    } yield id.convertTo[String]
+  }
+
+  // parse subworkflow IDs from the parent workflow's metadata and label each of them with the Submission ID
+  // so they can be found by future workflow metadata queries
+
+  private def labelSubWorkflowsWithSubmissionId(submissionId: String, executionServiceId: ExecutionServiceId, parentWorkflowMetadata: JsObject, userInfo: UserInfo): Future[Seq[ExecutionServiceLabelResponse]] = {
+    // traverse = stop on "first" failure and propagate
+    Future.traverse(parseSubWorkflowIdsFromMetadata(parentWorkflowMetadata.compactPrint)) { subWorkflowId =>
```
davidangb: I find it icky that we are printing a `JsObject` to string here, then in `parseSubWorkflowIdsFromMetadata` we immediately parse it back to a `JsObject`. It looks like `parseSubWorkflowIdsFromMetadata` is not used anywhere else? Why not have it accept a `JsObject`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
```diff
@@ -54,22 +53,108 @@ class ShardedHttpExecutionServiceCluster (readMembers: Map[ExecutionServiceId,Ex
...
+
+      case _ =>
+        // we don't have the execution service key because the workflow is not in the DB.  It might be a subworkflow.
+        // query all execution services for Workflow labels and search for a Submission match
+
+        val executionServiceLabelMap = readMembers.map { case (executionServiceId, executionServiceDao) =>
+          executionServiceDao.getLabels(workflowId, userInfo) map { labelResponse =>
+            (executionServiceId, labelResponse.labels)
+          }
+        }
```
davidangb: given what we know about cromwell sharding's legacy and especially the timing of when subworkflows were well-supported, we *expect* that all subworkflows will be on the cromwell submitServer, and none will be on the read-only extra shard. Therefore, this is making a probably-extraneous API call to the read-only shard for every subworkflow.

What if we changed this to query the expected (submit) shard first, and only if nothing is returned from that shard do we query the extra (read) shard? In the expected case this will cut down on API calls; in the corner case it will cause a delay due to the serial order of the queries.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2105,6 +2120,23 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  // used as part of the workflow metadata permission check - more detail at workflowMetadata()
+
+  // require submission to be present, but don't require the workflow to reference it
+  // if the workflow does reference the submission, return its executionServiceKey
+
+  private def withSubmissionAndWorkflowExecutionServiceKey[T](workspaceContext: SlickWorkspaceContext, submissionId: String, workflowId: String, dataAccess: DataAccess)(op: (Option[String]) => ReadWriteAction[T]): ReadWriteAction[T] = {
+    withSubmission(workspaceContext, submissionId, dataAccess) { _ =>
```
davidangb: `withSubmission` is really heavy to use in this situation. I think all you're doing here is validating that a submission with that id exists, correct? If so, is there something more optimized you can use? IIRC `withSubmission` loads all the workflows and entities and such, which you don't need.

davidangb (commented):
a few comments inline, lmk if you want to discuss.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
jmthibault79: ew - must have missed that on some refactoring round.  Will fix.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
jmthibault79: Good point.  I'll optimize.

I bet if we decided to hunt for them we could find many other examples like this in Rawls.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
jmthibault79: OK, makes sense

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: yeah, I'm sure there are a lot of existing cases of this - just don't want to add more!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -110,7 +110,10 @@ trait WorkflowComponent {
     }
 
     def getExecutionServiceIdByExternalId(externalId: String, submissionId: String): ReadAction[Option[String]] = {
-      uniqueResult[String](findWorkflowByExternalIdAndSubmissionId(externalId, UUID.fromString(submissionId)) map (_.executionServiceKey))
+      uniqueResult[WorkflowRecord](findWorkflowByExternalIdAndSubmissionId(externalId, UUID.fromString(submissionId))) map {
+        case Some(wr) => wr.executionServiceKey
+        case _ => None
+      }
```
davidangb: From a quick perusal ... is it possible the `Option[Option[String]]` issue comes from `uniqueResult` and not from a bug in Slick itself? If the column type is `Option[String]`, then `uniqueResult` looks to return `Option[Option[String]]`.

davidangb (approved):
:+1: code looks good. I haven't done functional testing.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
jmthibault79: Right - that's what I overlooked, causing the problem.

My concern is that I relied on the type system to catch this.  This function returned a Slick `Action` encapsulating an `Option[Option[String]]` but this compiled claiming to return a `ReadAction[Option[String]`.  Only stepping thru the debugger revealed that this actually returned a `ReadAction[Option[Option[String]]` at runtime.

I had a previous version which attempted to do the right thing collapsing the options via `flatMap (_.executionServiceKey)` but I removed it because that would NOT compile :(

jmthibault79: Could be the problem is in `ReadAction` instead: that one's mine.

gdlex4015 (approved):
Might make sense to test through Orchestration as well?

PR #916, opened 2018-05-09
Get workflow costs from BigQuery

ghost:
This is the main PR for https://github.com/DataBiosphere/firecloud-app/issues/12

- [x] Merge after https://github.com/broadinstitute/firecloud-develop/pull/1161

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -119,10 +119,23 @@ case class SubmissionStatusResponse(
...
+      submissionDate = submission.submissionDate,
+      submitter = rawlsUser.userEmail.value,
+      methodConfigurationNamespace = submission.methodConfigurationNamespace,
+      methodConfigurationName = submission.methodConfigurationName,
+      submissionEntity = submission.submissionEntity,
+      workflows = submission.workflows,
+      workflowCostMap = workflowCostMap,
+      status = submission.status,
+      workflowFailureMode = submission.workflowFailureMode
+    )
```
davidangb: I'm pondering this. Instead of having `workflowCostMap` be an additional field on `SubmissionStatusResponse`, what if we added a `cost` field to `Workflow` as an `Option[Float] = None`? `Workflow` is used in a lot of places and so this change could have wide impact, but it feels like a better encapsulation.

Either way, this will need an update to swagger to denote the new fields.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleBigQueryDAO.scala
```diff
@@ -0,0 +1,13 @@
...
+import org.broadinstitute.dsde.workbench.google.HttpGoogleBigQueryDAO
+import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.GoogleCredentialMode
+
+import scala.concurrent.ExecutionContext
+
+class MockHttpGoogleBigQueryDAO(appName: String,
+                                googleCredentialMode: GoogleCredentialMode,
+                                workbenchMetricBaseName: String)
+                               (implicit system: ActorSystem, executionContext: ExecutionContext)
+  extends HttpGoogleBigQueryDAO(appName, googleCredentialMode, workbenchMetricBaseName)
```
davidangb: this should override the relevant methods so it doesn't try to call Google (unless you really want this to call Google)

davidangb (commented):
first pass review comments ... I am continuing to review!

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
ghost: I think that I agree with you on this. I guess the big question would be when to hit BigQuery.

davidangb: oh, I don't think you have to change too much about that. the flow would be the same, and you'd still pass `workflowCostMap` as an argument when creating `SubmissionStatusResponse` ... but instead of `SubmissionStatusResponse` having a dedicated field, the costs in `workflowCostMap` would be annotated onto the workflows. That make sense?

davidangb: I think you'd end up with something like this (unproven, uncompiled):
```
  def getSubmissionStatus(workspaceName: WorkspaceName, submissionId: String): Future[PerRequestMessage] = {
    val submissionWithoutCosts = dataSource.inTransaction { dataAccess =>
        withSubmission(workspaceContext, submissionId, dataAccess) { submission =>
          withUser(submission.submitter, dataAccess) { user =>
            DBIO.successful(submission, user)
          }
        }
      }
    }

    submissionWithoutCosts flatMap {
      case (submission, user) => {
        val allWorkflowIds: Seq[String] = submission.workflows.flatMap(_.workflowId)
        submissionCostService.getWorkflowCosts(workspaceName.namespace, allWorkflowIds, GoogleProject(workspaceName.namespace)) map { costMap =>
          val costedWorkflows = submission.workflows.map { wf =>
            wf.workflowId match {
              case Some(wfid) => wf.copy(cost = costMap.get(wfid)) // TODO: this will only compile once cost is added as a field on Workflow
              case None => wf
            }
          }
          val costedSubmission = submission.copy(workflows = costedWorkflows)
          RequestComplete((StatusCodes.OK, new SubmissionStatusResponse(costedSubmission, costMap, user))) // TODO: remove costMap from here
        }
      }
    }
  }
```

davidangb: and after sleeping on this, I suggest we add `cost` to each workflow, *and* add a `cost` to the submission itself that has `costMap.values.sum`. It's technically duplicate data (callers of the API could do the sum themselves) but it feels like a nice convenience.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -128,6 +132,12 @@ object Boot extends App with LazyLogging {
       workbenchMetricBaseName = metricsPrefix
     )
 
+    val bigQueryDAO = new HttpGoogleBigQueryDAO(
+      gcsConfig.getString("appName"),
+      Json(FileUtils.readFileToString(new java.io.File(gcsConfig.getString("pathToBigQueryJson")), Charset.defaultCharset())),
+      metricsPrefix
+    )
```
davidangb: elsewhere in rawls.conf, we store the *contents* of a service account's json as a value - see `gcs.secrets` or `gcs.tokenSecretsJson`. If you followed that model - which would drive a change in firecloud-develop - you wouldn't have to read the file from the filesystem.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
ghost: Great point!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1456,16 +1461,25 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
         }
       }
     }
+
+    submissionWithoutCosts flatMap {
+      case (submission, user) => {
+        val allWorkflowIds: Seq[String] = submission.workflows.flatMap(_.workflowId)
+        submissionCostService.getWorkflowCosts(workspaceName.namespace, allWorkflowIds, GoogleProject(workspaceName.namespace)) map { costMap =>
+          RequestComplete((StatusCodes.OK, new SubmissionStatusResponse(submission, costMap, user)))
+        }
```
davidangb: I think we should add some error-trapping here - at least we should be intentional about what happens in the case of error.

My preference is that if we can successfully retrieve the submission (from mysql) but encounter an error getting costs (from BQ), then we should return the submission without cost info - which needs error-trapping.

If you argue that a BQ error should totally prevent the submission response, then we don't need error trapping :)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -25,6 +25,7 @@ import org.scalatest.{FlatSpec, Matchers, Suite}
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 
 import scala.concurrent.duration._
+import scala.collection.Seq
```
davidangb: is this necessary?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /*
+   * Queries BigQuery for compute costs associated with the workflowIds.
+   */
+  private def executeWorkflowCostsQuery(namespace: String,
+                       workflowIds: Seq[String],
+                       googleProject: GoogleProject): Future[util.List[TableRow]] = {
+
+    // TODO: need to protect against SQL injection. Can we use Slick? Hate to bring in Slick just for that.
```
davidangb: we do need to resolve the injection issue - let's talk in person.

ghost: I have a way! But need some assistance with sbt dependencies.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ghost: Done! Any point in writing to logs in the error case?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -122,7 +124,18 @@ case class SubmissionStatusResponse(
...
+      submissionId = submission.submissionId,
+      submissionDate = submission.submissionDate,
+      submitter = rawlsUser.userEmail.value,
+      methodConfigurationNamespace = submission.methodConfigurationNamespace,
+      methodConfigurationName = submission.methodConfigurationName,
+      submissionEntity = submission.submissionEntity,
+      workflows = submission.workflows,
+      status = submission.status,
+      workflowFailureMode = submission.workflowFailureMode
+    )
```
ghost, line 144: @davidangb will the orchestration swagger need to be updated as well?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1456,16 +1461,35 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
         }
       }
     }
+
+    submissionWithoutCosts flatMap {
+      case (submission, user) => {
+        val allWorkflowIds: Seq[String] = submission.workflows.flatMap(_.workflowId)
+        toFutureTry(submissionCostService.getWorkflowCosts(workspaceName.namespace, allWorkflowIds, GoogleProject(workspaceName.namespace))) map {
+          case Failure(_) =>
+            RequestComplete((StatusCodes.OK, new SubmissionStatusResponse(submission, user)))
```
ghost: Should this write to logs in case of failure?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
davidangb: yup, orch swagger will also need updates!

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4116,6 +4116,9 @@ definitions:
       status:
         $ref: '#/definitions/SubmissionStatus'
         description: Status
+      cost:
+        type: number
+        format: float
```
davidangb, line 4121: nitpick: add a description for this one too

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: yes, please log it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1475,8 +1474,18 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+          case Success(costMap) =>
+            val costedWorkflows = submission.workflows.map { workflow =>
+              workflow.workflowId match {
+                case Some(wfId) => workflow.copy(cost = costMap.get(wfId))
+                case None => workflow
+              }
+            }
+            val costedSubmission = submission.copy(cost = Option(costMap.values.sum), workflows = costedWorkflows)
```
davidangb: nitpick: I think `costMap.values.sum` will always have a value, even if `costMap` has no entries. So you could/should use `Some(costMap.values.sum)` instead of `Option`. This isn't a blocker.

davidangb (commented):
Very close ... a few minor comments inline. Looks like this needs a rebase, too. Once rebased, I'd like to see this passing tests!

core/src/main/resources/swagger/api-docs.yaml
ghost: Added on the line below 😄

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
ghost: Here's the issue for tracking: 
https://app.zenhub.com/workspace/o/databiosphere/firecloud-app/issues/63

ghost:
My fiab build with this branch was successful. jenkins retest

davidangb (approved):
:+1:

PR #919, opened 2018-05-15
Clean up some rawls compile warnings

davidangb:
no ticket.

Code cleanup - primarily migration from `JdbcDriver` to `JdbcProfile` - to get rid of *some* of the compile warnings. This doesn't come close to getting rid of all of them nor does this PR intend to do so ... incremental progress!

I thought about tackling the remainder of the `JavaConverters` deprecation warnings, but that would have made this PR much larger; I'd rather make a separate PR for that.

Read individual commits if desired to review smaller chunks.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -368,11 +368,11 @@ trait WorkspaceComponent {
     }
 
     def getUserSharePermissions(subjectId: RawlsUserSubjectId, workspaceContext: SlickWorkspaceContext): ReadWriteAction[Boolean] = {
-      workspaceUserShareQuery.filter(rec => rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId).countDistinct.result.map(rows => rows > 0) flatMap { hasSharePermission =>
+      workspaceUserShareQuery.filter(rec => rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId).distinct.length.result.map(rows => rows > 0) flatMap { hasSharePermission =>
         if(hasSharePermission) DBIO.successful(hasSharePermission)
         else rawlsGroupQuery.listGroupsForUser(RawlsUserRef(subjectId)).flatMap { userGroups =>
           val groupNames = userGroups.map(_.groupName.value)
-          workspaceGroupShareQuery.filter(rec => (rec.workspaceId === workspaceContext.workspaceId) && rec.groupName.inSetBind(groupNames)).countDistinct.result.map(rows => rows > 0)
+          workspaceGroupShareQuery.filter(rec => (rec.workspaceId === workspaceContext.workspaceId) && rec.groupName.inSetBind(groupNames)).distinct.length.result.map(rows => rows > 0)
         }
       }
     }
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/util/HttpClientUtils.scala
```diff
@@ -67,7 +67,7 @@ case class HttpClientUtilsInstrumented(implicit val materializer: Materializer,
   }
 }
 
-case class HttpClientUtilsGzipInstrumented(implicit val materializer: Materializer, requestCounter: (HttpRequest, HttpResponse) => Counter, requestTimer: (HttpRequest, HttpResponse) => Timer, actorRefFactory: ActorRefFactory, val executionContext: ExecutionContext) extends HttpClientUtils {
+case class HttpClientUtilsGzipInstrumented()(implicit val materializer: Materializer, requestCounter: (HttpRequest, HttpResponse) => Counter, requestTimer: (HttpRequest, HttpResponse) => Timer, actorRefFactory: ActorRefFactory, val executionContext: ExecutionContext) extends HttpClientUtils {
```
ghost, line 70: Minor nitpick: can remove these empty parens (I think this was IntelliJ)

ghost (approved):
Compiled with less warnings!

core/src/main/scala/org/broadinstitute/dsde/rawls/util/HttpClientUtils.scala
davidangb: I added these explicitly to remove a compile warning! Something along the lines of "case classes should have an explicit argument list"

ghost: Oh good to know!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -368,11 +368,11 @@ trait WorkspaceComponent {
     }
 
     def getUserSharePermissions(subjectId: RawlsUserSubjectId, workspaceContext: SlickWorkspaceContext): ReadWriteAction[Boolean] = {
-      workspaceUserShareQuery.filter(rec => rec.userSubjectId === subjectId.value && rec.workspaceId === workspaceContext.workspaceId).countDistinct.result.map(rows => rows > 0) flatMap { hasSharePermission =>
```
jmthibault79: Here we need to be careful we're not relying on the older behavior: https://github.com/slick/slick/issues/1651

core/src/main/scala/org/broadinstitute/dsde/rawls/util/HttpClientUtils.scala
jmthibault79: Can they be `case object`s then?  Not sure if that's compatible with the implicits.

davidangb: I'm pretty certain that would not work with the implicits.

davidangb:
@jmthibault79 I've reverted the `countDistinct` change and will refactor that separately, that work for you?

jmthibault79 (approved):

davidangb:
`rawls-build` job was mysteriously hosed and I aborted it after 3hr 40min. Let's try again: jenkins retest.

PR #15, opened 2015-05-15
Graph DAO with tests

mattsooknah:
First stab at a graph DAO using Tinkerpop and OrientDB. Also cleaned up the dataaccess package. 

Does not include: using remote DB server, smart transactions.

PR #929, opened 2018-05-21
Submission cost in list submissions API

ghost:
This implements [#15](https://app.zenhub.com/workspace/o/databiosphere/firecloud-app/issues/15)
Also added a test to confirm that we bypass BigQuery when an empty list of workflows is given. 

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
@@ -120,8 +120,8 @@ trait SubmissionComponent {
       } yield (submissionRec, methodConfigRec, entityRec)
 
       for {
-        workflowStates <- GatherStatusesForWorkspaceSubmissionsQuery.gatherWorkflowStatuses(workspaceContext.workspaceId)
-        states = workflowStates.groupBy(_.submissionId)
+        workflowStatusResponses <- GatherStatusesForWorkspaceSubmissionsQuery.gatherWorkflowIdsAndStatuses(workspaceContext.workspaceId)
+        states = workflowStatusResponses.groupBy(_.submissionId)
         recs <- query.result
         users <- rawlsUserQuery.load(recs.map { case (submissionRec, _, _) => RawlsUserRef(RawlsUserSubjectId(submissionRec.submitterId)) })
       } yield {
@@ -130,9 +130,14 @@ trait SubmissionComponent {
         recs.map { case (submissionRec, methodConfigRec, entityRec) =>
           val user = usersById(RawlsUserSubjectId(submissionRec.submitterId))
           val config = methodConfigurationQuery.unmarshalMethodConfig(methodConfigRec, Map.empty, Map.empty, Map.empty)
-          val subStatuses = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count).toMap
-
-          new SubmissionListResponse(unmarshalSubmission(submissionRec, config, entityRec.map(_.toReference), Seq.empty), user, subStatuses)
+          val wfStates = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count).groupBy(_._1) map {
+            case (status, statusToCount) => status -> statusToCount.map(_._2)
+          }
+          val statusCounts = wfStates map {
+            case (status, count) => status -> count.sum
+          }
+          val workflowIds = states.getOrElse(submissionRec.id, Seq.empty).flatMap(_.workflowId).sorted
+          new SubmissionListResponse(unmarshalSubmission(submissionRec, config, entityRec.map(_.toReference), Seq.empty), user, workflowIds, statusCounts)
         }
       }
     }
@@ -537,15 +542,15 @@ trait SubmissionComponent {
       val driver: JdbcProfile = SubmissionComponent.this.driver
 
       implicit val getSubmissionWorkflowStatusResponse = GetResult { r =>
-        SubmissionWorkflowStatusResponse(r.<<, r.<<, r.<<)
+        SubmissionWorkflowStatusResponse(r.<<, r.<< ,r.<<, r.<<)
       }
 
-      def gatherWorkflowStatuses(workspaceId: UUID) = {
-        sql"""select s.id, w.status, count(*) from SUBMISSION s
+      def gatherWorkflowIdsAndStatuses(workspaceId: UUID) = {
+        sql"""select s.id, w.external_id, w.status, count(*) from SUBMISSION s
                 join WORKFLOW w
                 on s.id = w.submission_id
                 where s.workspace_id = $workspaceId
-                group by s.id, w.status""".as[SubmissionWorkflowStatusResponse]
+                group by s.id, w.external_id, w.status""".as[SubmissionWorkflowStatusResponse]
       }
     }
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import java.sql.Timestamp
 import java.util.UUID
 
+import cats.implicits._
 import nl.grons.metrics.scala.Counter
 import org.broadinstitute.dsde.rawls.RawlsException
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -130,9 +130,14 @@ trait SubmissionComponent {
...
           val config = methodConfigurationQuery.unmarshalMethodConfig(methodConfigRec, Map.empty, Map.empty, Map.empty)
-          val subStatuses = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count).toMap
-
-          new SubmissionListResponse(unmarshalSubmission(submissionRec, config, entityRec.map(_.toReference), Seq.empty), user, subStatuses)
+          val wfStates = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count).groupBy(_._1) map {
+            case (status, statusToCount) => status -> statusToCount.map(_._2)
+          }
+          val statusCounts = wfStates map {
+            case (status, count) => status -> count.sum
+          }
```
davidangb: I'd reduce these two `val`s into a single call, by adding the `.sum` right into the first:
```
val statusCounts:Map[String, Int] = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count).groupBy(_._1) map {
  case (status, statusToCount) => status -> statusToCount.map(_._2).sum
}
```
and, if you feel like diving into a fairly minimal use of cats, you can replace the groupBy/map/case/sum:
```
import cats.implicits._

val statusCounts:Map[String, Int] = states.getOrElse(submissionRec.id, Seq.empty).map(x => x.workflowStatus -> x.count)
    .foldLeft(Map.empty[String,Int])(_|+|_)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1393,14 +1393,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
   }
 
-  def listSubmissions(workspaceName: WorkspaceName): Future[PerRequestMessage] =
-    dataSource.inTransaction { dataAccess =>
+  def listSubmissions(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
+    val costlessSubmissionListFuture = dataSource.inTransaction { dataAccess =>
       withWorkspaceContextAndPermissions(workspaceName, WorkspaceAccessLevels.Read, dataAccess) { workspaceContext =>
-        dataAccess.submissionQuery.listWithSubmitter(workspaceContext)
-          .map(RequestComplete(StatusCodes.OK, _))
+        dataAccess.submissionQuery.listWithSubmitter(workspaceContext).flatMap(DBIO.successful)
```
davidangb: I think the `.flatMap(DBIO.successful)` is extraneous here - just remove it

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1393,14 +1393,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+        toFutureTry(submissionCostService.getWorkflowCosts(costlessSubmission.workflowIds, workspaceName.namespace)) map {
+          case Failure(ex) =>
+            logger.error(s"Unable to get cost data for submission ${costlessSubmission.submissionId}", ex)
+            costlessSubmission
+          case Success(costs) =>
+            costlessSubmission.copy(cost = Some(costs.values.sum))
+        }
+      }) map (RequestComplete(StatusCodes.OK, _))
+    }
+  }
```
davidangb: I don't like this approach. This is making N request to BQ, where N is the number of submissions in the workspace. Could we instead make a single BQ request for all workflowIds, then parse/extract/annotate the costs in Scala? I worry about performance for N queries.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostServiceSpec.scala
```diff
@@ -23,9 +23,15 @@ class SubmissionCostServiceSpec extends FlatSpec with RawlsTestUtils {
...
       Await.result(submissionCostService.extractWorkflowCostResults(rows), 1 minute)
     }
   }
 
+  it should "bypass BigQuery with no workflow IDs" in {
+    val expected = Map.empty
+    assertResult(expected) {
+      Await.result(submissionCostService.getWorkflowCosts(Seq.empty, "test"), 1 minute)
+    }
+  }
```
davidangb: this doesn't actually test that we're bypassing BQ - it just tests that we get an empty map in the response. We coul easily have made a request to BQ but returned nothing.

davidangb (commented):
comments inline!

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostServiceSpec.scala
ghost: I'd actually like to argue that it does test bypassing BQ. My reason being that `MockGoogleBigQueryDAO`, which is what this `submissionCostService` will hit, throws an exception if the parameters passed to `startParameterizedQuery` are not equal to the fields `testProject`, `testParamQuery`, `testParameters` and `testParameterMode` in `MockGoogleBigQueryDAO`. Since this is version of `SubmissionCostService` isn't passing those exact parameters, we'll always see an exception if we try to hit BigQuery in this spec, unless we override `SubmissionCostService#executeWorkflowCostQuery`. I'm interested to hear what you think about this.

jbuabbud: I rescind my previous comment.

davidangb: you're absolutely right! I didn't follow this all the way down. Could you please add comments explaining this, since I expect it wouldn't be obvious to other readers too?

ghost: Done!

jmthibault79 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1393,14 +1393,31 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+    }
+
+    toFutureTry(costMapFuture) flatMap {
+      case Failure(ex) =>
+        logger.error("Unable to get cost data from BigQuery", ex)
+        Future.successful(RequestComplete(StatusCodes.OK, costlessSubmissionsFuture))
+      case Success(costMap) =>
+        costlessSubmissionsFuture map { costlessSubmissions =>
+          val costedSubmissions = costlessSubmissions map { costlessSubmission =>
+            costlessSubmission.copy(cost = Some(costlessSubmission.workflowIds.flatMap(id => costMap.get(id).value).sum))
```
davidangb: what does `value` do here?  You should be able to do just `costlessSubmission.workflowIds.flatMap(costMap.get).sum` (note I've also removed the `id =>` and the shorthand `_` since `get` is a single-argument function.

If you wanted to keep using `map`, you could do `costlessSubmission.workflowIds.map(costMap.getOrElse(_,0).sum`

davidangb (commented):
one fairly minor comment inline. Have you run this and manually looked at the results?

ghost:
@davidangb screen shot 2018-05-24 at 2 11 26 pm

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ghost: 🤦‍♂️ This was my hacky way of fixing a bug that was really caused by having `Some().sum`. Going with the first suggestion you made. Thanks for the comment 😄

davidangb (approved):
:+1: thanks!

PR #935, opened 2018-05-30
SubmissionListResponse.workflowIds is now an Option

davidangb:
In the `SubmissionListResponse`, make `workflowIds` an `Option`. This means that when serializing to json, we can omit this property entirely from the response.

As a side tweak, this PR also sets a `SubmissionListResponse`'s `cost` to `None` when no workflows exist; previously this was set to `Some(0)`.

It's a stretch, but if we can get this merged prior to release, we will have never rendered `workflowIds` in a response, making for less churn in our APIs.


ghost (approved):
Looks good!

PR #936, opened 2018-05-30
disable costs for submission list

davidangb:
we are seeing perf problem; let's disable this until we can quantify.

This was committed over a few PRs so it's not a simple revert; this PR disables the relevant BigQuery call but makes it easy to re-enable.

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
@@ -418,7 +418,9 @@ class SubmissionApiServiceSpec extends ApiServiceSpec {
     def expectedResponse(sub: Submission): SubmissionListResponse = {
       val wfCount = sub.workflows.length
       val statuses: Map[String, Int] = if (wfCount > 0) Map("Submitted" -> wfCount) else Map.empty
-      val runCost = if (wfCount == 0) None else Some(wfCount * 1.23f)  // mockSubmissionCostService.fixedCost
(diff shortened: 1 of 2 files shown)


ghost (approved):
👍

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
```diff
@@ -418,7 +418,9 @@ class SubmissionApiServiceSpec extends ApiServiceSpec {
     def expectedResponse(sub: Submission): SubmissionListResponse = {
       val wfCount = sub.workflows.length
       val statuses: Map[String, Int] = if (wfCount > 0) Map("Submitted" -> wfCount) else Map.empty
-      val runCost = if (wfCount == 0) None else Some(wfCount * 1.23f)  // mockSubmissionCostService.fixedCost
+      // TODO David An 2018-05-30: temporarily disabling cost calculations for submission list due to potential performance hit
+      // val runCost = if (wfCount == 0) None else Some(wfCount * 1.23f)  // mockSubmissionCostService.fixedCost
+      val runCost = None
```
jmthibault79, line 423: I think this is right but I'm suspicious there's some subtlety with the mocks - I'm gonna run a few of these tests.

davidangb: `SubmissionApiServiceSpec` passed for me locally - I confess I did not run the other specs.

jmthibault79: all tests with "Submission" and "WorkspaceService" in their names passed

jmthibault79 (approved):

PR #941, opened 2018-06-05
Add BigQuery user role to workspace writers

dmohs:
https://app.zenhub.com/workspace/o/databiosphere/saturn-ui/issues/202

Workspace creation documentation does not contain information about roles granted. It probably should, but that is a separate issue.

It's not clear what the best response is for when adding the roles fails. In this case, the workspace is nearly entirely usable, so it may not make sense to return an error code. On the other hand, an error did occur.

Idea for next iteration: Include status of role grants in workspace response along with a separate endpoint for re-granting the default roles. This would allow a user to monitor the grant status and take action if role granting fails after the initial creation of the workspace.

As far as I can tell, FISMA docs don't cover the specific permissions granted to users of the system, so no modifications are necessary.

Regarding testing, if we were to create the groups and ensure the roles were assigned, we'd have an expensive test that mostly tests Google's infrastructure. My preferred test would simply ensure that the workspace creation flow calls the necessary role-granting Google APIs. I can't see a reasonable way to do this in the current codebase.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -275,10 +275,13 @@ class HttpGoogleServicesDAO(
     }
     results.flatMap { workspaceInfo =>
       val accessLevels = Set(WorkspaceAccessLevels.Owner, WorkspaceAccessLevels.Write)
-      val groupEmails = accessLevels.map { workspaceInfo.accessGroupsByLevel.get(_).head.groupEmail }
-
-      Future.traverse(groupEmails) { email => addRoleToGroup(project.projectName, WorkbenchEmail(email.value), "bigquery.jobUser") }.flatMap(_ => results)
-    }
+      val groupEmails =
+        accessLevels.map { workspaceInfo.accessGroupsByLevel.get(_).head.groupEmail }
+      addPolicyBindings(
+        project.projectName,
+        Map("roles/bigquery.jobUser" -> groupEmails.map { _.value }.toList)
+      )
+    }.flatMap { _ => results }
   }
 
   def workspaceAccessGroupName(workspaceId: String, accessLevel: WorkspaceAccessLevel) = s"${workspaceId}-${accessLevel.toString}"
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -260,7 +260,7 @@ class HttpGoogleServicesDAO(
       _ <- insertInitialStorageLog(bucketName)
     } yield GoogleWorkspaceInfo(bucketName, accessGroups, intersectionGroups)
 
-    val results = bucketInsertion recoverWith {
+    bucketInsertion recoverWith {
       case regrets =>
         val groupsToRollback = for {
           accessGroupTries <- accessGroupInserts
@@ -273,15 +273,6 @@ class HttpGoogleServicesDAO(
         }
         groupsToRollback flatMap rollbackGroups flatMap (_ => Future.failed(regrets))
     }
-    results.flatMap { workspaceInfo =>
-      val accessLevels = Set(WorkspaceAccessLevels.Owner, WorkspaceAccessLevels.Write)
-      val groupEmails =
-        accessLevels.map { workspaceInfo.accessGroupsByLevel.get(_).head.groupEmail }
-      addPolicyBindings(
-        project.projectName,
-        Map("roles/bigquery.jobUser" -> groupEmails.map { e => s"group:${e.value}" }.toList)
-      )
-    }.flatMap { _ => results }
   }
 
   def workspaceAccessGroupName(workspaceId: String, accessLevel: WorkspaceAccessLevel) = s"${workspaceId}-${accessLevel.toString}"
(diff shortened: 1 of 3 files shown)


dmohs:
@bradtaylor I had to open a new PR to workaround a bug in our CI scripts. @jmthibault79 had this question:
> How does this interact with AoU's requirements regarding Project Owners' granting and denial of this role to non-PO users?

dmohs:
@MatthewBemis 
> regarding testing: automation tests are failing because you forked the repo and apparently we can't build the rawls image in that case

Thank you. Jackie is filing the bug for me.

> in terms of adding an automation test, is there an endpoint you can call to list your bigquery roles? having a WS creator or writer call that endpoint to verify that this code worked wouldn't be too much work but i don't know if that endpoint exists

But that would primarily exercise Google's code—It answers the question, did calling Google's "grant role" endpoint actually grant the role? I have relatively high confidence that their code works, so I'm not inclined to test it. To test this code, I'd want to assert something more like, "Creating a workspace caused 'grant role' to be called on groups X and Y."

jmthibault79:
There's an existing automation test for BigQuery roles at https://github.com/broadinstitute/firecloud-orchestration/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/OrchestrationApiSpec.scala#L29

dmohs:
@jmthibault79 @MatthewBemis Thanks Joel! So that test proves that calling Google's "grant role" endpoint works correctly. I wouldn't want to re-test that, since it's expensive. In theory, the only remaining thing to test would be whether creating a workspace causes the necessary call to Google's endpoint. However, that is very near to just saying, "did the code I typed get executed," so, in my judgement, the test doesn't add enough value to be included.

dmohs:
Jenkins retest

dmohs:
Jenkins retest

MatthewBemis (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -273,6 +273,15 @@ class HttpGoogleServicesDAO(
         }
         groupsToRollback flatMap rollbackGroups flatMap (_ => Future.failed(regrets))
     }
+    results.flatMap { workspaceInfo =>
+      val accessLevels = Set(WorkspaceAccessLevels.Owner, WorkspaceAccessLevels.Write)
+      val groupEmails =
+        accessLevels.map { workspaceInfo.accessGroupsByLevel.get(_).head.groupEmail }
```
dvoet: you can get rid of the get and head: `workspaceInfo.accessGroupsByLevel(_).groupEmail`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -273,6 +273,15 @@ class HttpGoogleServicesDAO(
...
+      addPolicyBindings(
+        project.projectName,
+        Map("roles/bigquery.jobUser" -> groupEmails.map { e => s"group:${e.value}" }.toList)
+      )
+    }.flatMap { _ => results }
```
dvoet: I would prefer mapping the result of addPolicyBindings to workspaceInfo instead of the flatmap:
```
    results.flatMap { workspaceInfo =>
      val accessLevels = Set(WorkspaceAccessLevels.Owner, WorkspaceAccessLevels.Write)
      val groupEmails =
        accessLevels.map { workspaceInfo.accessGroupsByLevel(_).groupEmail }
      addPolicyBindings(
        project.projectName,
        Map("roles/bigquery.jobUser" -> groupEmails.map { e => s"group:${e.value}" }.toList)
      ).map(_ => workspaceInfo)
    }
```

MatthewBemis (approved):
re-approved!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -43,6 +43,27 @@ object UserService {
     new UserService(userInfo, dataSource, googleServicesDAO, gpsDAO, gpsGroupSyncTopic, notificationDAO, samDAO, projectOwnerGrantableRoles)
 
   case class OverwriteGroupMembers(groupRef: RawlsGroupRef, memberList: RawlsGroupMemberList)
+
+  def getGoogleProjectOwnerGroupEmail(samDAO: SamDAO, project: RawlsBillingProject)(implicit ec: ExecutionContext): Future[RawlsGroupEmail] = {
```
dvoet, line 47: I appreciate your DRYness but I generally don't like the pattern of static functions like this. It certainly is convenient but it smells like there are some wrong divisions of labor that 2 "unrelated" things need to share pretty specific code.

I think the right answer is that there is some yet undiscovered ProjectService that knows how to do project creation stuff that is called by the CreatingBillingProjectMonitor, called directly from the whatever routes register and delete projects, and disappears from UserService.

I realize that is probably bigger than you intended for this PR so if you want to create a techdebt jira, so be it. (not implying that you created the techdebt, only discovered it)

dvoet (approved):
I approve as is. I would approve even more were the techdebt you discovered addressed.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
dmohs: I completely agree. Since Brad is going to need this functionality soon, and restructuring may take some thought, I'm inclined to let it slide and fix in a follow-up. I do think that at least identifying the duplication in code (even via static functions) is a step in the right direction.
Follow-up tickets:
https://broadinstitute.atlassian.net/browse/GAWB-3667
https://broadinstitute.atlassian.net/browse/GAWB-3666

PR #965, opened 2018-07-20
use GET instead of HEAD for bucket access check

davidangb:


dvoet (approved):
approved for the hotfix, for logging OK requests should be logged too

davidangb:
jenkins retest

rtitle:
👍

PR #968, opened 2018-07-26
cache parseWDL results

davidangb:
wants but does not require broadinstitute/firecloud-develop#1284, since this PR has fallback values in reference.conf

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


davidangb:
Unit tests passed, but two Selenium/UI test failures from https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8069/
```
[info] *** 2 TESTS FAILED ***
[error] Failed: Total 124, Failed 2, Errors 0, Passed 122, Ignored 5
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.analysis.MethodConfigSpec
[error] 	org.broadinstitute.dsde.firecloud.test.methodrepo.MethodRepoSpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```
My gut says these are unrelated, but the `MethodConfigSpec` is suspicious and I will re-run both manually.

davidangb:
`MethodConfigSpec` passed in a manual run: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8073/
and
`MethodRepoSpec` passed in a manual run: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8074/

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+        logger.debug(s"<parseWDL-cache> encountered cache miss for $key in $miss ms.")
+        val parseResult = threadedParse(wdl)
+        val parsetime = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> actively parsed WDL for $key in $parsetime ms.")
+        put(key)(parseResult)
+        val tock = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> cached result $key in $tock ms.")
+        parseResult
+    }
+  }
```
davidangb: 1) TODO: this should not cache potentially-transient errors. We want to retry those.

2) Instead of a bounded thread pool in `threadedParse` below, we could synchronize the `parse` method above, using `key` as the synchronization object. Thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+     then this mode is not recommended. If the network goes down, your app could hang forever!"
+   */
+  import scalacache.modes.sync._
+  private val underlyingCaffeineCache = Caffeine.newBuilder()
+    .maximumSize(cacheMaxSize)
+    .build[String, Entry[Try[WdlWorkflow]]]
+  implicit val customisedCaffeineCache: Cache[Try[WdlWorkflow]] = CaffeineCache(underlyingCaffeineCache)
+
+  // set up thread pool
+  val executorService: ExecutorService = Executors.newFixedThreadPool(threadPoolSize)
```
rtitle: I'm wondering if `newCachedTreadPool` would be more appropriate then `newFixedTheadPool`.

See: https://stackoverflow.com/questions/949355/executors-newcachedthreadpool-versus-executors-newfixedthreadpool

Basically `newFixedTheadPool` will _always_ keep 8 threads ready, no matter what. `newCachedTheadPool` will spin up new threads as needed and terminate old ones. It's possible for >8 people to load an uncached MC at the same time (thinking about the workshop use case) and have to wait in line if limited to 8 threads. `newCachedTreadPool` would ensure that all requests are handled (and subsequently cached).

I could be wrong, just explaining my thought process.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+    logger.debug(s"<parseWDL-cache> looking up $key ...")
+    get(key) match {
+      case Some(parseResult) =>
+        val tock = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> found cached result for $key in $tock ms.")
+        parseResult
+      case None =>
+        val miss = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> encountered cache miss for $key in $miss ms.")
+        val parseResult = threadedParse(wdl)
```
rtitle: Might want to add a `blocking` call, e.g. `blocking(threadedParse(wdl))` to inform the global execution context that a blocking call is being made.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+        val parsetime = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> actively parsed WDL for $key in $parsetime ms.")
+        put(key)(parseResult)
```
rtitle: `parseResult` could be a `Failure` right? Maybe we should only cache successes, to ensure failures are retried.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
rtitle: 1. Ah agreed, I commented to that effect above.
2. 🤷‍♂️. It's either add synchronization or potentially do unnecessary work if >1 people load an uncached MC at the same time. I'd probably lean towards the latter (e.g. NOT adding synchronization, the way it is now). IMO this shouldn't be such an expensive operation that we need to synchronize.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+        val tock = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> cached result $key in $tock ms.")
+        parseResult
+    }
+  }
+
+  // an alternate terse/idiomatic implementation, but can't get good logging
+  /*
+  def parse(wdl: String): Try[WdlWorkflow] = {
```
rtitle: This does seem nicer. :) We could still add timing logs to the `threadedParse` method if using this approach. The timing of cache hits or misses is probably small and not as interesting.

rtitle:
Left a few comments, looks good overall. Even discounting memory leaks, this should have a nice impact to users by reducing response time of these endpoints.

core/src/main/resources/reference.conf
```diff
@@ -36,6 +36,12 @@ submissionmonitor {
   trackDetailedSubmissionMetrics = true
 }
 
+wdl-parsing {
+  cache-max-size = 2500
```
rtitle: How many WDLs are there in FireCloud? Wondering if we could size our cache to potentially hold everything.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: totally understand your thought process! If the objective is simply to get this in an independent execution context, I agree - but my goal here was explicitly to put a bound on how many requests could be loading an uncached MC simultaneously - impose a bottleneck/queue to avoid contention issues. I'm open to hearing this bottleneck is unnecessary!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: what I really want to log is "cache hit" vs "cache miss" so we can at least manually get some metrics out of it. I don't see how to log a cache hit in this syntax, but I'll dig in!

core/src/main/resources/reference.conf
davidangb: currently in prod - 17139 in agora, unknown quantity in dockstor, 10760 distinct references from method configs in a workspace. We could certainly up the cache - I'm totally guessing at a cache size based on how many might be in active use.

cjllanwarne: Is there any way to detect duplicates or is that already unique WDLs?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,98 @@
...
+    get(key) match {
+      case Some(parseResult) =>
+        val tock = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> found cached result for $key in $tock ms.")
+        parseResult
+      case None =>
+        val miss = System.currentTimeMillis() - tick
+        logger.debug(s"<parseWDL-cache> encountered cache miss for $key in $miss ms.")
+        val parseResult = threadedParse(wdl)
+        val parsetime = System.currentTimeMillis() - tick
```
cjllanwarne: `tock`?

core/src/main/resources/reference.conf
davidangb: the 10760 is unique workflow ids - it's certainly possible that people have the exact same WDL syntax inside multiple unique IDs. I bet we could find that out but ... is it worth it getting that answer?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
cjllanwarne: If it's true that

- it's pretty rare to keep refreshing on a broken WDL 
- it's relatively easy to get a transient failure resolving remote files due to a network blip
- the sttp fix should make things better in general

then... I think it's ok to only cache successes.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
rtitle: Ah I see. However the underlying issue didn't seem to be related to concurrency, just total # of requests. So enforcing a concurrency limit of 8 wouldn't actually protect against the issue, and may have the side effect of increasing user wait time in high-concurrency situations.

Then again, a fixed thread pool might still be a good idea to protect Rawls against many concurrent requests to this expensive code.

I also wonder if `CallableParser` should recheck the cache before running, in case it had to wait in line and a previous thread already cached the result.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
rtitle: I don't really have a strong preference, it's just code style.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
dvoet: if it is fast to parse (sub second) then I without sync is probably fine. But the longer it takes, the bigger the window for multiple requests to come in and get us in the situation this fix is trying to fix.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,119 @@
...
+  /*
+  def parse(wdl: String): Try[ParsedWdlWorkflow] = {
+    sync.caching(generateCacheKey(wdl))(ttl = None) {
+      threadedParse(wdl)
+    }
+  }
+  */
+
+  private def threadedParse(wdl: String): Try[ParsedWdlWorkflow] = {
+    try {
```
dvoet: capital T Try then you don't need the catch

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,119 @@
...
+      blocking {
+        executorService.invokeAny(List(new CallableParser(wdl)).asJava, threadPoolTimeout, TimeUnit.SECONDS)
```
dvoet: I am worried about thread starvation here in the case where parsing takes too long... a timeout cancels the task but that does not guarantee anything if the executing code does not look for interrupts. Is the idea to limit concurrency? Is concurrency the actual issue? Would a synchronization approach per wdl be better?

davidangb: yeah, I am also leaning towards synchronizing, at least in the short term until we truly nail down all the issues. I really hate synchronization but this is a band aid PR.

davidangb: now with `synchronized`.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: for whatever reason I couldn't get this to compile; I think it's something about the nested `Try`s. I'll give it another shot.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: I moved to `synchronize` per-WDL - and with `synchronize` in place, I think we don't need the extra cache lookup in `CallableParser`, and it's safer and more appropriate to use `newCachedTheadPool`.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: now with `synchronized`.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,120 @@
...
+  implicit val customisedCaffeineCache: Cache[Try[ParsedWdlWorkflow]] = CaffeineCache(underlyingCaffeineCache)
+
+  // set up thread pool
+  val executorService: ExecutorService = Executors.newCachedThreadPool()
+
+  def parse(wdl: String): Try[ParsedWdlWorkflow] = {
+    val tick = System.currentTimeMillis()
+    val key = generateCacheKey(wdl)
+    logger.info(s"<parseWDL-cache> looking up $key ...")
+    key.synchronized {
```
dvoet: I am not 100% certain this does what you think. This will only work as intended if generateCacheKey returns the same String instance for the same value. You could [intern](https://stackoverflow.com/questions/1091045/is-it-good-practice-to-use-java-lang-string-intern) the string but I am not sure that is a good idea. You could also use a concurrent map of locks.

I would also do a cache lookup before the sync block and return a hit if there is one. If a miss, sync then cache lookup again, return a hit or do the real parse for a miss.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
dvoet: if synchronized do we need a thread pool at all?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
rtitle: +1
> I would also do a cache lookup before the sync block and return a hit if there is one. If a miss, sync then cache lookup again, return a hit or do the real parse for a miss.

davidangb: I'm interning the string inside of `generateCacheKey`, so it should be properly synchronizing. I do understand that interning strings has other implications but it felt like the right tradeoff here.

Agree on unsynchronized cache lookup then syncing the whole thing only on miss - I'll work on that.

dvoet: just a style point then... I would intern at the point of sync because that is where it actually matters...

geoffjentry: Would `TrieMap` be helpful in all of this?

davidangb: @geoffjentry explain? Are you thinking of maintaining our own map of locks?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
```diff
@@ -0,0 +1,131 @@
...
+        executorService.invokeAny(List(new CallableParser(wdl)).asJava, threadPoolTimeout, TimeUnit.SECONDS)
+      }
+    } catch {
+      case e: Exception => Failure(e)
+    }
+  }
+
+  private def generateCacheKey(wdl: String): String = {
+    // reduce key size to keep cache small. No attachment to the Murmur algorithm specifically.
+    MurmurHash3.stringHash(wdl).toString.intern()
```
aednichols: I worry that HTTP imports with an unstable URL - like `https://raw.githubusercontent.com/DataBiosphere/topmed-workflows/topmed-firecloud-demo/functional-equivalence-wdl/FunctionalEquivalence.wdl` - could affect correctness

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
geoffjentry: I hesitated to even make the comment as I hadn't followed this super closely and worried about it being anti-helpful, but I was pattern matching on seeing the convos about caching, locking, etc and wondering out loud if using a `TrieMap` for the cache would make sense? It's both thread safe and lock free.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
dvoet: if this is something that people actually do (use unstable URLs) they are open to a world of hurt... essentially provenance in firecloud is worthless

aednichols: I've seen mostly stable URLs outside of testing, but we were concerned enough about this that we warn users who come in via the "Launch with FireCloud" button in Dockstore
>Dockstore cannot guarantee that the WDL and Docker image referenced by this Workflow will not change. We advise you to review the WDL before future runs.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: now parsing outside the thread pool.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
davidangb: ah - @geoffjentry we're using a Caffeine-backed scalacache.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/wdlparsing/CachingThreadPoolWDLParser.scala
aednichols: FWIW I am perfectly happy to tell users, "don't do that" - however, I feel that is a PO decision

davidangb:
Three Selenium/UI failures that I believe are unrelated in https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8096/. I will re-run these manually:
```
[info] *** 3 TESTS FAILED ***
[error] Failed: Total 124, Failed 3, Errors 0, Passed 121, Ignored 5
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.library.UnpublishedWorkspaceSpec
[error] 	org.broadinstitute.dsde.firecloud.test.library.DataUseAwareSearchSpec
[error] 	org.broadinstitute.dsde.firecloud.test.security.AuthDomainGroupSpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

updates:
* UnpublishedWorkspaceSpec passed manually: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8097/
* DataUseAwareSearchSpec passed manually: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8098/
* AuthDomainGroupSpec passed manually: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/8099/

dvoet (approved):

rtitle (approved):
Only comment is regarding string interning - your comment references PermGen but it’s no longer true in Java 8 that interned strings are stored in PermGen. Interned strings go in main heap, and there’s a JVM setting to configure the string table size (defaults to some big number). This article has a good explanation of the differences between java versions:

http://java-performance.info/string-intern-in-java-6-7-8/

Based on this it might be ok to intern full wdls, especially since they’re stored as cache keys anyway. 

But I think the current approach is fine, this is more nit picky than anything. I approve as is. 👍

davidangb:
oh, nice @rtitle thanks for that info! I didn't know that.

PR #998, opened 2018-10-22
Phase4 test redo

dvoet:
This is *the* PR for moving workspace security to sam and cutting the cord to opendj!

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -406,410 +365,312 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
 
-  private def updateCatalogPermissions(catalogUpdates: Seq[WorkspaceCatalog], dataAccess: DataAccess, workspaceContext: SlickWorkspaceContext): ReadWriteAction[WorkspaceCatalogUpdateResponseList] = {
+  /**
+   * updates acls for a workspace
+   * @param workspaceName
+   * @param aclUpdates changes to make, if an entry already exists it will be changed to the level indicated in this
+   *                   Set, use NoAccess to remove an entry, all other preexisting accesses remain unchanged
+   * @return
+   */
+  def updateACL(workspaceName: WorkspaceName, aclUpdates: Set[WorkspaceACLUpdate], inviteUsersNotFound: Boolean): Future[PerRequestMessage] = {
```
dvoet, line 524: @MatthewBemis I took another stab at this implementation, thought you might want a look

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -73,6 +69,40 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 
+  def listUserPoliciesForResource(resourceTypeName: SamResourceTypeName, resourceId: String, userInfo: UserInfo): scala.concurrent.Future[Set[SamPolicyWithName]] = ???
```
dvoet: there are a few unimplemented calls to sam as yet... will get back to them once implemented in sam

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -22,7 +22,7 @@ import slick.jdbc.{GetResult, JdbcProfile}
 case class SubmissionRecord(id: UUID,
                             workspaceId: UUID,
                             submissionDate: Timestamp,
-                            submitterId: String,
+                            submitterEmail: String,
```
Qi77Qi, line 25: having type of this field as `WorkbenchEmail` seems better

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+case class SamResourceTypeName(value: String)
+
+object SamResourceTypeNames {
+  val billingProject = SamResourceTypeName("billing-project")
+  val managedGroup = SamResourceTypeName("managed-group")
+  val workspace = SamResourceTypeName("workspace")
+}
+
```
Qi77Qi, line 18: I would make it sealed
```
sealed abstract class SamResourceTypeName{
def value: String
}
object SamResourceTypeName{
  case object BillingProject extends SamResourceTypeName{
     def value: String = "billing-project"
  }
  case object ManagedGroup extends SamResourceTypeName{
     def value: String = ...
  }
  case object Workspace extends SamResourceTypeName{
   def value: String = ...
  }
}

```

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -0,0 +1,107 @@
...
+/*
+  Resource roles
+ */
+
+case class SamResourceRole(value: String) extends ValueObject
+
```
Qi77Qi, line 24: same comment as previous one

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -0,0 +1,107 @@
...
+  val owner = SamResourcePolicyName("owner")
+  val workspaceCreator = SamResourcePolicyName("workspace-creator")
+  val canComputeUser = SamResourcePolicyName("can-compute-user")
+}
+
+case class SamPolicy(memberEmails: Set[WorkbenchEmail], actions: Set[SamResourceAction], roles: Set[SamResourceRole])
+case class SamPolicyWithName(policyName: SamResourcePolicyName, policy: SamPolicy)
+case class SamPolicyWithNameAndEmail(policyName: SamResourcePolicyName, policy: SamPolicy, email: WorkbenchEmail)
+case class SamResourceWithPolicies(resourceId: String, policies: Map[SamResourcePolicyName, SamPolicy], authDomain: Set[String])
+case class SamResourceIdWithPolicyName(resourceId: String, accessPolicyName: SamResourcePolicyName, authDomains: Set[String], missingAuthDomains: Set[String], public: Option[Boolean])
```
Qi77Qi: does `public` really needs to be `Option[Boolean]` as opposed to be `Boolean`?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -0,0 +1,107 @@
...
+  val canCatalog = SamResourcePolicyName("can-catalog")
+}
+
+object SamBillingProjectPolicyNames {
+  val owner = SamResourcePolicyName("owner")
+  val workspaceCreator = SamResourcePolicyName("workspace-creator")
+  val canComputeUser = SamResourcePolicyName("can-compute-user")
+}
+
+case class SamPolicy(memberEmails: Set[WorkbenchEmail], actions: Set[SamResourceAction], roles: Set[SamResourceRole])
```
Qi77Qi, line 92: I would add `final` to all these case classes if you don't expect them to be extended

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -406,410 +365,312 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-      if (membersWithHigherExistingAccessLevelThanGranter.nonEmpty) {
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"You may not alter the access level of users with higher access than yourself. Please correct these entries: $membersWithHigherExistingAccessLevelThanGranter"))
-      }
-
-      val membersWithHigherAccessLevelThanGranter = actualAccessChangesToMake.filter { case (_, accessLevel) => accessLevel > userAccessLevel }
-
-      if (membersWithHigherAccessLevelThanGranter.nonEmpty) {
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"You may not grant higher access than your own access level. Please correct these entries: $membersWithHigherAccessLevelThanGranter"))
-      }
+  private def validateAclChanges(aclChanges: Set[WorkspaceACLUpdate], existingAcls: Set[WorkspaceACLUpdate]) = {
```
Qi77Qi, line 639: I would make this return `Either[RawlsExceptionWithErrorReport, Set[WorkspaceACLUpdate]]` if it's easy refatoring

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -406,410 +365,312 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-      }
-      if(membersWithComputePermission.nonEmpty && userAccessLevel < WorkspaceAccessLevels.Owner) {
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"You may not alter the compute permissions of users unless you are a workspace owner. Please correct these entries: $membersWithComputePermission"))
-      }
-      if(refsToUpdate.exists {
-        case (_, (WorkspaceAccessLevels.Read, _, Some(true))) => true
-        case _ => false
-      }) {
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Readers may not be granted compute permissions"))
+  private def sendACLUpdateNotifications(workspaceName: WorkspaceName, usersModified: Set[WorkspaceACLUpdate]) {
```
Qi77Qi: add return type and `=`...I think procedure syntax is deprecated

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  val callsToAddToPolicy = new ConcurrentLinkedDeque[(SamResourceTypeName, String, SamResourcePolicyName, String)]()
+  val callsToRemoveFromPolicy = new ConcurrentLinkedDeque[(SamResourceTypeName, String, SamResourcePolicyName, String)]()
+
+  override def registerUser(userInfo: UserInfo): Future[Option[UserStatus]] = {
+    userEmails.put(userInfo.userEmail.value, Option(userInfo.userSubjectId.value))
+    Future.successful(Option(UserStatus(RawlsUser(userInfo.userSubjectId, userInfo.userEmail), Map.empty)))
+  }
+
+  override def getUserIdInfo(userEmail: String, userInfo: UserInfo): Future[Either[Unit, Option[UserIdInfo]]] = {
```
Qi77Qi: the return type seems a bit suspicious...`Future[Option[UserIdInfo]]` is not enough?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceACL.scala
```diff
@@ -53,17 +54,31 @@ object WorkspaceAccessLevels {
...
-      case "PROJECT_OWNER" => ProjectOwner
-      case "OWNER" => Owner
-      case "WRITER" => Write
-      case "READER" => Read
-      case "NO ACCESS" => NoAccess
+      case accessLevel if accessLevel.equalsIgnoreCase("PROJECT_OWNER") => ProjectOwner
+      case accessLevel if accessLevel.equalsIgnoreCase("OWNER") => Owner
+      case accessLevel if accessLevel.equalsIgnoreCase("WRITER") => Write
+      case accessLevel if accessLevel.equalsIgnoreCase("READER") => Read
+      case accessLevel if accessLevel.equalsIgnoreCase("NO ACCESS") => NoAccess
```
Qi77Qi, line 73: would be nice if this function can return `Either` if refactoring isn't too much effort

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
MatthewBemis: i chose this type, so i'll chime in. there's three states that need to be represented, so `Option[UserIdInfo]]` isn't enough.

1) the email doesn't exist in the system at all: `Left(())`
2) the email exists and belongs to a managed group: `Right(None)`
3) the email exists and belongs to a user: `Right(Some(userIdInfo))`

all i could think of were ugly ways of representing this so if there is a more idiomatic way to represent it i'd love to learn about it!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
```diff
@@ -63,4 +63,6 @@
     <include file="changesets/20180409_index_audit_workflow_status.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20180425_optional_root_entity.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20180716_cromwell_backend_per_project.xml" relativeToChangelogFile="true"/>
+    <!--<include file="changesets/20180920_submitter_email.xml" relativeToChangelogFile="true"/>-->
```
MatthewBemis, line 66: remember to uncomment this too

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
MatthewBemis: backstory: it was only returning an option because during development the Sam endpoint wasn't yet updated to return that field. it can just be Boolean now

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
Qi77Qi: ah I see...I would make it an [ADT](http://tpolecat.github.io/presentations/algebraic_types.html#1) in this case.
something like:
```
sealed trait GetUserIdInfoResult extends Product with Serializable
final case object NotFound
final case object BelongToManagedGroup
final case class BelongToUser(userInfo: UserInfo)
```
And return `Future[GetUserIdInfoResult]`...it'll be easier if in the future we support more cases too

core/src/main/scala/org/broadinstitute/dsde/rawls/statistics/StatisticsService.scala
```diff
@@ -59,11 +59,9 @@ class StatisticsService(protected val userInfo: UserInfo, val dataSource: SlickD
       }
 
       DBIO.sequence(actions).flatMap { results =>
-        dataAccess.rawlsUserQuery.countUsers() flatMap { numUsers =>
-          entityStatistics.countEntitiesofTypeInNamespace(workspaceNamespace, workspaceName) map { entityStats =>
-            val allResults = results.toMap + ("currentTotalUsers" -> numUsers) + ("currentEntityStatistics" -> EntityStatistics(workspaceNamespace, workspaceName, entityStats))
-            RequestComplete(StatusCodes.OK, StatisticsReport(startDate, endDate, allResults))
-          }
+        entityStatistics.countEntitiesofTypeInNamespace(workspaceNamespace, workspaceName) map { entityStats =>
+          val allResults = results.toMap + ("currentEntityStatistics" -> EntityStatistics(workspaceNamespace, workspaceName, entityStats))
```
helgridly, line 63: who's this gonna affect?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -323,49 +290,39 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-      val results = query.map { case (permissionsPairs, managedGroupsForUser, ownerEmails, publicWorkspaces, submissionSummaryStats, workspaces) =>
-        val workspacesById = workspaces.groupBy(_.workspaceId).mapValues(_.head)
-        permissionsPairs.map { permissionsPair =>
-          val groupsForUser = managedGroupsForUser.map(group => ManagedGroupRef(group.groupName)).toSet
-
-          workspacesById.get(permissionsPair.workspaceId).map { workspace =>
-            def trueAccessLevel = {
-              if((workspace.authorizationDomain -- groupsForUser).nonEmpty) WorkspaceAccessLevels.NoAccess
-              else permissionsPair.accessLevel
+  def listWorkspaces(lite: Boolean): Future[PerRequestMessage] = {
```
helgridly: `lite`: unused?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -406,410 +365,312 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-          requireAccessIgnoreLock(workspaceContext.workspace, WorkspaceAccessLevels.Write, dataAccess) {
-            DBIO.sequence(workspaceContext.workspace.accessLevels.values.map {group =>
-              dataAccess.rawlsGroupQuery.flattenGroupMembership(group)
-            })
-          }
-        }
+  def validateUsersExist(userEmails: Set[String]): Future[Set[Option[String]]] = {
+    Future.traverse(userEmails) { email =>
+      samDAO.getUserIdInfo(email, userInfo).map {
+        case Left(()) => Option(email)
```
helgridly: without looking at the implementation... what the hell? why does this return a Left-of-Unit? it's a get function, shouldn't it be getting something?

helgridly (commented):
i can't honestly say i'm able to review this in any meaningful way, but i tried and made a few halfhearted comments

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: context: i put this in here with the intention of making fewer unnecessary calls to sam (for reading the owner policy for each workspace). didn't end up doing anything with it because i didn't have any proof that it'd actually be a noticeable improvement. would be interested to know how listWorkspaces performs with several hundred workspaces. if it's poor, then this would be useful for fc-ui, saturn, etc

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
dvoet: this is a database object representation, types need to match database types (unless we want to write the custom serializers which I don't want to right now). The way it is matches everything else so I am going to leave it alone.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: I am removing lite but I think my plan is to also remove the owners from the list workspace call anyway as it is not displayed.

core/src/main/scala/org/broadinstitute/dsde/rawls/statistics/StatisticsService.scala
dvoet: the stats end point has not been called by anything other than https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/metrics/MetricsActor.scala in the past 90 days and that doesn't care. As the comment says
> // out of all the metrics Rawls returns, we only care to extract the number of samples right

Qi77Qi (approved):

PR #1000, opened 2018-10-23
[experimental] speculative refactor of rawls, for discussion

helgridly:
I think this is where we should be aiming to get to with rawls within ~a year. Thoughts?


MatthewBemis (approved):
LGTM 👍

helgridly:
cold.

helgridly:
so much for thinking about the future!!!

PR #1006, opened 2018-11-02
[GAWB-3829] Create workflow collections for newly generated workspaces

helgridly:
Friends with https://github.com/broadinstitute/workbench-libs/pull/175 for autotest purposes.

**NOTE TO SELF**: update to a proper wb-libs hash once friend is merged and published.

Closes https://broadinstitute.atlassian.net/browse/GAWB-3834 too.

(third PR! chasing the doge merge train!)

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1699,6 +1699,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     withAttributeNamespaceCheck(attrNames)(op)
   }
 
+  private def createWorkflowCollectionForWorkspace(workspaceId: String) = {
```
helgridly, line 1723: this is the meat

helgridly:
ok my jenkins failure is the same as doge's so i'm declaring myself blocked here

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -39,6 +40,12 @@ object SamProjectRoles {
   val owner = SamResourceRole("owner")
 }
 
+object SamWorkflowCollectionRoles {
+  val owner = SamResourceRole("owner")
```
Qi77Qi, line 46: does it make sense to make this an ADT?
something like
```
sealed abstract class SamWorkflowCollectionRole{
  def role: SamResourceRole
}
object SamWorkflowCollectionRoles{
  case object Owner extends SamWorkflowCollectionRole{
    val role: SamResourceRole = SamResourceRole("owner")
  }  
  case object Writer extends SamWorkflowCollectionRole{
    val role: SamResourceRole = SamResourceRole("writer")
  }  
  case object Reader extends SamWorkflowCollectionRole{
    val role: SamResourceRole = SamResourceRole("reader")
  }
}
```
Would've been nice if `SamResourceRole` is also an ADT

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -86,6 +93,12 @@ object SamBillingProjectPolicyNames {
   val canComputeUser = SamResourcePolicyName("can-compute-user")
 }
 
+object SamWorkflowCollectionPolicyNames {
```
Qi77Qi, line 99: same comment as previous one

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1699,6 +1699,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      policyMap = workspacePolicies.map(pol => pol.policyName -> pol.email).toMap
+      _ <- samDAO.createResource(SamResourceTypeNames.workflowCollection, workspaceId, userInfo)
+      _ <- samDAO.overwritePolicy(SamResourceTypeNames.workflowCollection, workspaceId, SamWorkflowCollectionPolicyNames.workflowCollectionOwnerPolicyName,
+        SamPolicy(Set(policyMap(SamWorkspacePolicyNames.projectOwner), policyMap(SamWorkspacePolicyNames.owner)), Set.empty, Set(SamWorkflowCollectionRoles.owner)), userInfo)
+      _ <- samDAO.overwritePolicy(SamResourceTypeNames.workflowCollection, workspaceId, SamWorkflowCollectionPolicyNames.workflowCollectionWriterPolicyName,
+        SamPolicy(Set(policyMap(SamWorkspacePolicyNames.writer)), Set.empty, Set(SamWorkflowCollectionRoles.writer)), userInfo)
+      _ <- samDAO.overwritePolicy(SamResourceTypeNames.workflowCollection, workspaceId, SamWorkflowCollectionPolicyNames.workflowCollectionReaderPolicyName,
+        SamPolicy(Set(policyMap(SamWorkspacePolicyNames.reader)), Set.empty, Set(SamWorkflowCollectionRoles.reader)), userInfo)
```
Qi77Qi: can these 3 overwrite requests be done in parallel?

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -132,6 +130,31 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with FreeSpecLike with
       }
     }
 
+    "should create and destroy a workflow collection resource in Sam for a workspace" in {
```
dvoet, line 135: while this is ok as an integration test, it would be better as a unit test that verifies the correct sam api calls are made.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: or use samDAO.createResourceFull

dvoet: These policies should have members. The writer policy should have the email address of the workspace writer policy, the reader policy should have the email address of the workspace reader policy. The owner policy should have the email address of the workspace owner and project owner policies. I don't think we need a separate project owner policy on workflow collections.

dvoet (changes requested):
I don't see the call to delete the workflow-collection when the workspace is deleted.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: createResourceFull was a pain for some reason but I can take another look.

re adding the workspace policy members to the collection policies -- isn't that what passing in `SamPolicy(Set(policyMap(SamWorkspacePolicyNames.reader` does? The values of `policyMap` are the email addresses of the workspace policies (see line 1705).

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
helgridly: wouldn't that only test that i'd called the functions i just typed? imo this way tests the behaviour, which is better... plus it's resilient to implementation changes, like moving to `createResourceFull`

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
helgridly: @dvoet do you have an opinion here? i think this stuff is in libs too

helgridly:
@dvoet can you take another look please

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
dvoet: I look at these things as the equivalent of string constants, we have them mostly so we don't have the same string all over the place that is then hard to change. We never actually pattern match on them or do other things where an ADT would be useful. There are plenty of places where we swap around roles and policy names on purpose and an ADT would just get in the way. Seems like work for little or no gain.

But if it were an ADT, I would be fine with it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: yup, sorry, misread that... there are too many sets

but createResourceFull is the right thing. Not only is it fewer sam api calls, but createResource will also create a default policy for you which you don't want.

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
dvoet: this certainly tests more, namely that the sam dao is right, but integration tests are less accessible than unit tests. Where I think we should get is to use swagger to generate a sam client lib. Use that client lib in the sam integration level tests. Use the client libs in rawls knowing they are well tested. Then this should be a unit test and ease of use trumps behavior testing since we don't actually want to test sam's behavior.

but we are not there yet so I will let it go.

dvoet (approved):

dvoet (approved):

helgridly:
The rebasing never ends. Closing in favour of https://github.com/broadinstitute/rawls/pull/1020

PR #1032, opened 2019-01-23
read committed isolation updating workspace attr scratch

dvoet:
https://broadworkbench.atlassian.net/browse/WB-219

A deadlock on the workspace attribute scratch table has been causing timeouts. We think switching to read committed isolation level will help.

Note that this should be ok in the case of concurrent modification of the same workspace because there is an optimistic concurrency check.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -125,11 +125,11 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
   def createWorkspace(workspaceRequest: WorkspaceRequest): Future[Workspace] =
     withAttributeNamespaceCheck(workspaceRequest) {
-      dataSource.inTransaction { dataAccess =>
+      dataSource.inTransaction({ dataAccess =>
         withNewWorkspaceContext(workspaceRequest, dataAccess) { workspaceContext =>
           DBIO.successful(workspaceContext.workspace)
         }
-      }
+      }, TransactionIsolation.ReadCommitted) // read committed to avoid deadlocks on workspace attr scratch table
     }
 
   def getWorkspace(workspaceName: WorkspaceName): Future[PerRequestMessage] =
@@ -268,9 +268,9 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
         isCurator <- tryIsCurator(userInfo.userEmail)
         workspace <- getWorkspaceContext(workspaceName) flatMap { ctx =>
           withLibraryPermissions(ctx, operations, userInfo, isCurator) {
-            dataSource.inTransaction { dataAccess =>
+            dataSource.inTransaction ({ dataAccess =>
               updateWorkspace(operations, dataAccess)(ctx)
-            }
+            }, TransactionIsolation.ReadCommitted) // read committed to avoid deadlocks on workspace attr scratch table
           }
         }
         authDomain <- loadResourceAuthDomain(SamResourceTypeNames.workspace, workspace.workspaceId, userInfo)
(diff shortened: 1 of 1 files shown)


dvoet:
jenkins retest - prior run failed due to 412 Precondition Failed error deleting google group;

```
[DEBUG] [15:46:49.481] [scala-execution-context-global-44398] o.b.d.w.g.HttpGoogleDirectoryDAO - {"method":"DELETE","url":"https://www.googleapis.com/admin/directory/v1/groups/policy-e8b1e1c5-e089-4d40-a094-881c025391d6@quality.firecloud.org","time_ms":291,"statusCode":412}
[INFO] [15:46:49.481] [scala-execution-context-global-44398] o.b.d.w.g.HttpGoogleDirectoryDAO - retry-able operation failed: retries remain but predicate failed, not retrying
com.google.api.client.googleapis.json.GoogleJsonResponseException: 412 Precondition Failed
{
  "code" : 412,
  "errors" : [ {
    "domain" : "global",
    "location" : "If-Match",
    "locationType" : "header",
    "message" : "Condition not met",
    "reason" : "conditionNotMet"
  } ],
  "message" : "Condition not met"
}
```

helgridly (approved):
i approve that the code does the thing it says it does, but i don't like how RC continues to creep in as a cure-all for deadlocks

davidangb (approved):
:+1: don't know if this will fix the problem but changes look fine.

dvoet:
@helgridly can you explain why you don't like read committed as a solution to dead locks? The theory is that that mysql's implementation of repeatable read creates unexpected locks in unpredictable order leading to deadlocks so read committed is the right solution.

Honestly, I think read committed should be the default as it allows for more concurrency and mostly we don't care but it is a scary thing to change across the board.

dvoet:
jenkins retest

helgridly:
@dvoet my discomfort is borne of ignorance and an abundance of caution. we've generally been using RR and have selectively moved to RC only when things have been deadlocky. admittedly doing so hasn't caused us problems so far but every time we do it, i'm wary that changing the transaction semantics will have unexpected effects.

that said i'd never used a database until i started at broad so this is not my wheelhouse :) if we'd been using RC from the beginning i'd have more confidence in it, because we'd either know its pitfalls or have a history of it seeming to be fine.

"it is a scary thing" about sums up my feeling here.

breilly2:
@helgridly Doug and I chatted about this (see [slack thread](https://broadinstitute.slack.com/archives/C2B7JFJ4S/p1548363560223000) in #dsp-workbench from this morning). I agree 100% with your concerns. RC puts us at risk for a different class of problems. When the application makes multiple queries in a single transaction, the in-memory copy of the data may be inconsistent and/or validate database constraints which could cause the application to make very unexpected decisions (the `throw new Exception("this should never happen")` kind).

I suggested that we need to keep an eye on this and, if we continue to migrate transactions from RR to RC, we should:
1. Gather all of the DB-interfacing developers for a presentation/discussion of what it means with examples of potential errors (I'm happy to help with this).
2. Consider changing our default stance to RC, both for existing and new applications. As you said, it's easier to trust when it's like that from the beginning.

It's really important to have a good understanding of this because the transaction isolation declaration is higher in the call stack and often far from the queries and related code that need to consider the consequences of that choice.

helgridly:
yeah, i saw the thread :) i would support a move to RC for new applications; i think the effort to figure out what would happen if we moved existing things from RR to RC would probably be moderate, and we've had few (no?) instances of RC causing us issues, so maybe not worth doing until we need to. i still have a dream that one day we'll stop relying on rawls for this stuff, and it's generally been attribute updates that have been deadlocking, so fingers crossed that One Day :tm: this will magically resolve itself and vanish into the ether.

PR #1038, opened 2019-02-14
[WIP] GAWB-4074 upload cromwell metadata to gcs

Qi77Qi:
Part of https://broadinstitute.atlassian.net/browse/GAWB-4074

This will upload metadata to dsde project. Since rawls already have the credentials, and this change might be reverted in the future in Rawls, it seems reasonable to just reuse the existing project for keeping metadata objects. Let me know if this is not desired.

Currently, I added notification topic for this bucket in my test project, but that should probably be real hamm projects once we have them.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
@@ -162,7 +162,8 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
       "jes",
       "papi2",
       "executionStatus",
-      "backend"
+      "backend",
+      "attempt"
     ), Set.empty, true)
 
     for {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
@@ -148,6 +148,32 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
     }
   }
 
+  def callLevelMetadataForCostCalculation(submissionId: String, workflowId: String, execId: Option[ExecutionServiceId], userInfo: UserInfo): Future[JsObject] = {
+    val metadataParams = MetadataParams(Set(
+      "id",
+      "start",
+      "end",
+      "labels",
+      "executionEvents",
+      "runtimeAttributes",
+      "jobId",
+      "preemptible",
+      "callCaching", //We only need "hit" key in callCaching. Would be nice to filter everything else out, but not sure if cromwell supports that currently
+      "hit",
+      "jes",
+      "executionStatus",
+      "backend"
+    ), Set.empty, true)
+
+    for {
+      executionServiceId <- findExecService(submissionId, workflowId, userInfo, execId)
+      metadata <- getMember(executionServiceId).dao.callLevelMetadata(workflowId, metadataParams, userInfo)
+    } yield {
+      labelSubWorkflowsWithSubmissionId(submissionId, executionServiceId, metadata, userInfo)
+      metadata
+    }
+  }
+
   // ====================
   // facade-to-cluster entry points
   // ====================
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
@@ -168,10 +168,7 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
     for {
       executionServiceId <- findExecService(submissionId, workflowId, userInfo, execId)
       metadata <- getMember(executionServiceId).dao.callLevelMetadata(workflowId, metadataParams, userInfo)
-    } yield {
-      labelSubWorkflowsWithSubmissionId(submissionId, executionServiceId, metadata, userInfo)
-      metadata
-    }
+    } yield metadata
   }
 
   // ====================
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
@@ -158,9 +158,9 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
       "runtimeAttributes",
       "jobId",
       "preemptible",
-      "callCaching", //We only need "hit" key in callCaching. Would be nice to filter everything else out, but not sure if cromwell supports that currently
-      "hit",
+      "callCaching:hit",
       "jes",
+      "papi2",
       "executionStatus",
       "backend"
     ), Set.empty, true)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -296,9 +296,8 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
   private def uploadMetadataToGCS(submissionId: String, workflowId: String, petUser: UserInfo)(implicit executionContext: ExecutionContext): Future[Unit] = {
     for{
       metadata <- executionServiceCluster.callLevelMetadataForCostCalculation(submissionId, workflowId, None, petUser)
-//    If this is a subworkflow, we don't upload its metadata; else we upload it to GCS
-      _ <- if(metadata.fields.keys.toList.contains("parentWorkflowId")) Future.successful(())
-      else googleServicesDAO.storeObject(config.metadataBucketNamePrefix, GcsObjectName(workflowId), metadata.compactPrint.getBytes("UTF-8"))
+      // Rawls only monitor root level workflow
+      _ <- googleServicesDAO.storeCromwellMetadata(GcsObjectName(workflowId), metadata.compactPrint.getBytes("UTF-8"))
     } yield ()
   }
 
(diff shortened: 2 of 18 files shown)


dvoet:
bucket names have a global namespace... they don't necessarily need to contain uuids but certainly need to be unique between environments. I don't think that the token bucket has a uuid, though. I think it is the id of the service account. Generally though, you want it to be predictable so you are not creating new buckets all the time.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1561,7 +1564,17 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     // query the execution service(s) for the metadata
     execIdFutOpt flatMap {
       executionServiceCluster.callLevelMetadata(submissionId, workflowId, metadataParams, _, userInfo)
-    } map(RequestComplete(StatusCodes.OK, _))
+    } map {
+      metadata =>
+        // asynchronously upload metadata to GCS.
+        // Currently, this is only used for hamm, and may potentially be removed from RAWLS if design changes
+        uploadMetadataToGCS(workflowId, metadata.compactPrint.getBytes("UTF-8"))
+        RequestComplete(StatusCodes.OK, metadata)
```
dvoet, line 1568: 2 problems here:
1) reading the comment above, this may not retrieve sub workflow metadata so that might be missing
2) this function is called only when the user asks for the metadata which may never happen

Instead there should be code that executes when a workflow completes (probably in neighborhood of SubmissionMonitorActor.handleOutputs). And it probably needs to force getting sub workflows... but we ran into memory problems with lots of sub workflows... maybe we need to do our own traversal?

dvoet: the metadata api call to cromwell also takes an `includeKeys` parameter. I think we should figure out what keys we need and include only those to keep the json small (we definitely don't need all the inputs and outputs). Maybe lots of sub workflows won't be a problem if we only include the data we need.

ansingh7115: We're not using the inputs or outputs, and we only need a few of the execution events, not all, so our jsons will be significantly smaller.
For the keys that we're currently using:
At the workflow level - "id", "start", "end", "labels"
per call - "runtimeAttributes", "jobId", "executionEvents", "preemptible", "callCaching", "jes", "executionStatus", "backend", "attempt"
It's in hamm's `CromwellMetadataJsonCodec.scala` file

ansingh7115: From David An's `Cromwell metadata requests and FireCloud` [doc](https://docs.google.com/document/d/16Xv36NYiIlw6PJkBVyrxSRWbOpjpY7tI4UQGENZi7gc/edit#):

> the large majority of the payload is contained in its **inputs**, with the next largest portions being **outputs, executionEvents, and callCaching:hashes**. There are some additional fields such as **jes, runtimeAttributes, labels, backendLabels**, and **submittedFiles** that also contribute to the payload size, but are dwarfed by the main offenders.

We're not using the inputs or outputs, and we only need a few of the execution events, not all, so our jsons will be significantly smaller.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
```diff
@@ -148,6 +148,32 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
...
+      "papi2",
+      "executionStatus",
+      "backend"
+    ), Set.empty, true)
+
+    for {
+      executionServiceId <- findExecService(submissionId, workflowId, userInfo, execId)
+      metadata <- getMember(executionServiceId).dao.callLevelMetadata(workflowId, metadataParams, userInfo)
+    } yield {
+      labelSubWorkflowsWithSubmissionId(submissionId, executionServiceId, metadata, userInfo)
```
dvoet: why labelSubWorkflowsWithSubmissionId here?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -277,12 +278,26 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
   private def execServiceOutputs(workflowRec: WorkflowRecord, petUser: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[(WorkflowRecord, Option[ExecutionServiceOutputs])]] = {
     WorkflowStatuses.withName(workflowRec.status) match {
       case WorkflowStatuses.Succeeded =>
+        // asynchronously upload metadata to GCS.
+        // Currently, this is only used for hamm, and may potentially be removed from RAWLS if design changes
+        workflowRec.externalId.traverse{
+          workflowId =>
+            uploadMetadataToGCS(workflowRec.submissionId.toString, workflowId, petUser)
```
dvoet: this only runs for Succeeded workflows... I think money spent on failed workflows will also be interesting

dvoet: also since these Futures are not monitored by anything, we should handle and log any errors

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -277,12 +278,26 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-
       case _ => Future.successful(Option((workflowRec, None)))
     }
   }
 
+  private def uploadMetadataToGCS(submissionId: String, workflowId: String, petUser: UserInfo)(implicit executionContext: ExecutionContext): Future[Unit] = {
+    for{
+      metadata <- executionServiceCluster.callLevelMetadataForCostCalculation(submissionId, workflowId, None, petUser)
+//    If this is a subworkflow, we don't upload its metadata; else we upload it to GCS
+      _ <- if(metadata.fields.keys.toList.contains("parentWorkflowId")) Future.successful(())
```
dvoet: I don't think this will ever be true, a) we don't ask for the parentWorkflowId key and b) we only monitor the top level workflow

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -277,12 +278,26 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+      else googleServicesDAO.storeObject(GcsBucketName("cromwell_metadata"), GcsObjectName(workflowId), metadata.compactPrint.getBytes("UTF-8"))
```
dvoet: pull bucket name from config and there should be code to create the bucket on startup if it does not exist.

dvoet (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
Qi77Qi: I don't follow...Do you mean https://github.com/broadinstitute/rawls/pull/1038/files#diff-90caa3307e9614c29824ffa11e2fef55R285 already only has top level workflow?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
Qi77Qi: > code to create the bucket on startup if it does not exist.

I wasn't sure if I should do that. What's the benefit of creating bucket in app?
For setup, we need to 1. create the bucket, 2. create a pubsub topic for this bucket and create notification. Are we doing all the setup in app?
I used to do things like this in terraform

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
Qi77Qi: copy paste error

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
dvoet: We should do as much setup in app as possible. This makes deployment much easier given all of our environments, fiab, etc. How we use terraform is not particularly suited this this at the moment specifically since it excludes fiab. But also it is not tied to our release process.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
dvoet: I mean the condition `metadata.fields.keys.toList.contains("parentWorkflowId")` will never evaluate to true

dvoet: the only workflows that rawls monitors are the top level workflows, I don't see how we would hit this code path with a sub workflow

But besides that, in `callLevelMetadataForCostCalculation` `parentWorkflowId` is not one of the keys requested. So even it were a sub workflow would `parentWorkflowId` exist?

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ShardedHttpExecutionServiceCluster.scala
```diff
@@ -148,6 +148,29 @@ class ShardedHttpExecutionServiceCluster (readMembers: Set[ClusterMember], submi
...
+      "runtimeAttributes",
+      "jobId",
+      "preemptible",
+      "callCaching:hit",
+      "jes",
+      "papi2",
+      "executionStatus",
+      "backend"
+    ), Set.empty, true)
+
```
ansingh7115, line 168: I think this needs "attempt" for each call as well (the number of times we've attempted to run)

ansingh7115: Same for the associated Hamm PR

ansingh7115 (approved):

Qi77Qi:
jenkins retest

PR #1039, opened 2019-02-15
Automation test for PAPIv2

marctalbott:
[GAWB-4037](https://broadinstitute.atlassian.net/browse/GAWB-4037)

This includes the same changes as https://github.com/broadinstitute/rawls/pull/1047. The difference between the two PRs is that this one has an automation test. This automation test is reliant on the default backend being "PAPIv2". I separated this out into two PRs so that #1047 can be merged before the config is changed.

EDIT: The automation test doesn't actually rely on the default backend being PAPIv2. It simply checks that workflows will work if their docker image is a private one. This should be true for both PAPIv1 and PAPIv2, so this test can go in whenever, and it should pass both before and after the default backend config change occurs

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
@@ -16,6 +16,7 @@ import org.broadinstitute.dsde.rawls.util.{FutureSupport, MethodWiths, addJitter
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport, util}
 import akka.http.scaladsl.model.StatusCodes
 import com.typesafe.config.ConfigFactory
+import org.broadinstitute.dsde.rawls.model.CromwellBackends.CromwellBackend
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import spray.json.DefaultJsonProtocol._
@@ -44,7 +45,7 @@ object WorkflowSubmissionActor {
             requesterPaysRole: String,
             useWorkflowCollectionField: Boolean,
             useWorkflowCollectionLabel: Boolean,
-            defaultBackend: String): Props = {
+            defaultBackend: CromwellBackend): Props = {
     Props(new WorkflowSubmissionActor(dataSource, methodRepoDAO, googleServicesDAO, samDAO, dosResolver, executionServiceCluster, batchSize, credential, processInterval, pollInterval, maxActiveWorkflowsTotal, maxActiveWorkflowsPerUser, runtimeOptions, trackDetailedSubmissionMetrics, workbenchMetricBaseName, requesterPaysRole, useWorkflowCollectionField, useWorkflowCollectionLabel, defaultBackend))
   }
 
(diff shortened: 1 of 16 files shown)


dvoet (approved):
eventually we should replace dockerhub private image mtalbott/mtalbott-papi-v2 but I don't think it should hold up this PR

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
```diff
@@ -58,7 +58,8 @@ class WorkflowSubmissionSpec(_system: ActorSystem) extends TestKit(_system) with
     override val workbenchMetricBaseName: String = "test",
     val requesterPaysRole: String = requesterPaysRole,
     val useWorkflowCollectionField: Boolean = false,
-    val useWorkflowCollectionLabel: Boolean = false) extends WorkflowSubmission {
+    val useWorkflowCollectionLabel: Boolean = false,
+    val defaultBackend: String = "PAPIv2") extends WorkflowSubmission {
```
Qi77Qi: `String` is very easy to have typo and compiler won't help much to ensure it's correct..Would be nice if you can make defaultBackend an ADT, for instance
```
sealed abstract class CromwellBackend{
    def asString: String
}
object CromwellBackend {
   final case class JES extends CromwellBackend {def asString: String = "JES"}
   final case class PapiV2  extends CromwellBackend {def asString: String = "PAPIv2"}
}
```

This way, you won't need to write `PAPIv2` everywhere and worrying about at some place you might have `Papiv2`.

What's even better is moving this to workbench-libs model module, cuz I'm sure other services can benefit from it.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -73,7 +74,8 @@ class WorkflowSubmissionActor(val dataSource: SlickDataSource,
                               override val workbenchMetricBaseName: String,
                               val requesterPaysRole: String,
                               val useWorkflowCollectionField: Boolean,
-                              val useWorkflowCollectionLabel: Boolean) extends Actor with WorkflowSubmission with LazyLogging {
+                              val useWorkflowCollectionLabel: Boolean,
+                              val defaultBackend: String) extends Actor with WorkflowSubmission with LazyLogging {
```
Qi77Qi: Not directly related to your PR. It'd be nice at some point we can have a `WorkflowSubmissionActorConfig` case class so that we don't have to modify the constructor every time.

Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -505,4 +508,26 @@ object WorkflowFailureModes {
   case object NoNewCalls extends WorkflowFailureMode
 }
 
+object CromwellBackends {
+  val allCromwellBackends: Seq[CromwellBackend] = List(JES, PAPIv2)
```
Qi77Qi: I would make it a map, such as `val stringToComwellBackend = Map(JES.asString => JES, PAPIv2.asString => PAPIv2)`,  this way, when you need to get `ComwellBackend` from string, you can just `stringToComwellBackend.get("somestring")`, and this gives you an `Option`, so I would have define `withName` like this:
```
def withName(name: String): Option[CromwellBackend] = stringToComwellBackend.get(name.toLowerCase())
```
And leave whether or not throwing exception up to caller. This way, your `withNameOpt` can be `def withNameOpt(nameOpt: Option[String]): Option[CromwellBackend] = nameOpt.flatMap(name => withName(name))`

PR #1041, opened 2019-02-20
GAWB-4094 DM in FC

helgridly:
Fc-dev friend: https://github.com/broadinstitute/firecloud-develop/pull/1611

This closes CA-6 and CA-220.

# Extremely important note

When merging this, add the following info to the release ticket:

This release changes how billing projects are created. Any billing projects that are _being_ created at the moment the release drops will fail, so it's important to first **ensure no billing projects are being created before deploying**. In order to do this, run the following command:

`docker run -it --rm -v ~:/root broadinstitute/dsde-toolbox:dev mysql-connect.sh -p firecloud -e prod -a rawls "select count(*) from BILLING_PROJECT where CREATION_STATUS NOT IN ('Ready', 'Error')"`

If it does not show a count of 0, **do not deploy to production** -- wait 5 minutes and try again.


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -329,15 +335,21 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
               }
             }
 
-            createProjectOperation <- gcsDAO.createProject(projectName, billingAccount).recoverWith {
+            //NOTE: we're syncing this to Sam ahead of the resource actually existing. is this fine? (ps these are sam calls)
```
dvoet, line 346: the resource exists in sam but the project has not been created in google yet... should be fine

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
```diff
@@ -19,155 +22,87 @@ import scala.util.Success
...
-
-
-      _ <- datasource.inTransaction { dataAccess =>
-        // save project updates
-        val updatedProjects = maybeUpdatedProjects.toSet -- projects.toSet
-        dataAccess.rawlsBillingProjectQuery.updateBillingProjects(updatedProjects)
-      }
+  def checkPubSub()(implicit executionContext: ExecutionContext): Future[CreatingBillingProjectMonitorMessage] = {
+    for {
+      psMessages <- pubSubDAO.pullMessages(dmPubSubSubscription, 100)
```
dvoet: I don't care enough to warrant changing anything but since project create is relatively infrequent, it seems like doing them 100 at a time instead of 1 at a time is an over complication. In reality they are likely to come in 1 at a time.

dvoet (commented):
generally looks like a fine approach

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
helgridly: yeah i went for a big number assuming that lots of rawlses in fiab would potentially be sharing the same topic, and might have to churn through a lot of messages (esp. in the case of auto-tests)

but if we're putting the hostname in the topic then it doesn't matter so much

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -83,6 +84,11 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def AdminAddLibraryCurator(userEmail: RawlsUserEmail) = asFCAdmin { addLibraryCurator(userEmail) }
   def AdminRemoveLibraryCurator(userEmail: RawlsUserEmail) = asFCAdmin { removeLibraryCurator(userEmail) }
 
+  val dmPubSubTopic = dmConfig.getString("pubSubTopic")
```
Qi77Qi: I would create a `UserServiceConfig` case class, and initialize it in `boot`...if anything's wrong with config, we want to know sooner than later. It's also easier to write a single config test for all configurations required by the app that way. Something like https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala#L61

helgridly:
successful auto-tests https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/19000/

helgridly:
@dvoet @marctalbott can i get a final review, please

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -114,13 +114,30 @@ object Boot extends IOApp with LazyLogging {
...
+    }
+    val badPathBecauseBranchName = dmTemplatePath.contains("githubusercontent.com") &&
+      (dmTemplatePath.contains("/blob/master") || dmTemplatePath.contains("/blob/develop"))
+    if(badPathBecauseBranchName) {
+      logger.error("dmConfig.templatePath refers to a branch of a GitHub repo. This makes it impossible to know which template was used to create a project. Please use a specific commit instead.")
+    }
+    if(badPathBecauseGithub || badPathBecauseBranchName) {
+      sys.exit(1)
+    }
+
```
dvoet, line 127: first, pull this "Sanity check deployment manager template path" into a function with a name that reflects this comment.

second, throw an exception instead of sys.exit... sys.exit is a bad practice because it is unexpected.

dvoet:
is there a runtime dependency on github? What happens if github is down?

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -129,7 +146,7 @@ object Boot extends IOApp with LazyLogging {
         gcsConfig.getString("subEmail"),
         gcsConfig.getString("pathToPem"),
         gcsConfig.getString("appsDomain"),
-        gcsConfig.getString("groupsPrefix"),
+        dmConfig.getLong("orgID"),gcsConfig.getString("groupsPrefix"),
```
dvoet: new line after comma

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -138,13 +155,14 @@ object Boot extends IOApp with LazyLogging {
         gcsConfig.getString("billingPemEmail"),
         gcsConfig.getString("pathToBillingPem"),
         gcsConfig.getString("billingEmail"),
-        gcsConfig.getInt("bucketLogsMaxAge"),
+        gcsConfig.getString("billingGroupEmail"),gcsConfig.getInt("bucketLogsMaxAge"),
```
dvoet: new line after comma

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -218,5 +200,13 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
 
 case class GoogleWorkspaceInfo(bucketName: String, policyGroupsByAccessLevel: Map[WorkspaceAccessLevel, WorkbenchEmail])
-case class ProjectTemplate(policies: Map[String, Seq[String]], services: Seq[String])
-final case class HammCromwellMetadata(bucketName: GcsBucketName, topicName: ProjectTopicName)
\ No newline at end of file
+case class ProjectTemplate(policies: Map[String, Seq[String]])
+final case class HammCromwellMetadata(bucketName: GcsBucketName, topicName: ProjectTopicName)
+
+case object ProjectTemplate {
+  def from(projectTemplateConfig: Config, requesterPaysRole: String): ProjectTemplate = {
```
dvoet: this is only used in test code so probably should move there. Also `requesterPaysRole` is unused

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -538,23 +571,16 @@ class HttpGoogleServicesDAO(
...
 
-  /**
-    * NOTE: This function will returns "false" in both of the following cases:
-    * if you don't have sufficient scopes
-    *   - Google's JSON response body will contain "message": "Request had insufficient authentication scopes."
-    * if you're not authorized to see the billing account
-    *   - Google's JSON response body will contain "message" : "The caller does not have permission"
-    */
-  protected def credentialOwnsBillingAccount(credential: Credential, billingAccountName: String): Future[Boolean] = {
+  protected def billingAccountContainsFCBillingGroup(credential: Credential, billingAccountId: String): Future[Boolean] = {
```
dvoet: this function tests very specifically that there exists a policy with `billingGroupEmail` having `roles/billing.user`. It will not work if the email is granted the appropriate permissions in any other way: roles/billing.admin, a custom role, a more complicated group hierarchy.

Can we use https://cloud.google.com/billing/reference/rest/v1/billingAccounts/testIamPermissions instead to test that the service account we care about has appropriate permissions? This would mean pulling in a key for that SA because we need to auth as that identity.

helgridly:
@dvoet it'll fail to create the project and give an error

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -46,6 +46,10 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
     }
   }
 
+  protected def executeGoogleRqWithRetry[T](request: AbstractGoogleClientRequest[T])(implicit counters: GoogleCounters, histo: Histogram): Future[T] = {
```
dvoet: spell out request

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -705,28 +728,88 @@ class HttpGoogleServicesDAO(
...
-  override def createProject(projectName: RawlsBillingProjectName, billingAccount: RawlsBillingAccount): Future[RawlsBillingProjectOperationRecord] = {
-    implicit val service = GoogleInstrumentedService.Billing
-    val credential = getBillingServiceAccountCredential
+  def getDMConfigString(projectName: RawlsBillingProjectName, dmTemplatePath: String, properties: Map[String, JsValue]): String = {
+    import DeploymentManagerJsonSupport._
+    import cats.syntax.either._
+    import io.circe.yaml._
+    import io.circe.yaml.syntax._
+
+    val cc = ConfigContents(Seq(Resources(projectName.value, dmTemplatePath, properties)))
```
dvoet: better name than `cc`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -705,28 +728,88 @@ class HttpGoogleServicesDAO(
     })
   }
 
-  override def createProject(projectName: RawlsBillingProjectName, billingAccount: RawlsBillingAccount): Future[RawlsBillingProjectOperationRecord] = {
-    implicit val service = GoogleInstrumentedService.Billing
-    val credential = getBillingServiceAccountCredential
+  def getDMConfigString(projectName: RawlsBillingProjectName, dmTemplatePath: String, properties: Map[String, JsValue]): String = {
```
dvoet: name it `getDMConfigYamlString`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -218,5 +200,13 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 }
 
 case class GoogleWorkspaceInfo(bucketName: String, policyGroupsByAccessLevel: Map[WorkspaceAccessLevel, WorkbenchEmail])
-case class ProjectTemplate(policies: Map[String, Seq[String]], services: Seq[String])
-final case class HammCromwellMetadata(bucketName: GcsBucketName, topicName: ProjectTopicName)
\ No newline at end of file
+case class ProjectTemplate(policies: Map[String, Seq[String]])
```
dvoet: why is this still a map of policies?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -705,28 +728,88 @@ class HttpGoogleServicesDAO(
...
+      "projectViewersGroup" -> computeUserGroupEmail.value.toJson,
+      "requesterPaysRole" -> requesterPaysRole.toJson,
+      "highSecurityNetwork" -> false.toJson,
+      "fcProjectOwners" -> projectTemplate.policies("roles/owner").toJson,
+      "fcProjectEditors" -> projectTemplate.policies("roles/editor").toJson,
+      "labels" -> templateLabels
+    )
+
+    //config is a list of one resource: type=composite-type, name=whocares, properties=pokein
+    val confy = new ConfigFile().setContent(getDMConfigString(projectName, dmTemplatePath, properties))
```
dvoet: `config`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -705,28 +728,88 @@ class HttpGoogleServicesDAO(
...
+    val dconf = new TargetConfiguration().setConfig(confy)
```
dvoet: what does `dconf` mean?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -749,6 +832,16 @@ class HttpGoogleServicesDAO(
         }).map { op =>
           rawlsBillingProjectOperation.copy(done = toScalaBool(op.getDone), errorMessage = Option(op.getError).map(error => toErrorMessage(error.getMessage, error.getCode)))
         }
+
```
dvoet: should we get rid of the above cases?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
```diff
@@ -90,63 +90,30 @@ trait CreatingBillingProjectMonitor extends LazyLogging {
...
-          case operations: Seq[RawlsBillingProjectOperationRecord] if operations.forall(rec => rec.done) =>
-            // all operations completed but some failed
-            val messages = operations.collect {
-              case RawlsBillingProjectOperationRecord(_, operationName, _, true, Some(error), _) => s"Failure enabling api $operationName: ${error}"
-            }
-            val errorMessage = messages.mkString("[", "], [", "]")
-            logger.debug(s"errors enabling apis for project ${project.projectName}: $errorMessage")
-            Future.successful(project.copy(status = CreationStatuses.Error, message = Option(errorMessage)))
+          case Seq(RawlsBillingProjectOperationRecord(_, gcsDAO.CREATE_PROJECT_OPERATION, _, _, _, _)) =>
+            Future.successful(project.copy(status = CreationStatuses.Error, message = Option(s"failure processing new project ${project.projectName.value}: old-style project creation no longer supported")))
```
dvoet: what happens to any project creations in flight? Do we just ensure that there aren't any?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
```diff
@@ -90,63 +90,30 @@ trait CreatingBillingProjectMonitor extends LazyLogging {
...
+
+          case bp :: bp2 :: bps =>
```
dvoet: I don't understand this case

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -83,6 +84,9 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def AdminAddLibraryCurator(userEmail: RawlsUserEmail) = asFCAdmin { addLibraryCurator(userEmail) }
   def AdminRemoveLibraryCurator(userEmail: RawlsUserEmail) = asFCAdmin { removeLibraryCurator(userEmail) }
 
+  val dmTemplatePath = dmConfig.getString("templatePath")
+  val dmProject = dmConfig.getString("projectID")
```
dvoet: we should not be parsing config here. Instead of passing in a `Config` object in the constructor pass in a case class.

dvoet (changes requested):
is there anything new we need to test?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
helgridly: Yes.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
helgridly: This catches the cases I got rid of, where there's more than one in-flight operation for a project. This should only happen if there are projects creating the old way, and at the enable-APIs step, at the time we do this deploy and it switches over

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
helgridly: what's the country of origin? can you use it in a sentence?

:bee:

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
helgridly: DM still wants a list of owners and a list of editors, and I figured keeping them in conf would be better than putting them in the DM template directly. If you like, I can make this `case class ProjectTemplate(owners: Seq[String], editors: Seq[String])`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
helgridly: it seems to be used in `WorkflowSubmissionActor`

helgridly: or do you mean it's unused in the `from` def

helgridly: yes on removing the unused var. it's not obvious to me where in test code it'd go -- might as well keep it near its class definition?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: in the case where there are leftover project creations when we deploy this, i want the error to be caught by [these cases](https://github.com/broadinstitute/rawls/pull/1041/files#diff-5fc8bf1e934648fb8daf9dce39334029R112). so this code should probably return a successful future with an error message or something

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
dvoet: yes please... I agree we need it in config but a map is too free form given DM names those parameters

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: but as in the next comment, we should ensure that there are no project creations in flight when we deploy this new code. I think we have 2 options
1) [easiest] have our code support only the new way and destroy all evidence of the old way
2) [more robust] first release supports both simultaneously but all new project creations use the new way then a follow on release that removes the old way

I don't think there should be any middle ground that recognizes the old way and errors.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CreatingBillingProjectMonitor.scala
dvoet: see https://github.com/broadinstitute/rawls/pull/1041#discussion_r289855660

helgridly:
jenkins retest

dvoet (approved):
if you want to make a jira for the check billing account permissions thing, fine

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: So the `projectnumber@cloudservices.gserviceaccount.com` SA isn't one we can get keys for in order to auth as. I think our options here are:

1. Test for one of `terra-billing@terra.bio` and `terra-billing@firecloud.org` being in the bindings, and increase their scope to "at least BA User or higher"
2. Put the Rawls SA in the group and see if _it_ has permissions (but that's back to the Rawls SA being mega-powerful again)
3. I don't know if Google has an endpoint similar to `testIamPermissions` but for a different SA. I couldn't find one.

helgridly: IMO we want to test whether the user has done what we ask of them. If they somehow found out the `projectnumber@cloudservices` address and added that, then DM would _work_ but wouldn't be resilient to any changes we'd make further down the line. I take your point about more complicated group hierarchies, though.

helgridly: I did a bit more research. A possible new option:

4. We might be able to generate a [short-lived service account credential](https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials) for the `projectnumber@cloudservices` such that Rawls can temporarily become that SA and test IAM permissions.

helgridly: I've bugged this as [CA-314](https://broadworkbench.atlassian.net/browse/CA-314). We're going to have to solve this before release because Lukas is giving terra-billing@ permissions at the Broad org-level, which doesn't show in the BA IAM policy and will thus incorrectly fail this check.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: done

helgridly:
successful auto-test run https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/20873/

PR #1070, opened 2019-04-17
AuthDomains api test

aweng98:
Replace Firecloud-UI AuthDomainGroupSpec and AuthDomainMultiGroupSpec.

[QA-492](https://broadworkbench.atlassian.net/browse/QA-492) and  [QA-491](https://broadworkbench.atlassian.net/browse/QA-491)

depended on [workbench-lib PR 221](https://github.com/broadinstitute/workbench-libs/pull/221)

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
@@ -21,11 +21,9 @@ class AuthDomainGroupApiSpec extends FreeSpec with Matchers with WorkspaceFixtur
 
   val defaultUser: Credentials = UserPool.chooseCurator
   val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
-
-  // One Auth-Domain group tests
-
+  
   "A workspace" - {
-    "with one group inside of it" - {
+    "with one group in its auth domain" - {
 
       "can be created" in {
         val user = UserPool.chooseAuthDomainUser
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
@@ -1,0 +1,158 @@
+package org.broadinstitute.dsde.test.api
+
+import org.broadinstitute.dsde.test.util.AuthDomainMatcher
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
+import org.broadinstitute.dsde.workbench.fixture.{BillingFixtures, GroupFixtures, WorkspaceFixtures}
+import org.broadinstitute.dsde.workbench.service.util.Tags
+import org.broadinstitute.dsde.workbench.service.{AclEntry, Orchestration, Rawls, WorkspaceAccessLevel}
+import org.scalatest.{FreeSpec, Matchers}
+
+import scala.util.Try
+
+
+class AuthDomainGroupApiSpec extends FreeSpec with Matchers with WorkspaceFixtures with BillingFixtures with GroupFixtures {
+
+  /*
+  * Unless otherwise declared, this auth token will be used for API calls.
+  * We are using a curator to prevent collisions with users in tests (who are Students and AuthDomainUsers), not
+  *  because we specifically need a curator.
+  */
+
+  val defaultUser: Credentials = UserPool.chooseCurator
+  val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
+
+  // One Auth-Domain group tests
+
+  "A workspace" - {
+    "with one group inside of it" - {
+
+      "can be created" in {
+        val user = UserPool.chooseAuthDomainUser
+        implicit val authToken: AuthToken = authTokenDefault
+
+        withGroup("authDomain", List(user.email)) { authDomainName =>
+          withCleanBillingProject(user) { projectName =>
+            withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName)) { workspaceName =>
+
+              // user is one of the authdomain group members
+              groupNameToMembersEmails(authDomainName) should contain (user.email)
+              // user can access workspace and see the authdomain group
+              AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(authDomainName))(user.makeAuthToken())
+
+            }(user.makeAuthToken())
+          }
+        }
+      }
+
+      "can be cloned and retain the auth domain" taggedAs Tags.SmokeTest in {
+        val user = UserPool.chooseAuthDomainUser
+        implicit val authToken: AuthToken = authTokenDefault
+
+        withGroup("authDomain", List(user.email)) { authDomainName =>
+          withCleanBillingProject(defaultUser, userEmails = List(user.email)) { projectName =>
+            withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName),
+              List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+
+              val workspaceCloneName = workspaceName + "_clone"
+              register cleanUp Try(Rawls.workspaces.delete(projectName, workspaceCloneName)(user.makeAuthToken())).recover({
+                case _: Exception =>
+              })
+              // Note: this is not a passthrough to Rawls is because library needs to overwrite any publish and discoverableByGroups values
+              Orchestration.workspaces.clone(projectName, workspaceName, projectName, workspaceCloneName, Set(authDomainName))(user.makeAuthToken())
+
+              // the authdomain group should be found in cloned workspace
+              val groups = Rawls.workspaces.getAuthDomainsInWorkspace(projectName, workspaceCloneName)(user.makeAuthToken())
+              groups should contain theSameElementsAs List(authDomainName)
+
+              Rawls.workspaces.delete(projectName, workspaceCloneName)(user.makeAuthToken())
+            }
+          }
+        }
+      }
+
+
+      "when the user is not inside of the group" - {
+
+        "when the workspace is shared with them" - {
+          "can be seen but is not accessible" in {
+
+            val user = UserPool.chooseStudent
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("authDomain") { authDomainName =>
+              withCleanBillingProject(defaultUser) { projectName =>
+                withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName),
+                  List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+
+                  // user can see workspace but cannot access workspace
+                  AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+                }
+              }
+            }
+          }
+        }
+
+        "when the workspace is not shared with them" - {
+          "cannot be seen and is not accessible" in {
+
+            val user = UserPool.chooseStudent
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("authDomain") { authDomainName =>
+              withCleanBillingProject(defaultUser) { projectName =>
+                withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName)) { workspaceName =>
+
+                  // user cannot see workspace and user cannot access workspace
+                  AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+                }
+              }
+            }
+          }
+        }
+      }
+
+      "when the user is inside of the group" - {
+
+        "when the workspace is shared with them" - {
+          "can be seen and is accessible" in {
+
+            val user = UserPool.chooseAuthDomainUser
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("authDomain", List(user.email)) { authDomainName =>
+              withCleanBillingProject(defaultUser) { projectName =>
+                withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName),
+                  List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+
+                  // user can see workspace and user can access workspace
+                  AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(authDomainName))(user.makeAuthToken())
+                }
+              }
+            }
+          }
+        }
+
+        "when the workspace is not shared with them" - {
+          "cannot be seen and is not accessible" in {
+
+            val user = UserPool.chooseAuthDomainUser
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("AuthDomain", List(user.email)) { authDomainName =>
+              withCleanBillingProject(defaultUser) { projectName =>
+                withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName)) { workspaceName =>
+
+                  // user cannot see workspace and user cannot access workspace
+                  AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+                }
+              }
+            }
+          }
+        }
+      }
+
+    }
+  }
+
+}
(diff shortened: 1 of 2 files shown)


automation/project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ object Dependencies {
   val akkaV = "2.5.7"
   val akkaHttpV = "10.0.10"
 
-  val serviceTestV = "0.16-e73039a"
+  val serviceTestV = "0.16-a753e32-SNAP"
```
aweng98: It is the local published snapshot. test won't work in Jenkin until merge [workbench-lib #220](https://github.com/broadinstitute/workbench-libs/pull/220)

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
+import org.broadinstitute.dsde.workbench.fixture.{BillingFixtures, GroupFixtures, WorkspaceFixtures}
+import org.broadinstitute.dsde.workbench.service.util.Tags
+import org.broadinstitute.dsde.workbench.service.{AclEntry, Orchestration, Rawls, RestException, WorkspaceAccessLevel}
+import org.scalatest.{FreeSpec, Matchers}
+
+import scala.util.Try
+
+class AuthDomainGroupApiSpec extends FreeSpec with Matchers with WorkspaceFixtures with BillingFixtures with GroupFixtures {
```
aweng98: the test name is same as firecloud-ui test.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,417 @@
...
+
+        withGroup("AuthDomainOne", List(user.email)) { groupOne =>
+          withGroup("AuthDomainTwo", List(user.email)) { groupTwo =>
+
+            withCleanBillingProject(user) { projectName =>
+              withWorkspace(projectName, "GroupsApiSpec_workspace", Set(groupOne, groupTwo)) { workspaceName =>
+
+                // user is in members emails in two authdomain groups
+                groupNameToMembersEmails(groupOne) should contain allElementsOf List(user.email)
+                groupNameToMembersEmails(groupTwo) should contain allElementsOf List(user.email)
```
ansingh7115: for these you should just be able to do something like `... should contain (user.email)` instead of checking all elements of a 1 element list.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+      "can be created" in {
+        val user = UserPool.chooseAuthDomainUser
+        implicit val authToken: AuthToken = authTokenDefault
+
+        withGroup("authDomain", List(user.email)) { authDomainName =>
+          withCleanBillingProject(user) { projectName =>
+            withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName)) { workspaceName =>
+
+              // user is one of the authdomain group members
+              groupNameToMembersEmails(authDomainName) should contain allElementsOf List(user.email)
```
ansingh7115: here as well

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+          "cannot be seen and is not accessible" in {
+
+            val user = UserPool.chooseStudent
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("authDomain") { authDomainName =>
+              withCleanBillingProject(defaultUser) { projectName =>
+                withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName)) { workspaceName =>
+
+                  // user cannot see workspace but user cannot access workspace
```
aweng98: word `but` should be `and` in comment

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: fixed

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+          withCleanBillingProject(defaultUser, userEmails = List(user.email)) { projectName =>
+            withWorkspace(projectName, "AuthDomainGroupApiSpec_workspace", Set(authDomainName),
+              List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+
+              val workspaceCloneName = workspaceName + "_clone"
+              register cleanUp Try(Rawls.workspaces.delete(projectName, workspaceCloneName)(user.makeAuthToken())).recover({
+                case _: Exception =>
+              })
+              // Note: this is not a passthrough to Rawls is because library needs to overwrite any publish and discoverableByGroups values
+              Orchestration.workspaces.clone(projectName, workspaceName, projectName, workspaceCloneName, Set(authDomainName))(user.makeAuthToken())
```
marctalbott, line 60: How important is it that library overwrites the `publish` and `discoverableByGroups` values? Terra calls the Rawls clone endpoint directly to clone workspaces and I think this test should try to clone as similarly to how Terra does it as possible. If you decide to use the Rawls clone endpoint here, also change to the Rawls endpoint in the other two cloning tests in the multi group spec

aweng98:
jenkins retest

aweng98:
fiab-start failed.
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+          }
+        }
+      }
+
+      "can be cloned and retain the auth domain" taggedAs Tags.SmokeTest in {
+        val user = UserPool.chooseAuthDomainUser
+        implicit val authToken: AuthToken = authTokenDefault
+
+        withGroup("authDomain", List(user.email)) { authDomainName =>
+          withCleanBillingProject(defaultUser, userEmails = List(user.email)) { projectName =>
```
marctalbott, line 51: Why does `user` need to be a billing project user here, but not in the next test?

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+  *  because we specifically need a curator.
+  */
+
+  val defaultUser: Credentials = UserPool.chooseCurator
+  val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
+
+  // One Auth-Domain group tests
+
+  "A workspace" - {
+    "with one group inside of it" - {
```
marctalbott: I'd change the name here from "with one group inside of it" to "with one group in its auth domain" just so it's clear that the "it" here refers to the workspace's auth domain

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+   * We are using a curator to prevent collisions with users in tests (who are Students and AuthDomainUsers), not
+   *  because we specifically need a curator.
+   */
+
+  val defaultUser: Credentials = UserPool.chooseCurator
+  val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
+
+
+  "A workspace" - {
+    "with multiple groups inside of it" - {
```
marctalbott: Same comment here as the single group suite, can you rename this to something along the lines of "with multiple groups in its auth domain" just so the test name is a bit more clear

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+        "when the user is a billing project owner" - {
+          "can be seen but is not accessible" in {
+
+            val user = UserPool.chooseProjectOwner
+            implicit val authToken: AuthToken = authTokenDefault
+
+            withGroup("AuthDomainOne") { groupOne =>
+              withGroup("AuthDomainTwo") { groupTwo =>
+
+                withCleanBillingProject(user, userEmails = List(defaultUser.email)) { projectName =>
```
marctalbott, line 160: why is `defaultUser` included in the `userEmail` list here? I know it's needed so that `defaultUser`'s auth token can be used to create the workspace, but I'd rather see the implicit auth token just be `user`'s auth token for this test rather than use the default.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+                    // user can see workspace but user cannot access workspace
+                    AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+                  }
+                }
+              }
+            }
+          }
+
+          "when the user is a billing project owner" - {
+            "can be seen but is not accessible" in {
```
marctalbott, line 217: this test should include two auth domains, and `user` should be in one of them

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+                    // user can see workspace and user can access workspace
+                    AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupOne, groupTwo))(user.makeAuthToken())
+                  }
+                }
+              }
+            }
+          }
+
+          "and given writer access" - {
+            "the user has correct permissions" in {
```
marctalbott, line 284: I feel like this test should also check that the workspace is visible and accessible for the writer, not just that the workspace access level is writer

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
```diff
@@ -0,0 +1,418 @@
...
+                    AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+                  }
+                }
+              }
+            }
+          }
+        }
+
+        //TCGA controlled access workspaces use-case
+        "when the workspace is shared with the group" - {
```
marctalbott: This test is identical to the test above "a workspace with multiple groups inside of it when the user is in all of the groups when shared with one of the groups in the auth domain can be seen and is accessible by group member who is a member of both auth domain groups" ([line 326](https://github.com/broadinstitute/rawls/blob/4868778fc9e8a5d9300f4a25596c6a1945ae286c/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala#L326))

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+class AuthDomainGroupApiSpec extends FreeSpec with Matchers with WorkspaceFixtures with BillingFixtures with GroupFixtures {
+
+  /*
+  * Unless otherwise declared, this auth token will be used for API calls.
+  * We are using a curator to prevent collisions with users in tests (who are Students and AuthDomainUsers), not
+  *  because we specifically need a curator.
+  */
+
+  val defaultUser: Credentials = UserPool.chooseCurator
+  val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
```
ansingh7115, line 23: defaultUser and authTokenDefault should be removed, they're not being used anywhere

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
aweng98: @marctalbott  I think that's a business logic that I'm hesitating to change. I'd like to keep it behave in same way I found in firecloud-ui. Do you feel strongly about change to use Rawls clone?

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
aweng98: in this test, if `user` is not a owner of the billing project, it cannot create workspace. in next test, `authTokenDefault` is used to create workspace so the `user` doesn't have to a owner of billing project. Does that answer the question?

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
aweng98: they're used. line 81 for example.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
marctalbott: Hmm, yeah I'm not sure why I was comparing this test to the next test, this looks fine

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: bc `user` is a billing project owner, workspace wasn't created with its authToken so `user` can only see workspace in Workspaces list but not accessible. Change to `user` authToken changes the test.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
marctalbott: Fair point. @ansingh7115 thoughts? I know we want our new tests to reflect what Terra is actually doing (calling Rawls directly in this case), but we also want to keep the existing coverage and I suppose switching to the Rawls endpoint could potentially reduce coverage?

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
aweng98: yes, duplicated. I removed this test.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
marctalbott: My understanding is that a billing project owner is also granted owner permissions on workspaces created in the billing project. So `user` would have owner permissions on this workspace whether its auth token is used to create it or not. That being said, I'm not 100% sure, and this is a pretty minor change so I say ignore this comment

marctalbott (approved):

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupApiSpec.scala
ansingh7115: ah! You're right, I think I was thinking of a different PR. Sorry about that :)

ansingh7115 (approved):

PR #1074, opened 2019-04-22
AuthDomainGroupRoleSpec

aweng98:
Jira [QA-493](https://broadworkbench.atlassian.net/browse/QA-493)
depend on [workbench-lib PR 221](https://github.com/broadinstitute/workbench-libs/pull/221)

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -7,9 +7,9 @@ object Dependencies {
   val akkaV = "2.5.7"
   val akkaHttpV = "10.0.10"
 
-  val serviceTestV = "0.16-c44b1d3"
-  val workbenchGoogleV = "0.18-85787ba"
-  val workbenchModelV  = "0.10-52d614b"
+  val serviceTestV = "0.16-f63a6eb"
+  val workbenchGoogleV = "0.18-5941525"
+  val workbenchModelV  = "0.13-4c7acd5"
   val workbenchMetricsV  = "0.3-7ad0aa8"
 
   val workbenchModel: ModuleID = "org.broadinstitute.dsde.workbench" %% "workbench-model" % workbenchModelV
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
@@ -1,0 +1,220 @@
+package org.broadinstitute.dsde.test.api
+
+import org.broadinstitute.dsde.test.util.AuthDomainMatcher
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.config.UserPool
+import org.broadinstitute.dsde.workbench.fixture.{BillingFixtures, GroupFixtures, WorkspaceFixtures}
+import org.broadinstitute.dsde.workbench.service.BillingProject.BillingProjectRole
+import org.broadinstitute.dsde.workbench.service.Orchestration.groups.GroupRole
+import org.broadinstitute.dsde.workbench.service.{Rawls, RestException, Sam}
+import org.scalatest.{FreeSpec, Matchers}
+
+import scala.util.Try
+
+class AuthDomainGroupRoleSpec extends FreeSpec with WorkspaceFixtures with BillingFixtures with GroupFixtures with Matchers {
+
+  "removing permissions from workspace with auth domain" - {
+
+    "+ project owner, + group member, create workspace, - group member" in {
+
+      // a billing project owner and a member of the authdomain group
+      val billingProjectOwner = UserPool.chooseProjectOwner
+
+      // a billing project owner, a authdomain group owner and workspace owner
+      val student = UserPool.chooseStudent
+      implicit val studentToken: AuthToken = student.makeAuthToken()
+
+      withCleanBillingProject(billingProjectOwner, List(student.email)) { projectName =>
+        withGroup("group", List(billingProjectOwner.email)) { groupName =>
+          withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+            AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupName))(billingProjectOwner.makeAuthToken())
+
+            // remove a group member "billingProjectOwner" from group
+            Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, billingProjectOwner.email)
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(billingProjectOwner.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, + group member, create workspace, - project owner" in {
+
+      // a billing project owner, a authdomain group owner and workspace owner
+      val billingProjectOwner = UserPool.chooseProjectOwner
+      implicit val token: AuthToken = billingProjectOwner.makeAuthToken()
+
+      // a billing project owner and a member of authdomain group
+      val student = UserPool.chooseStudent
+
+      withCleanBillingProject(billingProjectOwner, ownerEmails = List(student.email)) { projectName =>
+        withGroup("group", List(student.email)) { groupName =>
+          withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+            AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupName))(student.makeAuthToken())
+
+            // remove "student" from billing project
+            Rawls.billing.removeUserFromBillingProject(projectName, student.email, BillingProjectRole.Owner)
+            AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, create workspace, + group member, - group member" in {
+
+      // a billing project owner and a member of the authdomain group
+      val billingProjectOwner = UserPool.chooseProjectOwner
+
+      // a billing project owner, a authdomain group owner and workspace owner
+      val student = UserPool.chooseStudent
+      implicit val token: AuthToken = student.makeAuthToken()
+
+      withCleanBillingProject(billingProjectOwner, List(student.email)) { projectName =>
+        withGroup("group") { groupName =>
+          withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(billingProjectOwner.makeAuthToken())
+
+            // add a group member "billingProjectOwner" to group
+            Sam.user.addUserToPolicy(groupName, GroupRole.Member.toString, billingProjectOwner.email)
+            register cleanUp Try(Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, billingProjectOwner.email)).recover {
+              case _: RestException =>
+            }
+            AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupName))(billingProjectOwner.makeAuthToken())
+
+            // remove a group member "billingProjectOwner" from group
+            Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, billingProjectOwner.email)
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(billingProjectOwner.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, + group member, - group member" in {
+
+      // a billing project owner, a authdomain group owner and workspace owner
+      val billingProjectOwner = UserPool.chooseProjectOwner
+      implicit val token: AuthToken = billingProjectOwner.makeAuthToken()
+
+      // a billing project owner and a member of authdomain group
+      val student = UserPool.chooseStudent
+
+      withCleanBillingProject(billingProjectOwner, ownerEmails = List(student.email)) { projectName =>
+        withGroup("group") { groupName =>
+          withCleanUp {
+            withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+              AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+
+              // add a group member "student" to group
+              Sam.user.addUserToPolicy(groupName, GroupRole.Member.toString, student.email)
+              register cleanUp Try(Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, student.email)).recover {
+                case _: RestException =>
+              }
+              AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupName))(student.makeAuthToken())
+
+              // remove a group member "student" from group
+              Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, student.email)
+              AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+            }
+          }
+        }
+      }
+    }
+
+    "+ project owner, + group member, - project owner" in {
+
+      // a billing project owner, a authdomain group owner and workspace owner
+      val billingProjectOwner = UserPool.chooseProjectOwner
+      implicit val token: AuthToken = billingProjectOwner.makeAuthToken()
+
+      // a billing project owner and a member of authdomain group
+      val student = UserPool.chooseStudent
+
+      withCleanBillingProject(billingProjectOwner, ownerEmails = List(student.email)) { projectName =>
+        withGroup("group") { groupName =>
+          withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+
+            // add a group member "student" to group
+            Sam.user.addUserToPolicy(groupName, GroupRole.Member.toString, student.email)
+            register cleanUp Try(Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, student.email)).recover {
+              case _: RestException =>
+            }
+            AuthDomainMatcher.checkVisibleAndAccessible(projectName, workspaceName, List(groupName))(student.makeAuthToken())
+
+            // remove "student" from billing project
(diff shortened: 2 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+class BillingProjectOwnerPermissionSpec extends FreeSpec with WorkspaceFixtures with BillingFixtures with GroupFixtures  with Matchers {
+
+  /*
+ * Unless otherwise declared, this auth token will be used for API calls.
+ * We are using a curator to prevent collisions with users in tests (who are Students and AuthDomainUsers), not
+ *  because we specifically need a curator.
+ */
+
+  val defaultUser: Credentials = UserPool.chooseCurator
+  val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
```
ansingh7115: All of the tests in this class declare their own implicit auth token so this isn't actually being used anywhere

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+
+
+  "removing permissions from billing project owners for workspaces with auth domains" - {
+
+    "+ project owner, + group member, create workspace, - group member" in {
+      val owner = UserPool.chooseProjectOwner
+      val creator = UserPool.chooseStudent
+      val user = owner
```
ansingh7115: Shouldn't need `val user = owner` or similar assignments. Just have owner and creator. Same in other tests

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+            Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, user.email)
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, + group member, create workspace, - project owner" in {
+      val owner = UserPool.chooseProjectOwner
+      val creator = owner
```
ansingh7115: `owner` isn't actually used anywhere. Take out `val creator = owner` and just have `val creator = UserPool.chooseProjectOwner`

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      }
+    }
+
+    "+ project owner, create workspace, + group member, - group member" in {
+      val owner = UserPool.chooseProjectOwner
+      val creator = UserPool.chooseStudent
+      val user = owner
+      implicit val token: AuthToken = creator.makeAuthToken()
+
+      withCleanBillingProject(owner, List(creator.email)) { projectName =>
```
ansingh7115: I'm seeing inconsistency around naming `owner` and `creator` and `user` where some tests have the creator be the owner and others have creator as the student. I would make these naming patterns more consistent.

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+            Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, user.email)
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, create workspace, + group member, - project owner" in {
+      val owner = UserPool.chooseProjectOwner
+      val creator = owner
```
ansingh7115: same here, owner isn't used, just have `val creator = UserPool.chooseProjectOwner`, check other tests too

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+      val user = UserPool.chooseStudent
+      implicit val token: AuthToken = creator.makeAuthToken()
+
+      withCleanBillingProject(creator, ownerEmails = List(user.email)) { projectName =>
+        withGroup("group") { groupName =>
+          withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(user.makeAuthToken())
+
+            Sam.user.addUserToPolicy(groupName, GroupRole.Member.toString, user.email)
```
ansingh7115: Sam doesn't need to be called for these tests. All the add / remove users from groups should be done through the rawls api. the original fc-ui tests were calling orch to do this so it should just be a pass-through

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: @ansingh7115 Intentionally, I wrote new tests as close as possible original `AuthDomainOwnersSpec` so I don't accidentally changed original tests' logic. Just want to be sure, are we sure about making suggested change? I do think suggested changes are going to help to make tests clearer and cleaner.

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: I don't know what Rawls api to use for add/remove users from groups. Can you tell me what they are and an example?

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: a user who is a student is also the billing project creator in this test. naming like this makes sense.

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: removed re-assignment here and in other tests.

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: updated here and in other tests.

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: ✔️

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
ansingh7115: It is not consistent with other tests in this class - for example the test right above it defines the `creator` as the project owner and the `user` as the student. _This_ test has the `owner` as the project owner and the `creator` as a student. The naming should be consistent from test to test. Either use `owner` and `creator` for all the tests or use `owner` and `user` for all the tests, but not a mixture of both

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
ansingh7115: hmm, looking more into it, seems orch is just calling sam directly for this rather than going through rawls. This is fine then!

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: removed `defaultUser` and `authTokenDefault`

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
aweng98: 👍 use `owner` and `creator` for all tests

marctalbott: I still find the `owner` vs `creator` naming confusing. Seeing 
```
val creator = UserPool.chooseProjectOwner
val owner = UserPool.chooseStudent
```
and then
```
val owner = UserPool.chooseProjectOwner
val creator = UserPool.chooseStudent
```
is hard to follow. A more descriptive name might help me because it is not apparent what the `owner` actually owns and what the `creator` has actually created.

aweng98: I wasn't the original author of tests so I'm guessing names meaning are:
creator - new owner of a clean billing project and also is project owner. this user creates new workspace and groups
owner - can be anyone (student, projectOwner, etc), added as owner of billing project after creator claimed it. in some tests, added as a member of group

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+
+import scala.util.Try
+
+class BillingProjectOwnerPermissionSpec extends FreeSpec with WorkspaceFixtures with BillingFixtures with GroupFixtures with Matchers {
+
+  "removing permissions from billing project owners for workspaces with auth domains" - {
+
+    "+ project owner, + group member, create workspace, - group member" in {
+
+      val owner = UserPool.chooseProjectOwner
```
marctalbott: `owner` is always the user who's permissions we care about. Sometimes they start as a billing project owner or a member of the auth domain or both. As such, `owner` should be changed to a different name like `user`, `student` or something else that indicates that this is the user who's permissions we are testing. On top of that, this user should always be chosen using `UserPool.chooseStudent`

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+      val creator = UserPool.chooseStudent
```
marctalbott: `creator` is always the user who creates the workspaces and is an owner on the billing project. We don't care about their permissions, so we should indicate that somehow. I suggest changing the name to `admin`, `workspaceCreator` or something else that indicates that they are a user with lots of permissions. Similarly, this user should always be chosen using `UserPool.chooseProjectOwner`

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+      implicit val token: AuthToken = creator.makeAuthToken()
+
+      withCleanBillingProject(owner, List(creator.email)) { projectName =>
```
marctalbott: The credentials for the `creator` user should always be passed first here. The email of the `owner` user can be passed in the `List` as required by the test

automation/src/test/scala/org/broadinstitute/dsde/test/api/BillingProjectOwnerPermissionSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+
+              Rawls.billing.removeUserFromBillingProject(projectName, owner.email, BillingProjectRole.Owner)
+              AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(owner.makeAuthToken())
+            }
+          }
+        }
+      }
+    }
+
+    "create workspace, + project owner, + group member, - group member" in {
```
marctalbott: This test and the three tests below it aren't actually doing what they state. When the workspace is created, the `owner` user should be neither a billing project owner nor an auth domain member. Those permissions should then be added via api calls in the order specified by the test. Currently, the `owner` user is already a billing project owner or auth domain member when the workspace is created, which means these tests are the same as previous tests in this suite.

marctalbott (changes requested):
Some suggestions for how to clarify the confusion we had, at least from my point of view. Let me know if you have any questions!

automation/project/Dependencies.scala
```diff
@@ -7,9 +7,9 @@ object Dependencies {
   val akkaV = "2.5.7"
   val akkaHttpV = "10.0.10"
 
-  val serviceTestV = "0.16-c44b1d3"
-  val workbenchGoogleV = "0.18-85787ba"
-  val workbenchModelV  = "0.10-52d614b"
+  val serviceTestV = "0.16-f63a6eb"
+  val workbenchGoogleV = "0.18-5941525"
+  val workbenchModelV  = "0.13-4c7acd5"
```
aweng98, line 12: try to do the right thing, I updated three dependencies to their latest version.

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+import org.broadinstitute.dsde.workbench.service.{Rawls, RestException, Sam}
+import org.scalatest.{FreeSpec, Matchers}
+
+import scala.util.Try
+
+class AuthDomainGroupRoleSpec extends FreeSpec with WorkspaceFixtures with BillingFixtures with GroupFixtures with Matchers {
+
+  "removing permissions from workspace with auth domain" - {
+
+    "+ project owner, + group member, create workspace, - group member" in {
```
marctalbott, line 17: Can we change this test so that we are using the `billingProjectOwner`'s token to create the auth domain group and the workspace and then we are using the `student`'s token to check the accessibility and visibility of the workspace? Basically switching `billingProjectOwner` and `student` in this test to be consistent with the other tests in this suite?

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+
+            // remove "student" from billing project
+            Rawls.billing.removeUserFromBillingProject(projectName, student.email, BillingProjectRole.Owner)
+            AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, create workspace, + group member, - group member" in {
```
marctalbott, line 69: Same comment on this test, switching `student` and `billingProjectOwner` would help with consistency

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+      withCleanBillingProject(billingProjectOwner) { projectName =>
+        withGroup("group", memberEmails = List(student.email)) { groupName =>
+          withCleanUp {
+            withWorkspace(projectName, "GroupApiSpec_workspace", Set(groupName)) { workspaceName =>
+              AuthDomainMatcher.checkNotVisibleNotAccessible(projectName, workspaceName)(student.makeAuthToken())
+
+              // add "student" to billing project with owner role
+              Rawls.billing.addUserToBillingProject(projectName, student.email, BillingProjectRole.Owner)
+              register cleanUp Try(Rawls.billing.removeUserFromBillingProject(projectName, student.email, BillingProjectRole.Owner)).recover {
+                case _: RestException =>
```
ansingh7115: Log the exception (here and in other places in this file where we are following this pattern) so it's not just lost

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+
+            // remove a group member "billingProjectOwner" from group
+            Sam.user.removeUserFromPolicy(groupName, GroupRole.Member.toString, billingProjectOwner.email)
+            AuthDomainMatcher.checkVisibleNotAccessible(projectName, workspaceName)(billingProjectOwner.makeAuthToken())
+          }
+        }
+      }
+    }
+
+    "+ project owner, + group member, - group member" in {
```
marctalbott: This test is a duplicate of the previous test and can be removed.

marctalbott (commented):
I agree with Doug's comment about removing implicit token usage. I think there may also be some test cases that are missing, but I'm having a hard time explaining them via text, so I'll find a time today to try and talk it over in person

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
aweng98: 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainGroupRoleSpec.scala
aweng98: 👍 Register a cleanUp here is redundant because fixtures `withGroup` and `withCleanBillingProject` should take care the cleanUp automatically.  I've removed this from all tests.

marctalbott (approved):

ansingh7115 (approved):

PR #1077, opened 2019-04-25
turn on highsec for all projects

helgridly:
Tough PR, this one. Only for the brave.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -786,7 +786,7 @@ class HttpGoogleServicesDAO(
       "projectOwnersGroup" -> ownerGroupEmail.value.toJson,
       "projectViewersGroup" -> computeUserGroupEmail.value.toJson,
       "requesterPaysRole" -> requesterPaysRole.toJson,
-      "highSecurityNetwork" -> false.toJson,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -786,7 +786,7 @@ class HttpGoogleServicesDAO(
       "projectOwnersGroup" -> ownerGroupEmail.value.toJson,
       "projectViewersGroup" -> computeUserGroupEmail.value.toJson,
       "requesterPaysRole" -> requesterPaysRole.toJson,
-      "highSecurityNetwork" -> false.toJson,
+      "highSecurityNetwork" -> true.toJson, //turned on for all new projects
```
gpolumbo-broad, line 789: Have you taken into account what you _mean_ by "true" in this context?  I mean, what is truth?  Is it not just a construct of human blah blah blah, whatever

gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
helgridly: yes

helgridly:
Closing this because we don't want to do it for now -- it'll overwhelm the flow logs.

PR #1079, opened 2019-04-25
CA-217 Remove prerequisite expressions

helgridly:
We never used prerequisites for anything, and their documentation in Swagger is wrong. Goodbye.

(there will be a firecloud-orch PR that will update rawls-model, but I need to wait until this branch is merged before i can make that.)

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
@@ -342,7 +342,9 @@ case class MethodConfiguration(
                    namespace: String,
                    name: String,
                    rootEntityType: Option[String],
-                   prerequisites: Map[String, AttributeString],
+                   //we used to have prereqs but did nothing with them. so we removed them.
+                   //leaving it as an option means users who are still sending an empty object don't suddenly have their code break.
+                   prerequisites: Option[Map[String, AttributeString]],
                    inputs: Map[String, AttributeString],
                    outputs: Map[String, AttributeString],
                    methodRepoMethod: MethodRepoMethod,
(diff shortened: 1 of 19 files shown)


model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -342,7 +342,9 @@ case class MethodConfiguration(
                    namespace: String,
                    name: String,
                    rootEntityType: Option[String],
-                   prerequisites: Map[String, AttributeString],
+                   //we used to have prereqs but did nothing with them. so we removed them.
+                   //leaving it as an option means users who are still sending an empty object don't suddenly have their code break.
```
MatthewBemis: is this true? i thought we tolerated extra keys

helgridly: honestly i'm not sure

helgridly:
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/19213/

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
helgridly: you're right, we do. will fix

helgridly: actually, i'm going to leave it as an option. it's possible that clients have been written that now expect us to return the prerequisites key, so i'm going to accept Nones coming in but put an empty map in the response as it comes out. otherwise it's a breaking API change.

marctalbott (approved):

dvoet (approved):
also rename the table in the DB _OLD or just drop it. I do note that there are 49 rows in the prod database which look like junk.

PR #1088, opened 2019-05-08
AuthDomainMultiGroupApiSpec: retry failed check with eventually

aweng98:
[QA-565](https://broadworkbench.atlassian.net/browse/QA-565)

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/AuthDomainMultiGroupApiSpec.scala
@@ -79,10 +79,9 @@ class AuthDomainMultiGroupApiSpec extends FreeSpec with Matchers with WorkspaceF
                 Orchestration.workspaces.clone(projectName, workspaceName, projectName, workspaceCloneName, authDomains)(user.makeAuthToken())
 
                 // two authdomain groups should be in cloned workspace
-                eventually {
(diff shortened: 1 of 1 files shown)


helgridly:
this implies that either:
- we don't wait for the HTTP request to /clone to finish before moving on to the next line, which i think would be a testing antipattern; or
- we do wait for the HTTP request to complete, but that it's doing some things asynchronously afterwards. in which case: is this documented and clear?

aweng98:
@helgridly I feel answer is in your second point. Test failure with error `"[NO ACCESS]" was not equal to "[WRITER]"` is intermittent in Alpha but not in dev. I don't know any documentation about this, but Sam does cache and cache time is a minute long (TTL). I used 150 seconds to be "very generously". Do you feel this is a real product issue, not a test issue?

helgridly:
I'm gonna drag in @dvoet here. I don't know why the cache would kick in on the clone of a fresh workspace -- I imagine there wouldn't be a value in the cache at all, but I might be missing something.

aweng98:
Use `eventually` only for failed check for now.

aweng98:
jenkins retest

aweng98:
@gpolumbo-broad any thoughts?

gdlex4015 (approved):

dvoet:
@helgridly the thing that sam caches is the list of groups the user is a member of so a cache miss is possible the way this test is structured.

dvoet (approved):

PR #1091, opened 2019-05-21
untag SmokeTest tag from test

aweng98:
a Firecloud-UI test converted to Rawls API test. Tag SmokeTest was in Firecloud_UI test and copied over in new API test. It is decided now by QA team that the tag is not needed anymore in API test.
Untagging SmokeTest tag for "A workspace with one group in its auth domain can be cloned and retain the auth domain".


aweng98:
jenkins retest

gdlex4015 (approved):

PR #1104, opened 2019-06-21
Ca 275 folders

dvoet:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -1181,9 +1181,13 @@ class HttpGoogleServicesDAO(
   override def addProjectToFolder(projectName: RawlsBillingProjectName, folderId: String): Future[Unit] = {
     implicit val service = GoogleInstrumentedService.CloudResourceManager
     val cloudResourceManager = getCloudResourceManager(getBillingServiceAccountCredential)
+
     retryWhen500orGoogleError( () => {
       val existingProject = executeGoogleRequest(cloudResourceManager.projects().get(projectName.value))
-      val folderResourceId = new ResourceId().setType(GoogleResourceTypes.Folder.value).setId(folderId)
+
+      // google is not consistent when dealing with folder ids. This api does not want the folder id to start with
+      // "folders/" but other apis return that or expect that. Strip it here in case it is present.
+      val folderResourceId = new ResourceId().setType(GoogleResourceTypes.Folder.value).setId(folderId.stripPrefix("folders/"))
       executeGoogleRequest(cloudResourceManager.projects().update(projectName.value, existingProject.setParent(folderResourceId)))
     })
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -858,7 +858,6 @@ class HttpGoogleServicesDAO(
   }
 
   override def pollOperation(operationId: OperationId): Future[OperationStatus] = {
-    val credential = getBillingServiceAccountCredential
     val dmCredential = getDeploymentManagerAccountCredential
 
     // this code is a colossal DRY violation but because the operations collection is different
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -54,7 +54,7 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, HttpClientUtilsStandard}
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.workbench.google2._
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.model.google.{GoogleProject, GoogleResourceTypes}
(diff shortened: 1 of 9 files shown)


marctalbott:
Changing this PR so that it will be merged into `ca-277-with-dminfc` instead of `ca-277-for-reals`. I cherry-picked the changes from `ca-277-for-reals` onto `ca-277-with-dminfc` which I had recently made off of `develop` so as to (hopefully) avoid messiness around merging to `develop` given the recent Deployment Manager change.

marctalbott:
Okay, changing it back because I didn't consider how messy changing the base would make this PR seeing as it now includes all of Hussein's commits from DM in the commit history for this PR. What I was trying to convey is that `ca-277-with-dminfc` is now the most up-to-date branch for service perimeters and that `ca-277-for-reals` is now out of date (and not "for-reals" at all). However, since I won't be here when you all actually perform the Git magic to merge this all together, I'm going to change the base back to how it was and let you all decide how you actually want to handle it. 

As an aside, I've found keeping track of all these branches for the same change and dealing with squashing, cherry-picking, etc. to be a bit complicated, so if anything I'm doing here is bad practice and there's an easier way please let me know!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1177,6 +1186,26 @@ class HttpGoogleServicesDAO(
...
+      val existingProject = executeGoogleRequest(cloudResourceManager.projects().get(projectName.value))
+
+      val folderResourceId = new ResourceId().setType(GoogleResourceTypes.Folder.value).setId(folderNumberOnly(folderId))
+      executeGoogleRequest(cloudResourceManager.projects().update(projectName.value, existingProject.setParent(folderResourceId)))
+    })
+  }
+
+  override def getFolderId(folderName: String): Future[Option[String]] = {
+    val credential = getBillingServiceAccountCredential
+    credential.refreshToken()
```
gpolumbo-broad, line 1203: Side effect call where we are disregarding the return value.  I'm also confused as to why we need to refresh the token in this method but not in other places like line `1191` above.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -86,9 +87,9 @@ trait RawlsBillingProjectComponent {
       }
     }
 
-    def listProjectsWithServicePerimeter(servicePerimeter: ServicePerimeterName): ReadWriteAction[Seq[RawlsBillingProject]] = {
+    def listProjectsWithServicePerimeterAndStatus(servicePerimeter: ServicePerimeterName, statuses: CreationStatus*): ReadWriteAction[Seq[RawlsBillingProject]] = {
```
gpolumbo-broad, line 90: Might we want a method in the future to `listProjectsWithServicePerimeter` regardless of what their status is?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -86,9 +87,9 @@ trait RawlsBillingProjectComponent {
...
       for {
-        projectRecords <- filter(_.servicePerimeter === servicePerimeter.value).result
+        projectRecords <- filter(rec => rec.servicePerimeter === servicePerimeter.value && rec.creationStatus.inSetBind(statuses.map(_.toString))).result
```
gpolumbo-broad, line 92: What happens if `statuses` is `Nil`?  Seems to me like this request would not return any records.  Just want to double check if that is what we want to happen or not?  My gut says it should either return all records in the specified perimeter or it should throw an error.

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
dvoet: we can write it or fix this one when/if we need it

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
dvoet: it will probably error

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: In this particular case we need an access token to make http calls. In the case of 1191 the credential object itself is passed to the google client lib and it takes care of getting the access token. If we don't call refreshToken first then getAccessToken below returns null.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
gpolumbo-broad: I guess the alternative is to take a collection of statuses so that at the very least, callers would need to pass in an empty collection and thus avoid the NPE.

dvoet: it will not be a NPE, it will be a database exception because an IN expression cannot be empty. The right fix would be to exclude the IN clause if the collection is empty... but right now I would just leave it alone

gpolumbo-broad: 👍

gpolumbo-broad (approved):

PR #1109, opened 2019-07-03
CA-379 add submission id to google labels so it shows up in the BQ cost export

helgridly:
I redid this because I didn't read the Cromwell documentation closely enough. The google labels go in workflow options, not workflow labels.


helgridly:
~i guess i could pass the google labels as a separate parameter and defer the jsonifying until the actual request
ie~

```def submitWorkflows(wdl: String, inputs: Seq[String], options: Option[String], labels: Option[Map[String, String]], googleLabels: Option[Map[String, String]], workflowCollection: Option[String], userInfo: UserInfo)```

edit: i am dumb, this is wrong

**Review by gpolumbo-broad — APPROVED** [review]

**Comment by gpolumbo-broad** [comment]
Approved as is, but I'd prefer your latter suggestion for isolating when the JSONification takes place

**Review by dvoet — APPROVED** [review]

**Comment by helgridly** [comment]
wow i did not read the documentation AT ALL, this goes in the workflow options not the workflow labels

File: core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala [hunk 1]
```diff
@@ -199,7 +199,8 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
       runtimeOptions,
       useCallCache,
       billingProject.cromwellBackend.getOrElse(defaultBackend),
-      workflowFailureMode
+      workflowFailureMode,
+      google_labels = Map("terra-submission-id" -> s"terra-${submissionId.toString}")
```
gpolumbo-broad, line 203: perhaps a nit, but do we want to use `terra` here?  There's Terra UI, and then the whole analysis platform backend services suite, which I'm sure some people call Terra, but others don't.   Like, when AoU starts submitting workflows, do we want their submissions called `terra-submissions` still?  Or might there be value in keeping the key something generic like, `submission-id`, and the value something that can have a namespacing prefix, like you have `terra-`

gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
helgridly: oh, sorry -- i just hit merge and the page refreshed and i saw this. my understanding is that everything is called terra now. it's certainly a case of "firecloud, powered by terra"; i'm not sure if that's the ultimate fate of AoU also.

secondly, this ending up on PAPI VMs and our bigquery cost export, so i don't really think end-users will be seeing this anyway

gpolumbo-broad: No worries.  I was totally fine with the PR, hence the approval, was just thinking about the labels in a potentially inconsequential way, so just wanted to ask.  It's all fine with me.

PR #1113, opened 2019-07-11
use submission id in bigquery cost lookup

helgridly:
Follow-on from the previous PR.

I still need to check that the query works; I've done a big submission in a fiab and am waiting for the export to bigquery.

I broke out the original function of "get me costs for this list of workflows" into "get me costs for this submission" and "get me costs for this one workflow", which more closely matches our usage. In the first case, we start by looking up the submission ID, and if that doesn't work, we fall back to executing the monster query. Over time this means we'll run more of the former than the latter, which won't make the results any smaller but the queries will be a lot more manageable.

I also passed through `_PARTITIONDATE` as that's an easy way to limit the amount of data BQ scans, thus making queries cheaper.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
@@ -62,8 +62,9 @@ class SubmissionCostService(tableName: String, serviceProject: String, bigQueryD
 
   private def partitionDateClause(submissionDate: Option[DateTime]): String = {
     (submissionDate map { d: DateTime =>
-      val date = d.toString(DateTimeFormat.forPattern("yyyy-MM-dd"))
-      s"AND _PARTITIONDATE >= $date"
+      //subtract a day so we never have to deal with timezones
+      val date = d.minusDays(1).toString(DateTimeFormat.forPattern("yyyy-MM-dd"))
+      s"""AND _PARTITIONDATE >= "$date""""
     }).getOrElse("")
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
```diff
@@ -17,34 +19,86 @@ object SubmissionCostService {
...
+
+    val querySql: String =
+      s"""SELECT wflabels.key, REPLACE(wflabels.value, "cromwell-", "") as `workflowId`, SUM(billing.cost)
+      |FROM `$tableName` as billing, UNNEST(labels) as wflabels
+      |CROSS JOIN UNNEST(billing.labels) as blabels
+      |WHERE blabels.value = "terra-$submissionId"
+      |AND wflabels.key = "cromwell-workflow-id"
+      |AND project.id = ?
+      |${partitionDateClause(submissionDate)}
+      |GROUP BY wflabels.key, workflowId""".stripMargin
```
helgridly, line 81: I will doublecheck this query once some test runs have been exported to BQ. It's a bit odd to be doing a cross join but it's what gets both the submission AND workflow labels in separate columns. If there are other more sane approaches I would love to hear them!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
```diff
@@ -17,34 +19,86 @@ object SubmissionCostService {
...
-
-    rowsFuture map { rowsOrNull =>
-      Option(rowsOrNull) match {
-        case Some(rows) => rows.asScala.map { row =>
-          // workflow ID is contained in the 2nd cell, cost is contained in the 3rd cell
-          row.getF.get(1).getV.toString -> row.getF.get(2).getV.toString.toFloat
-        }.toMap
-        case None => Map.empty[String, Float]
-      }
+  def extractCostResults(rows: util.List[TableRow]): Map[String, Float] = {
```
helgridly, line 53: I didn't like how this took a future and returned a future when it didn't need to. The caller can `map` if they want to.

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SubmissionCostService.scala
helgridly: I checked it and it works.

PR #1124, opened 2019-08-23
CA-464 chunkify testing permissions for billing accounts

helgridly:
The quota is 300 requests per minute (see [here](https://console.cloud.google.com/apis/api/cloudbilling.googleapis.com/quotas) for some project).

If someone had 300 billing projects and the turnaround time for one IAM test is 2 seconds, we could still hit the quota. (10 requests every 2 seconds = 300/min)

If we don't like this approach we could steal [the throttler from GPAlloc](https://github.com/broadinstitute/gpalloc/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/gpalloc/util/Throttler.scala).

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -664,8 +664,10 @@ class HttpGoogleServicesDAO(
     val cred = getUserCredential(userInfo)
     listBillingAccounts(cred) flatMap { accountList =>
       //some users have TONS of billing accounts, enough to hit quota limits.
-      //break the list of billing accounts up into chunks, and process the chunks serially.
-      //this should slow things down somewhat, though we may need to upgrade this to a throttle.
+      //break the list of billing accounts up into chunks.
+      //each chunk executes all its requests in parallel. the chunks themselves are processed serially.
+      //this limits the amount of parallelism (to 10 inflight requests at a time), which should should slow
+      //our rate of making requests. if this fails to be enough, we may need to upgrade this to an explicit throttle.
       val accountChunks: List[Seq[BillingAccount]] = accountList.grouped(10).toList
 
       //Iterate over each chunk.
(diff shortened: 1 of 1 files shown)


helgridly:
@rtitle i'm not super confident on whether my use of IO here is right so i'd appreciate the sanity check :)

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -659,14 +659,28 @@ class HttpGoogleServicesDAO(
...
+      //some users have TONS of billing accounts, enough to hit quota limits.
+      //break the list of billing accounts up into chunks, and process the chunks serially.
+      //this should slow things down somewhat, though we may need to upgrade this to a throttle.
+      val accountChunks: List[Seq[BillingAccount]] = accountList.grouped(10).toList
+
+      //Iterate over each chunk.
+      val allProcessedChunks: IO[List[Seq[RawlsBillingAccount]]] = accountChunks traverse { chunk =>
+
+        //Future.sequence each chunk (i.e. run all tests in the chunk in parallel)
+        IO.fromFuture(IO(Future.sequence(chunk map { acct =>
```
rtitle: I think
```
Future.sequence(chunk map { acct => ... })
```
is equivalent to
```
chunk traverse { acct => ... }
```

dvoet:
Have we asked google for a quota increase or a better way?

rtitle (approved):
The code looks good -- although making the point that even with chunking we could still hit quota without some sort of throttling/delay.

helgridly:
@dvoet I can certainly ask for a quota increase

dvoet:
Seems like the problem is that we have too much parallelism and that chunking limits that. Can you add something to that effect in the comments? Explain why chunking works.

helgridly:
@dvoet what are you looking for beyond the comment that's already there? https://github.com/broadinstitute/rawls/pull/1124/files#diff-5ecf29f76a772ac794b30c2ba0a65dd7R669

helgridly:
jenkins retest

dvoet:
The old code did them all in parallel. New code does 10 in parallel. The comment is unclear that all elements in a chunk execute in parallel. So really the size of the chunks governs the parallelism. Chunking makes it feel batchy to me which is not the case. I just want to make it clear that the intent is to limit parallelism.

PR #1133, opened 2019-09-10
skip tests should skip spinning up mysql too

helgridly:
This saves waiting around for a minute for something you'll never use.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -92,17 +92,25 @@ fi
 function make_jar()
 {
     echo "building jar..."
-    bash ./docker/run-mysql.sh start
+    if [ $SKIP_TESTS != "skip-tests" ]; then
+        bash ./docker/run-mysql.sh start
+    fi
 
     # Get the last commit hash of the model directory and set it as an environment variable
     GIT_MODEL_HASH=$(git log -n 1 --pretty=format:%h model)
 
     # make jar.  cache sbt dependencies. capture output and stop db before returning.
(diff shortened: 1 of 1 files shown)


docker/build.sh
```diff
@@ -92,17 +93,26 @@ fi
...
     GIT_MODEL_HASH=$(git log -n 1 --pretty=format:%h model)
 
     # make jar.  cache sbt dependencies. capture output and stop db before returning.
-    JAR_CMD=`docker run --rm --link mysql:mysql -e SKIP_TESTS=$SKIP_TESTS -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e GIT_COMMIT -e BUILD_NUMBER -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage /working/docker/install.sh /working`
+    DOCKER_RUN="docker run --rm"
+    if [ "$SKIP_TESTS" != "skip-tests" ]; then
+        DOCKER_RUN="$DOCKER_RUN --link mysql:mysql"
+    fi
+    DOCKER_RUN="$DOCKER_RUN -e SKIP_TESTS=$SKIP_TESTS -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e GIT_COMMIT -e BUILD_NUMBER -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage /working/docker/install.sh /working"
+    JAR_CMD=$($DOCKER_RUN)
```
RLuckom, line 109: it's weird that we assign the output to the `JAR_CMD` variable

RLuckom (approved):

docker/build.sh
helgridly: agreed. i actually changed it to something sensible but then decided it'd end up differing from all the other docker build scripts

dvoet (approved):

PR #1136, opened 2019-09-20
[draft] [noticket] Use an unbounded thread pool for the default dispatcher

helgridly:
This should stop it running out of threads.

fc-dev friend: https://github.com/broadinstitute/firecloud-develop/pull/1833

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -1,3 +1,13 @@
+default-dispatcher {
+  type = "Dispatcher"
+
+  executor = "org.broadinstitute.dsde.rawls.util.CachedThreadPoolExecutorConfigurator"
+
+  shutdown-timeout = 1s
+
+  throughput = 5
+}
+
 submission-monitor-dispatcher {
   # Dispatcher is the name of the event-based dispatcher
   type = Dispatcher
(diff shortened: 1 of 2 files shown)


gpolumbo-broad:
Is there a more thorough description somewhere for the what/why for this change?

helgridly:
This is part of a suite of changes I'm proposing to un-bottleneck Rawls. I don't have the full story yet (and I should probably make this a draft PR), but the short version is: under load, Rawls instances end up being pulled from the load balancer. This is because they timeout when the LB queries `/status`. I believe (note the uncertainty, this is hard to debug!) this is because akka's HTTP request-handling threads are being tied up servicing Gatling requests.

This change makes the number of akka-http request-handling threads functionally infinite. Along with (this PR)[https://github.com/broadinstitute/rawls/pull/1135] which makes less work run on those threads, akka-http should now be able to serve up responses to `/status`. This seems to be borne out in a test I ran on perf on Friday.

gpolumbo-broad:
Seems like it would be reasonable to write this up as a Jira issue

helgridly:
we're handling it on the performance team trello board

dvoet:
unbounded thread pool is generally a bad idea as it leads to general catastrophic exhaustion instead of a more graceful degradation. A bigger thread pool may be a good idea but not unbounded.

PR #1146, opened 2019-10-15
initial stackdriver tracing

kcibul:
For your consideration -- this is what tracing currently looks like.  This was done with the explicit style of parameter passing.  Please comment away in this draft PR and if there needs to be a discussion I'll add it to TPs

I think it's fairly clean looking as a part of a `for` (see HttpGoogleServicesDAO) but is a little harder to follow when used in the rawlsian "withXyz" style because of the long lines and indentation.

We could shorten the name of the tracing methods (e.g. t), which would reduce the visual clutter at the expense of approachability.

For an example of what this span looks like see [this StackDriver Trace GCP Console](https://console.cloud.google.com/traces/traces?project=broad-dsde-dev&organizationId=548622027621&start=1571164648727&end=1571168248727&tid=4b737663f131b14e7873b883039be58e&q=%2Fapi%2Fworkspaces)

There is a corresponding change in firecloud-develop for the config that I'll PR separately, but that's much more straightforward.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -38,6 +38,11 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 import cats.implicits._
+import io.opencensus.trace.Span
+import io.opencensus.scala.Tracing._
+import io.opencensus.trace.{AttributeValue => OpenCensusAttributeValue}
+import org.broadinstitute.dsde.rawls.util.OpenCensusDBIOUtils._
+
 
 /**
  * Created by dvoet on 4/27/15.
@@ -74,7 +79,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
   import dataSource.dataAccess.driver.api._
 
-  def CreateWorkspace(workspace: WorkspaceRequest) = createWorkspace(workspace)
+  def CreateWorkspace(workspace: WorkspaceRequest, parentSpan: Span = null) = createWorkspace(workspace, parentSpan)
   def GetWorkspace(workspaceName: WorkspaceName, params: WorkspaceFieldSpecs) = getWorkspace(workspaceName, params)
   def DeleteWorkspace(workspaceName: WorkspaceName) = deleteWorkspace(workspaceName)
   def UpdateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]) = updateWorkspace(workspaceName, operations)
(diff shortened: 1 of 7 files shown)


kcibul:
jenkins retest

kcibul:
jenkins retest

kcibul:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1985,15 +1990,15 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
             val defaultPolicies = Map(projectOwnerPolicy, ownerPolicy, writerPolicy, readerPolicy, shareReaderPolicy, shareWriterPolicy, canComputePolicy, canCatalogPolicy)
 
-            DBIO.from(samDAO.createResourceFull(SamResourceTypeNames.workspace, workspaceId, defaultPolicies, workspaceRequest.authorizationDomain.getOrElse(Set.empty).map(_.membersGroupName.value), userInfo))
+            traceDBIOWithParent("createResourceFull", parentSpan)(_ => DBIO.from(samDAO.createResourceFull(SamResourceTypeNames.workspace, workspaceId, defaultPolicies, workspaceRequest.authorizationDomain.getOrElse(Set.empty).map(_.membersGroupName.value), userInfo)))
```
rtitle, line 1993: It seems a little weird to me put Sam calls inside a DB transaction. That is not new in your PR, just commenting.

rtitle: I wonder if it would be better to do
```
DBIO.from(traceWithParent("createResourceFull", parentSpan)(_ => samDAO.createResourceFull(…))
```
since this _really_ isn't a database call..

rtitle (approved):
Looks good to me. The code feels a little noisy in places, but it seems like you're using the library correctly.

kcibul:
Any suggestions about the noise @rtitle ?  Since this will serve as the 'example' for folks we want it to be stylistically what we like

rtitle:
I had a few thoughts, not sure how great they are.

1. Use an implicit class. Instead of:
```
traceWithParent("insertInitialStorageLog", parentSpan)(_ => insertInitialStorageLog(bucketName))
```
this would let you write:
```
insertInitialStorageLog(bucketName).trace("insertInitialStorageLog", parentSpan)
```
However, it only works when you're _not_ threading through the `Span` variable so you'd still have to fall back to the other style. It might be more complexity than it's worth.

2. Shadow `span` variable. I found the `s1`s and `s2`s confusing:
```
traceWithParent("first", parentSpan) { s1 =>
  traceWithParent("second", s1) { s2 => 
    …
  }
}
```
We could just use the same `span` variable name at each layer so you're always using the right one (and shadowing ones that you shouldn't use):
```
traceWithParent("first", parentSpan) { span =>
  traceWithParent("second", span) { span => 
    …
  }
}
```

3. In general I'm not a fan of the "loan" pattern which this library seems to use:
```
withResource(…) { res =>
  doSomething(res)
}
```
When there are large chains it becomes hard to read. cats-effect lets you do similar with `Resource` but in a for-comprehension. However I think it's correct usage of this library (and Rawls uses this pattern internally anyway) so I think it's fine. Side project: make an opencensus-cats library :).

dvoet (approved):

kcibul:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: sam calls inside a db transaction is indeed one of the Bad Things we do, ticketed on app services as [AS-9](https://broadworkbench.atlassian.net/browse/AS-9)

helgridly:
Can we get this in?

kcibul:
thanks for ping -- i had to resolve a conflict and retesting

PR #1147, opened 2019-10-22
AS-90/AS-91: Avro upsert monitor

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -406,6 +380,40 @@ object Boot extends IOApp with LazyLogging {
...
+          requesterPaysRole,
+          useWorkflowCollectionField,
+          useWorkflowCollectionLabel,
+          defaultBackend,
+          methodConfigResolver,
+          GoogleProject(conf.getString("avroUpsertMonitor.pubSubProject")),
+          conf.getString("avroUpsertMonitor.pubSubTopic"),
+          conf.getString("avroUpsertMonitor.pubSubSubscription"),
+          conf.getString("avroUpsertMonitor.bucketName")
+        )
```
davidangb, line 408: so many arguments! Consider making a case class `AvroUpsertConfig` or whatever that encapsulates the pubsub project/topic/subscription and the bucket name, and pass that around?

Also, these will require settings in conf or fail, right?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,218 @@
...
+      acknowledgeMessage(ackId).map(_ => StartMonitorPass) pipeTo self
+
+    case Status.Failure(t) => throw t
+
+    case ReceiveTimeout =>
+      throw new WorkbenchException("AvroUpsertMonitorActor has received no messages for too long")
+
+    case x =>
+      logger.info(s"unhandled $x")
+//      self ! StartMonitorPass TODO: should this restart the monitor if we've got an unhandled message? I think so.
```
davidangb: agree it should restart; maybe `self ! None` to get the delay?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  private def initAvroUpsert(jobId: String, file: String): Future[Unit] = {
+    for {
+      avroJson <- readUpsertObject(s"$jobId/$file")
+      petSAJson <- samDAO.getPetServiceAccountKeyForUser(avroJson.namespace, RawlsUserEmail(avroJson.userEmail))
+      petUserInfo <- googleServicesDAO.getUserInfoUsingJson(petSAJson)
+      _ <- workspaceService.apply(petUserInfo).batchUpdateEntities(WorkspaceName(avroJson.namespace, avroJson.name), avroJson.payload.toSeq, true)
+    } yield ()
+  }
```
davidangb, line 205: 1) for debugging purposes, how about adding a log statement here with the jobid, userEmail, and avroJson.payload.toSeq.size? I'd also love to log the result of `batchUpdateEntities`.

2) since this sends `batchUpdateEntities` to `_` ... does it wait for the Future to complete or not? If not, it will ack the pubsub message and start a new monitor pass while this upsert is potentially still running. If it does wait for the Future, we could hit ReceiveTimeouts on the actor. I'm thinking it's okay to start a new monitor pass before this import completes .... just wanted to raise the issue to make sure it's all intentional

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  private def acknowledgeMessage(ackId: String) =
+    pubSubDao.acknowledgeMessagesById(pubSubSubscriptionName, Seq(ackId))
+
+  private def parseMessage(message: PubSubMessage) = {
+    val objectIdPattern = """"([^/]+)/([^/]+)"""".r
```
davidangb: this regex could move up to the `object AvroUpsertMonitor` level - or it should otherwise be a singleton, doesn't need to be recreated for each invocation of `parseMessage`. The compiler might handle this, and it's a micro-optimization, but still.

davidangb (commented):
comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,218 @@
...
+//      self ! StartMonitorPass TODO: should this restart the monitor if we've got an unhandled message? I think so.
+  }
+
+  private def initAvroUpsert(jobId: String, file: String): Future[Unit] = {
+    for {
+      avroJson <- readUpsertObject(s"$jobId/$file")
+      petSAJson <- samDAO.getPetServiceAccountKeyForUser(avroJson.namespace, RawlsUserEmail(avroJson.userEmail))
+      petUserInfo <- googleServicesDAO.getUserInfoUsingJson(petSAJson)
+      _ <- workspaceService.apply(petUserInfo).batchUpdateEntities(WorkspaceName(avroJson.namespace, avroJson.name), avroJson.payload.toSeq, true)
+    } yield ()
```
breilly2: Is this a typical convention (I'm not fluent in akka) or would it make sense to declare a specific `ImportDone` message?

MatthewBemis: I don't think it's necessarily atypical? But you make a good point by calling it out- an ImportDone message would be useful for writing back a log message w/ the success or failure. Will incorporate this into the next iteration!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,218 @@
...
+
+    case None =>
+      // there was no message so wait and try again
+      val nextTime = org.broadinstitute.dsde.workbench.util.addJitter(pollInterval, pollIntervalJitter)
+      system.scheduler.scheduleOnce(nextTime.asInstanceOf[FiniteDuration], self, StartMonitorPass)
+
+    case ((), ackId: String) =>
+      acknowledgeMessage(ackId).map(_ => StartMonitorPass) pipeTo self
+
+    case Status.Failure(t) => throw t
```
breilly2: Again, not fluent in akka... does this catch exceptions thrown by `initAvroUpsert`? Will this kill this actor? Is there a way we can explicitly tell Pub/Sub that this message should be redelivered rather than waiting for ack deadline to expire?

breilly2: I'll partially answer my own question... looks like there is a concept of `nack` but I don't immediately see it in the java client library.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -38,23 +45,31 @@ object BootMonitors extends LazyLogging {
...
-                   methodConfigResolver: MethodConfigResolver)(implicit cs: ContextShift[IO]): Unit = {
+                   methodConfigResolver: MethodConfigResolver,
+                   avroUpsertPubSubProject: GoogleProject,
+                   avroUpsertPubSubTopic: String,
+                   avroUpsertPubSubSubscription: String,
+                   avroUpsertBucketName: String,
+                   avroUpsertBatchSize: Int)(implicit cs: ContextShift[IO]): Unit = {
     //Reset "Launching" workflows to "Queued"
-    resetLaunchingWorkflows(slickDataSource)
+//    resetLaunchingWorkflows(slickDataSource)
```
MatthewBemis: ignore these commented out lines for now. i wanted a back rawls that didn't do all of these other things so i just commented it out

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -119,6 +134,25 @@ object BootMonitors extends LazyLogging {
...
+        googleServicesDAO,
+        samDAO,
+        googleStorage,
+        googlePubSubDAO,
+        avroUpsertPubSubProject,
+        avroUpsertPubSubTopic,
+        avroUpsertPubSubSubscription,
+        avroUpsertBucketName,
+        batchSize,
+        1
```
breilly2: Should we be starting more than 1 of these workers? Having only 1 protects us from a 2nd worker picking up a message that another has been working on past the ack deadline, but also allows us to do only 1 import at a time.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,237 @@
...
+    actorOf(AvroUpsertMonitor.props(pollInterval, pollIntervalJitter, workspaceService, googleServicesDAO, samDAO, googleStorage, pubSubDao, pubSubSubscriptionName, avroUpsertBucketName, batchSize))
+  }
+
+  override val supervisorStrategy =
+    OneForOneStrategy() {
+      case e => {
+        logger.error("unexpected error in avro upsert monitor", e)
+        // start one to replace the error, stop the errored child so that we also drop its mailbox (i.e. restart not good enough)
+        startOne()
+        Stop
```
breilly2, line 100: Does this stop the monitor, monitor supervisor, or both? Since `startOne` is called, this feels like it's in the context of the monitor supervisor, which gives me the impression that this `Stop` applies to the monitor supervisor as well... though I haven't spent any time learning about supervisor strategies so maybe I'm making poor assumptions about the contract of this function.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,237 @@
...
+      val nextTime = org.broadinstitute.dsde.workbench.util.addJitter(pollInterval, pollIntervalJitter)
+      system.scheduler.scheduleOnce(nextTime.asInstanceOf[FiniteDuration], self, StartMonitorPass)
+
+    case ((), ackId: String) => self ! None
+      acknowledgeMessage(ackId).map(_ => StartMonitorPass) pipeTo self
+//
+    case Status.Failure(t) => throw t
+
+    case ReceiveTimeout =>
+      throw new WorkbenchException("AvroUpsertMonitorActor has received no messages for too long")
```
breilly2: Is this an escape hatch in case the actor gets stuck for some unknown reason? Does throwing an exception here cause the actor to be replaced (according to the supervisor strategy above)? Does the time spent in `initAvroUpsert` count as activity?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,237 @@
...
+
+    case x =>
+      logger.info(s"unhandled $x")
+      self ! None
+  }
+
+  private def initAvroUpsert(jobId: String, file: String): Future[Unit] = {
```
breilly2: Nit: don't need `file` anymore.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,237 @@
...
+    for {
+      avroUpsertJson <- readUpsertObject(s"$jobId/upsert.json")
```
breilly2: If this future is created outside of the for comprehension and this line moved to right above the upsert loop, it can be processed in parallel with reading the metadata and getting the pet. Might be a win because this read could take some time and getting the pet involves some network calls.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
MatthewBemis: I am slightly concerned about having more than 1. I think the route we're going to go is to ack as soon as we load the JSON into memory, but PubSub isn't known for being an exactly-once messaging system so I'd be worried about duplicating heavy work with 2 workers and maybe running out of heap space at the same time.

As a compromise, I will make it configurable so that will buy us more time to test it out with 2 workers and then just change the config if we find that it's okay.

breilly2: I agree with your concern. I have been worried about the at-least-once delivery as well. Being able to handle only one of these imports at a time is an unfortunate consequence that I did not foresee. Given no other way to protect against that, I'm fine sticking with 1 worker for now. Low throughout is better than jeopardizing Rawls.

As an improvement, we could put a heartbeat signal in the job bucket to detect both repeated delivery and stalled jobs that need to be retried. Then we can optimize memory usage to make room for concurrent workers.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
MatthewBemis: I'm not well-versed in Akka either but I believe it just stops the particular child. From the `OneForOneStrategy` docs:

``` 
* Applies the fault handling `Directive` (Resume, Restart, Stop) specified in the `Decider`
 * to the child actor that failed, as opposed to [[akka.actor.AllForOneStrategy]] that applies
 * it to all children.
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
MatthewBemis: Yes exactly it's an escape hatch. 

And yes it seems to count because the timeout would be around 10-12 minutes and I've had an initAvroUpsert last as long as 14 mins in the case of a very low batchSize.

Will double check with a very low timeout, though.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
MatthewBemis: Good idea- done!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,234 @@
...
+        case "upsert.json" => initAvroUpsert(jobId, message.ackId).map(_ => ImportComplete) pipeTo self
+        case _ => acknowledgeMessage(message.ackId).map(_ => None) pipeTo self //some other file (i.e. success/failure log, metadata.json)
+      }
+
+    case None =>
+      // there was no message so wait and try again
+      val nextTime = org.broadinstitute.dsde.workbench.util.addJitter(pollInterval, pollIntervalJitter)
+      system.scheduler.scheduleOnce(nextTime, self, StartMonitorPass)
+
+    case ImportComplete => self ! None
```
breilly2: We just completed an import which may have taken a few minutes. Let's `StartMonitorPass` right away. `None` will wait a poll interval before checking the topic again.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,234 @@
...
+      self ! None
+  }
+
+  private def initAvroUpsert(jobId: String, ackId: String): Future[Unit] = {
+    //start this Future asap because it can be long-running
+    val readUpsertObjectFuture = readUpsertObject(s"$jobId/upsert.json")
+
+    for {
+      avroMetadataJson <- readMetadataObject(s"$jobId/metadata.json")
+      //ack the response after we load the json into memory. pro: don't have to worry about ack timeouts for long operations, con: if someone restarts rawls here the uspert is lost
```
breilly2, line 192: Another con: we won't retry if there's a temporary blip with setting up the pet or loading the upsert JSON.

davidangb: agree - I suggest doing the Sam + Google calls first to get the userinfo; if those succeed, start the read of upsert.json, ack the message, perform the upserts. This does incur a delay before starting to read upsert.json, but it protects against potentially intermittent Sam-related problems.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  private def acknowledgeMessage(ackId: String) =
+    pubSubDao.acknowledgeMessagesById(pubSubSubscriptionName, Seq(ackId))
+
+  private def parseMessage(message: PubSubMessage) = {
+    message.contents.parseJson.asJsObject.getFields("name").head.compactPrint match {
+      case objectIdPattern(jobId, file) => (jobId, file)
+      case m => throw new Exception(s"unable to parse message $m")
+    }
+  }
```
davidangb, line 215: talking out loud, and NOT a blocker for this PR ... I'm very slightly nervous about relying on `compactPrint` behavior. If the libraries ever change how they print out json to a String, this could break. We could skip the regex and get right at the data, something like:
```
  private def parseMessage(message: PubSubMessage) = {
    message.contents.parseJson.asJsObject.getFields("name").head match {
      case jss:JsString =>
        val pathParts = jss.value.split('/')
        if (pathParts.size == 2) (pathParts(0), pathParts(1)) else throw new Exception(s"message has wrong parts: $jss")
      case m => throw new Exception(s"unable to parse message $m")
    }
  }
```
again, no need to change anything for this PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  private def readMetadataObject(path: String): Future[AvroMetadataJson] = {
+    readObject[AvroMetadataJson](path)
+  }
+
+  private def readObject[T](path: String)(implicit reader: JsonReader[T]): Future[T] = {
+    googleStorage.getBlobBody(GcsBucketName(avroUpsertBucketName), GcsBlobName(path)).compile.to[Array].unsafeToFuture().map { byteArray =>
+      byteArray.map(_.toChar).mkString.parseJson.convertTo[T]
+    }
+  }
```
davidangb, line 230: for follow-on work, NOT for this PR: I think we all know this is one of the vulnerable spots, if the file we're reading is too big. Soon after this lands let's follow up with a streaming implementation. [Alpakka](https://doc.akka.io/docs/alpakka/current/data-transformations/json.html) might be a good solution. I envision something like starting a stream read, and as soon as we've collected `$batchSize` operations or the end of file, fire a message containing those upserts back to self. Alternately we could stream and collect all operations into a single Seq, if there's a reason to do that.

( repeat: not for this PR )

davidangb (approved):
looking really good, thanks for cranking on this. I have one comment about doing Sam work before reading the upsert, and a couple thoughts for the future. I'd love to see the sam/upsert ordering change, but I'd be fine with this PR as-is and making that change in a follow-on.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
davidangb: I totally agree this is a problem - I'd like to solve it "for real" in follow-on work that would incorporate a real persistence store beyond writing files to a jobid subdirectoy. If we had a real database we could do things like check job id status easily for end users, list the user's jobs ... and have a place to persist internal semaphores like "the message for this job has already been seen"

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
davidangb, line 230: Do you know if this call is retrieving the gzipped payload, or if it is being [transcoded](https://cloud.google.com/storage/docs/transcoding) and you're receiving the plaintext payload? Ideally we'd be sending gzip over the network to cut down on egress costs and download time.

MatthewBemis: It's really good that you called this out because I realize now in my testing I was actually receiving the plaintext payload...because the files I was placing in the bucket were plaintext, which is neither of the real use-cases.

Will look into getting this to work with gzipped payloads

davidangb (approved):
:+1:

PR #1163, opened 2019-12-06
[AS-134] Expose requireAccess as checkSamActionWithLock

helgridly:
In import-service-land we need to check to see if the user can actually import data. It is not as simple as "has the write action in Sam", as you can't import if the workspace is locked. Our current approach is to do an empty import first -- try it and see. This strikes me as Weird :tm: . So this PR essentially exposes `requireAccess` as an endpoint so we can test it and leave the implementation on the Rawls side.

I will open an Orch swagger PR if this gets thumbed.

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -2423,7 +2423,7 @@ paths:
       responses:
         '204':
           description: User can perform the given action on the workspace
-        '401':
+        '403':
           description: User may not perform the given action on the workspace (including if it doesn't exist)
           schema:
             $ref: '#/definitions/ErrorReport'
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -2418,6 +2418,41 @@ paths:
             - openid
             - email
             - profile
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/checkSamActionWithLock/{samActionName}':
+    get:
+      responses:
+        '204':
+          description: User can perform the given action on the workspace
+        '401':
+          description: User may not perform the given action on the workspace (including if it doesn't exist)
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      description: Check to see if the user has the given action on a workspace. Takes into account if the workspace is locked too.
+      tags:
+        - workspaces
+      summary: Check sam action with lock
+      operationId: checkSamActionWithLock
+      parameters:
+        - in: path
+          description: Workspace Namespace
+          name: workspaceNamespace
+          required: true
+          type: string
+        - in: path
+          description: Workspace Name
+          name: workspaceName
+          required: true
+          type: string
+        - in: path
+          description: Sam action
+          name: samActionName
+          required: true
+          type: string
+      security:
+        - authorization:
+            - openid
+            - email
+            - profile
   '/api/workspaces/{workspaceNamespace}/{workspaceName}/lock':
     put:
       responses:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -2425,6 +2425,8 @@ paths:
           description: User can perform the given action on the workspace
         '403':
           description: User may not perform the given action on the workspace (including if it doesn't exist)
+        '500':
+          description: Rawls Internal Error
           schema:
             $ref: '#/definitions/ErrorReport'
       description: Check to see if the user has the given action on a workspace in Sam. Takes into account if the workspace is locked too.
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -2418,7 +2418,7 @@ paths:
             - openid
             - email
             - profile
-  '/api/workspaces/{workspaceNamespace}/{workspaceName}/checkSamActionWithLock/{samActionName}':
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/checkIamActionWithLock/{samActionName}':
     get:
       responses:
         '204':
@@ -2427,11 +2427,11 @@ paths:
           description: User may not perform the given action on the workspace (including if it doesn't exist)
           schema:
             $ref: '#/definitions/ErrorReport'
-      description: Check to see if the user has the given action on a workspace. Takes into account if the workspace is locked too.
+      description: Check to see if the user has the given action on a workspace in Sam. Takes into account if the workspace is locked too.
       tags:
         - workspaces
-      summary: Check sam action with lock
-      operationId: checkSamActionWithLock
+      summary: Check IAM action with lock
+      operationId: checkIamActionWithLock
       parameters:
         - in: path
           description: Workspace Namespace
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1847,7 +1847,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
     //if we failed for any reason, the user can't do that thing on the workspace
     access.recover { case _ =>
-      RequestComplete(StatusCodes.Unauthorized) }
+      RequestComplete(StatusCodes.Forbidden) }
   }
 
   def listAllActiveSubmissions() = {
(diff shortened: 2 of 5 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2418,6 +2418,41 @@ paths:
             - openid
             - email
             - profile
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/checkSamActionWithLock/{samActionName}':
```
davidangb: I would prefer to not expose the codename "Sam" in API signatures.  Rename this?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1830,6 +1831,19 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      withWorkspaceContext(workspaceName, dataAccess, Some(WorkspaceAttributeSpecs(all=false))) { workspaceContext =>
+        requireAccess(workspaceContext.workspace, samAction) {
+          DBIO.successful(RequestComplete(StatusCodes.NoContent)) //if we get here, we passed all the hoops
+        }
+      }
+    }
+    //if we failed for any reason, the user can't do that thing on the workspace
+    testFuture.recover { case _ =>
+      RequestComplete(StatusCodes.Unauthorized) }
+  }
```
davidangb, line 1851: I know it's rampant throughout Rawls, but let's not add more: could you rewrite this so we're not making REST calls  within a db transaction?

core/src/main/resources/swagger/api-docs.yaml
helgridly: IAM?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: sure

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2132,6 +2152,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  //future-y version
+  private def requireAccessF[T](workspace: Workspace, requiredAction: SamResourceAction)(codeBlock: => Future[T]): Future[T] = {
```
ansingh7115: We now have two versions of requireAccess that both have extremely similar code. It would be nice if, in the future, we could get rid of calling Sam from inside db transactions - but that's definitely outside the scope of this ticket.

ansingh7115 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2132,6 +2152,26 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+            Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, accessDeniedMessage(workspace.toWorkspaceName))))
+          }
+          else {
+            Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound, noSuchWorkspaceMessage(workspace.toWorkspaceName))))
+          }
+        }
+      }
+    }
+  }
+
```
davidangb, line 2155: could you do something like this?
```
  private def doTheThingToRequireAccess(workspace: Workspace, requiredAction: SamResourceAction): Future[Boolean] = {
    samDAO.userHasAction(SamResourceTypeNames.workspace, workspace.workspaceId, requiredAction, userInfo) flatMap { hasRequiredLevel =>
      if (hasRequiredLevel) {
        if (Set(SamWorkspaceActions.write, SamWorkspaceActions.compute).contains(requiredAction) && workspace.isLocked)
          Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, s"The workspace ${workspace.toWorkspaceName} is locked.")))
        else Future.successful(true)
      } else {
        samDAO.userHasAction(SamResourceTypeNames.workspace, workspace.workspaceId, SamWorkspaceActions.read, userInfo) flatMap { canRead =>
          if (canRead) {
            Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, accessDeniedMessage(workspace.toWorkspaceName))))
          }
          else {
            Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound, noSuchWorkspaceMessage(workspace.toWorkspaceName))))
          }
        }
      }
    }
  }

  private def requireAccess[T](workspace: Workspace, requiredAction: SamResourceAction)(codeBlock: => ReadWriteAction[T]): ReadWriteAction[T] = {
    DBIO.from(doTheThingToRequireAccess(workspace, requiredAction)).flatMap{ _ => codeBlock}
  }

  private def requireAccessF[T](workspace: Workspace, requiredAction: SamResourceAction)(codeBlock: => Future[T]): Future[T] = {
    doTheThingToRequireAccess(workspace, requiredAction).flatMap{ _ => codeBlock}
  }
```

obv you'd want to rename `doTheThingToRequireAccess` and might not want to return a `Future[Boolean]` from it but hopefully you get the idea

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1832,15 +1832,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+    val access: Future[PerRequestMessage] = wsCtxFuture flatMap { workspaceContext =>
+      requireAccessF(workspaceContext.workspace, samAction) {
+        Future.successful(RequestComplete(StatusCodes.NoContent)) //if we get here, we passed all the hoops
       }
     }
+
     //if we failed for any reason, the user can't do that thing on the workspace
-    testFuture.recover { case _ =>
+    access.recover { case _ =>
       RequestComplete(StatusCodes.Unauthorized) }
```
davidangb: I think this should be a 403, not a 401

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1832,15 +1832,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-    val testFuture = dataSource.inTransaction { dataAccess =>
-      withWorkspaceContext(workspaceName, dataAccess, Some(WorkspaceAttributeSpecs(all=false))) { workspaceContext =>
-        requireAccess(workspaceContext.workspace, samAction) {
-          DBIO.successful(RequestComplete(StatusCodes.NoContent)) //if we get here, we passed all the hoops
-        }
+    val wsCtxFuture = dataSource.inTransaction { dataAccess =>
+      withWorkspaceContext(workspaceName, dataAccess, Some(WorkspaceAttributeSpecs(all = false))) { workspaceContext =>
+        DBIO.successful(workspaceContext)
+      }
+    }
```
davidangb, line 1839: NOT for this PR - just for saying it out loud: `withWorkspaceContext` itself contains REST calls to Sam, and is within the transaction. I don't wanna blow up scope too much so I'm ok with it - just wanna get it out there in case folks don't realize that.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: yeah -- it was a choice of "don't add another sam call inside a txn" or "write a duplicated version of this function".

edit:/ un-duplicated per DA's suggestion.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: > `withWorkspaceContext` itself contains REST calls to Sam

I don't see any?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: what is `verifyAccess`?

davidangb: oh hah sorry that  was supposed to be `doTheThingToRequireAccess`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: ah dammit I should maybe just go back to bed. I was thinking of `withWorkspaceContextAndPermissions`. My fault   here - sorry

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: updated the example to reflect what I really meant.

davidangb (approved):
:+1: thank you!

PR #1168, opened 2020-01-23
AS-137: Rawls feature branch for import service

ansingh7115:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -189,6 +189,7 @@ object Boot extends IOApp with LazyLogging {
         conf.getString("avroUpsertMonitor.pubSubProject"),
         workbenchMetricBaseName = metricsPrefix
       )
+      val importServiceDAO = new ImportServiceDAO(conf.getString("avroUpsertMonitor.server"))
 
       val bigQueryDAO = new HttpGoogleBigQueryDAO(
         appName,
@@ -399,6 +400,7 @@ object Boot extends IOApp with LazyLogging {
           samDAO,
           pubSubDAO,
           avroUpsertPubSubDAO,
+          importServiceDAO,
           appDependencies.googleStorageService,
           methodRepoDAO,
           dosResolver,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -189,7 +189,7 @@ object Boot extends IOApp with LazyLogging {
         conf.getString("avroUpsertMonitor.pubSubProject"),
         workbenchMetricBaseName = metricsPrefix
       )
-      val importServiceDAO = new ImportServiceDAO(conf.getString("avroUpsertMonitor.server"))
+      val importServiceDAO = new HttpImportServiceDAO(conf.getString("avroUpsertMonitor.server"))
 
       val bigQueryDAO = new HttpGoogleBigQueryDAO(
         appName,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -182,7 +182,7 @@ object Boot extends IOApp with LazyLogging {
       )
 
       // Avro upsert uses a different project for its pubsub topics
-      // remove when cut over to import service
+      // remove when cutting over to import service
       val arrowPubSubDAO = new HttpGooglePubSubDAO(
         clientEmail,
         pathToPem,
@@ -400,7 +400,7 @@ object Boot extends IOApp with LazyLogging {
           gcsDAO,
           samDAO,
           pubSubDAO,
-          arrowPubSubDAO, // remove when cut over to import service
+          arrowPubSubDAO, // remove when cutting over to import service
           importServiceDAO,
           appDependencies.googleStorageService,
           methodRepoDAO,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -182,11 +182,12 @@ object Boot extends IOApp with LazyLogging {
       )
 
       // Avro upsert uses a different project for its pubsub topics
-      val avroUpsertPubSubDAO = new HttpGooglePubSubDAO(
+      // remove when cut over to import service
+      val arrowPubSubDAO = new HttpGooglePubSubDAO(
         clientEmail,
         pathToPem,
         appName,
-        conf.getString("avroUpsertMonitor.pubSubProject"),
+        conf.getString("avroUpsertMonitor.arrowPubSubProject"),
         workbenchMetricBaseName = metricsPrefix
       )
       val importServiceDAO = new HttpImportServiceDAO(conf.getString("avroUpsertMonitor.server"))
@@ -399,7 +400,7 @@ object Boot extends IOApp with LazyLogging {
           gcsDAO,
           samDAO,
           pubSubDAO,
-          avroUpsertPubSubDAO,
+          arrowPubSubDAO, // remove when cut over to import service
           importServiceDAO,
           appDependencies.googleStorageService,
           methodRepoDAO,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
@@ -1,0 +1,69 @@
+package org.broadinstitute.dsde.rawls.dataaccess
+
+import java.util.UUID
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.client.RequestBuilding.Get
+import akka.stream.Materializer
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.dataaccess.ImportStatuses.ImportStatus
+import org.broadinstitute.dsde.rawls.model.{RawlsEnumeration, UserInfo, WorkspaceName}
+import org.broadinstitute.dsde.rawls.util.{HttpClientUtilsStandard, Retry}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ImportServiceJsonSupport {
+  import spray.json.DefaultJsonProtocol._
+
+  implicit val importServiceResponseFormat = jsonFormat1(ImportServiceResponse)
+}
+
+class HttpImportServiceDAO(url: String)(implicit val system: ActorSystem, val materializer: Materializer, val executionContext: ExecutionContext) extends ImportServiceDAO with DsdeHttpDAO with Retry {
+
+  val http = Http(system)
+  val httpClientUtils = HttpClientUtilsStandard()
+
+  def getImportStatus(importId: UUID, workspaceName: WorkspaceName, userInfo: UserInfo): Future[Option[ImportStatus]] = {
+    import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+    import spray.json.DefaultJsonProtocol._
+    import ImportServiceJsonSupport._
+
+    val requestUrl = s"$url/iservice/${workspaceName.namespace}/${workspaceName.name}/imports/$importId"
+    val importStatusResponse: Future[ImportServiceResponse] =  retry[ImportServiceResponse](when500) { () =>
+        executeRequestWithToken[ImportServiceResponse](userInfo.accessToken)(Get(requestUrl))
+    }
+
+    importStatusResponse.map { response =>
+      if( response.status.isEmpty ) {
+        logger.warn(s"Status not found for import with id $importId into workspace ${workspaceName.name} in namespace ${workspaceName.namespace} for user ${userInfo.userEmail}.")
+      }
+      response.status.map( statusString => ImportStatuses.withName(statusString))
+    }
+  }
+}
+
+
+case class ImportServiceResponse(status: Option[String])
+
+
+
+object ImportStatuses {
+  sealed trait ImportStatus extends RawlsEnumeration[ImportStatus] {
+    override def toString = getClass.getSimpleName.stripSuffix("$")
+    override def withName(name: String): ImportStatus = ImportStatuses.withName(name)
+  }
+
+  def withName(name: String): ImportStatus = name.toLowerCase match {
+    case "readyforupsert" => ReadyForUpsert
+    case "upserting" => Upserting
+    case "done" => Done
+    case "error" => Error()
+    case _ => throw new RawlsException(s"invalid ImportStatus [${name}]")
+  }
+
+  case object ReadyForUpsert extends ImportStatus
+  case object Upserting extends ImportStatus
+  case object Done extends ImportStatus
+  case class Error(message: String = "") extends ImportStatus
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
@@ -5,9 +5,10 @@ import java.util.UUID
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.client.RequestBuilding.Get
+import akka.http.scaladsl.marshalling.Marshal
 import akka.http.scaladsl.model.StatusCodes
 import akka.stream.Materializer
-import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.model.ImportStatuses
 import org.broadinstitute.dsde.rawls.model.ImportStatuses.ImportStatus
 import org.broadinstitute.dsde.rawls.model.{UserInfo, WorkspaceName}
@@ -32,11 +33,13 @@ class HttpImportServiceDAO(url: String)(implicit val system: ActorSystem, val ma
     import ImportServiceJsonSupport._
 
     val requestUrl = s"$url/${workspaceName.namespace}/${workspaceName.name}/imports/$importId"
+
     val importStatusResponse: Future[Option[ImportServiceResponse]] =  retry[Option[ImportServiceResponse]](when500) { () =>
         executeRequestWithToken[Option[ImportServiceResponse]](userInfo.accessToken)(Get(requestUrl)) recover {
           case notOK: RawlsExceptionWithErrorReport if notOK.errorReport.statusCode.contains(StatusCodes.NotFound) => None
         }
     }
+
     importStatusResponse.map { response =>
       response.map( statusString => ImportStatuses.withName(statusString.status))
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
@@ -5,10 +5,12 @@ import java.util.UUID
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.client.RequestBuilding.Get
+import akka.http.scaladsl.model.StatusCodes
 import akka.stream.Materializer
-import org.broadinstitute.dsde.rawls.RawlsException
-import org.broadinstitute.dsde.rawls.dataaccess.ImportStatuses.ImportStatus
-import org.broadinstitute.dsde.rawls.model.{RawlsEnumeration, UserInfo, WorkspaceName}
+import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.model.ImportStatuses
+import org.broadinstitute.dsde.rawls.model.ImportStatuses.ImportStatus
+import org.broadinstitute.dsde.rawls.model.{UserInfo, WorkspaceName}
 import org.broadinstitute.dsde.rawls.util.{HttpClientUtilsStandard, Retry}
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -16,7 +18,7 @@ import scala.concurrent.{ExecutionContext, Future}
 object ImportServiceJsonSupport {
   import spray.json.DefaultJsonProtocol._
 
-  implicit val importServiceResponseFormat = jsonFormat1(ImportServiceResponse)
+  implicit val importServiceResponseFormat = jsonFormat2(ImportServiceResponse)
 }
 
 class HttpImportServiceDAO(url: String)(implicit val system: ActorSystem, val materializer: Materializer, val executionContext: ExecutionContext) extends ImportServiceDAO with DsdeHttpDAO with Retry {
@@ -29,41 +31,20 @@ class HttpImportServiceDAO(url: String)(implicit val system: ActorSystem, val ma
     import spray.json.DefaultJsonProtocol._
     import ImportServiceJsonSupport._
 
-    val requestUrl = s"$url/iservice/${workspaceName.namespace}/${workspaceName.name}/imports/$importId"
-    val importStatusResponse: Future[ImportServiceResponse] =  retry[ImportServiceResponse](when500) { () =>
-        executeRequestWithToken[ImportServiceResponse](userInfo.accessToken)(Get(requestUrl))
+    val requestUrl = s"$url/${workspaceName.namespace}/${workspaceName.name}/imports/$importId"
+    val importStatusResponse: Future[Option[ImportServiceResponse]] =  retry[Option[ImportServiceResponse]](when500) { () =>
+        executeRequestWithToken[Option[ImportServiceResponse]](userInfo.accessToken)(Get(requestUrl)) recover {
+          case notOK: RawlsExceptionWithErrorReport if notOK.errorReport.statusCode.contains(StatusCodes.NotFound) => None
+        }
     }
-
     importStatusResponse.map { response =>
-      if( response.status.isEmpty ) {
-        logger.warn(s"Status not found for import with id $importId into workspace ${workspaceName.name} in namespace ${workspaceName.namespace} for user ${userInfo.userEmail}.")
-      }
-      response.status.map( statusString => ImportStatuses.withName(statusString))
+      response.map( statusString => ImportStatuses.withName(statusString.status))
     }
   }
+
 }
 
 
-case class ImportServiceResponse(status: Option[String])
+case class ImportServiceResponse(id: String, status: String)
 
 
-
-object ImportStatuses {
-  sealed trait ImportStatus extends RawlsEnumeration[ImportStatus] {
-    override def toString = getClass.getSimpleName.stripSuffix("$")
-    override def withName(name: String): ImportStatus = ImportStatuses.withName(name)
-  }
-
-  def withName(name: String): ImportStatus = name.toLowerCase match {
-    case "readyforupsert" => ReadyForUpsert
-    case "upserting" => Upserting
-    case "done" => Done
-    case "error" => Error()
-    case _ => throw new RawlsException(s"invalid ImportStatus [${name}]")
-  }
-
-  case object ReadyForUpsert extends ImportStatus
-  case object Upserting extends ImportStatus
-  case object Done extends ImportStatus
-  case class Error(message: String = "") extends ImportStatus
-}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.rawls.monitor
 
 import java.io.ByteArrayInputStream
+import java.util.UUID
 import java.util.zip.GZIPInputStream
 
 import akka.actor.SupervisorStrategy.{Escalate, Stop}
@@ -9,11 +10,12 @@ import akka.pattern._
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SamDAO}
+import org.broadinstitute.dsde.rawls.dataaccess.ImportStatuses.ImportStatus
+import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, ImportServiceDAO, ImportStatuses, SamDAO}
 import org.broadinstitute.dsde.rawls.google.GooglePubSubDAO
 import org.broadinstitute.dsde.rawls.google.GooglePubSubDAO.PubSubMessage
 import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.EntityUpdateDefinition
-import org.broadinstitute.dsde.rawls.model.{RawlsUserEmail, UserInfo, WorkspaceName}
+import org.broadinstitute.dsde.rawls.model.{RawlsUserEmail, WorkspaceName}
 import org.broadinstitute.dsde.rawls.monitor.AvroUpsertMonitorSupervisor.AvroUpsertMonitorConfig
 import org.broadinstitute.dsde.rawls.workspace.WorkspaceService
 import org.broadinstitute.dsde.workbench.google2.{GcsBlobName, GoogleStorageService}
@@ -23,10 +25,13 @@ import org.broadinstitute.dsde.workbench.util.FutureSupport
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 
-import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.{FiniteDuration, _}
 import scala.language.postfixOps
 import scala.util.{Failure, Success}
+import org.broadinstitute.dsde.rawls.model.UserInfo
+import org.joda.time.DateTime
+
+import scala.concurrent.{ExecutionContext, Future}
 
 /**
   * Created by mbemis on 10/17/19.
@@ -39,8 +44,10 @@ object AvroUpsertMonitorSupervisor {
   final case class AvroUpsertMonitorConfig(pollInterval: FiniteDuration,
                                            pollIntervalJitter: FiniteDuration,
                                            pubSubProject: GoogleProject,
-                                           pubSubTopic: String,
-                                           pubSubSubscription: String,
+                                           importRequestPubSubTopic: String,
+                                           importRequestPubSubSubscription: String,
+                                           importStatusPubSubTopic: String,
+                                           importStatusPubSubSubscription: String,
                                            bucketName: String,
                                            batchSize: Int,
                                            workerCount: Int)
@@ -50,6 +57,7 @@ object AvroUpsertMonitorSupervisor {
              samDAO: SamDAO,
              googleStorage: GoogleStorageService[IO],
              pubSubDao: GooglePubSubDAO,
+             importServiceDAO: ImportServiceDAO,
              avroUpsertMonitorConfig: AvroUpsertMonitorConfig)(implicit executionContext: ExecutionContext, cs: ContextShift[IO]): Props =
     Props(
       new AvroUpsertMonitorSupervisor(
@@ -58,6 +66,7 @@ object AvroUpsertMonitorSupervisor {
         samDAO,
         googleStorage,
         pubSubDao,
+        importServiceDAO,
         avroUpsertMonitorConfig))
 }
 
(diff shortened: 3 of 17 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -0,0 +1,69 @@
...
+      if( response.status.isEmpty ) {
+        logger.warn(s"Status not found for import with id $importId into workspace ${workspaceName.name} in namespace ${workspaceName.namespace} for user ${userInfo.userEmail}.")
+      }
+      response.status.map( statusString => ImportStatuses.withName(statusString))
+    }
+  }
+}
+
+
+case class ImportServiceResponse(status: Option[String])
```
MatthewBemis: can this be `Option[ImportStatus]` and then you can skip the parsing on line 41?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockImportServiceDAO.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import org.broadinstitute.dsde.rawls.dataaccess.ImportStatuses.ImportStatus
+import org.broadinstitute.dsde.rawls.model.{RawlsGroup, RawlsGroupRef, RawlsUser, UserInfo, WorkspaceName}
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.Future
+
+class MockImportServiceDAO extends ImportServiceDAO {
+
+  val groups: TrieMap[UUID, ImportStatus] = TrieMap()
```
MatthewBemis: `groups`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  val http = Http(system)
+  val httpClientUtils = HttpClientUtilsStandard()
+
+  def getImportStatus(importId: UUID, workspaceName: WorkspaceName, userInfo: UserInfo): Future[Option[ImportStatus]] = {
+    import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+    import spray.json.DefaultJsonProtocol._
+    import ImportServiceJsonSupport._
+
+    val requestUrl = s"$url/iservice/${workspaceName.namespace}/${workspaceName.name}/imports/$importId"
```
helgridly: FYI the `iservice` prefix is going to be dropped pretty soon -- it's a holdover from the cloud function days

ansingh7115:
Test is flakily passing. More investigation necessary

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
helgridly: For successful responses the format is `{"id": "some_uuid", "status": "Pending (or whatever)"}`, so if the response is 2xx you don't need to mark the status as optional.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -0,0 +1,69 @@
...
+    val importStatusResponse: Future[ImportServiceResponse] =  retry[ImportServiceResponse](when500) { () =>
+        executeRequestWithToken[ImportServiceResponse](userInfo.accessToken)(Get(requestUrl))
+    }
```
helgridly, line 41: I think this will fail for non-2xx responses, including a 404. (I'm comparing this code against `HttpMethodRepoDAO.getAgoraEntity` which calls `recover` on the http call in order to turn a 404 into a `Future(None)`)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+
+case class ImportServiceResponse(status: Option[String])
+
+
+
+object ImportStatuses {
```
helgridly: model classes like this usually live in the `model` package (as do their json support objects)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -80,16 +89,15 @@ class AvroUpsertMonitorSupervisor(workspaceService: UserInfo => WorkspaceService
     case Status.Failure(t) => logger.error("error initializing avro upsert monitor", t)
   }
 
-  def topicToFullPath(topicName: String) = s"projects/${avroUpsertMonitorConfig.pubSubProject.value}/topics/${avroUpsertMonitorConfig.pubSubTopic}"
```
helgridly, line 83: Where did this go? It's still being used further down

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -38,26 +43,29 @@ object AvroUpsertMonitorSupervisor {
 
   final case class AvroUpsertMonitorConfig(pollInterval: FiniteDuration,
                                            pollIntervalJitter: FiniteDuration,
-                                           pubSubProject: GoogleProject,
-                                           pubSubTopic: String,
-                                           pubSubSubscription: String,
-                                           bucketName: String,
+                                           importRequestPubSubTopic: String,
+                                           importRequestPubSubSubscription: String,
+                                           importStatusPubSubTopic: String,
```
helgridly: maybe call this `updateImportStatusXXX` to clarify that one is for reading and one is for writing?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -80,16 +89,15 @@ class AvroUpsertMonitorSupervisor(workspaceService: UserInfo => WorkspaceService
     case Status.Failure(t) => logger.error("error initializing avro upsert monitor", t)
   }
 
-  def topicToFullPath(topicName: String) = s"projects/${avroUpsertMonitorConfig.pubSubProject.value}/topics/${avroUpsertMonitorConfig.pubSubTopic}"
-
   def init =
     for {
-      _ <- pubSubDao.createSubscription(avroUpsertMonitorConfig.pubSubTopic, avroUpsertMonitorConfig.pubSubSubscription, Some(600)) //TODO: read from config
+      _ <- pubSubDao.createSubscription(avroUpsertMonitorConfig.importRequestPubSubTopic, avroUpsertMonitorConfig.importRequestPubSubSubscription, Some(600))
+      _ <- pubSubDao.createSubscription(avroUpsertMonitorConfig.importStatusPubSubTopic, avroUpsertMonitorConfig.importStatusPubSubSubscription, Some(600))
```
helgridly: if i understand correctly, this topic is for rawls to update the status of an import when it's done, which means it doesn't need a subscription, it just needs to publish to the topic

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -160,20 +172,24 @@ class AvroUpsertMonitorActor(
...
-            case Failure(t) =>  writeResult(s"$jobId/error.json", resultString(t.getMessage)).map(_ => ImportComplete) pipeTo self
+      logger.info(s"received async upsert message: $message")
+      //post to import service pubsub topic
+      val attributes = parseMessage(message)
+      for {
+        petUserInfo <- getPetServiceAccountUserInfo(attributes.workspace, attributes.userEmail)
+        importStatus <- importServiceDAO.getImportStatus(attributes.importId, attributes.workspace, petUserInfo)
+        _ <- importStatus match {
+            case Some(ImportStatuses.ReadyForUpsert) => {
+              publishMessageToUpdateImportStatus(attributes.importId, ImportStatuses.Upserting)
```
helgridly: there is a small chance that pubsub might deliver the message twice, but you're pulling the messages one by one and processing each in turn. so this will be fine on the condition that there is only ever one upsert monitor thread. if someone decides to be helpful and make more than one, the get-and-update import status isn't an atomic action so you might race here. that might be worth noting somewhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -987,6 +987,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
 
   def batchUpdateEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition], upsert: Boolean = false): Future[PerRequestMessage] = {
+    logger.info(s"Inside batchUpdateEntities workspaceName $workspaceName, with entities $entityUpdates with upsert $upsert")
```
MatthewBemis: do you want to leave these in?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,55 +210,88 @@ class AvroUpsertMonitorActor(
...
+    } yield ()
+  }
+
+  private def publishMessageToUpdateImportStatus(importId: UUID, importStatus: ImportStatus) = {
+    val now = DateTime.now
+    val requiredAttributes = Map("importId" -> importId.toString, "updateStatus" -> importStatus.toString, "updatedDate" -> now.toString)
+    val attributes = requiredAttributes ++ {importStatus match {
+      case error: ImportStatuses.Error => Map("errorMessage" -> error.message)
+      case _ => Map()
+    }}
```
helgridly: a case class `importId: UUID, status: ImportStatus, errorMessage: Optional[String] = None` with json support would make this function trivial

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
helgridly: also, what happens if any of these `publishMessageToUpdateImportStatus` calls fails? exception?

also also: this call returns a Future, I think. since this isn't in a `for`, you're not caring about what happens to it?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,55 +210,88 @@ class AvroUpsertMonitorActor(
       self ! None
   }
 
-  private def initAvroUpsert(jobId: String, ackId: String): Future[Unit] = {
+  private def getPetServiceAccountUserInfo(workspaceName: WorkspaceName, userEmail: RawlsUserEmail) = {
```
helgridly: this seems generic enough to live not-here

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,55 +210,88 @@ class AvroUpsertMonitorActor(
...
+      case _ => Map()
+    }}
+    val message = GooglePubSubDAO.MessageRequest("", attributes)
+    pubSubDao.publishMessages(importStatusPubSubTopic, Seq(message))
+  }
+
+
+  private def initUpsert(upsertFile: String, jobId: UUID, ackId: String, workspaceName: WorkspaceName, userInfo: UserInfo): Future[Unit] = {
+    for {
+      //ack the response after we load the json into memory. pro: don't have to worry about ack timeouts for long operations, con: if someone restarts rawls here the upsert is lost
```
helgridly, line 276: ...which leaves the upsert hung in `Upserting` forever, and the file left in the bucket.

we should have a plan to support resume-on-restart, even if the plan is "a ticket in the backlog"

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,55 +210,88 @@ class AvroUpsertMonitorActor(
...
+      upsertJson <- readUpsertObject(upsertFile)
```
helgridly: Not for this PR, but I did a little digging into streaming the json rather than downloading it.

According to [this StackOverflow](https://stackoverflow.com/a/50999067/2941784) it's possible to create a Java NIO `InputStream` from a bucket object. And [circe-fs2](https://github.com/circe/circe-fs2) claims to be able to create a stream from a JSON array-of-objects. There's probably a way to stitch these together, in which case you could grab a few and make a batch out of them.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
helgridly: residual logs from debugging in this file?

helgridly (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ansingh7115: yes :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: It's not being used anywhere else in this file. A method of the same name is, however, used in `HttpGooglePubSubDAO`, and that method is defined in that file.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockImportServiceDAO.scala
ansingh7115: copy pasta

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ansingh7115: nah - forgot to take them out

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
helgridly: I've really gotta stop reviewing with Ctrl+F. Sorry :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: Created this one: https://broadworkbench.atlassian.net/browse/AS-200

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: https://broadworkbench.atlassian.net/browse/AS-201

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: I agree - this type of method (getting a pet service account and then getting a token) is repeated in three other places in rawls: [WorkflowSubmissionActor](https://github.com/broadinstitute/rawls/blob/552baea34009bf5b56c399d92b32dc9e2586604f/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala#L307), [SubmissionMonitorActor](https://github.com/broadinstitute/rawls/blob/532491ad23174cea4fba8b2b0c5ec1cf753130bf/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala#L189), and [UserService](https://github.com/broadinstitute/rawls/blob/52027a74a0519f82fae4b2fb44ba331ae830185b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala#L223). I think it should be refactored, but am unsure of the best place to put it. If we put it in GoogleServicesDAO, we'd need to make it depend on a SamDAO, which doesn't feel right. We could make new util class?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: `publishMessageToUpdateImportStatus` calls `publishMessage` in `HttpGooglePubSubDAO` and that does a `retryWhen500OrGoogleError` [here](https://github.com/broadinstitute/rawls/blob/8b7e77cd1525bf804b1ec8705059d0e41102c011/google/src/main/scala/org/broadinstitute/dsde/rawls/google/HttpGooglePubSubDAO.scala#L85). Right now it would just error and then not do anything. If we are unable to write to the pub sub queue, what should happen?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpImportServiceDAO.scala
ansingh7115: Changed it to `case class ImportServiceResponse(id: String, status: String)`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
helgridly: util class ...or maybe the right answer is workbench-libs

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -38,34 +42,45 @@ object AvroUpsertMonitorSupervisor {
...
   final case class AvroUpsertMonitorConfig(pollInterval: FiniteDuration,
                                            pollIntervalJitter: FiniteDuration,
-                                           pubSubProject: GoogleProject,
-                                           pubSubTopic: String,
-                                           pubSubSubscription: String,
-                                           bucketName: String,
+                                           importRequestPubSubTopic: String,
+                                           importRequestPubSubSubscription: String,
+                                           updateImportStatusPubSubTopic: String,
+                                           updateImportStatusPubSubSubscription: String,
```
helgridly: you don't need this subscription

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -398,7 +400,8 @@ object Boot extends IOApp with LazyLogging {
           gcsDAO,
           samDAO,
           pubSubDAO,
-          avroUpsertPubSubDAO,
+          arrowPubSubDAO, // remove when cutting over to import service
+          importServiceDAO,
```
ansingh7115, line 404: We are passing in two pubSubDAOs for now

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -117,26 +139,36 @@ object AvroUpsertMonitor {
...
-                                  googleStorage: GoogleStorageService[IO],
-                                  pubSubDao: GooglePubSubDAO,
-                                  pubSubSubscriptionName: String,
-                                  avroUpsertBucketName: String,
-                                  batchSize: Int)(implicit cs: ContextShift[IO])
+                              val pollInterval: FiniteDuration,
+                              pollIntervalJitter: FiniteDuration,
+                              workspaceService: UserInfo => WorkspaceService,
+                              val googleServicesDAO: GoogleServicesDAO,
+                              val samDAO: SamDAO,
```
ansingh7115, line 158: these are vals because this class extends the `AuthUtil` trait

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -157,22 +189,15 @@ class AvroUpsertMonitorActor(
...
-          }
-        case _ =>
-          logger.info(s"found ignorable file $file message for job $jobId with ackId ${message.ackId} in message: $message")
-          acknowledgeMessage(message.ackId).map(_ => ImportComplete) pipeTo self //some other file (i.e. success/failure log, metadata.json)
+      // we received a message, so we will parse it and try to upsert it
+      logger.info(s"received async upsert message: $message")
+      if (message.attributes.contains("jobId")) {
+        importEntities(message)
+      } else {
+        arrowImportEntities(message)  // remove when cutting over to import service
```
ansingh7115, line 208: We decide whether to go down the CF path or the import service path based on whether the attributes contain a jobId. The CF solution has the metadata in the message body, while the import service has the metadata as attributes on the message

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -157,22 +189,15 @@ class AvroUpsertMonitorActor(
     case StartMonitorPass =>
       // start the process by pulling a message and sending it back to self
       pubSubDao.pullMessages(pubSubSubscriptionName, 1).map(_.headOption) pipeTo self
+      arrowPubSubDAO.pullMessages(arrowPubSubSubscriptionName, 1).map(_.headOption) pipeTo self
```
ansingh7115: At every pass, we pull messages from both topics.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -265,4 +344,71 @@ class AvroUpsertMonitorActor(
         Escalate
       }
     }
-}
+
```
ansingh7115, line 358: The old CF solution methods have been moved here, for easy deletion later.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
helgridly: How does this work? Does `pipeTo` place a message in the actor mailbox with the result of the computation or something? I'm trying to understand what happens if there's a message on both queues, since this is a mix of "two things sequentially" and "things that involve futures / actor messages", and I don't know the answer. Maybe a test to confirm that we process both messages if there's one on each pubsub?

helgridly (approved):
i have a big ole Q around polling two pubsubs at once, but other than that :+1:

core/src/main/scala/org/broadinstitute/dsde/rawls/util/AuthUtil.scala
```diff
@@ -0,0 +1,21 @@
...
+  val googleServicesDAO: GoogleServicesDAO
+
+  def getPetServiceAccountUserInfo(workspaceNamespace: String, userEmail: RawlsUserEmail)(implicit executionContext: ExecutionContext) = {
+    for {
+      petSAJson <- samDAO.getPetServiceAccountKeyForUser(workspaceNamespace, userEmail)
+      petUserInfo <- googleServicesDAO.getUserInfoUsingJson(petSAJson)
+    } yield {
+      petUserInfo
+    }
+  }
```
davidangb, line 19: would you file a ticket to update other parts of the Rawls code to use this trait, e.g. `SubmissionMonitorActor`?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -0,0 +1,188 @@
...
+
+    val blob = Await.result(googleStorage.unsafeGetBlobBody(bucketName, blobName).unsafeToFuture(), Duration.apply(10, TimeUnit.SECONDS))
+
+    // Publish message on the request topic
+    services.gpsDAO.publishMessages(importReadPubSubTopic, Seq(MessageRequest(sampleMessage, testAttributes(importId1))))
+
+    Thread.sleep(1000)
+
+    // check if correct message was posted on request topic
+    assert(services.gpsDAO.receivedMessage(importReadPubSubTopic, sampleMessage, 1))
```
davidangb: use `eventually` instead of a hardcoded `Thread.sleep`,  something  like:
```
    eventually(timeout(scaled(5000 milliseconds)), interval(scaled(250 milliseconds))) {
      assert(services.gpsDAO.receivedMessage(importReadPubSubTopic, sampleMessage, 1))  
    }
```
(apply this comment to all uses of `Thread.sleep`  in this test)

google/src/main/scala/org/broadinstitute/dsde/rawls/google/HttpGooglePubSubDAO.scala
```diff
@@ -110,9 +110,13 @@ class HttpGooglePubSubDAO(clientEmail: String,
       val messages = executeGoogleRequest(getPubSubDirectory.projects().subscriptions().pull(subscriptionToFullPath(subscriptionName), pullRequest)).getReceivedMessages
       if(messages == null)
         Seq.empty
-      else messages.asScala.map(message => PubSubMessage(message.getAckId,
-                                                         new String(message.getMessage.decodeData(), characterEncoding),
-                                                         message.getMessage.getAttributes.asScala.toMap))
+      else messages.asScala.map { message =>
+        val messageArray = message.getMessage.decodeData
+        val messageBody = if (messageArray == null) "" else new String(message.getMessage.decodeData(), characterEncoding)
```
davidangb: somewhat more Scala-idiomatic (and untested, of course!):
```
        val messageBody:String = Option(message.getMessage.decodeData()) match {
          case Some(decoded) => new String(decoded, characterEncoding)
          case None => ""
        }
```
or
```
Option(message.getMessage.decodeData()).map(new String(_, characterEncoding)).getOrElse("")
```

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -194,6 +194,28 @@ case class EntityCopyDefinition(
...
+    case "upserting" => Upserting
+    case "done" => Done
+    case "error" => Error
+    case _ => throw new RawlsException(s"invalid ImportStatus [${name}]")
+  }
+
+  case object ReadyForUpsert extends ImportStatus
+  case object Upserting extends ImportStatus
+  case object Done extends ImportStatus
+  //case class Error(message: String = "") extends ImportStatus
```
davidangb: nit: remove comment

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
davidangb: please add this as a code comment, for anyone reading the code in the future (even if the dual-approach isn't  long-lived)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
+        // end up with a race condition where multiple threads are attempting the same import / updating the status
+        // of the same import.
+        case Some(status) if status == ImportStatuses.ReadyForUpsert => {
+          publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Upserting, None)
+          toFutureTry(initUpsert(attributes.upsertFile, attributes.importId, message.ackId, attributes.workspace, petUserInfo)) map {
+            case Success(_) => markImportAsDone(attributes.importId, attributes.workspace, petUserInfo)
+            case Failure(t) => publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(t.getMessage))
+          }
+        }
+        case Some(status) => publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option("Import was not ready for upsert"))
```
davidangb: include the unexpected status in the error message for debugging: `s"Import was not ready for upsert: $status"`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
+    } yield ()
+  }
+
+
+  private def markImportAsDone(importId: UUID, workspaceName: WorkspaceName, userInfo: UserInfo) = {
+    for {
+      importStatus <- importServiceDAO.getImportStatus(importId, workspaceName, userInfo)
+      _ <- importStatus match {
+        case Some(status) if status == ImportStatuses.Upserting => publishMessageToUpdateImportStatus(importId, Option(status), ImportStatuses.Done, None)
+        case Some(status) => publishMessageToUpdateImportStatus(importId, Option(status), ImportStatuses.Error, Option("Upsert finished, but import status was not correct."))
```
davidangb: include the unexpected status in the error message

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
+        case None => publishMessageToUpdateImportStatus(importId, None, ImportStatuses.Error, Option("Import status not found"))
+      }
+    } yield ()
+  }
+
+  private def publishMessageToUpdateImportStatus(importId: UUID, currentImportStatus: Option[ImportStatus], newImportStatus: ImportStatus, errorMessage: Option[String]) = {
+    val updateImportStatus = UpdateImportStatus(importId.toString, newImportStatus.toString, currentImportStatus.map(_.toString), errorMessage)
+    val attributes = updateImportStatus.toJson.asJsObject.fields.map { case (attName, attValue) =>
+      // JsValue -> String, adds extra quotes for some reason
+      (attName, attValue.compactPrint.replace("\"","")) }
```
davidangb: how about:
```
    val attributes: Map[String, String] = updateImportStatus.toJson.asJsObject.fields.collect {
      case (attName:String, attValue:JsString) => (attName, attValue.value)
    }
```
this avoids the more-expensive String manipulation of compactPrint and replace

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
+      importStatus <- importServiceDAO.getImportStatus(importId, workspaceName, userInfo)
+      _ <- importStatus match {
+        case Some(status) if status == ImportStatuses.Upserting => publishMessageToUpdateImportStatus(importId, Option(status), ImportStatuses.Done, None)
+        case Some(status) => publishMessageToUpdateImportStatus(importId, Option(status), ImportStatuses.Error, Option("Upsert finished, but import status was not correct."))
+        case None => publishMessageToUpdateImportStatus(importId, None, ImportStatuses.Error, Option("Import status not found"))
+      }
+    } yield ()
+  }
+
+  private def publishMessageToUpdateImportStatus(importId: UUID, currentImportStatus: Option[ImportStatus], newImportStatus: ImportStatus, errorMessage: Option[String]) = {
```
davidangb, line 264: could you add a log message in here for debugging, a la:
```
logger.info(s"asking to change import job $importId from $currentImportStatus to $newImportStatus ${errorMessage.getOrElse("")}")
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
-      upsertResults <-
-        avroUpsertJson.grouped(batchSize).toList.traverse { upsertBatch =>
+      ws = workspaceService.apply(userInfo)
+      upsertResults <- {
+        upsertJson.grouped(batchSize).toList.traverse { upsertBatch =>
           logger.info(s"starting upsert for $jobId with ${upsertBatch.size} entities ...")
-          IO.fromFuture(IO(workspaceService.apply(petUserInfo).batchUpdateEntities(WorkspaceName(avroMetadataJson.namespace, avroMetadataJson.name), upsertBatch, true)))
-        }.unsafeToFuture
+          IO.fromFuture(IO(ws.batchUpdateEntities(workspaceName, upsertBatch, true)))
+        }.unsafeToFuture}
```
davidangb: for consideration ... it would be nice to know how many batches are in this job. It's extra runtime code; is it worth it for the log  message?
```
        val batches = upsertJson.grouped(batchSize).zipWithIndex.toList
        val numBatches = batches.size
        batches.traverse {
          case (upsertBatch, idx) =>
            logger.info(s"starting upsert $idx of $batchSize for $jobId with ${upsertBatch.size} entities ...")
            IO.fromFuture(IO(ws.batchUpdateEntities(workspaceName, upsertBatch, true)))
        }.unsafeToFuture
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -194,63 +219,116 @@ class AvroUpsertMonitorActor(
...
-  private def readUpsertObject(path: String): Future[Seq[EntityUpdateDefinition]] = {
-    readObject[Seq[EntityUpdateDefinition]](path, decompress = true)
-  }
 
-  private def readMetadataObject(path: String): Future[AvroMetadataJson] = {
-    readObject[AvroMetadataJson](path)
+  private def readUpsertObject(path: String) = {
+    val pathArray = path.split("/")
+    val bucketName = GcsBucketName(pathArray(0))
+    val blobName = GcsBlobName(pathArray(1))
```
davidangb: this would fail if we the path array has more than two parts.  Maybe  `GcsBlobName(pathArray.tail.mkString("/"))` ... or would we need `GcsBlobName(pathArray.tail.mkString("%2f"))`?

davidangb (commented):
comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/util/AuthUtil.scala
ansingh7115: Created: https://broadworkbench.atlassian.net/browse/AS-242

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: I tested by doing the following:

- Changed the config values for `pollInterval` and `pollJitter` so the monitor pass happens less frequently
- Added logging so that the actor logs when each monitor pass happens and at what time
- Stood up front and back rawls locally
- Called the old CF path by hitting the dev orch endpoint and the called the new import service path by making curl calls. Hit both of these endpoints many times very quickly (several times a second)
- Watched logs - looking at each block of logging between monitor passes

What I saw was that some of the monitor passes do receive messages from both the topics. 

Here's a slice of the log with one monitor pass - you'll see messages from both topics being recieved:
[Monitor Pass.txt](https://github.com/broadinstitute/rawls/files/4206383/Monitor.Pass.txt)

helgridly: Cool -- I learned something today :) Is this something you could do as a unit test?

ansingh7115: if i must

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: This would log for every single entity update - so if I uploaded a pfb with 5000 entities I would log 5000 lines. That would make the logs pretty ugly...
If you want to know the batch sizes - we could log that instead of every single thing in the batch

ansingh7115: Actually, this is fine, nvm, deleted my previous comment :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
ansingh7115: actually, I'm not sure - I think this would require having a monitor that just does 1 monitor pass and then stops.

ansingh7115:
jenkins retest

ansingh7115:
Swat tests fail because they are not using the correct fc-develop branch.
Here's a passing fiab run with the correct configs: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/30653/

helgridly (approved):

davidangb (commented):
one last outstanding comment about GCS paths, separators, and `tail` vs `pathArray(1)`. I think this could bite us if we change import service to use subfolders  - though if you know differently, let me know!

davidangb (approved):
:+1:

ansingh7115:
Swat test with correct fc-develop branch: 
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/30994/

PR #1169, opened 2020-01-29
AS-202: remove REST call from withWorkspaceContextAndPermissions db transaction

davidangb:
`withWorkspaceContextAndPermissions` is used in 36 places, and it contains an external REST call to Sam inside a database transaction. This PR refactors that approach - it makes one small  transaction to look up the workspace from the DB, then calls Sam outside of a transaction, and then continues on with pre-existing code.

This PR does ***NOT*** even try to remove _all_ REST calls inside transactions. I'm only focused on `withWorkspaceContextAndPermissions`, to keep scope and PR review small and manageable.

The implementation of this PR is:
* replace `withWorkspaceContextAndPermissions` loan-pattern function with the pre-existing  `getWorkspaceContextAndPermissions` that just returns a Future
* modify all calling code to ask for the `getWorkspaceContextAndPermissions` outside of a db transaction

-----

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -427,13 +427,13 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
   def updateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]): Future[PerRequestMessage] = {
     withAttributeNamespaceCheck(operations.map(_.name)) {
       for {
-        workspaceContext <- getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write)
-        updatedWorkspace <- dataSource.inTransaction({ dataAccess =>
-            updateWorkspace(operations, dataAccess)(workspaceContext)
+        ctx <- getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write)
+        workspace <- dataSource.inTransaction({ dataAccess =>
+            updateWorkspace(operations, dataAccess)(ctx)
         }, TransactionIsolation.ReadCommitted) // read committed to avoid deadlocks on workspace attr scratch table
-        authDomain <- loadResourceAuthDomain(SamResourceTypeNames.workspace, updatedWorkspace.workspaceId, userInfo)
+        authDomain <- loadResourceAuthDomain(SamResourceTypeNames.workspace, workspace.workspaceId, userInfo)
       } yield {
-        RequestComplete(StatusCodes.OK, WorkspaceDetails(updatedWorkspace, authDomain))
+        RequestComplete(StatusCodes.OK, WorkspaceDetails(workspace, authDomain))
       }
     }
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -200,7 +200,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     span.setStatus(Status.OK)
     span.end()
 
-    traceWithParent("withWorkspaceContextAndPermissionsF", parentSpan)(s1 => withWorkspaceContextAndPermissionsF(workspaceName, SamWorkspaceActions.read, Option(attrSpecs)) { workspaceContext =>
+    traceWithParent("getWorkspaceContextAndPermissions", parentSpan)(s1 => getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Option(attrSpecs)) flatMap { workspaceContext =>
       dataSource.inTransaction { dataAccess =>
 
         // maximum access level is required to calculate canCompute and canShare. Therefore, if any of
@@ -286,7 +286,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
   }
 
   def getBucketOptions(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
-    withWorkspaceContextAndPermissionsF(workspaceName, SamWorkspaceActions.read) { workspaceContext =>
+    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read) flatMap { workspaceContext =>
       dataSource.inTransaction { dataAccess =>
         DBIO.from(gcsDAO.getBucketDetails(workspaceContext.workspace.bucketName, RawlsBillingProjectName(workspaceContext.workspace.namespace))) map { details =>
           RequestComplete(StatusCodes.OK, details)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -565,8 +560,8 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     val (libraryAttributeNames, workspaceAttributeNames) = destWorkspaceRequest.attributes.keys.partition(name => name.namespace == AttributeName.libraryNamespace)
     withAttributeNamespaceCheck(workspaceAttributeNames) {
       withLibraryAttributeNamespaceCheck(libraryAttributeNames) {
-        dataSource.inTransaction({ dataAccess =>
-          withWorkspaceContextAndPermissions(sourceWorkspaceName, SamWorkspaceActions.read, dataAccess) { sourceWorkspaceContext =>
+        withWorkspaceContextAndPermissionsF(sourceWorkspaceName, SamWorkspaceActions.read) { sourceWorkspaceContext =>
+          dataSource.inTransaction({ dataAccess =>
```
davidangb, line 568: this is the general pattern in this PR: switch to the Future-ified version `withWorkspaceContextAndPermissionsF`,  then swap the ordering of `withWorkspaceContextAndPermissionsF` and  `dataSource.inTransaction` in the calling code.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -346,11 +346,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
   }
 
   def getWorkspaceContextAndPermissions(workspaceName: WorkspaceName, requiredAction: SamResourceAction): Future[SlickWorkspaceContext] = {
-    dataSource.inTransaction { dataAccess =>
-      withWorkspaceContextAndPermissions(workspaceName, requiredAction, dataAccess) { workspaceContext =>
-        DBIO.successful(workspaceContext)
-      }
-    }
+    withWorkspaceContextAndPermissionsF(workspaceName, requiredAction) { Future(_) }
   }
```
davidangb, line 351: This `getWorkspaceContextAndPermissions` method checks permissions, but doesn't return them. We probably want to rename it to something more accurate. At least for this draft, I'm not renaming so as to keep code changes  small.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2129,6 +2123,16 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+  private def withWorkspaceContextAndPermissionsF[T](workspaceName: WorkspaceName, requiredAction: SamResourceAction)(op: (SlickWorkspaceContext) => Future[T]): Future[T] = {
+    for {
+      workspaceContext <- dataSource.inTransaction { dataAccess =>
+        withWorkspaceContext(workspaceName, dataAccess) { ctx => DBIO.successful(ctx) }
+      }
+      opResult <- requireAccessF(workspaceContext.workspace, requiredAction) { op(workspaceContext) }
+    } yield opResult
+  }
+  // TODO: once everything is switched over to the Future-ified version, look for remaining calls to requireAccess and remove those too
+
```
davidangb: compare this Future-ified version to the `withWorkspaceContextAndPermissions` just previous to it

helgridly:
+1 for removing a bunch of badness from our codebase!

Q: Now that it's just a future, why does it still need to be a loan-pattern `withX` style function? There's no cleanup that needs to happen after the op, so it could just return `Future[SlickWorkspaceContext]`.

Rambling philosophical thoughts follow, please disregard:

1. How might we avoid this problem in future? It's easy to write and (can be) hard to extract ourselves from. Maybe making it clearer that this function opens a txn, but you might say that's evident from the signature, and it's easy to call another function and lose that context. Maybe something in our calling-external-services libraries that looks to see if we're in a txn -- but in our current architecture that'd require looking back over multiple execution context changes (would be relatively easy with a one-thread-per-request model though).

2. There is a little voice in my head that says "but how do we ensure that the permission is retained over the entire duration of the operation", but I don't buy that what we have today achieves that either -- or whether that's even possible without a change in approach to security that's more philosophical than I'm inclined to consider right now. I'd be curious to learn about ways to do this if they're obvious or well-known, though.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -200,100 +200,96 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     span.setStatus(Status.OK)
     span.end()
 
-    dataSource.inTransaction { dataAccess =>
-      traceDBIOWithParent("withWorkspaceContext", parentSpan)(s1 => withWorkspaceContext(workspaceName, dataAccess, Option(attrSpecs)) { workspaceContext =>
-        traceDBIOWithParent("requireAccess", s1)(s2 => requireAccess(workspaceContext.workspace, SamWorkspaceActions.read) {
-
```
davidangb, line 206: github UI shows a big diff in this file, but a lot of it is indentation changes. You may want to ignore whitespace.

davidangb:
@helgridly what's your opinion on the code comments below?
```
  def getBucketOptions(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
    // ******************************
    // is this construct:
    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read) flatMap { workspaceContext =>
    // better than this one?
    // withWorkspaceContextAndPermissionsF(workspaceName, SamWorkspaceActions.read) { workspaceContext =>
    // ******************************
        DBIO.from(gcsDAO.getBucketDetails(workspaceContext.workspace.bucketName, RawlsBillingProjectName(workspaceContext.workspace.namespace))) map { details =>
          RequestComplete(StatusCodes.OK, details)
        }
      }
    }
  }
```

helgridly:
IMO yes, because a pure-future version can be used in a [flat]map or a for-comp, whereas loan pattern is "loan pattern or bust"

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1986,6 +1981,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
+  // TODO: find and assess all usages. This is written to reside inside a DB transaction, but it makes external REST calls.
```
davidangb, line 1990: I will create Jira tickets for these TODOs

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -431,10 +428,9 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
   def updateWorkspace(workspaceName: WorkspaceName, operations: Seq[AttributeUpdateOperation]): Future[PerRequestMessage] = {
     withAttributeNamespaceCheck(operations.map(_.name)) {
       for {
+        ctx <- getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write)
         workspace <- dataSource.inTransaction({ dataAccess =>
-          withWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, dataAccess) {
-            updateWorkspace(operations, dataAccess)
-          }
+            updateWorkspace(operations, dataAccess)(ctx)
```
davidangb: by changing the transaction boundaries, this opens up the small chance of a race condition. We get the workspace in transaction A, then update the workspace in transaction B. Between those transactions, some other user could sneak in and perform a different update, which could get overwritten in transaction  B. Is this an acceptable tradeoff? Is it worth protecting against?

breilly2: (Drive-by comment because I like thinking about this type of thing)

In a lot of cases, last-in-wins can be fine. I can think of two cases where it's not:

1. If other user's decision to modify is based on information that the first user is modifying which invalidates other user's decision
2. If only the changed fields are updated in such a way that two users making separate updates could leave the entity in an inconsistent/invalid state

That said, if we're not doing optimistic locking with a rev column or something like that (I don't remember if we are) then we're already vulnerable to these problems. So the answer to "is it worth protecting against" depends on the fields that can be modified and whether this specific type of entity is vulnerable to either of these two scenarios. Can you think of any other potentially dangerous scenarios?

davidangb:
rawls-build failed in Jenkins, looks like it may be a transient dockerhub issue. Jenkins retest.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: note the last commit - 7f5b533e841ca70b7c6befe9d32b7d69749e7069 - in which I've re-retrieved the workspace for `updateLibraryAttributes`, `updateWorkspace`, and `cloneWorkspace` to avoid the race condition.

For these three APIs, this trades an extra db query to re-get the workspace for removing the Sam call from a transaction. I think this is worth it ... do you?

helgridly: Yes

helgridly (approved):

MatthewBemis (approved):

PR #1171, opened 2020-02-13
Add "method config deleted" flag to submissions list entries [WA-7]

aednichols:
This allows the UI to disable the "relaunch failures" button for those entries.


kshakir (commented):
Nice! I'm curious, there's a ton of discussion in WA-7. Which parts are addressed by this PR, for example:
- Will this flag be a signal to the Terra-UI to fix-in-the-browser the mangled names?
- Should we create separate tickets for Terra-UI's hide-deleted-unless-requested-for-CLIA-users, or is that part WA-7's A/C?

aednichols:
1. Sure, I can add that to my Terra PR (linked). It’s fully deterministic, if it’s deleted then chop off the 11 characters we added.
2. The scope of this ticket is to remove a button that doesn’t work, and maybe undo some name mangling. Even if we selectively leave the name mangling in place for CLIA users, there is absolutely nothing they can do with the method config, it is marked deleted in the database and can never run.

kshakir (approved):

salonishah11 (approved):

PR #1179, opened 2020-02-25
[CA-540] Step 3/3 Support delete billing project in Rawls

yonghaoy:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3323,7 +3323,7 @@ paths:
         - billing
         - users
       summary: delete billing project
-      operationId: deleteBillingProjects
+      operationId: deleteBillingProject
       security:
         - authorization:
             - openid
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3296,7 +3296,39 @@ paths:
             - openid
             - email
             - profile
-
+    delete:
+      responses:
+        '200':
+          description: Successfully delete billing project
+        '403':
+          description: You must be a project owner to delete billing project
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        '400':
+          description: Workspace Exists in this billing project
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        '500':
+          description: Rawls Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      parameters:
+        - name: projectName
+          description: Name of the billing project
+          required: true
+          type: string
+          in: path
+      description: delete billing project
+      tags:
+        - billing
+        - users
+      summary: delete billing project
+      operationId: deleteBillingProjects
+      security:
+        - authorization:
+            - openid
+            - email
+            - profile
   '/api/user/billingAccounts':
     get:
       responses:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3305,7 +3305,7 @@ paths:
           schema:
             $ref: '#/definitions/ErrorReport'
         '400':
-          description: Workspace Exists in this billing project
+          description: Project cannot be deleted because it contains workspaces.
           schema:
             $ref: '#/definitions/ErrorReport'
         '500':
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -203,8 +203,8 @@ trait WorkspaceComponent {
       loadWorkspaces(findByIdsQuery(workspaceIds), attributeSpecs)
     }
 
-    def listByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] = {
-      loadWorkspaces(findByNamespaceQuery(namespaceName))
+    def countByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Int] = {
+      findByNamespaceQuery(namespaceName).size.result
     }
 
     def delete(workspaceName: WorkspaceName): ReadWriteAction[Boolean] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -203,6 +203,10 @@ trait WorkspaceComponent {
       loadWorkspaces(findByIdsQuery(workspaceIds), attributeSpecs)
     }
 
+    def listByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] = {
+      loadWorkspaces(findByNamespaceQuery(namespaceName))
+    }
+
     def delete(workspaceName: WorkspaceName): ReadWriteAction[Boolean] = {
       uniqueResult[WorkspaceRecord](findByNameQuery(workspaceName)).flatMap {
         case None => DBIO.successful(false)
@@ -355,6 +359,10 @@ trait WorkspaceComponent {
 
     def findByIdsQuery(workspaceIds: Seq[UUID]): WorkspaceQueryType = {
       filter(_.id.inSetBind(workspaceIds))
+    }
+
+    private def findByNamespaceQuery(namespaceName: RawlsBillingProjectName): WorkspaceQueryType = {
+      filter(rec => (rec.namespace === namespaceName.value))
     }
 
     private def loadWorkspace(lookup: WorkspaceQueryType, attributeSpecs: Option[WorkspaceAttributeSpecs] = None): ReadAction[Option[Workspace]] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -228,7 +228,9 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
       projectUsers <- samDAO.listAllResourceMemberIds(SamResourceTypeNames.billingProject, projectName.value, ownerUserInfo)
       _ <- projectUsers.toList.traverse(destroyPet(_, projectName))
       _ <- samDAO.deleteResource(SamResourceTypeNames.billingProject, projectName.value, ownerUserInfo)
-      _ <- dataSource.inTransaction { dataAccess => dataAccess.rawlsBillingProjectQuery.delete(projectName) }
+      _ <- dataSource.inTransaction { dataAccess =>
+        dataAccess.rawlsBillingProjectQuery.delete(projectName)
+      }
     } yield {
       RequestComplete(StatusCodes.NoContent)
     }
@@ -253,16 +255,16 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def deleteBillingProject(projectName: RawlsBillingProjectName): Future[PerRequestMessage] = {
     for {
       // Check if workspace exists for this project.
-     workspaces <- dataSource.inTransaction { dataAccess =>
-        dataAccess.workspaceQuery.listByNamespace(projectName)
+      workspaceCount <- dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.countByNamespace(projectName)
       }
 
-     _<- if(!workspaces.isEmpty) {
-       Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, "There are workspaces exists in this project.")))
+      _<- if(workspaceCount > 0) {
+       Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, "Project cannot be deleted because it contains workspaces.")))
      } else {
       Future.successful(())
      }
-     // unregister then delete actual project in google
+      // unregister then delete actual project in google
       _<- unregisterBillingProjectWithUserInfo(projectName, userInfo)
       _<- gcsDAO.deleteProject(projectName)
     } yield RequestComplete(StatusCodes.NoContent)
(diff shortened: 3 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -225,14 +242,33 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     } yield ()
   }
 
-  def deleteBillingProject(projectName: RawlsBillingProjectName, ownerInfo: Map[String, String]): Future[PerRequestMessage] = {
+  def adminDeleteBillingProject(projectName: RawlsBillingProjectName, ownerInfo: Map[String, String]): Future[PerRequestMessage] = {
     // unregister then delete actual project in google
+    val ownerUserInfo = UserInfo(RawlsUserEmail(ownerInfo("newOwnerEmail")), OAuth2BearerToken(ownerInfo("newOwnerToken")), 3600, RawlsUserSubjectId("0"))
     for {
-      _ <- unregisterBillingProject(projectName, ownerInfo)
+      _ <- unregisterBillingProjectWithUserInfo(projectName, ownerUserInfo)
```
marctalbott: I think we should either call `unregisterBillingProjectWithOwnerInfo` and remove the `ownerUserInfo` val above, or remove `unregisterBillingProjectWithOwnerInfo` entirely since at the moment we aren't calling it at all

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -225,14 +242,33 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
 
+  def deleteBillingProject(projectName: RawlsBillingProjectName): Future[PerRequestMessage] = {
+    for {
+      // Check if workspace exists for this project.
+     workspaces <- dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listByNamespace(projectName)
+      }
+
+     _<- if(!workspaces.isEmpty) {
+       Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, "There are workspaces exists in this project.")))
```
marctalbott: I'm not sure that this is the right status code to return in this case. My mental model for 403 Forbidden is when the user doesn't have the correct permissions to perform an action. In this case, the user has the correct permissions, but can't delete the billing project because there are associated workspaces. Maybe a 400 Bad Request would be appropriate here? I'm definitely open to keeping it as a 403 if you think that's better though

yonghaoy: https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala#L96
Is the error where it comes from

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
yonghaoy: unregisterBillingProjectWithOwnerInfo is used by AdminUnregisterBillingProject, and AdminUnregisterBillingProject do pass into the ownerInfo

marctalbott: Ah, I see that now 👍. Is there a reason not to replace this call with `unregisterBillingProjectWithOwnerInfo` also?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: Sure, but in this case `requireProjectAction` returned true, but we are now rejecting the request because there's still a workspace in the project. The reason the call to delete the billing project is failing isn't because the caller is lacking permissions, rather it's because the caller hasn't yet deleted all the necessary billing projects

yonghaoy: Ah, sorry, I thought you are asking the line of error when user doesn't have permission. For workspace already exists, I also think 400 is better. Thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
yonghaoy: All the actual unregister logic goes to unregisterBillingProjectWithUserInfo(input is userInfo). 

When WithOwnerInfo calls unregisterBillingProjectWithUserInfo, WithOwnerInfo will convert map of ownerInfo into UserInfo. 
And for this method, it will just call unregisterBillingProjectWithUserInfo directly since we have UserInfo at hand. 

This do sounds confusion, any suggestion I can improve or any better name for those methods?  Thanks

marctalbott: My vote would be to call `unregisterBillingProjectWithOwnerInfo` to reduce code duplication (lines 247 and 216), but it's such a small difference so really your call.

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -225,14 +242,32 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
 
+  def deleteBillingProject(projectName: RawlsBillingProjectName): Future[PerRequestMessage] = {
+    for {
+      // Check if workspace exists for this project.
+     workspaces <- dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listByNamespace(projectName)
+      }
+
+     _<- if(!workspaces.isEmpty) {
+       Future.failed(new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, "There are workspaces exists in this project.")))
```
dvoet: Change error message to: "Project cannot be deleted because it contains workspaces."

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -203,6 +203,10 @@ trait WorkspaceComponent {
       loadWorkspaces(findByIdsQuery(workspaceIds), attributeSpecs)
     }
 
+    def listByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] = {
```
dvoet: we only need the count of workspaces

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -71,6 +85,7 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
     withEmptyTestDatabase { dataSource: SlickDataSource =>
       val project = defaultBillingProject.copy(googleProjectNumber = None)
       runAndWait(rawlsBillingProjectQuery.create(project))
+      runAndWait(workspaceQuery.save(workspace))
```
dvoet: why add this line?

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -149,4 +164,40 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
...
+
+      val mockSamDAO = mock[SamDAO]
+      when(mockSamDAO.userHasAction(SamResourceTypeNames.billingProject, project.projectName.value, SamBillingProjectActions.deleteBillingProject, userInfo)).thenReturn(Future.successful(true))
+
+      val userService = getUserService(dataSource, mockSamDAO)
+
+      val actual = userService.DeleteBillingProject(defaultBillingProjectName).failed.futureValue
+      assert(actual.isInstanceOf[RawlsExceptionWithErrorReport])
+      actual.asInstanceOf[RawlsExceptionWithErrorReport].errorReport.statusCode shouldEqual Option(StatusCodes.BadRequest)
```
dvoet: try this (and below too):
```
val actual = intercept[RawlsExceptionWithErrorReport] {
  userService.DeleteBillingProject(defaultBillingProjectName).futureValue
}
actual.errorReport.statusCode shouldEqual Option(StatusCodes.BadRequest)
```
intercept is a little cleaner for for expecting an exception

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -31,6 +32,19 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
...
+  val workspace = Workspace(
+    defaultBillingProjectName.value,
+    testData.wsName.name,
+    UUID.randomUUID().toString,
+    "aBucket",
+    Some("workflow-collection"),
+    currentTime(),
+    currentTime(),
+    "test",
+    Map.empty
```
dvoet: I think this is only required in `fail with a 400 when workspace exists in this billing project to be deleted`, define it in that function, not at the class level

dvoet (changes requested):
also add code to UserApiService and swagger

dvoet:
also add success case unit test in UserServiceSpec and api level unit tests in UserApiServiceSpec

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
yonghaoy: Done

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
yonghaoy: We want to test when there is a workspace exists. This line will insert a workspace with the billing project name.

yonghaoy: Removed

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
yonghaoy: Thanks!

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
yonghaoy: Done

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
yonghaoy: should start with Await.result(....

yonghaoy:
Jenkins retest

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3296,7 +3296,39 @@ paths:
...
+          description: Name of the billing project
+          required: true
+          type: string
+          in: path
+      description: delete billing project
+      tags:
+        - billing
+        - users
+      summary: delete billing project
+      operationId: deleteBillingProjects
```
dvoet: `deleteBillingProject` singular (no s at the end)

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -149,4 +151,81 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
...
+
+      val mockGcsDAO = mock[GoogleServicesDAO]
+      when(mockGcsDAO.getUserInfoUsingJson(petSAJson)).thenReturn(Future.successful(userInfo))
+      when(mockGcsDAO.deleteProject(project.projectName)).thenReturn(Future.successful())
+
+      val userService = getUserService(dataSource, mockSamDAO, gcsDAO = mockGcsDAO)
+      val actual = userService.DeleteBillingProject(defaultBillingProjectName).futureValue
+
+      verify(mockGcsDAO).deleteProject(project.projectName)
+      actual shouldEqual RequestComplete(StatusCodes.NoContent)
```
dvoet, line 182: also assert that the record has been removed from the rawls database and verify mockSamDAO.deleteResource was called

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -149,4 +151,81 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
...
+    }
+  }
+
+  // 403 when workspace exists in this billing project to be deleted
```
dvoet: remove this comment, it is wrong and the test name says it all anyway

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -149,4 +151,81 @@ class UserServiceSpec extends FlatSpecLike with TestDriverComponent with Mockito
...
+      val userService = getUserService(dataSource, mockSamDAO)
+
+      val actual = intercept[RawlsExceptionWithErrorReport] {
+        Await.result(userService.DeleteBillingProject(defaultBillingProjectName), Duration.Inf).asInstanceOf[RequestComplete[ErrorReport]]
+      }
+      actual.errorReport.statusCode shouldEqual Option(StatusCodes.BadRequest)
+    }
+  }
+
+  // 404 when user doesn't have permission to delete billing project.
```
dvoet: remove this comment, it is wrong and the test name says it all anyway

dvoet (approved):
👍 after these little things are addressed

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
yonghaoy: Done

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

yonghaoy:
Jenkins retest

PR #1174, opened 2020-02-13
New request fields must be defaulted or optional. WA-98 QA-1031

kshakir:
- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [ ] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
  [alpha](https://console.cloud.google.com/sql/instances/terraform-r4caezzc35c4tb7pgdhwkmme4y/backups?project=broad-dsde-alpha&organizationId=548622027621)
  DBs in Google Cloud Console
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] Tell your tech lead (TL) that the PR exists if they want to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Assign to submitter** to finalize
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Squash commits and merge to develop
[...]

PR #1182, opened 2020-02-28
Allow Rawls to accept null entityType, entityName [WA-135]

aednichols:
Fixes a subtle regression introduced in https://github.com/broadinstitute/rawls/pull/1174

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
@@ -375,12 +375,13 @@ class ExecutionJsonSupport extends JsonSupport {
         // so in that project one can just adjust the function call to jsonFormat*N*.
         methodConfigurationNamespace = fields("methodConfigurationNamespace").convertTo[String],
         methodConfigurationName = fields("methodConfigurationName").convertTo[String],
-        entityType = fields.get("entityType").map(_.convertTo[String]),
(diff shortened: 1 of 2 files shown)


grsterin (approved):

rushtong (approved):
👍

aednichols:
Thank you @rushtong !

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -375,12 +375,13 @@ class ExecutionJsonSupport extends JsonSupport {
         // so in that project one can just adjust the function call to jsonFormat*N*.
         methodConfigurationNamespace = fields("methodConfigurationNamespace").convertTo[String],
         methodConfigurationName = fields("methodConfigurationName").convertTo[String],
-        entityType = fields.get("entityType").map(_.convertTo[String]),
-        entityName = fields.get("entityName").map(_.convertTo[String]),
-        expression = fields.get("expression").map(_.convertTo[String]),
+        // We are flatMapping these because `null` is a valid input provided by the UI
```
breilly2, line 378: Sorry I'm a little late to this, but maybe it's a bug that the UI is sending `null`s? I have not investigated anything yet but I can if you think there may be a UI problem, @aednichols .

aednichols: We're talking the FireCloud UI, which didn't change within the timespan when the bug appeared

PR #1183, opened 2020-03-02
[CA-702] enable requester pays for linked bond service accounts

andy7i:
https://broadworkbench.atlassian.net/browse/CA-702

There are 3 things going on here:
1. apis to enable/disable requester pays within a workspace for linked service accounts including access control
1. updates wot workspace access control update logic to revoke RP when user lose access to workspaces
1. new database table and related logic to keep track of which workspace users are enabled in

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


dvoet:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -172,11 +172,39 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
     */
   def cleanupDMProject(projectName: RawlsBillingProjectName): Future[Unit]
 
+  /**
```
dvoet, line 173: changes in this file and HttpGoogleServicesDAO make removePolicyBindings work the same add with respect to retries and avoiding policy modifications if there are no actual changes

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3107,6 +3107,82 @@ paths:
...
+              - openid
+              - email
+              - profile
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/disableRequesterPaysForLinkedServiceAccounts':
+    put:
+      responses:
+        '204':
+          description: Successful Request
+        '403':
+          description: Insufficient permissions to enable Requester Pays for workspace.
```
gpolumbo-broad:
```suggestion
          description: Insufficient permissions to disable Requester Pays for workspace.
```

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3107,6 +3107,82 @@ paths:
           - openid
           - email
           - profile
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/enableRequesterPaysForLinkedServiceAccounts':
+    put:
+      responses:
+        '204':
+          description: Successful Request
+        '403':
+          description: Insufficient permissions to enable Requester Pays for workspace.
```
gpolumbo-broad: I think this comment could be better.  Can we be more specific about what permission is missing?  If a user makes this request and hits this permissions error, it's probably because they do not have permissions to add users to the Workspace's billing group.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3107,6 +3107,82 @@ paths:
...
+            $ref: '#/definitions/ErrorReport'
+        '404':
+          description: Workspace Not Found
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        '500':
+          description: Rawls Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        description: Enable Requester Pays for the user's linked service accounts in a workspace
```
gpolumbo-broad: I think this is the place where we can and should get pretty descriptive.  Will take all of your active Bond linked Service Accounts and add them to the right group so that they have permissions to incur cost on the workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -172,11 +172,39 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+    * Removes the IAM policies to the project's existing policies
```
gpolumbo-broad:
```suggestion
    * Removes the IAM policies from the project's existing policies
```

dvoet:
I have tested this manually

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpBondApiDAO.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.broadinstitute.dsde.rawls.util.{HttpClientUtilsStandard, Retry}
+
+import scala.concurrent.{ExecutionContext, Future}
+import BondJsonSupport._
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+
+case class BondServiceAccountEmail(client_email: String)
+case class BondResponseData(data: BondServiceAccountEmail)
+
+case class Providers(providers: List[String])
```
gpolumbo-broad, line 19: Seems like wrapping the `List[String]` into a class is unnecessary

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -996,7 +972,7 @@ class HttpGoogleServicesDAO(
   }
 
   override def removeRoleFromGroup(projectName: RawlsBillingProjectName, groupEmail: WorkbenchEmail, role: String): Future[Unit] = {
-    removePolicyBindings(projectName, Map(s"roles/$role" -> Seq(s"group:${groupEmail.value}")))
+    removePolicyBindings(projectName, Map(s"roles/$role" -> Set(s"group:${groupEmail.value}"))).void
```
gpolumbo-broad: Why the addition of the call to `.void` here?

core/src/main/resources/swagger/api-docs.yaml
andy7i:
```suggestion
          description: Insufficient permissions to disable Requester Pays for workspace.
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -172,11 +172,39 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+    * @return true if the policy was actually changed
+    */
+  def removePolicyBindings(projectName: RawlsBillingProjectName, policiesToRemove: Map[String, Set[String]]): Future[Boolean] = updatePolicyBindings(projectName) { existingPolicies =>
```
andy7i, line 177: I think I'm not understanding how `existingPolicies` is set. Would you mind explaining this some time?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -924,40 +924,19 @@ class HttpGoogleServicesDAO(
...
+    * 2) call updatePolicies
+    * 3) if updated policies are the same as existing policies return false, don't call google
+    * 4) if updated policies are different than existing policies update google and return true
+    *
+    * @param projectName google project name
+    * @param updatePolicies function (existingPolicies => updatedPolicies). May return policies with no members
+    *                       which will be handled appropriately when sent to google.
+    * @return true if google was called to update policies, false otherwise
+    */
+  override protected def updatePolicyBindings(projectName: RawlsBillingProjectName)(updatePolicies: Map[String, Set[String]] => Map[String, Set[String]]): Future[Boolean] = {
```
andy7i, line 929: intellij didn't find any uses of this overridden def (but found 2 uses of the base implementation). Is this being used in some other scala-magic-y way?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: why does this line end in `.void` but `addRoleToGroup` does not?:
```
  override def addRoleToGroup(projectName: RawlsBillingProjectName, groupEmail: WorkbenchEmail, role: String): Future[Boolean] = {
    addPolicyBindings(projectName, Map(s"roles/$role" -> Set(s"group:${groupEmail.value}")))
  }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: follow-on question: would this be used in a way similar to the way `updatePolicyBindings` is used in `GoogleServicesDAO.scala`, where `addPolicyBindings` calls `updatePolicyBindings` with an implementation of `(updatePolicies: Map[String, Set[String]] => Map[String, Set[String]])`? In other words, to use `updatePolicyBindings` here in `HttpGoogleServicesDAO.scala` we would define `addPolicyBindings` in this file in a similar way?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -68,6 +68,7 @@ object SamWorkspaceActions {
 object SamBillingProjectActions {
   val createWorkspace = SamResourceAction("create_workspace")
   val launchBatchCompute = SamResourceAction("launch_batch_compute")
+  val launchNotebookCluster = SamResourceAction("launch_notebook_cluster")
```
andy7i: just checking -- is this related to the ticket or just hitching a ride on this PR?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupService.scala
```diff
@@ -0,0 +1,56 @@
...
+    } yield {
+      emails
+    }
+  }
+
+  def revokeUserFromWorkspace(userEmail: RawlsUserEmail, workspaceName: WorkspaceName): Future[Seq[BondServiceAccountEmail]] = {
+    for {
+      emails <- dataSource.inTransaction { dataAccess => dataAccess.workspaceRequesterPaysQuery.listAllForUser(workspaceName, userEmail) }
+      _ <- revokeEmails(emails.toSet, userEmail, workspaceName)
+    } yield emails.map(BondServiceAccountEmail)
```
andy7i, line 36: funky syntax for typecasting each email string into a `BondServiceAccountEmail` (assuming that is what it's doing)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  object workspaceRequesterPaysQuery extends TableQuery(new WorkspaceRequesterPaysTable(_)) {
+    def insertAllForUser(workspaceName: WorkspaceName, userEmail: RawlsUserEmail, serviceAccountEmails: Set[BondServiceAccountEmail]): ReadWriteAction[Int] = {
+      for {
+        maybeWorkspaceId <- workspaceQuery.getWorkspaceId(workspaceName)
+        workspaceId = maybeWorkspaceId.getOrElse(throw new RawlsException(s"workspace not found $workspaceName"))
+        existingRecords <- existingRecordsForUserQuery(workspaceName, userEmail).result
+        saEmailsToInsert = serviceAccountEmails.map(_.client_email) -- existingRecords.map(_.serviceAccountEmail)
+        results <- workspaceRequesterPaysQuery ++= saEmailsToInsert.map(saEmail => WorkspaceRequesterPaysRecord(0, workspaceId, userEmail.value, saEmail))
```
andy7i, line 32: `WorkspaceRequesterPaysRecord(0,`... is this setting the `id` to 0? or does it do something else with it?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
```diff
@@ -0,0 +1,57 @@
...
+      } yield results.getOrElse(saEmailsToInsert.size) // results seems to be None when there are multiple inserts :(
+    }
+
+    def deleteAllForUser(workspaceName: WorkspaceName, userEmail: RawlsUserEmail): ReadWriteAction[Int] = {
+      existingRecordsForUserQuery(workspaceName, userEmail).delete
+    }
+
+    def userExistsInWorkspaceNamespace(namespace: String, userEmail: RawlsUserEmail): ReadAction[Boolean] = {
```
andy7i, line 40: not entirely sure I fully understand this function. It looks like it does a query to get any workspaces that also have requesterPays referencing that workspace, and then checks if any of those workspaces are in the given `namespace` and if the given `userEmail` exists as a requesterPays entry. So it seems like a check to see if the given user has any requesterPays entries in the given namespace?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    def listAllForUser(workspaceName: WorkspaceName, userEmail: RawlsUserEmail): ReadAction[Seq[String]] = {
+      existingRecordsForUserQuery(workspaceName, userEmail).map(_.serviceAccountEmail).result
+    }
+  }
+
+  private def existingRecordsForUserQuery(workspaceName: WorkspaceName, userEmail: RawlsUserEmail): Query[WorkspaceRequesterPaysTable, WorkspaceRequesterPaysRecord, Seq] = {
+    val workspaceSubquery = workspaceQuery.filter(ws => ws.namespace === workspaceName.namespace && ws.name === workspaceName.name).map(_.id)
+    workspaceRequesterPaysQuery.filter(_.workspaceId in workspaceSubquery).filter(_.userEmail === userEmail.value)
```
andy7i, line 55: the `_`s still get me good

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupServiceSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+    // add user to 2 workspaces in same namespace
+    runAndWait(workspaceRequesterPaysQuery.insertAllForUser(minimalTestData.workspace.toWorkspaceName, userInfo.userEmail, Set(expectedEmail)))
+    runAndWait(workspaceRequesterPaysQuery.insertAllForUser(minimalTestData.workspace2.toWorkspaceName, userInfo.userEmail, Set(expectedEmail)))
+    runAndWait(workspaceRequesterPaysQuery.userExistsInWorkspaceNamespace(minimalTestData.billingProject.projectName.value, userInfo.userEmail)) shouldBe true
+
+    // add user to mock google bindings
+    val initialBindings = Map(service.requesterPaysRole -> Set("serviceAccount:" + expectedEmail.client_email))
+    service.googleServicesDAO.asInstanceOf[MockGoogleServicesDAO].policies.put(minimalTestData.billingProject.projectName, initialBindings)
+
+    // remove use from 1 workspace and check that it did not get removed from google bindings
```
andy7i, line 80: confused me at first but it doesn't get removed from google bindings since the user still has requester pays in another workspace in the same namespace

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockBondApiDAO.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import org.broadinstitute.dsde.rawls.dataaccess.{BondResponseData, _}
+import org.broadinstitute.dsde.rawls.model._
+import scala.concurrent.{ExecutionContext, Future}
+
+class MockBondApiDAO(bondBaseUrl: String)(implicit executionContext: ExecutionContext) extends BondApiDAO {
+  override def getBondProviders(): Future[List[String]] = Future.successful(List("fence", "dcf-fence"))
+
+  override def getServiceAccountKey(provider: String, userInfo: UserInfo): Future[Option[BondResponseData]] = {
+    if (provider == "fence") {
```
andy7i, line 11: is this making any mock call to `fence` return None in order to emulate having one unlinked provider?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -516,6 +521,39 @@ class WorkspaceServiceSpec extends FlatSpec with ScalatestRouteTest with Matcher
...
+    populateWorkspacePolicies(services)
+
+    runAndWait(workspaceRequesterPaysQuery.insertAllForUser(testData.workspace.toWorkspaceName, testData.userWriter.userEmail, Set(BondServiceAccountEmail("foo@bar.com"))))
+
+    val aclUpdate = Set(WorkspaceACLUpdate(testData.userWriter.userEmail.value, WorkspaceAccessLevels.Read, None))
+    Await.result(services.workspaceService.updateACL(testData.workspace.toWorkspaceName, aclUpdate, false), Duration.Inf)
+
+    runAndWait(workspaceRequesterPaysQuery.listAllForUser(testData.workspace.toWorkspaceName, testData.userWriter.userEmail)) shouldBe empty
+  }
+
```
andy7i, line 556: is it worth having a test checking when a reader is upgraded to a writer/owner?

andy7i (commented):
phew! you were right -- there's a lot of new stuff. Most of the comments are clarifying questions and not blocking. 

Do we need the bond configs merged first? https://github.com/broadinstitute/firecloud-develop/tree/CA-702-adding-bond-to-rawls-config

Running tests locally.
Github won't let me approve the PR since I opened it, but here's a 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    def * = (id, workspaceId, userEmail, serviceAccountEmail) <> (WorkspaceRequesterPaysRecord.tupled, WorkspaceRequesterPaysRecord.unapply)
+  }
+
+  object workspaceRequesterPaysQuery extends TableQuery(new WorkspaceRequesterPaysTable(_)) {
+    def insertAllForUser(workspaceName: WorkspaceName, userEmail: RawlsUserEmail, serviceAccountEmails: Set[BondServiceAccountEmail]): ReadWriteAction[Int] = {
+      for {
+        maybeWorkspaceId <- workspaceQuery.getWorkspaceId(workspaceName)
+        workspaceId = maybeWorkspaceId.getOrElse(throw new RawlsException(s"workspace not found $workspaceName"))
+        existingRecords <- existingRecordsForUserQuery(workspaceName, userEmail).result
```
gpolumbo-broad, line 30: Is there a way to get Slick to do an `INSERT IF NOT EXISTS` query?

andy7i:
local tests results. looks like one unrelated test failed? I can rerun it after restarting my computer since it is starting to run at glacial speeds 

```
[info] ScalaTest
[info] Run completed in 24 minutes, 20 seconds.
[info] Total number of tests run: 1026
[info] Suites: completed 50, aborted 0
[info] Tests: succeeded 1025, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed: Total 1026, Failed 1, Errors 0, Passed 1025
[error] Failed tests:
[error]         org.broadinstitute.dsde.rawls.monitor.BucketDeletionMonitorSpec
[error] (rawlsCore / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
dvoet: `updatePolicyBindings` is a function that takes a function with a single parameter as an argument. The `=>` notation is saying that this is a function, the things on the left are the names of the parameters (just 1 in this case, `existingPolicies`) and the stuff on the right is the body of the function. The type of the parameter is not explicitly specified because `updatePolicyBindings` says what it must be.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: this is standard object-oriented programming, nothing scala special about it. `updatePolicyBindings` is defined without an implementation in the base class and and sub classes define the specific implementations according to their model. The http class does the real google stuff, the mock class does something in memory.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: the reason is that remove returns `Future[Unit]` but add returns `Future[Boolean]`. The question really is why are the return types different? They should not be, will fix.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
dvoet: actually, I think this was part of a dead end implementation approach

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupService.scala
dvoet: it is not casting per se although that is effectively what is going on. `BondServiceAccountEmail` has a single `String` arg constructor and `map` here is requires a function of the same shape so scala figures out what to do.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
dvoet: it is setting the id to 0. However slick knows that for inserts it should not include the primary key which is auto incrementing. See https://scala-slick.org/doc/3.2.0/queries.html#inserting paragraph starting with "When you include an AutoInc"

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupServiceSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+    val expectedEmail = BondServiceAccountEmail("bondSA")
+
+    // add user to 2 workspaces in same namespace
+    runAndWait(workspaceRequesterPaysQuery.insertAllForUser(minimalTestData.workspace.toWorkspaceName, userInfo.userEmail, Set(expectedEmail)))
+    runAndWait(workspaceRequesterPaysQuery.insertAllForUser(minimalTestData.workspace2.toWorkspaceName, userInfo.userEmail, Set(expectedEmail)))
+    runAndWait(workspaceRequesterPaysQuery.userExistsInWorkspaceNamespace(minimalTestData.billingProject.projectName.value, userInfo.userEmail)) shouldBe true
+
+    // add user to mock google bindings
+    val initialBindings = Map(service.requesterPaysRole -> Set("serviceAccount:" + expectedEmail.client_email))
+    service.googleServicesDAO.asInstanceOf[MockGoogleServicesDAO].policies.put(minimalTestData.billingProject.projectName, initialBindings)
```
gpolumbo-broad, line 78: Why do we need to manually add the user to the google policy bindings when adding the user to the workspace, but this happens automatically when deleting the user from the workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
dvoet: correct

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupServiceSpec.scala
dvoet: correct

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockBondApiDAO.scala
dvoet: correct, this is mocking out bond behavior in the case there is no linked account for the provider named "fence"

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpBondApiDAO.scala
dvoet: it is necessary because that is what the json looks like, it has a field named `providers`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
dvoet: I spent too long fighting with slick to figure that out. There were some posts in stack overflow that looked too magical.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/RequesterPaysSetupServiceSpec.scala
dvoet: It is more like just in time triggering for addition. The trigger will be starting a workflow or notebook. Doing it when adding to a workspace is not great because the user may not have linked accounts yet (or even be a registered user for that matter).

andy7i:
> local tests results. looks like one unrelated test failed? I can rerun it after restarting my computer since it is starting to run at glacial speeds
> 
> ```
> [info] ScalaTest
> [info] Run completed in 24 minutes, 20 seconds.
> [info] Total number of tests run: 1026
> [info] Suites: completed 50, aborted 0
> [info] Tests: succeeded 1025, failed 1, canceled 0, ignored 0, pending 0
> [info] *** 1 TEST FAILED ***
> [error] Failed: Total 1026, Failed 1, Errors 0, Passed 1025
> [error] Failed tests:
> [error]         org.broadinstitute.dsde.rawls.monitor.BucketDeletionMonitorSpec
> [error] (rawlsCore / Test / test) sbt.TestsFailedException: Tests unsuccessful
> ```

reran tests after restarting and everything passed.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
dvoet: nothing should happen with respect to requester pays in that case

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceRequesterPaysComponent.scala
gpolumbo-broad: the boolean expressions in `filter()` are being checked for each element in the collection.  In this context, the `_` is like saying `currentElement`.

andy7i (commented):
LGTM 👍

dvoet:
jenkins retest

gpolumbo-broad (approved):

dvoet (approved):

PR #1193, opened 2020-04-06
Fix hog limit bug [WA-173]

rsasch:
Don't use number of Cromwell servers to calculate how many workflows per user to allow.

Closes https://broadworkbench.atlassian.net/browse/WA-173

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -299,9 +299,8 @@ object Boot extends IOApp with LazyLogging {
         metricsPrefix
       )
 
-      val maxActiveWorkflowsTotal = conf.getInt(
-        "executionservice.maxActiveWorkflowsPerServer"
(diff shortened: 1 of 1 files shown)


aednichols (commented):
I think we should change the Rawls hog factor from 10 to 5 so that we remain at 4000 max WFs per user in production. This limit appears to be completely safe and there is a user now who has 4000 workflows.

rsasch:
> I think we should change the Rawls hog factor from 10 to 5 so that we remain at 4000 max WFs per user in production. This limit appears to be completely safe and there is a user now who has 4000 workflows.

Do you mean here: https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/rawls/rawls.conf.ctmpl#L314 ?

aednichols:
Yes, precisely.

rsasch:
Created https://github.com/broadinstitute/firecloud-develop/pull/2087

aednichols (approved):

salonishah11 (approved):

PR #1198, opened 2020-05-01
AS-302: use proper project for updating import service job status [risk: low]

davidangb:
When we originally developed this Rawls code, we were using a hand-rolled dev environment for import service, and the import-service-notify-${ENV} pubsub topic lived in broad-dsde-dev, alongside Rawls. 

That's not our final architecture - that pubsub topic lives in terra-importservice-${ENV}, alongside the import service GAE app.

Therefore, Rawls needs this PR in order to publish to the topic.


ansingh7115 (approved):
looks good!

MatthewBemis (approved):

PR #1203, opened 2020-05-13
AS-273 EnumerateSnapshots endpoint

zloery:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


zloery:
Snapshot-related tests are all passing, but I'm still running the full set of tests.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3251,6 +3251,44 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/definitions/EnumerateSnapshotRequestBody'
+      responses:
+        '200':
+          description: OK
+          schema:
+            $ref: '#/definitions/DataRepoSnapshotList'
+        '404':
+          description: Workspace or snapshot not found
```
davidangb: just "workspace not found"? I assume this will return an empty array if no snapshots are found.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3251,6 +3251,44 @@ paths:
...
           schema:
             $ref: '#/definitions/DataRepoSnapshot'
+    get:
+      requestBody:
+        description: |
+          Parameters for enumerating snapshots.
+        content:
+          application/json:
+            schema:
+              $ref: '#/definitions/EnumerateSnapshotRequestBody'
```
davidangb: 1. I think `requestBody` is OpenAPI 3, not supported in OpenAPI 2. This should move to `parameters` with `in: body`
2. Is this really a GET request with a body? We should avoid those, and move offset/limit to query params, assuming this remains as a GET

core/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -13,7 +17,16 @@ object DataRepoSnapshotReference {
   }
 }
 
+object DataRepoSnapshotListFromDataReferenceList {
+  def apply(ref: DataReferenceList): DataRepoSnapshotList = {
+    val buffer = for (r <- ref.getResources.asScala) yield DataRepoSnapshotReference(r)
```
davidangb, line 21: can `getResources` return null here, or would it return an empty collection?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -13,7 +17,16 @@ object DataRepoSnapshotReference {
...
+    DataRepoSnapshotList(buffer.toList)
+  }
+}
+
```
davidangb, line 25: stylistically, I would not make a new object class for this. 

I would make this a companion object to `DataRepoSnapshotList`. I realize that in doing so, you introduce ambiguity on the `apply` methods ... so I'd rename your custom method  to `from`:
```
object DataRepoSnapshotList {
  def from(ref: DataReferenceList): DataRepoSnapshotList = {
    ...
```

Then in SnapshotService, change:
```
Future.successful(DataRepoSnapshotListFromDataReferenceList(ref))
```
to
```
Future.successful(DataRepoSnapshotList.from(ref))
```

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -154,4 +154,45 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
       sealRoute(services.snapshotRoutes) ~>
       check { assertResult(StatusCodes.NotFound) {status} }
   }
+
+  it should "return 200 when a user lists all workspaces" in withTestDataApiServicesAndUser(testData.userReader.userEmail.value) { services =>
```
davidangb: nit: test name should be something like "... when a user lists snapshots in a workspace"

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -154,4 +154,45 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
...
+      EnumerateSnapshotRequestBody(
+        offset = 0,
+        limit = 10
+      )
+    )) ~>
+      sealRoute(services.snapshotRoutes) ~>
+      check {
+        val response = responseAs[DataRepoSnapshotList]
+        assertResult(StatusCodes.OK) {status}
+        assert(response.snapshots.size == 0)
```
davidangb: Ideally, the `MockWorkspaceManagerDAO.references` TrieMap that underlies this response would have some values in it, and this assert would look for size > 0. That would exercise all the json deserialization and class conversions.

davidangb (commented):
initial comments inline!

core/src/main/resources/swagger/api-docs.yaml
zloery: updated - it could also be "user lacks permissions"

core/src/main/resources/swagger/api-docs.yaml
zloery: I think you're right - moved these to parameters.
I can also just use query params, so I switched to those.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
zloery: WM always returns a DataReferenceList with an empty collection if there's no items

core/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
zloery: Ah! `companion object` was the term I was missing to google what was going on here, thanks!
My only motivation for making this a separate object class was avoiding the ambiguous `apply` issue - I wasn't sure if renaming the class or the `apply` method was a worse style violation, I'll switch this.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
zloery: Done!

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
zloery: Good point. I added two `create` calls before the list and check that both snapshots get returned. 
Style q: Is there a better way to organize this than nesting calls in `check` blocks?

zloery:
Along with those fixes, I also changed `limit` and `offset` to be explicitly required - I don't think we can use the OpenApi default values when calling from a client like this, and I really don't want to maintain separate values for defaults here and in WSM.

zloery:
Also, orch PR here: https://github.com/broadinstitute/firecloud-orchestration/pull/785

davidangb:
uh oh @zloery this needs a rebase ... I'll re-review once that's done!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -48,6 +49,13 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
     }
   }
 
+  def enumerateSnapshots(workspaceName: WorkspaceName, offset: Int, limit: Int): Future[DataRepoSnapshotList] = {
+    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Some(WorkspaceAttributeSpecs(all = false))).flatMap { workspaceContext =>
+      val ref = workspaceManagerDAO.enumerateDataReferences(workspaceContext.workspaceId, offset, limit, userInfo.accessToken)
+      Future.successful(DataRepoSnapshotList.from(ref))
+    }
+  }
```
davidangb, line 57: this is fine as-is, your call on whether or not to change it. For the sake of idiomatic scala ... 

this is currently written as: `getWorkspaceContextAndPermissions ... flatMap ... Future.successful()` but instead of using `flatMap` and creating an inside value of `Future`, you can just use `map`. So, this block could be:
```
  def enumerateSnapshots(workspaceName: WorkspaceName, offset: Int, limit: Int): Future[DataRepoSnapshotList] = {
    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
      val ref = workspaceManagerDAO.enumerateDataReferences(workspaceContext.workspaceId, offset, limit, userInfo.accessToken)
      DataRepoSnapshotList.from(ref)
    }
  }
```

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -155,41 +155,51 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
       check { assertResult(StatusCodes.NotFound) {status} }
   }
 
-  it should "return 200 when a user lists all workspaces" in withTestDataApiServicesAndUser(testData.userReader.userEmail.value) { services =>
-    Get(s"${testData.wsName.path}/snapshots", httpJson(
-      EnumerateSnapshotRequestBody(
-        offset = 0,
-        limit = 10
+  it should "return 200 when a user lists all snapshots in a workspace" in withTestDataApiServices { services =>
+    // First, create two workspaces
```
davidangb: nit: "First, create two data references"

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -155,41 +155,51 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
...
+          check {
+            val response = responseAs[DataRepoSnapshotReference]
+            assertResult(StatusCodes.Created) {status}
+            // Then, list them both
+            Get(s"${testData.wsName.path}/snapshots?offset=0&limit=10") ~>
+              sealRoute(services.snapshotRoutes) ~>
+              check {
+                val response = responseAs[DataRepoSnapshotList]
+                assertResult(StatusCodes.OK) {status}
+                assert(response.snapshots.size == 2)
```
davidangb, line 188: for deeper testing, you could do:
```
                 response.snapshots should contain theSameElementsAs List(
                  DataRepoSnapshot(
                    name = "foo",
                    snapshotId = "realsnapshot"
                  ),
                  DataRepoSnapshot(
                    name = "bar",
                    snapshotId = "realsnapshot2"
                  )
                )
```
(untested/uncompiled code, could be wrong)

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -155,41 +155,51 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
...
+              }
+          }
       }
+
   }
```
davidangb, line 194: couple comments on this test:
1. My original thought was that you'd edit `MockWorkspaceManagerDAO` and hardcode some fixture data inside the mock, then all unit tests for read-only functions could request data from the mock, know what to expect, and assert on it. That way, a read-only unit test wouldn't fail if there's a flaw in functions that write. I'm fine either way and wouldn't hold the PR on this design.
2. In looking at `MockWorkspaceManagerDAO`, there's a problem with its `references` member: it applies to all workspaces. If you use `MockWorkspaceManagerDAO.createDataReference` to add a snapshot for workspace A, then use `MockWorkspaceManagerDAO.enumerateDataReferences` to list snapshots for workspace B, you'll get workspace A's snapshot. I feel like this is going to bite us and we should solve it soon; if you want to solve it in a follow-on PR, that's fine

zloery: For 1, I'd lean towards using the write methods instead of hardcoding data - the unit tests are a bit less isolated, but there's one less place to maintain write behavior.
Agree on 2, I think it's enough of an issue to include in this PR. I updated the references map to use a `(workspaceId, referenceId)` tuple as a key instead of just `referenceId` which seems like a decent representation of the real system, but let me know if you'd rather take a different approach.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
zloery: One issue - the returned list is DataRepoSnapshotReferences, which have a lot more fields than the request model (some of which are generated by WM).
I verified the names from the inputs, though the snapshotIds are a little trickier since it requires parsing json.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
zloery: Done - I didn't realize you could do `map` on Futures that way. Thanks!

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
```diff
@@ -7,18 +7,20 @@ import akka.http.scaladsl.model.headers.OAuth2BearerToken
...
 class MockWorkspaceManagerDAO extends WorkspaceManagerDAO {
 
-  val references: TrieMap[UUID, DataReferenceDescription] = TrieMap()
+  val references: TrieMap[(UUID, UUID), DataReferenceDescription] = TrieMap()
 
   def mockGetWorkspaceResponse(workspaceId: UUID) = new WorkspaceDescription().id(workspaceId)
   def mockCreateWorkspaceResponse(workspaceId: UUID) = new CreatedWorkspace().id(workspaceId.toString)
-  def mockReferenceResponse(referenceId: UUID) = references.get(referenceId).getOrElse(throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not found")))
+  def mockReferenceResponse(workspaceId: UUID, referenceId: UUID) = references.get((workspaceId, referenceId)).getOrElse(throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not found")))
+  def mockEnumerateReferenceResponse(workspaceId: UUID) = new DataReferenceList().resources(references.filterKeys(_._1 == workspaceId).values.toList.asJava)
```
davidangb: ahh, thorny Scala thing here, which is not your fault at all: avoid `filterKeys` and `mapValues`.

Pretty much everywhere else in Scala, filtering/mapping a collection returns a materialized copy of the original with the appropriate transforms applied. `filterKeys` and `mapValues` don't - they return a view, more like Java behavior. This can lead to unexpected consequences and unpredictable bugs.

There are some cases where it's fine, but it's best to just avoid these altogether and never wonder.
https://stackoverflow.com/questions/14882642/scala-why-mapvalues-produces-a-view-and-is-there-any-stable-alternatives
https://github.com/scala/bug/issues/4776

I would rewrite this as:
```
    new DataReferenceList().resources(references.collect {
      case ((wsId, _), refDescription) if wsId == workspaceId => refDescription
    }.toList.asJava)
```

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
davidangb: ah of course you're right it's a different return object!

For verifying the names, how about:
```
assertResult(Set("foo", "bar")) { response.snapshots.map(_.name).toSet }
```

davidangb (commented):
so close, one important comment inline about `filterKeys` and one optional improvement to a test assertion

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
zloery: Huh, interesting. I'm sure there's a good use for that behavior, but I'll avoid it here.

davidangb (approved):
:+1: thanks for sticking this one out!

zloery:
> 👍 thanks for sticking this one out!

Thank you for all the help!

MatthewBemis (approved):
lgtm, just a heads up that you'll want to change the base to `develop`

Jenkins tests won't pass right now because of a quay.io issue, but once that goes green I'm a :+1:

zloery:
Jenkins retest

PR #1205, opened 2020-05-14
Data Repo enable/disable flag

davidangb:
Adds a config flag to enable/disable data repo features. This would allow us to merge the feature branch to develop, avoiding merge conflicts. We'd  be able to set the enable flag to true in dev, giving us a persistent environment in which to test, but leave the flag false (or empty) in other envs.

Of course, we may want to maintain shorter-lived, smaller feature branches while multiple people are coding in the same areas. My goal is to avoid a long-lived, giant feature branch.

When the config flag `dataRepo.enabled` is false or empty, this PR:
* never starts the snapshot routes, so those APIs are disabled
* serves the main api-docs.yaml swagger, which has no snapshot routes or models

When the config flag `dataRepo.enabled` is present and true, this PR:
* starts the snapshot routes,  making those APIs available
* merges data-repo-only.yaml and api-docs.yaml and serves the result in swagger UI, making the snapshot routes and models visible

This will require a firecloud-develop PR, to enable data repo only in dev.

NB: the `mergeYamls` functionality in this PR could be used to break our 5183-line api-docs.yaml into multiple pieces, but I have no intention of doing that in this PR.


MatthewBemis (approved):
I really like this! Looks good to me. We'll need to do this for orchestration as well.

I still need to open that firecloud-develop PR so this config value can go in there. After those things are done, let's get all of this merged!

zloery (approved):
Nice work on the yamls, lgtm!

PR #1207, opened 2020-05-14
entity provider refactor part 2 [no ticket; risk: low]

davidangb:
Aiming to make the entity provider and its surrounding classes more understandable and ready for multiple developers to contribute.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
@@ -4,7 +4,7 @@ import org.broadinstitute.dsde.rawls.entities.base.{EntityProvider, EntityProvid
 import org.broadinstitute.dsde.rawls.entities.datarepo.DataRepoEntityProvider
 import org.broadinstitute.dsde.rawls.entities.exceptions.DataEntityException
 import org.broadinstitute.dsde.rawls.entities.local.LocalEntityProvider
-import org.broadinstitute.dsde.rawls.model.{RawlsBillingProject, UserInfo, Workspace}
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
```diff
@@ -1,39 +1,56 @@
...
-  def getEntity(workspace: Workspace, entityRef: AttributeEntityReference) = ???
-
-  // internal method to figure out which entityprovider should be used for a given request
-  private def resolveProvider(workspace: Workspace): EntityProvider = {
-    // soon: inspect the workspace's registered snapshots plus request parameters,
-    // and route to the appropriate entity provider. This will likely require a method signature change
-    // to accept request details. For now, just find the builder for DefaultEntityProvider.
-    val defaultEntityProviderBuilder = providerBuilders.find(_.builds == typeTag[LocalEntityProvider])
+  // should we combine any of these arguments into a "EntityRequestArguments" case class,
+  // for encapsulation/abstraction?
```
dvoet: yes, I think so. There is a little uncertainty here and it will allow for some flexibility.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/EntityProviderBuilder.scala
```diff
@@ -14,5 +14,8 @@ trait EntityProviderBuilder[T <: EntityProvider] {
 
   /** create the EntityProvider this builder knows how to create.
     */
-  def build(workspace: Workspace): T
+  def build(workspace: Workspace,
```
dvoet: this should use `EntityRequestArguments` too

dvoet (approved):
add `EntityRequestArguments` then 👍

PR #1209, opened 2020-05-18
AS-321: move entity metadata to provider classes [risk: low]

davidangb:
moves the entity type metadata implementation out of EntityService and into the Entity Provider hierarchy. Does not attempt to implement for Data Repo yet; that will come in a following PR.

Doing this refactor first keeps both PRs smaller and focused.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -60,7 +60,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
 
   def createEntity(workspaceName: WorkspaceName, entity: Entity): Future[Entity] =
     withAttributeNamespaceCheck(entity) {
-      getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write) flatMap { workspaceContext =>
+      getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
         entityManager.resolveProvider(workspaceContext.workspace, userInfo).createEntity(entity)
       }
     }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -60,7 +60,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
 
   def createEntity(workspaceName: WorkspaceName, entity: Entity): Future[Entity] =
     withAttributeNamespaceCheck(entity) {
-      getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write) flatMap { workspaceContext =>
+      getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
```
ansingh7115, line 63: Should we do this for all `getWorkspaceContextAndPermissions` in here?

davidangb: yes, but I feel like that's out of scope of this PR (see also AS-7)

davidangb: or, sure - I can do it for all instances in `EntityService`, but not all instances everywhere

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -149,10 +149,9 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
     }
```
ansingh7115, line 150: Also out of scope perhaps? But EntityServiceSpec does not seem to have test coverage for a lot of the methods in this class. We should probably make a ticket for that

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -18,9 +18,15 @@ class LocalEntityProvider(workspace: Workspace, dataSource: SlickDataSource)
...
 
-  override def createEntity(entity: Entity): Future[Entity] = {
-    val workspaceContext = SlickWorkspaceContext(workspace)
+  private val workspaceContext = SlickWorkspaceContext(workspace)
+
+  override def entityTypeMetadata(): Future[Map[String, EntityTypeMetadata]] = {
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+    }
+  }
```
ansingh7115, line 27: LocalEntityProvider doesn't have any unit tests

davidangb: probably worth it to add unit tests; `EntityApiServiceSpec` tests all the features in here, though

davidangb:
unrelated fiab-start failure; jenkins retest

ansingh7115 (commented):
Looks good - put test related things in a separate ticket

ansingh7115 (approved):

PR #1212, opened 2020-05-19
AS-339: better debugging for jenkins builds [risk: low]

davidangb:
1. add debugging log statements to help diagnose jenkins build problems
2. send the stdout of `sbt test` to Jenkins console; since this was run in a subshell it needed a tweak to not suppress stdout. Previously we only saw stderr.
3. fix the ivy/sbt/coursier cache mounts, I don't expect this to show much difference.

Reviewer: if you're curious, click through to the jenkins build to see its output:
https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-build/2914/console

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -21,6 +21,8 @@ EOF
 # Enable strict evaluation semantics
 set -e
 
+echo "rawls docker/build.sh starting ..."
+
 # Set default variables
 DOCKER_CMD=
 BRANCH=${BRANCH:-$(git rev-parse --abbrev-ref HEAD)}  # default to current branch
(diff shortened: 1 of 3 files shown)


docker/build.sh
```diff
@@ -21,6 +21,8 @@ EOF
 # Enable strict evaluation semantics
 set -e
```
MatthewBemis, line 22: would it be useful to set `x` and/or `u` here as well?

MatthewBemis (approved):
One small question, otherwise looks good!

docker/build.sh
davidangb: I explicitly left out the `x`, as it would log the artifactory password; and this script already has decent `echo` statements. We could add it later if we find it's still hard to debug?

MatthewBemis: Hah! Good call then. I think it's fine as-is then

PR #1214, opened 2020-05-21
AS-321: entity types data repo [risk: low]

davidangb:
This PR:
* accepts `dataReference` and `billingProject` query params for the entity metadata api
  * generalized at the code level to accept these params for all entity apis; only used by metadata api
* adds a `DataRepoEntityProvider` and `DataRepoEntityProviderBuilder`
* adds a `DataRepoDAO` trait and `HttpDataRepoDAO` impl
  * implemented via data repo client library
* for entity metadata api, routes to `DataRepoEntityProvider` if `dataReference` query param is nonempty
* `DataRepoEntityProvider` implements:
  * request data reference description from Workspace Manager, based on the `dataReference` name
  * request snapshot description from Terra Data Repo, based on the snapshot id in Workspace Manager's data reference description
  * read tables and columns from Terra Data Repo's snapshot description


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.2.8
+sbt.version=1.3.10
```
davidangb, line 1: This PR adds the `datarepo-client` dependency. That has a transitive dependency on `javax.ws.rs:javax.ws.rs-api:2.1.1`. There is a bug in sbt 1.2 - actually in Ivy - that causes that transitive dependency to fail; see https://github.com/sbt/sbt/issues/3618.

Upgrading to sbt 1.3 addresses this problem, because sbt 1.3 uses Coursier, not Ivy. Alternate solution is to ask the data repo team to change their dependency, but the sbt 1.3 upgrade feels worthwhile anyway.

Note that broadinstitute/firecloud-develop#2138 enables coursier caching for local-environment servers. It's not technically required but it avoids re-downloading dependencies every time you start your local server.

EDIT: I've excluded `javax.ws.rs:javax.ws.rs-api:2.1.1` to avoid sbt assembly issues, so this sbt upgrade is no longer necessary ... but I've tested it and it seems worthwhile so I'm leaving it in.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+
+  it should "bubble up error if data repo errors" in {
+    val provider = createTestProvider(
+      dataRepoDAO = new SpecDataRepoDAO(Left(new bio.terra.datarepo.client.ApiException("whoops 2"))))
+
+    val ex = intercept[bio.terra.datarepo.client.ApiException] { provider.entityTypeMetadata() }
+    assertResult("whoops 2") { ex.getMessage }
+  }
+
+  // to-do: tests for different primary key values returned by data repo, once TDR supports this
```
ungwudik: How do we know when TDR starts supporting these? We should consider creating tickets for these tests (if they don't already exist) and linking them to the corresponding ticket on TDR team's JIRA.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpecSupport.scala
```diff
@@ -0,0 +1,123 @@
...
+import org.broadinstitute.dsde.rawls.model.{RawlsUserEmail, RawlsUserSubjectId, UserInfo, Workspace}
+import org.joda.time.DateTime
+import spray.json.{JsObject, JsString}
+
+import scala.collection.JavaConverters._
+
+trait DataRepoEntityProviderSpecSupport {
+
+  // default values for some important attributes
+  val wsid: UUID = UUID.randomUUID()
```
ungwudik: Can these two variables (`wsid` and `refid`) be updated to camelCase for consistency?

core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -1,6 +1,33 @@
...
+          description: Workspace Name
+          name: workspaceName
+          required: true
+          type: string
+        - in: query
+          description: Data reference name to query for entity metadata
+          name: dataReference
+          required: false
+          type: string
+        - in: path
```
MatthewBemis: shouldn't this also be `in: query`?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -1,24 +1,101 @@
...
+    // get snapshotId from reference name
+    val snapshotId = lookupSnapshotForName(dataReferenceName)
+
+    // contact TDR to describe the snapshot
+    val snapshotModel = dataRepoDAO.getSnapshot(snapshotId, userInfo.accessToken)
+
+    // reformat TDR's response into the expected response structure
+    val entityTypesResponse: Map[String, EntityTypeMetadata] = snapshotModel.getTables.asScala.map { table =>
+      val attrs: Seq[String] = table.getColumns.asScala.map(_.getName)
+      // TODO: AS-321 better way to describe the primary key, if null or more than one PK?
```
MatthewBemis: i'm most certainly missing the appropriate context, but does Data Repo somehow allow for null PKs or is that comment more related to an unexpected null making it's way into here?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -1,24 +1,101 @@
...
     throw new UnsupportedEntityOperationException("delete entities not supported by this provider.")
+
+  // not marked as private to ease unit testing
+  def lookupSnapshotForName(dataReferenceName: String): UUID = {
+    // contact WSM to retrieve the data reference specified in the request
+    val dataRef = workspaceManagerDAO.getDataReferenceByName(UUID.fromString(workspace.workspaceId),
+      ReferenceTypeEnum.DATAREPOSNAPSHOT.getValue,
+      dataReferenceName,
+      userInfo.accessToken)
+    // TODO: AS-321 verify we got one back (should be noop; request will throw if 0 found, but could use better error-trapping)
```
MatthewBemis: is there a reason to hold off on adding that error-trapping now?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -1,24 +1,101 @@
...
+
+    // verify reference object contains the instance and snapshotId keys
+    // TODO: AS-321 when broadinstitute/rawls#1219 merges, adopt its case class instead of using raw json objects
+    val cursor = refObj.hcursor
+    val refInstance: String = cursor.get[String]("instance").getOrElse(throw new DataEntityException(s"Reference value for $dataReferenceName does not contain an instance value."))
+    val refSnapshot: String = cursor.get[String]("snapshot").getOrElse(throw new DataEntityException(s"Reference value for $dataReferenceName does not contain a snapshot value."))
+
+    // verify the instance matches our target instance
+    // TODO: AS-321 is this the right place to validate this? We could add a "validateInstanceURL" method to the DAO itself, for instance
+    if (refInstance != dataRepoDAO.getBaseURL) {
```
MatthewBemis: out of an abundance of caution, should this be a little less strict, i.e. `equalsIgnoreCase`?

MatthewBemis (commented):
Just a few minor comments!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: I've updated the comment to reflect reality

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
davidangb: updated comments.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpecSupport.scala
davidangb: done

core/src/main/resources/swagger/data-repo-only.yaml
davidangb: done

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: updated comment

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: done

davidangb:
@ungwudik @MatthewBemis ready for re-review!

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
ungwudik: Looks like DR-1003 was completed yesterday? I'm not 100% sure what 'Done' means for their team, so adding "tests for entity/row counts returned by data repo" could be a part of a future PR.

ungwudik (approved):
Looks good to me, pending approvals from Matt and/or Anu.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -1,24 +1,104 @@
...
   override def createEntity(entity: Entity): Future[Entity] =
     throw new UnsupportedEntityOperationException("create entity not supported by this provider.")
 
   override def deleteEntities(entityRefs: Seq[AttributeEntityReference]): Future[Int] =
     throw new UnsupportedEntityOperationException("delete entities not supported by this provider.")
+
+  // not marked as private to ease unit testing
+  def lookupSnapshotForName(dataReferenceName: String): UUID = {
+    // contact WSM to retrieve the data reference specified in the request
+    val dataRef = workspaceManagerDAO.getDataReferenceByName(UUID.fromString(workspace.workspaceId),
```
ansingh7115, line 62: I believe by the time we get here, we've asked Sam whether the user has Read access to the workspace. Since snapshots have their own sam permissions - do we need to check that the user also has permissions on the snapshot? Or are we assuming that when a snapshot is associated with a workspace, that means anyone who can or will in the future have Read access to the workspace, should be able to see the snapshot?

davidangb: @ansingh7115 per PO business requirements, any user who has Read access to the workspace should be able to see all snapshot references in that workspace. This doesn't necessarily mean that the user can see the snapshot *contents* - and those permission checks will happen as soon as we connect to Data Repo. But as far as this specific `lookupSnapshotForName` method - any workspace user should be able to resolve the name into the snapshot UUID.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
davidangb: I saw that the jira ticket was marked Done, I did not see actual row count data in the Data Repo dev instance, so I haven't adopted it yet, I'll make a follow-on AS ticket to track this work.

MatthewBemis (approved):
:+1: pending passing tests

davidangb:
```
write /var/lib/docker/tmp/GetImageBlob612182488: no space left on device
```
in last fiab-start. jenkins retest.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
davidangb: tracking as AS-365

ansingh7115 (approved):

PR #1215, opened 2020-05-21
Travis should use same Scala version as sbt build [no ticket; risk: no]

davidangb:
See https://github.com/broadinstitute/rawls/blob/develop/project/Settings.scala#L54 ... Travis should use the same version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ dist: trusty
 language: scala
 
 scala:
-  - 2.12.8
+  - 2.12.10
 
 sudo: required
 


davidangb:
jenkins had fiab-start failures, but ... I'm not going to re-run. This is a Travis-only change.

davidangb:
jenkins retest

davidangb:
yeah, I keep getting fiab failures but still, this is just a Travis change.

zarsky-broad (approved):
Makes sense and looks good, scala version in the travis builds is indeed changed to 2.12.10!

PR #1220, opened 2020-05-28
AS-232: deleteEntities behind EntityProvider interface [risk: low]

davidangb:
This PR:
* moves the implementation of delete-entities behind the EntityProvider interface
  * with a noop implementation for DataRepo
* cleans up swagger a bit for the data repo-related APIs

### DIFF
--- a/core/src/main/resources/swagger/data-repo-only.yaml
+++ b/core/src/main/resources/swagger/data-repo-only.yaml
@@ -8,26 +8,26 @@ paths:
   '/api/workspaces/{workspaceNamespace}/{workspaceName}/entities':
     get:
       parameters:
-        - in: path
-          description: Workspace Namespace
-          name: workspaceNamespace
+        - $ref: '#/parameters/workspaceNamespacePathParam'
+        - $ref: '#/parameters/workspaceNamePathParam'
+        - $ref: '#/parameters/dataReferenceQueryParam'
+        - $ref: '#/parameters/billingProjectQueryParam'
+
+  '/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/delete':
+    post:
+      parameters:
+        - $ref: '#/parameters/workspaceNamespacePathParam'
+        - $ref: '#/parameters/workspaceNamePathParam'
+        - in: body
+          description: Set of entities to delete
+          name: entities
           required: true
-          type: string
-        - in: path
-          description: Workspace Name
-          name: workspaceName
-          required: true
-          type: string
-        - in: query
-          description: Data reference name to query for entity metadata
-          name: dataReference
-          required: false
-          type: string
-        - in: query
-          description: Billing project to use for queries to the data reference
-          name: billingProject
-          required: false
-          type: string
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/AttributeEntityReference'
+        - $ref: '#/parameters/dataReferenceQueryParam'
+        - $ref: '#/parameters/billingProjectQueryParam'
   '/api/workspaces/{namespace}/{name}/snapshots':
     post:
       responses:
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -96,19 +97,20 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
-            if (referringEntities != entRefs.toSet)
-              DBIO.successful(RequestComplete(StatusCodes.Conflict, referringEntities))
-            else {
-              dataAccess.entityQuery.hide(workspaceContext, entRefs).map(_ => RequestComplete(StatusCodes.NoContent))
-            }
-          }
+
+      // TODO: insert the billing project, if present. May want to use EntityRequestArguments or other container class.
+      // TODO: now with two methods building EntityRequestArguments, we probably want to factor that out into the
+      // EntityService constructor, or other higher-level method
```
davidangb, line 105: I don't want to tackle this refactoring until #1221 merges, will be in a follow-on PR

core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -8,26 +8,26 @@ paths:
...
-          name: workspaceNamespace
-          required: true
-          type: string
-        - in: path
-          description: Workspace Name
-          name: workspaceName
+        - $ref: '#/parameters/workspaceNamespacePathParam'
+        - $ref: '#/parameters/workspaceNamePathParam'
+        - $ref: '#/parameters/dataReferenceQueryParam'
+        - $ref: '#/parameters/billingProjectQueryParam'
```
davidangb, line 14: I moved a bunch of inlined params to a centralized `parameters:` section. There's a lot of cleanup like this that could happen in the main api-docs.yaml file; that's far out of scope of this PR.

core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -165,27 +137,59 @@ paths:
...
+    name: name
+    required: true
+    type: string
+  snapshotIdPathParam:
+    in: path
+    description: The snapshot ID
+    name: snapshotId
+    required: true
+    type: string
+  workspaceNamespacePathParam:
```
ansingh7115, line 168: you have both a `workspaceNamespacePathParam` as works and a `namespacePathParam` defined on line 162 and a similar pattern for workspaceName. Could we use just one naming convention for the name and namespace? Or is there a reason to have two?

davidangb: I had two because I just extracted them from the existing API definitions - but you're right, and I consolidated them by changing the API definitions (in swagger only)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -35,5 +37,19 @@ class LocalEntityProvider(workspace: Workspace, dataSource: SlickDataSource)
     }
   }
 
-  override def deleteEntities(entRefs: Seq[AttributeEntityReference]): Future[Int] = ???
```
zloery, line 38: I'm kind of amazed `???` works in scala, but neat!

zloery (approved):
LGTM

ansingh7115 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
davidangb: you'll get a runtime exception if you try to call it, but it's a nice way to stub things out!

PR #1224, opened 2020-06-01
AS-375: handle spaces in workspace names when talking to import service [risk: low]

davidangb:
properly encode spaces in workspace names/urls when talking to import service


ansingh7115 (approved):

davidangb:
swatomation passed all tests, but failed in fiab-stop. Calling it ok and merging.

PR #1230, opened 2020-06-05
AS-322: get-entity supports data repo [risk: low]

davidangb:
This PR:
* updates to the latest workbench-libs  google2 library, which has a new BigQuery client.
* moves the `getEntity.getEntity()` implementations behind the EntityProvider interface.
* adds a getEntity implementation for DataRepoEntityProvider. This implementation queries Workspace Manager and Data Repo to identify the named snapshot's underlying BigQuery tables, then queries BigQuery to get the specified row and returns it as a Rawls Entity.
* contains a ton of boilerplate/helper methods/utilities to make future entity-related BigQuery work easier.

outstanding TODOs, which I'll file as follow-on Jira tickets for future PRs:
* [ ] get billing project business requirements from PO and implement them (to which project should we bill BigQuery operations, if the user is a READER on the workspace?) [needs PO input]
* [ ] handle RECORD and repeated-value datatypes; these types are awkward in the current client and entity model [needs input from data repo team to find sample data]
* [ ] refactor and simplify where we resolve the EntityProvider impl to be used for any given request, and simplify how we pass arguments around [invasive, should be done in a separate PR]

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -25,7 +25,7 @@ import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.HttpWorkspaceMa
 import slick.basic.DatabaseConfig
 import slick.jdbc.JdbcProfile
 import org.broadinstitute.dsde.rawls.dataaccess.{ExecutionServiceDAO, _}
-import org.broadinstitute.dsde.rawls.entities.EntityService
+import org.broadinstitute.dsde.rawls.entities.{EntityManager, EntityService}
 import org.broadinstitute.dsde.rawls.genomics.GenomicsService
 import org.broadinstitute.dsde.rawls.google.{HttpGoogleAccessContextManagerDAO, HttpGooglePubSubDAO}
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
@@ -392,12 +392,14 @@ object Boot extends IOApp with LazyLogging {
         requesterPaysSetupService
       )
 
+      // create the entity manager.
+      val entityManager = EntityManager.defaultEntityManager(slickDataSource, workspaceManagerDAO, dataRepoDAO)
+
       val entityServiceConstructor: (UserInfo) => EntityService = EntityService.constructor(
         slickDataSource,
         samDAO,
         workbenchMetricBaseName = metricsPrefix,
-        workspaceManagerDAO,
-        dataRepoDAO
+        entityManager
       )
 
       val snapshotServiceConstructor: (UserInfo) => SnapshotService = SnapshotService.constructor(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -393,7 +393,7 @@ object Boot extends IOApp with LazyLogging {
       )
 
       // create the entity manager.
-      val entityManager = EntityManager.defaultEntityManager(slickDataSource, workspaceManagerDAO, dataRepoDAO)
+      val entityManager = EntityManager.defaultEntityManager(slickDataSource, workspaceManagerDAO, dataRepoDAO, samDAO)
 
       val entityServiceConstructor: (UserInfo) => EntityService = EntityService.constructor(
         slickDataSource,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -370,6 +370,9 @@ object Boot extends IOApp with LazyLogging {
       val bondApiDAO: BondApiDAO = new HttpBondApiDAO(bondConfig.getString("baseUrl"))
       val requesterPaysSetupService: RequesterPaysSetupService = new RequesterPaysSetupService(slickDataSource, gcsDAO, bondApiDAO, requesterPaysRole)
 
+      // create the entity manager.
+      val entityManager = EntityManager.defaultEntityManager(slickDataSource, workspaceManagerDAO, dataRepoDAO, samDAO, appDependencies.bigQueryServiceFactory)
+
       val workspaceServiceConstructor: (UserInfo) => WorkspaceService = WorkspaceService.constructor(
         slickDataSource,
         methodRepoDAO,
@@ -389,11 +392,9 @@ object Boot extends IOApp with LazyLogging {
         workbenchMetricBaseName = metricsPrefix,
         submissionCostService,
         workspaceServiceConfig,
-        requesterPaysSetupService
+        requesterPaysSetupService,
+        entityManager
       )
-
-      // create the entity manager.
-      val entityManager = EntityManager.defaultEntityManager(slickDataSource, workspaceManagerDAO, dataRepoDAO, samDAO)
 
       val entityServiceConstructor: (UserInfo) => EntityService = EntityService.constructor(
         slickDataSource,
@@ -520,7 +521,8 @@ object Boot extends IOApp with LazyLogging {
       httpClient <- BlazeClientBuilder(executionContext).resource
       googleServiceHttp <- GoogleServiceHttp.withRetryAndLogging(httpClient, metadataNotificationConfig)
       topicAdmin <- GoogleTopicAdmin.fromCredentialPath(pathToCredentialJson)
-    } yield AppDependencies[F](googleStorage, googleServiceHttp, topicAdmin)
+      bqServiceFactory = new GoogleBigQueryServiceFactory(blocker)
+    } yield AppDependencies[F](googleStorage, googleServiceHttp, topicAdmin, bqServiceFactory)
   }
 }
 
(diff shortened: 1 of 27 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -509,6 +509,8 @@ object Boot extends IOApp with LazyLogging {
     val googleApiUri = Uri.unsafeFromString(gcsConfig.getString("google-api-uri"))
     val metadataNotificationConfig = NotificationCreaterConfig(pathToCredentialJson, googleApiUri)
 
+    implicit val logger = Slf4jLogger.getLogger[F]
+
```
davidangb, line 513: the newer version of workbench-libs google2 requires this implicit logger.

project/Dependencies.scala
```diff
@@ -1,13 +1,13 @@
...
-  val akkaV = "2.5.19"
-  val akkaHttpV = "10.1.7"
+  val akkaV = "2.5.31"
+  val akkaHttpV = "10.1.12"
   val slickV = "3.3.2"
 
   val googleV = "1.22.0"
   val olderGoogleV = "1.20.0"   // TODO why do we have two google versions?  GAWB-2149
-  val workbenchGoogle2V = "0.6-4df42ca"
+  val workbenchGoogle2V = "0.10-ae0b940"
```
davidangb: we want this upgrade to take advantage of the newer BigQuery client from Google.

project/Dependencies.scala
```diff
@@ -1,13 +1,13 @@
 import sbt._
 
 object Dependencies {
-  val akkaV = "2.5.19"
-  val akkaHttpV = "10.1.7"
+  val akkaV = "2.5.31"
+  val akkaHttpV = "10.1.12"
```
davidangb, line 5: akka upgrades are not strictly necessary. The workbench-libs upgrade (below), which is necessary, somehow caused missing akka dependencies; upgrading these akka patch versions fixed it, and upgrading seems worthwhile.

davidangb:
grr, one single Rawls error on `org.broadinstitute.dsde.test.api.BillingApiSpec.A user with a billing account can create a new billing project` with `The project to be billed is associated with an absent billing account`, otherwise all swat tests passed. Jenkins retest, for good measure.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -0,0 +1,145 @@
...
+trait DataRepoBigQuerySupport {
+
+  /**
+   * translate a single BigQuery FieldValue to a single Rawls AttributeValue
+   *
+   * @param field the BQ schema's field definition
+   * @param fv the BQ field value
+   * @return the Rawls attribute value, post-translation
+   */
+  def fieldValueToAttributeValue(field: Field, fv: FieldValue): AttributeValue = {
```
ansingh7115, line 23: Out of curiosity, do you know why Field uses the Legacy SQL data types instead of the Standard ones?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -0,0 +1,145 @@
...
+      case x if fv.isNull =>
+        AttributeNull
+      case LegacySQLTypeName.FLOAT | LegacySQLTypeName.INTEGER | LegacySQLTypeName.NUMERIC =>
+        AttributeNumber(fv.getNumericValue)
+      case LegacySQLTypeName.BOOLEAN =>
+        AttributeBoolean(fv.getBooleanValue)
+      case LegacySQLTypeName.STRING =>
+        AttributeString(fv.getStringValue)
+      case LegacySQLTypeName.RECORD =>
+        // TODO: unclear what to do with RECORD types; they don't translate cleanly to the entity model
```
ansingh7115, line 34: For now, we could ask Product if anyone will be using the Record / Struct type and what those users use it for. 
Eventually (in the Framework implementation), I think we could translate this to some kind of JSON object.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: I don't know, and I find it frustrating. I haven't tried to look into it, though

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: I was thinking we could contact Jade and ask if they know of any examples of record types - and repeated values - and use that empirical knowledge to finish this implementation. I was also thinking `AttributeRawJson` might be the right answer.

davidangb:
most recent force-push is rebasing to develop, now that #1241 merged.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
ansingh7115: You could do `field.getType.getStandardType` and match on that instead if you'd rather not use Legacy SQL

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -0,0 +1,145 @@
...
+      case Failure(ex) =>
+        throw new DataEntityException(s"field ${field.getName} not found in row ${asDelimitedString(row)}")
+    }
+
+    val attribute = field.getMode match {
+      case Mode.REPEATED =>
+        fv.getRepeatedValue.asScala.toList match {
+          case x if x.isEmpty => AttributeValueEmptyList
+          case members =>
+            // can a list itself contain a list? Recursion here is difficult because of the Attribute/AttributeValue hierarchy
```
MatthewBemis, line 66: I am not well-versed in BigQuery (so hopefully this is even the correct documentation) but it looks like arrays of arrays are not supported per https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -0,0 +1,145 @@
...
+        List.empty[Entity]
+      } else {
+        val fieldDefs:List[Field] = schemaFields.iterator().asScala.toList
+        queryResults.iterateAll().asScala.map { row =>
+          val attrs = fieldDefs.map { field => fieldToAttribute(field, row) }.toMap
+          val entityName = attrs.find(_._1.name == primaryKey) match {
+            case Some((_, a: Attribute)) => AttributeStringifier.apply(a)
+            case None =>
+              // this shouldn't happen, since we validated the pk against the schema
+              throw new DataEntityException(s"could not find primary key column '$primaryKey' on record: ${asDelimitedString(schemaFields)}")
```
ansingh7115: This exception is the same as the one on line 94 - can we define it once and reuse?

davidangb: they're subtly different, and their wordings are subtly different. On line 94, we're checking to see if the primary key from the method arguments exists in the BigQuery schema - in other words, does the table have the requested column? That happens before looking at any of the rows, as a quick validation step.

On line 108, we're looping through and checking each individual row to see if that row contains the requested column. This should always pass, since if it's in the schema, each record should have it.

ansingh7115: Since this error is row-specific, then we should probably replace `${asDelimitedString(schemaFields)}` with `${asDelimitedString(row)}` so that users know which row is missing the primary key

ansingh7115 (approved):
Looks good - left two suggestions

MatthewBemis (approved):
This looks good to me! Didn't really go through the tests, I trust you on that :)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: good call, updated!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: @ansingh7115 do you prefer `field.getType.getStandardType`? It still returns a `LegacySQLTypeName` from `field.getType`, and then translates that via `getStandardType` ... so it just means we use a different enum in the `case` clauses, but doesn't change any functionality at all.

davidangb:
swat tests failed to get a fiab on which to run tests. Jenkins retest.

davidangb:
swat tests failed AGAIN to get a fiab on which to run tests. Jenkins retest.

davidangb:
one more time, now that we increased the fiab pool: jenkins retest

PR #1231, opened 2020-06-05
AS-395: be resilient to non-uuids in Sam resources during list-workspaces [risk: low]

davidangb:
Resilience for non-UUID resourceId values returned by Sam for the `SamResourceTypeNames.workspace` type.

Before this PR, Rawls would throw an error trying to call `UUID.fromString(...)` against these values when listing your workspaces, assuming you had access to the policy in Sam.

I searched the Rawls codebase for other instances of `SamDAO.getPoliciesForType` and this is the only relevant instance. However, there may be other use cases in Rawls where we do not properly catch exceptions on `UUID.fromString()`; looking for those is out of scope for this PR.

Alternatives:
* in `HttpSamDAO.getPoliciesForType`, if the caller asked for policies of type `workspace`, filter out non-UUID results before returning the result set. This would future-proof any other usages, but it feels too black-boxy to me.
* implement configurable validation in Sam, per resource type, and disallow creation of non-UUID policies at that level. This solution would also require cleaning up any pre-existing dirty data.


davidangb:
fiab-start failed on Thurloe; jenkins retest

MatthewBemis (approved):
Looks good to me. I also agree with going with this approach over the 2 alternatives mentioned in your PR description!

davidangb:
after passing all tests, the orch automation job hung during shutdown, timed out after 2 hours, and failed the test run. *sigh* one more time: jenkins retest

dvoet (approved):

PR #1233, opened 2020-06-12
skip tests: fix swagger defs in data-repo-only.yaml [no ticket; risk: low]

davidangb:
This fixes swagger definitions ***for the apis in data-repo-only.yaml*** and nowhere else. I think there are multiple problems with the remainder of the swagger defs; I'd like to get these in first and then figure out how to fix the others.

### DIFF
--- a/core/src/main/resources/swagger/data-repo-only.yaml
+++ b/core/src/main/resources/swagger/data-repo-only.yaml
@@ -8,57 +8,21 @@ paths:
   /api/workspaces/{workspaceNamespace}/{workspaceName}/entities:
     get:
       parameters:
-        - name: workspaceNamespace
-          in: path
-          description: The workspace namespace
-          required: true
-          schema:
-            type: string
-        - name: workspaceName
-          in: path
-          description: The workspace name
-          required: true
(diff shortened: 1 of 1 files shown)


core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -115,29 +69,24 @@ paths:
             '*/*':
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+      security:
```
zloery, line 72: This authorization block is the same for the `post` and `get` endpoints here, does the openapi spec let you pull this out a level?

zloery (approved):
LGTM, one possible change that might not be allowed by the spec

core/src/main/resources/swagger/data-repo-only.yaml
davidangb: I don't believe we can life it up one level, but we *should* put it at the top level as a default for all paths, then override the handful of paths that are different. I'd like to save that for a larger refactor, once the main api-docs.yaml is fixed up.

PR #1235, opened 2020-06-15
AS-322: upgrade google2 [risk: low]

davidangb:
This is part of AS-322 but does not complete that ticket.

We want the latest workbench-libs/google2 library, because it contains an updated BigQuery client; that client will give us pagination and a more modern approach to querying BQ.

This PR upgrades the google2 lib, with necessary compatibility fixes, but makes no other code changes. Let's keep this PR small and targeted and separate from business logic changes. I'll add the business logic for AS-322 in #1230, after this PR merges and I rebase that other one.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -517,8 +517,8 @@ object Boot extends IOApp with LazyLogging {
       googleStorage <- GoogleStorageService.resource[F](pathToCredentialJson, blocker, None, Option(serviceProject))
       httpClient <- BlazeClientBuilder(executionContext).resource
       googleServiceHttp <- GoogleServiceHttp.withRetryAndLogging(httpClient, metadataNotificationConfig)
-//      topicAdmin <- GoogleTopicAdmin.fromCredentialPath(pathToCredentialJson)
-    } yield AppDependencies[F](googleStorage, googleServiceHttp, null)
+      topicAdmin <- GoogleTopicAdmin.fromCredentialPath(pathToCredentialJson)
+    } yield AppDependencies[F](googleStorage, googleServiceHttp, topicAdmin)
   }
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -517,8 +517,8 @@ object Boot extends IOApp with LazyLogging {
       googleStorage <- GoogleStorageService.resource[F](pathToCredentialJson, blocker, None, Option(serviceProject))
       httpClient <- BlazeClientBuilder(executionContext).resource
       googleServiceHttp <- GoogleServiceHttp.withRetryAndLogging(httpClient, metadataNotificationConfig)
-      topicAdmin <- GoogleTopicAdmin.fromCredentialPath(pathToCredentialJson)
-    } yield AppDependencies[F](googleStorage, googleServiceHttp, topicAdmin)
+//      topicAdmin <- GoogleTopicAdmin.fromCredentialPath(pathToCredentialJson)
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -509,6 +509,8 @@ object Boot extends IOApp with LazyLogging {
     val googleApiUri = Uri.unsafeFromString(gcsConfig.getString("google-api-uri"))
     val metadataNotificationConfig = NotificationCreaterConfig(pathToCredentialJson, googleApiUri)
 
+    implicit val logger = Slf4jLogger.getLogger[F]
+
```
davidangb, line 513: the newer google2 lib requires this implicit Logger.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -342,7 +342,8 @@ class AvroUpsertMonitorActor(
     logger.info(s"reading ${if (decompress) "compressed " else ""}object $blobName from bucket $bucketName ...")
     val compiled = googleStorage.getBlobBody(bucketName, blobName).compile
 
-    compiled.to[Array].unsafeToFuture().map { byteArray =>
+    compiled.toList.unsafeToFuture().map { byteList =>
+      val byteArray = byteList.toArray
```
davidangb, line 346: the newer google2 lib removed `.to[T]` in favor of e.g. `.toList`

project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ object Dependencies {
 
   val googleV = "1.22.0"
   val olderGoogleV = "1.20.0"   // TODO why do we have two google versions?  GAWB-2149
-  val workbenchGoogle2V = "0.6-4df42ca"
+  val workbenchGoogle2V = "0.10-bd284e7"
```
davidangb, line 10: all the exclude rules below are a result of updating this library; its transitive dependencies caused some conflicts.

project/Dependencies.scala
```diff
@@ -19,7 +19,15 @@ object Dependencies {
   val excludeAkkaActor =        ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
   val excludeAkkaStream =       ExclusionRule(organization = "com.typesafe.akka", name = "akka-stream_2.12")
   val excludeWorkbenchModel =   ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-model_2.12")
-  val excludeWorkbenchUtil =   ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
+  val excludeWorkbenchUtil =    ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
```
davidangb, line 22: whitespace-only change

project/Merging.scala
```diff
@@ -8,6 +8,7 @@ object Merging {
     case PathList("org", "apache", xs @ _*) => MergeStrategy.last
     case PathList("com", "typesafe", xs @ _*) => MergeStrategy.last
     case PathList("com", "google", "auto", "value", xs @ _*) => MergeStrategy.last
+    case PathList("io", "sundr", xs @ _*) => MergeStrategy.last
```
davidangb, line 11: the google2 upgrade introduced an assembly conflict between classes in sundr-core and sundr-codegen. Excluding either caused problems, so I'm deduping here during assembly instead.

project/Dependencies.scala
```diff
@@ -19,7 +19,15 @@ object Dependencies {
...
+
+  val excludeBouncyCastle =     ExclusionRule(organization = "org.bouncycastle", name = s"bcprov-jdk15on")
```
MatthewBemis, line 24: interesting name lol

MatthewBemis (approved):
LGTM

PR #1240, opened 2020-06-19
fix data repo DAO base path [no ticket; risk: no]

davidangb:
#1232 broke Rawls' access to Data Repo, by telling the Data Repo DAO to use a human-readable name (e.g. `terra-dev`) as the base for data repo URLs. This PR fixes that!


zarsky-broad (approved):
Ah, I see.

PR #1243, opened 2020-06-26
remove hamm [no ticket; risk: low]

davidangb:
[Hamm](https://github.com/DataBiosphere/hamm) is an abandoned cost-tracking effort that is not in use. But, Rawls contains code to upload Hamm cost info to a bucket, where Hamm would have read it if Hamm was active. This Rawls code is chatty in the logs and pollutes the codebase. Let's remove it.

-----

remaining TODOs, for which I will file follow-on Jira tickets:
* [ ] remove Hamm's bucket and its contents, after confirming we don't want that data
* [ ] remove Hamm values from conf files


kshakir (approved):
No clue what this was... so uh LGTM

aednichols (changes requested):
I would prefer to see a "why?" in the PR summary for change tracking purposes

davidangb:
@aednichols I updated the PR summary, PTAL?

aednichols (approved):
TYVM LGTM 👍

MatthewBemis (approved):

PR #1246, opened 2020-07-01
AS-323: queryEntities supports data repo [risk: medium]

davidangb:
This PR adds support for Data Repo in the `entityQuery` API.

If the end user specifies a dataReference value for this API, via querystring, we will retrieve that reference from Workspace Manager, find its snapshot in Data Repo, and then formulate a paginated request to BigQuery to return results.

Fulltext search is not supported. If the user specifies search terms, they will get a BadRequest.

BigQuery, and its associated Java client, is lacking in support for SQL sanitization. I had to implement that myself, and very much want close inspection of that.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -183,14 +183,19 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
   def queryEntities(workspaceName: WorkspaceName, dataReference: Option[DataReferenceName], entityType: String, query: EntityQuery): Future[PerRequestMessage] = {
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
 
-      val requestArgs = EntityRequestArguments(workspaceContext, userInfo, dataReference)
+      val entityRequestArguments = EntityRequestArguments(workspaceContext, userInfo, dataReference)
 
-      entityManager.resolveProvider(requestArgs).queryEntities(entityType, query)
-        .map { response => RequestComplete(StatusCodes.OK, response) }
-        .recover {
-          case dee:DataEntityException => RequestComplete(dee.code, dee.getMessage)
-          case ex => RequestComplete(StatusCodes.InternalServerError, ex)
-        }
+      val queryFuture = for {
+        entityProvider <- entityManager.resolveProviderFuture(entityRequestArguments)
+        entities <- entityProvider.queryEntities(entityType, query)
+      } yield {
+        PerRequest.RequestComplete(StatusCodes.OK, entities)
+      }
+
+      queryFuture.recover {
+        case dee:DataEntityException => RequestComplete(dee.code, dee.getMessage)
+        case ex => RequestComplete(StatusCodes.InternalServerError, ex)
+      }
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -4,12 +4,12 @@ import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.{StatusCodes, Uri}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.entities.exceptions.{DeleteEntitiesConflictException, EntityNotFoundException}
+import org.broadinstitute.dsde.rawls.entities.exceptions.{DataEntityException, DeleteEntitiesConflictException, EntityNotFoundException}
 import org.broadinstitute.dsde.rawls.expressions.ExpressionEvaluator
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{AttributeUpdateOperation, EntityUpdateDefinition}
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport._
-import org.broadinstitute.dsde.rawls.model.{AttributeEntityReference, Entity, EntityCopyDefinition, EntityQuery, EntityQueryResponse, EntityQueryResultMetadata, ErrorReport, SamResourceTypeNames, SamWorkspaceActions, UserInfo, WorkspaceName, _}
+import org.broadinstitute.dsde.rawls.model.{AttributeEntityReference, Entity, EntityCopyDefinition, EntityQuery, ErrorReport, SamResourceTypeNames, SamWorkspaceActions, UserInfo, WorkspaceName, _}
 import org.broadinstitute.dsde.rawls.util.{AttributeSupport, EntitySupport, JsonFilterUtils, WorkspaceSupport}
 import org.broadinstitute.dsde.rawls.webservice.PerRequest
 import org.broadinstitute.dsde.rawls.webservice.PerRequest.{PerRequestMessage, RequestComplete}
@@ -44,7 +44,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
   def EvaluateExpression(workspaceName: WorkspaceName, entityType: String, entityName: String, expression: String) = evaluateExpression(workspaceName, entityType, entityName, expression)
   def GetEntityTypeMetadata(workspaceName: WorkspaceName, dataReference: Option[DataReferenceName]) = entityTypeMetadata(workspaceName, dataReference)
   def ListEntities(workspaceName: WorkspaceName, entityType: String) = listEntities(workspaceName, entityType)
-  def QueryEntities(workspaceName: WorkspaceName, entityType: String, query: EntityQuery) = queryEntities(workspaceName, entityType, query)
+  def QueryEntities(workspaceName: WorkspaceName, dataReference: Option[DataReferenceName], entityType: String, query: EntityQuery) = queryEntities(workspaceName, dataReference, entityType, query)
   def CopyEntities(entityCopyDefinition: EntityCopyDefinition, uri:Uri, linkExistingEntities: Boolean) = copyEntities(entityCopyDefinition, uri, linkExistingEntities)
   def BatchUpsertEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition]) = batchUpdateEntities(workspaceName, entityUpdates, true)
   def BatchUpdateEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition]) = batchUpdateEntities(workspaceName, entityUpdates, false)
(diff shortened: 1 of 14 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySanitizationSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  override protected def beforeAll(): Unit = {
+    // create a new variable to hold the size; this prevents Scalatest from attempting to output
+    // anything about the contents of the list itself, if the assertion fails
+    val nSize = naughtyStrings.size
+    assert(nSize > 0, "fixture data was empty; failing test. Size: " + nSize)
+
+    super.beforeAll()
+  }
```
davidangb, line 34: I create the `naughtyStrings` val at the class level, so that when Scalatest calculates its test names, the dynamic string `s"do the right things across ${naughtyStrings.size} ...` has the correct value. But I validate via a `beforeAll` that we successfully downloaded and parsed the data.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySanitizationSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.scalatest.{BeforeAndAfterAll, FreeSpec}
+import spray.json._
+
+class DataRepoBigQuerySanitizationSpec extends FreeSpec with DataRepoBigQuerySupport with BeforeAndAfterAll {
+
+  // read the list of naughty strings from https://github.com/minimaxir/big-list-of-naughty-strings.
+  // We expect a json array, containing base64-encoded strings.
+  val blnsData = Resources.toString(
+    new java.net.URL("https://raw.githubusercontent.com/minimaxir/big-list-of-naughty-strings/master/blns.base64.json"),
+    Charset.defaultCharset())
```
davidangb, line 16: reviewers: how do you feel about this test relying on downloading a third-party resource from github? we already rely on downloading third-party jars, etc. to compile, so it feels safe. I like downloading it each time, so that if it gets updated with new test cases, we automatically adopt those test cases.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -133,6 +140,86 @@ trait DataRepoBigQuerySupport {
...
+   * @param input the string to be sanitized
+   * @return the sanitized string
+   */
+  def sanitizeSql(input: String): String = {
+    if (input == null || input.isEmpty)
+      input
+    else
+      DataRepoBigQuerySupport.illegalBQChars.replaceAllIn(input, "")
+  }
+
```
davidangb, line 225: Please may I have lots of eyes on this sanitization function, plus its calling usages and test coverage?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -133,6 +140,86 @@ trait DataRepoBigQuerySupport {
...
+   * generates the SQL and config to send to BigQuery for use in the queryEntities() method
+   * @param dataProject project containing the BQ data
+   * @param viewName dataset/snapshot to query in BQ
+   * @param entityType table to query in BQ
+   * @param entityQuery user-supplied query criteria
+   * @return the object to pass to BQ to execute a query
+   */
+  def queryConfigForQueryEntities(dataProject: String, viewName: String, entityType: String, entityQuery: EntityQuery): QueryJobConfiguration = {
+    // generate BQ SQL for this entity
+    val query = s"SELECT * FROM `${sanitizeSql(dataProject)}.${sanitizeSql(viewName)}.${sanitizeSql(entityType)}` " +
```
dvoet: why do we sanitize dataProject, viewName and entityType (but we must verify that entityType exists in snapshot) since those come from data repo? For entityQuery.sortField we could verify that the field exists in the snapshot schema.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
dvoet: shouldn't we throw an exception instead of stripping?

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySanitizationSpec.scala
dvoet: given that in the unit test phase we already rely on github to download the code in the first place I am ok with the github dependency. I am a little concerned about depending on an externally controlled file that may be updated and break our tests at any time.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: I do both just for layered security. I don't think doing both HURTS - beyond the need to maintain the additional code. Do you think it's not worth it? FYI, I do already check that both the sort field and the entityType exists in the schema.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: fair point, I waffled on this choice

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySanitizationSpec.scala
MatthewBemis: It looks like they do occasionally update the list of naughty strings. IMO it might actually be a good thing, perhaps one day they'll introduce something that breaks our tests _but_ it's a legitimate breakage that we should be better protecting against. It might not be the most graceful way of finding out, but it seems like we're letting them do some work for us by adding different types of bad input. I do get the concern that maybe it disappears one day and agree that it's a shared concern with using third-party jars. 

tl;dr I'm good with this.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -74,7 +75,9 @@ class DataRepoEntityProvider(requestArguments: EntityRequestArguments, workspace
     // determine view name
     val viewName = snapshotModel.getName
     // generate BQ SQL for this entity
-    val query = s"SELECT * FROM `${dataProject}.${viewName}.${entityType}` WHERE $pk = @pkvalue;"
+    // TODO: prevent injection via dataProject, viewName, entityType vars. These are calculated from the snapshot so
```
MatthewBemis: this todo is resolved, no?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -97,6 +100,57 @@ class DataRepoEntityProvider(requestArguments: EntityRequestArguments, workspace
...
+      throw new UnsupportedEntityOperationException("term filtering not supported by this provider.")
+    }
+
+    // extract table definition, with PK, from snapshot schema
+    val tableModel = snapshotModel.getTables.asScala.find(_.getName == entityType) match {
+      case Some(table) => table
+      case None => throw new EntityTypeNotFoundException(entityType)
+    }
+
+    // determine project to be billed for the BQ job TODO: need business logic from PO!
```
MatthewBemis, line 101: what's the status of solidifying this business logic?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
MatthewBemis: +1 for throwing an exception and letting the caller decide how to fix it instead of fixing it on their behalf. sanitize and it's usage looks good to me as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -166,25 +166,17 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       }
     }
 
-  def queryEntities(workspaceName: WorkspaceName, entityType: String, query: EntityQuery): Future[PerRequestMessage] = {
+  def queryEntities(workspaceName: WorkspaceName, dataReference: Option[DataReferenceName], entityType: String, query: EntityQuery): Future[PerRequestMessage] = {
```
ansingh7115, line 183: will we be handling the billing project in a separate PR?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
ansingh7115: I support being extra cautious.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySanitizationSpec.scala
ansingh7115: Also agreed that it should generally be fine. 

In the rare case it disappears, it would a small fire - so maybe adding on something later to make this resilient to the json disappearing altogether (maybe our own list of strings) might be useful. 

Agreed that it should be allowed to fail if an update of the json to add more strings breaks our code.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
ansingh7115: sanitization looks good to me as well. I also vote for throwing an exception - stripping might end up with our code throwing an exception in a different place in a way that might be confusing to the user (user submits bad table name -> table name is stripped -> new table name can't be found in BQ -> user receives "[stripped table name] table not found error" -> user is confused why the table name in the error is different from their input)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: yes - see AS-440

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: waiting on PO; see AS-440

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: yes it is! I will remove the comment

MatthewBemis (approved):
The latest changes look good to me and cover all of my original comments. 👍

ansingh7115 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
adrazhi: @davidangb 

1. Is 64 just an arbitrary length to print out to logs/output in case illegal chars are found? 

2. I think it may be a good idea to quote user input to avoid any issues with BigQuery reserved keywords. https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#reserved_keywords. 
E.g. to use ORDER BY `${validateSql(entityQuery.sortField)}` (it's a bit of an edge case).  Another case would be when a sort field would not match the column name in the db but 
I guess that would not be so much of a problem since it'd return an error I guess.  Maybe something like: (https://henrikwarne.com/2020/03/22/secure-by-design/)

```
private static String quote(final String identifier) {
        return BIGQUERY_QUOTE_CHAR + identifier + BIGQUERY_QUOTE_CHAR;
    }
```



3. In https://github.com/broadinstitute/rawls/blob/4482e3a1ba25969341b7f6f937d13f275082e99e/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala#L15 
`[^a-zA-Z0-9\-_]` can be probably simplified to `[^\w\-]+`    (if you care :)) 

4. Where applicable (not the case on project IDs or table names), parameterized queries should be used to prevent SQLi. https://cloud.google.com/bigquery/docs/parameterized-queries

davidangb: > 1. Is 64 just an arbitrary length to print out to logs/output in case illegal chars are found?

Yes, exactly - if the input is invalid, we only output/log the beginning of the (sanitized) input, to prevent huge payloads affecting logs/output

> I think it may be a good idea to quote user input to avoid any issues with BigQuery reserved keywords.

Good point - I'll add a story to quote these inputs where they are used.

> Another case would be when a sort field would not match the column name in the db but
I guess that would not be so much of a problem since it'd return an error I guess.

Yes, this would return an error, we have business logic elsewhere to test for that.

> Where applicable (not the case on project IDs or table names), parameterized queries should be used to prevent SQLi. https://cloud.google.com/bigquery/docs/parameterized-queries

Wherever possible, we do use parameterized queries! This sanitization code is only used to validate user input in the places where BigQuery does not support parameters.

PR #1247, opened 2020-07-02
remove arrow/avro-import [no ticket; risk: low]

davidangb:
removes code that is currently unused. Once we switched over to using the new Import Service, these code paths were obsolete.

Still todo: remove these configs from rawls.conf via firecloud-develop.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -175,15 +175,6 @@ object Boot extends IOApp with LazyLogging {
         workbenchMetricBaseName = metricsPrefix
       )
 
-      // Avro upsert uses a different project for its pubsub topics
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -214,7 +205,7 @@ object Boot extends IOApp with LazyLogging {
       }
 
       val executionServiceConfig = conf.getConfig("executionservice")
-      val submissionTimeout = util.toScalaDuration(
+      val submissionTimeout = org.broadinstitute.dsde.rawls.util.toScalaDuration(
```
davidangb, line 208: not technically part of this PR; IntelliJ showed an error here until the full package was spelled out.

ansingh7115 (approved):
Looks good to me. Thanks for cleaning this up!

PR #1249, opened 2020-07-08
[CA-838] Workflow Configuration Validation for TDR snapshots

marctalbott:
WC validation without relationship validation

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionValidator.scala
@@ -1,15 +1,85 @@
 package org.broadinstitute.dsde.rawls.entities.base
 
+import akka.http.scaladsl.model.StatusCodes
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.GatherInputsResult
-import org.broadinstitute.dsde.rawls.model.{MethodConfiguration, ValidatedMethodConfiguration}
+import org.broadinstitute.dsde.rawls.model.{AttributeString, ErrorReport, MethodConfiguration, ValidatedMCExpressions, ValidatedMethodConfiguration}
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
 trait ExpressionValidator {
   /** validate a MC, skipping optional empty inputs, and return a ValidatedMethodConfiguration */
-  def validateMCExpressions(methodConfiguration: MethodConfiguration, gatherInputsResult: GatherInputsResult): Future[ValidatedMethodConfiguration]
+  def validateMCExpressions(methodConfiguration: MethodConfiguration,
+                            gatherInputsResult: GatherInputsResult)
+                           (implicit executionContext: ExecutionContext): Future[ValidatedMethodConfiguration] = {
+
+    Future {
+      val inputsToParse = gatherInputsResult.processableInputs map { mi => (mi.workflowInput.getName, AttributeString(mi.expression)) }
+      val (emptyOutputs, outputsToParse) = methodConfiguration.outputs.partition { case (_, expr) => expr.value.isEmpty }
+
+      val validated = validateMCExpressionsInternal(
+        inputs = inputsToParse.toMap,
+        outputs = outputsToParse,
+        rootEntityTypeOption = methodConfiguration.rootEntityType
+      )
+
+      // empty output expressions are also valid
+      val validatedOutputs = emptyOutputs.keys.toSet ++ validated.validOutputs
+
+      // a MethodInput which is both optional and empty is already valid
+      val emptyOptionalInputs = gatherInputsResult.emptyOptionalInputs map {
+        _.workflowInput.getName
+      }
+
+      ValidatedMethodConfiguration(methodConfiguration, validated.validInputs ++ emptyOptionalInputs, validated.invalidInputs, gatherInputsResult.missingInputs, gatherInputsResult.extraInputs, validatedOutputs, validated.invalidOutputs)
+    }
+  }
 
   /** validate a MC, skipping optional empty inputs, and return failure when any inputs/outputs are invalid */
-  def validateExpressionsForSubmission(methodConfiguration: MethodConfiguration, gatherInputsResult: GatherInputsResult): Future[Try[ValidatedMethodConfiguration]]
+  def validateExpressionsForSubmission(methodConfiguration: MethodConfiguration,
+                                       gatherInputsResult: GatherInputsResult)
+                                      (implicit executionContext: ExecutionContext): Future[Try[ValidatedMethodConfiguration]] = {
+
+    validateMCExpressions(methodConfiguration, gatherInputsResult).map { validated =>
+
+      Try {
+        if (validated.invalidInputs.nonEmpty || validated.missingInputs.nonEmpty || validated.extraInputs.nonEmpty || validated.invalidOutputs.nonEmpty) {
+          val inputMsg = if (validated.invalidInputs.isEmpty) Seq() else Seq(s"Invalid inputs: ${validated.invalidInputs.mkString(",")}")
+          val missingMsg = if (validated.missingInputs.isEmpty) Seq() else Seq(s"Missing inputs: ${validated.missingInputs.mkString(",")}")
+          val extrasMsg = if (validated.extraInputs.isEmpty) Seq() else Seq(s"Extra inputs: ${validated.extraInputs.mkString(",")}")
+          val outputMsg = if (validated.invalidOutputs.isEmpty) Seq() else Seq(s"Invalid outputs: ${validated.invalidOutputs.mkString(",")}")
+          val errorStr = (inputMsg ++ missingMsg ++ extrasMsg ++ outputMsg) mkString " ; "
+          throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, s"Validation errors: $errorStr"))
+        }
+        validated
+      }
+    }
+  }
+
+  private[local] def validateMCExpressionsInternal(inputs: Map[String, AttributeString],
+                                                   outputs: Map[String, AttributeString],
+                                                   rootEntityTypeOption: Option[String]): ValidatedMCExpressions = {
+    def validateAndPartition(m: Map[String, AttributeString], validateFunc: String => Try[Unit] ) = {
+      val validated = m map { case (key, attr) => (key, validateFunc(attr.value)) }
+      (
+        validated collect { case (key, scala.util.Success(_)) => key } toSet,
+        validated collect { case (key, scala.util.Failure(regret)) =>
+          regret match {
+            case reported: RawlsExceptionWithErrorReport => (key, reported.errorReport.message)
+            case _ => (key, regret.getMessage)
+          }
+        }
+      )
+    }
+
+    val (successInputs, failedInputs)   = validateAndPartition(inputs, validateInputExpr())
+    val (successOutputs, failedOutputs) = validateAndPartition(outputs, validateOutputExpr(rootEntityTypeOption))
+
+    ValidatedMCExpressions(successInputs, failedInputs, successOutputs, failedOutputs)
+  }
+
+  protected def validateInputExpr(rootEntityTypeOption: Option[String] = None)(expression: String): Try[Unit]
+
+  protected def validateOutputExpr(rootEntityTypeOption: Option[String])(expression: String): Try[Unit]
 }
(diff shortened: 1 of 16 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionValidator.scala
```diff
@@ -58,7 +58,8 @@ trait ExpressionValidator {
     }
   }
 
-  private def validateMCExpressionsInternal(inputs: Map[String, AttributeString],
+  // todo: made public for testing. any other way?
```
marctalbott: never resolved this or a similar comment on l. 22 of `LocalEntityExpressionValidator`

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidator.scala
```diff
@@ -0,0 +1,33 @@
...
+      parse the expression using ANTLR parser for local input expressions and walk the tree using `visit()` to examine
+      child nodes. If it finds an entityLookup node at any point, it fails unless allowRootEntity is true since
+      entity expressions are only allowed when running with the workspace data model
+     */
+    Try(terraExpressionParser.root()).flatMap(visitor.visit)
+  }
+
+  override protected def validateOutputExpr(rootEntityTypeOption: Option[String])(expression: String): Try[Unit] = {
+    val terraExpressionParser = AntlrTerraExpressionParser.getParser(expression)
+    val visitor = new LocalOutputExpressionValidationVisitor(rootEntityTypeOption)
```
dvoet: entity output expressions are not supported yet on snapshots so I think we should be passing in `None` to `LocalOutputExpressionValidationVisitor`

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityExpressionValidator.scala
```diff
@@ -1,97 +1,33 @@
...
+  // public for testing. any other way?
+  override def validateOutputExpr(rootEntityTypeOption: Option[String])(expression: String): Try[Unit] = {
+    val terraExpressionParser = AntlrTerraExpressionParser.getParser(expression)
+    val visitor = new LocalOutputExpressionValidationVisitor(rootEntityTypeOption)
+
+    for {
+      parseTree <- Try(terraExpressionParser.root())
+      _ <- visitor.visit(parseTree)
+      _ <- OutputExpression.validate(expression, rootEntityTypeOption)
+    } yield ()
```
dvoet: is this duplicate of what is in the data repo version?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidator.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.broadinstitute.dsde.rawls.entities.base.ExpressionValidator
+import org.broadinstitute.dsde.rawls.expressions.OutputExpression
+import org.broadinstitute.dsde.rawls.expressions.parser.antlr.{AntlrTerraExpressionParser, DataRepoInputExpressionValidationVisitor, LocalOutputExpressionValidationVisitor}
+
+import scala.util.Try
+
+class DataRepoEntityExpressionValidator(snapshotModel: SnapshotModel) extends ExpressionValidator {
+  override protected def validateInputExpr(rootEntityTypeOption: Option[String])(expression: String): Try[Unit] = {
+    val terraExpressionParser = AntlrTerraExpressionParser.getParser(expression)
+    val visitor = new DataRepoInputExpressionValidationVisitor(rootEntityTypeOption, snapshotModel)
```
dvoet, line 13: data repo input expressions are invalid if `rootEntityTypeOption` is `None`. That validation case should happen here, don't pass an `Option` to `DataRepoInputExpressionValidationVisitor` then the visitor code is simplified.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -0,0 +1,85 @@
...
+      case Nil => Success(currentTableModel)
+    }
+  }
+
+  // TODO: this is checking that the `fromTable` exists and that the relationship exists in one spot.  Not sure if we need a separate check for the table and another for the relationship
+  // Maybe we don't need to do all that though...it feels like we're getting into the territory of validating the TDR
+  // "relationship" schema at some point.  Maybe it's ok to just say, "If a relationship is defined with name X and
+  // from_table FOO, then we're going to assume that the RelationshipModel object is valid for this SnapshotModel."
+  private def maybeGetNextTableFromRelation(fromTable: TableModel, relationName: String): Option[TableModel] = {
+    maybeFindTableInSnapshotModel(fromTable.getName)
```
dvoet: I think until we implement relations this should just fail

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -0,0 +1,85 @@
...
+      // TODO: CA-939 implement relationship validation
+      // .map { tableModel =>
+      // if (relationship exists with name `relationName` with from_table `fromTable`) then
+      //     return Option(getRelationship(relationName).to_table())
+      // else
+      //     None
+      // }
```
dvoet: this pseudocode is not quite right... expressions can follow relationships bidirectionally

dvoet (changes requested):
some tests please

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityExpressionValidator.scala
marctalbott: At the moment, yes it appears to be just a duplicate. I could move it up into `ExpressionValidator`. edit: well I guess not strictly duplicated since we actually do care about passing `rootEntityTypeOption` for local entities and we will just be passing `None` for data repo entities.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
marctalbott: I'm not sure I fully understand what you mean by this, are there examples of relationship models and expressions that would illustrate it?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidator.scala
marctalbott: I think that would require us to do some checking that the input expression we are working with at this point is an entity expression. As I understand it, we don't know whether `expression` is an entity expression or some other input expression at this point. Passing the option to the visitor allows the visitor to only fail if the `rootEntityTypeOption` is `None` and `expression` is an entity expression.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
dvoet: https://docs.google.com/document/d/122Qu90bouAlfFUBeQwhjOGn3I2KOSqWY42uKG-9wW4I/edit#bookmark=id.dj2llc7trzea

gpolumbo-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
```diff
@@ -0,0 +1,193 @@
...
+    methodConfiguration.inputs.map(toMethodInput).toSeq
+  }
+
+  def toGatherInputs(exprs: Map[String, AttributeString]): GatherInputsResult = {
+    val methodInputs = exprs.map(toMethodInput)
+    GatherInputsResult(methodInputs.toSet, Set(), Set(), Set())
+  }
+
+  val rootTableName = "rootTable"
+  // todo: how to keep this in sync with the fixtures...
```
marctalbott: Does anyone have thoughts on this? These columns need to match up with the input expressions defined in the fixture in order for the tests to validate the input expressions correctly. Other than just leaving them here and hoping that they stay in sync, my other idea was adding some of these vals to `DataRepoExpressionFixture` (`rootTableName`, `rootTableColumns`, and `tables`) so that they're at least all defined in the same place. We could take it one step further and try to construct the lists of valid and invalid expressions using the values in `rootTableColumns` but that'd complicate things a bit

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  val emptyExpr = "this.empty" -> AttributeString("")
+
+  val oneEmpty = MethodConfiguration("dsde", "methodConfigEmptyExpr", Some(rootTableName), prerequisites=None,
+    inputs = toExpressionMap(validInputExpressions) + emptyExpr,
+    outputs = toExpressionMap(validOutputExpressions),
+    AgoraMethod("dsde", "three_step", 1))
+
+
+  it should "validateAndParseMCExpressions" in {
```
gpolumbo-broad: Some nits here seeing as we're creating this file now.  I know it was copied from the existing validator spec, but easier to clean this up now than later (if we want it cleaned up at all):

- rename `validateAndParseMCExpressions` - if anything, I would say we're just validating and the parsing is just a necessary component of validating.  Anyway, naming of tests would become more expressive if we do the following too:
- Can we break each assertion block below into its own test?  It seems unnecessary to have them all in one test.  If an assertion fails, processing of remaining tests won't happen.  Maybe that's what we want, but maybe not.  

Fully acknowledging these are nits and if you don't want to worry about them now, that's fine.

dvoet (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
dvoet: I would just move things to the fixture to keep them in the same place

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidator.scala
gpolumbo-broad: Or do we want to stub out `DataRepoOutputExpressionValidationVisitor` and just have that return failures for now?

gpolumbo-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
marctalbott: Good point, I will rename `validateAndParseMCExpressions` in the test. It's named after the function in `ExpressionValidator` which we have since changed to `validateMCExpressions`, so no point in keeping around the old and inaccurate function name.

As for breaking up the assertions, I think that's a good idea. It'll let us better describe what each assertion is meant to test and that'll make test failures more descriptive as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidator.scala
marctalbott: My only hesitation with writing a separate `DataRepoOutputExpressionValidationVisitor` is that it'd be a duplicate of `LocalOutputExpressionValidationVisitor` except with the `visitEntityLookup` changed to always return a Failure. We could write a different `OutputExpressionValidationVisitor` that handles the shared code (`defaultResult`, `aggregateResult`, workspace attributes and JSON) and just have the Local and DataRepo versions extend that and override `validateEntityLookup`

PR #1250, opened 2020-07-10
eager/early validation of snapshot UUID strings [no ticket; risk: no]

davidangb:
for snapshot GET and DELETE apis, perform early validation of the input UUID strings, before querying for workspaces.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -2,18 +2,20 @@ package org.broadinstitute.dsde.rawls.snapshot
 
 import java.util.UUID
 
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import bio.terra.workspace.model.DataReferenceDescription.{CloningInstructionsEnum, ReferenceTypeEnum}
 import com.google.api.client.auth.oauth2.Credential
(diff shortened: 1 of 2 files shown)


davidangb:
@MatthewBemis @zloery I put this together a while ago and forgot about it ... what do you think of this change?

zloery (approved):
LGTM, it's good to validate UUID strings as early as possible imo. Do you happen to know if there are other places we're calling stuff like `"${testData.wsName.path}/snapshots/doesntmatter"`?
One of the security-related changes I'm adding is switching to accepting UUIDs instead of taking strings and validating, so tests like that will start getting 400s.

MatthewBemis (approved):
I don't feel too strongly about this. I get wanting to know before we make a call to WSM so I certainly approve of it from that standpoint but it does end up duplicating validation logic that already exists in WSM

I give an indifferent thumb! :)

davidangb:
@MatthewBemis while you're correct this does duplicate validation that exists in WSM, Rawls contains tacit/indirect validation. Before this PR, we used `UUID.fromString(snapshotId)` when constructing the call to `workspaceManagerDAO.getDataReference` - so Rawls would throw an non-user-friendly error on invalid UUIDs before ever calling WSM. After this PR, Rawls throws a user-friendly error. I'd say we either need this PR, _or_ we pass the id on to WSM as a String and never try to turn it into a UUID.

@zloery I'm not aware of other places - but I haven't done an exhaustive search.

MatthewBemis:
Fair enough! I amend my previous indifference and am now in favor of this :)

PR #1254, opened 2020-07-20
AS-456: read batchupsert as stream [risk: low]

davidangb:
https://broadworkbench.atlassian.net/browse/AS-456

This PR:
* removes the code that attempted to read and parse potentially-large batchUpsert json files as a whole file
* replaces that code with a stream, with which we pull N entity upserts at a time and process them, then keep streaming through the file (N is 1000)
* adds unit test coverage
* adds logging statements to help with forensic debugging (it's almost too chatty now)
* adds `try/catch` and `recover` around the `initUpsert` method so we should now always fail any job with an error, mark it as Failed for the end user, and ack the relevant pubsub message that started the import
* adds detail to the responses we send to the user, indicating successes and failures

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -224,7 +224,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       }
     }
 
-  def batchUpdateEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition], upsert: Boolean = false): Future[PerRequestMessage] = {
+  def batchUpdateEntitiesInternal(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition], upsert: Boolean = false): Future[Traversable[Entity]] = {
     val namesToCheck = for {
       update <- entityUpdates
       operation <- update.operations
@@ -262,11 +262,16 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
             }
           }
 
-          saveAction.map(_ => RequestComplete(StatusCodes.NoContent))
+          saveAction
         }
       }
     }
   }
+
+  def batchUpdateEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition], upsert: Boolean = false): Future[PerRequestMessage] = {
+    batchUpdateEntitiesInternal(workspaceName, entityUpdates, upsert).map(_ => RequestComplete(StatusCodes.NoContent))
+  }
+
 
   /**
     * Applies the sequence of operations in order to the entity.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -4,15 +4,18 @@ import java.util.UUID
 
 import akka.actor.SupervisorStrategy.{Escalate, Stop}
 import akka.actor._
+import akka.http.scaladsl.model.StatusCodes
 import akka.pattern._
 import cats.effect.{ContextShift, IO}
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -262,12 +262,17 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
         }
       }
     }
   }
 
+  def batchUpdateEntities(workspaceName: WorkspaceName, entityUpdates: Seq[EntityUpdateDefinition], upsert: Boolean = false): Future[PerRequestMessage] = {
+    batchUpdateEntitiesInternal(workspaceName, entityUpdates, upsert).map(_ => RequestComplete(StatusCodes.NoContent))
+  }
+
+
```
davidangb, line 275: I factored this out into two methods, so that the PFB-import code can call `batchUpdateEntitiesInternal` and get something other than a `PerRequestMessage` back.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -256,24 +278,61 @@ class AvroUpsertMonitorActor(
...
+      }.sum
+
+      val failureReports:List[RawlsErrorReport] = upsertResults collect {
+        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport.causes
+      } flatten
+
+      // fail if nothing at all succeeded
+      if (numSuccesses == 0) {
+        // TODO: this could be a LOT of error reports, should we omit if large?
+        throw new RawlsExceptionWithErrorReport(RawlsErrorReport(StatusCodes.BadRequest, "All entities could not be updated.", failureReports))
```
aednichols: We used to have problems with big errors In Cromwell and found `take(100)` is a good balance between having something in the logs and not exploding

davidangb: hah! I was writing the `take(100)` implementation in the background and came back to see this comment ... perfect, thank you

aednichols: Haha nice!

MatthewBemis (approved):
everything LGTM

ansingh7115 (approved):
Great! I couldn't find much to criticize

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -256,23 +262,101 @@ class AvroUpsertMonitorActor(
...
+          logger.error(s"upsert process for $jobId FAILED after $elapsed ms: ${ex.getMessage}")
+          acknowledgeMessage(ackId) // just in case
+          Future.failed(ex)
+      }
+    } catch {
+      // try/catch for any synchronous exceptions, not covered by a Future.recover(). Potential for retry here, in case of transient failures
+      case ex:Exception =>
+        val elapsed = System.currentTimeMillis() - startTime
+        logger.error(s"upsert process for $jobId FAILED after $elapsed ms: ${ex.getMessage}")
+        acknowledgeMessage(ackId) // just in case
```
ansingh7115, line 358: We're calling acknowledgeMessage 3 times - once in the Try, once in the recoverWith and once in the catch. Could we call acknowledgeMessage once before the Try instead? Does it need to happen after getting the upsertStream?

davidangb: The first call, just after getting the stream, is where we originally *wanted* to call acknowledgeMessage. The calls in recover and catch are there just to be super safe ... if something goes wrong unexpectedly that we didn't otherwise plan for, we should ack the message so we don't have all these repeated "double delivery" log messages. That said ... lmk if you think we should _remove_ the calls in catch/recover, so that we would retry on those types of failures. I'm going to merge the PR as-is to move on to manual testing in dev, but could easily do a follow-on PR that changes where we ack.

PR #1255, opened 2020-07-22
AS-456: serialize upserts via Await [risk: low]

davidangb:
#1254 introduced a bug wherein batches of upserts execute in parallel, potentially meaning that later inserts fail due to unsatisfied references.

This is a less-than-ideal fix, using Await, to ensure that the codebase functions correctly until a more optimal fix can land. #1256 is preferable to this one, I have made this one just in case other fixes take too long. We don't want to release #1254 as-is.


davidangb:
closing in favor of #1256

PR #1256, opened 2020-07-22
AS-456: upsert batches sequentially via pause/resume of the stream [risk: low]

davidangb:
#1254 introduced a bug wherein batches of upserts execute in parallel, potentially meaning that later inserts fail due to unsatisfied references.

This PR fixes the bug by pausing stream processing until each upsert finishes, then resuming the stream.


MatthewBemis (approved):
Cool stuff!

ansingh7115 (approved):
Thanks - I spent a little bit of time reading about SignallingRef!

PR #1257, opened 2020-07-23
AS-456: attempt to stream harder [risk: low]

davidangb:
I suggest ignoring whitespace when viewing the diff!

This PR switches from using `map` to `evalMapChunk`, which indicates to fs2 that this is a side-effectful stream (where the side effect is the database write).

Now that it is an effectful stream, we use
```
val upsertResults = upsertFuturesStream.compile.toList.unsafeRunSync()
```
 to get the stream results, instead of
```
upsertFuturesStream.compile.toList.unsafeRunSync().sequence.map { upsertResults =>
```

this moves more of the processing inside fs2, which hopefully means it is more optimized/uses less memory, etc.


ansingh7115 (approved):
I can see nothing wrong with it

PR #1258, opened 2020-07-24
AS-456: read batchUpsert via fs2-blobstore, not workbench-libs [risk: low]

davidangb:
Use [fs2-blobstore](https://github.com/fs2-blobstore/fs2-blobstore#gcsstore) to read the batchUpsert json file as a true stream, instead of using workbench-libs, which reads the entire file into memory.

Accompanying changes to make various constructors and unit tests work


MatthewBemis (approved):
:shipit:

davidangb:
closing in favor of #1259

PR #1259, opened 2020-07-27
AS-456: use workbench libs streaming [risk: low]

davidangb:
A few changes in this PR:
1. Use true GCS blob streaming, by adopting the latest version of workbench-libs that incorporates broadinstitute/workbench-libs#333.
2. During the batchUpsert loop, send keepalive messages back to the actor. Otherwise, if the loop takes > 20 minutes (approx), the actor will kill itself because it hasn't received a message in too long. Now, it will only kill itself if a single batch of 1000 operations takes > 20 minutes.
3. During the batchUpsert loop, use a pet SA to perform each batch of db writes. Otherwise, if the total loop takes > 60 minutes (approx), the user's credentials will expire. Now, this limitation should be eliminated.


MatthewBemis (approved):
Looks good!

PR #1264, opened 2020-07-29
cleaner error messages for data repo integration [no ticket; risk: low]

davidangb:
some loose-end cleanup in this PR, in which we return cleaner error responses - including leaving out unhelpful stack traces - when calling the entityQuery API and:
* specifying a snapshot reference name that could not be found
* specifying an entityType that could not be found inside a snapshot

Previously, we returned stack traces that only included Akka forkJoin nonsense, and some of the exceptions were wrapped multiple times and then toString-ed, making it very hard to find the real messages. We also returned `InternalServerError` for these, when they really should be `BadRequest`.


davidangb:
one failure on `org.broadinstitute.dsde.test.api.AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain when the user is in all of the groups when shared with them can be seen and is accessible`; jenkins retest

davidangb:
second swat run, everything passed except one failure on `org.broadinstitute.dsde.test.api.RawlsApiSpec.Rawls should retrieve sub-workflow metadata and outputs from Cromwell`. Between the two runs and knowledge of what changed in this PR, I have confidence.

davidangb:
@dvoet do you have opinions on how this affects responses for expression/workflow related APIs? (I don't know how to test those - but see the change to unit tests)

PR #1267, opened 2020-08-06
[CA-939] validate method config TDR entity relationships

andy7i:
https://broadworkbench.atlassian.net/browse/CA-939

TDR now supports Relationships, which means that we can traverse tables in BQ. This PR provides the validation for relationships in MC expressions.

Note: `namespace: name` notation is not supported in BQ so this was removed.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -19,45 +22,72 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
...
+    maybeFindTableInSnapshotModel(fromTable.getName)
+      .flatMap { tableModel =>
+        relationships.find { relationship =>
+          relationship.getFrom.getTable == tableModel.getName && relationship.getFrom.getColumn == relationName
+        }.map { relationship =>
+          val nextTableName = relationship.getTo.getTable
+          return maybeFindTableInSnapshotModel(nextTableName)
+        }
+      }
+  }
```
dvoet: 3 things here:
1) try rewriting this as a for comprehension. More than one flapMap/map works better as a for comp
2) this needs to match the relationship in both directions. Either `relationship.getFrom.getTable == tableModel.getName` or `relationship.getTo.getTable == tableModel.getName`
3) `relationship.getFrom.getColumn == relationName` is not right, the relation name _is not_ the column name

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -22,7 +25,32 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
...
+                ErrorReport(StatusCodes.BadRequest, "Expressions with \"namespace: name\" are not valid for BigQuery")
+              ))
+          }
+          case None => Failure(new RawlsExceptionWithErrorReport(
+            ErrorReport(StatusCodes.BadRequest, "Expressions with \"namespace: name\" are not valid for BigQuery")
+          ))
+        }
+
+      }
+      case None => super.visitNamespace(ctx)
```
dvoet: sorry, I gave you bad advice... I forgot we need to allow namespaces on workspace expressions.

The better place to do this is in `validateEntityLookup` where you can check `entityLookupContext.attributeName().namespace()`.

The code that looks like `ctx.parent.parent.getChild(0).getText` is the part that is concerning to me here. Also the checking for "workspace.". Both feel like brute force approaches that depend on the grammar structure but don't use the parseTree. If the grammar changes in the future these things may break and the compiler will miss it.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -22,19 +24,23 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
...
-      case Some(rootTableModel) => {
+    (maybeFindTableInSnapshotModel(rootTableName), Option(entityLookupContext.attributeName().namespace())) match {
+      case (Success(rootTableModel), None) =>
         val relations = entityLookupContext.relation().asScala.toList
         traverseRelationsAndGetFinalTable(rootTableModel, relations).flatMap(finalTable => checkForAttributeOnTable(finalTable, entityLookupContext.attributeName()))
-      }
-      case None => Failure(new RawlsException(s"Root entity type [$rootTableName] is not a name of a table that exist within DataRepo Snapshot."))
+      case (Failure(regrets), _) => Failure(regrets)
+      case (_, Some(_)) => Failure(new RawlsExceptionWithErrorReport(
+        ErrorReport(StatusCodes.BadRequest, "Expressions with \"namespace: name\" are not valid for BigQuery")
```
dvoet: say "not valid with snapshots" in stead of "not valid for BigQuery"

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -44,20 +50,55 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
     if (tableColumns.exists(_.getName == attributeName)) {
       Success()
     } else {
-      Failure(new RawlsException(s"Missing attribute `${attributeName}` on table `${tableModel.getName}`"))
+      Failure(new RawlsExceptionWithErrorReport(
+        ErrorReport(StatusCodes.BadRequest, s"Missing attribute `${attributeName}` on table `${tableModel.getName}`")
```
dvoet: error message should be `s"Column ${attributeName} does not exist on table ${tableModel.getName}"`

dvoet (approved):
a couple wording changes on error messages then 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -22,19 +24,23 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
   override def visitEntityLookup(ctx: EntityLookupContext): Try[Unit] = {
     rootEntityType match {
       case Some(rootTableName) => validateEntityLookup(rootTableName, ctx)
-      case None => Failure(new RawlsException("Expressions beginning with \"this.\" are only allowed when running with workspace data model. However, workspace attributes can be used."))
+      case None => Failure(new RawlsExceptionWithErrorReport(
+        ErrorReport(StatusCodes.BadRequest, "Expressions beginning with \"this.\" are only allowed when running with workspace data model. However, workspace attributes can be used.")
```
andy7i, line 28: // it seems like it would get here only if methodConfiguration.rootEntityType is not set, so this error message doesn't make sense to me. Shouldn't it be something like "Missing Root Entity Type. Please select one and try again."

gpolumbo-broad: I had to think about this too.  However, I think the point here is that here we're validating an individual Expression, not the MethodConfiguration.  And in particular, we know we're validating a DataRepo Expression, so I think this statement is ok.  Whether it could be wordsmithed for clarity is something maybe product or UX could help us understand if users know what the problem is and how to fix it when they see this message in context.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -22,19 +24,23 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
...
+      ))
     }
   }
 
   // Valid DataRepo EntityLookups mean that each of the relationships exist and the final attribute exists as a column
   // on the final table
   private def validateEntityLookup(rootTableName: String, entityLookupContext: EntityLookupContext): Try[Unit] = {
-    maybeFindTableInSnapshotModel(rootTableName) match {
-      case Some(rootTableModel) => {
+    (maybeFindTableInSnapshotModel(rootTableName), Option(entityLookupContext.attributeName().namespace())) match {
```
gpolumbo-broad: I think this would read clearer as two separate checks.  First check if there's a namespace specified and if it's valid, then continue on to `maybeFindTableInSnapshotModel()`

I don't know if the FP/Scala purists agree with that approach, but I prefer the style of breaking down each step and calling it out explicitly.  I might even break out the namespace validation into its own method so I can name it, `validateEntityNamespace` to be explicitly clear about what that step is doing.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
```diff
@@ -44,20 +50,55 @@ class DataRepoInputExpressionValidationVisitor(rootEntityType: Option[String],
...
   }
 
-  // TODO: CA-939 support relationships once that information is published by data repo
   private def traverseRelationsAndGetFinalTable(currentTableModel: TableModel, relations: List[RelationContext]): Try[TableModel] = {
     relations match {
       case Nil => Success(currentTableModel)
-      case _ => Failure(new RawlsException(s"Relationships are not currently supported."))
+      case nextRelationContext :: remainingRelations => {
+        val nextRelationName = nextRelationContext.attributeName.name.getText
+        maybeGetNextTableFromRelation(currentTableModel, nextRelationName).flatMap(traverseRelationsAndGetFinalTable(_, remainingRelations))
```
gpolumbo-broad, line 78: I don't think this is tail recursive, unless scala does some magic with `flatMap` behind the scenes to make it so.

https://alvinalexander.com/scala/fp-book/tail-recursive-algorithms/

I don't think we need to worry about it here since most expressions I think would only result in a handful (literally single digits) of recursive calls.  That said, I would say it is good practice to always try to make your recursive functions tail recursive.  

You can try adding the `@tailrec` annotation to this method to see what the compiler thinks, maybe scala knows how to do something special with `flatMap` to make it tail recursive, but I don't know.  The way I'm reading this it is the flatten part of the `flatMap` that is going to cause the problem here.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
```diff
@@ -61,6 +62,30 @@ class DataRepoEntityExpressionValidatorSpec extends FlatSpec with TestDriverComp
...
     AgoraMethod("dsde", "three_step", 1))
 
+  /* Set up for Relationships */
+  val toRootTableAndColumn: RelationshipTermModel = new RelationshipTermModel().table(rootTableName).column("root_table_column")
+  val toSecondTableAndColumn: RelationshipTermModel = new RelationshipTermModel().table(secondTableName).column("second_table_column")
+  val toThirdTableAndColumn: RelationshipTermModel = new RelationshipTermModel().table(thirdTableName).column("third_table_column")
+  val fromRootTableAndColumnToSecondTable: RelationshipTermModel = new RelationshipTermModel().table(rootTableName).column("root_second")
+  val fromSecondTableAndColumnToThirdTable: RelationshipTermModel = new RelationshipTermModel().table(secondTableName).column("second_third")
+  val fromSecondTableAndColumnToRootTable: RelationshipTermModel = new RelationshipTermModel().table(secondTableName).column("second_root")
+  val fromThirdTableAndColumnToSecondTable: RelationshipTermModel = new RelationshipTermModel().table(thirdTableName).column("third_second")
```
gpolumbo-broad: I'm having a hard time with these names and making sense out of what the schema looks like in my head.  Maybe describe the schema some other way in a comment?  And/or rename the columns?  Maybe consider some testing schema that we inherently already understand, like a couple of table of Books and Authors and Author Addresses or something like that, that the reader of this will inherently have some understanding of how it all ties together.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
```diff
@@ -116,6 +141,33 @@ class DataRepoEntityExpressionValidatorSpec extends FlatSpec with TestDriverComp
...
+    actualValid.invalidInputs shouldBe 'empty
+    actualValid.invalidOutputs shouldBe 'empty
+  }
+
+  // should never get here if TDR is not buggy
+  it should "fail if the linked table does not exist for relationship traversals" in {
+    val providerWithMultipleTables = createTestProvider(snapshotModel = createSnapshotModel(defaultFixtureTables, relationships))
+    val expressionValidatorWithMultipleTables: ExpressionValidator = providerWithMultipleTables.expressionValidator
+
+    val actualValid = expressionValidatorWithMultipleTables.validateMCExpressions(allValidWithRelationships, toGatherInputs(allValidWithRelationships.inputs)).futureValue
```
gpolumbo-broad: Naming this `actualValid` has tripped me a couple of times reading these tests.  Consider renaming to something like `validationResults`

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityExpressionValidatorSpec.scala
```diff
@@ -116,6 +141,33 @@ class DataRepoEntityExpressionValidatorSpec extends FlatSpec with TestDriverComp
...
+    actualValid.invalidInputs.size shouldBe 23
+    actualValid.invalidOutputs shouldBe 'empty
+  }
+
+  it should "fail if the relationship does not exist for relationship traversals" in {
+    val providerWithMultipleTables = createTestProvider(snapshotModel = createSnapshotModel(multipleTables, List.empty))
+    val expressionValidatorWithMultipleTables: ExpressionValidator = providerWithMultipleTables.expressionValidator
+
+    val actualValid = expressionValidatorWithMultipleTables.validateMCExpressions(allValidWithRelationships, toGatherInputs(allValidWithRelationships.inputs)).futureValue
+    actualValid.invalidInputs.size shouldBe 23
```
gpolumbo-broad: Same as above re: magic number

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
```diff
@@ -34,22 +34,52 @@ trait DataRepoExpressionFixture {
...
-    "this.library:cohort",
-    "this.library:cohort1",
-    "this.arbitrary:whatever",
     "this.underscores_are_ok",
     "this._",
     """["foo","bar", this.gvcf]""",
     """["a",{"more":{"elaborate":this.gvcf}}]""",
     """{"more":{"elaborate":{"reference1": this.gvcf, "path":"gs://abc/123"}}}""",
-    """["foo", "bar", 123, ["array", this.gvcf, this.library:cohort], false]""",
-    """["foo", "bar", 123, ["array", this.gvcf, [this.library:cohort]], false, ["abc", this.with-dash]]"""
```
gpolumbo-broad, line 46: Why were these removed (and the others up on lines 37-39)?  I know that the namespaced attributes are not valid for DataRepo expressions, but these are valid expressions for the LocalEntityProvider.   This means we might need to creates some different lists that build for DataRepo vs. Local expressions.

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
```diff
@@ -66,8 +96,17 @@ trait DataRepoExpressionFixture {
   // parseable input expressions that are invalid (don't fit the schema, relations, any other reason?)
   val invalidInputExpressions: Seq[String] = Seq(
     "this.column_does_not_exist",
-    "this.case_sample.foo:ref.bar:attribute", // invalid for now -- relations
-    """{"level1": "easy", "other-levels": {"level2": this.gvcf, "level3": [this.library:cohort, "extremely difficult", this.library:cohort.entity]}}""", // invalid -- relations
+    "this.relationship_does_not_exist.second_table_column",
+    "this.rootTableToSecondTable.column_does_not_exist",
+    "this.library:cohort", // namespace:name is not allowed for BQ column/table names
+    "this.library:cohort1",
+    "this.arbitrary:whatever",
```
gpolumbo-broad, line 103: Again, these are only invalid TDR Entity Expressions

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
```diff
@@ -116,10 +155,26 @@ trait DataRepoExpressionFixture {
...
     new TableModel().name(defaultFixtureRootTableName).columns(defaultFixtureRootTableColumns.map(new ColumnModel().name(_)).asJava)
   )
-}
\ No newline at end of file
+
+  val rootTableName = "rootTable"
+  val rootTableColumns = List("root_table_column", "gvcf", "with-dash", "underscores_are_ok", "_", "case_sample", "root_second")
+
+  val secondTableName = "secondTable"
+  val secondTableColumns = List("second_table_column", "second_third", "second_root")
```
gpolumbo-broad: If possible, would like to rename these columns again, as I struggle with this naming convention

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
andy7i: slack conversation about this: https://broadinstitute.slack.com/archives/GSHUGJ5RD/p1596826199127100

I'm okay with leaving it for now. It seemed off to me so I just wanted to make sure it was ok

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
andy7i: we actually had it as 2 separate checks at first, but opted for this approach to consolidate the various error states. This way it's more clear what the behavior is when it succeeds and when it fails. 

I do see your point about it being tougher to read! I think breaking out validateEntityNamespace would help make it easier to read and I'll give that a shot.

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoInputExpressionValidationVisitor.scala
andy7i: yeah, it did have the tail recursion (that you added?) there before, but with the flatmap it gave an error:

```
Error:(62, 5) could not optimize @tailrec annotated method traverseRelationsAndGetFinalTable: it contains a recursive call not in tail position
    relations match {
```

Doug and I removed it since we figured (as you point out) that most expressions should be pretty short.

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
andy7i: I think they are separated. There's a Fixture for local expressions here: https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/LocalExpressionFixture.scala

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
gpolumbo-broad: I'm still concerned about the fact that we added these namespaced expressions to the list of invalid expressions because I think these same fixtures are used in both TDR and Local Entity Expression tests

gpolumbo-broad (commented):
Maybe we need to get together with @marctalbott too and talk about how to sort out the valid/invalid expressions in the fixtures?

core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/DataRepoExpressionFixture.scala
marctalbott: There's a different ExpressionFixture file called LocalExpressionFixture.scala so we can differentiate between what's valid for TDR and what's valid for the local entity provider

andy7i: Maybe I'm missing something, but Intellij tells me that `DataRepoExpressionFixture` is only used by `DataRepoEntityExpressionValidatorSpec` and `LocalEntityLocalExpressionValidatorSpec` uses `LocalExpressionFixture`. Is there something else that I am not considering?

andy7i:
jenkins retest -- `bucket did not become readable`

gpolumbo-broad (approved):

PR #1271, opened 2020-08-18
update wsm client to 0.2.0 [AS-391 part 2a]

zarsky-broad:
See https://github.com/DataBiosphere/terra-workspace-manager/pull/85/; the real change here is making DataRepoSnapshot an actual type instead of a string.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -1,35 +1,85 @@
...
+      SNAPSHOT -> JsString(snap.getSnapshot)
+    )
+
+    def read(json: JsValue) = {
+      json.asJsObject.getFields(INSTANCE_NAME, SNAPSHOT) match {
+        case Seq(JsString(instanceName), JsString(snapshot)) =>
+          new DataRepoSnapshot().instanceName(instanceName).snapshot(snapshot)
+        case _ => throw new DeserializationException("DataRepoSnapshot expected")
+      }
+    }
```
davidangb, line 37: good call on using the spray-json de/serializers here, this keeps Rawls consistent in its approaches.

I am a bit wary about the lack of null handling. For writes, you might want to create a reusable method at the `DataReferenceModelJsonSupport ` object level:
```
  private def stringOrNull(in: String): JsValue = Option(in) match {
    case None => JsNull
    case Some(s) => JsString(s)
  }
```
and then your writes can take advantage of it:
```
    def write(snap: DataRepoSnapshot) = JsObject(
      INSTANCE_NAME -> stringOrNull(snap.getInstanceName),
      SNAPSHOT -> stringOrNull(snap.getSnapshot)
    )
```
you could extend `stringOrNull` to include `UUID`s or other classes that need serialization.

For reads, the concern is that if any of the json values are null, the `match` clauses won't match. For instance, if instanceName is null, then
```
json.asJsObject.getFields(INSTANCE_NAME, SNAPSHOT) match {
        case Seq(JsString(instanceName), JsString(snapshot)) =>
```
will not match, because you'll have a `Seq(JsNull, JsString)` instead of a `Seq(JsString, JsString)`. This might be what we WANT - to throw a `DeserializationException` if anything is null - but I want to make sure that's the intended behavior.

(this comment applies to all the new `*Format` classes in this PR)

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -1,35 +1,85 @@
...
+    def write(snap: DataRepoSnapshot) = JsObject(
+      INSTANCE_NAME -> JsString(snap.getInstanceName),
+      SNAPSHOT -> JsString(snap.getSnapshot)
+    )
+
+    def read(json: JsValue) = {
+      json.asJsObject.getFields(INSTANCE_NAME, SNAPSHOT) match {
+        case Seq(JsString(instanceName), JsString(snapshot)) =>
+          new DataRepoSnapshot().instanceName(instanceName).snapshot(snapshot)
+        case _ => throw new DeserializationException("DataRepoSnapshot expected")
```
davidangb: nit: in Scala, if you are creating an instance of a case class, you don't need - and it's idiomatic to omit - the `new` keyword:
```
case _ => throw DeserializationException("DataRepoSnapshot expected")
```
( this applies elsewhere in the PR too )

davidangb (commented):
see inline!

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
zarsky-broad: Added null handling for writing; I think we explicitly want to treat nulls as an error when reading, yes.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/JsonSupport.scala
```diff
@@ -199,4 +203,13 @@ class JsonSupport {
       }
     }
   }
+
+  implicit object UUIDFormat extends JsonFormat[UUID] {
+    override def write(uuid: UUID) = JsString(uuid.toString)
+    override def read(json: JsValue): UUID = json match {
+      case JsString(str) =>
+        Try(UUID.fromString(str)).fold[UUID](t => throw DeserializationException(s"Couldn't parse $str into a UUID: ${t.toString}"), identity)
+      case x => throw DeserializationException(s"UUID can only be parsed from a string, got $x")
```
davidangb: a few comments:
1) As a rule, we are trying to avoid echoing user input in error messages, as it can lead to security issues. Since this is a reusable format, it could easily end up being used in a way that echoes user input. I'd leave `$str` out of the error message. Or to help with developer usability, maybe we only output `${str.take(8)}`.
2) the previous comment is especially true in the `case x` branch ... you definitely want to use `${x.getClass.getName}` in the error message, not `$x` 
3) wow, I thought we already had a UUID json codec in the codebase somewhere. I can't find it though. Hmmm .... 
4) purely for style and Scala familiarity, your call on whether or not to change: I find `fold` mechanics to be less readable and unwieldy unless they're really needed. I would probably have written this as:
```
Try(UUID.fromString(str)).getOrElse(throw DeserializationException(s"Couldn't parse value into a UUID"))
```

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -5,28 +5,34 @@ import java.util.UUID
...
 case class DataReferenceName(value: String) extends ValueObject
 case class NamedDataRepoSnapshot(name: DataReferenceName, snapshotId: String)
 
-object DataReferenceModelJsonSupport {
+object DataReferenceModelJsonSupport extends JsonSupport {
+  private def stringOrNull(in: Any): JsValue = in match {
+    case None => JsNull
+    case str: String => JsString(str)
+    case notStr => JsString(notStr.toString)
+  }
```
davidangb, line 20: Ah, this illustrates that this PR should have some unit tests - for individual `*Format` codecs and even just for this method.
```
val n: String = null
stringOrNull(n)
```
will result in a NPE, because it will attempt to execute the `case notStr` branch. In Scala, `null` and `None` are not the same thing ... you need to wrap nulls in an `Option` in order to match `None`:
```
  private def stringOrNull(in: Any): JsValue = Option(in) match {
    case None => JsNull
    case Some(str: String) => JsString(str)
    case Some(notStr) => JsString(notStr.toString)
  }
```

davidangb (commented):
more comments inline!

davidangb (approved):
:+1: once tests pass ... thanks for sticking it out!

MatthewBemis (approved):

PR #1282, opened 2020-10-01
took out config fallback to martha_v2 [WA-352]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/WA-352


rsasch:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -264,9 +264,8 @@ object Boot extends IOApp with LazyLogging {
         gcsConfig.getString("notifications.topicName")
       )
 
-      val marthaConfig: Config = conf.getConfig("martha")
-      val marthaBaseUrl: Option[String] = marthaConfig.getStringOption("baseUrl")
-      val marthaUrl = marthaBaseUrl.map(_ + "/martha_v3").getOrElse(marthaConfig.getString("baseUrl_v2"))
+      val marthaBaseUrl: String = conf.getConfig("martha").getString("baseUrl")
+      val marthaUrl: String = marthaBaseUrl.flatMap(_ + "/martha_v3")
```
salonishah11: flatMapping on a String would split the string into characters and append `/martha_v3` at the end of each character. 
So for example:
```
val strVal = "hello"
val path = strVal.flatMap(_ + "/world")
println(path)
```
will result in:
```
h/worlde/worldl/worldl/worldo/world
```

Instead one can do 
```
val marthaUrl: String = s"$marthaBaseUrl/martha_v3"
```
Code:
```
val strVal = "hello"
val path = s"$strVal/world"
println(path)
```
will result in:
```
hello/world
```

salonishah11 (changes requested):

kshakir (commented):
Concur with Saloni. Re-ping me here when ready for re-review.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
rsasch: Thank you!  Done.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -264,9 +264,8 @@ object Boot extends IOApp with LazyLogging {
         gcsConfig.getString("notifications.topicName")
       )
 
-      val marthaConfig: Config = conf.getConfig("martha")
-      val marthaBaseUrl: Option[String] = marthaConfig.getStringOption("baseUrl")
-      val marthaUrl = marthaBaseUrl.map(_ + "/martha_v3").getOrElse(marthaConfig.getString("baseUrl_v2"))
+      val marthaBaseUrl: String = conf.getConfig("martha").getString("baseUrl")
```
kshakir:
```suggestion
      val marthaBaseUrl: String = conf.getString("martha.baseUrl")
```

kshakir (approved):

salonishah11 (approved):

PR #1296, opened 2020-10-26
AS-438: log warning if Data Repo query results encounter STRUCT fields [risk: low]

davidangb:
The original scope for AS-438 was to support repeated fields and `STRUCT` fields in BigQuery results for Data Repo snapshots.

Data Repo does not support `STRUCT` fields. For these, I reviewed the code and left in place what we had: we'll just `toString` any we find and leave it at that. Since Data Repo doesn't support them and we don't expect to encounter them, this should be good enough. I did add a log statement so we can detect if we ever see these.

I located some repeated fields in BigQuery and tested empirically. The untested code we had in place previously worked as expected, so now that it's tested there's no changes needed there.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
@@ -6,6 +6,7 @@ import akka.http.scaladsl.model.StatusCodes
 import bio.terra.datarepo.model.{RelationshipModel, SnapshotModel, TableModel}
 import com.google.cloud.bigquery.Field.Mode
 import com.google.cloud.bigquery._
+import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.entities.datarepo.DataRepoBigQuerySupport._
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -216,6 +218,13 @@ trait DataRepoBigQuerySupport {
 
       val schemaFields = queryResults.getSchema.getFields
 
+      // check results for data we don't handle well. For now this is just RECORD types.
+      if (schemaFields.asScala.exists(_.getType == LegacySQLTypeName.RECORD)) {
+        logger.warn(s"query results for entity type $entityType contains one or more fields " +
+          s"with ${LegacySQLTypeName.RECORD.toString} datatypes; these " +
+          s"are unsupported and will be serialized to a string value.")
+      }
+
```
davidangb, line 227: This log message lets us know we hit a `STRUCT`, but it doesn't log the snapshot id or the workspace .... we don't have access to that info without changing a bunch of method signatures. Is this worth it?

ansingh7115 (approved):

PR #1314, opened 2020-11-13
[CA-946] Introducing, the MC-RBS (Resource Buffering Service)

andy7i:
https://broadworkbench.atlassian.net/browse/CA-946

Adds client library for RBS, with retries when we get a 500.

Config changes will be in a separate firecloud-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2354

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/config/RbsConfig.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.rawls.config
+
+import com.typesafe.config.Config
+
+case class RbsConfig(
```
marctalbott: In general in this PR I think I'd prefer `ResourceBuffer` or something like it over `Rbs`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1834,12 +1835,47 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     }
   }
 
-  // TODO: https://broadworkbench.atlassian.net/browse/CA-946
-  // This method should talk to the Resource Buffering Service, get a Google Project, and then set the appropriate
-  // Billing Account on the project and any other things we need to set on the Project.
-  private def getGoogleProjectFromRBS: Future[GoogleProjectId] = {
-    // TODO: Implement for real
-    Future.successful(GoogleProjectId(UUID.randomUUID.toString.substring(0, 8) + "_fake_proj_name"))
+  // todo: find this a home
```
marctalbott: Maybe `ResourceBufferModel`?

core/src/main/scala/org/broadinstitute/dsde/rawls/rbs/RbsService.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, UserInfo}
+
+import scala.concurrent.ExecutionContext
+
+object RbsService {
+  def constructor(rbsDao: RbsDAO, serviceAccountCreds: Credential)(userInfo: UserInfo)(implicit executionContext: ExecutionContext): RbsService = {
+    new RbsService(rbsDao, userInfo, serviceAccountCreds)
+  }
+}
+// todo: delete this file? might not need this if we're just using RBS under the hood.
```
marctalbott: I think it depends on how much work needs to get done here beyond just making calls to RBS. `RbsDAO` should just be the API calls and some retry logic like you have it in my opinion. In its current state, I'd agree that we could probably use `RbsDAO` directly from `WorkspaceService`, but I could see wanting to move stuff like `generateHandoutRequestId`, `getProjectPoolId` and `getGoogleProjectFromRBS` into this class

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1030,6 +1032,16 @@ class HttpGoogleServicesDAO(
       .build()
   }
 
+  def getResourceBufferServiceAccountCredential: Credential = {
+    new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(resourceBufferPemEmail)
+      .setServiceAccountScopes(Seq(ComputeScopes.CLOUD_PLATFORM).asJava)
```
andy7i: not entirely sure what scopes are needed here, but it errors if there are no scopes.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/ResourceBufferSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+
+class ResourceBufferSpec extends AsyncFlatSpec {
+
+
+  val userInfo = UserInfo(RawlsUserEmail("owner-access"), OAuth2BearerToken("token"), 123, RawlsUserSubjectId("123456789876543212345"))
+  val resourceBufferDAO = new MockResourceBufferDAO
+  val resourceBufferService = new ResourceBufferService(resourceBufferDAO, userInfo)
+
+
+  "getGoogleProjectFromRBS" should "get a Google Project ID" in {
```
andy7i: I wasn't really sure if any unit tests would be useful / worth the effort but can gladly add more if someone has more ideas

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -15,13 +15,14 @@ import org.broadinstitute.dsde.rawls.config.{DataRepoEntityProviderConfig, Deplo
 import org.broadinstitute.dsde.rawls.coordination.UncoordinatedDataSourceAccess
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
+import org.broadinstitute.dsde.rawls.dataaccess.resourcebuffer.ResourceBufferDAO
```
andy7i, line 18: intellij's optimize imports did some stuff

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
```diff
@@ -0,0 +1,20 @@
...
+case class ResourceBufferConfig(
+                                 url: String,
+                                 regularProjectPoolId: ProjectPoolId,
+                                 servicePerimeterProjectPoolId: ProjectPoolId
+                               )
+
+object ResourceBufferConfig{
+  def apply(conf: Config): ResourceBufferConfig = {
+    ResourceBufferConfig(
+      conf.getString("url"),
```
marctalbott, line 15: Where is the firecloud-develop PR that adds these values to config?

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.rawls.config
+
+import com.typesafe.config.Config
+import org.broadinstitute.dsde.rawls.model.ProjectPoolId
+
+case class ResourceBufferConfig(
+                                 url: String,
+                                 regularProjectPoolId: ProjectPoolId,
+                                 servicePerimeterProjectPoolId: ProjectPoolId
```
marctalbott: What's different about the regular projects and the service perimeter projects? I thought that we were adding projects to the perimeter once we got them, not that they'd be in the perimeter already?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ResourceBufferModel.scala
```diff
@@ -0,0 +1,10 @@
+package org.broadinstitute.dsde.rawls.model
+
+case class ProjectPoolId(value: String)
```
marctalbott, line 3: What's the difference between a `PoolId` and a `ProjectPoolId`?

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -164,7 +165,9 @@ object Boot extends IOApp with LazyLogging {
         cleanupDeploymentAfterCreating = dmConfig.cleanupDeploymentAfterCreating,
         terraBucketReaderRole = gcsConfig.getString("terraBucketReaderRole"),
         terraBucketWriterRole = gcsConfig.getString("terraBucketWriterRole"),
-        accessContextManagerDAO = accessContextManagerDAO
+        accessContextManagerDAO = accessContextManagerDAO,
+        resourceBufferPemFile = gcsConfig.getString("pathToResourceBufferPem"),
```
marctalbott: What is this service account? I would've thought that we'd be making requests to RBS as the Rawls service account?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/HttpResourceBufferDAO.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  private def getApiClient(accessToken: String): ApiClient = {
+    val client: ApiClient = new ApiClient()
+    client.setBasePath(baseUrl)
+    client.setAccessToken(accessToken)
+
+    client
+  }
+
+  override def getProjectPoolId(projectPoolType: ProjectPoolType.ProjectPoolType): ProjectPoolId = {
```
marctalbott: Not sure that this should be part of the DAO. I think the public methods in this class should just be actual requests to RBS. This seems like it belongs in the ResourceBufferService or I could also see it as a helper function for the enum (`toProjectPoolId`)

core/src/main/scala/org/broadinstitute/dsde/rawls/resourcebuffer/ResourceBufferService.scala
```diff
@@ -0,0 +1,37 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+object ResourceBufferService {
+  def constructor(resourceBufferDAO: ResourceBufferDAO)(userInfo: UserInfo)(implicit executionContext: ExecutionContext): ResourceBufferService = {
+    new ResourceBufferService(resourceBufferDAO, userInfo)
+  }
+}
+class ResourceBufferService(resourceBufferDAO: ResourceBufferDAO, protected val userInfo: UserInfo) {
+
+  def GetGoogleProjectFromRBS(projectPoolType: ProjectPoolType): Future[GoogleProjectId] = getGoogleProjectFromRBS(projectPoolType)
```
marctalbott: I know this is the pattern that we use elsewhere in Rawls, but I can't say I remember _why_ we use it? It's always felt unnecessary to me, but maybe that's because I just don't get it

core/src/main/scala/org/broadinstitute/dsde/rawls/resourcebuffer/ResourceBufferService.scala
```diff
@@ -0,0 +1,37 @@
...
+
+    val handoutRequestId = generateHandoutRequestId(userInfo, projectPoolId)
+
+    resourceBufferDAO.handoutGoogleProject(PoolId(projectPoolId.value), handoutRequestId)
+  }
+
+  //  handoutRequestId:
+  //        The unique identifier presented by the client for a resource request.
+  //        Using the same handoutRequestId in the same pool would get the same resource back.
+  private def generateHandoutRequestId(userInfo: UserInfo, projectPoolId: ProjectPoolId): String = {
```
marctalbott: It might be worth making a case class for this too so that we aren't passing Strings around

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1030,6 +1032,16 @@ class HttpGoogleServicesDAO(
...
 
+  def getResourceBufferServiceAccountCredential: Credential = {
+    new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(resourceBufferPemEmail)
+      .setServiceAccountScopes(workbenchLoginScopes.asJava) // might be ok with null scopes but adding this here per recommendation
+      .setServiceAccountPrivateKeyFromPemFile(new java.io.File(resourceBufferPemFile))
+      .build()
+  }
```
dvoet: why is this in `GoogleServicesDAO`?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1860,8 +1856,10 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       case _ => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"Billing Account is missing or invalid for Billing Project: ${billingProject}"))
     }
 
+    val resourceBufferService = new ResourceBufferService(resourceBufferDAO, userInfo)
```
marctalbott: Is this how we create services elsewhere? I'd prefer to see it get created in the constructor rather than down here

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/HttpResourceBufferDAO.scala
```diff
@@ -0,0 +1,64 @@
...
+class HttpResourceBufferDAO(config: ResourceBufferConfig, clientServiceAccountCreds: Credential)
+                           (implicit val system: ActorSystem, val materializer: Materializer,
+                            val executionContext: ExecutionContext) extends ResourceBufferDAO with Retry with LazyLogging {
+
+  private val baseUrl = config.url
+
+  override def handoutGoogleProject(poolId: PoolId, handoutRequestId: String): Future[GoogleProjectId] = {
+    retry(when500) { () =>
+      Future {
+        val resource = handoutResourceGeneric(poolId, handoutRequestId, OAuth2BearerToken(clientServiceAccountCreds.getAccessToken))
```
dvoet: I think you need a call to `refreshToken()` before call `getAccessToken`

marctalbott (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/HttpResourceBufferDAO.scala
```diff
@@ -0,0 +1,64 @@
...
+        GoogleProjectId(resource.getCloudResourceUid.getGoogleProjectUid.getProjectId)
+      }
+    }
+  }
+
+  protected def when500(throwable: Throwable): Boolean = {
+    throwable match {
+      case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(_.intValue / 100 == 5)
```
dvoet: I don't think RBS will throw a `RawlsExceptionWithErrorReport`

core/src/main/scala/org/broadinstitute/dsde/rawls/resourcebuffer/ResourceBufferService.scala
dvoet: this pattern is a remnant of the migration from spray to akka http. We don't need to propagate it in new service classes

core/src/main/scala/org/broadinstitute/dsde/rawls/resourcebuffer/ResourceBufferService.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, PoolId, ProjectPoolId, ProjectPoolType, UserInfo}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ResourceBufferService {
+  def constructor(resourceBufferDAO: ResourceBufferDAO)(userInfo: UserInfo)(implicit executionContext: ExecutionContext): ResourceBufferService = {
+    new ResourceBufferService(resourceBufferDAO, userInfo)
+  }
+}
+class ResourceBufferService(resourceBufferDAO: ResourceBufferDAO, protected val userInfo: UserInfo) {
```
dvoet: do we need a service class or can `ResourceBufferDAO` just be called from `WorkspaceService` where it is needed

core/src/main/scala/org/broadinstitute/dsde/rawls/resourcebuffer/ResourceBufferService.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  def GetGoogleProjectFromRBS(projectPoolType: ProjectPoolType): Future[GoogleProjectId] = getGoogleProjectFromRBS(projectPoolType)
+
+  def getGoogleProjectFromRBS(projectPoolType: ProjectPoolType = ProjectPoolType.Regular): Future[GoogleProjectId] = {
+
+    val projectPoolId: ProjectPoolId = resourceBufferDAO.getProjectPoolId(projectPoolType)
+
+    val handoutRequestId = generateHandoutRequestId(userInfo, projectPoolId)
```
dvoet: we should not be generating random request ids. We should use the workspace id so that if we end up retrying for the same workspace we get the same project and don't leak projects

dvoet (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
andy7i: I needed to futz around with decoding the secret, but it's here: https://github.com/broadinstitute/firecloud-develop/pull/2354

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
andy7i: RBS gives projects with APIs enabled (for example, this is the [AoU project pool configuration](https://github.com/DataBiosphere/terra-rbs/blob/master/src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml)). The idea here was that the service perimeter could have different APIs enabled. Is this not useful/necessary to have?

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
andy7i: RBS has a client SA instead of having each individual app being added to a whitelist individually. Each of the kernel apps will have a client SA like this.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: It looked like all existing service account Credentials were being set up here. Do you think it makes more sense to stick it in HttpResourceBufferDAO?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/HttpResourceBufferDAO.scala
andy7i: since I would need to pass in the config, I'm not sure if I can make it part of the Enumeration, but maybe if I change it to a sealed trait?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ResourceBufferModel.scala
andy7i: RBS doesn't actually have a concept of ProjectPoolId, so I separated them so that we can ensure that we aren't pulling from the wrong type of pool. Since Rawls might not be pulling any other types of resources from RBS, this might not be super useful

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
marctalbott: I don't have access to the link to the AoU project pool configuration, so I might be misunderstanding this.

I don't know of any difference in APIs that need to be enabled for a project to be eligible to be added to a service perimeter. That said, we do currently support a couple of small variations in projects (highSecurity, flowLogs, privateIpGoogleAccess -- all the boolean flags currently on the [createBillingProject endpoint](https://rawls.dsde-dev.broadinstitute.org/#/billing/createBillingProjectFull)). I could envision having different project pools to support those flags, but none of those flags should preclude a project from being added to a perimeter.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ResourceBufferModel.scala
marctalbott: In that case, I'd vote to remove `PoolId`. If we ever start using RBS for other resources, we could add distinct `{ResourceName}PoolId` case classes then if we want to ensure that we aren't pulling from the wrong type of pool. At the moment, this feels like an unnecessary distinction to me

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/ResourceBufferSpec.scala
marctalbott: Greg's PR has a couple of tests that `createWorkspace` and `cloneWorkspace` can get a project from RBS that could be implemented

andy7i:
I'm a bit worried about the interchangeable naming of RBS (aka Resource Buffering Service / ResourceBuffer / Buffer / MC RBS), some of which is definitely on me.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/ResourceBufferModel.scala
s-rubenstein: I am going to second this, this smells a bit to me of premature optimization.

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
dvoet: need to have highSecurity network to be in a perimeter. In any case, we probably want a different name than `servicePerimeterProjectPoolId` just now sure what.

core/src/test/resources/reference.conf
```diff
@@ -98,6 +98,14 @@ dataRepo {
   enabled = true
 }
 
+resourceBuffer {
+  projectPool {
+    regular = "dev-project-regular-v1"
+    servicePerimeter = "dev-project-serviceperimeter-v1"
+  }
+  url = "https://rbs.dsde-dev.broadinstitute.orgLOCAL"
```
s-rubenstein: Should this have a `:` between `.org` and `LOCAL`?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/ResourceBufferSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.broadinstitute.dsde.rawls.config.ResourceBufferConfig
+import org.broadinstitute.dsde.rawls.mock.MockResourceBufferDAO
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, ProjectPoolType, RawlsUserEmail, RawlsUserSubjectId, UserInfo}
+import org.broadinstitute.dsde.rawls.resourcebuffer.ResourceBufferService
+import org.scalatest.flatspec.AsyncFlatSpec
+
+class ResourceBufferSpec extends AsyncFlatSpec {
+
+
+  val userInfo = UserInfo(RawlsUserEmail("owner-access"), OAuth2BearerToken("token"), 123, RawlsUserSubjectId("123456789876543212345"))
```
s-rubenstein: I am assuming the `RawlsUserSubjectId` is not important here? Could we have it generate a random number string instead? (Just to show that it is irrelevant?)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/resourcebuffer/ResourceBufferSpec.scala
s-rubenstein: Yeah, I would rather test this inside of another test, than give it its on test. (Adding something like `verifyResourceBufferService isCalled`

s-rubenstein (commented):
This all looks reasonable to me. Mainly seconding Marc's comments with a few notes.

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
andy7i: Maybe this is something that we need:
```
        enableNetworkMonitoring:
          description: |-
            Make network traffic is measured. If enabled, turn on flow logs, enable private google access.
            See https://docs.google.com/document/d/1ccz2kzDL68CPofZ-b95ykQIudCgh7OCSFJ4Ym4Oz0r0/edit for more context.
```

https://github.com/DataBiosphere/terra-resource-buffer/blob/eed0f78856b9a7c7b5f6fd77f9bcfc30de63afdc/src/main/resources/config/resource_schema.yaml#L102

in which case, would something like `networkMonitoredProjectPoolId` or `flowLogsProjectPoolId` work?

core/src/test/resources/reference.conf
andy7i: I think this file is used for unit tests

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ResourceBufferConfig.scala
dvoet: how about `exfiltrationControlledPoolId`

s-rubenstein (approved):
LGTM

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -124,7 +124,9 @@ class HttpGoogleServicesDAO(
                              cleanupDeploymentAfterCreating: Boolean,
                              terraBucketReaderRole: String,
                              terraBucketWriterRole: String,
-                             override val accessContextManagerDAO: AccessContextManagerDAO)(implicit val system: ActorSystem, val materializer: Materializer, implicit val executionContext: ExecutionContext, implicit val cs: ContextShift[IO], implicit val timer: Timer[IO]) extends GoogleServicesDAO(groupsPrefix) with FutureSupport with GoogleUtilities {
+                             override val accessContextManagerDAO: AccessContextManagerDAO,
+                             resourceBufferPemFile: String,
```
dvoet: we should use the json instead of pem. That makes the firecloud-develop stuff simpler. Use `GoogleCredential.fromStream(...).toBuilder.setServiceAccountScopes(...).build()`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1049,6 +1050,12 @@ class HttpGoogleServicesDAO(
       .build()
   }
 
+  def getResourceBufferServiceAccountCredential: Credential = {
```
gpolumbo-broad: Barring how long a built `GoogleCredential` is valid for, I'm wondering if we can do all of this just one time during `Boot`.  Tests might want to use different JSON files, but the running App will only ever get 1 credential file at startup, so we have no need to keep making this API call to Google, unless the credential expires, in which case we might be able to keep track of the age of the credential and request it only when necessary to avoid extra calls to Google.

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1857,16 +1850,21 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     * @param span
     * @return Future[(GoogleProjectId, GoogleProjectNumber)] of the project that we claimed from RBS
     */
-  private def setupGoogleProject(billingProject: RawlsBillingProject, span: Span = null): Future[(GoogleProjectId, GoogleProjectNumber)] = {
+  private def setupGoogleProject(billingProject: RawlsBillingProject, span: Span = null, workspaceId: String) = {
```
gpolumbo-broad: Keep the return type

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2105,7 +2103,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
             s1.putAttribute("workspaceId", OpenCensusAttributeValue.stringAttributeValue(workspaceId))
 
             traceDBIOWithParent("getPolicySyncStatus", s1)(_ => DBIO.from(samDAO.getPolicySyncStatus(SamResourceTypeNames.billingProject, workspaceRequest.namespace, SamBillingProjectPolicyNames.owner, userInfo).map(_.email))).flatMap { projectOwnerPolicyEmail =>
-              traceDBIOWithParent("setupGoogleProject", s1)(_ => DBIO.from(setupGoogleProject(billingProject, s1))).flatMap { case (googleProjectId, googleProjectNumber) =>
+              traceDBIOWithParent("setupGoogleProject", s1)(_ => DBIO.from(setupGoogleProject(billingProject, s1, workspaceId))).flatMap { case (googleProjectId, googleProjectNumber) =>
```
gpolumbo-broad, line 2106: Should we consider using `WorkspaceName` as the `handoutRequestId` here?  The `workspaceId` is a UUID we just generated here and there isn't much to tie it back to the original request.  For example, if the `saveNewWorkspace` action should fail on the next line, we won't know the `workspaceId` except for maybe/hopefully in the logs it will be in the failed SQL statement or something.  If we use `workspaceRequest.toWorkspaceName` here, it will hopefully make it easier to track/correlate the request.  

Oh, and I guess all of this is on the assumption that the purpose of the `handoutRequestId` is purely for tracking/auditing/troubleshooting requests for Resources from RBS

gpolumbo-broad (commented):
Could also tests to `WorkspaceServiceSpec`.  There are a few `pending` tests in there related to getting the Project from RBS for Create and Clone workspace.  I think they can just do like some of the other tests and use Mockito to check that when we create/clone a workspace, there is exactly 1 call to `ResourceBufferService.getGoogleProjectFromBuffer` or to `ResourceBufferDAO.handoutGoogleProject` (either is probably fine, and I'd say choose whichever one is easier to Spy on.  There should be some existing tests doing the same thing for `MockGoogleServicesDAO` you can copy)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
dvoet: can change this to a `lazy val`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: `handoutRequestId` is not at all for tracking/auditing/troubleshooting, it is functional. RBS will always give back the same google project for a given `handoutRequestId`. It is a way to avoid dropping resources on the floor in the case a retries.

So the code is right as is. Consider the case that a user creates a workspace, deletes it, then creates it again. We want to be sure RBS gives us back a different project.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: changing to a lazy val

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -150,7 +150,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def getBucketServiceAccountCredential: Credential
 
-  def getResourceBufferServiceAccountCredential: Credential
+  val getResourceBufferServiceAccountCredential: Credential
```
dvoet: can you leave this as def?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1303,7 +1304,14 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
...
   }
 
-  it should "claim a Google Project from Resource Buffering Service" in pending
+  it should "claim a Google Project from Resource Buffering Service" in withTestDataServices { services =>
+    val newWorkspaceName = "space_for_workin"
+    val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
+
+    val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+
+    verify(services.resourceBufferService, Mockito.atLeastOnce()).getGoogleProjectFromBuffer(any[ProjectPoolType], any[String])
```
gpolumbo-broad:
```suggestion
    verify(services.resourceBufferService).getGoogleProjectFromBuffer(any[ProjectPoolType], any[String])
```
Leaving the second parameter blank means it will check that the method was called exactly once. 

Of course same goes for the `cloneWorkspace` test

gpolumbo-broad (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
andy7i: cool, I see that now. TIL!

gpolumbo-broad (approved):

PR #1318, opened 2020-11-24
CA-1102 - refactor create workspace

gpolumbo-broad:
Primary concerns:

Are we doing all the right steps in the right order?
Can we parallelize things?  I think we can and I think Cats IO is the way to do it better, I am just not very familiar with IO, so would want some help in doing that.

Below is the ordering of the "old" flow prior to the refactoring.  Note that most steps happen sequentially and if a step fails at any time, an exception will be thrown that halts the subsequent steps.  There is obviously finer detail of actions inside some of these methods, but I omitted the fine details of those methods here if they did not change in the refactoring. 
Double pipe || indicates step runs in parallel with each other || step at same level of nesting
```
1. withAttributeNamespaceCheck
2.   withBillingProjectContext
3.     withNewWorkspaceContext (starts db transaction)
4.       requireCreateWorkspaceAccess
5.         maybeRequireBillingProjectOwnerAccess
6.           workspaceQuery.findByName
7.             generate workspaceId (UUID)
8.             generate bucket name
9.             samDAO.getPolicySyncStatus for Billing Project Owner policy and return email address for policy group
10.              setupGoogleProject
11.                getGoogleProjectFromBuffer
12.                gcsDAO.setBillingAccountForProject
13.                getGoogleProjectNumber
14.              saveNewWorkspace
15.                workspaceQuery.createOrUpdate
16.                samDAO.createResourceFull
17.                ||createWorkflowCollectionForWorkspace
18.                ||Future.traverse(syncPolicyToGoogle) for each Workspace Policy that requires sync with Google
19.              something about gathering policy Email addresses and accounting for authDomains
20.              gcsDAO.setupWorkspace (Creates bucket and bucket IAM and logging)
21. commit db transaction (or did this happen in step 14?)
22. maybeUpdateGoogleProjectsInPerimeter (because we need to wait for the Workspace to be persisted before we can add to perimeter)
```

And for `cloneWorkspace`:

```
1.  withAttributeNamespaceCheck
2.    withLibraryAttributeNamespaceCheck
3.      withBillingProjectContext
4.        getWorkspaceContextAndPermissions (getWorkspaceIfUserHasAction)
5.          Start DB Transaction
6.            getWorkspaceContextAndPermissions (calling again in transaction in case Workspace was modified since last call)
7.              samDAO.getResourceAuthDomain (for Source Workspace)
8.                withClonedAuthDomain (are auth domains for clone compatible with source)
9.                 mergeWorkspaceAttributes
10.                  withNewWorkspaceContext (starts db transaction)
11.                    requireCreateWorkspaceAccess
12.                      maybeRequireBillingProjectOwnerAccess
13.                        workspaceQuery.findByName
14.                          generate workspaceId (UUID)
15.                          generate bucket name
16.                          samDAO.getPolicySyncStatus for Billing Project Owner policy and return email address for policy group
17.                            setupGoogleProject
18.                              getGoogleProjectFromBuffer
19.                              gcsDAO.setBillingAccountForProject
20.                              getGoogleProjectNumber
21.                            saveNewWorkspace
22.                              workspaceQuery.createOrUpdate
23.                              samDAO.createResourceFull
24.                              ||createWorkflowCollectionForWorkspace
25.                              ||Future.traverse(syncPolicyToGoogle) for each Workspace Policy that requires sync with Google
26.                            something about gathering policy Email addresses and accounting for authDomains
27.                            gcsDAO.setupWorkspace (Creates bucket and bucket IAM and logging)
28.                 entityQuery.copyAllEntities
29.                 copyAllMethodConfigs
30. commit db transaction (or did this happen in step 21?)
31. maybeUpdateGoogleProjectsInPerimeter (because we need to wait for the Workspace to be persisted before we can add to perimeter)
32. copyFilesWithPrefix -> return immediately, do not wait for Future
```


core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -111,15 +76,16 @@ trait WorkspaceSupport {
...
+  def getWorkspaceIfHasUserHasAction(workspaceName: WorkspaceName, requiredAction: SamResourceAction, attributeSpecs: Option[WorkspaceAttributeSpecs] = None): Future[Workspace] = {
     for {
       workspaceContext <- getWorkspaceContext(workspaceName, attributeSpecs)
       _ <- accessCheck(workspaceContext, requiredAction, ignoreLock = false) // throws if user does not have permission
     } yield workspaceContext
   }
 
+  // Not sure why this isn't just called `getWorkspace`.  Also, why doesn't this just call
+  //    `dataAccess.workspaceQuery.findByName(workspaceName, attributeSpecs)`
+  // directly since that's all that it does
```
davidangb: I didn't want to duplicate the `workspaceQuery.findByName` implementation in two places, with risk of them getting out of sync, so I had `getWorkspaceContext` delegate to `withWorkspaceContext`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    * @param workspaceRequest
+    * @param span
+    * @return
+    */
+  def createWorkspace(workspaceRequest: WorkspaceRequest, span: Span = null): Future[Workspace] = {
+    validateAttributeNamespace(workspaceRequest.attributes.keys)
+    createWorkspaceInternal(workspaceRequest, span)
+  }
+
+  // TODO: What's up with CloneWorkspace in Orch? Need to make sure this is doing same things as Orch
```
davidangb: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala#L254

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    * @param parentSpan
+    * @return
+    */
+  private def requireCreateWorkspaceAccess(workspaceRequest: WorkspaceRequest, parentSpan: Span = null): Future[Boolean] = {
+    val projectName = RawlsBillingProjectName(workspaceRequest.namespace)
+    traceWithParent("checkUserCanCreateWorkspace", parentSpan)(_ => samDAO.userHasAction(SamResourceTypeNames.billingProject, projectName.value, SamBillingProjectActions.createWorkspace, userInfo)).map {
+      case true => true
+      case false => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, s"You are not authorized to create a workspace in billing project ${workspaceRequest.toWorkspaceName.namespace}"))
+    }
+  }
```
davidangb, line 258: For consistency, these util methods should have the same signature - other methods return `Future[Unit]`, this one returns a `Boolean` but can only ever return true

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    val billingProjectOwnerPolicyEmailFuture = getBillingProjectOwnerPolicyEmail(workspaceRequest, span)
+    val getBillingProjectFuture = getBillingProjectForNewWorkspace(workspaceRequest.namespace, span)
+
+    for { // These things happen in parallel, but they all need to finish before moving on
+      _ <- checkUserCanCreateWorksapce
+      _ <- checkWorkspaceNameIsUnique
+      billingProjectOwnerPolicyEmail <- billingProjectOwnerPolicyEmailFuture
+      billingProject <- getBillingProjectFuture
+    } yield (billingProjectOwnerPolicyEmail, billingProject)
+  }
```
davidangb: an interesting side effect of running these in parallel is the possibility of information leakage. Consider a user who has zero permissions to create a workspace in billing project X. Now consider that user is trying to create a workspace named "foo". The user can't see that a workspace X/foo exists, because they don't have permissions.

If `checkCreateWorkspacePermissions` is slow because it's querying Sam, but `checkIfWorkspaceWithNameAlreadyExists` is fast because it's only going to the local database ... the user could end up with an error message of "Workspace ${workspaceName} already exists", when otherwise they would not have been able to know if that workspace existed or not.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+
+    for {
+      (billingProjectOwnerPolicyEmail, billingProject) <- firstSteps(workspaceRequest, span)
+      (workspace, workspaceResourceResponse) <- secondSteps(workspaceRequest, billingProject, billingProjectOwnerPolicyEmail, noWorkspaceOwner, authDomains, span)
+      _ <- thirdSteps(workspace, billingProject, workspaceResourceResponse, billingProjectOwnerPolicyEmail, authDomains, span)
+    } yield workspace
+  }
+
+  private def firstSteps(workspaceRequest: WorkspaceRequest, span: Span = null): Future[(WorkbenchEmail, RawlsBillingProject)] = {
+    val checkUserCanCreateWorksapce = checkCreateWorkspacePermissions(workspaceRequest, span)
```
davidangb: nit: spelling of "workspace"

davidangb (commented):
I'm not close to finishing review, but here's the comments I have so far ...

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -111,15 +76,16 @@ trait WorkspaceSupport {
 
   //WorkspaceContext helpers
 
-  // function name may be misleading. This returns the workspace context and checks the user's permission,
-  // but does not return the permissions.
-  def getWorkspaceContextAndPermissions(workspaceName: WorkspaceName, requiredAction: SamResourceAction, attributeSpecs: Option[WorkspaceAttributeSpecs] = None): Future[Workspace] = {
+  def getWorkspaceIfHasUserHasAction(workspaceName: WorkspaceName, requiredAction: SamResourceAction, attributeSpecs: Option[WorkspaceAttributeSpecs] = None): Future[Workspace] = {
```
dvoet: too many `Has`es in the name?

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
dvoet: the word `Context` is used because there used to be such a thing as `WorkspaceContext` that was more than just `Workspace`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+object WorkspaceCreator {
+  def constructor(dataSource: SlickDataSource,
+                  samDAO: SamDAO,
+                  gcsDAO: GoogleServicesDAO,
+                  config: WorkspaceServiceConfig)
+                 (userInfo: UserInfo)
+                 (implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext): WorkspaceCreator = {
+    new WorkspaceCreator(userInfo, dataSource, samDAO, gcsDAO, config)
+  }
+}
```
dvoet: You don't need this object

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+                       val gcsDAO: GoogleServicesDAO,
+                       val config: WorkspaceServiceConfig)
+                      (implicit val system: ActorSystem, val materializer: Materializer, protected val executionContext: ExecutionContext)
+  extends WorkspaceSupport with AttributeSupport with LibraryPermissionsSupport with Retry with LazyLogging {
+
+  import dataSource.dataAccess.driver.api._
+
+  /**
+    * Public method to be used for creating a new Workspace.  The bulk of the creation process is performed by
+    * `createWorkspaceInternal`
```
dvoet: get rid of the part of the comment "The bulk of the creation process is performed by `createWorkspaceInternal`"

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+      //we also don't want to block returning a response on this call because it's already a slow endpoint
+      copyBucketContents(sourceWorkspace, destWorkspace, cloneWorkspaceRequest.copyFilesWithPrefix)
+
+      destWorkspace
+    }
+  }
+
+  // Temporary method for easily switching between the parallel/sequential versions of createWorkspaceInternal while
+  // this is still in development/testing
+  private def createWorkspaceInternal(workspaceRequest: WorkspaceRequest, span: Span = null): Future[Workspace] = {
```
dvoet: why do you need this function? Why can't `cloneWorkspace` just call `createWorkspace`?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    */
+  private def validateAuthDomainsForDestWorkspace(sourceWorkspaceId: String, destWorkspaceAuthDomains: Set[ManagedGroupRef]): Future[Set[ManagedGroupRef]] = {
+    samDAO.getResourceAuthDomain(SamResourceTypeNames.workspace, sourceWorkspaceId, userInfo).map { sourceAuthDomainStrings =>
+      val sourceAuthDomains = sourceAuthDomainStrings.map(n => ManagedGroupRef(RawlsGroupName(n))).toSet
+      unionAuthDomains(sourceAuthDomains, destWorkspaceAuthDomains)
+    }
+  }
+
+  /**
+    * Ever Auth Domain listed in sourceWorkspaceADs must be present in destWorkspaceADs or else an exception will be
```
dvoet:
```suggestion
    * Every Auth Domain listed in sourceWorkspaceADs must be present in destWorkspaceADs or else an exception will be
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    * thrown.  If valid, then return the Union of sourceWorkspaceADs and destWorkspaceADs.
+    *
+    * TODO: If sourceWorkspaceADs is a subset of destWorkspaceADs, isn't the Union equivalent to destWorkspaceADs?
```
dvoet: yes, I would just inline this function

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    validateAttributeNamespace(workspaceAttributeNames)
+    validateLibraryAttributeNamespaces(libraryAttributeNames)
+
+    for {
+      sourceWorkspace <- getWorkspaceIfHasUserHasAction(sourceWorkspaceName, SamWorkspaceActions.read)
+      destAuthDomains <- validateAuthDomainsForDestWorkspace(sourceWorkspace.workspaceId, destWorkspaceRequest.authorizationDomain.getOrElse(Set.empty))
+      newAttrs = sourceWorkspace.attributes ++ destWorkspaceRequest.attributes
+      cloneWorkspaceRequest = destWorkspaceRequest.copy(authorizationDomain = Option(destAuthDomains), attributes = newAttrs)
+      destWorkspace <- createWorkspaceInternal(cloneWorkspaceRequest)
+      _ <- copyEntitiesAndMethodConfigs(sourceWorkspace, destWorkspace)
```
dvoet, line 77: An inadvertent change here is that creating the workspace and copying the source contents are separate transactions. The copy could fail leaving a created but empty workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+    * bucket
+    *
+    * @param filePrefix
+    * @param sourceWorkspace
+    * @param destWorkspace
+    */
+  private def copyBucketContents(sourceWorkspace: Workspace, destWorkspace: Workspace, filePrefix: Option[String]): Future[List[Option[StorageObject]]] = {
+    filePrefix match {
+      case Some(prefix) => gcsDAO.listObjectsWithPrefix(sourceWorkspace.bucketName, prefix).flatMap { objectsToCopy =>
+        Future.traverse(objectsToCopy) { objectToCopy => gcsDAO.copyFile(sourceWorkspace.bucketName, objectToCopy.getName, destWorkspace.bucketName, objectToCopy.getName) }
```
dvoet, line 218: probably a weakness in the original code: doing `Future.traverse` on a possibly large list has the potential to take down a rawls instance. Depending on the execution context it can a) block other work from happening by hogging the queue or b) exhaust memory by causing too many threads to be created.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+  private def requireBillingProjectOwnerAccess(workspaceRequest: WorkspaceRequest, parentSpan: Span): Future[Unit] = {
+    traceWithParent("listUsersRolesOnBillingProject", parentSpan)(_ => samDAO.listUserRolesForResource(SamResourceTypeNames.billingProject, workspaceRequest.namespace, userInfo)).map { billingProjectRoles =>
+      if (!billingProjectRoles.contains(SamProjectRoles.owner))
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Missing ${SamProjectRoles.owner} role on billing project '${workspaceRequest.namespace}'."))
+    }
+  }
+
+  // Legacy code calls samDAO.getPolicySyncStatus just to get the Owner Policy Email.  Not exactly sure why it is done
+  // this way instead of just searching up the Policy email.  This way checks that there's sync information for the
+  // policy, and perhaps that's useful?
```
dvoet, line 278: the caller may not have read_policies action on the billing-project

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+        Option(WorkspaceAccessLevels.ProjectOwner -> billingProjectOwnerPolicyEmail)
+      } else {
+        WorkspaceAccessLevels.withPolicyName(policyName.value).map(_ -> policyEmail)
+      }
+    }.flatten.toMap
+
+    traceWithParent("gcsDAO.setupWorkspace", span)(_ => gcsDAO.setupWorkspace(userInfo, workspace.googleProjectId, policyEmails, workspace.bucketName, getLabels(authDomains.toList), span))
+  }
+
+  private def getLabels(authDomain: List[ManagedGroupRef]): Map[String, String] = authDomain match {
```
dvoet, line 530: get what kind of labels?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+                         authDomains: Set[ManagedGroupRef],
+                         span: Span = null): Future[Unit] = {
+    val policyNameToEmailMap = samCreateResourceResponse.accessPolicies.map(x => SamResourcePolicyName(x.id.accessPolicyName) -> WorkbenchEmail(x.email)).toMap
+    for { // Run all of these in parallel
+      _ <- Future.sequence(Seq(
+        createWorkflowCollectionForWorkspace(workspace.workspaceId, policyNameToEmailMap, span),
+        syncWorkspacePoliciesWithGoogle(workspace, policyNameToEmailMap, authDomains, span),
+        setupGoogleProjectForWorkspace(workspace, policyNameToEmailMap, billingProjectOwnerPolicyEmail, authDomains, span),
+        maybeUpdateGoogleProjectsInPerimeter(billingProject)))
+    } yield ()
```
davidangb, line 154: nit: `firstSteps` runs a few futures in parallel by assigning each to a val, then using a for-comp calling each val. Here in `thirdSteps` you use `Future.sequence`. I think we should use only one approach for consistency and readability.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+            dataAccess.methodConfigurationQuery.get(sourceWorkspace, methodConfigShort.namespace, methodConfigShort.name).flatMap { methodConfig =>
+              dataAccess.methodConfigurationQuery.create(destWorkspace, methodConfig.get)
+            }
+          }
+          DBIO.seq(inserts: _*)
+        } andThen {
+        DBIO.successful()
+      }
+    }
+  }
```
davidangb, line 205: First, I'm noting Doug's comment about clone operations now being in separate transactions and potentially resulting in half-cloned workspaces. If and only if we are comfortable with operations being in separate transactions ... I'd definitely break this one up. Method configs and entities don't need to be tied in the same transaction.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
```diff
@@ -0,0 +1,688 @@
...
+  private def createWorkspaceInternal_parallel(workspaceRequest: WorkspaceRequest, span: Span = null): Future[Workspace] = {
+    val authDomains = workspaceRequest.authorizationDomain.getOrElse(Set.empty)
+    val noWorkspaceOwner = workspaceRequest.noWorkspaceOwner.getOrElse(false)
+
+    for {
+      (billingProjectOwnerPolicyEmail, billingProject) <- firstSteps(workspaceRequest, span)
+      (workspace, workspaceResourceResponse) <- secondSteps(workspaceRequest, billingProject, billingProjectOwnerPolicyEmail, noWorkspaceOwner, authDomains, span)
+      _ <- thirdSteps(workspace, billingProject, workspaceResourceResponse, billingProjectOwnerPolicyEmail, authDomains, span)
+    } yield workspace
+  }
```
davidangb: thinking out loud, not for this PR: at what point is it safe to return to the user? Could we return a response to the user after `secondSteps`, then let `thirdSteps` continue asynchronously? We'd have to handle failures in the `thirdSteps` work somehow; I don't know how that  would look to the user.

davidangb (commented):
first pass complete, here's the rest of my comments

davidangb:
This PR will likely resolve AS-222.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Yeah, I am open to suggestions here.

The original goal of this refactor was to isolate the db transactions from external API calls, and one of my primary concerns about the resulting code is that by doing this, we have changed the behavior.  So what is acceptable here and what is not?

I also tried to preserve the idea that `cloneWorkspace` does the same thing as `createWorkspace`, plus the couple of extra steps before/after as needed.  To achieve what you're looking for here, avoiding the partially cloned workspace, I think Stairway would be the ideal solution.  Short of that, I think this will take some deeper thought or some finer interweaving of `create` and `clone` maybe.  I will take a look at this now to see if there's an elegant solution.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: yeah, totally, and this actually did trigger a test failure.  I put things back to be mostly sequential for now.  Hopefully in the future we can find ways to parallelize this more.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Excellent question.  It is because of when `validateAttributeNamespace` is getting called in create and clone.  

If you look at `cloneWorkspace`, you can see that it calls `validateAttributeNamespace` and `validateLibraryAttributeNamespaces` prior to doing some other "preliminary" stuff before calling `createWorkspaceInternal`.  I did it this way because I didn't dig too deep into what the `validate*AttributeNamespace`actually does and I was generally trying to preserve the ordering of operations from the old code and new.  

This is one of those things where if you can help me confirm that a little reordering of validations and auth checks is OK here, then we can make this change, so please let me know what you think.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Sure, I didn't mess around in this block of code, just copied it over.  I'm inclined to just let it be for now so we can get through the bulk of this refactor, but if you want to dig into this and try to fix it now, just say the words.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Look like labels placed on the bucket.  Again, this was just code I copied over and didn't dig into unless I needed to.  Once again, if you want to clean this up now, just say the words otherwise I'm inclined to leave it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Understood.  Not sure what we want to do here.  The fact that we've isolated the db transactions from the API calls has a cost, and it's this risk of a partially created Workspace.  Short of getting Stairway in here, I don't have a good idea for a solution to this right now.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceCreator.scala
gpolumbo-broad: Per the other comments about ordering of things (and tests that broke) I put this back to being _mostly_ sequential.  I would love to talk about parallelizing though, but to do that, I'd rather sit with you and/or Doug and make sure we're on the same page about what can be parallelize and what needs to happen sequentially.  The good news is that _hope_ that the bulk of this refactoring has broken out the steps involved in such a way that making those decisions should hopefully be a lot easier now.

davidangb (commented):
The code in here looks ok, and it shows as passing tests. All good there.

I absolutely think this is the right direction to go, but I am concerned about behavior changes with partially-created workspaces. Most specifically, I want to be sure that there are no cases in which a partially-created workspace results in the workspace being visible to the public or to users other than the one that created it. Even if there's no sensitive data in the workspace yet, it's a bad look if users start seeing extra workspaces. Practically, this means that all steps that change/set permissions must be ordered from most restrictive to least restrictive - at no point should an error leave a workspace in a less-restrictive state. That said, I think the ordering is correct on this front.

I also wonder if we need to implement any of these:
* add a "state" column (whatever you want to name it) to the WORKSPACE table. Insert the workspace as "creating" and only once all steps have finished, update the workspace to "ready". Modify all other select queries to filter out "creating" workspaces.
* add a try/catch or recover phase to `createWorkspaceInternal` and clean up any partial creations. We could maybe do this by implementing a resilient "deleteWorkspace" method, and  just calling "resilientDeleteWorkspace" to clean up anything left behind.

gpolumbo-broad:
Great suggestions @davidangb , thanks.  I'll look at this some more to see what I can think of.  These were my same concerns too and it seems obvious now that I should have paid more attention to them during the original refactor.  

I don't know much about Stairway, but it sounds like the right solution for this sort of thing, no?  I will start looking at that next to see how easy or hard it might be to incorporate here.

davidangb:
@gpolumbo-broad Stairway would give us the ability to handle a saga transaction and rollback of said transaction - absolutely. However it would still require us to write the rollback code - e.g. to delete any buckets that we created. And it would still be vulnerable - if only for a few seconds/minutes - to partially-created workspaces. If we persisted a workspace to the DB, that workspace would be returned by any DB queries until Stairway rolled it back.

gpolumbo-broad:
Ah, ok, that makes sense.  I really don't know much of anything about Stairway yet other than the elevator pitch.  

It is sounding to me like we do not want to move ahead with the PR until we find a way to reduce the likelihood and risk of partially created Workspaces.  I'll see if there's something I can do.

davidangb:
The code restructuring in this PR is valuable in and of itself, and it risks getting out of date/needing ugly rebases if it lingers. Brainstorming: you could potentially wrap the entire `createWorkspaceInternal` in a transaction. This would replicate previous behavior and allow the code restructuring to be merged. From there, work on removing that outer transaction through one of the means we've discussed. (this is just an idea, I am open to other options)

gpolumbo-broad:
Closing this out.  Would be cool to do but this is just not a priority now.

PR #1319, opened 2020-11-25
Retrieve location for a regional bucket [BW-438]

salonishah11:
This PR **only** retrieves compute zones for a regional bucket i.e. bucket whose location type is `region`. For buckets that are `multi-region` (which are default type of bucket created by Rawls) we pass default runtime zone attributes to Cromwell while submitting a workflow as before.

Closes https://broadworkbench.atlassian.net/browse/BW-438

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -232,6 +232,9 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def testBillingAccountAccess(billingAccount: RawlsBillingAccountName, userInfo: UserInfo): Future[Boolean]
 
+  /**
+    * Returns location of a regional bucket. If the bucket's location type is `multi-region`, it returns None
+    */
   def getRegionForRegionalBucket(bucketName: String): Future[Option[String]]
 
   def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[List[String]]
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -231,6 +231,8 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def getFolderId(folderName: String): Future[Option[String]]
 
   def testBillingAccountAccess(billingAccount: RawlsBillingAccountName, userInfo: UserInfo): Future[Boolean]
+
+  def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[Option[List[String]]]
 }
 
 object GoogleApiTypes {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -232,6 +232,8 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def testBillingAccountAccess(billingAccount: RawlsBillingAccountName, userInfo: UserInfo): Future[Boolean]
 
+  def getRegionForRegionalBucket(bucketName: String): Future[Option[String]]
+
   def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[List[String]]
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -232,7 +232,7 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def testBillingAccountAccess(billingAccount: RawlsBillingAccountName, userInfo: UserInfo): Future[Boolean]
 
-  def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[Option[List[String]]]
+  def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[List[String]]
 }
 
 object GoogleApiTypes {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -406,9 +406,13 @@ class HttpGoogleServicesDAO(
 
   override def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[Option[List[String]]] = {
     implicit val service = GoogleInstrumentedService.Storage
-    val getter = getComputeManager(getBucketServiceAccountCredential).regions().get(googleProject.value, region)
-    retryWithRecoverWhen500orGoogleError(() => { Option(executeGoogleRequest(getter).getZones.asScala.toList) }) {
-      case e: HttpResponseException => None
+    val getter = getComputeManager(getBillingServiceAccountCredential).regions().get(googleProject.value, region)
+    retryWithRecoverWhen500orGoogleError(() => {
+      val zonesAsResourceUrls = executeGoogleRequest(getter).getZones.asScala.toList
+      val zones = zonesAsResourceUrls.map(_.split("/").last)
+      Option(zones)
+    }) {
+      case e => None
     }
   }
 
@@ -1036,7 +1040,7 @@ class HttpGoogleServicesDAO(
   }
 
   def getBillingServiceAccountCredential: Credential = {
-    new GoogleCredential.Builder()
+    val abc = new GoogleCredential.Builder()
       .setTransport(httpTransport)
       .setJsonFactory(jsonFactory)
       .setServiceAccountScopes(Seq(ComputeScopes.CLOUD_PLATFORM).asJava) // need this broad scope to create/manage projects
@@ -1044,6 +1048,10 @@ class HttpGoogleServicesDAO(
       .setServiceAccountPrivateKeyFromPemFile(new java.io.File(billingPemFile))
       .setServiceAccountUser(billingEmail)
       .build()
+
+    println(s"******* FIND ME TOKEN getBillingServiceAccountCredential: ${abc.getAccessToken}")
+
+    abc
   }
 
   def toGoogleGroupName(groupName: RawlsGroupName) = s"${proxyNamePrefix}GROUP_${groupName.value}@${appsDomain}"
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -401,9 +401,11 @@ class HttpGoogleServicesDAO(
 
   override def getBucket(bucketName: String)(implicit executionContext: ExecutionContext): Future[Option[Bucket]] = {
     implicit val service = GoogleInstrumentedService.Storage
-    val getter = getStorage(getBucketServiceAccountCredential).buckets().get(bucketName)
-    retryWithRecoverWhen500orGoogleError(() => { Option(executeGoogleRequest(getter)) }) {
-      case e: HttpResponseException => None
+    retryWithRecoverWhen500orGoogleError(() => {
+      val getter = getStorage(getBucketServiceAccountCredential).buckets().get(bucketName)
+      Option(executeGoogleRequest(getter))
+    }) {
+      case _: HttpResponseException => None
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -410,12 +410,12 @@ class HttpGoogleServicesDAO(
   }
 
   override def getRegionForRegionalBucket(bucketName: String): Future[Option[String]] = {
-    getBucket(bucketName) map { maybeBucket =>
-      val bucket = maybeBucket.getOrElse(throw new RawlsException(s"Failed to retrieve bucket `$bucketName`"))
-      bucket.getLocationType match {
+    getBucket(bucketName) map {
+      case Some(bucket) => bucket.getLocationType match {
         case SingleRegionLocationType => Option(bucket.getLocation.toLowerCase)
         case _ => None
       }
+      case None => throw new RawlsException(s"Failed to retrieve bucket `$bucketName`")
     }
   }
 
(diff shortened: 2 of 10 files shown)


aednichols (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
```diff
@@ -263,6 +263,23 @@ class WorkflowSubmissionSpec(_system: ActorSystem) extends TestKit(_system) with
...
+    }
+
+    withWorkspaceContext(testData.regionalWorkspace) { ctx =>
+      val (workflowRecs, submissionRec, workspaceRec) = getWorkflowSubmissionWorkspaceRecords(testData.regionalSubmission, testData.regionalWorkspace)
+
+      Await.result(workflowSubmission.submitWorkflowBatch(WorkflowBatch(workflowRecs.map(_.id), submissionRec, workspaceRec)), Duration.Inf)
+
+      val workflowOptions = mockExecCluster.getDefaultSubmitMember.asInstanceOf[MockExecutionServiceDAO].submitOptions.map(_.parseJson.convertTo[ExecutionServiceWorkflowOptions])
+
+      workflowOptions.get.default_runtime_attributes shouldBe Some(JsObject(Map("zones" -> JsString("europe-north1-a europe-north1-b europe-north1-c"))))
```
cjllanwarne: Is this a little bit of a hostage to fortune if those zones ever change (eg if they add `europe-north1-d`)? Can we say something more like `zones.nonEmpty && zones.forall { z => z.startsWith("europe-north1-") }`?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockGoogleServicesDAO.scala
```diff
@@ -223,4 +223,21 @@ class MockGoogleServicesDAO(groupsPrefix: String,
     Future.successful(billingAccount == accessibleBillingAccountName)
   }
 
+  override def getRegionForRegionalBucket(bucketName: String): Future[Option[String]] = {
+    Future.successful {
+      bucketName match {
+        case "fc-regional-bucket" => Option("europe-north1")
```
cjllanwarne: Just FWIW I think the real API returns this as `EUROPE-NORTH1`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -231,6 +231,13 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
   def getFolderId(folderName: String): Future[Option[String]]
 
   def testBillingAccountAccess(billingAccount: RawlsBillingAccountName, userInfo: UserInfo): Future[Boolean]
+
+  /**
+    * Returns location of a regional bucket. If the bucket's location type is `multi-region`, it returns None
```
cjllanwarne, line 236: Is that right? How do we know how to clone the region requirements for multi-regional buckets?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -398,9 +401,33 @@ class HttpGoogleServicesDAO(
...
+      val getter = getStorage(getBucketServiceAccountCredential).buckets().get(bucketName)
+      Option(executeGoogleRequest(getter))
+    }) {
+      case _: HttpResponseException => None
+    }
+  }
+
+  override def getRegionForRegionalBucket(bucketName: String): Future[Option[String]] = {
+    getBucket(bucketName) map { maybeBucket =>
+      val bucket = maybeBucket.getOrElse(throw new RawlsException(s"Failed to retrieve bucket `$bucketName`"))
```
cjllanwarne: A `map` into a `getOrElse` feels like it could be neater as a `match`?

You could also `fold` if you really want to, though I've never been a big fan of `fold` on `Option`s

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -398,9 +401,33 @@ class HttpGoogleServicesDAO(
...
+      bucket.getLocationType match {
+        case SingleRegionLocationType => Option(bucket.getLocation.toLowerCase)
```
cjllanwarne: Is there a reason for the `toLowerCase` here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -398,9 +401,33 @@ class HttpGoogleServicesDAO(
...
+      }
+    }
+  }
+
+  override def getComputeZonesForRegion(googleProject: GoogleProjectId, region: String): Future[List[String]] = {
+    implicit val service = GoogleInstrumentedService.Storage
+    retryWithRecoverWhen500orGoogleError(() => {
+      val getter = getComputeManager(getBucketServiceAccountCredential).regions().get(googleProject.value, region)
+      val zonesAsResourceUrls = executeGoogleRequest(getter).getZones.asScala.toList
+      zonesAsResourceUrls.map(_.split("/").last)
```
cjllanwarne, line 431: Maybe you could add a comment on why you need to do this `split`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -398,9 +401,33 @@ class HttpGoogleServicesDAO(
...
+    }) {
+      case e => throw new RawlsException(s"Something went wrong while retrieving zones for region `$region` under Google " +
+        s"project `${googleProject.value}`. Error: ${ExceptionUtils.getMessage(e)}")
```
cjllanwarne: Can you pass the error as a `causedBy` to the `RawlsException` to preserve them both in stack trace hierarchy?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -186,6 +186,25 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
...
+    for {
+      regionOption <- googleServicesDAO.getRegionForRegionalBucket(bucketName)
+      runtimeOptions <- {
+        regionOption match {
+          case Some(region) =>
+            for {
+              zones <- googleServicesDAO.getComputeZonesForRegion(GoogleProjectId(googleProjectId), region)
+            } yield {
+              Option(JsObject("zones" -> JsString(zones.mkString(" "))))
+            }
```
cjllanwarne: This looks more like it could be a simple `map` rather than a one-line `for` comprehension?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -186,6 +186,25 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
...
+          case None =>
+            // if the location is `multi-region` we want the default zones from config
+            Future.successful(defaultRuntimeOptions)
```
cjllanwarne, line 207: Probably fine for now, but it's only a matter of time before someone creates an `EU` multi-regional bucket.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -186,6 +186,25 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
     }
   }
 
+  def getRuntimeOptions(googleProjectId: String, bucketName: String)(implicit executionContext: ExecutionContext): Future[Option[JsValue]] = {
```
cjllanwarne, line 189: This seems to be assuming that the only entry in `defaultRuntimeOptions` is `zones`. We probably want to merge with the default options instead of generating a new set.

cjllanwarne (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
salonishah11: That is a good point. I will update it 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
salonishah11: Can you explain this more? We are only supporting creating/cloning workspaces (outside us) in single region only no at the moment?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockGoogleServicesDAO.scala
salonishah11: That is correct. But in the actual implementation we convert it to lowercase before returning it. The reason we convert it to lowercase is that `.regions().get(..)` in [here](https://github.com/broadinstitute/rawls/blob/889918b9e49916d7ac9832539854f89b4b033e3f/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L425-L426) expects the input to be in lowercase. If it is in any other format it throws an error

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
salonishah11: Yes. The reason we convert it to lowercase is that `.regions().get(..)` in [here](https://github.com/broadinstitute/rawls/blob/889918b9e49916d7ac9832539854f89b4b033e3f/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L425-L426) expects the input to be in lowercase. If it is in any other format it throws an error

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
salonishah11: I am not sure how to use `causedBy`. So something like this
```
throw new (RawlsException(s"Something went wrong while retrieving zones for region `$region` under Google " +
        s"project `${googleProject.value}`.") causedBy e)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
salonishah11: Currently the only entry in `defaultRuntimeOptions` is `zones` based on [rawls.conf in fc-develop](https://github.com/broadinstitute/firecloud-develop/blob/6f37dd2a61ed30bc1fb9ae9bc2707a48e38cfa1c/base-configs/rawls/rawls.conf.ctmpl#L346) and what is being passed to SubmissionActor in rawls [here](https://github.com/broadinstitute/rawls/blob/45a0b0b5b6bd1731bc61736ec71a85a9c2bbfd9d/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala#L176). So instead of merging it we replaced it

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cjllanwarne: Maybe comment that here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cjllanwarne: Most (many?) exceptions accept a second parameter like `new RawlsException(msg: String, causedBy: Throwable)`

cjllanwarne: If `RawlsException` doesn't have that as a constructor argument then maybe we could add it?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
cjllanwarne: I have a feeling that the reference disk work might need/want to use workflow options too. It's also a bit risky IMO to leave an option in the config file that ignores half of its contents if someone changes it.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
salonishah11: So I refactored the code a bit. Instead of doing `.toLowerCase` here, I shifted it to `getComputeZonesForRegion()` function where it made more sense. I have also added a comment there stating why `.toLowerCase` is needed

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
salonishah11: `RawlsException` does have a `cause` field wherein you can pass the throwable. I have used that 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
salonishah11: We will validate that the region passed during workspace creation is a single region only and not multi-region like `EU` or `US`

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
salonishah11: I have modified the code to only change the `zones` in the `defaultRuntimeOptions` and not generate a new set

salonishah11: I also modified the test to check that if another default runtime option is passed than it should not be overridden by the get-compute-zones logic

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
```diff
@@ -276,7 +279,16 @@ class WorkflowSubmissionSpec(_system: ActorSystem) extends TestKit(_system) with
...
+      val expectedRuntimeOptions = workflowOptions.get.default_runtime_attributes.get.asJsObject
+      val actualZones: JsValue = expectedRuntimeOptions.fields("zones")
+      val actualZonesList: List[String] = actualZones.convertTo[String].split(" ").toList
+
+      actualZonesList should not be (empty)
+      actualZonesList.foreach { z => z should startWith("europe-north1-") }
+
+      // check that other runtime option passed has not been overridden
+      val anotherRuntimeOption = expectedRuntimeOptions.fields("another-option")
+      anotherRuntimeOption should be (JsString("10GB"))
```
cjllanwarne, line 291: 💯

cjllanwarne (approved):

PR #1321, opened 2020-12-01
Add support for creating and cloning workspace buckets in a specified region [BW-427][BW-443]

aednichols:
This PR provides an optional parameter called `bucketLocation` for workspace creation. It also adds support for non-default bucket locations when the workspace is cloned.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
@@ -320,9 +320,7 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
               callZones foreach { _.split(",") foreach { zone => zone should startWith (europeNorth1ZonesPrefix) } }
 
               workerAssignedExecEvents should not be (empty)
-              workerAssignedExecEvents foreach { event =>
-                event should include (europeNorth1ZonesPrefix)
-              }
+              workerAssignedExecEvents foreach { event => event should include (europeNorth1ZonesPrefix) }
             }
           }
         }
@@ -416,9 +414,7 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
               callZones foreach { _.split(",") foreach { zone => zone should startWith (europeNorth1ZonesPrefix) } }
 
               workerAssignedExecEvents should not be (empty)
-              workerAssignedExecEvents foreach { event =>
-                event should include (europeNorth1ZonesPrefix)
-              }
+              workerAssignedExecEvents foreach { event => event should include (europeNorth1ZonesPrefix) }
             }
           }
         }
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
@@ -220,6 +220,101 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
 
     }
 
+    "should be able to run sub-workflow tasks in non-US regions" in {
+      implicit val token: AuthToken = studentBToken
+
+      // this will run scatterCount^levels workflows, so be careful if increasing these values!
+      val topLevelMethod: Method = methodTree(levels = 2, scatterCount = 3)
+
+      withCleanBillingProject(studentB) { projectName =>
+        withWorkspace(projectName, "rawls-subworkflows-in-regions", workspaceRegion = Option("europe-north1")) { workspaceName =>
+          withCleanUp {
+            Orchestration.methodConfigurations.createMethodConfigInWorkspace(
+              projectName, workspaceName,
+              topLevelMethod,
+              topLevelMethodConfiguration.configNamespace, topLevelMethodConfiguration.configName, topLevelMethodConfiguration.snapshotId,
+              topLevelMethodConfiguration.inputs(topLevelMethod), topLevelMethodConfiguration.outputs(topLevelMethod), topLevelMethodConfiguration.rootEntityType)
+
+            Orchestration.importMetaData(projectName, workspaceName, "entities", SingleParticipant.participantEntity)
+
+            // it currently takes ~ 5 min for google bucket read permissions to propagate.
+            // We can't launch a workflow until this happens.
+            // See https://github.com/broadinstitute/workbench-libs/pull/61 and https://broadinstitute.atlassian.net/browse/GAWB-3327
+
+            Orchestration.workspaces.waitForBucketReadAccess(projectName, workspaceName)
+
+            val submissionId = Rawls.submissions.launchWorkflow(
+              projectName,
+              workspaceName,
+              topLevelMethodConfiguration.configNamespace,
+              topLevelMethodConfiguration.configName,
+              "participant",
+              SingleParticipant.entityId,
+              "this",
+              useCallCache = false,
+              deleteIntermediateOutputFiles = false
+            )
+            // clean up: Abort submission
+            register cleanUp Rawls.submissions.abortSubmission(projectName, workspaceName, submissionId)
+
+            // may need to wait for Cromwell to start processing workflows.  just take the first one we see.
+
+            val submissionPatience = PatienceConfig(timeout = scaled(Span(8, Minutes)), interval = scaled(Span(30, Seconds)))
+            implicit val patienceConfig: PatienceConfig = submissionPatience
+
+            // Get workflow IDs from submission details
+            val firstTierWorkflowIds: List[String] = eventually {
+              val (status, workflows) = Rawls.submissions.getSubmissionStatus(projectName, workspaceName, submissionId)
+              withClue(s"queue status: ${getQueueStatus()}, submission status: ${getSubmissionResponse(projectName, workspaceName, submissionId)}") {
+                status should (be("Submitted") or be("Done")) // very unlikely it's already done, but let's handle that case.
+                workflows.length should be(3)
+                workflows
+              }
+            }
+
+            // Fetch ALL call metadata from ALL subworkflows:
+            val callMetadataSections: List[JsonNode] = eventually {
+              for {
+                firstWorkflowId: String <- firstTierWorkflowIds
+                cromwellMetadata = Rawls.submissions.getWorkflowMetadata(projectName, workspaceName, submissionId, firstWorkflowId)
+
+                // validate options contain the :
+                submittedOptions = parseWorkflowOptionsFromMetadata(cromwellMetadata)
+                _ = submittedOptions should include ("europe-north1a")
+                _ = submittedOptions should include ("europe-north1b")
+                _ = submittedOptions should include ("europe-north1c")
+
+                subworkflowMetadata: JsonNode <- parseCallsFromMetadata(cromwellMetadata)
+                callMetadataWithinSubworkflows: JsonNode <- parseCallsFromMetadata(subworkflowMetadata)
+
+                _ = callMetadataWithinSubworkflows.get("executionStatus") should be("Done")
+
+              } yield callMetadataWithinSubworkflows
+            }
+
+
+            // verify that Rawls can retrieve the sub-sub-workflow's metadata without throwing an exception.
+
+            eventually {
+              Rawls.submissions.getWorkflowMetadata(projectName, workspaceName, submissionId, firstSubSubWorkflowId)
+            }
+
+            // verify that Rawls can retrieve the workflows' outputs from Cromwell without error
+            // https://github.com/DataBiosphere/firecloud-app/issues/157
+
+            val outputsTimeout = Timeout(scaled(Span(10, Seconds)))
+            eventually(outputsTimeout) {
+              Rawls.submissions.getWorkflowOutputs(projectName, workspaceName, submissionId, firstWorkflowId)
+              // nope https://github.com/DataBiosphere/firecloud-app/issues/160
+              //Rawls.submissions.getWorkflowOutputs(projectName, workspaceName, submissionId, firstSubWorkflowId)
+              //Rawls.submissions.getWorkflowOutputs(projectName, workspaceName, submissionId, firstSubSubWorkflowId)
+            }
+          }
+        }
+      }
+
+    }
+
 //    Disabling this test until we decide what to do with it. See AP-177
 
 //    "should retrieve metadata with widely scattered sub-workflows in a short time" in {
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
@@ -50,6 +50,21 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
     Await.result(find, 1.minute)
   }
 
+  def parseCallsFromMetadata(metadata: String): List[JsonNode] = {
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    // "calls" is a top-level key of metadata, whose value is a JSON object.
+    // Get that object's values.
+    val calls: List[JsonNode] = mapper.readTree(metadata).get("calls").elements().asScala.toList
+
+    // Call values are arrays of scatter shards.
+    // Get the shards, which are JSON objects.
+    val scatterShards: List[JsonNode] = calls flatMap { c =>
+      c.elements().asScala.toList
+    }
+  }
+
   def parseSubWorkflowIdsFromMetadata(metadata: String): List[String] = {
     /*
     Workflow metadata has this structure:
@@ -64,19 +79,7 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
       },  ...
     }
     */
-
-    val mapper = new ObjectMapper()
-    mapper.registerModule(DefaultScalaModule)
-
-    // "calls" is a top-level key of metadata, whose value is a JSON object.
-    // Get that object's values.
-    val calls: List[JsonNode] = mapper.readTree(metadata).get("calls").elements().asScala.toList
-
-    // Call values are arrays of scatter shards.
-    // Get the shards, which are JSON objects.
-    val scatterShards: List[JsonNode] = calls flatMap { c =>
-      c.elements().asScala.toList
-    }
+    val scatterShards: List[JsonNode] = parseCallsFromMetadata(metadata)
 
     // Return the values corresponding to each scatter shard's "subWorkflowId" key
     scatterShards map { s =>
@@ -100,6 +103,14 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
     val outputFieldJson = Option(outputsJson.get(outputField))
       .getOrElse(fail(s"Unable to find metadata output $outputField"))
     outputFieldJson.asText()
+  }
+
+  def parseWorkflowOptionsFromMetadata(metadata: String): String = {
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    val metadataJson = mapper.readTree(metadata)
+    metadataJson.get("submittedFiles").get("options").asText()
   }
 
   // if these prove useful anywhere else, they should move into workbench-libs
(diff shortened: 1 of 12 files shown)


salonishah11:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -771,6 +771,24 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
       }
   }
 
+  it should "return 400 if the bucket location requested is in an invalid format" in withTestDataApiServices { services =>
+    val newWorkspace = WorkspaceRequest(
+      namespace = testData.wsName.namespace,
+      name = "newWorkspace",
+      attributes = Map.empty,
+      authorizationDomain = None,
+      bucketLocation = Option("US")
```
cjllanwarne, line 780: Just to confirm - when we clone a workspace that's currently in the `US` multi-region, we don't even specify the location, right?

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -102,6 +105,26 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
 
+  def parseWorkflowOptionsFromMetadata(metadata: String): String = {
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    val metadataJson = mapper.readTree(metadata)
+    metadataJson.get("submittedFiles").get("options").asText()
+  }
+
+  def parseRuntimeAttributeKeyFromMetadata(metadata: List[JsonNode], attributeKey: String): List[String] = {
```
cjllanwarne:
```suggestion
  def parseRuntimeAttributeKeyFromCallMetadata(metadata: List[JsonNode], attributeKey: String): List[String] = {
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -102,6 +105,26 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+    metadata.map(_.get("runtimeAttributes").get("zones").asText())
+  }
+
+  def parseWorkerAssignedExecEvents(metadata: List[JsonNode]): List[String] = {
```
cjllanwarne:
```suggestion
  def parseWorkerAssignedExecEventsFromCallMetadata(metadata: List[JsonNode]): List[String] = {
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -102,6 +105,26 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+    metadata.flatMap{ call =>
```
cjllanwarne:
```suggestion
    metadata.flatMap { call =>
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
cjllanwarne: Personal preference: I would probably make this function (and `parseWorkerAssignedExecEvents` below) operate on a single call metadata `JsonNode` and move the `map` up into the calling function. My reasoning is, it keeps the "transform one entry" logic separate from the "iterate over a list" logic.

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -102,6 +105,26 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+      val flattenedExecutionEvents = call.get("executionEvents").elements().asScala.toList
```
cjllanwarne: Why "flattened"?

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -209,6 +235,192 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
 
     }
 
+    "should be able to create workspace and run sub-workflow tasks in non-US regions" in {
```
cjllanwarne, line 236: TOL (the answer may well be "no"): is there enough similarity between these multiple tests to extract a common pattern?

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -209,6 +235,192 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+
+              workerAssignedExecEvents should not be (empty)
+              workerAssignedExecEvents foreach { event => event should include (europeNorth1ZonesPrefix) }
+            }
+          }
+        }
+      }
+    }
+
+    "should be able to run sub-workflow tasks in a cloned workspace in non-US regions" in {
```
cjllanwarne, line 331: If the "cloned workspace" test works, do we still need the "before we cloned it workspace" version of the test?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5669,6 +5669,9 @@ components:
           type: boolean
           description: Optional, false if not specified. If true, the workspace is created with a Billing Project owner but no workspace owner. Requires being a Billing Project owner.
           default: false
+        bucketLocation:
+          type: string
+          description: Region in which bucket attached to the workspace should be created. It is optional. If not provided, the bucket will be created in `us-central1`
```
cjllanwarne:
```suggestion
          description: Region (NOT multi-region) in which bucket attached to the workspace should be created. If not provided, the bucket will be created in the 'US' multi-region.
```

cjllanwarne: You can get the "It is optional" by adding a `required: false` field under `bucketLocation` in the yaml.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -242,7 +249,22 @@ class HttpGoogleServicesDAO(
...
+        traceId = Option(traceId),
+        bucketPolicyOnlyEnabled = true,
+        logBucket = Option(GcsBucketName(getStorageLogsBucketName(googleProject))),
+        location = bucketLocation
+      ).compile.drain.unsafeToFuture().recoverWith{
+        case e: StorageException if e.getCode == 400 =>
+          val message = s"Workspace creation failed. Error trying to create bucket `$bucketName` in Google project " +
+            s"`${googleProject.value}` in region `${bucketLocation.getOrElse("US (default)")}`."
+          val errorReport = ErrorReport(statusCode = StatusCodes.BadRequest, message)
+          throw new RawlsExceptionWithErrorReport(errorReport)
```
cjllanwarne, line 266:
```suggestion
          throw new RawlsExceptionWithErrorReport(errorReport, e)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -143,6 +143,21 @@ trait WorkspaceSupport {
...
 
+  def withWorkspaceBucketRegionCheck[T](bucketRegion: Option[String])(op: => Future[T]): Future[T] = {
+    bucketRegion match {
+      case Some(region) =>
+        // if the user specifies a region for the workspace bucket, it must be in the proper format for a single region
+        val singleRegionPattern = "[A-Za-z]+-[A-Za-z]+[0-9]+"
+        if (region.matches(singleRegionPattern)) op
+        else {
+          val err = ErrorReport(statusCode = StatusCodes.BadRequest, message = s"Workspace Bucket Location should be " +
+            s"of format: $singleRegionPattern.")
```
cjllanwarne: I think this might be more user friendly if we say `"Bucket location must be a single (not multi-) region in the google regions list (https://cloud.google.com/compute/docs/regions-zones)"` (ie rather than expecting them to interpret the regex)

cjllanwarne (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
rsasch: When a workspace is cloned, the location of the new workspace bucket is set to the same as the original (see https://github.com/broadinstitute/rawls/blob/d72a788693bf9892210f49871bdb7dfa16b148a0/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L593).

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
rsasch: 🤷 dunno, changing to `executionEvents`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
rsasch: the constructor for `RawlsExceptionWithErrorReport` takes only one argument (see https://github.com/broadinstitute/rawls/blob/d72a788693bf9892210f49871bdb7dfa16b148a0/model/src/main/scala/org/broadinstitute/dsde/rawls/RawlsException.scala#L7)

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
rsasch: The two tests are different because one runs a workflow in a newly-created workspace, while the other runs a workflow in a cloned workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
rsasch: We specifically chose this class because it doesn't return the whole error, and we wanted to avoid showing the entire stack trace to the user.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
rsasch: Because the check doesn't actually compare the region submitted to the list of regions from Google, we're hesitant to put that in the error message.  How about `"Workspace bucket location must be a single (not multi-) region of format: [A-Za-z]+-[A-Za-z]+[0-9]+"`?

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
salonishah11: Some portion of the code could be extracted but it didn't seem that the result would be worth the effort since it is test code.

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -50,6 +50,21 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
     Await.result(find, 1.minute)
   }
 
+  def parseCallsFromMetadata(metadata: String): List[JsonNode] = {
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    // "calls" is a top-level key of metadata, whose value is a JSON object.
```
cjllanwarne:
```suggestion
    // "calls" is a top-level key of metadata, whose value is a JSON object mapping call names to shard lists.
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -50,6 +50,21 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+    // Get that object's values.
+    val calls: List[JsonNode] = mapper.readTree(metadata).get("calls").elements().asScala.toList
+
+    // Call values are arrays of scatter shards.
+    // Get the shards, which are JSON objects.
```
cjllanwarne:
```suggestion
    // Call values are arrays of scatter shards.
    // Flatten this structure to get a single list containing all the shards.
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -209,6 +233,192 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
...
+            //    - the zones for each job that were determined by Cromwell and
+            //    - the worker assigned for the tasks
+            // belong to `europe-north1`
+            eventually {
+              val subWorkflowsMetadata = subWorkflowIds map { Rawls.submissions.getWorkflowMetadata(projectName, workspaceName, submissionId, _) }
+              val subWorkflowCallMetadata = subWorkflowsMetadata flatMap parseCallsFromMetadata
+              val callZones = subWorkflowCallMetadata map { parseRuntimeAttributeKeyFromCallMetadata(_, "zones") }
+              val workerAssignedExecEvents = subWorkflowCallMetadata flatMap { parseWorkerAssignedExecEventsFromCallMetadata }
+
+              subWorkflowsMetadata foreach { x => x should include (""""executionStatus":"Done"""") }
```
cjllanwarne, line 310: TOL: if you move everything other than the "is done" check to happen _after_ the `eventually`, the flow would be:

* Wait for completion
* Check once to make sure everything looks correct.

Right now the flow is:

* Wait for everything to look correct.

It's a small difference but it means you'd hopefully get a failure result faster if something goes wrong

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
cjllanwarne: My thought was just that if you're able to "create and clone" a workspace in a different region, and everything works correctly, does that automatically imply that the "just create" case must also work? (and therefore that we don't need to test it separately).

I think you're definitely being more thorough, fwiw, I was just wondering whether we could economize if these tests turn out to be extremely slow.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -143,6 +143,21 @@ trait WorkspaceSupport {
...
 
+  def withWorkspaceBucketRegionCheck[T](bucketRegion: Option[String])(op: => Future[T]): Future[T] = {
+    bucketRegion match {
+      case Some(region) =>
+        // if the user specifies a region for the workspace bucket, it must be in the proper format for a single region
+        val singleRegionPattern = "[A-Za-z]+-[A-Za-z]+[0-9]+"
+        if (region.matches(singleRegionPattern)) op
+        else {
+          val err = ErrorReport(statusCode = StatusCodes.BadRequest, message = s"Workspace bucket location must be a single " +
+            s"(not multi-) region of format: $singleRegionPattern.")
```
cjllanwarne, line 154: TOL still a little scary IMO. But I guess if we're only expecting this to be used by API power-users then that's ok (because most users would use a neat little UI element)

cjllanwarne (approved):
Only TOL-level comments remain

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
panentheos: TOL: Not commenting on this specific case, but as a general principle I don't think we should apply lower standards to test code vs application code, since they're both artifacts that we build and maintain.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -588,32 +590,39 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-                  val newAttrs = sourceWorkspaceContext.attributes ++ destWorkspaceRequest.attributes
-
-                  withNewWorkspaceContext(destWorkspaceRequest.copy(authorizationDomain = Option(newAuthDomain), attributes = newAttrs), dataAccess) { destWorkspaceContext =>
-                    dataAccess.entityQuery.copyAllEntities(sourceWorkspaceContext, destWorkspaceContext) andThen
-                      dataAccess.methodConfigurationQuery.listActive(sourceWorkspaceContext).flatMap { methodConfigShorts =>
-                        val inserts = methodConfigShorts.map { methodConfigShort =>
-                          dataAccess.methodConfigurationQuery.get(sourceWorkspaceContext, methodConfigShort.namespace, methodConfigShort.name).flatMap { methodConfig =>
-                            dataAccess.methodConfigurationQuery.create(destWorkspaceContext, methodConfig.get)
+          // if the source bucket is a regional bucket, retrieve the region as the destination bucket also needs to be created in the same region
+          val bucketLocationFuture: Future[Option[String]] = for {
```
panentheos, line 594: Question: What's the thinking around adding explicit types for `val`s vs letting them be inferred? (I'm assuming it could be inferred here, if we wanted)

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
salonishah11: Yes. Once we open this to all the users we can refine this and as you said with a neat UI element

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
salonishah11: I would still prefer to be more thorough, at least for now as this is a new feature. We can always come back and remove a test if we realize it's being slow or not needed anymore

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
salonishah11: That's a good point. I will refactor it 👍

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
panentheos: TOL: Some questions I find helpful in evaluating these decisions: "What exactly is likely to break in the future?" E.g. what kind of bug do we think will happen, such that this test is the best/only way to catch it?

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -143,6 +143,21 @@ trait WorkspaceSupport {
     }
   }
 
+  def withWorkspaceBucketRegionCheck[T](bucketRegion: Option[String])(op: => Future[T]): Future[T] = {
```
panentheos, line 146: I'm curious why we wrote this as a 'with' as opposed to just a plain function that validates and possibly throws? We don't seem to be doing anything that requires Futures here.

panentheos (approved):
Looks good.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
salonishah11: It's following what it's [surrounding code is doing](https://github.com/broadinstitute/rawls/blob/360fab8c12f920b6102415947f9f588b86d5cc60/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L176-L183) as well as the naming convention in Rawls when you want your code to do stuff and then run the block of code passed to it as argument. So here we say check the bucket region and then run the code block specified as `op: => Future[T]`. So I personally understand it as run the code with bucket region check.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
salonishah11: The rule of thumb I **try** to follow is to explicitly show the data types when the type can't be inferred by simply reading the code or if it would it make reading and understanding code easier without having to ask IntelliJ about it

salonishah11:
jenkins retest

salonishah11:
jenkins retest

PR #1340, opened 2021-01-06
[CA-964] add billing project name validation

marctalbott:
Ticket: [CA-964](https://broadworkbench.atlassian.net/browse/CA-964)
This validation is also enforced at the UI level, but now we are enforcing it in Rawls as well.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -19,7 +19,7 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.user.UserService._
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, RoleSupport, UserWiths}
 import org.broadinstitute.dsde.rawls.webservice.PerRequest.{PerRequestMessage, RequestComplete}
-import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport, StringValidationUtils}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -435,6 +436,7 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   }
 
   private def validateCreateProjectRequest(createProjectRequest: CreateRawlsBillingProjectFullRequest): Future[Unit] = {
+    validateBillingProjectName(createProjectRequest.projectName.value)
```
dvoet: This is not great from a FP and `Future` perspective. The return value is `Future[Unit]`. The expectation is that it returns a Future which could be failure or success. This changes it so it returns a `Future` or throws an exception. This works because the calling code is tolerant. It would be better to return a `Future.failed` as the code below does.

dvoet (changes requested):

model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala
```diff
@@ -30,6 +32,16 @@ trait StringValidationUtils {
     }
   }
 
+  private lazy val billingProjectNameRegex = "[A-z0-9_-]{6,30}".r
```
marctalbott, line 35: could also move this into `UserService`, I left it here with the other string validators, but it's a little different as the only one to return a `Future` so if you would rather see it closer to the project creation code that uses it I'd be fine with that

s-rubenstein: I agree with you that this seems like the right place for this.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -218,6 +218,36 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
       }
   }
 
+  it should "return 400 when creating a project with a name that is too short" in withEmptyDatabaseAndApiServices { services =>
```
s-rubenstein, line 221: Non-Blocking: Just flagging here that it is deeply uncomfortable to me that these tests are effectively duplicated in two places.

s-rubenstein (approved):
Pending tests passing, this looks good to me.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
marctalbott: Agreed, I don't love the duplicated tests. I could move the tests into `UserServiceSpec`, but there would still need to be duplication to be able to test both versions of billing project creation. It'd probably be easier to remove the duplication if it's all contained within the versioned API spec, but I'd be fine with moving the duplication to `UserServiceSpec` instead if that's preferred

dvoet (approved):

PR #854, opened 2018-02-09
Create method configs for Dockstore, Agora [GAWB-3100]

aednichols:
To try this out, you can spin up UI -> Orch -> Front Rawls and use the APIs below.

Dockstore configs show up on the Method Configurations page and look only mildly broken.

Of course, you can't actually do anything with them until GAWB-3101.

To use the new functionality from Orch, https://github.com/broadinstitute/firecloud-orchestration/pull/607 is required; but since these changes are strictly additive, this PR can safely be merged first (i.e. everything that is documented and currently works in Orch, will keep working).

Still needs
- [ ] Signoff on the changes to the API relative to the tech doc!
- [ ] Update tech doc
- [x] Fix existing tests
- [x] Tests for the new things
- [ ] Update Rawls Swagger
- [ ] Update Orch Swagger
- [x] Small [UI PR](https://github.com/broadinstitute/firecloud-ui/pull/1180) to prettify Method Configurations Page

---

`POST /api/workspaces/{workspaceNamespace}/{workspaceName}/methodconfigs` now accepts and does the right thing with any of
```
"methodRepoMethod": { // Agora - backwards compatible
  "methodNamespace": "adam-methods",
  "methodName": "test",
   "methodVersion": 555
}
```
```
"methodRepoMethod": {
  "sourceRepo": "agora",
  "methodNamespace": "adam-methods",
  "methodName": "test",
   "methodVersion": 555
}
```
```
"methodRepoMethod": {
  "sourceRepo": "dockstore",
  "methodPath": "broadinstitute/wdl/Validate-Bams",
  "methodVersion": "develop"
}
```
```
"methodRepoMethod": {
  "methodUri": "dockstore://broadinstitute%2Fwdl%2FValidate-Bams/develop",
}
```
```
"methodRepoMethod": {
  "methodUri": "agora://adam-methods/test/555",
}
```

`GET /api/workspaces/{workspaceNamespace}/{workspaceName}/methodconfigs` now returns
```
// All existing fields plus two new ones
"methodRepoMethod": {
  "sourceRepo": "agora",
  "methodUri": "agora://adam-methods/test/555",
  "methodNamespace": "adam-methods", 
  "methodName": "test",
  "methodVersion": 555
}
```
for Agora MC's, and never returns any Dockstore configs.

`GET /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/methodconfigs` is new and returns both Agora and Dockstore configs, like
```
"methodRepoMethod": {
[...]

PR #1351, opened 2021-01-27
allow creating method configs for dockstore tools [CROM-6688]

panentheos:
This adds a new 'dockstoretools` repo definition which allows working with method configs that point to dockstore 'tools' (as opposed to 'workflows'). Verified by creating a method config from the swagger, then configuring it and submitting it with a slightly modified local UI.

Note that this will require followup UI works to handle the new source repo.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
@@ -23,6 +23,8 @@ import org.broadinstitute.dsde.rawls.config.MethodRepoConfig
 import org.broadinstitute.dsde.rawls.model.MethodRepoJsonSupport._
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport._
 import org.broadinstitute.dsde.rawls.model.StatusJsonSupport._
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets.UTF_8
 
 /**
  * @author tsharpe
@@ -53,6 +55,10 @@ class HttpMethodRepoDAO(agoraConfig: MethodRepoConfig[Agora.type], dockstoreConf
     getAgoraEntity(s"$agoraServiceURL/configurations/$namespace/$name/$version",userInfo)
   }
 
+  private def noneIfNotFound[T]: PartialFunction[Throwable, Option[T]] = {
+    case notOK: RawlsExceptionWithErrorReport if notOK.errorReport.statusCode.contains(StatusCodes.NotFound) => None
+  }
+
   override def getMethod( method: MethodRepoMethod, userInfo: UserInfo ): Future[Option[WDL]] = {
     method match {
       case agoraMethod: AgoraMethod =>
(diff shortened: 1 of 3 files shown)


salonishah11 (changes requested):
This looks great! It would be good to add some unit tests for this. It looks like there are some tests for Dockstore and Agora in [WorkspaceModelSpec.scala](https://github.com/broadinstitute/rawls/blob/develop/model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala). Maybe you could leverage that and add some more for DockstoreTools?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -380,6 +381,35 @@ object DockstoreMethod {
...
+    } else {
+      throw new RawlsException(s"Could not generate a method URI from DockstoreToolsMethod with path \'$methodPath\', version \'$methodVersion\'")
+    }
+  }
+
+  override def repo: MethodRepository = DockstoreTools
+}
+
+object DockstoreToolsMethod {
+  def apply(uri: String): DockstoreToolsMethod = {
```
aednichols, line 408: Can we DRY out the `apply` method between `DockstoreToolsMethod` and `DockstoreMethod`? Looks similar or identical.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -392,15 +422,20 @@ case object Dockstore extends MethodRepository {
...
 
   def withName(name: String): Option[MethodRepository] = name.toLowerCase match {
     case Agora.scheme => Some(Agora)
     case Dockstore.scheme => Some(Dockstore)
+    case DockstoreTools.scheme => Some(DockstoreTools)
     case _ => None
   }
 
-  val all: Set[MethodRepository] = Set(Agora, Dockstore)
+  val all: Set[MethodRepository] = Set(Agora, Dockstore, DockstoreTools)
```
aednichols, line 443: It would be nice to throw in a comment, perhaps here, explaining the domain concept difference between `Dockstore` and `DockstoreTools`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
```diff
@@ -68,21 +74,23 @@ class HttpMethodRepoDAO(agoraConfig: MethodRepoConfig[Agora.type], dockstoreConf
             tool: GA4GHTool <- maybeTool
           } yield WdlUrl(tool.url) // We submit the Github URL to Cromwell so that relative imports work.
         }
+      case DockstoreToolsMethod(methodPath, methodVersion) => {
+        val url = s"${dockstoreServiceURL}/ga4gh/v1/tools/${URLEncoder.encode(methodPath, UTF_8.name)}/versions/${URLEncoder.encode(methodVersion, UTF_8.name)}/WDL/descriptor"
```
aednichols: Might be nice to stash this big URL in a method that mirrors `getDockstoreMethod` and `ga4ghDescriptorUrl` for consistency

aednichols (changes requested):
This looks basically correct to me, just a few minor consistency/organization/documentation comments.

I originally created the `MethodRepoMethod` and `MethodRepository` super-classes back in 2018 [0][1], fun to see them being extended.

[0] https://github.com/broadinstitute/rawls/pull/854
[1] https://github.com/broadinstitute/rawls/pull/873

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
panentheos: I thought about that, but since the value is only used here I came to the opinion that reducing the code complexity by removing a layer of indirection was a win, even if it was slightly less consistent with the existing case.

Thoughts? (What about inlining the other one as well?)

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
panentheos: Fair enough, I'll pull out a helper function.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
panentheos: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpMethodRepoDAO.scala
aednichols: My thinking with the indirection is that `ga4ghDescriptorUrl` is a natural member of the domain object `DockstoreMethod`. It can be used to dereference a method and is itself constructed from other members of the same object. I would prefer we keep that pattern but don't feel _too_ strongly.

panentheos: Thanks for the input, I'll maintain the existing pattern.

I'd like to continue exploring this concept as we go forward, since I believe in general that structuring code according to its current usage patterns yields a smaller and simpler code base, which is more amenable to reading and changing.

aednichols: Sounds good, we don't want to end up with a `DockstoreMethodFactory` or any other excessive OO

aednichols (approved):
Changes look great to me 👍 thank you!

salonishah11 (approved):
💯

PR #1355, opened 2021-01-28
AS-494: Entity Statistics Cache

MatthewBemis:
The logic for the monitor takes the following into consideration:

* I didn't want to overwhelm back rawls or the DB with tons of backfilling so workspaces with a NULL cacheLastUpdated field are ignored. The way I get around this is that the first time a cache request comes in for a workspace, update the cacheLastUpdated field to lastModified-1 so it will get picked up by the monitor
* The lastModified field on the workspace serves as an etag of sorts. If entities in a workspace are updated mid-sweep, the old lastModified date is passed in and saved as the new cacheLastUpdated date. If cacheLastUpdated were to use the time that the sweep was completed, we may end up in situations where the cache stays stale.

It's also resilient to reboots, utilizing the DB to decide what needs updating rather than actor messaging or anything.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -62,21 +62,24 @@ object BootMonitors extends LazyLogging {
...
-    startSubmissionMonitorSupervisor(
-      system,
-      submissionMonitorConfig,
-      dataSourceAccess,
-      samDAO,
-      gcsDAO,
-      shardedExecutionServiceCluster,
-      metricsPrefix
-    )
+//    startSubmissionMonitorSupervisor(
```
MatthewBemis: Ignore this, just commented out because I wanted a back rawls locally but didn't want all of these other monitors spamming my logs

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -370,8 +376,18 @@ trait WorkspaceComponent {
       filter(rec => (rec.namespace === namespaceName.value))
     }
 
+    def listOutdatedEntityCaches(limit: Int): ReadAction[Seq[(UUID, Timestamp)]] = {
+      //We don't necessarily want all of the workspaces back, just the ones that we can handle in one pass
+      //We also order by lastModified so we can handle the ones that are the longest out of date
+      filter(rec => (rec.entityCacheLastUpdated < rec.lastModified)).take(limit).sortBy(_.lastModified.asc).map { ws => (ws.id, ws.lastModified) }.result
```
MatthewBemis: There are a few cases where lastModified may be updated even when entities aren't updated. All of the cases are:

* Updating entities directly
* A submission completing (may update entities, might not- hard to know)
* A workspace attribute being updated
* A method config being updated

I was on the fence about it being worth it to make this more specific, but it might be worth addressing to reduce any potential load. I think those events are probably lower volume, however.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
 
   private val workspaceContext = workspace
 
-  override def entityTypeMetadata(): Future[Map[String, EntityTypeMetadata]] = {
+  override def entityTypeMetadata(useCache: Boolean): Future[Map[String, EntityTypeMetadata]] = {
     dataSource.inTransaction { dataAccess =>
-      dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+      dataAccess.workspaceQuery.isEntityCacheDefined(workspaceContext.workspaceIdAsUUID).flatMap { cacheEntriesExistOpt =>
+        val cacheEntriesExist = cacheEntriesExistOpt.getOrElse(false)
```
MatthewBemis: this logic is important. if the user tries to use the cache and it isn't populated, we'll route the request through to the original queries and mark the workspace so it gets picked up by the monitor. this means that some cache requests may still be slow, but only temporarily

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
davidangb: `filter(rec => (rec.entityCacheLastUpdated < rec.lastModified))` does not return records where `entityCacheLastUpdated` is null, is that correct?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -0,0 +1,70 @@
...
+      _ <- recordsToUpdate.toList.traverse { workspaceToUpdate =>
+        IO.fromFuture(IO(updateStatisticsCache(workspaceToUpdate._1, workspaceToUpdate._2)))
+      }.unsafeToFuture()
+    } yield ()
+
+    checkFuture.failed.foreach {
+      // there was a failure, log it and it will retry later
+      t: Throwable => logger.error("Error updating statistics cache", t)
+    }
+  }
```
davidangb, line 64: as written, this processes `workspacesPerSweep` every `pollInterval`, for instance 5 workspaces every minute. To keep the cache as up-to-date as possible, I'd love to see this move off the scheduler and instead process as fast as it can. When `sweep()` finishes, if `recordsToUpdate` had `workspacesPerSweep` rows, immediately start another pass by sending `Sweep` to self. If `recordsToUpdate` was empty or had fewer than `workspacesPerSweep` rows, wait for `pollInterval` before running again.

As a possible enhancement - instead of just waiting for `pollInterval` before sweeping  again, find N workspaces whose  `entity_cache_last_updated` is  null, and set them to a value that will trigger them to populate their cache. This will take care of the backlog of unpopulated workspaces!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210129_add_entity_statistics_cache_tables.xml
```diff
@@ -0,0 +1,61 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="ENTITY_TYPE_COUNT" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="create_WORKSPACE_ENTITY_TYPE_STATISTICS_FK">
+        <addForeignKeyConstraint baseColumnNames="WORKSPACE_ID" baseTableName="WORKSPACE_ENTITY_TYPE_STATISTICS" constraintName="FK_WETS_WORKSPACE" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="WORKSPACE"/>
```
davidangb: set both of the foreign keys in this changeset to onDelete="CASCADE" ( I think that's the right syntax, pls double-check ). Else, you'll need to manually delete the rows when the workspace is deleted, or you'll get errors deleting the workspace due to fk constraints.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
davidangb: not sure this will make a difference in practice, but perhaps this should sort by `(lastModified-entityCacheLastUpdated) desc, lastModified asc`  ... process the most out-of-date workspaces first.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+        }
+        else if(useCache && !cacheEntriesExist) {
+          dataAccess.workspaceQuery.updateCacheLastUpdated(workspaceContext.workspaceIdAsUUID, new Timestamp(workspaceContext.lastModified.getMillis - 1)).flatMap { _ =>
+            dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+          }
+        }
+        else {
+          dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+        }
+      }
```
davidangb, line 46: I'd really like to track metrics on this, so we can judge success/improvement:
* how fast are the non-cached queries?
* how fast are the cached queries?
* how many user requests are satisfied by cache?
* how many user requests asked for cache but had no cache to pull from?
* how many user requests specified useCache=false (the default)?

(could do this in a separate PR if needed; I'm not sure of the best place to persist these metrics)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -0,0 +1,70 @@
...
+        _ <- dataAccess.entityTypeStatisticsQuery.deleteAllForWorkspace(workspaceId)
+        _ <- dataAccess.entityTypeStatisticsQuery.batchInsert(workspaceId, entityTypesWithCounts)
+        //update entity attribute statistics
+        entityTypesWithAttrNames <- dataAccess.entityQuery.getAttrNamesAndEntityTypes(workspaceId)
+        _ <- dataAccess.entityAttributeStatisticsQuery.deleteAllForWorkspace(workspaceId)
+        _ <- dataAccess.entityAttributeStatisticsQuery.batchInsert(workspaceId, entityTypesWithAttrNames)
+        //update cache update date
+        _ <- dataAccess.workspaceQuery.updateCacheLastUpdated(workspaceId, timestamp)
+      } yield ()
+    }
```
davidangb, line 80: this pattern - a shared table where we delete an entire workspace's worth of rows and insert an entire workspace's worth of rows - is the same pattern we're seeing with ENTITY_ATTTRIBUTE_SCRATCH and deadlocks. We shouldn't have problems in this implementation because  everything runs through the  `EntityStatisticsCacheMonitor`, essentially single-threaded ... but we need to watch out for that if we ever  implement concurrent cache updates.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+            }
+          }
+        }
+        else if(useCache && !cacheEntriesExist) {
+          dataAccess.workspaceQuery.updateCacheLastUpdated(workspaceContext.workspaceIdAsUUID, new Timestamp(workspaceContext.lastModified.getMillis - 1)).flatMap { _ =>
+            dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+          }
+        }
+        else {
+          dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
```
davidangb: even if `useCache` is false, if we know the cache is up to date, is there a reason to NOT use the cache?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -0,0 +1,70 @@
...
+  private def sweep() = {
+    val checkFuture = for {
+      recordsToUpdate <- datasource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listOutdatedEntityCaches(limit)
+      }
+
+      _ <- recordsToUpdate.toList.traverse { workspaceToUpdate =>
+        IO.fromFuture(IO(updateStatisticsCache(workspaceToUpdate._1, workspaceToUpdate._2)))
+      }.unsafeToFuture()
+    } yield ()
```
davidangb: add logging for each workspace processed, that includes at least the workspace id and the gap between workspace.lastModified and workspace.entityCacheLastUpdated. I think we'll want to be able to monitor how often we update cache on each workspace and how out of date the caches  are.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
davidangb: a future possibility is to add a database trigger to the ENTITY table; every insert/update/delete could mark the parent workspace as needing a cache update.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+            typesAndAttrsQ map { typesAndAttrs =>
+              (typesAndCounts.keySet ++ typesAndAttrs.keySet) map { entityType =>
+                (entityType, EntityTypeMetadata(
+                  typesAndCounts.getOrElse(entityType, 0),
+                  entityType + Attributable.entityIdAttributeSuffix,
+                  typesAndAttrs.getOrElse(entityType, Seq()).map (AttributeName.toDelimitedName).sortBy(_.toLowerCase)))
+              } toMap
+            }
+          }
+        }
```
davidangb: in the case where we satisfy the user's request from cache, I'd be curious to record metrics about cache freshness. Is it worth going to the db so we can log (or send metrics somewhere) the value of `workspace.lastModified - workspace.entityCacheLastUpdated`?

Another thought - in the implementation of `isEntityCacheDefined`, check to  see if `workspace.lastModified = workspace.entityCacheLastUpdated` instead of checking `workspace.entityCacheLastUpdated is null`

davidangb (commented):
first-round comments inline! Feel free to ping with questions, I know this is a complicated one.

I'd love to see some tests for the monitor that builds the caches as well as tests for the LocalEntityProvider changes, too.

core/src/test/resources/reference.conf
```diff
@@ -70,6 +70,11 @@ entityUpsert {
   maxContentSizeBytes = 67108864
 }
 
+entityStatisticsCache {
+  workspacesPerSweep = 5
```
ansingh7115: How did you arrive at 5?  that seems like a small number

ansingh7115:
Are UI changes going in accompanying this or are we merging this first and seeing if the UI changes are necessary? (Joy has updated [this ticket](https://broadworkbench.atlassian.net/browse/AS-591) with UX designs)

Do we have a plan for monitoring production after this goes live? And what happens if we need to rollback for some reason - will we be able to do that easily?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
MatthewBemis: That's a good point but I would like to leave that for a quick follow-up ticket if this looks good in production. I'd like to keep the risk as low as possible by making this purely opt-in and explicit to start

core/src/test/resources/reference.conf
MatthewBemis: It felt like a safe number to start with. My main reason for keeping it low is in case multiple large workspace got picked up in the same sweep. Overriding in rawls.conf is quick enough so we could easily bump it up over time. I haven't really tested this on a database with production-level numbers, so I'm erring on the side of caution.

Although given one of David An's comments in the monitor, I may be re-thinking how the sweeping works in general.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
MatthewBemis: I really like this idea, I've now implemented it in a recent commit (with the extra enhancement to handle backlogged workspaces)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: That's correct

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
MatthewBemis: done

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
MatthewBemis: I'm making this change but noting here that doing so means that we'll never use the cache if it's out of date. Also renamed to `isEntityCacheCurrent`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: I don't believe I can use the `-` operator in NumericColumnExtensionMethods on these columns because they're not NumericColumns

MatthewBemis:
@ansingh7115 this can go in ahead of any UX changes. it's purely opt-in, and with one of the recent commits to only use the cache if it's up to date, the only UI change that we'll need is to call the entity metadata endpoint with `useCache=true`. The rollback procedure for this would be to update the UI to no longer call with that param. I also just added a kill-switch in the rawls config so we can disable the cache entirely if it causea problems in the DB or in back rawls.

@ansingh7115 & @davidangb this is ready for re-review. i have deferred all monitoring and metrics for a follow-up task. let me know if you think that's a bad move. we can probably set up some decent Logit alerts pretty easily as in the interim

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: (that said, I could do it in memory or write raw sql, but I don't think it's worth it)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,33 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
 
   private val workspaceContext = workspace
 
-  override def entityTypeMetadata(): Future[Map[String, EntityTypeMetadata]] = {
+  override def entityTypeMetadata(useCache: Boolean): Future[Map[String, EntityTypeMetadata]] = {
     dataSource.inTransaction { dataAccess =>
-      dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+      dataAccess.workspaceQuery.isEntityCacheCurrent(workspaceContext.workspaceIdAsUUID).flatMap { isCacheCurrent =>
+        if(useCache && isCacheCurrent) {
```
MatthewBemis: I realized in my sleep that the kill switch I added yesterday should also be factored in here. Will add that once I get my day started

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -0,0 +1,112 @@
...
+        }
+      }.unsafeToFuture()
+    } yield {
+      updateResults.foreach { case (workspaceId, lastModified, cacheUpdated) =>
+        logger.info(s"Updated workspace $workspaceId. Cache was ${cacheUpdated - lastModified} millis out of date")
+      }
+      logger.info(s"Sweep complete. Updated entity cache for ${recordsToUpdate.length} workspace(s), with the limit set to $limit")
+      if(recordsToUpdate.length < limit) HandleBacklog
+      else Sweep
+    }
```
davidangb: I think this still needs some kind of poll interval or delay. While there are workspaces to update - including the backlog  -  it's good to work as fast as possible (I wonder if we want a few-millisecond delay just so we don't overwhelm the db). But ... once we've caught up on all workspaces, we should not be sweeping as fast as possible; there should be a delay once we've caught up.

davidangb (commented):
one quick note, will do a more in-depth review tomorrow!

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1760,6 +1760,12 @@ paths:
           required: true
           schema:
             type: string
+        - name: useCache
+          in: query
+          description: True to read entity statistics from cache (eventually consistent), false to force latest statistics (slower). First time reading from the cache may force latest statistics. Default is false.
+          schema:
+            type: boolean
+            default: false
```
davidangb, line 1768: nit: since the code now checks if the cache is up to date, you can remove the "eventually consistent" comment

core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -10,6 +10,12 @@ paths:
       parameters:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
         - $ref: '#/components/parameters/workspaceNamePathParam'
+        - name: useCache
+          in: query
+          description: True to read entity statistics from cache (eventually consistent), false to force latest statistics (slower). First time reading from the cache may force latest statistics. Default is false.
+          schema:
+            type: boolean
+            default: false
```
davidangb, line 18: same, remove "eventually consistent"

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -370,8 +376,30 @@ trait WorkspaceComponent {
...
+    }
+
+    def listBackloggedEntityCaches(limit: Int): ReadAction[Seq[(UUID, Timestamp)]] = {
+      filter(rec => rec.entityCacheLastUpdated.isEmpty).take(limit).map { ws => (ws.id, ws.lastModified) }.result
+    }
```
davidangb: add `.sortBy(_.lastModified.desc)`, so we're catching up on the most-recently updated workspaces first?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+          typesAndCountsQ flatMap { typesAndCounts =>
+            typesAndAttrsQ map { typesAndAttrs =>
+              (typesAndCounts.keySet ++ typesAndAttrs.keySet) map { entityType =>
+                (entityType, EntityTypeMetadata(
+                  typesAndCounts.getOrElse(entityType, 0),
+                  entityType + Attributable.entityIdAttributeSuffix,
+                  typesAndAttrs.getOrElse(entityType, Seq()).map(AttributeName.toDelimitedName).sortBy(_.toLowerCase)))
+              } toMap
+            }
+          }
```
davidangb: this block is copy-pasted from `EntityComponent` - could it be extracted into a reusable method that accepts `typesAndCountsQ`  and `typesAndAttrsQ` as arguments?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -29,9 +31,34 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+  override def entityTypeMetadata(useCache: Boolean): Future[Map[String, EntityTypeMetadata]] = {
     dataSource.inTransaction { dataAccess =>
-      dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+      dataAccess.workspaceQuery.isEntityCacheCurrent(workspaceContext.workspaceIdAsUUID).flatMap { isCacheCurrentOpt =>
+        //If the cache doesn't exist; return full query results and mark the workspace for an update
+        if(isCacheCurrentOpt.isEmpty) {
+          dataAccess.workspaceQuery.updateCacheLastUpdated(workspaceContext.workspaceIdAsUUID, new Timestamp(workspaceContext.lastModified.getMillis - 1)).flatMap { _ =>
+            dataAccess.entityQuery.getEntityTypeMetadata(workspaceContext)
+          }
+        }
```
davidangb: 1. I'm wondering what happens here, in the case where `getEntityTypeMetadata` is too slow ... is there a case in which `entityQuery.getEntityTypeMetadata` fails, thus the transaction gets rolled back, thus `updateCacheLastUpdated` is also rolled back and we don't mark this workspace as needing its cache calculated?

2. If the cache exists, but is out of date ... the call to `updateCacheLastUpdated` will change the cache-last-updated value, obscuring any calculation of how out of date it was previously.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -0,0 +1,112 @@
...
+    val checkFuture = for {
+      recordsToUpdate <- dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listOutdatedEntityCaches(limit)
+      }
+
+      updateResults <- recordsToUpdate.toList.traverse { case (workspaceId, lastModified) =>
+        IO.fromFuture(IO(updateStatisticsCache(workspaceId, lastModified))).map { _ =>
+          (workspaceId, lastModified.getTime, System.currentTimeMillis())
+        }
+      }.unsafeToFuture()
```
davidangb: probably need to mess with the akka receive timeout/inactivity timeout. If `updateStatisticsCache` is slow, the `EntityStatisticsCacheMonitorActor` could see no messages for an extended period and shut itself down due to inactivity.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
```diff
@@ -128,5 +135,116 @@ class LocalEntityProviderSpec extends AnyWordSpecLike with Matchers with TestDri
...
+      val typeCountCache = runAndWait(dataSource.dataAccess.entityTypeStatisticsQuery.getAll(workspaceContext.workspaceIdAsUUID))
+      val attrNamesCache = runAndWait(dataSource.dataAccess.entityAttributeStatisticsQuery.getAll(workspaceContext.workspaceIdAsUUID))
+
+      //assert that there is something in the cache for this workspace
+      typeCountCache should not be Map.empty
+      attrNamesCache should not be Map.empty
+
+      entityTypeMetadataResult should contain theSameElementsAs expectedResultWhenUsingFullQueries
+    }
+
```
davidangb, line 224: another simple smoke test:
* Create workspace, populate entities
* get entity metadata (cache does not yet exist)
* populate cache
* get entity metadata, compare result to previous, should be the same.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
davidangb: I'm worried about a case where we have a problem workspace. I'm not sure what condition would cause this, but if we have a workspace that consistently fails to calculate its cache ... the monitor can get stuck on that workspace, continually trying and failing to process it, rolling back the transaction, and moving on to the next sweep ... in which the same problem workspace is again processed. We'd be stuck in a loop.

davidangb (commented):
more comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
ansingh7115: For a workspace like this - could we log that it failed to process and set its cache values to null or 0?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
MatthewBemis: 1. That's a good point, there is probably room for improvement for error-handling here. I will put some more thought into this.

2. That shouldn't happen. If `isCacheCurrentOpt.isEmpty` = `false` and `isCacheCurrentOpt.getOrElse(false)` = `false`, then it will fall to the `else` branch at the end which does not update the timestamp. `isCacheCurrentOpt` is an `Option[Boolean]` that has three modes that dictate: cache is up to date, cache is not up to date, cache does not exist. Looking at it now, the readability isn't the greatest. I'll see if there's a better way to represent these states. At the very least I will add a test for this.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
MatthewBemis: LocalEntityProvider doesn't handle any of the work to populate the cache but I just added a test that verifies that we update the timestamp for a workspace without an existing cache.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
MatthewBemis: I've added a concept of a delayed sweep if we're caught up

MatthewBemis: I'm trying to think of a state that would allow us to mark these workspaces as ignored. Null or 0 will still cause them to be picked up by the monitor. I think setting the time to the future could technically work with some modified logic but that feels very cryptic to me so I'm not going to do that. This requires some more thought.

MatthewBemis: Thought about this a bunch more. I ended up going with Anu's suggestion of using 0 as the timestamp. I was a bit hesitant to use a magic/cryptic value but I can't think of any other way to possibly represent an ignored workspace given the way that cache state is currently represented. I really want to avoid complicating the implementation by adding another database column or introducing more comprehensive states. IMO everything should be driven off of timestamps. Hopefully these cases are extremely rare. At least in these cases, the endpoint will just default to the non-cached version and still return the correct result.

LMK what you think.

MatthewBemis:
@ansingh7115 @davidangb please take another look, I think I've handled all of the comments by now

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
davidangb: code here looks correct. For readability, it would be great if `cacheStatus` was an enum; I wouldn't hold the PR just for that.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
```diff
@@ -246,5 +246,37 @@ class LocalEntityProviderSpec extends AnyWordSpecLike with Matchers with TestDri
       entityTypeMetadataResult should contain theSameElementsAs expectedResultWhenUsingFullQueries
     }
 
+    "update the entity cache last updated field if we ask for the cache but the last updated field is null" in withLocalEntityProviderTestDatabase { dataSource =>
+      val workspaceContext = runAndWait(dataSource.dataAccess.workspaceQuery.findById(localEntityProviderTestData.workspace.workspaceId)).get
+      val localEntityProvider = new LocalEntityProvider(workspaceContext, slickDataSource, true)
+
+      //Update the entityCacheLastUpdated field to be identical to lastModified, so we can test our scenario of having a fresh cache
+      runAndWait(workspaceQuery.updateCacheLastUpdated(workspaceContext.workspaceIdAsUUID, null))
```
davidangb: nit: the code comment here is misleading

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
```diff
@@ -246,5 +246,37 @@ class LocalEntityProviderSpec extends AnyWordSpecLike with Matchers with TestDri
...
+    }
+
+    "not update the last updated field if the cache exists but is out of date" in withLocalEntityProviderTestDatabase { dataSource =>
+      val workspaceContext = runAndWait(dataSource.dataAccess.workspaceQuery.findById(localEntityProviderTestData.workspace.workspaceId)).get
+      val localEntityProvider = new LocalEntityProvider(workspaceContext, slickDataSource, true)
+
+      val outdatedTimestmap = new Timestamp(workspaceContext.lastModified.getMillis - 1)
+
+      //Update the entityCacheLastUpdated field to be identical to lastModified, so we can test our scenario of having a fresh cache
+      runAndWait(workspaceQuery.updateCacheLastUpdated(workspaceContext.workspaceIdAsUUID, outdatedTimestmap))
```
davidangb: same here - please update code comment

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
davidangb: my apologies, I didn't mean the test had to go in this spec!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -61,7 +67,8 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
         logger.info(s"Updated workspace $workspaceId. Cache was ${cacheUpdated - lastModified} millis out of date")
       }
       logger.info(s"Sweep complete. Updated entity cache for ${recordsToUpdate.length} workspace(s), with the limit set to $limit")
-      if(recordsToUpdate.length < limit) HandleBacklog
+      if(recordsToUpdate.isEmpty) ScheduleDelayedSweep
+      else if(recordsToUpdate.length < limit) HandleBacklog
```
davidangb: I believe there's still a logic problem here. If the number of workspaces needing a cache update is an even multiple of `limit`, then this code will toggle between `Sweep` and `ScheduleDelayedSweep`, but never reach `HandleBacklog`, at least until a user triggers some workspace to need a cache update and the counts change.

Some thoughts on simplifying the business logic:
* make the `entity_cache_last_updated` db column non-null, with a default value of 1, and ask liquibase to set all current workspaces to 1 when it runs (for safety, it could set all workspaces to 1 where entity_cache_last_updated != last_modified)
* remove the `limit` configuration value and process workspaces one at a time - you'll always either get 1 or 0.

davidangb (commented):
some more comments inline. The more I think about it the more I really want tracing or other metrics - let's not hold this PR for those but let's try to get them in before this makes it to prod.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
ansingh7115: I think this is fine - I would hope that a workspace getting into a problematic state where the cache fails to calculate should be extremely rare. We should document somewhere in the code what a 0 timestamp designates, as well.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210129_add_entity_statistics_cache_tables.xml
```diff
@@ -0,0 +1,61 @@
...
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="create_WORKSPACE_ENTITY_ATTRIBUTE_STATISTICS_INDEX">
+        <createIndex indexName="IDX_WORKSPACE_ENTITY_ATTRIBUTE_STATS" tableName="WORKSPACE_ENTITY_ATTRIBUTE_STATISTICS" unique="false">
+            <column name="WORKSPACE_ID"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="add-cache-updated-ws">
+        <addColumn tableName="WORKSPACE">
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02"/>
```
MatthewBemis: `1970-01-01 00:00:01` is the earliest timestamp that MySQL supports, so we'll use that one to "ignore" workspaces.

hence why `1970-01-01 00:00:02` is the default

MatthewBemis:
@davidangb @ansingh7115 please take another look. i've simplified the logic in the monitor and LocalEntityProvider per feedback, and updated/added tests to go along with it

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210129_add_entity_statistics_cache_tables.xml
davidangb: golden opportunity here for an easter egg ... somebody's birthday? Some other important date?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -376,28 +376,19 @@ trait WorkspaceComponent {
...
-      //0 is a "magic" timestamp that means that the cache should not be updated because it produces
-      //an error that breaks the statistics cache monitor. These are unexpected cases that would
-      //need manual intervention.
-      filter(rec => (rec.entityCacheLastUpdated < rec.lastModified && rec.entityCacheLastUpdated > new Timestamp(0))).take(limit).sortBy(_.lastModified.asc).map { ws => (ws.id, ws.lastModified) }.result
+    def findMostOutdatedEntityCacheAfter(timestamp: Timestamp): ReadAction[Option[(UUID, Timestamp)]] = {
+      // Find the workspace that has the entity cache that is the most out of date:
+      // A. Workspace has a cacheLastUpdated date that is not current ("current" means equal to lastModified)
+      // B. cacheLastUpdated is after @param timestamp
+      // C. Ordered by lastModified from oldest to newest. Meaning, return the workspace that was modified the longest ago
+      uniqueResult[(UUID, Timestamp)](filter(rec => rec.entityCacheLastUpdated < rec.lastModified && rec.entityCacheLastUpdated > timestamp).sortBy(_.lastModified.asc).take(1).map { ws => (ws.id, ws.lastModified) })
```
davidangb, line 384: I know you didn't want to drop to plain SQL for this, but I wanted to figure out what it would entail. Getting the workspace id into a UUID is a big pain but the rest of it works as expected and allows sorting on the (lastModified-entityCacheLastUpdated) value.

```suggestion
      // Find the workspace that has the entity cache that is the most out of date:
      // A. Workspace has a cacheLastUpdated date that is not current ("current" means equal to lastModified)
      // B. cacheLastUpdated is after @param timestamp
      // C. Ordered by the difference between lastModified and cacheLastUpdated
      implicit object SetUUID extends SetParameter[UUID] { def apply(v: UUID, pp: PositionedParameters) { pp.setBytes(driver.columnTypes.uuidJdbcType.toBytes(v)) } }

      implicit val getIdTimestampTuple: GetResult[(UUID, Timestamp)] = GetResult[(UUID, Timestamp)] { r => (
        driver.columnTypes.uuidJdbcType.fromBytes(r.nextBytes()), r.<<) }
      val query =
        sql"""select id, last_modified
                from WORKSPACE
                where last_modified > ${timestamp}
                  AND last_modified > entity_cache_last_updated
                order by (last_modified - entity_cache_last_updated) desc
                limit 1"""
      uniqueResult[(UUID, Timestamp)](query.as[(UUID, Timestamp)])

```

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210129_add_entity_statistics_cache_tables.xml
```diff
@@ -54,7 +54,7 @@
 
     <changeSet logicalFilePath="dummy" author="mbemis" id="add-cache-updated-ws">
         <addColumn tableName="WORKSPACE">
-            <column name="entity_cache_last_updated" type="TIMESTAMP(6)"/>
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02"/>
```
davidangb: add `nullable=false`? Or whatever the exact syntax is

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -376,28 +376,19 @@ trait WorkspaceComponent {
...
-    // None: cache does not exist
-    def isEntityCacheCurrent(workspaceId: UUID): ReadAction[Option[Boolean]] = {
-      val queryResult = uniqueResult[(Timestamp, Option[Timestamp])](filter(rec => rec.id === workspaceId).map(ws => (ws.lastModified, ws.entityCacheLastUpdated)))
+    def isEntityCacheCurrent(workspaceId: UUID): ReadAction[Boolean] = {
+      val queryResult = uniqueResult[(Timestamp, Timestamp)](filter(rec => rec.id === workspaceId).map(ws => (ws.lastModified, ws.entityCacheLastUpdated)))
       queryResult.map {
-        case Some((lastModified, Some(entityCacheLastUpdated))) => Option(lastModified.equals(entityCacheLastUpdated))
-        case _ => None
+        case Some((lastModified, entityCacheLastUpdated)) => lastModified.equals(entityCacheLastUpdated)
+        case _ => false
```
davidangb, line 391: if `entityCacheLastUpdated` is now non-null, could you calculate the boolean of `ws.lastModified == ws.entityCacheLastUpdated` directly in the query and eliminate the match clause?

ansingh7115 (approved):
Went over the new changes and they look good to me!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -47,51 +45,20 @@ class EntityStatisticsCacheMonitorActor(val dataSource: SlickDataSource, val lim
...
     dataSource.inTransaction { dataAccess =>
-      for {
-        backloggedWorkspaces <- dataAccess.workspaceQuery.listBackloggedEntityCaches(limit)
-        numMarked <- DBIO.sequence(backloggedWorkspaces map { case (workspaceId, lastModified) =>
-          dataAccess.workspaceQuery.updateCacheLastUpdated(workspaceId, new Timestamp(lastModified.getTime - 1))
+      //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
+      dataAccess.workspaceQuery.findMostOutdatedEntityCacheAfter(new Timestamp(1000)).flatMap {
+        case Some((workspaceId, lastModified)) => DBIO.from(updateStatisticsCache(workspaceId, lastModified).map { _ =>
+          logger.info(s"Updated entity cache for workspace $workspaceId. Cache was ${System.currentTimeMillis() - lastModified.getTime}ms out of date.")
+          Sweep
```
davidangb, line 57: if you accept my plain-SQL suggestion for `findMostOutdatedEntityCacheAfter`, you could have that return its staleness calculation, and log the actual staleness here instead of using the current time for the diff.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitorSpec.scala
```diff
@@ -33,43 +42,111 @@ class EntityStatisticsCacheMonitorSpec(_system: ActorSystem) extends TestKit(_sy
...
+    val monitor = new EntityStatisticsCacheMonitor {
+      override val dataSource: SlickDataSource = slickDataSource
+      override implicit val executionContext: ExecutionContext = defaultExecutionContext
+      override val standardPollInterval: FiniteDuration = util.toScalaDuration(testConf.getDuration("entityStatisticsCache.standardPollInterval"))
+    }
+
+    val workspaceContext = runAndWait(slickDataSource.dataAccess.workspaceQuery.findById(localEntityProviderTestData.workspace.workspaceId)).get
+    val localEntityProvider = new LocalEntityProvider(workspaceContext, slickDataSource, cacheEnabled = true)
+
+    //Update the entityCacheLastUpdated field to be olader than lastModified, so we can test our scenario of having a stale cache
```
davidangb: nit: typo "olader"

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210129_add_entity_statistics_cache_tables.xml
MatthewBemis: I am failing to be creative right now :'(

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: I think there would still need to be a match somewhere even if I pushed that logic into the query because I'd still need to pass this through `uniqueResult` to coerce it into a single result

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: I'm inclined to push this off until the metrics follow-up work, and decide then if we should adopt this. I don't think this will have much effect in practice but I do think it could be useful for returning slightly more accurate metrics which is why I want to hold off until then.

davidangb (approved):
looks good, let's get it into dev and see how it does, and make follow-on PRs as necessary.

PR #1361, opened 2021-02-10
support updating snapshots [AS-512]

zarsky-broad:
This adds support for the new wsm snapshot patch endpoint.

You'll notice a significant change in the openapi definition file; this is because when I was adding this endpoint I discovered that the security stanza was repeated for every endpoint, and was the same for many of them, so I made that the default. The only endpoints that still have the stanza are those that are unauthorized or those that require the billing scope.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


zarsky-broad:
I missed a spec, it's unhappy, so I'll figure it out, but the logic should be fine to review.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -38,8 +38,12 @@ class HttpWorkspaceManagerDAO(baseWorkspaceManagerUrl: String)(implicit val syst
     getWorkspaceApi(accessToken).deleteWorkspace(workspaceId)
   }
 
-  override def createDataReference(workspaceId: UUID, name: DataReferenceName, referenceType: ReferenceTypeEnum, reference: DataRepoSnapshot, cloningInstructions: CloningInstructionsEnum, accessToken: OAuth2BearerToken): DataReferenceDescription = {
-    getWorkspaceApi(accessToken).createDataReference(new CreateDataReferenceRequestBody().name(name.value).referenceType(referenceType).reference(reference).cloningInstructions(cloningInstructions), workspaceId)
+  override def createDataReference(workspaceId: UUID, name: DataReferenceName, referenceDescription: Option[String], referenceType: ReferenceTypeEnum, reference: DataRepoSnapshot, cloningInstructions: CloningInstructionsEnum, accessToken: OAuth2BearerToken): DataReferenceDescription = {
+    getWorkspaceApi(accessToken).createDataReference(new CreateDataReferenceRequestBody().name(name.value).referenceDescription(referenceDescription.orNull).referenceType(referenceType).reference(reference).cloningInstructions(cloningInstructions), workspaceId)
+  }
```
davidangb, line 43: could you create a `DataReferenceDescription` value class parallel to `DataReferenceName`, to maintain the strong typing in these APIs?

core/src/main/resources/swagger/data-repo-only.yaml
```diff
@@ -200,22 +188,40 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UpdateDataReferenceRequestBody'
+      responses:
+        204:
+          description: Successful Request
+          content: { }
```
davidangb: omit `content` entirely to indicate an empty response (search for "Empty Response Body" here: https://swagger.io/docs/specification/describing-responses/)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
davidangb: and - I realize this would require changes in workspace manager - should we require a description, but give it a sensible default (empty string, maybe, or equal to the name)? It would avoid all of the `Option` and `null` handling.

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
```diff
@@ -55,6 +55,15 @@ class MockWorkspaceManagerDAO extends WorkspaceManagerDAO {
     mockEnumerateReferenceResponse(workspaceId)
   }
 
+  override def updateDataReference(workspaceId: UUID, referenceId: UUID, updateInfo: UpdateDataReferenceRequestBody, accessToken: OAuth2BearerToken): Unit = {
+    val existingRef = references.getOrElse((workspaceId, referenceId), throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not found")))
+    references.update((workspaceId, referenceId), existingRef.name(
+      if (updateInfo.getName != null) updateInfo.getName else existingRef.getName
+    ).referenceDescription(
+      if (updateInfo.getReferenceDescription != null) updateInfo.getReferenceDescription else existingRef.getReferenceDescription
```
davidangb: once a description is non-null, is there a way to set it back to null? As written, this will always maintain the existing description, instead of allowing someone to null it out.

I am totally aware this is a mock. Does the mock not reflect actual runtime behavior, or does the runtime behavior need some reconsideration to allow updating to null?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala
```diff
@@ -306,30 +287,77 @@ class SnapshotApiServiceSpec extends ApiServiceSpec {
...
+        Get(s"${baseSnapshotsPath}/${response.getReferenceId}") ~>
+          sealRoute(services.snapshotRoutes) ~>
+          check {
+            val response = responseAs[DataReferenceDescription]
+            assertResult(StatusCodes.OK) {status}
+            assert(response.getName == "foo2")
+            assert(response.getReferenceDescription == "bar2")
+          }
+      }
+  }
```
davidangb, line 310: since this test has multiple assertions of the same type (status codes), please add clues to assist with test-failure debugging, e.g. `assertResult(StatusCodes.Created, "Unexpected snapshot creation status") {status}`

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -19,6 +19,10 @@ object DataReferenceModelJsonSupport extends JsonSupport {
     case Some(notStr) => JsString(notStr.toString)
   }
 
+  def getOptionalField(jsObject: JsObject, fieldName: String): String = {
+    jsObject.getFields(fieldName).headOption.map(_.convertTo[String]).orNull
+  }
```
davidangb: As a Scala method, this should never return a `null`; it could return an `Option`. If calling code needs to interface with Java methods that require nulls, the calling code can add their own `.orNull` handling in place.

This also has a big assumption in the `convertTo` which could be unsafe - especially because the method name doesn't indicate anything about Strings, it only talks about generic Fields.

One possible reimplementation:
```
  def getOptionalStringField(jsObject: JsObject, fieldName: String): Option[String] = {
    jsObject.fields.get(fieldName) match {
      case Some(s:JsString) => Option(s.value)
      case _ => None
    }
  }
```

finally, can this method be private? That would make it easier for it to include assumptions.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModelSpec.scala
```diff
@@ -53,6 +65,18 @@ class DataReferenceModelSpec extends AnyFreeSpec with Matchers {
...
+        assertResult { s"""{"name":null,"referenceDescription":"foo"}""".parseJson } {
+          new UpdateDataReferenceRequestBody().referenceDescription("foo").toJson
+        }
+      }
+
+      "UpdateDataReferenceRequestBody with no parameters should fail" in {
+        assertThrows[DeserializationException] {
+          s"""{}""".parseJson.convertTo[UpdateDataReferenceRequestBody]
+        }
+      }
```
davidangb, line 74: for completeness, add one more test that includes both a name and description

davidangb (commented):
see comments inline!

zarsky-broad:
Switching back to draft, going to tweak WSM so it defaults to an empty string rather than `null`

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
zarsky-broad: We'll no longer ever have null - default is now empty string.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
zarsky-broad: Rescoped and renamed.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
zarsky-broad: I think my latest changes cover this sufficiently?

davidangb (approved):

MatthewBemis (approved):

zarsky-broad:
jenkins retest

PR #1364, opened 2021-02-19
[CA-1094] Update addProjectToServicePerimeter for v2 workspaces

marctalbott:
Ticket: [CA-1094](https://broadworkbench.atlassian.net/browse/CA-1094)
This breaks out the service perimeter logic into a separate service that both `UserService` and `WorkspaceService` can use to update perimeters. I ended up re-writing the `overwriteGoogleProjectsInPerimeter` function because I found that it wasn't throwing exceptions on failures from Google like it was supposed to.

There is an associated fc-dev branch that I will merge into the PPW feature branch -- PR to come

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/config/ServicePerimeterServiceConfig.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/config/ServicePerimeterServiceConfig.scala
@@ -1,0 +1,31 @@
+package org.broadinstitute.dsde.rawls.config
+
+import com.typesafe.config.Config
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectNumber, ServicePerimeterName}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+
+final case class ServicePerimeterServiceConfig(staticProjectsInPerimeters: Map[ServicePerimeterName, Seq[GoogleProjectNumber]] = Map.empty,
+                                               pollInterval: FiniteDuration,
+                                               pollTimeout: FiniteDuration)
+
+case object ServicePerimeterServiceConfig {
+  def apply[T <: ServicePerimeterServiceConfig](conf: Config): ServicePerimeterServiceConfig = {
+    val gcsConfig = conf.getConfig("gcs")
+    val staticProjectConfig = gcsConfig.getConfig("servicePerimeters.staticProjects")
+
+    val mappings: Map[ServicePerimeterName, Seq[GoogleProjectNumber]] = staticProjectConfig.entrySet().asScala.map { entry =>
+      val key: ServicePerimeterName = ServicePerimeterName(entry.getKey.replace("\"", ""))
+      val value: Seq[GoogleProjectNumber] = staticProjectConfig.getStringList(entry.getKey).asScala.map(GoogleProjectNumber)
+      key -> value
+    }.toMap
+
+    val pollInterval = Duration.fromNanos(gcsConfig.getDuration("servicePerimeters.pollInterval").toNanos)
+    val pollTimeout = Duration.fromNanos(gcsConfig.getDuration("servicePerimeters.pollTimeout").toNanos)
+
+    new ServicePerimeterServiceConfig(mappings, pollInterval, pollTimeout)
+  }
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
@@ -1,0 +1,84 @@
+package org.broadinstitute.dsde.rawls.serviceperimeter
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.StatusCodes
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.rawls.config.ServicePerimeterServiceConfig
+import org.broadinstitute.dsde.rawls.dataaccess._
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectNumber, ServicePerimeterName, Workspace}
+import org.broadinstitute.dsde.rawls.util.Retry
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class ServicePerimeterService(dataSource: SlickDataSource, gcsDAO: GoogleServicesDAO, config: ServicePerimeterServiceConfig)
+                             (implicit val system: ActorSystem, protected val executionContext: ExecutionContext)
+  extends LazyLogging with Retry {
+
+  /**
+    * In its own transaction, look up all of the Workspaces contained in Billing Projects that use the specified
+    * ServicePerimeterName
+    *
+    * @param servicePerimeterName
+    * @return
+    */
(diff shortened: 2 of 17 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
```diff
@@ -0,0 +1,91 @@
...
+      projectNumberStrings = projectNumbers.map(_.value).toSet
+      operation <- gcsDAO.accessContextManagerDAO.overwriteProjectsInServicePerimeter(servicePerimeterName, projectNumberStrings)
+      result <- retryUntilSuccessOrTimeout(failureLogMessage = s"Google Operation to update Service Perimeter: ${servicePerimeterName} was not successful")(config.pollInterval, config.pollTimeout) { () =>
+        gcsDAO.pollOperation(OperationId(GoogleApiTypes.AccessContextManagerApi, operation.getName)).map {
+          case OperationStatus(false, _) => Future.failed(new RawlsException(s"Google Operation to update Service Perimeter ${servicePerimeterName} is still in progress..."))
+          // TODO: If the operation to update the Service Perimeter failed, we need to consider the possibility that
+          // the list of Projects in the Perimeter may have been wiped or somehow modified in an undesirable way.  If
+          // this happened, it would be possible for Projects intended to be in the Perimeter are NOT in that
+          // Perimeter anymore, which is a problem.
```
gpolumbo-broad, line 71: I know this comment was here before, but any thoughts on what we could do?  I can think of a couple of things:

- When this exception is thrown, email the Cloud Integration team
- Write a Stackdriver monitor or something that watches for this exception and alerts us
- Diff the projects in the perimeter at the start of the method to what is in the perimeter now and log/alert that information

I don't know, these all might be terrible ideas.  Maybe others aren't as worried about it as I am.  My understanding is that if this does get messed up it's very bad.  The condition may be very unlikely and we might not know a situation that can cause it.  

I guess if Doug or anyone else can justify the risk, I would like to put that justification into the comment.  Something that asserts, "Yes, we know there's a potential hole/risk here.  Here's our justification for why we're not doing more to prevent it or sound alarms if it should occur."

Of course, we don't even know all the various things that could go wrong with the Operation that gets us to this line of code, many of which we might not be concerned about.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -644,10 +645,18 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
         case None => gcsDAO.getGoogleProject(billingProject.googleProjectId).map(googleProject => GoogleProjectNumber(googleProject.getProjectNumber.toString))
       }
 
+      // all v2 workspaces in the specified Terra billing project will already have their own Google project number, but any v1 workspaces should store the Terra billing project's Google project number
+      workspaces <- dataSource.inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listWithBillingProject(projectName)
+      }
+      v1Workspaces = workspaces.filterNot(_.googleProjectNumber.isDefined)
```
gpolumbo-broad, line 663: https://github.com/broadinstitute/rawls/blob/v2_workspaces/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20201106_workspace_google_project_columns.xml

I think this scenario is unlikely to occur given the liquibase.  Therefore I'm inclined to remove this conditional, but it's probably not hurting anything either.

gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
marctalbott: I like the idea of a Stackdriver monitor, that seems like a simple and effective way to notify us.

I also wonder if we should lock down access to the projects if the perimeter is messed up? It'd be super disruptive of course, especially for projects that remained in the perimeter, so I suppose it depends on how strongly we need to try and prevent exfiltration

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: If the Terra billing project doesn't have its Google project number stored when that liquibase changeset runs, then I believe we'd still need this conditional. I realize now I also need to update the Terra billing project with the Google project number and service perimeter name

core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.broadinstitute.dsde.rawls.dataaccess._
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectNumber, ServicePerimeterName, Workspace}
+import org.broadinstitute.dsde.rawls.util.Retry
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+object ServicePerimeterService {
+  def constructor(dataSource: SlickDataSource, gcsDAO: GoogleServicesDAO, config: ServicePerimeterServiceConfig)
```
dvoet: this is not used, remove it

core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
dvoet: I don't think an error in the API call means the perimeter is corrupted. Let's start with an alert.

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/config/ServicePerimeterServiceConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.typesafe.config.Config
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectNumber, ServicePerimeterName}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+
+final case class ServicePerimeterServiceConfig(staticProjectsInPerimeters: Map[ServicePerimeterName, Seq[GoogleProjectNumber]] = Map.empty,
+                                               pollInterval: FiniteDuration = 5 seconds,
```
dvoet: I think 5s is too slow for production. How about 1s?

marctalbott: I can change it to 1s in the firecloud-develop PR. I'm going to remove the default specified here because I think it's a little misleading to have this when it's really specified in config

core/src/main/scala/org/broadinstitute/dsde/rawls/serviceperimeter/ServicePerimeterService.scala
marctalbott: wrote a ticket for a stackdriver alert https://broadworkbench.atlassian.net/browse/CA-1210

PR #1367, opened 2021-02-23
Ensure BQ dataset for snapshot references [AS-593]

Shakespeared:
Testing is complete as of now, and this is ready for final review 

Sibling PR for workbench libs is here: https://github.com/broadinstitute/workbench-libs/pull/528


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
       val dataRepoReference = new DataRepoSnapshot().instanceName(terraDataRepoInstanceName).snapshot(snapshot.snapshotId)
       val ref = workspaceManagerDAO.createDataReference(workspaceContext.workspaceIdAsUUID, snapshot.name, ReferenceTypeEnum.DATA_REPO_SNAPSHOT, dataRepoReference, CloningInstructionsEnum.NOTHING, userInfo.accessToken)
 
-      Future.successful(ref)
+      val datasetUuid = UUID.randomUUID().toString.replace('-', '_')
+
+      val defaultIamRoles = Acl.Role.OWNER -> Seq((clientEmail, Acl.Entity.Type.USER))
```
Shakespeared: defaultIamRoles is where we're planning to add the Delta Layer SA

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+
+      val accessPolicies = Seq(SamWorkspacePolicyNames.owner, SamWorkspacePolicyNames.writer, SamWorkspacePolicyNames.reader)
+
+      val datasetLabels = Map("workspace_id" -> workspaceContext.workspaceId, "snapshot_name" -> snapshot.name.value, "snapshot_id" -> snapshot.snapshotId)
+
+      // create BQ dataset, get workspace policies from Sam, and add those Sam policies to the dataset IAM
+      val IOresult = for {
```
Shakespeared: Also considering extracting some of this to make it more readable

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6234,3 +6234,4 @@ components:
             email: email authorization
             profile: profile authorization
             https://www.googleapis.com/auth/cloud-billing: cloud billing authorization
+            https://www.googleapis.com/auth/bigquery.readonly: BigQuery read only authorization
```
davidangb: why is this scope needed? I'll read through the rest of the PR and see if the answer is obvious.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
       val ref = workspaceManagerDAO.createDataReference(workspaceContext.workspaceIdAsUUID, snapshot.name, ReferenceTypeEnum.DATA_REPO_SNAPSHOT, dataRepoReference, CloningInstructionsEnum.NOTHING, userInfo.accessToken)
 
-      Future.successful(ref)
+      val datasetUuid = UUID.randomUUID().toString.replace('-', '_')
+
+      val defaultIamRoles = Acl.Role.OWNER -> Seq((clientEmail, Acl.Entity.Type.USER))
+
+      val accessPolicies = Seq(SamWorkspacePolicyNames.owner, SamWorkspacePolicyNames.writer, SamWorkspacePolicyNames.reader)
+
+      val datasetLabels = Map("workspace_id" -> workspaceContext.workspaceId, "snapshot_name" -> snapshot.name.value, "snapshot_id" -> snapshot.snapshotId)
```
davidangb: reference names are editable. It's fine to add it as a label, and could be helpful - as long as we don't count on it being immutable.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+
+      // create BQ dataset, get workspace policies from Sam, and add those Sam policies to the dataset IAM
+      val IOresult = for {
+        samPolicies <- IO.fromFuture(IO(samDAO.listPoliciesForResource(SamResourceTypeNames.workspace, workspaceContext.workspaceId, userInfo)))
+        projectOwnerPolicy = samPolicies.filter(_.policyName == SamWorkspacePolicyNames.projectOwner).head.policy.memberEmails
+        filteredSamPolicies = samPolicies.filter(samPolicy => accessPolicies.contains(samPolicy.policyName)).map(_.email) ++ projectOwnerPolicy
+        samAclBindings = Acl.Role.READER -> filteredSamPolicies.map{ filteredSamPolicyEmail =>(filteredSamPolicyEmail, Acl.Entity.Type.GROUP) }.toSeq
+        aclBindings = Map(samAclBindings, defaultIamRoles)
+        createdDataset <- bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace)).use(_.createDataset(datasetUuid, datasetLabels))
+        updatedDataset <- bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace)).use(_.setDatasetIam(datasetUuid, aclBindings))
```
davidangb: create the GoogleBigQueryService once,  then use it twice:
```
        bqService = bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace))
        _ <- bqService.use(_.createDataset(datasetUuid, datasetLabels))
        _ <- bqService.use(_.setDatasetIam(datasetUuid, aclBindings))
```
also note that I changed `createdDataset` and `updatedDataset` to `_`; I don't think those are used anywhere, so no need to save them to a variable

davidangb: also, if `createDataset` succeeds but `setDatasetIam` throws an error for whatever reason, does this leave an orphaned BQ dataset in the user's project? Bigger picture - if creating the snapshot reference succeeds, but creating the BQ dataset or reference fails, does the workspace have a snapshot reference without its Delta Layer companion?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
       val dataRepoReference = new DataRepoSnapshot().instanceName(terraDataRepoInstanceName).snapshot(snapshot.snapshotId)
       val ref = workspaceManagerDAO.createDataReference(workspaceContext.workspaceIdAsUUID, snapshot.name, ReferenceTypeEnum.DATA_REPO_SNAPSHOT, dataRepoReference, CloningInstructionsEnum.NOTHING, userInfo.accessToken)
 
-      Future.successful(ref)
+      val datasetUuid = UUID.randomUUID().toString.replace('-', '_')
```
davidangb: nit: call this "datasetName" or similar? I realize we're using a UUID to populate the name, but this is definitely the dataset name.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -40,7 +44,28 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+        filteredSamPolicies = samPolicies.filter(samPolicy => accessPolicies.contains(samPolicy.policyName)).map(_.email) ++ projectOwnerPolicy
+        samAclBindings = Acl.Role.READER -> filteredSamPolicies.map{ filteredSamPolicyEmail =>(filteredSamPolicyEmail, Acl.Entity.Type.GROUP) }.toSeq
+        aclBindings = Map(samAclBindings, defaultIamRoles)
+        createdDataset <- bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace)).use(_.createDataset(datasetUuid, datasetLabels))
+        updatedDataset <- bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace)).use(_.setDatasetIam(datasetUuid, aclBindings))
+      } yield {}
+      IOresult.unsafeToFuture().map { _ =>
+        workspaceManagerDAO.createBigQueryDataset(workspaceContext.workspaceIdAsUUID, new DataReferenceRequestMetadata().name(datasetUuid).cloningInstructions(CloningInstructionsEnum.NOTHING), new GoogleBigQueryDatasetUid().projectId(workspaceContext.namespace).datasetId(datasetUuid), userInfo.accessToken)
+        ref
+      }
```
davidangb: How will future code find the BigQuery reference for a given TDR snapshot reference? WSM may eventually store relationships between resources, but it doesn't yet. Perhaps the BigQuery reference should be named using the TDR snapshot reference's id, something like "deltalayer-${snapshotId}"?

davidangb (commented):
first-pass comments inline!

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: Rawls doesn't actually need this scope, but WSM does downstream for when we create the uncontrolled resource. WSM does a check to make sure that the user can view the dataset. We were seeing failures in WSM with the default scopes, which was confusing because when we tested with a token from `gcloud auth print-access-token`, it worked. When we inspected the scopes, we saw that the token from `gcloud` had some additional scopes, including the Cloud Platform scope. Cloud Platform is obviously way too broad of a scope to ask for, so we tightened it down to just BigQuery.ReadOnly.

An alternative may be to _not_ require this scope, and instead make the call to WSM as the user's pet, and we can ask Sam to generate the pet token with this scope.

I am slightly leaning towards that solution so we can keep the number of scopes that we request from the user to the minimum possible set.

davidangb: hm, I agree: we shouldn't ask the end user for additional scopes just to add the uncontrolled resource. I would rather solve this at the WSM layer by using a pet or some other solution.

Shakespeared:
@davidangb this is ready for re-review now!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SamDAO.scala
```diff
@@ -76,4 +78,9 @@ object SamDAO {
   case object NotFound extends GetUserIdInfoResult
   case object NotUser extends GetUserIdInfoResult
   final case class User(userIdInfo: UserIdInfo) extends GetUserIdInfoResult
-}
\ No newline at end of file
+  val defaultScopes = Set(
+    "https://www.googleapis.com/auth/userinfo.email",
+    "https://www.googleapis.com/auth/userinfo.profile"
+  )
+  val bigQueryScope = "https://www.googleapis.com/auth/bigquery.readonly"
```
MatthewBemis: nit: let's change this to `bigQueryReadOnlyScope`

MatthewBemis:
jenkins retest

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: we've replaced this with a pet token that has the appropriate scope

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/SamDAO.scala
```diff
@@ -76,4 +78,9 @@ object SamDAO {
   case object NotFound extends GetUserIdInfoResult
   case object NotUser extends GetUserIdInfoResult
   final case class User(userIdInfo: UserIdInfo) extends GetUserIdInfoResult
-}
\ No newline at end of file
+  val defaultScopes = Set(
+    "https://www.googleapis.com/auth/userinfo.email",
+    "https://www.googleapis.com/auth/userinfo.profile"
+  )
+  val bigQueryReadOnlyScope = "https://www.googleapis.com/auth/bigquery.readonly"
```
davidangb:
```
  com.google.api.services.oauth2.Oauth2Scopes.USERINFO_EMAIL
  com.google.api.services.oauth2.Oauth2Scopes.USERINFO_PROFILE
  com.google.api.services.bigquery.BigqueryScopes.BIGQUERY_READONLY
```
should give you Google's definitions for these scopes, instead of hardcoding them

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -41,7 +46,29 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+
+      val datasetLabels = Map("workspace_id" -> workspaceContext.workspaceId, "snapshot_id" -> snapshot.snapshotId)
+
+      // create BQ dataset, get workspace policies from Sam, and add those Sam policies to the dataset IAM
+      val IOresult = for {
+        samPolicies <- IO.fromFuture(IO(samDAO.listPoliciesForResource(SamResourceTypeNames.workspace, workspaceContext.workspaceId, userInfo)))
+        projectOwnerPolicy = samPolicies.filter(_.policyName == SamWorkspacePolicyNames.projectOwner).head.policy.memberEmails
+        filteredSamPolicies = samPolicies.filter(samPolicy => accessPolicies.contains(samPolicy.policyName)).map(_.email) ++ projectOwnerPolicy
+        samAclBindings = Acl.Role.READER -> filteredSamPolicies.map{ filteredSamPolicyEmail =>(filteredSamPolicyEmail, Acl.Entity.Type.GROUP) }.toSeq
+        aclBindings = defaultIamRoles + samAclBindings
```
davidangb: assuming I'm reading this correctly ... end users will only have read access to these datasets, is that correct? I wonder if we'll want to give them any greater access at any point; certainly not for this PR.

davidangb (commented):
minor comment/question inline.

I know you're still thinking about tests, so I haven't really reviewed test coverage.

I'm still worried about orphaned datasets in SnapshotService. If we create the dataset in BQ, but then get an error on the call to WSM, the dataset will remain in the user's project but we won't have any reference to it. I think that area of code needs more exception/recover code. One option is to adopt Stairway and its rollback of saga transactions, but at minimum this needs some custom rollback code.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockBigQueryServiceFactory.scala
```diff
@@ -110,4 +116,10 @@ class MockGoogleBigQueryService(queryResponse: Either[Throwable, TableResult]) e
       case Right(results) => IO.pure(results)
     }
   }
+
+  override def createDataset(datasetName: String, labels: Map[String, String], aclBindings: Map[Acl.Role, Seq[(WorkbenchEmail, Entity.Type)]]): IO[DatasetId] = {
+    val dataset = DatasetInfo.newBuilder(datasetName).build().getDatasetId
+    IO.pure(dataset)
```
ansingh7115: tiny: Can this be one line? 
`IO.pure(DatasetInfo.newBuilder(datasetName).build().getDatasetId)`
Not a big deal tho - might be clearer as two

ansingh7115:
Looks like Jenkins build step is failing on sbt assembly - I wonder if this is because the google2 version in rawls hasn't been updated in three months, and in the intervening time there have been some updates to akka dependencies in workbench-libs

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
Shakespeared: That is correct, and I think that's what we want (as of now at least)

MatthewBemis:
jenkins retest due to transient error

MatthewBemis:
a completely different yet still unrelated test failed. jenkins retest

automation/project/Dependencies.scala
```diff
@@ -61,6 +64,7 @@ object Dependencies {
     workbenchServiceTest,
 
     dataRepo,
+    workspaceManager,
```
davidangb: this is funky. I wouldn't expect this file to show diffs at all. Have you rebased?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -65,4 +68,9 @@ class HttpWorkspaceManagerDAO(baseWorkspaceManagerUrl: String)(implicit val syst
     getWorkspaceApi(accessToken).enumerateReferences(workspaceId, offset, limit)
   }
 
+  override def createBigQueryDataset(workspaceId: UUID, metadata: DataReferenceRequestMetadata, dataset: GoogleBigQueryDatasetUid, accessToken: OAuth2BearerToken): BigQueryDatasetReference = {
+    val createBigQueryDatasetReference = new CreateBigQueryDatasetReferenceRequestBody().dataset(dataset).metadata(metadata)
+    getReferencedGcpResourceApi(accessToken).createBigQueryDatasetReference(createBigQueryDatasetReference, workspaceId)
+  }
```
davidangb, line 74: nit: rename this method to `createBigQueryDatasetReference`? Want to be clear it's a reference, not a controlled resource.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+        fakeRawlsClientEmail,
+        fakeDeltaLayerStreamerEmail
+      )(userInfo)
+
+      try {
+        Await.result(snapshotService.createSnapshot(workspace.toWorkspaceName, NamedDataRepoSnapshot(DataReferenceName("foo"), Option(DataReferenceDescriptionField("foo")), "bar")), Duration.Inf)
+      }
+      catch {
+        case e: RuntimeException => "WSM snapshot info not returned"
+      }
```
davidangb: instead of try/catch, assert that this actually throws an exception, so you know it's working as expected. Something like:
```
      val createException = intercept[RuntimeException] {
        Await.result(snapshotService.createSnapshot(workspace.toWorkspaceName, NamedDataRepoSnapshot(DataReferenceName("foo"), Option(DataReferenceDescriptionField("foo")), "bar")), Duration.Inf)
      }
      createException.getMessage shouldBe "my expected message"
```

Also ... could you update the mock as well as this code to use a checked exception, not `RuntimeException`? I believe this throws `bio.terra.workspace.client.ApiException` in real usage.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -41,7 +46,29 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+        bqService = bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace))
+        _ <- bqService.use(_.createDataset(datasetName, datasetLabels, aclBindings))
+        petToken <- IO.fromFuture(IO(samDAO.getPetServiceAccountToken(GoogleProjectId(workspaceName.namespace), SamDAO.defaultScopes + SamDAO.bigQueryReadOnlyScope, userInfo)))
+      } yield { petToken }
+      IOResult.unsafeToFuture().map { petToken =>
+        workspaceManagerDAO.createBigQueryDataset(workspaceContext.workspaceIdAsUUID, new DataReferenceRequestMetadata().name(datasetName).cloningInstructions(CloningInstructionsEnum.NOTHING), new GoogleBigQueryDatasetUid().projectId(workspaceContext.namespace).datasetId(datasetName), OAuth2BearerToken(petToken))
+        ref
+      }
```
davidangb: This method is still vulnerable to orphaning resources. As I read it, this method:
1. Creates the snapshot reference
2. Creates BQ dataset
3. Creates dataset reference

(with other details elided, like getting policies from Sam and getting a pet token)

If any step fails, the previous steps will be orphaned. That is:
* if creating the BQ dataset fails, the user will have a snapshot reference without a Delta Layer companion
* if creating the dataset reference fails, the user will have an orphaned dataset in their workspace with no reference to it

The two paths I can see to solving this are to adopt Stairway, or to wrap this entire method in a try/catch or Future.recover, and in the failure case attempt to delete all three resources (snapshot reference, BQ dataset, dataset reference), ignoring any 404s from the delete attempts.

This cleanup code may drive changes to the accompanying tests, of course.

davidangb (commented):
More comments inline, this is getting very close - I think it's really just handling the saga transaction that is left!

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
Shakespeared: mockito is really fussy about checked exceptions, I wasn't able to get it working with one at all (that's the only reason I used the ApiException). I'd much prefer to use a checked exception here, but I would need someone who knows more about mockito to pair on that with me because I tried and couldn't figure it out

MatthewBemis:
jenkins retest - rawls build failed due to infra issue

MatthewBemis:
jenkins retest due to some infra issue

MatthewBemis:
fiab tests are a PITA

jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -41,7 +48,38 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+        aclBindings = calculateDatasetAcl(samPolicies)
+        bqService = bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(workspaceName.namespace))
+        _ <- bqService.use(_.createDataset(datasetName, datasetLabels, aclBindings))
+        petToken <- IO.fromFuture(IO(samDAO.getPetServiceAccountToken(GoogleProjectId(workspaceName.namespace), SamDAO.defaultScopes + SamDAO.bigQueryReadOnlyScope, userInfo)))
+      } yield { petToken }
+
+      IOResult.unsafeToFuture().recover {
+        case t: Throwable =>
+          //fire and forget this undo, we've made our best effort to fix things at this point
+          IO.pure(workspaceManagerDAO.deleteDataReference(workspaceContext.workspaceIdAsUUID, ref.getReferenceId, userInfo.accessToken)).unsafeToFuture()
```
davidangb: this should be `deleteBigQueryDataset`, shouldn't it? Roll back creation of the BQ dataset, not of its reference?

Also I wonder if it is worth adding error handling into these rollback steps and logging any problems with the rollback, so we are aware of any problem workspaces. It should still be fire-and-forget - the *user* shouldn't have to wait for the rollback to continue - but having logging might be useful.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+        Await.result(snapshotService.createSnapshot(workspace.toWorkspaceName, NamedDataRepoSnapshot(DataReferenceName("foo"), Option(DataReferenceDescriptionField("foo")), "bar")), Duration.Inf)
+      }
+      createException.getMessage shouldBe "oh no!"
+
+      verify(mockSamDAO, times(0)).listPoliciesForResource(any[SamResourceTypeName], any[String], any[UserInfo])
+      verify(mockBigQueryServiceFactory, times(0)).getServiceFromCredentialPath(fakeCredentialPath, GoogleProject(workspace.namespace))
+      verify(mockWorkspaceManagerDAO, times(1)).createDataReference(any[UUID], any[DataReferenceName], any[Option[DataReferenceDescriptionField]], any[ReferenceTypeEnum], any[DataRepoSnapshot], any[CloningInstructionsEnum], any[OAuth2BearerToken])
+      verify(mockWorkspaceManagerDAO, times(0)).createBigQueryDatasetReference(any[UUID], any[DataReferenceRequestMetadata], any[GoogleBigQueryDatasetUid], any[OAuth2BearerToken])
+    }
+
```
davidangb, line 113: could you add one more test here, for `should "not leave an orphaned bq dataset if getting the pet service account token fails"`?

davidangb (commented):
see inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
MatthewBemis: We still need the `deleteDataReference` to delete the reference to the snapshot, but you're still right to call this out. We didn't discuss the possibility of `createDataset` succeeding but then `getPetServiceAccountToken` failing.

I think we can leave this `recover` as-is, but move `getPetServiceAccountToken` down to a more appropriate place. I don't think there's a god reason to have it in `IOResult` and then emit the token. I think that's just an artifact of earlier development.

ansingh7115: I think we might want to do both. Delete the big query dataset and the reference. Unless we're okay with some references existing without the dataset?

MatthewBemis: I am operating under the assumption that if Google returns an error code on `createDataset`, then they won't have created the dataset. So doing a `deleteDataset` as part of this `recover` is unnecessary IMO.

To be clear on the changes that I think should happen:

The only change I think we should make to this step of the recover logic is to move `getPetServiceAccountToken` out of `IOResult` just so there's no ambiguity on why `IOResult` actually failed.

davidangb: I retract my original comment, I misread - I got confused between the *snapshot* reference and the *BigQuery* reference (maybe add some code comments)? That said I agree that moving the pet SA token out of the first IOResult could be good.

MatthewBemis: Latest commit moves the pet token SA stuff down and adds a test for it.

We discussed adding logging but we really didn't like number of nested recovers that popped up as a result. It made the code unwieldy.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
MatthewBemis: done

davidangb (approved):
:+1: this was a big complicated one ... nice job!

Last thought ... when should this merge? Do you feel good about merging it anytime, or should we hold it until after the first TDR/WSM integration goes to prod, just in case of any issues?

ansingh7115 (approved):
Awesome :)

PR #1370, opened 2021-02-25
AS-649: Remove vestigial actor pattern from service classes

MatthewBemis:
(ignoring whitespace may make the diffs clearer, particularly for the cases where permission checks were pushed down into business logic)

Nearly every API call in Rawls has an extra step in it's code path: the API layer calls a top-level method in the service layer, which then directly calls another method in that same service layer which contains the actual implementation. There's no reason that this extra step of calling a top-level method needs to happen anymore. It's an artifact from our days of using actors: each of these redundant top-level methods used to be an akka actor message.

The reason that this pattern still exists is because it made the git diff much more reasonable to understand back when we were migrating rawls to akka-http and Scala 2.12. But I've now found that it's confusing to developers who are unfamiliar with this history, and there's really no reason to keep it around.

There were also some cases where permission checks were done in that top-level method, rather than in the actual implementation. With the removal of the top-level methods, I've pushed these permission checks down into their implementations to be more consistent with the majority of the other code.

PR to do the same thing in firecloud-orchestration: https://github.com/broadinstitute/firecloud-orchestration/pull/836

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


dvoet:
I am not going to take a stand on this but I think the permissions checks are generally better off separate from the business logic. That way the business logic can be reused more easily. We have had great success in Sam keeping all the permissions checks at the API/Route layer. That might be a little cumbersome in Rawls because of more complex requirements or database lookups. The other thing I look for in permission checks is that happen at a consistent place in the call stack. That's why there were mostly at the message handling layer.

🥇 on this effort.

The next thing you can do is get rid of PerRequest ;)

dvoet (approved):

davidangb (approved):
lgtm!!! I suggest at least showing this to the batch workflows team, whether or not they want to review, so they're aware of the structural changes.

PR #1372, opened 2021-03-01
update to jdk 11 [AS-594]

zarsky-broad:
Paired with https://github.com/broadinstitute/firecloud-develop/pull/2454...


zarsky-broad:
Successful automation run (with paired firecloud-develop PR): https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/46321/

project/Merging.scala
```diff
@@ -9,6 +9,7 @@ object Merging {
     case PathList("com", "typesafe", _ @ _*) => MergeStrategy.last
     case PathList("com", "google", "auto", "value", _ @ _*) => MergeStrategy.last
     case PathList("io", "sundr", _ @ _*) => MergeStrategy.last
+    case PathList("javax", "activation", _ @ _*) => MergeStrategy.first
```
zarsky-broad: This comes in through clients built with swagger codegen. I'm not positive that this is the most correct merge strategy, but I don't _think_ it matters hugely.

project/Testing.scala
```diff
@@ -22,6 +23,9 @@ object Testing {
 
   val commonTestSettings: Seq[Setting[_]] = List(
 
+    testOptions in Test += Tests.Setup(() =>
+      sys.props += "mockserver.logLevel" -> "WARN"
```
zarsky-broad, line 27: The default log verbosity for mockserver logs every mock defined, for every test that defines it. Since this goes to the console for jenkins jobs, it makes the logs _huge_, to the point that the browser has trouble searching through them.

project/Testing.scala
```diff
@@ -55,7 +59,7 @@ object Testing {
     (testOnly in Test) := ((testOnly in Test) dependsOn validMySqlHost).evaluated,
 
     parallelExecution in Test := false
-  ) ++ MinnieKenny.testSettings
+  ) ++ (if (sys.props.getOrElse("secrets.skip", false) != "true") MinnieKenny.testSettings else List())
```
zarsky-broad: I chose this option name somewhat arbitrarily, if it doesn't make sense it can be changed. Alternatively, we could remove Minnie-Kenny being run by sbt; Orch doesn't have it, and it's run manually in the github CI.

automation/Dockerfile-tests
```diff
@@ -1,4 +1,4 @@
-FROM bigtruedata/sbt
+FROM broadinstitute/scala-baseimage:jdk11-2.12.12-1.4.7
```
kshakir, line 1: Assuming container-based compilation is absolutely required, someday I hope this custom image gets pushed to a public GCR or anywhere else besides docker.io. We have so many tests in the terra multi-repo that are failing due to [pull rate limits](https://www.docker.com/increase-rate-limits) from docker.io these days. 😢 

Or for easier patching of individual components, just use a more flexible base container like focal and then download java and the [0.016GB sbt package](https://github.com/sbt/sbt/releases). While Java gets patches much slower, Scala and SBT get patches once a month. For example: sbt 1.4.8 is already out.

project/Merging.scala
kshakir: Interesting. These generated client libraries are including the classes, not referring to them as a dependency? I'm surprised OpenAPI doesn't have support for transitive dependencies. Is there a link to an issue that can be included as a comment?

project/Testing.scala
kshakir: To confirm: Orch doesn't check for the secrets leaks commited in `minniekenny.gitconfig`? Is that a future ticket that can be linked in a comment?

kshakir (approved):
A bunch of ramblings based on scar tissue from other repos but LGTM overall

automation/Dockerfile-tests
zarsky-broad: So:
1. Eventually, there's going to be Broad GCR coordinates for this image, @choover-broad is working on something for that...
2. This is basically just git (which I debated about) and that little sbt package on top of the base adoptopenjdk image, all the existing images were running mystery meat openjdk's, the idea was to get us on the one we're officially using.
3. SBT will automatically download the version of itself that you specify, so it matters a lot less what version of sbt (and scala for that matter, although it's bigger) is baked into the image.

project/Testing.scala
zarsky-broad: Orch doesn't have any minnie-kenny anything of any kind; I'm not aware of any discussion around it, @davidangb do you know of anything?

core/src/test/scala/org/broadinstitute/dsde/rawls/RawlsTestUtils.scala
```diff
@@ -71,8 +71,8 @@ trait RawlsTestUtils extends Suite with TestDriverComponent with Matchers {
   }
 
   // MockServer's .withBody doesn't have a built-in string contains feature.  This serves that purpose.
-  def mockServerContains(text: String): StringBody = {
-    val anythingWithNewlines = "((.|\n|\r)*)"
-    StringBody.regex(anythingWithNewlines + Regex.quote(text.toString) + anythingWithNewlines)
+  def mockServerContains(text: String): RegexBody = {
+    // "(?s)" turns on DOTALL mode, where a "." matches a line break as well as any character
+    new RegexBody("(?s).*" + Regex.quote(text) + ".*")
```
davidangb, line 76: why did this have to change? Was there a java 8 -> compatibility issue here?

project/Dependencies.scala
```diff
@@ -87,7 +87,7 @@ object Dependencies {
   val scalaUri: ModuleID =        "io.lemonlabs"                  %% "scala-uri"            % "3.0.0"
   val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.2.2" % "test"
   val mockito: ModuleID =         "org.scalatestplus"             %% "mockito-3-4"          % "3.2.2.0" % Test
-  val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "3.9.2" % "test"
+  val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "5.11.2" % "test"
```
davidangb, line 90: why did this have to change?

project/Merging.scala
zarsky-broad: There's a series of issues referenced [here](https://github.com/eclipse-ee4j/jaf/issues/18) that lead to [Jackson](https://github.com/FasterXML/jackson-modules-base/issues/112) and then to [swagger-core](https://github.com/swagger-api/swagger-core/issues/3836), tbh I don't understand this all well enough to know what would be a useful comment.

zarsky-broad: There's a resolution to that last one, which will require updates to the version of swagger used to generate the clients, but I'm not exactly sure I understand what will need to happen.

davidangb: I much prefer using exclude rules instead of the blunt-intstrument merge strategy. Do you know where the conflicts are, and could you add excludes for those instead?

fwiw, we recently also had assembly issues in orch on javax.activation, though that was via javax.mail: broadinstitute/firecloud-orchestration#838

project/Settings.scala
```diff
@@ -78,13 +82,14 @@ object Settings {
...
     libraryDependencies ++= googleDependencies
-  ) ++ versionSettings ++ noPublishSettings
+  ) ++ versionSettings ++ noPublishSettings ++ java11BuildSettings
 
   //the full list of settings for the rawlsModel project (see build.sbt)
   //coreDefaultSettings (inside commonSettings) sets the project name, which we want to override, so ordering is important.
   //thus commonSettings needs to be added first.
   val modelSettings = cross212and213 ++ commonSettings ++ List(
     name := "rawls-model",
+    javacOptions ++= Seq("--release", "8"), // has to publish a java 8 artifact
```
davidangb, line 92: did you figure out who is consuming this? Does it actually need to be java 8?

davidangb (commented):
some questions/comments inline

core/src/test/scala/org/broadinstitute/dsde/rawls/RawlsTestUtils.scala
zarsky-broad: The new version of mockserver doesn't support `Stringbody.regex`, it has `RegexBody` instead.

project/Dependencies.scala
zarsky-broad: The old version was unhappy with Java 11, upgrading was the easiest way to deal with that, as far as @MatthewBemis and I could see.

project/Settings.scala
zarsky-broad: There are [a few consumers](https://github.com/search?p=1&q=%22rawls-model%22&type=Code), but the main potential problem would be workbench-libs, if I understand correctly

project/Merging.scala
zarsky-broad: I didn't experiment with that, I think I'd need some help to try it.

project/Testing.scala
ansingh7115: Awesome!!

project/Testing.scala
ansingh7115: I don't see any tickets in the backlog. It is used by most other services (like rawls and sam) so maybe we should add it

project/Merging.scala
ansingh7115: +1 for using exclusion rules

project/Dependencies.scala
```diff
@@ -108,8 +108,10 @@ object Dependencies {
 
   val accessContextManager = "com.google.apis" % "google-api-services-accesscontextmanager" % "v1beta-rev55-1.25.0"
 
-  val workspaceManager = excludeGuavaJDK5("bio.terra" % "workspace-manager-client" % "0.13.0-SNAPSHOT")
-  val dataRepo = excludeGuavaJDK5("bio.terra" % "datarepo-client" % "1.0.44-SNAPSHOT")
+  def swaggerClientExcludes(m: ModuleID): ModuleID = m.exclude("jakarta.activation", "jakarta.activation-api")
+
+  val workspaceManager = swaggerClientExcludes("bio.terra" % "workspace-manager-client" % "0.13.0-SNAPSHOT")
+  val dataRepo = swaggerClientExcludes("bio.terra" % "datarepo-client" % "1.0.44-SNAPSHOT")
```
zarsky-broad: @kshakir @davidangb @asingh7115 Preferable?

project/Testing.scala
davidangb: for my clarification: this discussion of orch is tangential to this PR, right?

yeah, if orch does not have secrets-detection, it absolutely needs it. I'd prefer to check with appsec first on the progress of their centralized/cross-repo secrets detection before committing to any work, if appsec is already handling it let's just let them.

automation/Dockerfile-tests
davidangb: my take is that even though sbt *will* automatically download whatever it needs, the entire point of adding sbt to a baseimage is to preload as many dependencies as possible at docker-build time, and avoid re-downloading those dependencies on every build of every service that uses this image. So, counting on sbt to download what it needs defeats the purpose - if we're relying on that, I could see omitting it entirely and downloading it at runtime, avoiding the maintenance of the baseimage.

build.sbt
```diff
@@ -50,6 +50,7 @@ lazy val rawls = project.in(file("."))
 // This appears to do some magic to configure itself. It consistently fails in some environments
 // unless it is loaded after the settings definitions above.
 Revolver.settings
+Global / excludeLintKeys += debugSettings // To avoid lint warning
```
davidangb, line 53: what is the lint warning? Can we fix the source code instead of skipping the linting?

project/Dependencies.scala
davidangb: much better, glad this is working via excludes!

nit, optional: rename `swaggerClientExcludes` to e.g. `excludeJakartaActivation`. If it's only excluding one dependency, make it obvious in the name, similar to `excludeGuavaJDK5`. If it's a bundle of excludes, name it after who uses it, similar to `workbenchGoogleExcludes`.

lastly: do these no longer need to exclude guava-jdk5?

project/Testing.scala
kshakir: I think it's tangential. The question came from the switch to skip testing for git-secrets via the line above that says `sys.props.getOrElse("secrets.skip"`. From the comments, it sounded like orch was the repo/app that would add use this toggle instead of adding secrets tests?

build.sbt
zarsky-broad: The warning's been there since forever, it's from those Revolver settings above, so I don't necessarily want to get into it, since they're apparently magic

project/Testing.scala
davidangb: @zarsky-broad please confirm, but my understanding is this toggle is specific to Rawls, and is called from the change in this PR in docker/install.sh.

build.sbt
davidangb: > I don't necessarily want to get into it

ok, I pulled branches and ran it myself, the lint warning is:
```
[warn] there's a key that's not used by any other settings/tasks:
[warn]  
[warn] * Global / debugSettings
[warn]   +- /Users/davidan/work/src/rawls/build.sbt:52
[warn]  
[warn] note: a setting might still be used by a command; to exclude a key from this `lintUnused` check
[warn] either append it to `Global / excludeLintKeys` or call .withRank(KeyRanks.Invisible) on the key
```
this is present (and fixed by the exclude) on both this PR branch with java 11 and on the develop branch with java 1.8, so it isn't related to the java 11 upgrade as far as I can tell.

davidangb (commented):
A few more comments inline. Where I think things stand on this:
* read optional comment at https://github.com/broadinstitute/rawls/pull/1372#discussion_r591633123, choose if you want to address it
* get swatomation tests to passing

thanks for sticking with it!

project/Dependencies.scala
zarsky-broad: They don't actually include guava-jdk5, that's not included in anything vaguely recent. ~I went with _why_ they have the exclusion over _what_ is being excluded, but I could see inverting it...~ Edit: just re-read your comment, totally misread those sentences...

zarsky-broad:
Successful swatomation with firecloud-develop built from accompanying PR: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/46375/

build.sbt
zarsky-broad: Ok, so should I roll this line back? I was just trying to get clean console output.

project/Testing.scala
zarsky-broad: Yeah. The old build image for Rawls didn't include git at all, so the secrets check was skipped silently. The new image has git, so the secrets check was failing here because it didn't have git-secrets. I added an optional flag to skip it that's only used in the `install.sh`.

The alternative is to not have sbt run the check at all, since it's explicitly run in the Github CI workflow.

zarsky-broad: Ah, I see the problem:

> Orch doesn't have it, and it's run manually in the github CI.

Orch has no minnie-kenny at all. Separately, the rawls github workflow runs minnie-kenny.

project/Dependencies.scala
zarsky-broad: renamed!

project/Testing.scala
gpolumbo-broad: I think there's a scenario here where you're comparing a Boolean to a String. My quick test in the repl shows that this can cause problems. Recommend coercing everything the same type.

gpolumbo-broad (changes requested):

ansingh7115 (approved):
Thanks for switching to using exclusion rules!

gpolumbo-broad (approved):

davidangb (approved):

PR #1388, opened 2021-03-18
update automation tests to newer workspace-manager-client [QA-1415]

davidangb:
QA-1415

Tests were failing because the workspace-manager-client in use was old and did not include recent changes to model classes. The tests hit problems attempting to deserialize json responses from WSM into the old model classes.

This PR updates to use the most recent version of the client.

Note that this is a test-only problem; it does not indicate a problem with the app itself. The app had already updated the client lib.

Tests pass for me when run locally with this branch.

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -25,7 +25,8 @@ object Dependencies {
   val workbenchGoogle2: ModuleID = "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V
   val workbenchServiceTest: ModuleID = "org.broadinstitute.dsde.workbench" %% "workbench-service-test" % serviceTestV % "test" classifier "tests" excludeAll(workbenchExclusions:_*)
 
-  val dataRepo: ModuleID = "bio.terra" % "datarepo-client" % "1.0.44-SNAPSHOT"
+  val dataRepo: ModuleID         = "bio.terra" % "datarepo-client" % "1.0.44-SNAPSHOT"
+  val workspaceManager: ModuleID = "bio.terra" % "workspace-manager-client" % "0.16.0-SNAPSHOT"
 
   val rootDependencies = Seq(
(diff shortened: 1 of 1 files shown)


automation/project/Dependencies.scala
```diff
@@ -52,7 +53,9 @@ object Dependencies {
     "org.seleniumhq.selenium" % "selenium-java" % "3.8.1" % "test",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
     "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-14278f08f"
-      exclude("com.typesafe.scala-logging", "scala-logging_2.11") exclude("com.typesafe.akka", "akka-stream_2.11"),
+      exclude("com.typesafe.scala-logging", "scala-logging_2.11")
+      exclude("com.typesafe.akka", "akka-stream_2.11")
+      exclude("bio.terra", "workspace-manager-client"),
```
davidangb, line 58: workspace-manager-client was included as a transitive dependency from rawls-model. I've excluded it from rawls-model and included it explicitly above, so we have more visibility into and control over which version we are using.

davidangb:
one failure in orch swatomation:
```
org.broadinstitute.dsde.test.api.orch.PFBImportSpec.Orchestration should synchronously return an error when attempting to import a PFB file via import service for readers of a workspace
```

that failure is totally unrelated to this PR

gdlex4015 (approved):

ungwudik (approved):

ungwudik:
> one failure in orch swatomation:
> 
> ```
> org.broadinstitute.dsde.test.api.orch.PFBImportSpec.Orchestration should synchronously return an error when attempting to import a PFB file via import service for readers of a workspace
> ```
> 
> that failure is totally unrelated to this PR

FYI - We are tracking this issue here in https://broadworkbench.atlassian.net/browse/QA-1230

PR #1391, opened 2021-03-19
querying snapshot tables with zero rows should return success [AS-698]

davidangb:
It is possible for a Data Repo snapshot table to contain zero rows.

Previously, if you requested the first page of results from such a table, we would throw an error ("requested page 1 is greater than the number of pages 0").

After this PR, requesting page 1 of a zero-row table returns success, with 0 results. Requesting page 2 or greater of such a table will still throw an error.

This is only relevant to snapshots. It is not possible for Rawls-based entities to have a table with zero rows.

This is a regression from #1386.

Tested by running local UI against local rawls.

REVIEWER: I suggest ignoring whitespace in the diff; I moved around lines of code in `DataRepoEntityProvider.scala` but the functional change is fairly small.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
@@ -15,7 +15,7 @@ import org.broadinstitute.dsde.rawls.entities.datarepo.DataRepoBigQuerySupport._
 import org.broadinstitute.dsde.rawls.entities.exceptions.{DataEntityException, UnsupportedEntityOperationException}
 import org.broadinstitute.dsde.rawls.expressions.parser.antlr.{AntlrTerraExpressionParser, DataRepoEvaluateToAttributeVisitor, LookupExpressionExtractionVisitor, ParsedEntityLookupExpression}
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.GatherInputsResult
-import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, AttributeEntityReference, AttributeNull, AttributeNumber, AttributeString, AttributeValue, AttributeValueList, AttributeValueRawJson, Entity, EntityQuery, EntityQueryResponse, EntityTypeMetadata, ErrorReport, GoogleProjectId, SubmissionValidationEntityInputs}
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -131,24 +131,31 @@ class DataRepoEntityProvider(snapshotModel: SnapshotModel, requestArguments: Ent
...
+        queryResults <- runBigQuery(queryConfigBuilder, petKey, GoogleProject(googleProject.value))
+      } yield {
+        // translate the BQ results into a Rawls query result
+        queryResultsToEntities(queryResults, entityType, pk)
+      }
+      resultIO.unsafeToFuture()
+    }
 
-    val queryConfigBuilder = queryConfigForQueryEntities(dataProject, viewName, entityType, finalQuery)
+    futurePage map { page => EntityQueryResponse(finalQuery, metadata, page) }
```
zarsky-broad, line 158: would we want to get rid of this blank line?

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderQueryEntitiesSpec.scala
```diff
@@ -1,11 +1,11 @@
 package org.broadinstitute.dsde.rawls.entities.datarepo
 
-import bio.terra.datarepo.model.TableModel
+import bio.terra.datarepo.model.{ColumnModel, TableModel}
 import com.google.cloud.PageImpl
 import com.google.cloud.bigquery._
 import org.broadinstitute.dsde.rawls.TestExecutionContext
 import org.broadinstitute.dsde.rawls.dataaccess.MockBigQueryServiceFactory
-import org.broadinstitute.dsde.rawls.dataaccess.MockBigQueryServiceFactory.{createTestTableResult, createKeyList}
+import org.broadinstitute.dsde.rawls.dataaccess.MockBigQueryServiceFactory.{createKeyList, createTestTableResult}
```
zarsky-broad, line 8: I'd love to get some linting going so this can't happen...

zarsky-broad (approved):
Seems like a reasonable solution to the problem that won't break backward compatibility; I would definitely want someone who's more comfortable with our Scala testing norms (and Futures, I know very little about how they're used) to review as well. Logic checks out, though!

ansingh7115 (approved):
looks good, could find nothing wrong

PR #1393, opened 2021-03-20
upgrade datarepo client lib [AS-628, AS-692]

davidangb:
* Upgrades the data repo client lib to 1.41.0, which is the version currently in prod
* Includes the `jersey-hk2` library as a dependency, which the newer data repo client requires
* Adds a unit test to exercise the data repo client connectivity/runtime success

Tested by running Rawls locally and manually hitting the entity-type-metadata API and the entity-query API, both of which connect to Data Repo. And of course the new unit test.

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.rawls.entities.datarepo
 
+import akka.http.scaladsl.model.StatusCodes
 import bio.terra.datarepo.model.{ColumnModel, RelationshipModel, RelationshipTermModel, TableModel}
 import com.google.cloud.PageImpl
 import com.google.cloud.bigquery._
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -111,7 +111,8 @@ object Dependencies {
   def excludeJakartaActivationApi(m: ModuleID): ModuleID = m.exclude("jakarta.activation", "jakarta.activation-api")
 
   val workspaceManager = excludeJakartaActivationApi("bio.terra" % "workspace-manager-client" % "0.16.0-SNAPSHOT")
-  val dataRepo = excludeJakartaActivationApi("bio.terra" % "datarepo-client" % "1.0.44-SNAPSHOT")
+  val dataRepo = excludeJakartaActivationApi("bio.terra" % "datarepo-client" % "1.41.0-SNAPSHOT")
+  val dataRepoJersey = "org.glassfish.jersey.inject" % "jersey-hk2" % "2.32"
```
davidangb, line 115: without the additional `jersey.inject` library, the upgraded data repo client encountered `Unexpected error: InjectionManagerFactory not found.` errors at runtime.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderSpec.scala
```diff
@@ -793,6 +799,35 @@ class DataRepoEntityProviderSpec extends AsyncFlatSpec with DataRepoEntityProvid
...
+        .withStatusCode(StatusCodes.OK.intValue)
+    )
+
+    val dataRepoDAO = new HttpDataRepoDAO("mock", s"http://localhost:$mockPort")
+    val snapshotResponse = dataRepoDAO.getSnapshot(mockSnapshotId, userInfo.accessToken)
+    mockServer.stopAsync()
+
+    snapshotResponse.getId shouldBe mockSnapshotId.toString
+  }
+
```
davidangb, line 830: This test fails when the `jersey.inject` library is not included in dependencies (see below), but passes when it is included. We really need this test, because the other tests we have that connect to data repo only run nightly in alpha - without this test, PRs passed as green even though we had runtime failures.

project/Dependencies.scala
MatthewBemis: is there a reason that the data repo client lib does not include all of its own dependencies?

MatthewBemis (approved):
LGTM, just curious about one thing

project/Dependencies.scala
davidangb: Good question! I raised the issue to them and they weren't aware of it.

PR #1394, opened 2021-03-23
handle nulls when serializing StackTraceElement [AS-696]

davidangb:
I still don't fully understand why, but when running on the JRE, stacktraces contain null fileNames, while when running on the JDK those fileNames are populated. Compare the partial stacktrace examples below.

This nulls inside stacktraces caused problems when serializing an `ErrorReport` to the response. Our `RootJsonFormat[StackTraceElement] ` was not null-safe when writing; now it is.

partial stacktrace on JRE:
```
    {
      "className": "scala.concurrent.impl.CallbackRunnable",
      "fileName": "Promise.scala",
      "lineNumber": 64,
      "methodName": "run"
    },
    {
      "className": "java.util.concurrent.ForkJoinTask$RunnableExecuteAction",
      "fileName": null,
      "lineNumber": -1,
      "methodName": "exec"
    },
```

same stacktrace on JDK:
```
    {
      "className": "scala.concurrent.impl.CallbackRunnable",
      "fileName": "Promise.scala",
      "lineNumber": 64,
      "methodName": "run"
    },
    {
      "className": "java.util.concurrent.ForkJoinTask$RunnableExecuteAction",
      "fileName": "ForkJoinTask.java",
      "lineNumber": 1426,
      "methodName": "exec"
    },
```

This is a follow-on PR to #1392. That other PR solved the problem for one specific click path. Hopefully this PR solves it for all paths.

**REVIEWER:** note that you cannot easily reproduce this running Rawls locally, because our local-run scripts use the JDK as their base image. https://docs.google.com/document/d/1H1ib08LHs6XNNBMltWkmJedsTIMc0sblhAFsbNGT3kc/edit has instructions on running/reproducing locally.

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
@@ -868,10 +868,10 @@ class WorkspaceJsonSupport extends JsonSupport {
     val LINE_NUMBER = "lineNumber"
 
     def write(stackTraceElement: StackTraceElement) =
-      JsObject( CLASS_NAME -> JsString(stackTraceElement.getClassName),
-                METHOD_NAME -> JsString(stackTraceElement.getMethodName),
-                FILE_NAME -> JsString(stackTraceElement.getFileName),
-                LINE_NUMBER -> JsNumber(stackTraceElement.getLineNumber) )
+      JsObject( CLASS_NAME -> Option(stackTraceElement.getClassName).map(JsString(_)).getOrElse(JsNull),
+                METHOD_NAME -> Option(stackTraceElement.getMethodName).map(JsString(_)).getOrElse(JsNull),
+                FILE_NAME -> Option(stackTraceElement.getFileName).map(JsString(_)).getOrElse(JsNull),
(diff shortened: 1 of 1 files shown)


davidangb:
error during fiab-start, may have been an ssh disconnect. Jenkins retest.

kshakir (commented):
Where is `def read(json: JsValue)` used? It looks like it's now out of sync with the modified `def write(stackTraceElement: StackTraceElement)`, where the latter may return `JsNull` values.

davidangb:
@kshakir that `read()` method can be used if Rawls is calling to another service - e.g. Sam or Cromwell - and that service responds with an ErrorReport that Rawls tries to deserialize. I don't _think_ there are any cases where Rawls serializes and then deserializes the same ErrorReport object instance, so I don't know if write() and read() need to be round-tripped. That said, I updated read() a bit. Good catch!

kshakir (approved):
`write` and `read` are NOT symmetric as of commit `8725f23`. Ex: Potentially `CLASS_NAME` could be `null` will be written as a `JsNull` that will error upon read. But I'm going to give this a "LGTM" as from what I've gathered from the AS-696 report is that this asymmetry should not be a problem anytime soon 🤞

MatthewBemis (approved):

PR #1397, opened 2021-03-24
remove unwanted billingProject/dataReference params

davidangb:
**Note that this PR is looking to merge into `mb-as-668`, not into `develop`!!** Since that other PR has a lot of swagger changes, I thought it would be easier to combine all this work into that PR.

This PR covers AS-687 and AS-680. It removes the billingProject and dataReference query param from the delete-entity API, and removes the billingProject query param from the entity-type-metadata API.


MatthewBemis (approved):
This seems reasonable to me

zarsky-broad (approved):
Makes sense to me.

PR #1399, opened 2021-03-25
Tweak github test action [no ticket]

davidangb:
This fixes a minor annoyance in github actions test checks - which I wrote in the first place, so it fixes my bad.

Before this PR, GHA checks would fail on the "Fail if tests failed" job, which required extra clicks to find the individual unit test failure(s).

After this PR, the "Run tests" job is the one that fails, and github should open right to a summary of the test failure.

**Before:**
testfailures-before

**After:**
testfailures-after


MatthewBemis (approved):
Super glad to see this, thanks for fixing it! Not a github workflows expert but this looks fine to me

PR #1404, opened 2021-03-26
[CA-1224] Adds google project display name and tests

s-rubenstein:
related to: https://github.com/broadinstitute/workbench-libs/pull/597

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2125,6 +2130,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
               case Some(ba) if !billingProject.invalidBillingAccount => ba
               case _ => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"Billing Account is missing or invalid for Billing Project: ${billingProject}"))
             }
+            val workspaceName = WorkspaceName(workspaceRequest.namespace, workspaceRequest.name)
```
andy7i, line 2148: maybe outside of the scope of this ticket, but the case class `WorkspaceName` seems ambiguous, especially when one of the params is the workspace name. I wonder if we can rename the case class to FullWorkspaceName or at least rename the val :/

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2134,7 +2140,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
               resource <- createWorkspaceResourceInSam(workspaceId, billingProjectOwnerPolicyEmail, workspaceRequest, parentSpan)
               policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail] = resource.accessPolicies.map(x => SamResourcePolicyName(x.id.accessPolicyName) -> WorkbenchEmail(x.email)).toMap
               (googleProjectId, googleProjectNumber) <- traceDBIOWithParent("setupGoogleProject", parentSpan)(_ => DBIO.from(
-                  setupGoogleProject(billingProject, billingAccount, workspaceId, policyEmailsByName, billingProjectOwnerPolicyEmail, parentSpan)))
+                  setupGoogleProject(billingProject, billingAccount, workspaceId, workspaceName, policyEmailsByName, billingProjectOwnerPolicyEmail, parentSpan)))
```
andy7i, line 2158: are there changes updating `setupGoogleProject`?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1364,6 +1364,26 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
     verify(services.resourceBufferService).getGoogleProjectFromBuffer(any[ProjectPoolType], any[String])
   }
 
+  it should "Update a Google Project name after claiming from Resource Buffering Service" in withTestDataServices { services =>
+    val newWorkspaceName = "space_for_workin"
+    val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
+
+    val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+
+    verify(services.gcsDAO).updateGoogleProjectName(any[GoogleProjectId], ArgumentMatchers.eq(s"${testData.testProject1Name.value}/${newWorkspaceName}"))
```
andy7i: suggestion: create a `val namespace = testData.testProject1Name.value`

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1364,6 +1364,26 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
...
+  }
+
+  it should "Apply labels to a Google Project after claiming from Resource Buffering Service" in withTestDataServices { services =>
+    val newWorkspaceName = "space_for_workin"
+    val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
+
+    val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+
+    verify(services.gcsDAO).addGoogleProjectLabel(any[GoogleProjectId], ArgumentMatchers.eq("workspaceNamespace"), ArgumentMatchers.eq(newWorkspaceName))
+    verify(services.gcsDAO).addGoogleProjectLabel(any[GoogleProjectId], ArgumentMatchers.eq("workspaceName"), ArgumentMatchers.eq(testData.testProject1Name.value))
```
andy7i: the namespace and name values are swapped

```suggestion
    verify(services.gcsDAO).addGoogleProjectLabel(any[GoogleProjectId], ArgumentMatchers.eq("workspaceNamespace"), ArgumentMatchers.eq(testData.testProject1Name.value))
    verify(services.gcsDAO).addGoogleProjectLabel(any[GoogleProjectId], ArgumentMatchers.eq("workspaceName"), ArgumentMatchers.eq(newWorkspaceName))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1030,46 @@ class HttpGoogleServicesDAO(
...
+    for {
+      project <- getGoogleProject(googleProject)
+      _ = project.getLabels.put(key, value)
+      _ <- executeGoogleRequestWithRetry(cloudResourceManager.projects().update(googleProject.value, project))
+    } yield {
+
+    }
+  }
+
+  override def updateGoogleProjectName(googleProject: GoogleProjectId, name: String): Future[Unit] = {
```
andy7i: suggestion: rename or add comment that this is the google project display name

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1030,46 @@ class HttpGoogleServicesDAO(
...
+    implicit val service = GoogleInstrumentedService.CloudResourceManager
+
+    val cloudResourceManagerServiceAccountCredential = getCloudResourceManagerServiceAccountCredential
+    val cloudResourceManager = getCloudResourceManager(cloudResourceManagerServiceAccountCredential)
+    for {
+      _ <- executeGoogleRequestWithRetry(cloudResourceManager.projects().update(googleProject.value, new Project().setName(name)))
```
andy7i: It looks like we need to do some string validation because the displayname looks pretty limited :/

```
  /**
   * The user-assigned display name of the Project. It must be 4 to 30 characters. Allowed
   * characters are: lowercase and uppercase letters, numbers, hyphen, single-quote, double-quote,
   * space, and exclamation point.
   *
   * Example: My Project Read-write.
   * @param name name or {@code null} for none
   */
  public Project setName(java.lang.String name) {
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1030,46 @@ class HttpGoogleServicesDAO(
     }
   }
 
+  override def addGoogleProjectLabel(googleProject: GoogleProjectId, key: String, value: String): Future[Unit] = {
```
andy7i: do we do any string validation for key, value? some prior code example we did in DM: https://github.com/broadinstitute/gcp-dm-templates/blob/master/firecloud_project.py#L371

and buffer: https://github.com/DataBiosphere/terra-resource-buffer/blob/17f08a85c249546732975a998f3c7d261fa13e02/src/main/java/bio/terra/buffer/service/resource/flight/CreateProjectStep.java#L116

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1030,46 @@ class HttpGoogleServicesDAO(
...
+    implicit val service: GoogleInstrumentedService.Value = GoogleInstrumentedService.CloudResourceManager
+
+    val cloudResourceManagerServiceAccountCredential = getCloudResourceManagerServiceAccountCredential
+    val cloudResourceManager = getCloudResourceManager(cloudResourceManagerServiceAccountCredential)
+    for {
+      project <- getGoogleProject(googleProject)
+      _ = project.getLabels.put(key, value)
```
andy7i: just double checking if this does save the label to the map within the `project` object. Seems like it's a java map and I don't know what the behavior would be OTTOMH.

andy7i (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1030,46 @@ class HttpGoogleServicesDAO(
     }
   }
 
+  override def addGoogleProjectLabel(googleProject: GoogleProjectId, key: String, value: String): Future[Unit] = {
+    implicit val service: GoogleInstrumentedService.Value = GoogleInstrumentedService.CloudResourceManager
+
+    val cloudResourceManagerServiceAccountCredential = getCloudResourceManagerServiceAccountCredential
+    val cloudResourceManager = getCloudResourceManager(cloudResourceManagerServiceAccountCredential)
```
gpolumbo-broad: perhaps a nit, but it looks like these two lines of code are used 3 times now in `addGoogleProjectLabel`, `updateGoogleProjectName`, and `deleteGoogleProject`.  It's only 2 lines of code, so it's not that big of a deal, but we could consider extracting.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
gpolumbo-broad: Could we refactor this to take a `Map[String, String]` of labels so that we can add several at once and only make 1 call to Google?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
gpolumbo-broad: I know I can probably Google this myself too and perhaps I will, but why for updating the labels, do we need to first call `getGoogleProject`, update the labels, and then `update` the project.  Whereas in `updateGoogleProjectName`, we do not get the project first and instead just create a new `Project` and specify the field we want updated?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1927,6 +1928,10 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     for {
       googleProjectId <- traceWithParent("getGoogleProjectFromBuffer", span)(_ => resourceBufferService.getGoogleProjectFromBuffer(projectPoolType, workspaceId))
       _ <- traceWithParent("updateGoogleProjectBillingAccount", span)(_ => gcsDAO.updateGoogleProjectBillingAccount(googleProjectId, Option(billingAccount)))
+      _ <- traceWithParent("updateGoogleProjectDisplayName", span)(_ => gcsDAO.updateGoogleProjectName(googleProjectId, workspaceName.toString))
+      _ <- traceWithParent("addGoogleProjectLabel: namespace", span)(_ => gcsDAO.addGoogleProjectLabel(googleProjectId, "workspaceNamespace", workspaceName.namespace))
+      _ <- traceWithParent("addGoogleProjectLabel: name", span)(_ => gcsDAO.addGoogleProjectLabel(googleProjectId, "workspaceName", workspaceName.name))
+      _ <- traceWithParent("addGoogleProjectLabel: id", span)(_ => gcsDAO.addGoogleProjectLabel(googleProjectId, "workspaceId", workspaceId))
```
gpolumbo-broad: As noted above, I think we can consolidate all the label calls into a single call.  Wondering if we can't also consolidate all of the calls to update the Google Project into a single call.  So set the Billing Account, Name, Labels all into a single `updateGoogleProject` call.  

Taking it a step further, `addGoogleProjectLabel` calls `getGoogleProject` as does `getGoogleProjectNumber`.  Could we optimize to make one call to get the Google Project (and therefore the Google Project Number) and then we can use that instance to make any updates to set the Labels, Billing Account, and Name, and then make one single call back to Google to update the Google Project.  

I think reducing the amount of chatter back and forth with Google is something we should generally aim for whenever we can to reduce overall latency, potential quota issues, and just overhead in general.

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
gpolumbo-broad: @andy7i, yeah, certainly outside the scope of this ticket.  I'm sure this is used in a ton of places but I 100% agree with you that the name is confusing and it throws me off all the time and I'd LOVE to change it.  Can you make a tech debt ticket?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
gpolumbo-broad: I have a slight preference towards commenting.  Since this is the `GoogleServicesDAO`, it is specifically about getting/setting values in Google land and this _is_ what the field is called on the Google Project.  That said, any time I need to do something with the Google Project Name or ID I reference this page:  https://cloud.google.com/resource-manager/docs/creating-managing-projects#before_you_begin which disambiguates `Project ID`, `Project Name`, and `Project Number`.  Some commenting here would be helpful and perhaps a link to the Google documentation.  Yeah, it might move, but it's the canonical source for what these things are and what they mean right now, and that's the best we can do.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: yeah definitely, I didn't realize it but `WorkspaceName` is used in 100 places  -- created https://broadworkbench.atlassian.net/browse/CA-1264

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: Combined the calls that I could, reducing ~7 calls to 2. 

The billing account uses a different service (cloud billing) so that can't be combined with the other calls here. Just in case, I did some investigation, but didn't see a way to batch calls to google across different gcp services (for example, cloud billing APIs vs cloud resource manager APIs).

If we need to, I think there could be some additional optimization for gcsDAO.removePolicyBindings and updateGoogleProjectIam to combine some of the calls, but it get messy so we discussed not spending too much time on that.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: RBS gives us projects with existing labels, so I think this is the right call so we don't replace any existing labels.  I've added a comment about this to clarify

andy7i: Not sure if this was working but I changed the logic to use more scala.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: extracted as suggested and replaced it in some other areas where `getCloudResourceManager` was being used (sometimes with a SA that wasn't the CloudResourceManager SA)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
andy7i: added validation ; waiting on answers from PO about how this should look https://broadworkbench.atlassian.net/browse/CA-1224

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
s-rubenstein: Yeah, I had been thinking about that. I think it is also safer from a management perspective (what happens if an asynchronous call futzes with the labels.) Was thinking it would be cleanest just to have an `updateGoogleProject` call, but for some reason that wasn't the current pattern in the codebase, so I tried to stick with what was existing. Much happier combining though.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1927,7 +1930,8 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     for {
       googleProjectId <- traceWithParent("getGoogleProjectFromBuffer", span)(_ => resourceBufferService.getGoogleProjectFromBuffer(projectPoolType, workspaceId))
       _ <- traceWithParent("updateGoogleProjectBillingAccount", span)(_ => gcsDAO.updateGoogleProjectBillingAccount(googleProjectId, Option(billingAccount)))
-      googleProjectNumber <- traceWithParent("getProjectNumberFromGoogle", span)(_ => getGoogleProjectNumber(googleProjectId))
+      googleProjectLabels = Map("workspaceNamespace" -> workspaceName.namespace, "workspaceName" -> workspaceName.name, "workspaceId" -> workspaceId)
```
s-rubenstein: Opt: I think my preference would be to call labelSafeString on these here, but I am okay with it as is atm. I don't like that the `setUpProjectInCloudResourceManagerAndGetGoogleProjectNumber` changes labels around on users of the method. I would prefer that function to have minimal side effects, and that seems like a fairly obscure/hidden one.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1951,12 +1955,27 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      googleProjectNumber = gcsDAO.getGoogleProjectNumberFromOption(googleProjectId, Option(GoogleProjectNumber(googleProject.getProjectNumber.toString)))
+
+      // RBS projects already come with some labels. In order to not lose those, we need to combine those existing labels with the new labels
+      existingLabels = googleProject.getLabels.asScala
+
+      updatedLabels = existingLabels ++ validLabels
+
+      // create a Project with fields that we want to update. Then send that to gcsDAO to update the project
+      googleProjectWithValuesToUpdate = new Project()
+        .setName(gcsDAO.googleProjectNameSafeString("Terra-managed Google project"))
```
s-rubenstein: TOL: This seems like minimal information here, would we want some sort of identifier on which Terra billing project it was associated with? I just wonder if we can give some information to users at all. Although maybe it doesn't matter. Do we expect users to care about their google project?

If so, who does care? Where will this be shown? I know we were going to show google project id in the workspace (in case users need that information for gs commands via the terminal or something), but I am wondering where this will really be used, especially since our users won't have direct access.

Maybe we/support are the users of this?

gpolumbo-broad: I know we're waiting on feedback from Janki here, but whatever name we choose here, I think our preference should be to make it unique to each Workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1927,7 +1930,8 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     for {
       googleProjectId <- traceWithParent("getGoogleProjectFromBuffer", span)(_ => resourceBufferService.getGoogleProjectFromBuffer(projectPoolType, workspaceId))
       _ <- traceWithParent("updateGoogleProjectBillingAccount", span)(_ => gcsDAO.updateGoogleProjectBillingAccount(googleProjectId, Option(billingAccount)))
-      googleProjectNumber <- traceWithParent("getProjectNumberFromGoogle", span)(_ => getGoogleProjectNumber(googleProjectId))
+      googleProjectLabels = Map("workspaceNamespace" -> workspaceName.namespace, "workspaceName" -> workspaceName.name, "workspaceId" -> workspaceId)
+      googleProjectNumber <- traceWithParent("setUpProjectInCloudResourceManagerAndGetGoogleProjectNumber", span)(_ => setUpProjectInCloudResourceManagerAndGetGoogleProjectNumber(googleProjectId, googleProjectLabels))
```
gpolumbo-broad: This feels a little clunky to me now.  Perhaps we try something like:
```suggestion
      googleProject <- traceWithParent("setUpProjectInCloudResourceManager", span)(_ => setUpProjectInCloudResourceManager(googleProjectId, googleProjectLabels))
      googleProjectNumber = gcsDAO.getGoogleProjectNumberFromOption(googleProjectId, Option(GoogleProjectNumber(googleProject.getProjectNumber.toString)))
```
Of course `setUpProjectInCloudResourceManager` would need to be updated to return the project, but that looks trivial.

My preference is to try not to have a single function doing too much.  The important thing this function is doing is the project setup.  The fact that it returns a Project or a Project Number is secondary, so l think this code becomes more maintainable if the steps are separated.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -232,17 +233,55 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+    *
+    * For more info see: https://cloud.google.com/resource-manager/reference/rest/v1/projects
+    * @param name
+    * @return
+    */
+  def googleProjectNameSafeString(name: String): String = {
+    name.toLowerCase.replaceAll("[^a-z0-9\\-'\" !]", "-").take(30)
+  }
+
+  def getGoogleProjectNumberFromOption(googleProjectId: GoogleProjectId, maybeGoogleProjectNumber: Option[GoogleProjectNumber]): GoogleProjectNumber = maybeGoogleProjectNumber match {
```
gpolumbo-broad: There's a couple of things I'm unsure about with this and I don't know what the answers are yet either unfortunately:

1. I'm not a huge fan of passing in the `googleProjectId` just for generating the error message.
2. I'm trying to ask myself what the true purpose is of this method.  Obviously it's DRY'ing up some code, that's good.  But the real "work" this method is doing for us doesn't seem to be "getting the Project Number", to me it seems to be about consistently handling the null case where our Project is missing the Number.  

I'm wondering if we could do something like `Option(googleProject.getProjectNumber).getOrElse(throw new RawlsExceptionWithErrorReport(...))`.  I think the things that I'd want to address are a consistent way of getting the `Long` project number off of the `googleProject`, turning it into a `GoogleProjectNumber`, and when it's null, generating the right ErrorReport.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -232,17 +233,55 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+    case None => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadGateway, s"Failed to retrieve Google Project Number for Google Project ${googleProjectId}"))
+    case Some(longProjectNumber) => GoogleProjectNumber(longProjectNumber.toString)
```
gpolumbo-broad:
```suggestion
    case Some(longProjectNumber) => longProjectNumber
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -232,17 +233,60 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+  def googleProjectNameSafeString(name: String): String = {
+    name.replaceAll("[^a-zA-Z0-9\\-'\" !]", "-").take(30)
+  }
+
+  /**
+    * Handles getting the google project number from the google [[Project]]
+    * @param googleProject
+    * @return GoogleProjectNumber
+    */
+  def getGoogleProjectNumberFromGoogleProject(googleProject: Project): GoogleProjectNumber = googleProject.getProjectNumber match {
```
gpolumbo-broad: nit: I think it's ok to call this just `getGoogleProjectNumber` and it will read a just as well

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -88,11 +88,13 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
...
       val mockGcsDAO = mock[GoogleServicesDAO](RETURNS_SMART_NULLS)
-      val googleProjectNumber = 42L
-      when(mockGcsDAO.getGoogleProject(project.googleProjectId)).thenReturn(Future.successful(new Project().setProjectNumber(googleProjectNumber)))
+      val googleProjectNumber = GoogleProjectNumber("42")
+      val googleProject = new Project().setProjectNumber(googleProjectNumber.value.toLong)
+      when(mockGcsDAO.getGoogleProject(project.googleProjectId)).thenReturn(Future.successful(googleProject))
       val folderId = "folders/1234567"
       when(mockGcsDAO.getFolderId(defaultServicePerimeterName.value.split("/").last)).thenReturn(Future.successful(Option(folderId)))
       when(mockGcsDAO.addProjectToFolder(project.googleProjectId, folderId)).thenReturn(Future.successful(()))
+      when(mockGcsDAO.getGoogleProjectNumberFromGoogleProject(googleProject)).thenReturn(googleProjectNumber)
```
gpolumbo-broad: I'm not sure you need to mock this?  I think that function should be self contained and should just work now?

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: I guess I was thinking that calling `labelSafeString()` outside of the function: 1) would require devs to know that there is a need to use `labelSafeString()` and to remember to use it before updating labels and 2) could result in obscure google error messages if we accidentally don't run `labelSafeString` and send an invalid label.

it does look a bit cleaner being outside the function so I updated to that

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1031,20 +1029,35 @@ class HttpGoogleServicesDAO(
...
+    * @param googleProjectId project to update
+    * @param googleProjectWithUpdates [[Project]] with values to update. For example, a (new Project().setName("ex")) will update the name of the googleProjectId project.
+    * @return
+    */
+  override def updateGoogleProject(googleProjectId: GoogleProjectId, googleProjectWithUpdates: Project): Future[Unit] = {
+    implicit val service = GoogleInstrumentedService.CloudResourceManager
+    val cloudResourceManager: CloudResourceManager = getCloudResourceManagerWithCloudResourceManagerServiceAccount
+
+    for {
+      _ <- executeGoogleRequestWithRetry(cloudResourceManager.projects().update(googleProjectId.value, googleProjectWithUpdates))
```
andy7i: I kind of agree with codecov here. It would give me a bit of confidence if we had a unit test here to know that we did in fact call `cloudResourceManager.projects().update()` but I don't see any other unit tests for GoogleServicesDAO. Maybe the right thing to do is to create an integration test?

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
andy7i: I think because mockGcsDAO is a mockito mock:
```
val mockGcsDAO = mock[GoogleServicesDAO](RETURNS_SMART_NULLS)
```

it requires any call to be stubbed (even with my updates to `getGoogleProjectNumber`). Error message:

```
org.scalatest.exceptions.TestFailedException: The future returned an exception of type: org.mockito.exceptions.verification.SmartNullPointerException, with message: 
You have a NullPointerException here:
-> at org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceComponent$workspaceQuery$.updateGoogleProjectNumber(WorkspaceComponent.scala:270)
because this method call was *not* stubbed correctly:
-> at org.broadinstitute.dsde.rawls.user.UserService.$anonfun$addProjectToServicePerimeter$9(UserService.scala:658)
googleServicesDAO.getGoogleProjectNumber(
    {"projectNumber" = 42L}
);
```

andy7i (commented):
I can't comment on the earlier threads anymore.. 
I confirmed with janki about how the display name and labels should look
I also changed `getGoogleProjectNumberFromOption` but am considering inlining those instead like you were suggesting.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: Janki and I went over the questions we had in the jira ticket and we decided to go with:

**display name**
* use the format “< namespace >--< name >” truncated to 30 chars

**Labels**
* use labels, truncating at 63 chars and replace all illegal characters with “-”

which should be reflected in the code now

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
s-rubenstein: Yeah, I also looked for that… an integration test seems like a good idea.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -232,17 +233,60 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
   }
 
+  /**
+    * Convert a map of labels to legal gcp label text. Runs [[labelSafeString]] on all keys and values in the map.
+    * @param m Map of label key value pairs
+    * @param prefix
+    * @return
+    */
+  def labelSafeMap(m: Map[String, String], prefix: String = ""): Map[String, String] = m.map { case (key, value) =>
+    labelSafeString(key, prefix) -> labelSafeString(value, prefix) }
```
s-rubenstein, line 263: Should this have the same default as `labelSafeString`? I am wondering if it will override the default and do it all with `""` as the prefix, which seems unintuitive to me.

s-rubenstein (commented):
Overall, looks good to me, with one comment. I can't hit approve cause I am the PR author.

andy7i (approved):
sky's 👍

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1364,6 +1365,42 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
...
+    val newWorkspaceName = "space_for_workin"
+    val workspaceRequest = WorkspaceRequest(newWorkspaceNamespace, newWorkspaceName, Map.empty)
+    val captor = ArgumentCaptor.forClass(classOf[Project])
+
+    val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+
+    verify(services.gcsDAO).updateGoogleProject(ArgumentMatchers.eq(GoogleProjectId("project-from-buffer")), captor.capture())
+    val capturedProject = captor.getValue.asInstanceOf[Project] // Explicit cast needed since Scala type interference and capturing parameters with Mockito don't play nicely together here
+
+    val expectedNewLabels = services.gcsDAO.labelSafeMap(Map("workspaceNamespace" -> newWorkspaceNamespace, "workspaceName" -> newWorkspaceName, "workspaceId" -> workspace.workspaceId), "")
```
gpolumbo-broad: I would just hard code the expected values here rather than using `labelSafeMap`.  I would also give the workspace name a couple of special characters so that you can make sure the `createWorkspace` call creates the labels with the appropriate special character replacement.  Maybe even make a long workspace name too to make sure it truncates properly.

gpolumbo-broad (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspaceApiSpec.scala
```diff
@@ -100,6 +100,40 @@ class WorkspaceApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLi
         Rawls.billingV2.deleteBillingProject(billingProjectName)
       }
 
+      "to create a v2 workspaces (in a v2 billing project) with properly updated labels and display name" in {
```
gpolumbo-broad: Do we need this test?  This test suite already tests that we can create projects and `WorkspaceServiceSpec` already tests that the labels and display name are set correctly.  I don't see this test adding value or covering functionality that we don't already have covered.

If you _do_ think we need this test, then:

nit: for the sake of making progress, I don't think we should do this now, but I think this test should be named differently and located elsewhere.  I would rather call this test something like, "Rawls should set labels on the underlying Google Project when creating a new Workspace".  I feel like these clearly articulates the logic under test.  As it is currently written, it seems to me that this is a test about Rawls Billing Project owners and what they can do.  The fact that they can create Workspaces, to me, is one test.  The fact that when we create Workspaces, labels (and display name) are set is a totally different test.  

Again, it's a nit, no need to change it now, just something to think about as we create and organize tests.

gpolumbo-broad (approved):

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspaceApiSpec.scala
andy7i: I was able to set up the test name to what you suggested and the test seems happy. It seemed easy enough to move now so I opted to do that instead of waiting, in case we forgot.

(We talked about this in office time: I think we do need this test. The main difference between this test and the other tests is that this is an integration test. I think it provides value in that we don't really validate labels on a real project anywhere else. This test will ensure that we update real projects correctly. It saved me the hassle of finding the project in GCP console to verify if the project was updated correctly and should ensure that our logic for updating the project doesn't break in the future.)

PR #1405, opened 2021-03-29
increase timeout for avro upsert monitor unit tests [AS-706]

davidangb:
_... commit 1, increase test timeouts ..._
unit test run 1: passed
unit test run 2: passed
unit test run 3: passed
unit test run 4: passed
unit test run 5: passed
unit test run 6: passed (an unrelated test in DataAccessSpec failed, but avro tests were fine)
unit test run 7: **boo, AvroUpsertMonitor still had a failure**
unit test run 8: passed
unit test run 9: passed
_... commit 2, increase timeouts and add clues ..._
unit test run 10: **boo, AvroUpsertMonitor still had a failure**
_... commit 3, different topics/subscriptions for each test ..._ 
unit test run 11: passed
unit test run 12: passed
unit test run 13: passed
unit test run 14: passed
unit test run 16: passed
unit test run 17: passed
unit test run 18: **boo, AvroUpsertMonitor still had a failure**
_... commit 4, more debug logging ..._ 
unit test run 19: passed
unit test run 20: passed
unit test run 21: **boo, AvroUpsertMonitor still had a failure**
_... commit 5, even yet more debug logging ..._ 
unit test run 22: passed
unit test run 23: passed
unit test run 24: passed
unit test run 25: passed
unit test run 26: passed
unit test run 27: passed (an unrelated test in DataAccessSpec failed, but avro tests were fine)
unit test run 28: passed
unit test run 30: passed
unit test run 31: **boo, AvroUpsertMonitor still had a failure**
_... commit 6, debug logging again again again ..._ 
unit test run 32: **boo, AvroUpsertMonitor still had a failure**
_... commit 7, found the futures ..._ 
unit test run 33: passed

failed in run 6:
```
[info] DataAccessSpec:
[info] DataAccess
[info] - should not deadlock due to too few threads *** FAILED ***
[info]   A timeout occurred waiting for a future to complete. Waited 10 seconds. (DataAccessSpec.scala:69)
```

failed in run 7:
```
[info] AvroUpsertMonitorSpec:
[info] AvroUpsertMonitor
[info] - should mark the import job as failed on valid json that doesn't match our model *** FAILED ***
[...]

PR #1407, opened 2021-03-31
AvroUpsertMonitorSpec setUp needs to wait for futures [AS-706]

davidangb:
`AvroUpsertMonitorSpec` had a `setUp` method that created pubsub topics and subscriptions in the mock pubsub dao. These calls to create are `Future`s, but the method did not wait for them to complete.

Under just the right conditions - maybe 10% of the time? - one of the necessary assets (usually the subscription) would not be created before a test started to use it, and a test would fail.

In this PR, we wait for those futures to complete.

In a separate draft PR #1405, I saw 6 failures in 32 test runs. I'm going to run the unit tests multiple times in this PR to ensure they pass. Here are the results for each unit test run:
1. ✅  passed
2. ✅  passed
3. ✅  passed
4. ✅  passed
5. ✅  passed
6. ✅  passed
7. ✅  passed
8. ✅  passed
9. ✅  passed
10. ✅  passed
11. ✅  passed

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
@@ -96,11 +96,21 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
   )
 
   def setUp(services: TestApiService) = {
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

sagehen03 (approved):
Looks good to me.  Sneaky bug, as I ran the old version of the test at least 5 times locally and could not induce a failure.

PR #1410, opened 2021-04-07
AS-709 Handle exceptions from WSM

ungwudik:
- Handle exceptions from WSM endpoints
- Also, return 404 (and handle exceptions) instead of 500 when attempting to update a non-existent snapshot

Created AS-718 as follow-up work.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -107,6 +107,8 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
   def updateSnapshot(workspaceName: WorkspaceName, snapshotId: String, updateInfo: UpdateDataReferenceRequestBody): Future[Unit] = {
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
+      // check that snapshot exists before updating it. If the snapshot does not exist, the GET attempt will throw a 404
+      // TODO: these WSM APIs are in the process of being deprecated. We should update with new APIs as they become available
       workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
       workspaceManagerDAO.updateDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, updateInfo, userInfo.accessToken)
     }
@@ -115,6 +117,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
   def deleteSnapshot(workspaceName: WorkspaceName, snapshotId: String): Future[Unit] = {
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
+      // check that snapshot exists before deleting it. If the snapshot does not exist, the GET attempt will throw a 404
       workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
       workspaceManagerDAO.deleteDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
     }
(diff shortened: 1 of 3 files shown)


ungwudik:
Jenkins Automation Tests failed, but those don't look related to this PR. I assume it's due to flakiness.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -107,6 +107,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
   def updateSnapshot(workspaceName: WorkspaceName, snapshotId: String, updateInfo: UpdateDataReferenceRequestBody): Future[Unit] = {
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
+      workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
```
sagehen03, line 112: This seems unusual to me and unrelated to handling WSM exceptions.  Is there some necessary side effect we get from calling `getDataReference`?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -43,6 +44,8 @@ object RawlsApiService extends LazyLogging {
       case rollback:MySQLTransactionRollbackException =>
         logger.error(s"ROLLBACK EXCEPTION, PROBABLE DEADLOCK: ${rollback.getMessage} [${rollback.getErrorCode} ${rollback.getSQLState}] ${rollback.getNextException}", rollback)
         complete(StatusCodes.InternalServerError -> ErrorReport(rollback))
+      case wsmApiException: ApiException =>
```
sagehen03, line 47: seems reasonable to me :)

sagehen03 (commented):
I'm still new to Scala and Rawls, but I think it would be nice to add a unit test to confirm that you're getting the change you expect.  At the top of `SnapshotApiServiceSpec` you should be able to  override a method like getDataReference and have that method for a particular argument throw a ApiException.  Let me know if that sounds like something you'd be interested in and I can try to add more detail.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: `updateDataReference` returns a 20x even if the reference does not exist, so we need `getDataReference` here in order to generate the 404 in the case of nonexistent references.

these WSM APIs are in the process of being deprecated and replaced by newer-version APIs, so I don't think it's worth changing the WSM behavior. That said, we should probably add explanatory code comments/TODOs here.

davidangb (commented):
code change looks good. Are there any tests that can/should be written or updated?

ungwudik:
> I'm still new to Scala and Rawls, but I think it would be nice to add a unit test to confirm that you're getting the change you expect. At the top of `SnapshotApiServiceSpec` you should be able to override a method like getDataReference and have that method for a particular argument throw a ApiException. Let me know if that sounds like something you'd be interested in and I can try to add more detail.

We do have a mocked [unit test here](https://github.com/broadinstitute/rawls/blob/09d8572256be1a06e0bc44a95e0220e325f46d7d/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiServiceSpec.scala#L339) (see P.S./context below) for verifying that attempting to update a non-existent snapshot returns a `404`. But if you meant specifically testing that `bio.terra.workspace.client.ApiException` is thrown, then we can also discuss that/its value.


**P.S.**
Some context that may help: The unit test in `SnapshotApiServiceSpec` was (stay with me) ahead of the `updateSnapshot` function that it's testing. This is because:
1. The function was originally returning 404 as expected, and this behavior matched the test. So far, so good.
2. Then, the behavior changed in WSM causing it to start returning 204 when updating a non-existent snapshot. This 204 bubbled up to Rawls as well.
3. But after #2 happened, the mocked test stayed the same (i.e. still returning the expected 404). This meant there was a mismatch between the functionality AND what the test was doing.
4. We later decided that Rawls should return 404 even if WSM returns a 204. This is why I'm updating the functionality here, but the test already matches the expectation.

Let me know if you have any questions or need clarification regarding this context. I'm happy to discuss offline as well.

ungwudik:
> code change looks good. Are there any tests that can/should be written or updated?

See my response to Drew about test [here](https://github.com/broadinstitute/rawls/pull/1410#issuecomment-815973602). That said, I've updated `MockWorkspaceManagerDAO.scala` to remove the now-unnecessary `throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not found"))`, and have also added a condition for updating snapshot reference.

ungwudik:
> > code change looks good. Are there any tests that can/should be written or updated?
> 
> See my response to Drew about test [here](https://github.com/broadinstitute/rawls/pull/1410#issuecomment-815973602). That said, I've updated `MockWorkspaceManagerDAO.scala` to remove the now-unnecessary `throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not found"))`, and have also added a condition for updating snapshot reference.

Let me know if you have further feedback.

sagehen03:
That all makes sense @ungwudik  Thanks for the explanation approving now!  I do think it would make sense to remove the RawlsException being thrown from the mock since that should happen via the Rawls exception handler now.

sagehen03 (approved):

davidangb (approved):
:+1: Updating the mock to not throw 404 is great, thank you!

PR #1412, opened 2021-04-08
[CA-1206] Attempt to fix leak in project deletion

s-rubenstein:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -320,7 +320,10 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
       if (projectChildren.contains(SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))) {
         for {
           _ <- deletePetsInProject(GoogleProjectId(projectName.value), userInfoForSam)
-          _ <- gcsDAO.deleteProject(GoogleProjectId(projectName.value))
+          _ <- toFutureTry(gcsDAO.deleteProject(GoogleProjectId(projectName.value))) map {
+            case Failure(t) => logger.warn(s"failure deleting google project ${projectName} from google during project deletion.", t)
+            case Success(b) => b
+          }
           _ <- samDAO.deleteResource(SamResourceTypeNames.googleProject, projectName.value, userInfoForSam)
         } yield ()
       } else {


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -320,7 +320,10 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
       if (projectChildren.contains(SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))) {
         for {
           _ <- deletePetsInProject(GoogleProjectId(projectName.value), userInfoForSam)
-          _ <- gcsDAO.deleteProject(GoogleProjectId(projectName.value))
+          _ <- toFutureTry(gcsDAO.deleteProject(GoogleProjectId(projectName.value))) map {
+            case Failure(t) => logger.warn(s"failure deleting google project ${projectName} from google during project deletion.", t)
```
s-rubenstein, line 324: I don't really like this, it leaves a chance of us leaving orphaned google projects around. However, it does fix the leaks in Sam's db. However, is there a better way for us to do this? Maybe putting it on a queue to get cleaned up asynchronously, or flagging the project in our DB somehow?

I didn't want to just naively put the samDAO delete before the gcsDAO delete, because it would also leave us with orphaned resources, the same as this. That may be cleaner though? (If so, I can do that, remove the try logic here, and then we can rely on looking at error logs in deleteProject to tell us that there are orphaned projects)

andy7i: Maybe retries?

I think the main issue with CA-1206 was leaks in the Rawls DB, not Sam DB. (I think leaks in the Sam DB, while messy, won't cause too many issues.)

When we were cleaning up problem projects, we found that they were deleted from google, but not rawls. Did the logs for delete project show anything? (Just want to make sure this is causing the error and not Sam on the next line or maybe even Rawls.)

andy7i (commented):
do we need to update anything for v2_workspaces?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
s-rubenstein: So, `gcsDAO.deleteProject` already retries internally. I suspect that means that retries aren't the issue. I have not checked logs yet, but my hypothesis is that the issue is that the execution flow is getting blocked on this line, due to errors from google, and then not continuing to any other deletions.

PR #1415, opened 2021-04-09
delete BQ dataset when a snapshot is deleted [AS-584]

Shakespeared:
delete BQ dataset and dataset reference when a snapshot reference is deleted from a workspace. 

We didn't implement the case of deleting an entire workspace because we weren't able to get the dataset IDs from the API yet. The enumerate endpoint didn't give us the right info, but it looks like it is in progress. We will create a follow-up ticket to handle that.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -122,9 +122,12 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
       workspaceManagerDAO.deleteDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
 
       val datasetName = generateDatasetName(snapshotRef.getReferenceId)
-      val datasetRef = workspaceManagerDAO.getBigQueryDatasetReferenceByName(workspaceContext.workspaceIdAsUUID, datasetName, userInfo.accessToken)
-      workspaceManagerDAO.deleteBigQueryDatasetReference(workspaceContext.workspaceIdAsUUID, datasetRef.getMetadata.getReferenceId, userInfo.accessToken)
-      deleteBigQueryDataset(workspaceName, datasetName).unsafeToFuture()
+      deleteBigQueryDataset(workspaceName, datasetName).unsafeToFuture().map { _ =>
+        val datasetRef = workspaceManagerDAO.getBigQueryDatasetReferenceByName(workspaceContext.workspaceIdAsUUID, datasetName, userInfo.accessToken)
+        workspaceManagerDAO.deleteBigQueryDatasetReference(workspaceContext.workspaceIdAsUUID, datasetRef.getMetadata.getReferenceId, userInfo.accessToken)
+      }.recover {
+        case t: Throwable => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"Your snapshot reference was deleted, but an error occurred while deleting its Delta Layer companion dataset. Error: ${t.getMessage}"))
+      }
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -42,7 +42,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
       val dataRepoReference = new DataRepoSnapshot().instanceName(terraDataRepoInstanceName).snapshot(snapshot.snapshotId)
       val snapshotRef = workspaceManagerDAO.createDataReference(workspaceContext.workspaceIdAsUUID, snapshot.name, snapshot.description, ReferenceTypeEnum.DATA_REPO_SNAPSHOT, dataRepoReference, CloningInstructionsEnum.NOTHING, userInfo.accessToken)
 
-      val datasetName = "deltalayer_" + snapshotRef.getReferenceId.toString.replace('-', '_')
+      val datasetName = generateDatasetName(snapshotRef)
 
       val datasetLabels = Map("workspace_id" -> workspaceContext.workspaceId, "snapshot_id" -> snapshot.snapshotId)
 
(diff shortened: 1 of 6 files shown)


MatthewBemis:
jenkins retest ; rawls build failed for some reason

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -118,8 +118,13 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       // check that snapshot exists before deleting it. If the snapshot does not exist, the GET attempt will throw a 404
-      workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+      val snapshotRef = workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
       workspaceManagerDAO.deleteDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+
+      val datasetName = generateDatasetName(snapshotRef.getReferenceId)
+      val datasetRef = workspaceManagerDAO.getBigQueryDatasetReferenceByName(workspaceContext.workspaceIdAsUUID, datasetName, userInfo.accessToken)
+      workspaceManagerDAO.deleteBigQueryDatasetReference(workspaceContext.workspaceIdAsUUID, datasetRef.getMetadata.getReferenceId, userInfo.accessToken)
+      deleteBigQueryDataset(workspaceName, datasetName).unsafeToFuture()
```
davidangb: 1) if we're relying on the `name` of the dataset reference in WSM to correlate snapshots to DL datasets ... what happens if someone updates the name of that reference? Is there any failsafe for this case? At minimum, log a warning here if we can't find the DL dataset.

2) What's the error-handling story here? How should the API respond if we successfully delete the snapshot reference, but fail to delete either the dataset or its reference?

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -238,6 +238,48 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+      Await.result(snapshotService.deleteSnapshot(workspace.toWorkspaceName, snapshotUUID.toString), Duration.Inf)
+
+      verify(mockWorkspaceManagerDAO, times(1)).getDataReference(ArgumentMatchers.eq(workspace.workspaceIdAsUUID), ArgumentMatchers.eq(snapshotUUID),any[OAuth2BearerToken])
+      verify(mockWorkspaceManagerDAO, times(1)).deleteDataReference(ArgumentMatchers.eq(workspace.workspaceIdAsUUID), ArgumentMatchers.eq(snapshotUUID), any[OAuth2BearerToken])
+      verify(mockWorkspaceManagerDAO, times(1)).getBigQueryDatasetReferenceByName(ArgumentMatchers.eq(workspace.workspaceIdAsUUID), ArgumentMatchers.eq(deltaLayerDatasetName), any[OAuth2BearerToken])
+      verify(mockWorkspaceManagerDAO, times(1)).deleteBigQueryDatasetReference(ArgumentMatchers.eq(workspace.workspaceIdAsUUID), any[UUID], any[OAuth2BearerToken])
+      verify(mockBigQueryServiceFactory, times(1)).getServiceFromCredentialPath(any[String], any[GoogleProject])
+
+    }
+
```
davidangb, line 282: when we add integration tests (AS-688), should probably add an integration test for the delete case too.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
Shakespeared: @davidangb 1. The `name` that we're using is keyed off the `referenceId` not the human-readable snapshot name, so it can't be changed, so I think that should be ok. 

2. @MatthewBemis and I thought of basically either retrying the deleteBigQueryDataset call and if that fails reporting an error to the user, or not retrying and just throwing an error right away if that fails. Neither of those really solve the problem of orphaned resources though, and if it's a google problem a retry might not help anyway… do you have any better ideas?

davidangb: The case I'm worried about for line 125 is the case where we've automatically created the dataset reference and given it a name of `deltalayer_xxxxxxxx_xxxx_Mxxx_Nxxx_xxxxxxxxxxxx`, but at some point a human user edits that reference and gives it a name like `myFancyBigQueryDatasetReference`. Then, we'll never find it.

I guess this is only an issue if we expose the ability to rename dataset references. Currently we expose the ability to rename snapshot references, but I guess that we don't have to expose both - then, the danger would only be to internal users who have direct access to WSM itself.

davidangb: on the error-handling ... it it feasible to include a helpful message in an error response? Something like "Your snapshot reference was deleted, but an error occurred while deleting its Delta Layer companion." Also, I say swap the order of operations so that we delete the BigQuery dataset first, then delete the reference to it second ... that way, if we hit an error deleting the dataset reference, at least we've deleted the dataset and won't be costing the user any money.

Shakespeared: @davidangb These are the options that @MatthewBemis and I can think of. Right now we're leaning towards option 2, unless we can find out more information about 4 that makes it seem usable. 

> 1) prevent renaming references altogether
>  - downsides: seems arbitrary, prevents the legitimate use-case of wanting to rename something that's not a delta layer reference
>  - benefits: would definitely prevent references from being renamed and "lost" in all use cases
>  
> 2) prevent renaming through the UI
> - downsides: internal users could still rename and "lose" references through WSM directly
> - benefits: no implementation cost, doesn't disable any features
>
> 3) rawls keeps track of state for reference names
> - downsides: duplication of information, more heavy-weight solution, internal users could still rename and "lose"  references through WSM directly
> - benefits: doesn't disable any features
>
> 4) use the BQ labels feature to keep track of original reference name
> - downsides: the API for this seems underdeveloped, doesn't seem like labels are intended to be used that way
> - upsides: would be resilient, would not disable any features

davidangb: well, we already prevent renaming dataset references via the UI and the API, because we haven't exposed any of those APIs, so I think we're set on option 2 - we just have to beware if we ever decide to expose that functionality.

On option 4  - I think that's certainly possible as a fallback implementation, see https://googleapis.dev/python/bigquery/latest/usage/datasets.html#listing-datasets. If we find this is a problem in the future, we could implement lookup-by-label.

davidangb (approved):
lgtm, assuming there's some followup to add the integration test(s)

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -118,8 +118,16 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
-      workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+      val snapshotRef = workspaceManagerDAO.getDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
       workspaceManagerDAO.deleteDataReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+
+      val datasetName = generateDatasetName(snapshotRef.getReferenceId)
+      deleteBigQueryDataset(workspaceName, datasetName).unsafeToFuture().map { _ =>
+        val datasetRef = workspaceManagerDAO.getBigQueryDatasetReferenceByName(workspaceContext.workspaceIdAsUUID, datasetName, userInfo.accessToken)
+        workspaceManagerDAO.deleteBigQueryDatasetReference(workspaceContext.workspaceIdAsUUID, datasetRef.getMetadata.getReferenceId, userInfo.accessToken)
+      }.recover {
+        case t: Throwable => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"Your snapshot reference was deleted, but an error occurred while deleting its Delta Layer companion dataset. Error: ${t.getMessage}"))
```
ansingh7115: In the case where we delete a data reference by the BQ dataset deletion fails, can we log the name of the snapshot and/or BQ dataset? That way, if we need to clean up orphaned BQ datasets, we'll be able to grab them

ansingh7115 (approved):
Left one small comment

Shakespeared:
jenkins retest

PR #1420, opened 2021-04-14
Move upserts/updates to EntityProvider classes [AS-586]

davidangb:
part of AS-586 but does not complete that ticket.

This PR should not change any functionality. It is purely a refactoring/reorganization, which will be necessary to build the real functionality. I thought it would be useful to do the reorg in a standalone PR to make it easier to review.

At a high level, this PR enables the entity batchUpdate and batchUpsert (which is called for point correction) APIs for data repo integration. It does this by moving the implementation out of EntityService and into LocalEntityProvider, with appropriate changes to the EntityProvider trait. In detail:
* the batchUpsert and batchUpdate APIs now accept `dataReference` and `billingProject` query params. I have intentionally not updated swagger yet, since these params don't work fully yet.
* the `batchUpdateEntities()` and `batchUpsertEntities()` functions move to the `EntityProvider` trait and the `LocalEntityProvider` and `DataRepoEntityProvider` implementations
    * `DataRepoEntityProvider` is currently a no-op implementation that throws immediately; this will change in future PRs when we build the real functionality
* the utility function `applyOperationsToEntity()` moves from `EntityService` to `AttributeSupport`, where both `EntityService` and `LocalEntityProvider` can access it
* the `getWorkspaceContextAndPermissions` permission check for upserts/updates moves a bit higher in the chain, out of `batchUpdateEntitiesInternal` and into the top-level methods in `EntityService`. This mirrors other provider-enabled EntityService methods, and keeps `LocalEntityProvider` from having to query workspaces.

This illustrates the changes:
Entity Provider Flow

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -180,7 +180,6 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
   }
 
   def copyEntities(entityCopyDef: EntityCopyDefinition, uri: Uri, linkExistingEntities: Boolean): Future[PerRequestMessage] =
-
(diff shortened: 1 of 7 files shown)


ungwudik (approved):
Looks good to me.

MatthewBemis (approved):

PR #1422, opened 2021-04-23
pipe DeltaLayerWriter through to DataRepoEntityProvider [AS-586]

davidangb:
This is another foundational PR for AS-586 but does not come close to completing that ticket. Because this touches a lot of files I thought it would be easier to review on its own. This PR should not change any functionality.

In this PR:
* create a new `DeltaLayerWriter` trait with Gcs and Mock implementations
* initialize the `GcsDeltaLayerWriter` with a GCS storage client and the name of the Delta Layer bucket. This depends on broadinstitute/firecloud-develop#2518 which ~I will make sure to merge first~ is now merged.
* pass the `GcsDeltaLayerWriter` instance through EntityManager and DataRepoEntityProviderBuilder to DataRepoEntityProvider, so that a future PR can use it to write point corrections
* update tests to use the `MockDeltaLayerWriter` for syntax compatibility
* as of this PR, the `DeltaLayerWriter` implementations are pure noops, I will add functionality later.


davidangb:
jenkins retest

MatthewBemis:
jenkins retest ; looks like there were some issues downloading images from dockerhub

MatthewBemis (approved):
The plumbing looks good to me 👍

PR #1424, opened 2021-05-05
AS-721: Import Service/PFB Import should show errors to user on partial success

ansingh7115:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -379,9 +379,9 @@ class AvroUpsertMonitorActor(
         case Success(entityList) => entityList.size
       }.sum
 
-      val failureReports:List[RawlsErrorReport] = upsertResults collect {
-        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport.causes
-      } flatten
+      val failureReports: List[RawlsErrorReport] = upsertResults collect {
+        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport
+      }
 
       // this could be a LOT of error reports, we don't want to send an enormous packet back to the caller.
       // Cap the failure reports at 100.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -262,7 +262,7 @@ class AvroUpsertMonitorActor(
                 baseMsg
               else
                 baseMsg + s" First 100 failures are: $failureMessages"
-              publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Done, Option(msg))
+              publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(msg))
             case Failure(t) => publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(t.getMessage))
           }
         }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -379,9 +379,13 @@ class AvroUpsertMonitorActor(
         case Success(entityList) => entityList.size
       }.sum
 
+      // if the failure has underlying causes, use those; else use the parent failure.
+      // unresolved entity references only have useful error messages in the underlying causes, not the parent failure
+      // therefore to deliver the best messages to the user we need to handle both cases
       val failureReports: List[RawlsErrorReport] = upsertResults collect {
-        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport
-      }
+        case Failure(regrets:RawlsExceptionWithErrorReport) if regrets.errorReport.causes.nonEmpty => regrets.errorReport.causes
+        case Failure(regrets:RawlsExceptionWithErrorReport) => Seq(regrets.errorReport)
+      } flatten
 
       // this could be a LOT of error reports, we don't want to send an enormous packet back to the caller.
       // Cap the failure reports at 100.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -258,11 +258,12 @@ class AvroUpsertMonitorActor(
             case Success(importUpsertResults) =>
               val failureMessages = stringMessageFromFailures(importUpsertResults.failures, 100)
               val baseMsg = s"Successfully updated ${importUpsertResults.successes} entities; ${importUpsertResults.failures.size} updates failed."
-              val msg = if (importUpsertResults.failures.isEmpty)
-                baseMsg
-              else
-                baseMsg + s" First 100 failures are: $failureMessages"
-              publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(msg))
+              if (importUpsertResults.failures.isEmpty)
+                publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Done, Option(baseMsg))
+              else {
+                val msg = baseMsg + s" First 100 failures are: $failureMessages"
+                publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(msg))
+              }
             case Failure(t) => publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Error, Option(t.getMessage))
           }
         }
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -316,4 +316,82 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+//
+//    }
+
+    // upsert will fail; check that a pubsub message was published to set the import job to error.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      val statusMessages = Await.result(services.gpsDAO.pullMessages(importWriteSubscriptionName, 1), Duration.apply(10, TimeUnit.SECONDS))
+      assert(statusMessages.exists { msg =>
+        msg.attributes.get("importId").contains(importId1.toString) &&
+          msg.attributes.get("newStatus").contains("Error") &&
+          msg.attributes.get("action").contains("status")
```
davidangb: could you also check here for the presence of the "First 100 failures..." in the error message?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -316,4 +316,82 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+//        assertResult(upsertQuantity) { entitiesOfType.size }
+//        upsertRange foreach { idx =>
+//          val name = s"avro-entity-$idx"
+//          val entity = Entity(name, entityType, Map(AttributeName("default", "avro-attribute") -> AttributeString("foo")))
+//          val actual = entitiesOfType.find(_.name == name)
+//          assertResult(Some(entity)) { actual }
+//        }
+//      }
+//
+//    }
```
davidangb: delete instead of commenting out

davidangb (commented):
nice simple change!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -378,9 +379,9 @@ class AvroUpsertMonitorActor(
         case Success(entityList) => entityList.size
       }.sum
 
-      val failureReports:List[RawlsErrorReport] = upsertResults collect {
-        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport.causes
-      } flatten
+      val failureReports: List[RawlsErrorReport] = upsertResults collect {
+        case Failure(regrets:RawlsExceptionWithErrorReport) => regrets.errorReport
+      }
```
ansingh7115: This change is because sometimes there are no causes in the error report, which leads to the error report not being added at all. This does mean that we're not reporting back every cause of every error, but instead reporting the top level error message. 

Another solution might be to do both - add both the top level message and the underlying causes.

davidangb:
```suggestion
      // if the failure has underlying causes, use those; else use the parent failure.
      // unresolved entity references only have useful error messages in the underlying causes, not the parent failure
      // therefore to deliver the best messages to the user we need to handle both cases
      val failureReports: List[RawlsErrorReport] = upsertResults collect {
        case Failure(regrets:RawlsExceptionWithErrorReport) if regrets.errorReport.causes.nonEmpty => regrets.errorReport.causes
        case Failure(regrets:RawlsExceptionWithErrorReport) => Seq(regrets.errorReport)
      } flatten
```

your PR would cause a functional change, most importantly around unresolved entity references. The previous code would return a message like "entityType entityName not found", but your PR would make that less helpful, returning "Could not resolve some entity references". I think my suggestion resolves this for all cases.

NB: I suppose it would be even better if we could return the conjunction "Could not resolve some entity references: entityType entityName not found"!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -316,4 +322,44 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+        msg.attributes.get("importId").contains(importId1.toString) &&
+          msg.attributes.get("newStatus").contains("Error") &&
+          msg.attributes.get("action").contains("status") &&
+          msg.attributes.get("errorMessage").get.contains("Successfully updated 1000 entities; 1 updates failed. First 100 failures are: Invalid input")
+      })
+    }
+
+  }
+
 }
```
davidangb, line 412: here's a unit test to check for the contents of error messages on unresolved references:

```suggestion
}

 it should "bubble up useful error message if upserts result in partial failure" in withTestDataApiServices { services =>
    val timeout = 30000 milliseconds
    val interval = 250 milliseconds
    val importId1 = UUID.randomUUID()

    // add the imports and their statuses to the mock importserviceDAO
    val mockImportServiceDAO =  setUp(services)
    mockImportServiceDAO.imports += (importId1 -> ImportStatuses.ReadyForUpsert)

    val successfulBatch = createUpsertOpsList(upsertRange(1000))

    // failure creates an entity that refers to a non-existent entity
    val ref = AttributeEntityReference("test-type", "this-entity-does-not-exist")
    val op: AttributeUpdateOperation = AddUpdateAttribute(AttributeName.withDefaultNS("intentionallyBadReference"), ref)
    import spray.json._
    import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.AttributeUpdateOperationFormat
    implicit val attributeFormat: AttributeFormat = new AttributeFormat with TypedAttributeListSerializer
    val opJsonString = op.toJson.compactPrint

    val failureBatch = s"""{"name": "avro-entity-failure", "entityType": "failme", "operations": [$opJsonString]}"""
    val contents = makeOpsJsonString(successfulBatch :+ failureBatch)

    // Store upsert json file
    Await.result(googleStorage.createBlob(bucketName, GcsBlobName(importId1.toString), contents.getBytes()).compile.drain.unsafeToFuture(), Duration.apply(10, TimeUnit.SECONDS))

    // Make sure the file saved properly
    Await.result(googleStorage.unsafeGetBlobBody(bucketName, GcsBlobName(importId1.toString)).unsafeToFuture(), Duration.apply(10, TimeUnit.SECONDS))

    // Publish message on the request topic
    services.gpsDAO.publishMessages(importReadPubSubTopic, List(MessageRequest(importId1.toString, testAttributes(importId1))))

    // check if correct message was posted on request topic. This will start the upsert attempt.
    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
      assert(services.gpsDAO.receivedMessage(importReadPubSubTopic, importId1.toString, 1))
    }

    // upsert will fail; check that a pubsub message was published to set the import job to error.
    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
      val statusMessages = Await.result(services.gpsDAO.pullMessages(importWriteSubscriptionName, 1), Duration.apply(10, TimeUnit.SECONDS))
      assert(statusMessages.exists { msg =>
        msg.attributes.get("importId").contains(importId1.toString) &&
          msg.attributes.get("newStatus").contains("Error") &&
          msg.attributes.get("action").contains("status") &&
          msg.attributes.get("errorMessage").get.contains("Successfully updated 1000 entities; 1 updates failed. First 100 failures are: test-type this-entity-does-not-exist not found")
      })
    }
  }
```

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -316,4 +323,90 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+    }
+
+    // upsert will fail; check that a pubsub message was published to set the import job to error.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      val statusMessages = Await.result(services.gpsDAO.pullMessages(importWriteSubscriptionName, 1), Duration.apply(10, TimeUnit.SECONDS))
+      assert(statusMessages.exists { msg =>
+        msg.attributes.get("importId").contains(importId1.toString) &&
+          msg.attributes.get("newStatus").contains("Error") &&
+          msg.attributes.get("action").contains("status") &&
+          msg.attributes.get("errorMessage").get.contains("Successfully updated 1000 entities; 1 updates failed. First 100 failures are: Invalid input")
```
sagehen03: Intellij makes the suggestion below.  Test still seems to work afterward.
```suggestion
          msg.attributes.("errorMessage").contains("Successfully updated 1000 entities; 1 updates failed. First 100 failures are: Invalid input")
```

davidangb (approved):
:+1: assuming tests pass - I think that unit test failure is a fluke.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -316,4 +323,90 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+    }
+
+    // upsert will fail; check that a pubsub message was published to set the import job to error.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      val statusMessages = Await.result(services.gpsDAO.pullMessages(importWriteSubscriptionName, 1), Duration.apply(10, TimeUnit.SECONDS))
+      assert(statusMessages.exists { msg =>
+        msg.attributes.get("importId").contains(importId1.toString) &&
+          msg.attributes.get("newStatus").contains("Error") &&
+          msg.attributes.get("action").contains("status") &&
+          msg.attributes.get("errorMessage").get.contains("Successfully updated 1000 entities; 1 updates failed. First 100 failures are: test-type this-entity-does-not-exist not found")
```
sagehen03: Same comment as above and it looks like we can collapse all `msg.attributes.get(key)` to `msg.attributes(key)` so long as we're confident the key is present in the map which it seems like we are.  All of that being said, I don't think it's worth spending much time on this :)

sagehen03 (approved):
Looks good to me!

PR #1425, opened 2021-05-06
optimize workspaceAttributeQuery.rewriteAttrsAction [AS-751]

davidangb:
`workspaceAttributeQuery.rewriteAttrsAction` is called from `EntityComponent.save` and `WorkspaceComponent.save`, which means it exists in many places, including the batchUpsert API.

Changes to `workspaceAttributeQuery.rewriteAttrsAction` in this PR:
* Don't issue noop queries. If there are no attributes to delete, don't issue a `DELETE from table where false` statement. Same with updates; if there are no attributes to update, don't issue a statement at all. Inserts already bypassed the db.
* When adding new attributes to entities, insert those attributes into ENTITY_ATTRIBUTE and be done. Previously, we would insert them into ENTITY_ATTRIBUTE, then insert them into ENTITY_ATTRIBUTE_TEMP, then update ENTITY_ATTRIBUTE with the (exact same) value from ENTITY_ATTRIBUTE_TEMP.
* When updating pre-existing attribute values on an entity, only update those rows that actually changed. Previously, we would update _*every attribute*_ on the entity whether or not it changed, by inserting both the changed and unchanged attributes into ENTITY_ATTRIBUTE_TEMP and updating ENTITY_ATTRIBUTE based on what's in ENTITY_ATTRIBUTE_TEMP.

Consider a data table with 25 columns. If a user issued a point correction to one column, we now update 1 row instead of 25.

Test coverage seems generally good for high-level functionality - "can I save an entity?" for instance. I've added low-level targeted tests to verify the specific changes in this PR.

This does NOT optimize workflow outputs. That code uses a separate path, which I will look at in the future.

If you run this locally, I found it useful to toggle on logging for `slick.jdbc.JdbcBackend.statement` and `slick.jdbc.JdbcBackend.parameter`; see https://scala-slick.org/doc/3.3.3/config.html#logging

One more use case: consider a Terra data table of aliquots, with 2434 rows and 6 columns. As a user, I download a TSV of that table, change one value, and reupload the TSV. This is a not-uncommon use case, though users tend to edit more than one value in the TSV before re-uploading. As measured by Rawls:
* before this PR: 14,613 rows inserted to ENTITY_ATTRIBUTE_TEMP, 14,613 rows updated in ENTITY_ATTRIBUTE
* after this PR: 1 row inserted to ENTITY_ATTRIBUTE_TEMP, 1 rows updated in ENTITY_ATTRIBUTE

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -381,14 +381,6 @@ trait AttributeComponent {
             !existingAttributes.exists(equalRecords(_, v)) // if the attribute exists and is unchanged, don't update it
       }
 
-      // N.B. attributesToIgnore is only used for debugging/logging! TODO: delete this entire block when done debugging
-//      val attributesToIgnore = existingKeys -- attributesToInsert.keys -- attributesToUpdate.keys -- attributesToDelete.keys
-//
-//      System.err.println(s"********** CALCULATED attributesToInsert: ${attributesToInsert.keys.map(_.name).toList.sorted}")
-//      System.err.println(s"********** CALCULATED attributesToDelete: ${attributesToDelete.keys.map(_.name).toList.sorted}")
-//      System.err.println(s"********** CALCULATED attributesToUpdate: ${attributesToUpdate.keys.map(_.name).toList.sorted}")
-//      System.err.println(s"********** CALCULATED attributesToIgnore: ${attributesToIgnore.map(_.name).toList.sorted}")
-//
       patchAttributesAction(attributesToInsert.values,
         attributesToUpdate.values,
         attributesToDelete.values.map(_.id),
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -330,14 +330,8 @@ trait AttributeComponent {
 
     def patchAttributesAction(inserts: Traversable[RECORD], updates: Traversable[RECORD], deleteIds: Traversable[Long], insertFunction: Seq[RECORD] => String => WriteAction[Int]) = {
       for {
-        _ <- if (deleteIds.nonEmpty)
-                deleteAttributeRecordsById(deleteIds.toSeq)
-              else
-                DBIO.successful(0)
-        _ <- if (inserts.nonEmpty)
-                batchInsertAttributes(inserts.toSeq)
-              else
-                DBIO.successful(0)
+        _ <- if (deleteIds.nonEmpty) deleteAttributeRecordsById(deleteIds.toSeq) else DBIO.successful(0)
+        _ <- if (inserts.nonEmpty) batchInsertAttributes(inserts.toSeq) else DBIO.successful(0)
         updateResult <- if (updates.nonEmpty)
                           AlterAttributesUsingScratchTableQueries.updateAction(insertFunction(updates.toSeq))
                         else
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -338,14 +338,50 @@ trait AttributeComponent {
...
+      // insert attributes which are in save but not exists
+      val attributesToInsert = toSaveAttrMap.filterKeys(! existingKeys.contains(_))
+
+      // delete attributes which are in exists but not save
+      val attributesToDelete = existingAttrMap.filterKeys(! toSaveAttrMap.keySet.contains(_))
+
+      // update attributes which:are in both to-save and currently-exists, but have different values.
+      // note that currently-existing attributes will have a populated id e.g. "1234", but to-save will have an id of "0"
+      // therefore, create a collection of unmarshaled existing attributes to use for comparison, ignoring ids
+      val existingUnmarshalled = existingAttributes.toList.map(unmarshalValue)
```
MatthewBemis: `unmarshalValue` only unmarshals attribute values (string, boolean, number, json) so this will skip over other types like attribute entity references and i _think_ `AttributeNull` which explains both of the unit test failures

davidangb: ah, good catch. I can address that - this should be fixable!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -329,23 +329,62 @@ trait AttributeComponent {
...
+      val attributesToDelete = existingAttrMap.filterKeys(! toSaveAttrMap.keySet.contains(_))
+
+      // update attributes which are in both to-save and currently-exists, but have different values.
+      // note that currently-existing attributes will have a populated id e.g. "1234", but to-save will have an id of "0"
+      // therefore, use a comparison function that looks at everything except id
+      // note this does not compare transactionId for AttributeScratchRecords. We do not expect AttributeScratchRecords
+      // here, and transactionId will eventually be going away, so don't bother
+      // TODO: should this compare function move closer to the AttributeRecord class?
+      def equalRecords(left: RECORD, right: RECORD): Boolean = {
+        // compare everything except id
```
MatthewBemis, line 363: i was wondering why you hadn't done something like `left.copy(id=0).equals(right.copy(id=0))` but i see now that switching this to take an `AttributeRecord` to do that would probably be annoying and is _not_ worth the effort right now

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
```diff
@@ -726,4 +729,191 @@ class AttributeComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+    val existingAttributes = Seq(
+      baseRec.copy(id = 1, name = "attr1", valueString = Some("value1")),
+      baseRec.copy(id = 2, name = "attr2", valueString = None, valueNumber = Some(2)),
+      baseRec.copy(id = 3, name = "attr3", valueString = None, valueBoolean = Some(true)))
+
+    val spiedAttrQuery = spy(entityAttributeQuery)
+
+    runAndWait(spiedAttrQuery.rewriteAttrsAction(existingAttributes, existingAttributes, entityAttributeTempQuery.insertScratchAttributes))
+
+    val insertsCaptor: ArgumentCaptor[Traversable[EntityAttributeRecord]] = ArgumentCaptor.forClass(classOf[Traversable[EntityAttributeRecord]])
```
MatthewBemis, line 745: TIL about another power of mockito

MatthewBemis (approved):
LGTM

ungwudik (approved):
This looks like a great improvement! 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
davidangb: yeah - `copy` isn't even available on `AttributeRecord` (which is a trait); it's only available on the implementing case classes `EntityAttributeRecord`, `WorkspaceAttributeRecord`, etc. I considered a pattern-match here to work with those, but it felt more complicated and potentially less performant than just a swath of `==` clauses.

PR #1426, opened 2021-05-06
handleOutputs logging [AS-750]

davidangb:
helpful logging inside `handleOutputs` so we can quantify how often workflow outputs write to - and therefore lock - entity attr tables and/or workspace attr tables

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -403,7 +403,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
             else if (entityUpdates.nonEmpty)
               logger.info("handleOutputs writing to entity attributes only")
             else
-              logger.debug("handleOutputs writing to neither workspace nor entity attributes; could be errors")
+              logger.info("handleOutputs writing to neither workspace nor entity attributes; could be errors")
 
         // save everything to the db
         _ <- saveWorkspace(dataAccess, updatedEntitiesAndWorkspace)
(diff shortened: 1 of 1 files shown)


davidangb:
@ansingh7115 I know this is your ticket, I had this code locally and thought it might help!

ansingh7115 (approved):
Only downside here is that we are increasing the number of things we're doing inside of a transaction! But I don't think this should be a significant burden

rsasch (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -393,6 +393,18 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+        workspacesToUpdate = updatedEntitiesAndWorkspace.collect { case Left((_, Some(workspace))) => workspace }
+        entityUpdates = updatedEntitiesAndWorkspace.collect { case Left((Some(entityUpdate), _)) if entityUpdate.upserts.nonEmpty => entityUpdate }
+        _ = if (workspacesToUpdate.nonEmpty && entityUpdates.nonEmpty)
+              logger.info("handleOutputs writing to both workspace and entity attributes")
+            else if (workspacesToUpdate.nonEmpty)
+              logger.info("handleOutputs writing to workspace attributes only")
+            else if (entityUpdates.nonEmpty)
+              logger.info("handleOutputs writing to entity attributes only")
+            else
+              logger.debug("handleOutputs writing to neither workspace nor entity attributes; could be errors")
```
aednichols: Should be consistent? Or maybe even `error` since it's a possible error?
```suggestion
              logger.info("handleOutputs writing to neither workspace nor entity attributes; could be errors")
```

aednichols (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: my goal here was to not spam logs too much - I believe that errors are logged elsewhere, so I didn't want to add anything here

davidangb: buuut ... I think you're right, I'll make it info - even in the case where it doesn't write to entities or workspaces, it will still write to the db to save errors, that's good to track too

aednichols (approved):

davidangb:
`org.broadinstitute.dsde.test.api.AuthDomainSpec.AuthDomains should clone a workspace if the source has a multi-group auth domain and user is in all groups` failed in swatomation, which I cannot fathom is a result of this PR, but for due process: Jenkins retest

PR #1427, opened 2021-05-07
AS-744: Use Temp Tables for Workspace Attributes

sagehen03:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
```diff
@@ -38,14 +38,24 @@ class SlickDataSource(val databaseConfig: DatabaseConfig[JdbcProfile])(implicit
...
   // creates the ENTITY_ATTRIBUTE_TEMP for use by this transaction, executes the transaction, drops the temp table
   def inTransactionWithAttrTempTable[T](f: (DataAccess) => ReadWriteAction[T], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[T] = {
 
     val callerAction = f(dataAccess).transactionally.withTransactionIsolation(isolationLevel)
 
     val callerActionWithTempTables = (for {
       _ <- dropTempTable
+      _ <- dropWorkspaceAttributeTempTable
       _ <- createTempTable
+      _ <- createWorkspaceAttributeTempTable
```
davidangb: up for debate: should we have separate `inTransactionWithEntityAttrTempTable` and `inTransactionWithWorkspaceAttrTempTable` methods, and isolate the entity/workspace temp tables?

On one hand, it's wasteful to create the entity temp table when all you need is the workspace temp table, and vice versa - why knowingly ask the db to do extra work? Keeping these separate would also make the code more self-documenting - at the point of calling `dataSource.inTransactionWithSomeAttrTempTable` it would be clear what's happening.

On the other hand, this requires a proliferation of similar methods, and could be a premature optimization.

davidangb: That ^^ was 8am thinking. Slightly more awake: could the choice of which temp tables to create be an argument to this method? Create (or find+reuse) an enum or some constants to represent `ENTITY` and `WORKSPACE`, then pass those in:
```
dataSource.inTransactionWithAttrTempTable(ENTITY){ ... sql code here ... }
dataSource.inTransactionWithAttrTempTable(WORKSPACE){ ... sql code here ... }
dataSource.inTransactionWithAttrTempTable(ENTITY,WORKSPACE){ ... sql code here ... }
```
and conditionally create the relevant temp tables based on the arguments

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
```diff
@@ -523,11 +485,11 @@ class AttributeComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
     val insert = WorkspaceAttributeRecord(3, workspaceId, AttributeName.defaultNamespace, "test3", None, None, Option(false), None, None, None, None, deleted = false, None)
 
     //test insert and update
-    runAndWait(workspaceAttributeQuery.patchAttributesAction(Seq(insert), Seq(update), Seq(), workspaceAttributeScratchQuery.insertScratchAttributes))
+    runAndWait(workspaceAttributeQuery.patchAttributesAction(Seq(insert), Seq(update), Seq(), workspaceAttributeTempQuery.insertScratchAttributes))
     assertExpectedRecords(Seq(existing.head, update, insert):_*)
 
     //test delete
-    runAndWait(workspaceAttributeQuery.patchAttributesAction(Seq(), Seq(), existing.map(_.id), workspaceAttributeScratchQuery.insertScratchAttributes))
+    runAndWait(workspaceAttributeQuery.patchAttributesAction(Seq(), Seq(), existing.map(_.id), workspaceAttributeTempQuery.insertScratchAttributes))
```
davidangb, line 497: could you create a unit test for workspaces similar to the s"handle $parallelism simultaneous writes to their own entity attribute temp tables" test for entities? Do you think that's overkill?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
sagehen03: This seems like a good plan.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala
sagehen03: It is for sure an opportunity to learn some more scala so probably worth doing on that account alone.

ansingh7115:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
```diff
@@ -19,6 +19,10 @@ object DataSource {
   }
 }
 
+object AttributeTempTableType extends Enumeration {
```
ansingh7115, line 22: Are there ever times where we will want to use temp tables for both? For example, in `SubmissionMonitorActor`, we call `handleOutputs`, which can write to both workspace and entity attributes.

sagehen03: Yes, I think the code should do the right thing in that case.  `handleOuputs` is wrapped by  `datasource.inTransactionWithAttrTempTable` which should create both temp tables assuming `UncoordinatedDataSourceAccess` is used.  Is that a good assumption?

sagehen03:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
ansingh7115: Ah! Okay - I commented with only a partial understanding with what was going on! Sorry

sagehen03:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
sagehen03: No worries at all!

sagehen03:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
```diff
@@ -30,26 +34,46 @@ class SlickDataSource(val databaseConfig: DatabaseConfig[JdbcProfile])(implicit
...
     } yield {
-      origResult
-    }).andFinally(dropTempTable)
+      origResult match {
+        case scala.util.Success(s) => s
+        case scala.util.Failure(ex) => throw ex
+      }
+    }
+
+
```
davidangb: nit: lots of whitespace here

also, I suggest changing the log message at line 80 to something like:
```
logger.error(s"Transaction with temporary tables failed for (${tempTableTypes.mkString(",")}). Message: ${t.getMessage}")
```

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -81,7 +81,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
             }
             updateAction.map(RequestComplete(StatusCodes.OK, _))
           }
-        }
+        }, Set(AttributeTempTableType.Entity))
```
davidangb: looking for opinions ... should we stick with the `Entity`, `Workspace` after the database code, or go for a syntax something like:
```
dataSource.inTransactionWithAttrTempTable(Entity, Workspace) { dataAccess =>
  ...
}
```
wondering if either is easier to read/write/understand. In the few places we override `TransactionIsolation`, that comes after the database code.

davidangb (commented):
two comments inline, I am going to think hard about tests ... and that said, have you looked at
https://github.com/broadinstitute/rawls/blob/992a08d46d1f1fe698cd7226cb30d8374033ddee/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala#L806 ? I assume that test has been passing ...

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
ansingh7115: I would vote for the latter. I think if I was scanning through code, having them together indicates they're together. I've sometimes thought we should change TransactionIsolation to come before the block, too, but that's out of scope.

sagehen03: I don't have an opinion, and David it sounds like you like the change enough to throw it out for feedback--seems like 1.5 votes in favor of the change with none against--I'll make the change.

sagehen03:
> two comments inline, I am going to think hard about tests ... and that said, have you looked at
> 
> https://github.com/broadinstitute/rawls/blob/992a08d46d1f1fe698cd7226cb30d8374033ddee/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala#L806
> ? I assume that test has been passing ...

That test uses code that's only invoked in tests to repeatedly retry upon a RawlsConcurrentModificationException so I don't feel like it's a good test to verify that things will work as expected in production.

davidangb (approved):
:+1:. I (like you) am struggling to figure out a useful, effective test for this above and beyond the test coverage we already have. I'm pretty confident in the implementation, but I suggest merging it just after a weekly release, to give it the most burn-in time in dev/alpha/perf before it heads to prod.

ansingh7115 (approved):

PR #1428, opened 2021-05-07
[CA-1331][CA-1332] Fix a couple bugs in the workspace billing account monitor

marctalbott:
Ticket: [CA-1331](https://broadworkbench.atlassian.net/browse/CA-1331), [CA-1332](https://broadworkbench.atlassian.net/browse/CA-1332)

This PR does a few things (including two out of scope things listed at the bottom that I found while making the primary changes):

- Adds error message populated when the billing account cannot be updated on a workspace's Google project. This allows us to not retry endlessly when there is a problem. A user can always manually retry by sending another request
- Makes the monitor compatible with v1 and v2 workspaces by updating the sql query to include v1 workspaces and return their Google project id (that of the Terra billing project) and deduplicating the results to only send a single updateBillingAccount request per v1 workspace
- removes the unique constraint on the `google_project_number` column in the `WORKSPACE` table because v1 workspaces will share a Google project and have identical Google project numbers as a result
- switches update billing account endpoints to `/billingAccount` instead of `/billing-account` for consistency with the rest of Rawls

  - [ ] **...and Orchestration's Swagger too!** TODO!
- [] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20201106_workspace_google_project_columns.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20201106_workspace_google_project_columns.xml
@@ -5,7 +5,7 @@
         <renameColumn tableName="WORKSPACE" oldColumnName="GOOGLE_PROJECT" newColumnName="GOOGLE_PROJECT_ID" columnDataType="VARCHAR(254)" />
         <addColumn tableName="WORKSPACE">
             <column name="GOOGLE_PROJECT_NUMBER" type="VARCHAR(254)">
-                <constraints nullable="true" unique="true" />
+                <constraints nullable="true"/>
             </column>
         </addColumn>
         <sql>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -328,11 +328,11 @@ trait WorkspaceComponent {
       findByGoogleProjectIdQuery(googleProjectId).map(_.currentBillingAccountOnGoogleProject).update(newBillingAccount.map(_.value))
     }
 
-    def updateWorkspaceBillingAccountErrorMessage(googleProjectId: GoogleProjectId, errorMessage: String): WriteAction[Int] = {
+    def updateWorkspaceBillingAccountErrorMessages(googleProjectId: GoogleProjectId, errorMessage: String): WriteAction[Int] = {
       findByGoogleProjectIdQuery(googleProjectId).map(_.billingAccountErrorMessage).update(Option(errorMessage))
     }
 
-    def deleteWorkspaceBillingAccountErrorMessage(namespace: RawlsBillingProjectName): WriteAction[Int] = {
+    def deleteAllWorkspaceBillingAccountErrorMessagesInBillingProject(namespace: RawlsBillingProjectName): WriteAction[Int] = {
       findByNamespaceQuery(namespace).map(_.billingAccountErrorMessage).update(None)
     }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -26,14 +26,15 @@ case class WorkspaceRecord(
                             workflowCollection: Option[String],
                             createdDate: Timestamp,
                             lastModified: Timestamp,
-  entityCacheLastUpdated: Timestamp,
+                            entityCacheLastUpdated: Timestamp,
                             createdBy: String,
                             isLocked: Boolean,
                             recordVersion: Long,
                             workspaceVersion: String,
                             googleProjectId: String,
                             googleProjectNumber: Option[String],
-                            currentBillingAccountOnGoogleProject: Option[String]) {
+                            currentBillingAccountOnGoogleProject: Option[String],
+                            billingAccountErrorMessage: Option[String]) {
   def toWorkspaceName: WorkspaceName = WorkspaceName(namespace, name)
 }
 
@@ -64,11 +65,12 @@ trait WorkspaceComponent {
     def workspaceVersion = column[String]("workspace_version")
     def googleProjectId = column[String]("google_project_id")
     def googleProjectNumber = column[Option[String]]("google_project_number")
+    def currentBillingAccountOnGoogleProject = column[Option[String]]("billing_account_on_google_project", O.Length(254))
+    def billingAccountErrorMessage = column[Option[String]]("billing_account_error_message")
 
     def uniqueNamespaceName = index("IDX_WS_UNIQUE_NAMESPACE_NAME", (namespace, name), unique = true)
 
-    def currentBillingAccountOnGoogleProject = column[Option[String]]("billing_account_on_google_project", O.Length(254))
-    def * = (namespace, name, id, bucketName, workflowCollection, createdDate, lastModified, entityCacheLastUpdated, createdBy, isLocked, recordVersion, workspaceVersion, googleProjectId, googleProjectNumber, currentBillingAccountOnGoogleProject) <> (WorkspaceRecord.tupled, WorkspaceRecord.unapply)
+    def * = (namespace, name, id, bucketName, workflowCollection, createdDate, lastModified, entityCacheLastUpdated, createdBy, isLocked, recordVersion, workspaceVersion, googleProjectId, googleProjectNumber, currentBillingAccountOnGoogleProject, billingAccountErrorMessage) <> (WorkspaceRecord.tupled, WorkspaceRecord.unapply)
   }
 
   /** raw/optimized SQL queries for working with workspace attributes
@@ -307,22 +309,31 @@ trait WorkspaceComponent {
 
       loadWorkspaces(workspaces)
     }
-    def listWorkspaceGoogleProjectsToUpdateWithNewBillingAccount(): ReadAction[Seq[(GoogleProjectId, Option[RawlsBillingAccountName])]] = {
+
+    def listWorkspaceGoogleProjectsToUpdateWithNewBillingAccount(): ReadAction[Set[(GoogleProjectId, Option[RawlsBillingAccountName])]] = {
       val query = for {
         billingProject <- rawlsBillingProjectQuery if !billingProject.invalidBillingAccount
         workspace <- workspaceQuery if workspace.namespace === billingProject.projectName &&
-          workspace.workspaceVersion === WorkspaceVersions.V2.value &&
+          workspace.billingAccountErrorMessage.isEmpty &&
           !(workspace.currentBillingAccountOnGoogleProject === billingProject.billingAccount ||
             (workspace.currentBillingAccountOnGoogleProject.isEmpty && billingProject.billingAccount.isEmpty))
       } yield (workspace.googleProjectId, billingProject.billingAccount)
       query.result.map(results => results.map {
         case (googleProjectId, newBillingAccount) =>
           (GoogleProjectId(googleProjectId), newBillingAccount.map(RawlsBillingAccountName))
-      })
+      }.toSet)
     }
 
     def updateWorkspaceBillingAccount(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): WriteAction[Int] = {
       findByGoogleProjectIdQuery(googleProjectId).map(_.currentBillingAccountOnGoogleProject).update(newBillingAccount.map(_.value))
+    }
+
+    def updateWorkspaceBillingAccountErrorMessage(googleProjectId: GoogleProjectId, errorMessage: String): WriteAction[Int] = {
+      findByGoogleProjectIdQuery(googleProjectId).map(_.billingAccountErrorMessage).update(Option(errorMessage))
+    }
+
+    def deleteWorkspaceBillingAccountErrorMessage(namespace: RawlsBillingProjectName): WriteAction[Int] = {
+      findByNamespaceQuery(namespace).map(_.billingAccountErrorMessage).update(None)
     }
 
     /**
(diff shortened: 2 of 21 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -597,6 +594,18 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
 
   }
 
+  //todo: i don't know what to call this, soliciting recommendations -- updateBillingAccountInternalInternal (^2?)
```
marctalbott: Looking for help on the name of this function please!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20201106_workspace_google_project_columns.xml
```diff
@@ -5,7 +5,7 @@
         <renameColumn tableName="WORKSPACE" oldColumnName="GOOGLE_PROJECT" newColumnName="GOOGLE_PROJECT_ID" columnDataType="VARCHAR(254)" />
         <addColumn tableName="WORKSPACE">
             <column name="GOOGLE_PROJECT_NUMBER" type="VARCHAR(254)">
-                <constraints nullable="true" unique="true" />
+                <constraints nullable="true"/>
```
gpolumbo-broad, line 8: Can we make sure we put this constraint back on during the PPW migration.  I'm fine tacking it onto an existing ticket or create its own ticket, just don't want to forget it.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitor.scala
```diff
@@ -45,11 +45,15 @@ class WorkspaceBillingAccountMonitor(datasource: SlickDataSource, gcsDAO: Google
     for {
       _ <- gcsDAO.updateGoogleProjectBillingAccount(googleProjectId, newBillingAccount).recoverWith {
         case e: RawlsExceptionWithErrorReport if e.errorReport.statusCode == Option(StatusCodes.Forbidden) && newBillingAccount.isDefined =>
-          datasource.inTransaction( { dataAccess =>
+          dataSource.inTransaction( { dataAccess =>
             dataAccess.rawlsBillingProjectQuery.updateBillingAccountValidity(newBillingAccount.get, isInvalid = true)
           }).map(_ => throw e)
+        case e =>
+          dataSource.inTransaction { dataAccess =>
+            dataAccess.workspaceQuery.updateWorkspaceBillingAccountErrorMessage(googleProjectId, e.getMessage)
```
gpolumbo-broad: Let's log the failure as well.  Yes, the message is in the database, but since it can be cleared/overwritten, it might be useful to be able to check the logs for the failure message too if we're trying to debug something like intermittent failures or who knows what.  

As a general practice, I think it's a good idea to start with always logging when you're handling an exception.  Then if you can convince yourself that a the log message is not needed for this scenario, then you can remove it.  I may even be inclined to add a log statement at line 49 too in this case.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -307,24 +309,33 @@ trait WorkspaceComponent {
...
-      })
+      }.toSet)
     }
 
     def updateWorkspaceBillingAccount(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): WriteAction[Int] = {
       findByGoogleProjectIdQuery(googleProjectId).map(_.currentBillingAccountOnGoogleProject).update(newBillingAccount.map(_.value))
     }
 
+    def updateWorkspaceBillingAccountErrorMessage(googleProjectId: GoogleProjectId, errorMessage: String): WriteAction[Int] = {
+      findByGoogleProjectIdQuery(googleProjectId).map(_.billingAccountErrorMessage).update(Option(errorMessage))
```
gpolumbo-broad, line 332: I think this is only right for v1 workspaces, no?  This finds all the Workspaces in a project and then then sets them all to have the same error message.  That will be problematic for v2 Workspaces.  

If the reason why it errored is because Terra doesn't have the right permissions on the Billing Account, then yeah, probably all the v2 Workspaces will end up with the same error.  But I think we need to allow that there could be other reasons why the Billing Account change could fail for a particular Google Project (v2 Workspace) and we need to be able to be specific to a Workspace/Google Project.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
gpolumbo-broad: `updateBillingAccountInDb`?  Or `updateBillingProjectsBillingAccount`?  My preference is that if it ends up lengthy, that's ok, so long as it is clear.  Long and clear/expressive is better than short and possibly vague/ambiguous.  I would also generally try to avoid abbreviations in most cases for the same reason.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -307,24 +309,33 @@ trait WorkspaceComponent {
...
           (GoogleProjectId(googleProjectId), newBillingAccount.map(RawlsBillingAccountName))
-      })
+      }.toSet)
     }
 
     def updateWorkspaceBillingAccount(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): WriteAction[Int] = {
       findByGoogleProjectIdQuery(googleProjectId).map(_.currentBillingAccountOnGoogleProject).update(newBillingAccount.map(_.value))
     }
 
+    def updateWorkspaceBillingAccountErrorMessage(googleProjectId: GoogleProjectId, errorMessage: String): WriteAction[Int] = {
```
gpolumbo-broad: naming nit for this and `deleteWorkspaceBillingAccountErrorMessage` that you pluralize them if they remain written as they are.  So:

`updateWorkspaceBillingAccountErrorMessage` -> `updateWorkspaceBillingAccountErrorMessages`

It's subtle, but I think it makes it clear that you're going be updating the `errorMessages` for many Workspaces, not just one.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -551,8 +552,9 @@ case class WorkspaceDetails(namespace: String,
                             workspaceVersion: WorkspaceVersion,
                             googleProjectId: GoogleProjectId,
                             googleProjectNumber: Option[GoogleProjectNumber],
-                            billingAccount: Option[RawlsBillingAccountName]) {
-  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProjectId, googleProjectNumber, billingAccount)
+                            billingAccount: Option[RawlsBillingAccountName],
+                            billingAccountErrorMessage: Option[String]) {
+  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProjectId, googleProjectNumber, billingAccount, billingAccountErrorMessage) // todo: do we just add the error message on here? what does that mean for the api?
```
gpolumbo-broad, line 557: I think it should be OK since it's just adding a new field to the returned data, which I think is a backwards compatible API change.

We should add it to the schema definition in `api-docs.yaml` for the `WorkspaceDetails`: https://github.com/broadinstitute/rawls/blob/406d6101654222980f1424767fb0188c162d7a01/core/src/main/resources/swagger/api-docs.yaml#L5267

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      runAndWait(workspaceQuery.createOrUpdate(secondV1Workspace))
+      runAndWait(rawlsBillingProjectQuery.updateBillingAccount(billingProject.projectName, Option(newBillingAccount)))
+
+      val actor = createWorkspaceBillingAccountMonitor(dataSource, failingGcsDAO)
+
+      eventually (timeout = timeout(Span(10, Seconds))) {
+        runAndWait(workspaceQuery.findByName(secondV1Workspace.toWorkspaceName)).getOrElse(fail("workspace not found"))
+          .billingAccountErrorMessage shouldBe Option(failureMessage)
+        runAndWait(workspaceQuery.findByName(v1Workspace.toWorkspaceName)).getOrElse(fail("workspace not found"))
+          .billingAccountErrorMessage shouldBe Option(failureMessage)
```
gpolumbo-broad, line 138: And I think add one more check:  

`runAndWait(workspaceQuery.findByName(v2Workspace.toWorkspaceName)).getOrElse(fail("workspace not found"))
          .billingAccountErrorMessage shouldBe None`

This is based on my comment above about how I think v2 Workspaces should get their own specific message.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+
+      runAndWait(workspaceQuery.createOrUpdate(v1Workspace))
+      runAndWait(workspaceQuery.createOrUpdate(v2Workspace))
+      runAndWait(rawlsBillingProjectQuery.updateBillingAccount(billingProject.projectName, Option(newBillingAccount)))
+
+      val actor = createWorkspaceBillingAccountMonitor(dataSource, failingGcsDAO)
+
+      eventually (timeout = timeout(Span(10, Seconds))) {
+        runAndWait(rawlsBillingProjectQuery.load(billingProject.projectName)).getOrElse(fail("project not found"))
+          .invalidBillingAccount shouldBe true
```
gpolumbo-broad, line 173: In this scenario, we do not set the `billingAccountErrorMessage`s on the Workspaces, so I think we should add another assertion to collect the error messages on the Billing Project's workspaces and confirm that they're all blank.

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
marctalbott: > This finds all the Workspaces in a **Google** project and then then sets them all to have the same error message. That will be problematic for v2 Workspaces.

This should work for both v1 and v2 workspaces since it finds workspaces by their Google project id, not Terra billing project

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -307,24 +309,33 @@ trait WorkspaceComponent {
...
+      findByGoogleProjectIdQuery(googleProjectId).map(_.billingAccountErrorMessage).update(Option(errorMessage))
+    }
+
+    def deleteWorkspaceBillingAccountErrorMessage(namespace: RawlsBillingProjectName): WriteAction[Int] = {
+      findByNamespaceQuery(namespace).map(_.billingAccountErrorMessage).update(None)
+    }
```
andy7i: it seems like this clears out the billingAccountErrorMessage for all workspaces in a billing project. Why would we do it this way vs. clearing it out for each individual workspaces?

Would it be possible for a billing project with 5 workspaces to have 3 that succeeded and 2 that failed?

andy7i: I guess to answer my own question, we use this only when updating the billing account on a billing project, so it makes sense that we want to clear out the error messages on all workspaces in the billing project. 

Maybe we should rename the function name, maybe deleteAllWorkspaceBillingAccountErrorMessagesInBillingProject

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -597,6 +594,18 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+  private def doDatabaseThings(projectName: RawlsBillingProjectName, billingAccount: Option[RawlsBillingAccountName]): Future[Option[RawlsBillingProject]] = {
```
andy7i: nit: projectName => billingProjectName ?

but... it looks like it's used all over this file so I'll leave it up to you whether we should make any changes.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+import org.scalatest.time.{Seconds, Span}
+import org.scalatest.{BeforeAndAfterAll, OptionValues}
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.concurrent.ExecutionContext.global
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.language.postfixOps
+
+class WorkspaceBillingAccountMonitorSpec(_system: ActorSystem) extends TestKit(_system) with MockitoSugar with AnyFlatSpecLike with Matchers with TestDriverComponent with BeforeAndAfterAll with Eventually with OptionValues {
```
andy7i, line 28: nice

andy7i (approved):
👍  with comments addressed

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20201106_workspace_google_project_columns.xml
marctalbott: Sure! Added a ticket to the cleanup epic https://broadworkbench.atlassian.net/browse/CA-1346

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
marctalbott: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
marctalbott: 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: might as well do it right and move us in the right direction

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: Agreed, thanks!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitorSpec.scala
marctalbott: Good call!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitorSpec.scala
marctalbott: 👍

gpolumbo-broad (approved):

PR #1432, opened 2021-05-13
AS-728: Don't use deprecated WSM APIs in Rawls

ansingh7115:
This PR's changes are:
- Updating the version of Workspace Manager in Dependencies.scala
- Changing HttpWorkspaceManagerDAO.scala to use the new APIs. These endpoints used to return `DataReferenceDescription` and `DataReferenceList`, which have been replaced with `DataRepoSnapshotResource` and `ResourceList` respectively 
- V2 routes are added to SnapshotApiService.scala. Both V1 and V2 routes call the same methods, but V1 routes convert the response into the old, expected response objects.
- SnapshotApiServiceSpec now has two sets of tests for V1 and V2 routes
- New swagger for V2 routes
- The rest of it is test fallout

PR #1435, opened 2021-05-13
Write point correction to GCS for Delta Layer [AS-769]

davidangb:
Objective of this PR, for AS-769: complete a spike that connects a user's API call for a point correction all the way to writing the point correction file to the Delta Layer source bucket. 

Out of scope, targeted for AS-770: populating that file with all the necessary data and validating the point correction's contents.

The API underlying point corrections is [batchUpsert](https://rawls.dsde-dev.broadinstitute.org/#/entities/batch_upsert_entities). We should also tackle [batchUpdate](https://rawls.dsde-dev.broadinstitute.org/#/entities/batch_update_entities) at some point because they are so similar, but I also considered that out of scope for this PR.

Reviewer: if you want to run this locally, I suggest adding
```
        - $ref: '#/components/parameters/dataReferenceQueryParam'
```
to the swagger definition for the batchUpsert API; this will allow you to easily test batchUpsert against a snapshot reference. I have not updated swagger as part of this PR in order to keep this functionality somewhat obscured, since it is incomplete.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
@@ -41,6 +41,10 @@ class GcsDeltaLayerWriter(val storageService: GoogleStorageService[IO],
       // file multiple times?
       case t: HttpResponseException =>
         logger.warn(s"encountered error [${t.getStatusMessage}] with status code [${t.getStatusCode}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
+        throw t
+      case e: Exception =>
+        logger.warn(s"encountered error [${e.getMessage}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
+        throw e
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
@@ -1,12 +1,46 @@
 package org.broadinstitute.dsde.rawls.deltalayer
 
 import cats.effect.IO
-import org.broadinstitute.dsde.workbench.google2.GoogleStorageService
+import fs2._
+import org.broadinstitute.dsde.rawls.model.DeltaInsert
+import org.broadinstitute.dsde.workbench.google2.{GcsBlobName, GoogleStorageService}
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
+import spray.json._
+import spray.json.DefaultJsonProtocol._
 
 class GcsDeltaLayerWriter(val storageService: GoogleStorageService[IO],
                           val sourceBucket: GcsBucketName)
   extends DeltaLayerWriter {
 
+  // calculate the GCS path to which we should save the file
+  private[deltalayer] def filePath(writeObject: DeltaInsert): GcsBlobName = {
+    // workspace/${workspaceId}/reference/${referenceId}/insert/${insertId}.json
+    GcsBlobName(s"workspace/${writeObject.destination.workspaceId}/reference/${writeObject.destination.referenceId}/insert/${writeObject.insertId}.json")
+  }
+
+  // generate the string representation (json) of the file
+  private[deltalayer] def serializeFile(writeObject: DeltaInsert): String = {
+    val updateJson = writeObject.inserts.toJson.prettyPrint
+
+    // TODO: this should use real JsonFormat classes! For now, since the model is in such flux, just hand-roll
+    // the json
+    s"""{
+       |  "insertId": "${writeObject.insertId}",
+       |  "workspaceId": "${writeObject.destination.workspaceId}",
+       |  "referenceId": "${writeObject.destination.referenceId}",
+       |  "insertTimestamp": "${writeObject.insertTimestamp.toString}",
+       |  "insertingUser": "${writeObject.insertingUser.value}",
+       |  "inserts": $updateJson
+       |}
+       |""".stripMargin.parseJson.prettyPrint
+  }
+
+  override def writeFile(writeObject: DeltaInsert): Unit = {
+    // TODO: set overwrite value and traceId value
+    val writePipe = storageService.streamUploadBlob(sourceBucket, filePath(writeObject))
+    val fileContents = serializeFile(writeObject)
+    // TODO: retries and error-handling
+    text.utf8Encode(Stream.emit(fileContents)).through(writePipe).compile.drain.unsafeRunSync()
+  }
 
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.rawls.deltalayer
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.Uri
 import cats.effect.IO
 import com.google.api.client.http.HttpResponseException
 import fs2._
@@ -13,7 +14,7 @@ import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 
 class GcsDeltaLayerWriter(val storageService: GoogleStorageService[IO],
                           val sourceBucket: GcsBucketName,
@@ -23,27 +24,33 @@ class GcsDeltaLayerWriter(val storageService: GoogleStorageService[IO],
 
   implicit val service: GoogleInstrumentedService.Value = GoogleInstrumentedService.Storage
 
-  override def writeFile(writeObject: DeltaInsert): Unit = {
+  override def writeFile(writeObject: DeltaInsert): Future[Uri] = {
 
     val destinationPath = filePath(writeObject)
     val fileContents = serializeFile(writeObject)
 
     retryWithRecoverWhen500orGoogleError(() => {
+      // create the destination pipe to which we will write the file
       // set traceId equal to insertId for easy correlation
       val writePipe = storageService.streamUploadBlob(sourceBucket, destinationPath,
-        overwrite = false,
         traceId = Some(TraceId(writeObject.insertId)))
 
-      text.utf8Encode(Stream.emit(fileContents)).through(writePipe).compile.drain.unsafeRunSync()
+      // stream the file contents to the destination GcsBlob pipe
+      text.utf8Encode(Stream.emit(fileContents)).through(writePipe)
+        .compile.drain.unsafeRunSync()
+
+      // return Uri to the file we just wrote
+      Uri.apply(s"gs://${sourceBucket.value}/${destinationPath.value}")
+
     }) {
-      // are there any special error-handling cases we should include? Maybe 409 in case we've written the
-      // file multiple times?
       case t: HttpResponseException =>
+        // are there any special error-handling cases we should include? Maybe 409 in case we've written the
+        // file multiple times?
         logger.warn(s"encountered error [${t.getStatusMessage}] with status code [${t.getStatusCode}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
         throw t
-      case e: Exception =>
-        logger.warn(s"encountered error [${e.getMessage}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
-        throw e
+      case t: Throwable =>
+        logger.warn(s"encountered error [${t.getMessage}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
+        throw t
     }
   }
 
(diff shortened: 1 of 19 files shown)


project/Dependencies.scala
```diff
@@ -101,7 +101,7 @@ object Dependencies {
   val workbenchGoogleMocks: ModuleID =  workbenchGoogleExcludes("org.broadinstitute.dsde.workbench" %% "workbench-google" % workbenchGoogleV % "test" classifier "tests")
   val workbenchGoogle2: ModuleID =      workbenchGoogleExcludes("org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V)
   val workbenchGoogle2Tests: ModuleID = workbenchGoogleExcludes("org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V % "test" classifier "tests")
-  val googleStorageLocal: ModuleID = "com.google.cloud" % "google-cloud-nio" % "0.111.0-alpha" % "test"
+  val googleStorageLocal: ModuleID = "com.google.cloud" % "google-cloud-nio" % "0.122.11" % "test"
```
davidangb, line 104: this upgrade is necessary to get the underlying support for workbench-libs' `streamUploadBlob` method

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
```diff
@@ -1,12 +1,80 @@
...
+        logger.warn(s"encountered error [${t.getMessage}] when writing delta file [${sourceBucket.value}/${destinationPath.value}]")
+        throw t
+    }
+  }
+
+  // calculate the GCS path to which we should save the file
+  private[deltalayer] def filePath(writeObject: DeltaInsert): GcsBlobName = {
+    // workspace/${workspaceId}/reference/${referenceId}/insert/${insertId}.json
+    GcsBlobName(s"workspace/${writeObject.destination.workspaceId}/reference/${writeObject.destination.referenceId}/insert/${writeObject.insertId}.json")
+  }
```
davidangb, line 60: Delta Layer doesn't care about the file path; I added the workspace/reference/insert subdirectories to make it easier for humans to debug.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -213,6 +214,9 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
                           }
       } yield {
         entities
+      }) recover {
+        case dle: DeltaLayerException =>
+          throw new RawlsExceptionWithErrorReport(ErrorReport(dle.code, dle.getMessage, dle))
```
davidangb, line 219: without this `recover`, any DeltaLayerExceptions will be obscured and their status codes dropped; all exceptions would result in 500s to the user

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -409,6 +414,36 @@ class DataRepoEntityProvider(snapshotModel: SnapshotModel, requestArguments: Ent
...
+    // create DeltaInsert object
+    // TODO AS-770: populate with everything we need, including model versioning
+    val dest = Destination(workspaceId = requestArguments.workspace.workspaceIdAsUUID,
+                           referenceId = dataReference.getReferenceId)
+    val ins = DeltaInsert(version = "v0",
+                          insertId = UUID.randomUUID(),
+                          insertTimestamp = new DateTime(),
+                          insertingUser = requestArguments.userInfo.userSubjectId,
+                          destination = dest,
+                          inserts = entityUpdates)
```
davidangb, line 431: I expect the `Destination` and `DeltaInsert` classes to change in AS-770; I didn't put too much thought into them for this PR.

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriterSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+  implicit val actorSystem: ActorSystem = ActorSystem("GcsDeltaLayerWriterSpec")
+
+  behavior of "GcsDeltaLayerWriter"
+
+  it should "write the specified file to the bucket" in {
+    val bucket = GcsBucketName("successful-write")
+
+    // explicitly create the storage so we can query it directly
+    val storage = LocalStorageHelper.getOptions.getService
+    val deltaLayerWriter = getGcsWriter(bucket, Some(storage))
```
davidangb, line 37: `LocalStorageHelper` is provided by Google to assist with testing; it mocks the real cloud-based storage. See https://javadoc.io/doc/com.google.cloud/google-cloud-nio/latest/com/google/cloud/storage/contrib/nio/testing/LocalStorageHelper.html

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriterSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    val expected = GcsBlobName(s"workspace/${workspaceId}/reference/${referenceId}/insert/${insertId}.json")
+
+    val actual = deltaLayerWriter.filePath(testInsert)
+
+    actual shouldBe expected
+
+  }
+
+  it should "serialize DeltaInsert correctly" is (pending) // NB not worth testing until model class is stable
```
davidangb, line 116: implement this test as part of AS-770, once model classes are defined

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsStorageTestSupport.scala
```diff
@@ -0,0 +1,29 @@
...
+    val db = storage.getOrElse(LocalStorageHelper.getOptions().getService())
+    val blocker = Blocker.liftExecutionContext(testExecutionContext)
+    val semaphore = Semaphore[IO](1).unsafeRunSync
+
+    val localStorage = GoogleStorageInterpreter[IO](db, blocker, Some(semaphore))
+
+    new GcsDeltaLayerWriter(localStorage, bucket, "metricsPrefix")
+  }
+
+}
```
davidangb, line 29: most of this implementation lifted from workbench-libs: https://github.com/broadinstitute/workbench-libs/blob/72f888e21f24f6cb2e542c0982a0e05df5488a94/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreterSpec.scala#L176

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/MockDeltaLayerWriter.scala
```diff
@@ -1,5 +1,10 @@
 package org.broadinstitute.dsde.rawls.deltalayer
+import akka.http.scaladsl.model.Uri
+import org.broadinstitute.dsde.rawls.model.DeltaInsert
 
-class MockDeltaLayerWriter extends DeltaLayerWriter {
+import scala.concurrent.Future
 
+class MockDeltaLayerWriter extends DeltaLayerWriter {
+  // TODO: flesh out for unit tests, ideally using a mocked Google storage service
+  override def writeFile(writeObject: DeltaInsert): Future[Uri] = ???
```
davidangb, line 9: there exist spots in the test code that need a `DeltaLayerWriter` in a constructor, but don't ever use it - so here's a mock with zero implementation

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriterSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    // calculate expected file path and contents
+    val expectedPath = deltaLayerWriter.filePath(testInsert)
+    val expectedContents = deltaLayerWriter.serializeFile(testInsert)
+
+    // write the object via Delta Layer
+    deltaLayerWriter.writeFile(testInsert) map { _ =>
+      // confirm we wrote the file correctly. The LocalStorageHelper in-memory storage engine
+      // seems to have some delays, so use eventually to allow for delays and retries
+      eventually (timeout(Span(30, Seconds)), interval(Span(500, Millis))) {
```
ungwudik, line 53: Is there any guidance (or any general info) for how long this usually takes? Wondering if there's room for this step to potentially become flaky as-is.

ungwudik (approved):
Looks good to me. But it looks like a PR that could use at least 2 reviews.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriter.scala
```diff
@@ -1,12 +1,80 @@
 package org.broadinstitute.dsde.rawls.deltalayer
 
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.{StatusCodes, Uri}
 import cats.effect.IO
-import org.broadinstitute.dsde.workbench.google2.GoogleStorageService
+import com.google.api.client.http.HttpResponseException
```
ansingh7115: Remove unused imports

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -18,18 +20,21 @@ import org.broadinstitute.dsde.rawls.entities.exceptions.{DataEntityException, U
...
 import spray.json.{JsArray, JsBoolean, JsNull, JsNumber, JsObject, JsString, JsValue}
 
+import java.util.UUID
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class DataRepoEntityProvider(snapshotModel: SnapshotModel, requestArguments: EntityRequestArguments,
+class DataRepoEntityProvider(snapshotModel: SnapshotModel, dataReference: DataReferenceDescription,
```
ansingh7115: Per my PR, `DataReferenceDescription` is going to be replaced with `DataRepoSnapshotResource`. I think here, instead of passing in `DataReferenceDescription`, we could pass in `MetadataResource`, once we switch over.

In terms of merge order, I think we should probably merge AS-728 first, what do you think?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -409,6 +414,36 @@ class DataRepoEntityProvider(snapshotModel: SnapshotModel, requestArguments: Ent
...
+      case se: StorageException =>
+        val throwCode = StatusCodes.getForKey(se.getCode).getOrElse(StatusCodes.InternalServerError)
+        throw new DeltaLayerException(s"StorageException in Delta Layer: ${se.getMessage}", se, throwCode)
+      case t: Throwable =>
+        throw new DeltaLayerException(s"Error in Delta Layer: ${t.getClass.getName}: ${t.getMessage}", t)
+    }
+    // This method signature claims to return Traversable[Entity] - and we leave it that way for compatibility -
+    // but note that in practice we don't return any entities. That's good - we don't have access to the
+    // updated entities yet because they will be updated asynchronously by the code that reads the file
+    // we just wrote.
```
ansingh7115, line 447: _Sigh._ This makes me feel like we're shoehorning delta layer functionality into existing entity code - but that's a discussion separate from this PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderBuilder.scala
```diff
@@ -78,6 +78,8 @@ class DataRepoEntityProviderBuilder(workspaceManagerDAO: WorkspaceManagerDAO, da
     // trigger any exceptions
     val dataRef = dataRefTry.get
```
ansingh7115, line 78: Maybe not part of this PR, but we currently differentiate the response between when a reference does not exist in a workspace and when the user does not have access to the reference (see lines 72). Is this behavior okay or do we want to obscure for security reasons?

ansingh7115 (approved):
Looks good! Only the unused imports comment is blocking

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriterSpec.scala
davidangb: @ungwudik I am not aware of any info on this. I am basing my changes on what I saw empirically from running the tests a bunch of times. When I ran them I saw delays of less than 5 seconds (usually < 2), so I thought a 30-second timeout would be ok.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
davidangb: @ansingh7115 I agree - merge yours first, once that's in I will deal with merging and updating the models in use.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProviderBuilder.scala
davidangb: I think this is ok in this case. Any user who reaches this point has read access on the workspace, which means they can list the references that exist in this workspace. So they already have the ability to know which references do/do not exist; we're not leaking info.

davidangb:
@ansingh7115 @ungwudik now rebased and updated, since #1432 merged. Quick re-review for the latest changes?

ungwudik (approved):
LGTM

ansingh7115 (approved):

PR #1436, opened 2021-05-14
CA-1226: Allow user to manage project spend report configuration

MatthewBemis:
Note that this merges into the feature branch `FEATURE_CA-1218`. I am planning to merge all backend changes into that branch for now, and then decide whether or not to feature-flag them based on where the other pieces are at

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -58,12 +58,12 @@ object UserService {
...
+class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSource, protected val gcsDAO: GoogleServicesDAO, notificationDAO: NotificationDAO, samDAO: SamDAO, bqServiceFactory: GoogleBigQueryServiceFactory, pathToCredentialJson: String, requesterPaysRole: String, protected val dmConfig: DeploymentManagerConfig, protected val projectTemplate: ProjectTemplate)(implicit protected val executionContext: ExecutionContext) extends RoleSupport with FutureSupport with UserWiths with LazyLogging with StringValidationUtils {
   implicit val errorReportSource = ErrorReportSource("rawls")
 
   import dataSource.dataAccess.driver.api._
 
-  def requireProjectAction(projectName: RawlsBillingProjectName, action: SamResourceAction)(op: => Future[PerRequestMessage]): Future[PerRequestMessage] = {
```
MatthewBemis, line 66: This and the related changes below are because I don't want to continue using the outdated `PerRequest` pattern for the new endpoints. I have a PR [here](https://github.com/broadinstitute/rawls/pull/1408) that refactors Rawls and completely removes that pattern, but I'm waiting for a less disruptive time to push it through (i.e. after PPW)

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -36,28 +37,44 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
               }
             }
         } ~
-          pathPrefix("members") {
-            pathEnd {
-              get {
+        pathPrefix("spendReportConfiguration") {
```
MatthewBemis, line 40: If you've made it here, I'd recommend reviewing this file with whitespace changes turned off

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -89,7 +89,7 @@ trait RawlsApiService //(val workspaceServiceConstructor: UserInfo => WorkspaceS
   implicit val executionContext: ExecutionContext
   implicit val materializer: Materializer
 
-  val baseApiRoutes = workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutes ~ billingRoutesV2 ~ notificationsRoutes ~ servicePerimeterRoutes ~ snapshotRoutes
+  val baseApiRoutes = workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutesV2 ~ billingRoutes ~ notificationsRoutes ~ servicePerimeterRoutes ~ snapshotRoutes
```
MatthewBemis, line 92: This change is because the order that the routes were concatenated was causing some ambiguity and made these 2 new endpoints unreachable. 

`/api/billing/v2/{projectId}/spendReportConfiguration`
overlaps with
`/api/billing/{projectId}/{workbenchRole}/{email}`

for both `PUT` and `DELETE`.

The side effect is that calling the latter won't work if your project name is `v2`, but given that you'd be forced to call it with an invalid `workbenchRole` and `email`, I don't see that as a problem. I don't think that v2 is even a valid project name either.

project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ object Dependencies {
 
   val googleV = "1.22.0"
   val olderGoogleV = "1.20.0"   // TODO why do we have two google versions?  GAWB-2149
-  val workbenchGoogle2V = "0.19-3a6194b"
+  val workbenchGoogle2V = "0.21-c8093717-SNAP"
```
MatthewBemis: TODO: will remove the `SNAP` version once the workbench-libs PR is merged.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockBigQueryServiceFactory.scala
```diff
@@ -122,4 +122,14 @@ class MockGoogleBigQueryService(queryResponse: Either[Throwable, TableResult]) e
   }
 
   override def deleteDataset(datasetName: String): IO[Boolean] = IO.pure(true)
+
+  override def getTable(datasetName: String, tableName: String): IO[Option[Table]] = {
+    if(tableName.equals("gcp_billing_export_v1_table_does_not_exist")) IO.none
+    else IO.pure(Some(null))
```
MatthewBemis, line 128: I'm not too happy about this, but the `Some(null)` is completely intentional here. Creating an instance of a `Table` or a `Dataset` (for the method below) for mock purposes is surprisingly annoying. Since we don't actually use these objects for anything (we basically just call these methods to test for existence), returning `Some(null)` seemed like a somewhat okay way to deal with simulating that these methods had succeeded. It's also just test code, not production code, so I'm not too concerned.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -58,12 +58,12 @@ object UserService {
...
+  def requireProjectAction[T](projectName: RawlsBillingProjectName, action: SamResourceAction)(op: => Future[T]): Future[T] = {
```
MatthewBemis, line 66: This and the related changes below are because I don't want to continue using the outdated `PerRequest` pattern for the new endpoints. I have a PR [here](https://github.com/broadinstitute/rawls/pull/1408) that refactors Rawls and completely removes that pattern, but I'm waiting for a less disruptive time to push it through (i.e. after PPW). For now, I have chosen to not make the problem worse with this PR

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -61,7 +61,7 @@ object Boot extends IOApp with LazyLogging {
     startup() *> ExitCode.Success.pure[IO]
 
   private def startup(): IO[Unit] = {
-    implicit val log4CatsLogger: _root_.io.chrisdavenport.log4cats.Logger[IO] = Slf4jLogger.getLogger[IO]
+    implicit val log4CatsLogger = Slf4jLogger.getLogger[IO]
```
MatthewBemis, line 64: just fixing some breakage from the workbench-libs upgrade

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -391,6 +391,81 @@ paths:
...
+        description: project spend configuration information
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/SpendReportConfiguration'
+      responses:
+        204:
+          description: No Content
+        400:
+          description: Either the project does not have an active billing account, the specified dataset does not exist, or the specified table does not exist
```
gpolumbo-broad: I am trying to think of a better way to word this description.  I think there's a lot that could go here.  The dataset and specifically named table need to exist and Terra needs to be able to read from that table.  Not sure about the Billing Account stuff here.  I'll try to get back to this and suggest a better wording after reading the actual logic behind this endpoint.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -123,6 +123,10 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
     */
   def listBillingAccountsUsingServiceCredential(implicit executionContext: ExecutionContext): Future[Seq[RawlsBillingAccount]]
 
+  def getProjectBillingAccount(projectName: RawlsBillingProjectName, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```
gpolumbo-broad: We are trying to be more explicit whenever we use the overloaded term, `project`, and say either `billingProject` or `googleProject`.  So this signature would be something like:

```suggestion
  def getBillingProjectsBillingAccountName(billingProjectName: RawlsBillingProjectName, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```

Or maybe:
```suggestion
  def getBillingAccountNameForBillingProject(billingProjectName: RawlsBillingProjectName, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```

In a more general sense, I think we should be very precise with our naming here because I got confused reading this at first about what it was returning.  The fact that this is just returning a Billing Account _Name_ threw me off since we are usually dealing with IDs.

gpolumbo-broad: And I haven't gotten into the calling code yet, but this method is going to need to change or it is going to break when we get into PPW.  The reason being that in PPW, each Workspace will have its own Google Project.  All of the Workspaces (and therefore, Google Projects) _should_ have the same Billing Account set on them as specified by the Billing Project.  However, it is not guaranteed that this will be the case.  Updating the Billing Account for all of the Workspaces/GoogleProjects in a Billing Project is going to become an asynchronous operation. 

There is a possibility that at least one Google Projects errors out when updating the Billing Account.  Furthermore, during an update operation, there will naturally be some Google Projects with the old Billing Account still set and some that have already been updated with the new Billing Account.  

TL;DR - After PPW launches, it won't make sense to ask `GoogleServicesDAO` to get the Billing Account for a Billing Project, because it is possible that a Billing Project may have multiple different Billing Accounts specified on the Google Projects within it.  That is not a state we want the data to remain in, but it is nonetheless a valid state.

So although weird, maybe we add this method and mark it as deprecated right away or something.  Something to tell us that this thing is going to need to change real soon.  And maybe a Jira ticket is needed to update this in the PPW branch.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -123,6 +123,10 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
...
+
+  def getParentBillingAccount(billingProjectId: String, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```
gpolumbo-broad: In the spirit of precise naming as stated above:

```suggestion
  def getBillingAccountForGoogleProject(googleProjectId: String, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -391,6 +391,81 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/billing/v2/{projectId}/spendReportConfiguration:
+    put:
+      tags:
+        - billing_v2
+      summary: set the spend configuration for the project
```
andy7i: nit: "project" can be ambiguous so we should try to specify "billing project" vs "google project". I think we should update that in the rest of the PR as well.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5758,6 +5833,15 @@ components:
             Caller must have the add_project action for this service perimeter in
             Sam.
       description: ""
+    SpendReportConfiguration:
+      required:
+        - datasetName
```
andy7i, line 5838: just curious, why use an object with 1 string instead of adding a parameter on the API? (Is it because we expect to store more data in the `SpendReportConfiguration` object later?)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -121,6 +123,17 @@ trait RawlsBillingProjectComponent {
       }.toMap)
     }
 
+    def setBillingProjectExport(projectName: RawlsBillingProjectName, datasetName: Option[String], tableName: Option[String]): WriteAction[Int] = {
```
gpolumbo-broad: nit: slowly but surely trying to use specific names `projectName` -> `billingProjectName`.  This is applicable in other places in your PR too, but I'll just say it here.

Don't go doing a global find&replace in Rawls for `projectName` though, it's used everywhere!  That's a naming refactoring we can tackle on its own some day, for now we thought we'd just not propagate the ambiguous naming at least.

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -315,6 +315,8 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system)
         gcsDAO,
         notificationDAO,
         samDAO,
+        MockBigQueryServiceFactory.ioFactory(),
+        "/fake/Credential/Path/credentials.json",
```
andy7i: nit: maybe worth putting into the test config? 🤷

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -105,6 +105,8 @@ case class CreateRawlsBillingProjectFullRequest(
   privateIpGoogleAccess: Option[Boolean],
   servicePerimeter: Option[ServicePerimeterName])
 
+case class BillingProjectSpendConfiguration(datasetName: String)
```
gpolumbo-broad, line 108: This line raises questions in my head about why we need this here.  And if we need it here, maybe we should always wrap the datasetname and datasettable in a Configuration object?   I'm not even sure what I'm asking here, but I think maybe it's a discussion?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
andy7i:
```suggestion
  def getParentBillingAccount(billingAccountId: String, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```

same for the subclasses

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -391,6 +391,81 @@ paths:
...
+            type: string
+      requestBody:
+        description: project spend configuration information
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/SpendReportConfiguration'
+      responses:
+        204:
+          description: No Content
```
andy7i:
```suggestion
          description: Success
```

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -391,6 +391,81 @@ paths:
...
+      parameters:
+        - name: projectId
+          in: path
+          description: Project ID
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: No Content
```
andy7i:
```suggestion
          description: Success
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -134,11 +147,11 @@ trait RawlsBillingProjectComponent {
     }
 
     private def marshalBillingProject(billingProject: RawlsBillingProject): RawlsBillingProjectRecord = {
-      RawlsBillingProjectRecord(billingProject.projectName.value, billingProject.status.toString, billingProject.billingAccount.map(_.value), billingProject.message, billingProject.cromwellBackend.map(_.value), billingProject.servicePerimeter.map(_.value), billingProject.googleProjectNumber.map(_.value), billingProject.invalidBillingAccount)
+      RawlsBillingProjectRecord(billingProject.projectName.value, billingProject.status.toString, billingProject.billingAccount.map(_.value), billingProject.message, billingProject.cromwellBackend.map(_.value), billingProject.servicePerimeter.map(_.value), billingProject.googleProjectNumber.map(_.value), billingProject.invalidBillingAccount, None, None)
```
andy7i: this `None, None` doesn't seem right since it will never marshal the spendReportDataset params?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -121,6 +123,17 @@ trait RawlsBillingProjectComponent {
...
+      rawlsBillingProjectQuery
+        .filter(_.projectName === projectName.value)
+        .map(bp => (bp.spendReportDataset, bp.spendReportTable))
+        .update(datasetName, tableName)
+    }
+
+    def clearBillingProjectExport(projectName: RawlsBillingProjectName): WriteAction[Int] = {
+      setBillingProjectExport(projectName, None, None)
+    }
+
```
andy7i: would names like `setSpendReportFields` / `clearSpendReportFields` be more accurate?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -996,6 +996,7 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
 
     override def save() = {
       DBIO.seq(
+        rawlsBillingProjectQuery.create(billingProject),
```
andy7i, line 999: (just curious) was it creating workspaces without a valid workspace before? I'm surprised it didn't error before unless this bypasses db validations

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -233,4 +236,169 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
...
+      val userService = getUserService(dataSource, mockSamDAO)
+
+      Await.result(userService.setProjectSpendConfiguration(billingProject.projectName, spendReportDatasetName), Duration.Inf) shouldEqual 1
+
+      val spendReportDatasetInDb = runAndWait(dataSource.dataAccess.rawlsBillingProjectQuery.filter(_.projectName === billingProject.projectName.value).map(_.spendReportDataset).result)
+
+      spendReportDatasetName shouldEqual spendReportDatasetInDb.head.get
+    }
+  }
+
```
andy7i, line 257: would it be worth adding a test to make sure users get a 404 for billing projects that exist that the user has no access to?

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -233,4 +236,169 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
...
+      val mockSamDAO = mock[SamDAO](RETURNS_SMART_NULLS)
+      when(mockSamDAO.userHasAction(SamResourceTypeNames.billingProject, billingProject.projectName.value, SamBillingProjectActions.alterSpendReportConfiguration, userInfo)).thenReturn(Future.successful(true))
+
+      val userService = getUserService(dataSource, mockSamDAO)
+
+      Await.result(userService.setProjectSpendConfiguration(billingProject.projectName, spendReportDatasetName), Duration.Inf) shouldEqual 1
+
+      val spendReportDatasetInDb = runAndWait(dataSource.dataAccess.rawlsBillingProjectQuery.filter(_.projectName === billingProject.projectName.value).map(_.spendReportDataset).result)
+
+      spendReportDatasetName shouldEqual spendReportDatasetInDb.head.get
```
andy7i: should we also verify that spendReportTableName gets saved properly? either in this test or a separate one

andy7i: it seems like most of the tests don't check `spendReportTableName`. Is it sufficient just to check `spendReportDatasetName`?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     }
   }
 
+  def setProjectSpendConfiguration(projectName: RawlsBillingProjectName, datasetName: String): Future[Int] = {
+
+    validateBigQueryDatasetName(datasetName)
```
andy7i, line 311: nit: could we move this inside `requireProjectAction`? I know it's just a string validation but it throws me off seeing it outside a permissions check

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+
+    requireProjectAction(projectName, SamBillingProjectActions.alterSpendReportConfiguration) {
+      val bqService = bqServiceFactory.getServiceFromCredentialPath(pathToCredentialJson, GoogleProject(projectName.value))
```
andy7i: the billing project name won't necessarily be the name of a google project, but that might be outside the scope of this ticket..

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+        _ <- bqService.use(_.getDataset(datasetName)).unsafeToFuture().map {
+          case None => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The dataset $datasetName could not be found."))
+          case dataset => dataset
+        }
+
+        //The billing account ID in the DB can't be trusted, so we'll get it directly from the source of truth
+        billingAccountIdOpt <- gcsDAO.getProjectBillingAccount(projectName, userInfo)
+
+        //Get the parent billing account for the projects billing account
+        parentBillingAccountId <- billingAccountIdOpt match {
```
andy7i:
```suggestion
        rootBillingAccountId <- billingAccountIdOpt match {
```
maybe rename this? since it is looking for the top-most level (which may be a parent or not)

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+        //Get the table and validate that it exists and that we have permission to see it
+        tableName = s"gcp_billing_export_v1_${parentBillingAccountId.stripPrefix("billingAccounts/")}"
+        table <- bqService.use(_.getTable(datasetName, tableName)).unsafeToFuture()
+
+        res <- if(table.isDefined) {
+                //Isolate the db txn so we're not running any REST calls inside of it
+                dataSource.inTransaction { dataAccess =>
+                  dataAccess.rawlsBillingProjectQuery.setBillingProjectExport(projectName, Option(datasetName), Option(tableName))
+                }
+              } else throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The billing export table gcp_billing_export_v1_${parentBillingAccountId} in dataset ${datasetName} could not be found."))
```
andy7i:
```suggestion
              } else throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The billing export table ${tableName} in dataset ${datasetName} could not be found."))
```

andy7i (commented):
happy to hop on a call/slack to answer any questions about my comments/nits!

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
MatthewBemis: there's no FK from workspace namespace to billing project name, so it flew under the radar until i actually needed it

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
MatthewBemis: ah yeah, I decided part-way through that I was going to start calling these things "spend report configurations" since the "export" terminology in the ticket was pretty confusing. I'll harmonize things across the board 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: I put it outside of the permissions check because I would prefer to short-circuit the work ASAP and avoid a REST call if the dataset name is invalid

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: good call, I was struggling to come up with a good word to use here but that's definitely better!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: in what cases can that happen?

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: I leaned in this direction because I think it's just a bit simpler to deal with. It's also future-proof as you pointed out (although I don't know that we'd ever need it to be)

If I was going to take it in as a path parameter I would probably change the routes to something like:

`PUT /api/billing/v2/{projectId}/spendReportConfiguration/dataset/{datasetName}`
`DELETE /api/billing/v2/{projectId}/spendReportConfiguration/dataset`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
MatthewBemis: good eye :)

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
MatthewBemis: I think this case is covered with the existing tests where `userHasAction` returns `false`

They assert for 403 Forbidden because that's what the pre-existing `requireProjectAction` does. Perhaps that should be switched to a 404 at some point but it will consistently return 403 regardless of the reason (lack of action, or lack of overall access, resource does not exist at all) so I don't think it's leaky, at least.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+
+      for {
+        //Get the dataset to validate that it exists and that we have permission to see it
+        _ <- bqService.use(_.getDataset(datasetName)).unsafeToFuture().map {
+          case None => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The dataset $datasetName could not be found."))
+          case dataset => dataset
+        }
+
+        //The billing account ID in the DB can't be trusted, so we'll get it directly from the source of truth
+        billingAccountIdOpt <- gcsDAO.getProjectBillingAccount(projectName, userInfo)
```
gpolumbo-broad: Maybe a question for Janki to verify that Billing Account on the Billing Project matches the Billing Account on the Google Project during this process.

core/src/main/resources/swagger/api-docs.yaml
andy7i: I think that makes sense. I forgot that the param is in the path. Feels like path params should be for things like entities in rawls as opposed to data being stored on an entity. That way, the API is called on the entity and has data passed in.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
andy7i: v2 billing projects don't have their own google project. This will be the default for new projects starting with PPW. Workspaces' google projects will have auto-generated IDs.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -49,7 +49,7 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
       null,
       samDAO,
       MockBigQueryServiceFactory.ioFactory(),
-      "/fake/Credential/Path/credentials.json",
+      testConf.getString("gcs.pathToCredentialJson"),
```
andy7i, line 52: I forgot to mention we can update this line in (3) other files as well

andy7i (approved):
one more update, then 👍  from me!

thanks for the thorough updates!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
```diff
@@ -123,6 +123,10 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
     */
   def listBillingAccountsUsingServiceCredential(implicit executionContext: ExecutionContext): Future[Seq[RawlsBillingAccount]]
 
+  def getProjectBillingAccount(projectName: RawlsBillingProjectName, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
+
+  def getRootBillingAccount(billingAccountId: String, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Option[String]]
```
gpolumbo-broad: Could we comment on this explaining/linking to https://cloud.google.com/billing/docs/concepts#subaccounts .  It's a thing I hadn't really realized existed until recently and certainly it's relevant here.  

Furthermore, I don't think we need the parent/root account.  My understanding from talking with Google and one of the 3rd party vendors managing Billing Accounts for the NIH is that when the BiqQuery spend data export is set up, it happens at the subaccount level.  I do not know if it needs to happen at the subaccount or if it can happen for both.   At the very least, what I can tell you is that for a subset of Billing Accounts, the data exporting is going to be at the subaccount level.  I wouldn't be surprised if the Broad's Billing Account is set up to export at the root level, but I don't know.

gpolumbo-broad: Once again, I would rename the method to be specific about whether it is returning the Root Billing Account's name or ID.  It is not clear from reading the implementation.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,53 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+
+        //Get the parent billing account for the projects billing account
+        rootBillingAccountId <- billingAccountIdOpt match {
+                                    case Some(billingAccountId) => gcsDAO.getRootBillingAccount(billingAccountId, userInfo).map(_.getOrElse(billingAccountId))
+                                    case None => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The billing project ${projectName.value} is not linked to an active billing account."))
+                                  }
+
+        //Get the table and validate that it exists and that we have permission to see it
+        tableName = s"gcp_billing_export_v1_${rootBillingAccountId.stripPrefix("billingAccounts/")}"
```
gpolumbo-broad: Apologies for not having read all the Google docs on this yet, but is this ALWAYS the case?  That even if I configure the BigQuery data export for my subaccount, the spend data will end up in a table named after the Root Billing Account?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -36,28 +37,44 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
+          pathEnd {
+            put {
+              entity(as[BillingProjectSpendConfiguration]) { spendConfiguration =>
                 complete {
-                  userServiceConstructor(userInfo).getBillingProjectMembers(RawlsBillingProjectName(projectId))
+                  userServiceConstructor(userInfo).setBillingProjectSpendConfiguration(RawlsBillingProjectName(projectId), spendConfiguration.datasetName).map(_ => StatusCodes.NoContent)
```
gpolumbo-broad, line 45: Let's talk about v2 billing in Slack or in a Zoom or something.

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: Following up on this, it looks like it's _not_ the case. See https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables#data-schema where <BILLING_ACCOUNT_ID> will be the subaccount

Most notably:

`For resellers: For usage costs generated by a Cloud Billing subaccount, this is the ID of the subaccount, not the ID of the parent reseller Cloud Billing account.`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
MatthewBemis: You're right that the root account isn't actually needed. Per documentation [here](https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables#data-schema)

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: One thing that may be useful here is to link to a Terra support article that has troubleshooting information with the different failure modes. I think that this is going to be a rather dense description and I think that I would prefer to have that live in Terra-support-land. The actual error message from Rawls will have more specific information, too.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
MatthewBemis: I've updated the code a little bit to try to clarify this stuff more and referenced a new Jira ticket that I've created.

As discussed over Zoom, it definitely feels like there's going to need to be some refactoring once PPW goes live. There are some things that can be prepped ahead of time but I think there are some re-factorings that are going to be inevitable. 

The thing that concerns me the most is this multi-billing-account state that you've described. If it's the case that the Google projects within a billing project can temporarily map to different billing accounts, then I'm not sure how this is going to work without significant refactoring.

This will require product input but IMO a reasonable way to deal with this inconsistent state would be to prevent users from setting a spend configuration on a billing project that has inconsistent billing accounts across its Google projects. It's already going to be the case that the spend configuration will be cleared when the billing account changes, so there's already some "downtime" on spend reporting anyway in these circumstances.

I offer this solution knowing that I don't have much insight into how frequent a billing project could end up in an inconsistent state and for how long it could last.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
MatthewBemis: I think this comment thread may be of interest to you: https://github.com/broadinstitute/rawls/pull/1436#discussion_r635342537

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
gpolumbo-broad: Yikes, wonder if we should  create a ticket to create a FK constraint between Billing Project name and Workspace namespace?  Seems like we should unless one of the gurus, aka Doug, tells us there was some legitimate reason why they didn't add this previously.  I wouldn't be surprised either way.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -302,6 +302,54 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     }
   }
 
+  def setBillingProjectSpendConfiguration(billingProjectName: RawlsBillingProjectName, datasetName: String): Future[Int] = {
+
+    //Note that this assumes that the mapping between a Google Project and a Billing Project is 1:1.
+    //This will not be the case once PPW goes live. See Jira ticket CA-1363 for details.
```
gpolumbo-broad, line 308: Will thumb the PR, but don't forget to fill in details for this ticket.

gpolumbo-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
MatthewBemis: I don't think it was intentionally done this way. My best guess is that it's because workspace "namespaces" existed before billing projects did. There was no need for that column to have a FK to anything because there was no other table to point to at the time. Once we implemented billing projects, we just never switched the column to point to a unique billing project ID. All of this was super early on in Rawls history and some of those oversights still live on :)

PR #1437, opened 2021-05-18
optimize entity rewrites when no attributes changed [AS-763]

davidangb:
Use case: a user has a workspace containing a collection of 2000 data entities of type "sample." The user downloads a TSV of those entities, modifies one row, and re-uploads the TSV in order to tweak their data. This is a not-uncommon click path for users.

Before this PR: Rawls recalculated the "all_attribute_values" string for ALL 2000 entities, and issues 2000 individual `update ENTITY ... ` sql statements, overwriting each entity - including the potentially large "all_attribute_values" column - and incrementing each entity's record version.

After this PR: Rawls only recalculates the "all_attribute_values" string for the one entity that actually had its attributes change, and only issues one `update ENTITY ...` sql statement. This single entity is the only one whose record version is incremented.

-----

I did take a look at the other suggestion in AS-763 - to fail fast on `updateTrials` if any of the entities the user is trying to update do not exist (and it's an update, not an upsert). I don't think that's a big win - it doesn't change database interaction, and it only affects in-memory data munging - so I didn't make changes there for fear of regression.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -450,8 +450,9 @@ trait EntityComponent {
         preExistingEntityRecs <- getEntityRecords(workspaceContext.workspaceIdAsUUID, entities.map(_.toReference).toSet)
         savingEntityRecs <- entityQueryWithInlineAttributes.insertNewEntities(workspaceContext, entities, preExistingEntityRecs.map(_.toReference)).map(_ ++ preExistingEntityRecs)
         referencedAndSavingEntityRecs <- lookupNotYetLoadedReferences(workspaceContext, entities, savingEntityRecs.map(_.toReference)).map(_ ++ savingEntityRecs)
-        _ <- rewriteAttributes(entities, savingEntityRecs.map(_.id), referencedAndSavingEntityRecs.map(e => e.toReference -> e.id).toMap)
-        _ <- entityQueryWithInlineAttributes.optimisticLockUpdate(preExistingEntityRecs, entities)
+        actuallyUpdatedEntityIds <- rewriteAttributes(entities, savingEntityRecs.map(_.id), referencedAndSavingEntityRecs.map(e => e.toReference -> e.id).toMap)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -450,8 +450,9 @@ trait EntityComponent {
         preExistingEntityRecs <- getEntityRecords(workspaceContext.workspaceIdAsUUID, entities.map(_.toReference).toSet)
         savingEntityRecs <- entityQueryWithInlineAttributes.insertNewEntities(workspaceContext, entities, preExistingEntityRecs.map(_.toReference)).map(_ ++ preExistingEntityRecs)
         referencedAndSavingEntityRecs <- lookupNotYetLoadedReferences(workspaceContext, entities, savingEntityRecs.map(_.toReference)).map(_ ++ savingEntityRecs)
-        _ <- rewriteAttributes(entities, savingEntityRecs.map(_.id), referencedAndSavingEntityRecs.map(e => e.toReference -> e.id).toMap)
-        _ <- entityQueryWithInlineAttributes.optimisticLockUpdate(preExistingEntityRecs, entities)
+        actuallyUpdatedEntityIds <- rewriteAttributes(entities, savingEntityRecs.map(_.id), referencedAndSavingEntityRecs.map(e => e.toReference -> e.id).toMap)
+        actuallyUpdatedPreExistingEntityRecs = preExistingEntityRecs.filter(e => actuallyUpdatedEntityIds.contains(e.id))
+        _ <- entityQueryWithInlineAttributes.optimisticLockUpdate(actuallyUpdatedPreExistingEntityRecs, entities)
```
davidangb, line 455: in other words:
* ask `rewriteAttributes` which entities actually had changes;
* subset `preExistingEntityRecs` to only those that actually had changes;
* perform `optimisticLockUpdate` on only those that actually had changes.

ansingh7115 (approved):
Good tests!

sagehen03 (approved):
Looks good to me.  My only feedback is likely beyond the scope of this effort, but to the non initiated `rewriteAttributes` seems like a confusing name.  It seems like we're first determining what needs to be written back to the the database and and then writing it.

PR #1440, opened 2021-05-19
AS-774: Workflows in Requester Pays enabled workspaces fail

ansingh7115:
See https://broadworkbench.atlassian.net/browse/AS-774 on how to reproduce. I have a workspace on dev that I can share with you if you'd rather not go through the hassle of creating your own RP workspace.

This PR fixes a bug in Requester Pays workspaces where workflows immediately fail after a submission starts because, under the hood, they're trying to get the workspace bucket without specifying a project to be billed.

I added an API test that tests running a submission in an RP workspace, which I realized we actually didn't have.


ansingh7115:
jenkins retest

PR #1443, opened 2021-05-21
populate Delta Layer GCS write file with all necessary data [AS-770]

davidangb:
Jira ticket AS-770 is a followup to AS-669. In this PR, we populate AS-669's stub file with all the information Delta Layer will need - the target BQ dataset, the timestamp, the content of the upserts themselves, etc.

This PR also includes validation of user-provided upserts in order to throw decent errors.

remaining work:
- [x] proper JSON serialization of `DeltaInsert` objects
- [x] centralize BQ dataset naming and populate dataset name into `DeltaInsert`
- [x] add support for values that are null or lists
- [ ] ~add support for values that are raw json~ _(won't do)_
- [x] assess test coverage to see if we need any more
- [x] self-review for code comments, formatting, readability, etc.
- [ ] ~remove stack traces from new tests, to keep test output clean?~ _(won't do, affects too many other code paths)_

A sample of what actually gets written to GCS:
```
{
  "destination": {
    "bqDataset": "deltalayer_34542b10_e69a_4f02_84c2_6b4b8c6fc852",
    "datasetProject": "general-dev-billing-account",
    "workspaceId": "60bd54fe-5ca2-4763-b9f5-03a0a5083f89"
  },
  "insertId": "4fb4b39b-8013-49c2-ac08-eaca904ca9c6",
  "insertTimestamp": "2021-05-26T19:15:54.187331Z",
  "inserts": [{
    "datarepoRowId": "40d43454-2eb6-465a-8bb0-14d95605f3a3",
    "name": "new1",
    "value": 123
  }, {
    "datarepoRowId": "40d43454-2eb6-465a-8bb0-14d95605f3a3",
    "name": "new2",
    "value": "two"
  }, {
    "datarepoRowId": "40d43454-2eb6-465a-8bb0-14d95605f3a3",
    "name": "new3",
    "value": false
  }, {
    "datarepoRowId": "29579c37-15ad-470c-95f9-366934052186",
    "name": "another1",
    "value": 987
  }, {
    "datarepoRowId": "29579c37-15ad-470c-95f9-366934052186",
    "name": "another2",
    "value": "two two two"
  }, {
    "datarepoRowId": "29579c37-15ad-470c-95f9-366934052186",
    "name": "another3",
    "value": true
  }],
  "source": {
    "insertingUser": "111137720113699110757",
    "referenceId": "34542b10-e69a-4f02-84c2-6b4b8c6fc852"
  }
}
```


core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.rawls.deltalayer
+
+import java.util.UUID
+
+object DeltaLayer {
+  def generateDatasetName(datasetReferenceId: UUID) = {
+    "deltalayer_" + datasetReferenceId.toString.replace('-', '_')
+  }
+}
```
davidangb, line 9: moved out of `SnapshotService` so it can be easily referenced from multiple places

core/src/test/scala/org/broadinstitute/dsde/rawls/deltalayer/GcsDeltaLayerWriterSpec.scala
```diff
@@ -100,10 +104,10 @@ class GcsDeltaLayerWriterSpec extends AnyFlatSpec with GcsStorageTestSupport wit
...
     val referenceId = UUID.randomUUID()
 
-    val dest = Destination(workspaceId, referenceId)
-    val attrUpdates: Seq[AttributeUpdateOperation] = Seq(AddUpdateAttribute(AttributeName.withDefaultNS("attrName"), AttributeString("attrValue")))
-    val entityUpdates: Seq[EntityUpdateDefinition] = Seq(EntityUpdateDefinition("name", "type", attrUpdates))
-    val testInsert = DeltaInsert("vTest", insertId, new DateTime(), RawlsUserSubjectId("1234"), dest, entityUpdates)
+    val source = InsertSource(referenceId, RawlsUserSubjectId("1234"))
+    val dest = InsertDestination(workspaceId, "", GoogleProjectId(""), None)
+    val attrUpdates = Seq(DeltaRow(UUID.randomUUID(), "attrName", JsString("attrValue")))
+    val testInsert = DeltaInsert(insertId, Instant.now(), source, dest, attrUpdates)
```
davidangb, line 110: lots of model updates in this spec because the model classes changed

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceDeltaLayerSpec.scala
```diff
@@ -134,7 +134,7 @@ class EntityApiServiceDeltaLayerSpec extends ApiServiceSpec with GcsStorageTestS
           status
         }
         // issue batchUpsert against snapshot reference
-        val update1 = EntityUpdateDefinition("newSample", "Sample", Seq(AddUpdateAttribute(AttributeName.withDefaultNS("newAttribute"), AttributeString("foo"))))
+        val update1 = EntityUpdateDefinition(UUID.randomUUID().toString, "Sample", Seq(AddUpdateAttribute(AttributeName.withDefaultNS("newAttribute"), AttributeString("foo"))))
```
davidangb, line 137: these tests started failing once we started validating UUIDs

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,132 @@
...
+  val ERR_EMPTY_DATATYPE = "This method requires at least one string, boolean, or numeric scalar value to update."
+
+  /**
+   * Inspects the incoming EntityUpdateDefinitions and throws exceptions if any of those updates
+   * are illegal.
+   * @param entityUpdates the updates to validate
+   * @return the original set of updates, if all are valid
+   */
+  def validateEntityUpdates(entityUpdates: Seq[EntityUpdateDefinition]): Seq[EntityUpdateDefinition] = {
+    // verify that all entity names are legal UUIDs - they should be datarepo_row_ids
```
ansingh7115, line 33: I think there's a bunch of mapping, partitioning and collecting that's going on here that could maybe be combined? For example, you're doing `allOps.partition` on line 44 and `allOps.collect` on line 58 and getting all the AddUpdateAttribute operations both time.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,132 @@
...
+   * @return whether or not Delta Layer supports this Attribute's type
+   */
+  private def isSupportedDataType(attr: Attribute): Boolean = attr match {
+    case AttributeBoolean(_) => true
+    case AttributeNumber(_) => true
+    case AttributeString(_) => true
+    case AttributeNull => true
+    case AttributeValueEmptyList => true
+    case AttributeValueList(elems) => elems forall isSupportedDataType
+    case _ =>
```
ansingh7115, line 88: Could we combine cases like: `case AttributeBoolean(_) | AttributeNumber(_) | AttributeString(_) | AttributeNull | AttributeValueEmptyList => true`

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,132 @@
...
+      // we expect all calls to use datarepo_row_id as entity name
+      val datarepoRowId = Try(UUID.fromString(update.name)) match {
+        case Success(id) => id
+        case Failure(_) => throw new DeltaLayerException(ERR_INVALID_ENTITYNAME, code = BadRequest)
+      }
+      update.operations map { op =>
+        // strip "default." from attr name
+        val name = op.name.namespace match {
+          case AttributeName.defaultNamespace => op.name.name
+          case _ => toDelimitedName(op.name)
```
ansingh7115: I _think_ you can just do 
```
val name = toDelimitedName(op.name)
```
since `toDelimitedName(...)` contains the same logic for the `defaultNamespace` case as what you're doing here.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: done! I listed them all out for readability … my idea of readability may be in the minority :)

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: switched! I *_swear_* I investigated `toDelimitedName` and thought it didn't do what I needed, but … it absolutely does. I also added a unit test for this, since it didn't have coverage.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: looking …

davidangb: ok, refactored this, hopefully it's better now!

ansingh7115 (approved):
Thanks for making the changes - tests look comprehensive to me!

davidangb:
Jenkins automation failed contacting Vault; this looks transient. Jenkins retest.

davidangb:
… aaaaand, one test failure which should be completely unrelated:
```
org.broadinstitute.dsde.test.api.AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain when the user is in all of the groups when shared with one of the groups in the auth domain can be seen but is not accessible by group member who is a member of only one auth domain group
```

one more time: Jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
sagehen03: I see that there are two files named DeltaLayer.scala in this PR, which is a little surprising to my non-scala experienced self.  What are the conventions for how to name files and what goes in them?

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Inspects the incoming EntityUpdateDefinitions and throws exceptions if any of those updates
+   * are illegal.
+   * @param entityUpdates the updates to validate
+   * @return the original set of updates, if all are valid
+   */
+  def validateEntityUpdates(entityUpdates: Seq[EntityUpdateDefinition]): Seq[EntityUpdateDefinition] = {
+    // verify that all entity names are legal UUIDs - they should be datarepo_row_ids
+    val uuidAttempts = entityUpdates.map(upd => Try(UUID.fromString(upd.name)))
```
sagehen03, line 34: It looks like in tests at least that the `EntityUpdateDefinition's` name field is not always a UUID.  Is that correct?  If yes, and there's no easy way out out of that name field having being a UUID in some cases and a generic string in others then so be it, but in general I've found these sort of context dependent semantics often cause bugs.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    validUpdates.flatMap { update =>
+      // we expect all calls to use datarepo_row_id as entity name
+      val datarepoRowId = Try(UUID.fromString(update.name)) match {
+        case Success(id) => id
+        case Failure(_) => throw new DeltaLayerException(ERR_INVALID_ENTITYNAME, code = BadRequest)
+      }
+      update.operations map { op =>
+        // strip "default:" from attr name
+        val name = toDelimitedName(op.name)
```
sagehen03, line 115: Curious as to why we remove the default namespace qualifier?

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
```diff
@@ -0,0 +1,126 @@
...
+        val value = op match {
+          case aua:AddUpdateAttribute => attributeFormat.writeAttribute(aua.addUpdateAttribute)
+          case _ => throw new DeltaLayerException(ERR_INVALID_OPERATION, code = BadRequest)
```
sagehen03: Is this case caught earlier by `validateEntityUpdates`?  Not saying we should remove it necessarily; belt and suspenders is usually nice in the face of future refactorings.

sagehen03 (approved):
I have some a few questions for the sake of my Rawls' and Scala edification, but I see no issues to hold up merging insofar as I can spot them :)

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: we remove the default namespace qualifier mostly for consistency with elsewhere in the stack. If you view the Data tab of a workspace, for instance, all the columns you see in a data table are "foo" and "bar," not "default:foo" and "default:bar."

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: good catch, I will review tests and look for places where the name is not a UUID.

`EntityUpdateDefinition` itself - which is used for both Rawls-local entities and TDR snapshot/Delta Layer entities - only requires a string, and many Rawls-local entities are simple strings. In practice with TDR/Delta Layer, we expect that string to contain a UUID.

davidangb: hm, I could not find any instances of TDR/Delta Layer tests where `EntityUpdateDefinition` uses a non-UUID name.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
davidangb: yes - that case is redundant and already covered by `validateEntityUpdates`. I don't like the extra code here, but the alternates were less appealing:
* use a non-exhaustive match
* use an explicit type cast
* move the logic out of validateEntityUpdates`, which spreads validation across multiple methods and confounds unit testing

I will add a code comment for future devs!

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
davidangb: hm, I can't speak authoritatively on conventions, but I have renamed the model file from DeltaLayer.scala to DeltaLayerModel.scala, which matches other model files. Good call.

davidangb:
@sagehen03 I believe I have addressed all your comments - I know you already thumbed, do you want to look again? Thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayerTranslator.scala
sagehen03: Ok, I think I was missing that `EntityUpdateDefinition` is used for Rawls-local entities too.  This was the code in `EntityApiServiceSpec` that I originally identified as not using an UUID: 

`val update1 = EntityUpdateDefinition(testData.sample1.name, testData.sample1.entityType, Seq(AddUpdateAttribute(AttributeName.withDefaultNS("newAttribute"), AttributeString("bar"))))`

This might be one of those cases where I might argue code reuse is good so long as it doesn't make the code harder to follow.

PR #1445, opened 2021-05-25
CA-1229 + CA-1339: Allow users to update the billing account for billing projects

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -151,7 +151,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
         404:
-          description: Project Not Found
+          description: Billing Project Not Found
           content:
             'application/json':
               schema:
@@ -179,7 +179,7 @@ paths:
         200:
           description: Billing account successfully removed
         404:
-          description: Project Not Found
+          description: Billing project Not Found
           content:
             'application/json':
               schema:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -142,7 +142,7 @@ paths:
               $ref: '#/components/schemas/UpdateRawlsBillingAccountRequest'
         required: true
       responses:
-        204:
+        200:
           description: Billing account successfully updated
         400:
           description: Billing account not found
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -143,7 +143,7 @@ paths:
         required: true
       responses:
         204:
-          description: No Content
+          description: Billing account successfully updated
         400:
           description: Billing account not found
           content:
@@ -177,7 +177,7 @@ paths:
             type: string
       responses:
         200:
-          description: No Content
+          description: Billing account successfully removed
         404:
           description: Project Not Found
           content:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -120,6 +120,76 @@ paths:
             - profile
             - https://www.googleapis.com/auth/cloud-billing
       x-codegen-request-body-name: createProjectRequest
+  /api/billing/v2/{projectId}/billingAccount:
+    put:
+      tags:
+        - billing_v2
+      summary: update billing account on billing project
+      description: update billing account on billing project
+      operationId: updateBillingProjectBillingAccount
+      parameters:
+        - name: projectId
+          in: path
+          description: Id of the billing project
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: update billing account request
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/UpdateRawlsBillingAccountRequest'
+        required: true
+      responses:
+        204:
+          description: No Content
+        400:
+          description: Billing account not found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Project Not Found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          description: Internal Server Error
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+    delete:
+      tags:
+        - billing_v2
+      summary: deletes the billing account on billing project
+      description: deletes the billing account on billing project
+      operationId: deleteBillingProjectBillingAccount
+      parameters:
+        - name: projectId
+          in: path
+          description: Id of the billing project
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: No Content
+        404:
+          description: Project Not Found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          description: Internal Server Error
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/billing/{projectId}/members:
     get:
       tags:
@@ -5918,6 +5988,14 @@ components:
           items:
             type: string
           description: the roles the caller has on the project
+    UpdateRawlsBillingAccountRequest:
+      required:
+        - billingAccount
+      type: object
+      properties:
+        billingAccount:
+          type: string
+          description: The name of the billing account to use
     RawlsBillingProjectTransfer:
       required:
         - bucket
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -53,6 +53,22 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
+            put {
+              entity(as[UpdateRawlsBillingAccountRequest]) { updateProjectRequest =>
+                complete{
+                  userServiceConstructor(userInfo).updateBillingProjectBillingAccount(RawlsBillingProjectName(projectId), updateProjectRequest).map(_ => StatusCodes.OK)
+                }
+              }
+            } ~
+              delete {
+                complete {
+                  userServiceConstructor(userInfo).deleteBillingAccount(RawlsBillingProjectName(projectId)).map(_ => StatusCodes.OK)
```
MatthewBemis, line 67: I originally had these as `No Content` but it looks like `v2_workspaces` returns `OK` with a body.

Any objections to returning `OK` but with no body for now? I figure that is probably the path of least resistance for harmonizing these two epics once they're both live.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -120,6 +120,76 @@ paths:
             - profile
             - https://www.googleapis.com/auth/cloud-billing
       x-codegen-request-body-name: createProjectRequest
+  /api/billing/v2/{projectId}/billingAccount:
```
MatthewBemis, line 123: This swagger change is ripped right from `v2_workspaces`. Only change is that I return `OK` with no body instead of `OK` with a `RawlsBillingProjectResponse`. I left another comment on that in the route layer

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -461,6 +461,36 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     }
   }
 
+  def updateBillingProjectBillingAccount(billingProjectName: RawlsBillingProjectName, updateAccountRequest: UpdateRawlsBillingAccountRequest): Future[Unit] = {
```
MatthewBemis, line 464: `v2_workspaces` returns a `PerRequestMessage` but I'd like to stop proliferating that pattern so I don't use that here. If y'all would find it easier to mesh with, I am happy to make this return a `PerRequestMessage` as well and then take care of it all in my currently-paused [PerRequest refactor](https://github.com/broadinstitute/rawls/pull/1408) after PPW goes live.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -461,6 +461,36 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+    requireProjectAction(billingProjectName, SamBillingProjectActions.updateBillingAccount) {
```
MatthewBemis, line 467: From what I can tell, this permission check is not present in the v2_workspaces branch. Should I just file a Jira for it to be added?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
s-rubenstein: Out of curiosity, what's the rationale in removing that pattern?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
s-rubenstein: That seems fine to me, for the time being. I generally prefer returning the updated object to returning an empty body, because it gives clients the ability to base their updates on the resulting object, rather than guessing or making a fresh api call.

s-rubenstein (approved):
This looks reasonable to me. Mostly just philosophical conversations about response types, which definitely aren't blocking.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: The main reason is that it obfuscates the return types of many of the methods in the service layer, which can make the code harder to reason about and test. It also causes status codes to bleed more deeply into the business logic when preferably that would be taken care of in the API route layer.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
MatthewBemis: Sorry forgot to respond to this. The reason that I first chose to return nothing here is that the format of what I would be able to return here would be quite different than what the endpoint will return once v2_workspaces is merged. I could get close, but it would require bringing in a bunch of other code from v2_workspaces.

My thinking is that it is better to go from no response body to _some_ response body, rather than going from some response body that sets expectations for field names, values, etc and then switching to something else.

In general, I totally agree with you about trying to return the updated object when applicable.

gpolumbo-broad: I think I'm in agreement with you Matt.  Better to return nothing for now and then add to it.  Technically, we might consider that an API version bump, but I think it's ok here.  If we actually change the contents of what is returned, then I think we'd need to version the API again with PPW, and I don't think anybody wants to do that.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -401,4 +404,132 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
...
+      Await.result(userService.updateBillingProjectBillingAccount(billingProject.projectName, newBillingAccountRequest), Duration.Inf) shouldEqual ()
+
+      val spendReportDatasetInDb = runAndWait(dataSource.dataAccess.rawlsBillingProjectQuery.filter(_.projectName === billingProject.projectName.value).map(row => (row.spendReportDataset, row.spendReportTable)).result)
+
+      //assert that the spend report configuration has been fully cleared
+      spendReportDatasetInDb.head shouldEqual (None, None)
+
+      verify(mockGcsDAO, times(1)).setGoogleProjectBillingAccount(ArgumentMatchers.eq(googleProjectName), any[Option[RawlsBillingAccountName]], any[UserInfo])(any[ExecutionContext])
+    }
+  }
```
gpolumbo-broad, line 432: Matt, I know you pulled all this over from the PPW branch, but I am inclined to add a test for when we call the Update Billing Account endpoint with an empty/blank Billing Account field.  At the moment, I think that's something somebody could do and I think it would result in a failure when checking if we have access to the Billing Account, and I guess that's fine, but that's the sort of thing we should probably test because callers of the API may come to depend on that behavior.

MatthewBemis: These tests were actually not lifted from the PPW branch. I can look at adding that test case, though.

MatthewBemis: I am writing a test for this scenario and truthfully I don't think that it's valuable. This would test Google's behavior, which I can certainly mock out, but the assertion would just be that the mock did what I told it to do.

gpolumbo-broad: That makes sense.  I guess from my perspective, the long story short is that I'd rather see us have some kind of validation check(s) on the Billing Account that was submitted.  And whether we do that is based on standards/expectations we already set for other fields maybe as well as how the code behaves today.  

What validations should we check?  Do we know the min/max lengths and maybe regex that a Billing Account should match?  We certainly know that if it's an empty string, it's not valid, and I'd rather explicitly catch that.  

My preference would be that overall, we do good validation on values entered rather than counting on making the call to Google to catch whether we entered a bad value.  But I'm ok leaving it up to you to decide if you want to add this at this time.  It wasn't specified in the original ticket, and we can add it later if people are concerned about the behavior or the UX proves problematic.  In general, I do just want to put in my plug for doing good validation when we can.

tl;dr - my preference is to do good validation, but I'm OK with you making the call on this one

gpolumbo-broad (approved):

PR #1450, opened 2021-06-10
Attribute update safety limits [BW-689]

aednichols:
This PR addresses two issues related to the Terra entity table lock wait timeout incident.

1. Do not attempt to issue entity update queries including more than 20k attributes
2. Because we do not attempt to write those updates in the first place... we no longer retry them infinitely either

The 20k number was based on the top 5 updates in prod in the last 14 days, with a 2x margin on top. Because it's config based, we can easily lower it on demand if we see instability; I am erring on the side of not blocking users whose stuff works fine today.

```
Updating 1335 attribute values
Updating 2662 attribute values
Updating 7154 attribute values
Updating 8470 attribute values
Updating 1089 attribute values
```

I computed the above numbers with the following log query, having learned that Logs Explorer supports regular expressions:
```
resource.type="k8s_container"
resource.labels.cluster_name="terra-prod"
resource.labels.namespace_name="terra-prod"
textPayload =~ "\d\d\d\d attribute values for entity" OR textPayload =~ "\d\d\d\d attribute values for workspace"
resource.labels.container_name="rawls-backend-app"
```


core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -290,7 +290,7 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system)
     withDataOp { dataSource =>
       val execServiceCluster: ExecutionServiceCluster = MockShardedExecutionServiceCluster.fromDAO(executionServiceDAO, dataSource)
 
-      val config = SubmissionMonitorConfig(250.milliseconds, trackDetailedSubmissionMetrics = true)
+      val config = SubmissionMonitorConfig(250.milliseconds, trackDetailedSubmissionMetrics = true, 20000)
```
davidangb, line 293: nit: anywhere you've hardcoded 20000, should it pull from src/test/resources/reference.conf instead? This isn't a big deal, don't hold the PR for this, it might avoid confusion in the future

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -694,6 +694,87 @@ class SubmissionMonitorSpec(_system: ActorSystem) extends TestKit(_system) with
...
+          AttributeString("entry 10"),
+          AttributeString("entry 11")
+    ))))))), this))
+
+    val workflowRecord = runAndWait(
+      workflowQuery.findWorkflowByExternalIdAndSubmissionId(
+        testData.submissionMaxWorkspaceAttributes.workflows.head.workflowId.get,
+        UUID.fromString(testData.submissionMaxWorkspaceAttributes.submissionId)).result).head
+
+    val errorMessage = "Cannot save outputs to workspace because workflow's attribute count of 15 exceeds Terra maximum of 10."
```
davidangb, line 729: for my edification: why is this 15 when I only see 11 entries just above it at line 708?

davidangb (approved):
one non-blocking question, one non-blocking comment inline, thanks for pushing this through! Definitely needs a release note/heads up to support/user outreach.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -463,4 +463,75 @@ class WorkspaceModelSpec extends AnyFreeSpec with Matchers {
...
+      //             AttributeString("gs://my-workflow/shard-0/cacheCopy/some_file.tsv"),
+      //             AttributeString("gs://my-workflow/shard-1/cacheCopy/some_file.tsv"),
+      //             AttributeString("gs://my-workflow/shard-2/cacheCopy/some_file.tsv"),
+      //             AttributeString("gs://my-workflow/shard-3/cacheCopy/some_file.tsv"),
+      //             AttributeString("gs://my-workflow/shard-4/cacheCopy/some_file.tsv"),
+      //             AttributeString("gs://my-workflow/shard-5/cacheCopy/some_file.tsv")
+      //         )
+      //       )
+      //     )
+      // )
```
cjllanwarne: Are you intending to uncomment this, or should it be removed?

ansingh7115 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -471,14 +471,28 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-          logger.debug(s"Updating ${attributeCount(update.upserts)} attribute values for entity ${update.entityRef.entityName} of type ${update.entityRef.entityType} in ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}. First 100: ${update.upserts.take(100)}")
-        }
-        optWs foreach { workspace: Workspace =>
-          logger.debug(s"Updating ${attributeCount(workspace.attributes)} attribute values for workspace in ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}. First 100: ${workspace.attributes.take(100)}")
-        }
 
-        Left(updates)
+        val entityAttributeCount = optEntityUpdates map { update: WorkflowEntityUpdate =>
+          val cnt = attributeCount(update.upserts)
+          logger.debug(s"Updating $cnt attribute values for entity ${update.entityRef.entityName} of type ${update.entityRef.entityType} in ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}. ${safePrint(workspace.attributes)}")
```
cjllanwarne, line 477: TOL: Since `debug` is presumably rare, you might want to rephrase this (and others) in the `logger.debug(template, val1, val2, ...)` format

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -471,14 +471,28 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+
+        val workspaceAttributeCount = optWs map { workspace: Workspace =>
+          val cnt = attributeCount(workspace.attributes)
+          logger.debug(s"Updating $cnt attribute values for workspace in ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}. ${safePrint(workspace.attributes)}")
+          cnt
+        } getOrElse 0
+
+        if (entityAttributeCount > config.attributeUpdatesPerWorkflow) {
+          logger.error(s"Cancelled update of $entityAttributeCount entity attributes for workflow ${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}.")
+          Right((workflowRecord, Seq(AttributeString(s"Cannot save outputs to entity because workflow's attribute count of $entityAttributeCount exceeds Terra maximum of ${config.attributeUpdatesPerWorkflow}."))))
```
cjllanwarne, line 489: No change needed because I suspect this is long-preexisting, but FYI for future projects, it's more canonical to have `Left` be the bad side and `Right` be the good side of an `Either` (sometimes referred to as right-bias, and reflected in things like "map" on `Either`s mapping values on the RHS rather than the LHS)

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
cjllanwarne, line 477: It looks like `safePrint` defaults to 10 whereas this line used to print the first 100. Is that ok?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -61,7 +61,11 @@ object BootMonitors extends LazyLogging {
 
     //Boot submission monitor supervisor
     val submissionmonitorConfigRoot = conf.getConfig("submissionmonitor")
-    val submissionMonitorConfig = SubmissionMonitorConfig(util.toScalaDuration(submissionmonitorConfigRoot.getDuration("submissionPollInterval")), submissionmonitorConfigRoot.getBoolean("trackDetailedSubmissionMetrics"))
+    val submissionMonitorConfig = SubmissionMonitorConfig(
+      util.toScalaDuration(submissionmonitorConfigRoot.getDuration("submissionPollInterval")),
```
cjllanwarne, line 65: TOL: Probably a project for another time but `net.ceedubs.ficus` has helper functions like `get[Duration]("submissionPollInterval")` to help with this kind of thing

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -40,6 +40,18 @@ object Attributable {
     map.values.map(countAttributes).sum
   }
 
+  def safePrint(map: AttributeMap, depth: Int = 10): String = {
+    def safePrintInner(attr: Attribute): String = {
+      attr match {
+        case attr: AttributeListElementable => attr.toString
+        case attrList: AttributeList[_] => attrList.list.take(depth).toString
```
cjllanwarne: Does this work if the first element of `attrList` is itself a 20,000-element list?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -40,6 +40,18 @@ object Attributable {
...
+      }
+    }
+
+    val keys = map.keys.take(depth)
+    val values = map.values.take(depth).map(safePrintInner)
```
cjllanwarne, line 55: This seems a little confusing - so a depth of 10 would actually print "the first 10 values for each of the first 10 keys"?

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -463,4 +463,75 @@ class WorkspaceModelSpec extends AnyFreeSpec with Matchers {
       Attributable.attributeCount(map6) shouldBe 6
     }
   }
+
+  "Safe printer" - {
+    "should safely print attributes" in {
+      val map1 = Map(
+        AttributeName("namespace1", "name1") -> AttributeString("value1")
+      )
+      val map2: AttributeMap = Map.empty
```
cjllanwarne: Could we maybe give these readable names like `emptyMap` or `mapOf1`?

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
aednichols: I agree it's a bit confusing, these tests share a workspace that already has some attributes: https://github.com/broadinstitute/rawls/blob/53b88c392dd779bbe805683eae79853aaefd20a1/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala#L255-L260

davidangb: does this mean that when a workflow writes to workspace attributes, all pre-existing attributes are rewritten even if they haven't changed? If that's the case, we should fix that (obviously not for this PR)

aednichols: Looks like it, yeah: https://github.com/broadinstitute/rawls/blob/81d6a688a77c90489e1e3a0b3377b4015e358122/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala#L504-L512

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
aednichols: [This](https://github.com/broadinstitute/rawls/blob/3f3fff74701a4618493bb4ad04d8f5a6bcbfc79e/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala#L64-L68) was the only place I found where we actually instantiate this class with config values. The tests all seem to be hard-coded, strange as that may seem.

davidangb: huh. Well then, consistency is good - no objections!

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
aednichols: Neither, I intended it as a permanent comment - hence the indented start to the `//`

Do you disapprove?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
aednichols: Yes, I was the author of the previous line also and because virtually all lists contain repeitive data 10 is more than enough to figure out what's happening

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
aednichols: Yep that confused me too!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
aednichols: I think this has come up before and I found that our library actually recommends against it: https://github.com/lightbend/scala-logging#string-interpolation

Are you suggesting because this applies to us? "tools [...] that use the log message format as grouping key"

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: Lists of lists are not supported (see comment in `WorkspaceModelSpec.scala`)

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aednichols: Yeah, basically I'm trying to eliminate the possibility that any dimension of nesting has a high number of values. It was definitely challenging finding a way to do this that was both informative _and_ wouldn't hog CPU concatenating strings. Open to suggestions.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
cjllanwarne: Ah, ok, consider this a TOL level comment then. 

Maybe just changing `// Does not compile:` at the start to `// For example, this would not compile:` would help my quick-scan brain have an easier time telling that it's here deliberately?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
cjllanwarne: Huh, ok, fine as it is then

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
cjllanwarne: You even left me the clues to put this together and I didn't quite manage it! 🤦

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
cjllanwarne: Short of renaming it to `depthAndWidth` I don't think there's much to do here. Maybe a comment on how to use the parameter? Or maybe it's simple enough that people should be able to work it out on their own?

cjllanwarne (approved):
👍  No blockers, but a few responses to comments that you can act on or ignore, at your preference.

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
aednichols: Agreed & implemented!

aednichols:
jenkins retest

PR #1451, opened 2021-06-10
add extra info to waitUntilSubmissionComplete fail message [no ticket; test code only]

davidangb:
Adds extra info to the `waitUntilSubmissionComplete` output for automation tests. If a test fails because a submission doesn't complete in time, we now have more info.

Before this PR, error messages looked like:
> Last failure message: Monitoring submission workspace-namespace/workspace name/123e4567-e89b-12d3-a456-426614174000 until finish false was not equal to true.

After this PR, they will look like:
> Last failure message: Monitoring submission [workspace-namespace/workspace name/123e4567-e89b-12d3-a456-426614174000] until finish; actual status was [Running] with first 10 workflow ids [123e4567-e89b-12d3-a456-426614174000, 123e4567-e89b-12d3-a456-426614174000]. Scalatest-generated error message is: List(Done, Aborted) did not contain element "Running"


sagehen03 (approved):
seems good to me!

ungwudik (approved):
Thanks for adding this!

PR #1455, opened 2021-06-15
feature branch: Rawls entity attribute sharding

davidangb:
This PR represents the feature branch for the ENTITY_ATTRIBUTE sharding effort. We will be making additional PRs targeting this branch; when we merge those other PRs we can see the results here.

-----

This PR:
* creates 64 sharded entity attribute tables `ENTITY_ATTRIBUTE_xx_xx`
* copies all rows from `ENTITY_ATTRIBUTE` to the appropriate shard `ENTITY_ATTRIBUTE_xx_xx`
* renames the now-obsolete `ENTITY_ATTRIBUTE` to `ENTITY_ATTRIBUTE_archived` for later deletion
* updates all SQL statements to properly target the appropriate shard, based on the containing workspace id

Note that we may choose to incrementally move workspaces to shards, instead of the all-workspaces-at-once approach in this PR, so much may change!

Lots to do before this would be ready for production:
- [ ] ~create proper liquibase, instead of hand-rolled SQL~
- [x] create more shards, leading candidate is 256 shards but we might do 16
- [x] resolve all TODOs inline in the code
- [x] see if any of the code that uses `ENTITY_ATTRIBUTE_ALL_SHARDS` can read directly from a shard instead
- [ ] refactor, clean up, polish, document
- [ ] add extra unit test coverage as appropriate
- [ ] start running this through perf/concurrency tests
- [ ] as much validation as possible, both human clicking and automated tests
- [ ] practice data migrations against a prod-like db to understand downtime
- [ ] socialize the solution
- [ ] update playbooks


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="create_sharded_entity_tables">
+        <sql stripComments="true">
+            -- TODO: davidan turn the create-table code into a stored proc; call the stored proc repeatedly
```
ansingh7115: What happens if we tried x shards and then went "actually, we need more!" - would it be easy to increase the number of shards? Would it be worth it to make that process easy?

davidangb: creating the additional tables is pretty trivial, but re-allocating the existing rows into their appropriate shards is a big pain - it needs downtime and is a tricky SQL operation. I'd much rather overestimate the number of shards and avoid a second data migration.

davidangb:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -178,7 +180,30 @@ trait AttributeComponent {
...
+  def entityAttributeShardQuery(workspaceId: UUID): EntityAttributeShardQuery = {
+    // TODO: davidan is it ok to create a new EntityAttributeShardQuery each time? Should we pre-create one for each shard and reuse it?
+    new EntityAttributeShardQuery(determineShard(workspaceId))
+  }
+  def entityAttributeShardQuery(workspace: Workspace): EntityAttributeShardQuery = {
+    entityAttributeShardQuery(workspace.workspaceIdAsUUID)
+  }
+  def entityAttributeShardQuery(entityRec: EntityRecord): EntityAttributeShardQuery = {
+    entityAttributeShardQuery(entityRec.workspaceId)
+  }
```
davidangb, line 216: is it ok to create a new EntityAttributeShardQuery each time? Should we pre-create one for each shard and reuse it?

davidangb:
woo! all tests green and passing.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -0,0 +1,271 @@
...
+            and don't have to rollback any progress up to that changeset. -->
+    <changeSet logicalFilePath="dummy" author="davidan" id="create_and_populate_shard_0_07">
+        <sql>call createAndPopulateShard('0_07');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_0_07;</rollback>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="create_and_populate_shard_0_8f">
+        <sql>call createAndPopulateShard('0_8f');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_0_8f;</rollback>
+    </changeSet>
```
davidangb: the copy-paste in this file is super ugly and as a programmer I hate it. But I think it's the right call to have an individual liquibase changeset for each shard, and I don't know how else to do this via liquibase.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -0,0 +1,271 @@
...
+                SET @sql_text:=CONCAT('INSERT INTO ENTITY_ATTRIBUTE_',@tableSuffix,'(id, name, value_string, value_number, value_boolean,'
+                    ' value_entity_ref, list_index, owner_id, list_length, namespace, VALUE_JSON, deleted, deleted_date)'
+                    'SELECT ea.id, ea.name, ea.value_string, ea.value_number, ea.value_boolean,'
+                    ' ea.value_entity_ref, ea.list_index, ea.owner_id, ea.list_length, ea.namespace, ea.VALUE_JSON, ea.deleted, ea.deleted_date'
+                    ' FROM ENTITY_ATTRIBUTE ea, ENTITY e, WORKSPACE w'
+                    ' WHERE e.workspace_id = w.id'
+                    ' AND ea.owner_id = e.id'
+                    ' AND shardIdentifier(hex(w.id)) = ''',@tableSuffix,''';');
+                PREPARE stmt from @sql_text;
+                EXECUTE stmt;
```
davidangb: TODO: disable unique/foreign key checks and defer indexes until after the insert is done?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
sagehen03: Truthfully, I don't know enough about Scala and Rawls to make an assertion here, but I'd lean towards yes.  Re-using an instance per shard would seemingly alleviate CPU and GC related problems and I haven't heard of either of those being issues for Rawls.  Sharing instances could also introduce thread safety issues.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -0,0 +1,271 @@
...
+                DECLARE shardIdPart1 CHAR(1);
+                DECLARE shardIdPart2 CHAR(2);
+
+                SET stringToInspect = SUBSTRING(workspaceId,1,2);
+                SET secondCharToInspect = SUBSTRING(stringToInspect,2,1);
+
+                -- first part of the shard id is equal to the first char of the workspace id, 16 possible values
+                SET shardIdPart1 = SUBSTRING(stringToInspect,1,1);
+                -- second part of the shard id produces 2 values: 0-7 and 8-f
+                IF secondCharToInspect >= '0' &amp;&amp; secondCharToInspect &lt;= '7' THEN
```
sagehen03: I see `>=` and `&lt;=` on this line.. if liquibase lets us use > and < I'd say use them rather the xml escaped versions.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
sagehen03: The command line tools for MySQL and Postgres all seem to do this.  I think we are going to be surprised by how long it takes to conduct operations on 10s and 100s of millions of records so my guess it's probably worth the extra code in terms of the time it will save us.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
sagehen03: It's awesome to have all of your database changes in source control alongside the application's code, but I do feel this change is so big that we're going to run into problems trying to use liquibase here.  But, if we can use liquibase, then this seems like the right approach.

sagehen03 (approved):
I'm pleasantly surprised that there's so little code here!  I know we're studying other pathways, but doing the migration this way via liquibase would equate to down time upon service reboot, correct? Just wanted to make sure I understood that detail.  If I've got that detail right, I don't think you're considering this, but merging this and crossing our fingers in the hopes that we have downtime less than a few hours  seems like a bad idea to me.

I think it's unavoidable, but having the logic to determine the proper shard in both the MySQL function and in Scala isn't DRY so makes me a little nervous. I guess if the MySQL version is only used for the migration there will be plenty of integration testing so a unit test to confirm the two are in agreement probably isn't necessary.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -0,0 +1,282 @@
...
+    <changeSet logicalFilePath="dummy" author="davidan" id="create_shard_functions_and_procedures">
+        <sql stripComments="true" endDelimiter="/">
+            -- function to produce a shard identifier from a workspace uuid string
+            -- important: the values produced by this function must match exactly the values produced by Rawls scala code
+            -- and the values inserted into 'shards' above
+            DROP FUNCTION IF EXISTS shardIdentifier /
+            CREATE FUNCTION shardIdentifier(workspaceId CHAR(32))
+            RETURNS CHAR(4) DETERMINISTIC
+            BEGIN
+                -- this function produces 32 shards
```
MatthewBemis: 64 shards and the agreed upon naming are for a later PR I assume?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
```diff
@@ -32,11 +34,21 @@ trait DataAccess
 
   import driver.api._
 
+  // only called from TestDriverComponent
   def truncateAll: WriteAction[Int] = {
     // important to keep the right order for referential integrity !
     // if table X has a Foreign Key to table Y, delete table X first
 
-    TableQuery[EntityAttributeTable].delete andThen             // FK to entity
+    // davidan: instead of specific ordering, could we instead SET FOREIGN_KEY_CHECKS = 0; truncate tables ...; SET FOREIGN_KEY_CHECKS = 1; ?
```
MatthewBemis, line 53: I have a feeling the reason that we don't already do this is because we didn't think to do it before. Agreed that it might be nice to simplify this function at some point later on

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
MatthewBemis: I think there is a way to implement new Liquibase tags in Scala or Java that do whatever you want. Not sure if that's any better than this for something one-off, though

MatthewBemis (approved):

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
davidangb: yup! AS-957, AS-958

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
davidangb: it looks like the custom tags allow you to implement custom logic within a changeset … what I really want is to be able to programmatically generate changesets. (what I did for this one is use Scastie to loop over the shard ids and output the changesets, which I then cut and paste into the file)

davidangb:
latest Jenkins swat test passed except for one unrelated failure in orch, on `PublishedWorkspaceSpec.a user with publish permissions can publish a workspace can clone a published workspace cloned workspace should default to visible to 'all users'`:
```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 1607, groupName -> billing-project_gpalloc-qa-master-gcnmqwt_owner, still used in group(s): -> {workspace_18b221da-6ee4-49e5-bcce-15a49fdd83cb_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}
```

davidangb:
Jenkins retest, for good measure

davidangb:
one swat failure in `org.broadinstitute.dsde.test.api.AuthDomainGroupRoleSpec.removing permissions from workspace with auth domain create workspace, + group member, + project owner, - group member`:

```
{"causes":[{"causes":[],"message":"gpalloc-qa-master-encrmwm/GroupApiSpec_workspace 4b4ffdda-d87e-47c5-9ae2-ab9d4327c4a0 does not exist or you do not have permission to use it","source":"rawls","stackTrace":[],"statusCode":404}],"message":"gpalloc-qa-master-encrmwm/GroupApiSpec_workspace 4b4ffdda-d87e-47c5-9ae2-ab9d4327c4a0 does not exist or you do not have permission to use it","source":"Rawls","stackTrace":[],"statusCode":404}
```

jenkins retest.

davidangb:
two failures this time:

org.broadinstitute.dsde.test.api.AuthDomainSpec.AuthDomains should not allow changing a workspace's Realm if it exists
```
{"causes":[{"causes":[],"message":"gpalloc-qa-master-fgcx5vn/AuthDomains fd6fb1f9-173c-4552-be3d-825f2bd87145 does not exist or you do not have permission to use it","source":"rawls","stackTrace":[],"statusCode":404}],"message":"gpalloc-qa-master-fgcx5vn/AuthDomains fd6fb1f9-173c-4552-be3d-825f2bd87145 does not exist or you do not have permission to use it","source":"Rawls","stackTrace":[],"statusCode":404}
```

org.broadinstitute.dsde.test.api.orch.AsyncImportSpec.Orchestration should import a TSV asynchronously for updates
timeout

Jenkins retest

davidangb:
and this time, three failures:
1. org.broadinstitute.dsde.test.api.AuthDomainGroupApiSpec.A workspace bucket should not be accessible to project owners via projectViewer Google role
2. org.broadinstitute.dsde.test.api.AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain when the user is in all of the groups when shared with them can be seen and is accessible
3. org.broadinstitute.dsde.test.api.WorkspaceApiSpec.Rawls should not allow readers to launch workflows

all with the same root cause failure, which looks like:
```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 2314, groupName -> billing-project_gpalloc-qa-master-mcatyfk_owner, still used in group(s): -> {workspace_075aec0f-2324-42a1-9e2f-c90beeec10cb_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}
```

so, over the last three swat test runs, all three failed but with different reasons each time. In other words, all tests have passed but it took multiple runs to do so. One more: Jenkins retest

davidangb:
and, finally, a fully-passing test run!

PR #1456, opened 2021-06-16
Delete entity optimizations [AS-808]

davidangb:
This PR optimizes Rawls' interaction with the database, when a user deletes multiple entities from a workspace. Note that we do not actually delete rows from the db; we update those rows to set a deleted flag. So, we use the terminology "hide" instead of delete.

The relevant Rawls API for this is https://rawls.dsde-dev.broadinstitute.org/#/entities/delete_entities.

**Before this PR**, Rawls would:
* re-query the `ENTITY` table to retrieve all entities-to-be-hidden, just before hiding them (1 SQL statement)
* for each of N entities to be hidden, re-query to list all of that entity's attributes (N SQL statements)
* for each of M attributes to be hidden, call the relatively CPU-expensive `getSufficientlyRandomSuffix()` method
* for each of M attributes to be hidden, issue an individual `UPDATE` to the db (M SQL statements)
* for each of N entities to be hidden, call the relatively CPU-expensive `getSufficientlyRandomSuffix()` method
* for each of N entities to be hidden, issue an individual `UPDATE` to the db (N SQL statements)

This equates to `2N + (N*M) + 1` SQL statements. So, **if the user chose to hide 20 entities, each of which had 10 attributes, we would issue 241 SQL statements and invoke `getSufficientlyRandomSuffix()` 220 times.**. The number of SQL statements and method invocations is O(N) to the number of entities*attributes.

**After this PR**, Rawls will:
* _not_ re-query the `ENTITY` table to retrieve all entities-to-be-hidden, just before hiding them
* _not_ re-query to list each entity's attributes
* call `getSufficientlyRandomSuffix()` twice, once for the entire batch of attributes and once for the entire batch of entities
* issue one SQL statement to hide all relevant attributes
* issue one SQL statement to hide all relevant entities

**This is a total of 2 SQL statements and 2 invocations of `getSufficientlyRandomSuffix()`**. This is a fixed number, not O(N), though the SQL statements involved become longer as the number of entities rises.

The new SQL statements involved are … 

To hide entities:
```
update ENTITY set deleted=1, deleted_date=?, name=CONCAT(name, ?) where deleted=0 AND workspace_id=? and (entity_type, name) in ((?, ?),(?, ?),(?, ?))
```

To hide attributes:
```
update ENTITY_ATTRIBUTE ea join ENTITY e on ea.owner_id = e.id
    set ea.deleted=1, ea.deleted_date=?, ea.name=CONCAT(ea.name, ?)
    where e.workspace_id=? and ea.deleted=0 and (e.entity_type, e.name) in ((?, ?),(?, ?),(?, ?))
```
(both examples illustrate hiding three entities, but will support any number)

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -176,7 +176,7 @@ trait EntityComponent {
         val renameSuffix = "_" + getSufficientlyRandomSuffix(1000000000) // 1 billion
         val deletedDate = new Timestamp(new Date().getTime)
         // issue bulk rename/hide for all entities
-        val baseUpdate = sql"""update ENTITY set deleted=1, deleted_date=$deletedDate, name=CONCAT(name, $renameSuffix) where workspace_id = $workspaceId and (entity_type, name) in ("""
+        val baseUpdate = sql"""update ENTITY set deleted=1, deleted_date=$deletedDate, name=CONCAT(name, $renameSuffix) where deleted=0 AND workspace_id=$workspaceId and (entity_type, name) in ("""
         val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { ref => sql"(${ref.entityType}, ${ref.entityName})" })
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
       }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -76,6 +76,8 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
   override def deleteEntities(entRefs: Seq[AttributeEntityReference]): Future[Int] = {
     dataSource.inTransaction { dataAccess =>
       // withAllEntities throws exception if some entities not found; passes through if all ok
+      // TODO: davidan: withAllEntities issues individual SELECT statements for each entity, in order to collect
+      // error messages on which are missing. This could be significantly optimized.
```
davidangb, line 80: I would like to tackle this in a separate PR to avoid PR-scope-creep

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -611,30 +634,32 @@ trait EntityComponent {
...
       }
     }
 
+    // currently unused, except from hideEntityAttributes(entityId: Long) above
     private def hideEntityAttribute(attrRec: EntityAttributeRecord): WriteAction[Int] = {
       val currentTime = new Timestamp(new Date().getTime)
       entityAttributeQuery.filter(_.id === attrRec.id).map(rec => (rec.deleted, rec.name, rec.deletedDate)).update(true, renameForHiding(attrRec.id, attrRec.name), Option(currentTime))
     }
 
+    // currently unused
```
davidangb: should I just delete these three unused methods?

MatthewBemis: I'm a vote in favor of removing them

MatthewBemis (approved):
LGTM, huge improvement

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
ansingh7115: also in favor of removing

ansingh7115 (approved):
Looks good to me

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
davidangb: follow-on PR is #1457

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: removed!!

PR #1457, opened 2021-06-16
optimize withAllEntities [AS-808]

davidangb:
Optimizes database interactions for `withAllEntityRefs`, previously named `withAllEntities`.

This method is only called in the code path for deleting/hiding entities. The intention of the method is to validate that all of the entities specified by the end user for deletion do actually exist in the workspace and are active.

Before this PR, this method issued N individual SQL statements, one for each entity to be deleted. The SQL statement itself was a three-table join and returned many extraneous columns:
```
select e.id, e.name, e.entity_type, e.workspace_id, e.record_version, e.deleted, e.deleted_date,
    a.id, a.namespace, a.name, a.value_string, a.value_number, a.value_boolean, a.value_json, a.value_entity_ref, a.list_index, a.list_length, a.deleted, a.deleted_date,
    e_ref.id, e_ref.name, e_ref.entity_type, e_ref.workspace_id, e_ref.record_version, e_ref.deleted, e_ref.deleted_date
    from ENTITY e
    left outer join ENTITY_ATTRIBUTE a on a.owner_id = e.id and a.deleted = e.deleted
    left outer join ENTITY e_ref on a.value_entity_ref = e_ref.id where e.name = ? and e.entity_type = ? and e.workspace_id = ?
```

After this PR, this method issues one and only one SQL statement, which is now slimmed down to a single table and only returns the two columns we need:
```
select entity_type, name from ENTITY where workspace_id = ? and deleted = 0 and (entity_type, name) in ((?, ?),(?, ?),(?, ?))
```

This reduces SQL statements from O(N) to 1. 

To be fair, the 1 sql statement can grow to have a large `IN` clause, but that seems worth it.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/EntitySupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/EntitySupport.scala
@@ -38,10 +38,8 @@ trait EntitySupport {
       val notFound = entities diff found
       if (notFound.nonEmpty) {
         // generate error messages
-        val failures = notFound.map { e =>
-          Failure(new RawlsException(s"${e.entityType} ${e.entityName} does not exist in ${workspaceContext.toWorkspaceName}"))
-        }
-        val err = ErrorReport(statusCode = StatusCodes.BadRequest, message = (Seq("Entities were not found:") ++ failures) mkString System.lineSeparator())
+        val failureMessages = notFound.map { e => s"${e.entityType} ${e.entityName} does not exist in ${workspaceContext.toWorkspaceName}" }
+        val err = ErrorReport(statusCode = StatusCodes.BadRequest, message = (Seq("Entities were not found:") ++ failureMessages) mkString System.lineSeparator())
         DBIO.failed(new RawlsExceptionWithErrorReport(err))
       } else {
         op(found)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/util/EntitySupport.scala
```diff
@@ -31,20 +31,18 @@ trait EntitySupport {
...
+    dataAccess.entityQuery.getActiveRefs(workspaceContext.workspaceIdAsUUID, entities.toSet) flatMap { found =>
+      // were any of the user's entities not found in our query?
+      val notFound = entities diff found
+      if (notFound.nonEmpty) {
+        // generate error messages
+        val failureMessages = notFound.map { e => s"${e.entityType} ${e.entityName} does not exist in ${workspaceContext.toWorkspaceName}" }
+        val err = ErrorReport(statusCode = StatusCodes.BadRequest, message = (Seq("Entities were not found:") ++ failureMessages) mkString System.lineSeparator())
         DBIO.failed(new RawlsExceptionWithErrorReport(err))
+      } else {
+        op(found)
```
davidangb, line 45: this might be easier to review as a split diff instead of unified diff

sagehen03 (approved):
Looks good to me!

davidangb:
automation tests encountered gpalloc failures; this is a known issue. The previous run (before rebasing) passed. Holding off on re-test for now until gpalloc issues are resolved.

davidangb:
jenkins retest

MatthewBemis (approved):
Nice!

MatthewBemis:
i'll give it a jenkins retest while i'm here although it seems unlikely to pass due to ongoing unrelated issues

davidangb:
well, that was a failure in fiab-start, let's give this one more shot: jenkins retest

PR #1458, opened 2021-06-17
Logging in tests that run workflows in non-US regions [QA-1502]

salonishah11:
This PR only adds logging to below tests so that if we see timeouts again in future, they might help debug faster:

- Rawls should be able to run sub-workflow tasks in a cloned workspace in non-US regions
- Rawls should be able to create workspace and run sub-workflow tasks in non-US regions

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
@@ -272,9 +272,11 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
               deleteIntermediateOutputFiles = false
             )
 
-            logger.info(s"****** create workspace and run sub-workflow tasks in non-US regions: Returned Submission ID: $submissionId ******")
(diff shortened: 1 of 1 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -269,7 +271,12 @@ class RawlsApiSpec extends TestKit(ActorSystem("MySpec")) with AnyFreeSpecLike w
               useCallCache = false,
               deleteIntermediateOutputFiles = false
             )
-            // clean up: Abort submission
+
+            logger.info(s"Submission in $projectName/$workspaceName returned submission ID: $submissionId")
+
+            // Clean up: Abort submission. This doesn't abort the submission immediately, but only registers it to be aborted
+            // as part of clean up process. Once the test finishes (or times out), an abort request is sent to Cromwell to undo any
+            // side effects of the test. If the workflow has reached terminal state, Cromwell will return 404 for the abort request
```
aednichols, line 279: Nice

aednichols (approved):

cjllanwarne (approved):

PR #1459, opened 2021-06-18
improve rewriteAttrsAction speed [AS-803]

davidangb:
When working with large entities/attributes, the `rewriteAttrsAction` had a perf bottleneck in its comparison of attributes provided by the user and attributes pre-existing in the db. This PR significantly improves performance in this case.

The bottleneck was my bad - it was introduced in #1425. I believe the bottleneck only applies to entities with long strings; so that PR helped in some cases but made it worse in others.

Using the "performance test" included in this PR, I measured before and after the changes in this PR. There are three use cases in the test:
* starting with 0 entities, load 2304 entities with 59 attributes each. This is the "load" test.
* after those entities load, send a batchUpsert of those same 2304 entities/59 attributes, with only one entity/attribute containing a changed value. This mimics downloading a TSV, changing one value, and re-uploading it. This is the "change" test.
* delete all 2304 entities. This is the "delete" test.

Repeating the above tests over 20 iterations, I saw:

&nbsp; | after, ms | before, ms
-- | -- | --
load mean | 3970 | 3653
load min | 3180 | 3059
load max | 5317 | 4483
load sttdev | 492 | 404
change mean | **2934** | **308287**
change min | **2094** | **181005**
change max | **4878** | **345957**
change stddev | **678** | **42672**
delete mean | 6947 | 6304
delete min | 6333 | 5893
delete max | 7521 | 7108
delete stddev | 370 | 308

The tests at the bottom of `AttributeComponentSpec` validate this PR for correctness.

Reviewer: this PR contains both the important runtime fix and the "perf" test to validate it. The perf test must be run manually, it is not by default part of the unit test suite. If you have any objections to the perf test, I would happily split that off into a separate PR and just merge the runtime fix.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -360,14 +360,13 @@ trait AttributeComponent {
       // note this does not include transactionId for AttributeScratchRecords. We do not expect AttributeScratchRecords
       // here, and transactionId will eventually be going away, so don't bother
       object ComparableRecord {
-        def fromRecord[OWNER_ID](rec: RECORD): ComparableRecord[OWNER_ID] = {
-          // TODO: davidan: is the asInstanceOf safe here?? I'm not sure why rec.ownerId doesn't compile on its own
-          new ComparableRecord[OWNER_ID](rec.ownerId.asInstanceOf[OWNER_ID], rec.namespace, rec.name, rec.valueString, rec.valueNumber, rec.valueBoolean,
+        def fromRecord(rec: RECORD): ComparableRecord = {
+          new ComparableRecord(rec.ownerId, rec.namespace, rec.name, rec.valueString, rec.valueNumber, rec.valueBoolean,
             rec.valueJson, rec.valueEntityRef, rec.listIndex, rec.listLength, rec.deleted, rec.deletedDate)
         }
       }
 
-      case class ComparableRecord[OWNER_ID] (
+      case class ComparableRecord (
         ownerId: OWNER_ID,
         namespace: String,
         name: String,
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -89,6 +89,7 @@ object Dependencies {
   val scalatest: ModuleID =       "org.scalatest"                 %% "scalatest"            % "3.2.2" % "test"
   val mockito: ModuleID =         "org.scalatestplus"             %% "mockito-3-4"          % "3.2.2.0" % Test
   val mockserverNetty: ModuleID = "org.mock-server"               % "mockserver-netty"      % "5.11.2" % "test"
+  val breeze: ModuleID =          "org.scalanlp"                  %% "breeze"               % "1.2" % "test"
```
davidangb, line 92: introduced - to test code only - to help calculate stddev/mean. There are other ways to do that, this seemed simplest.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -363,32 +392,10 @@ trait AttributeComponent {
...
-        left.ownerId == right.ownerId &&
-        left.deleted == right.deleted &&
-        left.deletedDate == right.deletedDate
-      }
-
       val attributesToUpdate = toSaveAttrMap.filter {
         case (k, v) =>
             existingKeys.contains(k) && // if the attribute doesn't already exist, don't attempt to update it
-            !existingAttributes.exists(equalRecords(_, v)) // if the attribute exists and is unchanged, don't update it
+            !existingAttributesSet.contains(ComparableRecord.fromRecord(v)) // if the attribute exists and is unchanged, don't update it
```
davidangb, line 398: 1) let Scala compare case classes instead of manually inspecting the fields of the `RECORD` class
2) use `Set.contains` to look for existing attributes instead of `Traversable.exists`

MatthewBemis (approved):
LGTM. I read the test but did not run it myself because a) I trust the results you posted, and b) the actual code change itself is what I care the most about

davidangb:
merging with one thumb instead of two as an override; will get a second review soon after this is released as a hotfix.

PR #1463, opened 2021-06-22
Delta Layer companion dataset lifecycle: create [AS-724]

davidangb:
This is PR 1 of 2 for AS-724.

Highlights of this PR:
* moves create- and delete-BigQuery dataset code out of SnapshotService and into a new DeltaLayer class, so that those methods can be called from multiple places
* changes naming of the Delta Layer companion dataset so that there is one companion per workspace, not one per reference.
* changes behavior of SnapshotService so that on snapshot-reference-create, we attempt to create the Delta Layer companion dataset but ignore 409 conflicts if it already exists. Previously, we would create a new companion dataset for each reference.
* changes behavior of SnapshotService so that we no longer delete the companion dataset when a snapshot reference is deleted. Companion deletion will move to the delete-workspace event, in an upcoming PR.
* adds test coverage for the create- and delete-companion methods

REVIEWERS: is it ok to merge this PR before adding the delete-companion step to delete-workspace? That would mean until the second PR merges, we wouldn't delete companion datasets. The empty dataset should be free, and we don't have many users of this feature. Also, this PR has no functionality to delete "old-style" per-reference datasets; those will hang around, also being free. Is that ok? I want to push this PR through since it is already big, but if you think I need to hold it to get other functionality in, I understand.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -24,7 +24,7 @@ import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.HttpWorkspaceMa
 import slick.basic.DatabaseConfig
 import slick.jdbc.JdbcProfile
 import org.broadinstitute.dsde.rawls.dataaccess.{ExecutionServiceDAO, _}
-import org.broadinstitute.dsde.rawls.deltalayer.GcsDeltaLayerWriter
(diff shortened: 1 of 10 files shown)


davidangb:
some strange db errors, maybe transient? `Communications link failure`, then `PROCEDURE testdb.dropEntityAttributeTempTable does not exist`. Jenkins retest.

davidangb:
no fiab hosts available, let's try again: Jenkins retest

MatthewBemis (approved):
LGTM. I like the re-org. And no need to wait until deletes are implemented to merge this, IMO

sagehen03 (approved):
I appreciate `DeltaLayerSpec`!

PR #1465, opened 2021-06-25
[CA-1413] Sr mt fix billing account

s-rubenstein:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/README.md
+++ b/automation/README.md
@@ -43,9 +43,11 @@ FIAB=true ./config/docker-rsync-local-ui.sh
 
 #### From IntelliJ
 
-In IntelliJ, go to `Run` > `Edit Configurations...`, select `ScalaTest` under `Defaults`, and make sure there is a `Build` task configured to run before launch.
+In IntelliJ, go to `Run` > `Edit Configurations...`, select `ScalaTest` under `Templates`, and make sure there is a `Build` task configured to run before launch.
 
-Now, simply open the test spec, right-click on the class name or a specific test string, and select `Run` or `Debug` as needed. A good one to start with is `GoogleSpec` to make sure your base configuration is correct. All test code lives in `automation/src/test/scala`. FireCloud test suites can be found in `automation/src/test/scala/org/broadinstitute/dsde/firecloud/test`.
+Now, simply open the test spec, right-click on the class name or a specific test string, and select `Run` or `Debug` as needed. A good one to start with is `SnapshotAPISpec > should be able to contact Data Repo` to make sure your base configuration is correct. All test code lives in `automation/src/test/scala`. FireCloud test suites can be found in `automation/src/test/scala/org/broadinstitute/dsde/firecloud/test`.
+
+In order to run the tests, you need to open the Automation folder as a separate IntelliJ project. (File -> Open -> select the `automation` folder.)
 
 #### From the command line
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -282,7 +282,8 @@ object Boot extends IOApp with LazyLogging {
           requesterPaysRole,
           dmConfig,
           projectTemplate,
-          servicePerimeterService
+          servicePerimeterService,
+          RawlsBillingAccountName(gcsConfig.getString("adminRegisterBillingAccountId"))
         )
       val genomicsServiceConstructor: (UserInfo) => GenomicsService =
         GenomicsService.constructor(slickDataSource, gcsDAO)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -32,8 +32,8 @@ import scala.util.{Failure, Success}
 object UserService {
   val allUsersGroupRef = RawlsGroupRef(RawlsGroupName("All_Users"))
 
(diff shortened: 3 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -282,7 +282,8 @@ object Boot extends IOApp with LazyLogging {
           requesterPaysRole,
           dmConfig,
           projectTemplate,
-          servicePerimeterService
+          servicePerimeterService,
+          RawlsBillingAccountName(gcsConfig.getString("adminRegisterBillingAccountId"))
```
s-rubenstein, line 286: We looked at abstracting out a GCS config, but decided that would blow up very quickly on us.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -361,7 +361,7 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def adminRegisterBillingProject(xfer: RawlsBillingProjectTransfer): Future[PerRequestMessage] = {
     asFCAdmin {
       val billingProjectName = RawlsBillingProjectName(xfer.project)
-      val project = RawlsBillingProject(billingProjectName, CreationStatuses.Ready, Option(RawlsBillingAccountName("billingAccounts/00293C-5DEA2D-6887E7")), None)
+      val project = RawlsBillingProject(billingProjectName, CreationStatuses.Ready, Option(adminRegisterBillingAccountName), None)
```
s-rubenstein: This is now different to match gpalloc's billing account.

automation/README.md
```diff
@@ -43,9 +43,11 @@ FIAB=true ./config/docker-rsync-local-ui.sh
 
 #### From IntelliJ
 
-In IntelliJ, go to `Run` > `Edit Configurations...`, select `ScalaTest` under `Defaults`, and make sure there is a `Build` task configured to run before launch.
+In IntelliJ, go to `Run` > `Edit Configurations...`, select `ScalaTest` under `Templates`, and make sure there is a `Build` task configured to run before launch.
```
s-rubenstein, line 46: We were noticing the automation readme had some issues while trying to set it up, so decided to yak-shave and update it here.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1876,13 +1876,19 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
     for {
       googleProjectId <- traceWithParent("getGoogleProjectFromBuffer", span)(_ => resourceBufferService.getGoogleProjectFromBuffer(projectPoolType, workspaceId))
+      _ = logger.info(s"Moving google project ${googleProjectId} to folder.")
```
s-rubenstein, line 1879: We decided to leave our log messages in, because people kept needing to add them, so we decided they were useful enough to stay.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -361,7 +361,7 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def adminRegisterBillingProject(xfer: RawlsBillingProjectTransfer): Future[PerRequestMessage] = {
     asFCAdmin {
       val billingProjectName = RawlsBillingProjectName(xfer.project)
-      val project = RawlsBillingProject(billingProjectName, CreationStatuses.Ready, Option(RawlsBillingAccountName("billingAccounts/00293C-5DEA2D-6887E7")), None)
+      val project = RawlsBillingProject(billingProjectName, CreationStatuses.Ready, Option(adminRegisterBillingAccountId), None)
```
gpolumbo-broad, line 366: I never noticed that this endpoint existed, so in addition to reviewing your changes, I'm trying to wrap my head around what this api is doing.  

It is for transferring ownership of a Billing Project to a new user, right?  And for some reason, we expect to be able to ask the new owner for their email address and a bearer token, but we don't ask them for the Billing Account they want set?  Instead, we just default it to some Billing Account that we own and pay for and then expect the user to go in and update it?  

Wouldn't it be better if we just made the Billing Account part of `RawlsBillingProjectTransfer` and add that to the request payload?  I would argue that we don't even need to worry about versioning the endpoint if we do this because it's just an admin endpoint.

s-rubenstein: I am also not 100% sure what the endpoint is doing. I think it is intended for the GPAlloc projects, to re-register them after they've been put back in the pool.

It arguably would be reasonable to do that. I would, however, probably push that we should version it regardless, both out of good practice, and since we would otherwise need to make changes in every automated script that uses it, or they would end up broken.

s-rubenstein: (The change here was @marctalbott's idea, thus why I didn't go trying to fully understand it)

automation/README.md
jacmrob: love the initiative 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
gpolumbo-broad: Maybe let's put a comment on this method then saying that we believe this endpoint is primarily used by gpalloc.  The points I would want to emphasize are:

1.  Since we think gpalloc is the only user of this, that is why the Billing Account is hard coded.
2.  When we complete https://broadworkbench.atlassian.net/browse/CA-1104 we can investigate whether we should also delete this method.
3.  IF there are other use cases for this endpoint outside of gpalloc, we should update this comment to identify them

jacmrob (approved):

gpolumbo-broad (approved):

PR #1467, opened 2021-06-29
HTTP request log level error -> debug [BW-706]

aednichols:
We formerly had this at `debug` and were perplexed as to why it wasn't showing up.

Upgrading to `error` did fix the problem of logs going missing, but had undesired side effects in Sentry.

`debug` didn't work only because we did not know about adding the class to `logback.xml`.

### DIFF
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -84,6 +84,13 @@
         <appender-ref ref="Sentry" />
     </logger>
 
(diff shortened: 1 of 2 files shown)


salonishah11 (approved):

cahrens (approved):
This looks straightforward to me. What about the Jenkins automation failure though?

aednichols:
jenkins retest

PR #1474, opened 2021-07-09
[AS-787] 

sagehen03:
It's working for me locally, but I'm looking for guidance on my scala (not idiomatic I'm sure) and on what do about versioning/preserving the old API response.  ~~For now I've replaced the v2 snapshot list end point in place, which is probably wrong, but I didn't want to guess.  It should be easy enough to move things around based on reviewer consensus :crossed_fingers:~~ David looked at the production logs and we've had no callers to the v2 list end point in the last 30 days so seems safe to replace in place. 

Here's some sample JSON responses:

```
//get
{
  "attributes": {
    "instanceName": "terra",
    "snapshot": "6b073ea9-caea-4a64-9cbf-745d17d5a0c8"
  },
  "metadata": {
    "cloningInstructions": "COPY_NOTHING",
    "description": null,
    "name": "testshot",
    "resourceId": "6173b23e-2f87-4fde-9970-747de1a076c1",
    "resourceType": "DATA_REPO_SNAPSHOT",
    "stewardshipType": "REFERENCED",
    "workspaceId": "8f080501-96c3-4aee-8a73-bfab347a2f65"
  }
}

//list 
{ "gcpDataRepoSnapshots": [
  {
    "attributes": {
      "instanceName": "terra",
      "snapshot": "6b073ea9-caea-4a64-9cbf-745d17d5a0c8"
    },
    "metadata": {
      "cloningInstructions": "COPY_NOTHING",
      "description": null,
      "name": "testshot",
      "resourceId": "6173b23e-2f87-4fde-9970-747de1a076c1",
      "resourceType": "DATA_REPO_SNAPSHOT",
      "stewardshipType": "REFERENCED",
      "workspaceId": "8f080501-96c3-4aee-8a73-bfab347a2f65"
    }
  }
]}
```

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -53,7 +53,7 @@ object Dependencies {
     "org.scalatest"       %%  "scalatest"     % "3.2.2"   % "test",
     "org.seleniumhq.selenium" % "selenium-java" % "3.8.1" % "test",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
-    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-d060dbbc"
+    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-8223bdce-SNAP"
       exclude("com.typesafe.scala-logging", "scala-logging_2.11")
       exclude("com.typesafe.akka", "akka-stream_2.11")
       exclude("bio.terra", "workspace-manager-client"),
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -53,7 +53,7 @@ object Dependencies {
     "org.scalatest"       %%  "scalatest"     % "3.2.2"   % "test",
     "org.seleniumhq.selenium" % "selenium-java" % "3.8.1" % "test",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
-    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-14278f08f"
+    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-d060dbbc"
       exclude("com.typesafe.scala-logging", "scala-logging_2.11")
       exclude("com.typesafe.akka", "akka-stream_2.11")
       exclude("bio.terra", "workspace-manager-client"),
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4587,21 +4587,6 @@ components:
           $ref: '#/components/schemas/ResourceMetadata'
         attributes:
           $ref: '#/components/schemas/DataRepoSnapshotAttributes'
-    ResourceAttributesUnion:
-      type: object
-      required: [ gcpDataRepoSnapshot ]
-      properties:
-        gcpDataRepoSnapshot:
-          $ref: '#/components/schemas/DataRepoSnapshotAttributes'
-    ResourceDescription:
-      type: object
-      required: [ metadata, resourceAttributes ]
-      properties:
-        metadata:
-          $ref: '#/components/schemas/ResourceMetadata'
-        resourceAttributes:
-          $ref: '#/components/schemas/ResourceAttributesUnion'
-      description: resource description
     SnapshotListResponse:
       type: object
       required: [ gcpDataRepoSnapshots ]
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4250,7 +4250,7 @@ paths:
           content:
             '*/*':
               schema:
-                $ref: '#/components/schemas/DataRepoSnapshotResourceList'
+                $ref: '#/components/schemas/SnapshotListResponse'
         404:
           description: Workspace not found or user lacks permissions
           content:
@@ -4602,15 +4602,15 @@ components:
         resourceAttributes:
           $ref: '#/components/schemas/ResourceAttributesUnion'
       description: resource description
-    DataRepoSnapshotResourceList:
+    SnapshotListResponse:
       type: object
-      required: [ resources ]
+      required: [ gcpDataRepoSnapshots ]
       properties:
-        resources:
+        gcpDataRepoSnapshots:
           type: array
-          description: A list of snapshot Data Resource definitions from Workspace Manager
+          description: A list of DataRepoSnapshotResources from Workspace Manager
           items:
-            $ref: '#/components/schemas/ResourceDescription'
+            $ref: '#/components/schemas/DataRepoSnapshotResource'
       description: List of snapshot resources
     ErrorReport:
       required:
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -29,7 +29,7 @@ object SnapshotService {
     new SnapshotService(userInfo, dataSource, samDAO, workspaceManagerDAO, bqServiceFactory, terraDataRepoUrl, pathToCredentialJson, clientEmail, deltaLayerStreamerEmail)
   }
 
-  val GCP_SNAPSHOTS_KEY = "gcpDataRepoSnapshots"
+  final val GCP_SNAPSHOTS_KEY = "gcpDataRepoSnapshots"
 
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -16,7 +16,9 @@ import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectId, NamedD
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 
+import scala.collection.mutable.ListBuffer
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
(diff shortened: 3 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiService.scala
```diff
@@ -69,7 +70,6 @@ trait SnapshotApiService extends UserInfoDirectives {
         parameters("offset".as[Int], "limit".as[Int]) { (offset, limit) =>
           complete {
             snapshotServiceConstructor(userInfo).enumerateSnapshots(WorkspaceName(workspaceNamespace, workspaceName), offset, limit)
-              .map(resourceListToDataReferenceList)
```
sagehen03, line 72: We should delete resourceListToDataReferenceList in the unlikely event we replace v2 enumerate snapshots in place.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -87,13 +89,23 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
     }
   }
 
-  def enumerateSnapshots(workspaceName: WorkspaceName, offset: Int, limit: Int): Future[ResourceList] = {
+  def massageSnapshots(references: ResourceList): Seq[DataRepoSnapshotResource] = {
+    val res = new ListBuffer[DataRepoSnapshotResource]()
+    references.getResources.forEach(r => { val massaged = new DataRepoSnapshotResource
```
sagehen03: I want to do a references.getResources.map one liner, but it seems like there's some java/scala interop fu that I'm missing as there's no map method on `getResources`

davidangb: `getResources` returns a `java.util.List`, so Scala's `map` isn't available. The java/scala interop fu you're looking for is `import scala.collection.JavaConverters._` … see https://docs.scala-lang.org/overviews/collections/conversions-between-java-and-scala-collections.html for details. Once you've imported that, you have `asScala` available to you, and you can do:
```
    references.getResources.asScala.map{r =>
      val massaged = new DataRepoSnapshotResource
      massaged.setAttributes(r.getResourceAttributes.getGcpDataRepoSnapshot)
      massaged.setMetadata(r.getMetadata)
      massaged
    }
```

this eliminates the use of the `mutable.ListBuffer`, which is good - any mutable classes are un-idiomatic in Scala (of course *_sometimes_* you need one, but it's the exception)

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -87,13 +89,23 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+
+  def enumerateSnapshots(workspaceName: WorkspaceName, offset: Int, limit: Int): Future[Seq[DataRepoSnapshotResource]] = {
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       Try(workspaceManagerDAO.enumerateDataRepoSnapshotReferences(workspaceContext.workspaceIdAsUUID, offset, limit, userInfo.accessToken)) match {
-        case Success(references) => references
+        case Success(references) => massageSnapshots(references)
         // if we fail with a 404, it means we have no stub in WSM yet. This is benign and functionally equivalent
         // to having no references, so return the empty list.
-        case Failure(ex: bio.terra.workspace.client.ApiException) if ex.getCode == 404 => new ResourceList()
+        case Failure(ex: bio.terra.workspace.client.ApiException) if ex.getCode == 404 => List()
```
davidangb: open for debate: I prefer to be explicit about types, so I'd use `Seq.empty[DataRepoSnapshotResource]` instead of `List()`; you could also use `Seq()`. Some prefer the terseness/readability of the shorter `Seq()` but I prefer to see it written out. Scala will do the right thing in either case.

davidangb (commented):
Scala commentary inline.

As far as changing the v2 api in place vs. creating a new v3 api, the fallback position is to create a v3 api; that's SOP. In this case, however, these endpoints are only used by a few internal people … I _think_. I suggest the right thing to do is analyze logs and see how often/by whom these are called and make a determination based on that.

sagehen03:
> Scala commentary inline.
> 
> As far as changing the v2 api in place vs. creating a new v3 api, the fallback position is to create a v3 api; that's SOP. In this case, however, these endpoints are only used by a few internal people … I _think_. I suggest the right thing to do is analyze logs and see how often/by whom these are called and make a determination based on that.

I don't think I have access to the prod access logs yet.  How can I check?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -87,13 +91,24 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+  //AS-787 - rework the data so that it's in the same place in the JSON with a list and get snapshot responses
+  def massageSnapshots(references: ResourceList): Map[String, Seq[DataRepoSnapshotResource]] = {
+    val snapshots = references.getResources.asScala.map { r =>
+      val massaged = new DataRepoSnapshotResource
+      massaged.setAttributes(r.getResourceAttributes.getGcpDataRepoSnapshot)
+      massaged.setMetadata(r.getMetadata)
+      massaged
+    }
+    Map(SnapshotService.GCP_SNAPSHOTS_KEY -> snapshots)
+  }
```
davidangb, line 85: best practice is to make a new case class with a "gcpDataRepoSnapshots" member, then return an instance of that class instead of a `Map`. It's a bit more overhead but it keeps everything type-safe, and other code such as the unit tests doesn't have to be searching through a map for a well-known key.

davidangb (commented):
Implementation code looks good to me, modulo my comment about using a case class instead of a Map.

This needs a swagger update - the current response payload as documented is now incorrect.

Before this merges, let's make sure to:
* doublecheck that the v2 API is not used by anyone; I can do this by searching logs
* check with Janki, now that she's back, to ensure PO is ok with changing in place
* ensure we have good release notes, Janki may want to weigh in here
* should we doublecheck with Hornet to make sure this works for them?

sagehen03:
jenkins retest

davidangb (changes requested):
ahhhhh one thing I just thought of: this is going to break nightly automation in `SnapshotApiSpec`, specifically anything that relies on https://github.com/broadinstitute/rawls/blob/83421bec9fb8c00dd35fe01aad34717e3136740f/automation/src/test/scala/org/broadinstitute/dsde/test/api/SnapshotAPISpec.scala#L259.

Note that the nightly automation is unfortunately a separate IntelliJ project / sbt build, so it's a bit of a pain to work with

sagehen03:
> Implementation code looks good to me, modulo my comment about using a case class instead of a Map.
> 
> This needs a swagger update - the current response payload as documented is now incorrect.
> 
> Before this merges, let's make sure to:
> 
>     * doublecheck that the v2 API is not used by anyone; I can do this by searching logs
> 
>     * check with Janki, now that she's back, to ensure PO is ok with changing in place
> 
>     * ensure we have good release notes, Janki may want to weigh in here
> 
>     * should we doublecheck with Hornet to make sure this works for them?

I've updated the swagger and switched to a case class for the response.  Please verify that those changes are ok with our conventions since all the automated things seem to be telling me I'm good (i.e. did I choose the right file to hold the case class? are my swagger changes sufficient?) 

I don't see an obvious release notes file in the rawls codebase.  How do I provide release notes?  

I am still awaiting my prod account; so David or someone else suitable will need to check to verify that the current API is not in use.

I showed Olivia the JSON response at the very beginning of this process and she said it was fine.  I just messaged her about the small change we introduced by putting the list of snapshots under the gcpDataRepoSnapshots key--she says the changes are fine.

I will coordinate with David and Janki over slack.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4602,15 +4602,15 @@ components:
...
+      required: [ gcpDataRepoSnapshots ]
       properties:
-        resources:
+        gcpDataRepoSnapshots:
           type: array
-          description: A list of snapshot Data Resource definitions from Workspace Manager
+          description: A list of DataRepoSnapshotResources from Workspace Manager
           items:
-            $ref: '#/components/schemas/ResourceDescription'
+            $ref: '#/components/schemas/DataRepoSnapshotResource'
```
davidangb, line 4630: I think `ResourceDescription` is now unused and could be deleted, which in turn would make `ResourceAttributesUnion` unused and deletable. Fine if you'd rather do this in a follow-on PR along with deleting `SnapshotApiService.resourceListToDataReferenceList`

davidangb (changes requested):
these changes look good! will approve once the automation code is fixed too.

Release notes are documented in Jira, fyi

core/src/main/resources/swagger/api-docs.yaml
sagehen03: Good point.

sagehen03:
jenkins retest

sagehen03:
jenkins retest

automation/project/Dependencies.scala
```diff
@@ -53,7 +53,7 @@ object Dependencies {
     "org.scalatest"       %%  "scalatest"     % "3.2.2"   % "test",
     "org.seleniumhq.selenium" % "selenium-java" % "3.8.1" % "test",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
-    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-14278f08f"
+    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-d060dbbc"
```
davidangb:
```suggestion
    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-8223bdce-SNAP"
```

in case the version got lost in Slack threads, here it is

davidangb (approved):
:+1: let's make sure to keep an eye on the SnapshotAPISpec nightly test against alpha and ensure it passes (it should!)

MatthewBemis (approved):

PR #1475, opened 2021-07-13
delete companion on workspace deletion [AS-724]

davidangb:
When a workspace is deleted, also delete that workspace's Delta Layer companion dataset.

The dataset-deletion is tolerant of the dataset not existing, so it's safe to issue the delete-dataset request every time.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -376,7 +376,7 @@ object Boot extends IOApp with LazyLogging {
         shardedExecutionServiceCluster,
         conf.getInt("executionservice.batchSize"),
         workspaceManagerDAO,
-        dataRepoDAO,
+        deltaLayer,
         methodConfigResolver,
         gcsDAO,
         samDAO,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
@@ -66,7 +66,7 @@ class DeltaLayer(bqServiceFactory: GoogleBigQueryServiceFactory, deltaLayerWrite
                        datasetLabels: Map[String, String], aclBindings: Map[Acl.Role, Seq[(WorkbenchEmail, Entity.Type)]]): IO[DatasetId] = {
     val bqServiceResource = bqServiceFactory.getServiceForProject(googleProjectId)
     bqServiceResource.use(_.createDataset(datasetName, datasetLabels, aclBindings))
-    // TODO: AS-724: the GoogleBigQueryService.createDataset method, inside workbench-libs, always returns null for the DatasetId's project.
+    // the GoogleBigQueryService.createDataset method, inside workbench-libs, always returns null for the DatasetId's project.
     // the dataset is created in the correct project, but the return value is misleading; it results in "null" where
     // we expect something more informative, such as the log message "successfully created Delta Layer companion
     // BigQuery dataset my-dataset-name in project null" in createDatasetIfNotExist below.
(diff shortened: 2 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -376,7 +376,7 @@ object Boot extends IOApp with LazyLogging {
         shardedExecutionServiceCluster,
         conf.getInt("executionservice.batchSize"),
         workspaceManagerDAO,
-        dataRepoDAO,
+        deltaLayer,
```
davidangb, line 379: the `dataRepoDAO` was unused in WorkspaceService, so I just replaced it with the deltaLayer object.

core/src/main/scala/org/broadinstitute/dsde/rawls/deltalayer/DeltaLayer.scala
```diff
@@ -116,12 +116,14 @@ class DeltaLayer(bqServiceFactory: GoogleBigQueryServiceFactory, deltaLayerWrite
...
+    * @return true if dataset was deleted, false if it was not found (and throws on other errors)
     */
-  def deleteDataset(workspace: Workspace): IO[Boolean] = {
+  def deleteDataset(workspace: Workspace): Future[Boolean] = {
     val bqService = bqServiceFactory.getServiceForProject(workspace.googleProject)
     val datasetName = DeltaLayer.generateDatasetNameForWorkspace(workspace)
-    bqService.use(_.deleteDataset(datasetName))
+    // the GoogleBigQueryService.deleteDataset call below is already 404-safe; no need to trap
+    // for the case where the dataset doesn't exist
+    bqService.use(_.deleteDataset(datasetName)).unsafeToFuture()
```
davidangb, line 126: I moved the `.unsafeToFuture()` here instead of inside callers e.g. SnapshotService. This provides consistency between `deleteDataset` and `createDataset` - both return `Future` now.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -64,7 +64,10 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
               case t: Throwable =>
                 //fire and forget these undos, we've made our best effort to fix things at this point
                 for {
-                  _ <- deltaLayer.deleteDataset(workspaceContext).unsafeToFuture()
+                  // TODO: AS-724 is it worth deleting the companion and its ref here? If the user attempted to add a snapref,
+                  // they are likely to do so again, and we can just reuse whatever companion was already created.
+                  // This will prevent continual creation/deletion of the same-named dataset.
+                  _ <- deltaLayer.deleteDataset(workspaceContext)
```
davidangb, line 70: I would like to consider simplifying SnapshotService by not creating a WSM reference for the companion dataset, and then removing all the accompanying error-handling cases and pet-retrieval. BUT, I'd like to consider that for a future PR and leave it as-is in this one to keep scope down. For that potential future PR … does this sound like a good idea?

MatthewBemis: Are you suggesting simplifying the class by moving that logic elsewhere or by removing the code entirely? I am assuming the latter. I think that's probably fine. It kinda felt like this was just using WSM for the sake of using WSM

davidangb: the latter - don't create the WSM reference-to-companion-dataset at all. I don't think we need it, now that we have one companion per workspace.

MatthewBemis: I'm on board with that 👍

MatthewBemis (approved):

sagehen03 (approved):
:+1:

PR #1476, opened 2021-07-13
look up snapshot references by referenced snapshot id [AS-778]

davidangb:
This PR adds a new `referencedSnapshotId` query parameter to the v2 list-snapshots API. When supplied by an end user, it returns only those snapshot references that refer to the supplied snapshot id.

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4301,7 +4301,6 @@ paths:
             '*/*':
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-
   /api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/v2/{snapshotId}:
     parameters:
       - $ref: '#/components/parameters/workspaceNamespacePathParam'
@@ -4635,7 +4634,7 @@ components:
           description: A list of DataRepoSnapshotResources from Workspace Manager
           items:
             $ref: '#/components/schemas/DataRepoSnapshotResource'
-            description: List of snapshot resources
+      description: List of snapshot resources
     ErrorReport:
       required:
         - causes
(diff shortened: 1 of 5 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4295,6 +4295,43 @@ paths:
             '*/*':
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/v2/query/{referencedSnapshotId}:
```
MatthewBemis: I think instead of this being its own endpoint, it may be better to support `referencedSnapshotId` as one of the query params on the existing listSnapshots endpoint

davidangb: oooh hmm I like that, lemme work on that!

ansingh7115 (approved):
LGTM, aside from Matt's comment

core/src/main/resources/swagger/api-docs.yaml
davidangb: Ahh I'm remembering now why I didn't do it this way.

The existing listSnapshots endpoint supports pagination. We pass through the pagination params to WSM, and WSM generates a paginated SQL statement to the db.

My current implementation of search-for-referencedId does not support pagination. Because the search logic is all in Rawls, it's a different beast and can't rely on SQL's `limit`/`offset`. So I see a few options:

1. Leave PR as-is, with a separate endpoint and no pagination
2. Implement pagination inside Rawls for the search functionality
3. Implement search-for-referencedSnapshotId inside WSM and pass through pagination to WSM

I guess option 2 is my choice, @MatthewBemis @ansingh7115 what do you think?

MatthewBemis: blah....that's a PITA. I was really hoping that we'd get pagination for free here. I am fine with 2 or 3. I think I would prefer option 3 because it feels like a generic field filtering option would be more appropriate living in WSM....but I also think it's probably the most expedient route to go with option 2, so 👍

MatthewBemis (approved):

ansingh7115 (approved):

PR #1477, opened 2021-07-14
code and logic cleanup for Delta Layer companion reference [AS-724]

davidangb:
Follwup to #1475.

Don't create a WSM reference for the Delta Layer companion dataset; we don't need this. By not creating this reference, we simplify the error-handling/cleanup logic inside the create-snapshot-reference method.

This PR also includes a handful of syntax tweaks to resolve IntelliJ warnings, sorry for lumping it all together.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -1,22 +1,18 @@
 package org.broadinstitute.dsde.rawls.snapshot
 
-import java.util.UUID
 import akka.http.scaladsl.model.StatusCodes
-import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import bio.terra.workspace.model._
 import cats.effect.{ContextShift, IO}
-import com.google.cloud.bigquery.Acl
-import com.google.cloud.bigquery.Acl.Entity
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
-import org.broadinstitute.dsde.rawls.dataaccess.{GoogleBigQueryServiceFactory, SamDAO, SlickDataSource}
+import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
 import org.broadinstitute.dsde.rawls.deltalayer.DeltaLayer
-import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectId, NamedDataRepoSnapshot, SamPolicyWithNameAndEmail, SamResourceTypeNames, SamWorkspaceActions, SamWorkspacePolicyNames, UserInfo, WorkspaceAttributeSpecs, WorkspaceName}
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, NamedDataRepoSnapshot, SamWorkspaceActions, UserInfo, WorkspaceAttributeSpecs, WorkspaceName}
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
+import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
@@ -48,7 +44,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
       deltaLayer.createDatasetIfNotExist(workspaceContext, userInfo).recover {
         case t: Throwable =>
           // something went wrong creating the companion dataset
-          logger.warn(s"Error creating Delta Layer companion dataset for workspace ${workspaceName}: ${t.getMessage}")
+          logger.warn(s"Error creating Delta Layer companion dataset for workspace $workspaceName: ${t.getMessage}")
           // since companion dataset creation failed, try to clean up this snapshot reference so the user can try again
           Try(workspaceManagerDAO.deleteDataRepoSnapshotReference(wsid, snapshotRef.getMetadata.getResourceId, userInfo.accessToken)) match {
             case Success(_) =>
@@ -107,7 +103,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       // check that snapshot exists before deleting it. If the snapshot does not exist, the GET attempt will throw a 404
-      val snapshotRef = workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+      workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
       workspaceManagerDAO.deleteDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
     }
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -35,48 +35,36 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
 
   def createSnapshot(workspaceName: WorkspaceName, snapshot: NamedDataRepoSnapshot): Future[DataRepoSnapshotResource] = {
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).flatMap { workspaceContext =>
-      if(!workspaceStubExists(workspaceContext.workspaceIdAsUUID, userInfo)) {
-        workspaceManagerDAO.createWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)
+      val wsid = workspaceContext.workspaceIdAsUUID // to avoid UUID parsing multiple times
+      // create the stub workspace in WSM if it does not already exist
+      if(!workspaceStubExists(wsid, userInfo)) {
+        workspaceManagerDAO.createWorkspace(wsid, userInfo.accessToken)
       }
-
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -1,22 +1,18 @@
...
-import org.broadinstitute.dsde.rawls.dataaccess.{GoogleBigQueryServiceFactory, SamDAO, SlickDataSource}
+import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
 import org.broadinstitute.dsde.rawls.deltalayer.DeltaLayer
-import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectId, NamedDataRepoSnapshot, SamPolicyWithNameAndEmail, SamResourceTypeNames, SamWorkspaceActions, SamWorkspacePolicyNames, UserInfo, WorkspaceAttributeSpecs, WorkspaceName}
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, NamedDataRepoSnapshot, SamWorkspaceActions, UserInfo, WorkspaceAttributeSpecs, WorkspaceName}
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
+import java.util.UUID
```
davidangb, line 15: changes above are the result of IntelliJ's "Optimize Imports"

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -35,48 +31,36 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
-
-      val referenceId = snapshotRef.getMetadata.getResourceId
+      // create the requested snapshot reference
+      val snapshotRef = workspaceManagerDAO.createDataRepoSnapshotReference(wsid, snapshot.snapshotId, snapshot.name,
+        snapshot.description, terraDataRepoInstanceName, CloningInstructionsEnum.NOTHING, userInfo.accessToken)
 
       // attempt to create the BQ dataset, which might already exist
-      val createDatasetIO = IO.fromFuture(IO(deltaLayer.createDatasetIfNotExist(workspaceContext, userInfo)))
-
-      createDatasetIO.unsafeToFuture().recover {
```
davidangb, line 49: previous code turned the `Future` into an `IO`, then called `.unsafeToFuture()` on it, I removed those extra jumps

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -35,48 +31,36 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+            case Success(_) =>
+              throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError,
+                s"Unable to create snapshot reference in workspace ${workspaceContext.workspaceId} due to problems creating " +
+                  s"the Delta Layer companion dataset. Error: [${t.getMessage}]"))
+            case Failure(ex) =>
+              throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError,
+                s"Error while creating snapshot reference in workspace ${workspaceContext.workspaceId}. Additionally, there " +
+                  s"was an error cleaning up the snapshot reference. The reference may be in an unusable state. Original error " +
+                  s"during Delta Layer companion dataset creation was: [${t.getMessage}]. " +
+                  s"Error during snapshot reference cleanup: [${ex.getMessage}]."))
```
davidangb, line 59: previous code turned the cleanup into a Future and called it asynchronously; I turned it back into a synchronous operation so we can offer more precise error messages to the user.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -119,7 +103,7 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       // check that snapshot exists before deleting it. If the snapshot does not exist, the GET attempt will throw a 404
-      val snapshotRef = workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
+      workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
```
davidangb, line 106: the `snapshotRef` val was unused

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -66,10 +65,9 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
       verify(mockSamDAO, times(1)).listPoliciesForResource(any[SamResourceTypeName], any[String], any[UserInfo])
       verify(mockBigQueryServiceFactory, times(1)).getServiceForProject(workspace.googleProject)
       verify(mockWorkspaceManagerDAO, times(1)).createDataRepoSnapshotReference(any[UUID], any[UUID], any[DataReferenceName], any[Option[DataReferenceDescriptionField]], any[String], any[CloningInstructionsEnum], any[OAuth2BearerToken])
-      verify(mockWorkspaceManagerDAO, times(1)).createBigQueryDatasetReference(any[UUID], any[ReferenceResourceCommonFields], any[GcpBigQueryDatasetAttributes], any[OAuth2BearerToken])
```
davidangb, line 69: we no longer create a WSM dataset reference, nothing to verify

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -152,102 +150,7 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
-
-      // see SnapshotService line 65, "fire and forget these undos, we've made our best effort to fix things at this point"
-      // because the call to workspaceManagerDAO.deleteDataRepoSnapshotReference is fire-and-forget, we can't count on it happening
-      // by this point of this test. We need an eventually here:
-      eventually {
-        verify(mockWorkspaceManagerDAO, times(1)).deleteDataRepoSnapshotReference(any[UUID], any[UUID], any[OAuth2BearerToken])
-      }
-
-    }
```
davidangb, line 205: this test is no longer needed - since we don't create a WSM dataset reference for the companion, there's no way for the WSM dataset reference creation to fail, making this test invalid.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -152,102 +150,7 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
-      }
-      createException.errorReport.statusCode shouldBe Some(StatusCodes.InternalServerError)
-      assert(createException.errorReport.message.contains("Unable to create snapshot reference in workspace"))
-
-      verify(mockSamDAO, times(1)).listPoliciesForResource(any[SamResourceTypeName], any[String], any[UserInfo])
-      verify(mockBigQueryServiceFactory, times(2)).getServiceForProject(workspace.googleProject)
-      verify(mockWorkspaceManagerDAO, times(1)).createDataRepoSnapshotReference(any[UUID], any[UUID], any[DataReferenceName], any[Option[DataReferenceDescriptionField]], any[String], any[CloningInstructionsEnum], any[OAuth2BearerToken])
-      verify(mockWorkspaceManagerDAO, times(1)).deleteDataRepoSnapshotReference(any[UUID], any[UUID], any[OAuth2BearerToken])
-      verify(mockWorkspaceManagerDAO, times(0)).createBigQueryDatasetReference(any[UUID], any[ReferenceResourceCommonFields], any[GcpBigQueryDatasetAttributes], any[OAuth2BearerToken])
-    }
```
davidangb, line 248: this test is also no longer needed - since we don't create a WSM dataset reference for the companion, there's no need to call Sam to get a pet token, therefore there's no way for calling Sam to fail, so the test is invalid

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
sagehen03: Is it ok to use that feature? I've been avoiding it since imports like `import scala.collection.JavaConverters._` aren't directly referenced in code but seem to be required to give java collections access to scala idioms.  Am I misunderstanding things?

MatthewBemis (approved):
lgtm

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -27,15 +26,15 @@ import scala.concurrent.duration.Duration
...
 class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSugar with TestDriverComponent {
 
-  implicit val cs = IO.contextShift(global)
+  implicit val cs: ContextShift[IO] = IO.contextShift(global)
 
   //test constants
-  val fakeRawlsClientEmail = WorkbenchEmail("fake-rawls-service-account@serviceaccounts.google.com")
-  val fakeDeltaLayerStreamerEmail = WorkbenchEmail("fake-rawls-service-account@serviceaccounts.google.com")
+  val fakeRawlsClientEmail: WorkbenchEmail = WorkbenchEmail("fake-rawls-service-account@serviceaccounts.google.com")
+  val fakeDeltaLayerStreamerEmail: WorkbenchEmail = WorkbenchEmail("fake-rawls-service-account@serviceaccounts.google.com")
```
sagehen03, line 33: Why are you specifying the types on the lhs here?  I feel like I was corrected when I did that at one point :)  Just curious.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: I've also seen some cases in which Optimize Imports breaks things in the code - usually around implicits. So I always double-check its work and use it sparingly - in this case I think it's fine

sagehen03 (approved):
LGTM, especially any PR that deletes more code than it adds.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
davidangb: heh - IntelliJ throws (ignorable) warnings if these aren't present, and I was trying to make IntelliJ happy. It does make the Scala more verbose … hey, maybe I should just turn off that warning in IntelliJ …

sagehen03: I prefer the verbosity personally (the IDE does all the the typing anyhow), but probably best to stick with whatever the preponderant pattern is in the code base.

PR #1482, opened 2021-07-20
[CA-1431] Add parameter to api-docs in swagger for pre-emptive implem…

s-rubenstein:
…entation.

This supports AoU implementing the asynchronous updates for change billing account by adding the swagger field before it is actually being populated in the API response. This is technically still correct, because in the current world it is impossible for a workspace to have an incorrect billing account. Thus, currently, it should always be empty (or null), so we are safe without making any code changes. This is why we are moving forwards with using this field for AoU first, rather than `workspacesWithCorrectBillingAccount`

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -6294,6 +6294,21 @@ components:
           items:
             type: string
           description: the roles the caller has on the project
+        workspacesWithIncorrectBillingAccount:
+          type: array
+          items:
+            $ref: '#/components/schemas/WorkspaceBillingAccount'
+    WorkspaceBillingAccount:
+      required:
+        - workspaceName
+      type: object
+      properties:
+        workspaceName:
+          type: string
+          description: the name of the workspace
+        currentBillingAccountOnGoogleProject:
+          type: string
+          description: the billing project associated with the workspace
     UpdateRawlsBillingAccountRequest:
       required:
         - billingAccount


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6294,6 +6294,21 @@ components:
           items:
             type: string
           description: the roles the caller has on the project
+        workspacesWithIncorrectBillingAccount:
```
marctalbott, line 6297: So we aren't also adding `workspacesWithCorrectBillingAccount` because then we'd need to populate it? Will adding it via the PPW feature branch later make things difficult for AoU?

s-rubenstein: Yes, exactly. `workspacesWithCorrectBillingAccount` wouldn't be correct unless we made some changes to the code. My goal is to minimize our work as much as possible.

Adding it later won't make things difficult, because they are both optional fields. This just lets AoU know the format to check with, and then they can do something like:
`if(workspacesWithIncorrectBillingAccount == null || workspacesWithIncorrectBillingAccount.isEmpty()) {doSomething()}`, and that will fill their use case of ensuring the update is done.

gpolumbo-broad (approved):

marctalbott (approved):

PR #1483, opened 2021-07-20
[CA-1429] Remove parameters for v2 only

s-rubenstein:
This removes `privateIpGoogleAccess` and `highSecurityNetwork` from the v2 create billing project API, since those are going to be on by default in the new PPW world.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5845,6 +5845,25 @@ components:
             Caller must have the add_project action for this service perimeter in
             Sam.
       description: ""
+    CreateRawlsV2BillingProjectFullRequest:
+      required:
+        - billingAccount
+        - projectName
+      type: object
+      properties:
+        projectName:
+          type: string
+          description: the name of the project to create
+        billingAccount:
+          type: string
+          description: the id of the billing account to use, must start with 'billingAccounts/'
+        servicePerimeter:
+          type: string
+          description: The fully qualified name of the GCP service perimeter to put
+            this project into in the form accessPolicies/[POLICY NUMBER]/servicePerimeters/[NAME].
+            Caller must have the add_project action for this service perimeter in
+            Sam.
+      description: ""
     RawlsBillingProjectMembership:
       required:
         - projectName
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5866,7 +5866,7 @@ components:
         enableFlowLogs:
           type: boolean
           description: Optional, false if not specified. If true, enable flow logs
-            within the high security network. Requires highSecurityNetwork = true.
+            within the high security network.
       description: ""
     RawlsBillingProjectMembership:
       required:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -268,7 +268,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: '#/components/schemas/CreateRawlsBillingProjectFullRequest'
+              $ref: '#/components/schemas/CreateRawlsV2BillingProjectFullRequest'
         required: true
       responses:
         201:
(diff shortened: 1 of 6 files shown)


s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

andy7i:
could you add the ticket # and a description of changes? I think this ticket also included some open questions so the answers to those should also be documented somewhere

s-rubenstein:
Ah, I missed the questions in the comments of the ticket. Anything that is critical like that should probably end up in the body of the ticket, in the future. I will check in with all those folks and document.

andy7i:
thanks for adding in more details on the PR!

it's a bit hidden, but from the body of the [ticket](https://broadworkbench.atlassian.net/browse/CA-1429):
>If enableFlowLogs should be on by default, then we can remove all three and use a single RBS pool, containing projects with all 3 enabled. (I think this would also remove the need for a separate pool for service perimeter projects.)
>
>If we keep enableFlowLogs as a parameter, we will need to thread that parameter through so that we get a project from the correct pool. (Bonus: We may also want to rename the pool so that it's clear that it isn't used only for service perimeter projects.)

I think we still need to handle this second pool. The fact that tests pass without it is probably not a good sign.. Maybe worth pulling that into a separate ticket / PR?

s-rubenstein:
Yeah, that feels like a separate thread of work. I can pull that out into another PR and keep it in this ticket? (That feels fine to me at least)

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -271,58 +264,6 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
...
-      sealRoute(services.billingRoutesV2) ~>
-      check {
-        assertResult(StatusCodes.Created) {
-          status
-        }
-      }
-  }
-
-  it should "return 201 when creating a project with a highSecurityNetwork with enableFlowLogs and privateIpGoogleAccess turned on" in withEmptyDatabaseAndApiServices { services =>
-    val projectName = RawlsBillingProjectName("test_good")
```
andy7i, line 314: do we have tests covering projects with enableFlowLogs turned on? Or will you cover this in the subsequent PR?

andy7i (approved):
1 question about tests, but if test cases exist or will exist, then 👍

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5845,6 +5845,29 @@ components:
...
+        servicePerimeter:
+          type: string
+          description: The fully qualified name of the GCP service perimeter to put
+            this project into in the form accessPolicies/[POLICY NUMBER]/servicePerimeters/[NAME].
+            Caller must have the add_project action for this service perimeter in
+            Sam.
+        enableFlowLogs:
+          type: boolean
+          description: Optional, false if not specified. If true, enable flow logs
+            within the high security network. Requires highSecurityNetwork = true.
```
marctalbott: nit: no need to include `Requires highSecurityNetwork = true`

marctalbott (commented):
We should remove these parameters from `develop` as well, not just the feature branch. The v2 billing API is already merged to `develop`, so this PR should be merged into `develop` not the feature branch.

In a separate but related PR to the feature branch, we should update the create workspace call to pull from different RBS pools depending on if we want flow logs enabled.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
s-rubenstein: We do have tests here: https://github.com/broadinstitute/rawls/pull/1483/files#diff-874ce0961fead919b34abf612968996e00aa7ebecfeed0ba7aa4e34c7f587814R238. They don't test any special behavior though, and maybe they should.

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

PR #1485, opened 2021-07-21
fix rawls-model compile errors in automation by excluding from serviceTest [AS-787]

davidangb:
This PR adds an exclusion rule to workbench-libs' `serviceTest` - in automation - to exclude `rawls-model`.

workbench-libs currently depends on an older version `0.1-2356e282` of `rawls-model`; see https://github.com/broadinstitute/workbench-libs/blob/1d672fd6638d43947464d029053ed7c450628f7a/project/Dependencies.scala#L90. Automation was trying to use a newer version. Scala/sbt was having problems with both of these versions in its dependency tree, resulting in confounding errors where compile could not find classes inside `ExecutionModel.scala`.

Excluding rawls-model from workbench-libs fixes this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,10 @@ VAULT_TOKEN=$(cat ~/.vault-token) ARTIFACTORY_USERNAME=dsdejenkins ARTIFACTORY_P
 
 You can view what is in the artifactory here: https://broadinstitute.jfrog.io/broadinstitute/webapp/#/home
 
-After publishing, update [model/CHANGELOG.md](model/CHANGELOG.md) properly.
+After publishing:
+* update [model/CHANGELOG.md](model/CHANGELOG.md) properly
+* update the rawls-model dependency in the automation subdirectory, and ensure that sbt project is healthy
+* update the rawls-model dependency in workbench-libs serviceTest, and ensure that sbt project is healthy
 
 
 ## Troubleshooting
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -85,7 +85,10 @@ VAULT_TOKEN=$(cat ~/.vault-token) ARTIFACTORY_USERNAME=dsdejenkins ARTIFACTORY_P
 
 You can view what is in the artifactory here: https://broadinstitute.jfrog.io/broadinstitute/webapp/#/home
 
-After publishing, update [model/CHANGELOG.md](model/CHANGELOG.md) properly.
+After publishing:
+* update [model/CHANGELOG.md](model/CHANGELOG.md) properly
+* update the rawls-model dependency in the automation subdirectory, and ensure that sbt project is healthy
+* update the rawls-model dependency in workbench-libs serviceTest, and ensure that sbt project is healthy
```
davidangb, line 91: workbench-libs PR: broadinstitute/workbench-libs#696

automation/project/Dependencies.scala
```diff
@@ -53,7 +54,7 @@ object Dependencies {
     "org.scalatest"       %%  "scalatest"     % "3.2.2"   % "test",
     "org.seleniumhq.selenium" % "selenium-java" % "3.8.1" % "test",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
-    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-14278f08f"
+    "org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-90eae81cd"
```
davidangb, line 57: I just published this release version, from the head of develop branch

MatthewBemis (approved):

sagehen03 (approved):
ah, the old multiple version classloader non-determinism--that makes a lot of sense.  Thank you for figuring that out!  At some I'd like to discuss with you all of the surprises I found on AS-787.  There's still a lot that I don't understand.

PR #1444, opened 2021-05-24
Add memoryRetryMultiplier parameter to submission create endpoint [BW-281]

cjllanwarne:
In analogy with #1310.

Still TODO:
- [ ] Update orchestration
- [ ] Update workbench-libs(?)

Original checklist:
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [x] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
  [alpha](https://console.cloud.google.com/sql/instances/terraform-r4caezzc35c4tb7pgdhwkmme4y/backups?project=broad-dsde-alpha&organizationId=548622027621)
  DBs in Google Cloud Console
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] Tell your tech lead (TL) that the PR exists if they want to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Assign to submitter** to finalize
[...]

PR #1488, opened 2021-07-22
update to workbench-libs serviceTest 0.19 [AS-787]

davidangb:
following up on #1485 and broadinstitute/workbench-libs#696, this PR updates Rawls automation tests to use the latest version of workbench-libs' serviceTest library.

The upgrade from 0.18 to 0.19 of serviceTest requires adding `useReferenceDisks` and `memoryRetryMultiplier` parameters to `Rawls.submissions.launchWorkflow`; see https://github.com/broadinstitute/workbench-libs/blob/develop/serviceTest/CHANGELOG.md and broadinstitute/workbench-libs#631

`useReferenceDisks` has a default of `false` per #1310: https://github.com/broadinstitute/rawls/blob/0de4f198495c0fa28c7818433bf7f8eb87981359/model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala#L31

`memoryRetryMultiplier` has a default of `1.0` per #1444: https://github.com/broadinstitute/rawls/blob/0de4f198495c0fa28c7818433bf7f8eb87981359/model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala#L32

so I've used those values everywhere I needed to update signatures.

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -7,7 +7,7 @@ object Dependencies {
   val akkaHttpV     = "10.2.0"
   val jacksonV      = "2.11.3"
 
-  val serviceTestV = "0.18-23f5ae7"
+  val serviceTestV = "0.19-5edfa7a"
   val workbenchGoogleV = "0.21-24dcd92"
   val workbenchGoogle2V = "0.21-e17afdf"
   val workbenchModelV  = "0.14-65bba14"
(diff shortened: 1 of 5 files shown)


davidangb:
@cjllanwarne as the author of the workbench-libs changes, do you have bandwidth to review this PR?

ungwudik (approved):
LGTM.

I don't think this change will particularly affect `SnapshotAPISpec` in a way that it doesn't affect other tests. But I'm wondering if it's worth running `SnapshotAPISpec` locally to confirm things look good there since the test doesn't (currently) run in FiaBs like the others.

davidangb:
@ungwudik I just ran those tests, and they passed!

cjllanwarne (approved):
Very nice, and thank you!

PR #1492, opened 2021-08-05
[CA-1463] delete rawls billing project before deleting sam BP resource

andy7i:
CA-1463
Hopefully a workaround for CA-1206. 

Our investigations have found that it's possible sam resources for the *workspace* can hang around if any of the deletion steps fail. If that happens, then we get issues when we try to delete the *billing project* resource in sam, since the billing project owners group is still used in the workspace project owners group.  Deleting the rawls resource before deleting the sam resource should work as a workaround for the issues seen in 1206. 

This doesn't address the root cause of CA-1206 (which may be related to deleteWorkspace) but should unblock billing project deletion in rawls (and therefore prevent more instances of AoU being unable to delete projects as described in CA-1206).

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -258,10 +258,10 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
    * */
   def unregisterBillingProjectWithUserInfo(projectName: RawlsBillingProjectName, ownerUserInfo: UserInfo): Future[PerRequestMessage] = {
     for {
-      _ <- samDAO.deleteResource(SamResourceTypeNames.billingProject, projectName.value, ownerUserInfo)
       _ <- dataSource.inTransaction { dataAccess =>
         dataAccess.rawlsBillingProjectQuery.delete(projectName)
       }
+      _ <- samDAO.deleteResource(SamResourceTypeNames.billingProject, projectName.value, ownerUserInfo) // Moving this to the end so that the rawls record is cleared even if there are issues clearing the Sam resource (CA-1206)
     } yield {
       RequestComplete(StatusCodes.NoContent)
     }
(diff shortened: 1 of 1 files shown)


s-rubenstein (commented):
Are we worried at all about leaving dangling Sam resources? I know usually I try to delete the db object that's in the service with the method last to try to avoid the case where we clean up the db, and then get an error calling the dependent service, because in that case there's an error down the line, but no information in the calling service's database that anything is wrong.

andy7i:
> Are we worried at all about leaving dangling Sam resources? I know usually I try to delete the db object that's in the service with the method last to try to avoid the case where we clean up the db, and then get an error calling the dependent service, because in that case there's an error down the line, but no information in the calling service's database that anything is wrong.

The danging sam resource could cause issues if it collides with a new billing project with the same name. I'm not sure if the request would fail or if the request would succeed (in which case that's a potential security risk; I think we enforce some level of requiring billing project names to be unique though). We should look into this, but I'd propose that we do that separately (and let this change go in first) since this issue already affects prod and this change shouldn't make anything worse, unless we expect rawls db calls to fail.

jacmrob (approved):
Curious how we test if the theoretical workaround is actually working

andy7i:
> Curious how we test if the theoretical workaround is actually working

@jacmrob yeah I probably undersold it a bit. This will definitely fix ONE issue, where any problems during workspace deletion will make billing project deletion fail. This issue is the leading theory for why CA-1206 is happening. I've tested this locally against dev and verified that the BP gets deleted in rawls even if the workspace exists in sam.

s-rubenstein (approved):
This looks fine. I am okay with this moving forward and seeing if it helps. Would you mind adding an error handler for if the sam call errors, to log a warning or something? Just something so that this is flagged to us/we can potentially track it if we are getting a lot of dangling sam projects?

jacmrob:
I think this can be merged despite the codecov - it's just a logging line

andy7i:
created new tickets for followup actions:
https://broadworkbench.atlassian.net/browse/CA-1470
https://broadworkbench.atlassian.net/browse/CA-1472

PR #1493, opened 2021-08-05
[CA-1464] logging delete workspace failures to help debug CA-1206

andy7i:
CA-1464

Changelog:
* Adds logging messages in deleteWorkspace so it is easier for us to track down the root cause of the failed deleteWorkspace calls. 
* makes deleting in Workspace Manager fail if it's not the expected 404. (was this a bug?)
* removes the mixing of try (sync) and future (async) 

---

error message examples:

```
// before:
rawls [ERROR] [08/05/2021 16:31:33.696] [scala-execution-context-global-76] [akka.actor.ActorSystemImpl(rawls)] HttpMethod(DELETE) http://rawls-sbt:8080/api/workspaces/aj-ca-1206-6/aj-test-ws-2020-copy7: 500 Internal Server Error entity: {"causes":[],"message":"aj-test-failures","source":"rawls","stackTrace":[]}
```

```
// after:
rawls [ERROR] [16:31:33.560] [scala-execution-context-global-73] o.b.d.r.workspace.WorkspaceService - Unexpected failure deleting workspace (while deleting workflowCollection in Sam) for workspace `aj-ca-1206-6/aj-test-ws-2020-copy7`
rawls org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport: ErrorReport(rawls,aj-test-failures,None,List(),List(),None)
rawls 	at org.broadinstitute.dsde.rawls.dataaccess.HttpSamDAO.$anonfun$deleteResource$1(HttpSamDAO.scala:159)
rawls 	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
rawls 	at scala.util.Success.$anonfun$map$1(Try.scala:255)
rawls 	at scala.util.Success.map(Try.scala:213)
rawls 	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
rawls 	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
rawls 	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
rawls 	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
rawls 	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
rawls 	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
rawls 	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
rawls 	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
rawls 	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
rawls 	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
rawls [WARN] [16:31:33.648] [scala-execution-context-global-73] io.sentry.dsn.Dsn - *** Couldn't find a suitable DSN, Sentry operations will do nothing! See documentation: https://docs.sentry.io/clients/java/ ***
rawls [ERROR] [08/05/2021 16:31:33.696] [scala-execution-context-global-76] [akka.actor.ActorSystemImpl(rawls)] HttpMethod(DELETE) http://rawls-sbt:8080/api/workspaces/aj-ca-1206-6/aj-test-ws-2020-copy7: 500 Internal Server Error entity: {"causes":[],"message":"aj-test-failures","source":"rawls","stackTrace":[]}
```

---

I decided not to refactor deleteWorkspace too much right now, since we should probably figure out the root cause first before we make a lot of changes.

---

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -378,7 +378,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
             logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}. Received ${e.getCode}: [${e.getResponseBody}]")
             Future.successful()
           }
-          else throw e
+          throw e
         }
       }
       // Delete the Delta Layer companion dataset, if it exists
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -372,13 +372,13 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
       // Delete workspace manager record (which will only exist if there had ever been a TDR snapshot in the WS)
       _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
-        //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle 404 exceptions here
-        case e: ApiException if e.getCode == StatusCodes.NotFound.intValue => {
-          Future.successful()
-        }
-        case t:Throwable => {
-          logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}`", t)
-          throw t
+        //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle all exceptions here
+        case e: ApiException => {
+          if(e.getCode != StatusCodes.NotFound.intValue) {
+            logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}. Received ${e.getCode}: [${e.getResponseBody}]")
+            Future.successful()
+          }
+          else throw e
         }
       }
       // Delete the Delta Layer companion dataset, if it exists
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -363,6 +363,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
             logger.warn(s"Unexpected failure deleting workspace in Workspace Manager. Received ${e.getCode}: [${e.getResponseBody}]")
             Future.successful()
           }
+          throw e
         }
       }
       // Delete the Delta Layer companion dataset, if it exists
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -372,8 +372,8 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
       // Delete workspace manager record (which will only exist if there had ever been a TDR snapshot in the WS)
       _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
-        //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle 404 exceptions here
-        case e: ApiException if e.getCode == StatusCodes.NotFound.intValue => {
+        //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle 401 exceptions here
+        case e: ApiException if e.getCode == StatusCodes.Unauthorized.intValue => {
           Future.successful()
         }
         case t:Throwable => {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -349,7 +349,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
     for {
 
       workflowsToAbort <- deletionFuture recoverWith {
-        case t:Throwable => { // logging added as part of investigation into https://broadworkbench.atlassian.net/browse/CA-1206
+        case t:Throwable => {
           logger.warn(s"Unexpected failure deleting workspace (while running `deletionFuture`) for workspace `${workspaceName}`", t)
           throw t
         }
@@ -357,7 +357,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
 
       // Abort running workflows
       aborts = Future.traverse(workflowsToAbort) { wf => executionServiceCluster.abort(wf, userInfo) } recoverWith {
-        case t:Throwable => { // logging added as part of investigation into https://broadworkbench.atlassian.net/browse/CA-1206
+        case t:Throwable => {
           logger.warn(s"Unexpected failure deleting workspace (while aborting workflows) for workspace `${workspaceName}`", t)
           throw t
         }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -348,26 +348,52 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
       }
     for {
 
-      workflowsToAbort <- deletionFuture
+      workflowsToAbort <- deletionFuture recoverWith {
+        case t:Throwable => { // logging added as part of investigation into https://broadworkbench.atlassian.net/browse/CA-1206
```
s-rubenstein: Nit: I don't think you need these comments. Regardless of why the logs were added, they seem like good things to have around, and I would want to keep them around even after 1206 merges.

s-rubenstein (approved):
This looks good

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -348,26 +348,52 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
         //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle all exceptions here
         case e: ApiException => {
           if(e.getCode != StatusCodes.NotFound.intValue) {
-            logger.warn(s"Unexpected failure deleting workspace in Workspace Manager. Received ${e.getCode}: [${e.getResponseBody}]")
+            logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}. Received ${e.getCode}: [${e.getResponseBody}]")
+            Future.successful()
           }
-          Success(())
+          throw e
```
s-rubenstein, line 381: Why do we want to throw here where we didn't before? Is it just for our records?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
jacmrob: +1

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -348,26 +348,52 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
+      _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
         //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle all exceptions here
         case e: ApiException => {
           if(e.getCode != StatusCodes.NotFound.intValue) {
-            logger.warn(s"Unexpected failure deleting workspace in Workspace Manager. Received ${e.getCode}: [${e.getResponseBody}]")
+            logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}. Received ${e.getCode}: [${e.getResponseBody}]")
+            Future.successful()
           }
-          Success(())
+          else throw e
```
andy7i: Looking for a review from AS on this, to make sure this new behavior (throwing exception on non-404s) looks reasonable.

jacmrob: why throw the error here?

jacmrob (commented):
mostly LG,  just one question

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: I think this may be a bug. The old behavior would return success on ANY exception. I think the intention was probably to return success ONLY on 404. We were hiding any exceptions that weren't 404s.

davidangb: * is there a reason to move this from a `Try` to a `Future`? Either works, and if you want to use `Future` for consistency, I understand … is there some other reason?
* the logic as you have written it is incorrect; you are swallowing all non-404 errors and rethrowing all 404s. This is backwards - we should swallow 404s. I think this is why you have so many test failures.
* you're correct that you're making this part of the code stricter. As originally written, we swallowed ALL errors but only logged the non-404s; now we will throw on non-404s (after fixing the previous bullet point). This feels like a good change, now that workspace manager has been live and stable for a while.

davidangb (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: how about this for the `recoverWith` handler?
```
        //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle 404 exceptions here
        case e: ApiException if e.getCode == StatusCodes.NotFound.intValue => {
          Future.successful()
        }
        case t:Throwable => {
          logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}`", t)
          throw t
        }
```

andy7i: love it, thank you!

> the logic as you have written it is incorrect; you are swallowing all non-404 errors and rethrowing all 404s. This is backwards - we should swallow 404s. I think this is why you have so many test failures.

I missed the `!=` in the original code :/

> is there a reason to move this from a Try to a Future? Either works, and if you want to use Future for consistency, I understand … is there some other reason?

Rob and I looked at this code together and explained that we shouldn't be mixing Trys (which are synchronous) and Futures (which are async) if possible.

andy7i: updated this logic based on David An's recommendation

davidangb (approved):
thank you!

andy7i:
~pushed up changes and they appear on the [branch](https://github.com/broadinstitute/rawls/tree/aj--logging-delete-workspace-failures--ca-1464) but not in the PR probably due to a github outage https://www.githubstatus.com/incidents/rmfrw9dfbtbp~
the missing commit appears above this comment now

jacmrob (approved):

davidangb (changes requested):
changing 404 to 401 on the delete-workspace call to WSM is not correct. WSM returns a 404 if the workspace is not found.

In the example you used, you reference a 401 code that includes HTML in its response. HTML responses are only returned by Apache. This is a sign that something else was wrong in the fiab.

davidangb:
wait … WSM isn't even present in FiaBs. So the FiaB tests are calling … some instance of WSM that isn't meant to be called by tests. I'm not surprised there's lots of failures.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -348,26 +348,51 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
-      _ = Try(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
+      _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, userInfo.accessToken)).recoverWith {
         //this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle all exceptions here
         case e: ApiException => {
           if(e.getCode != StatusCodes.NotFound.intValue) {
-            logger.warn(s"Unexpected failure deleting workspace in Workspace Manager. Received ${e.getCode}: [${e.getResponseBody}]")
+            logger.warn(s"Unexpected failure deleting workspace (while deleting in Workspace Manager) for workspace `${workspaceName}. Received ${e.getCode}: [${e.getResponseBody}]")
           }
-          Success(())
+          Future.successful()
```
andy7i, line 380: @davidangb reverted the behavior here back so it logs exceptions but does not throw exceptions.

davidangb (approved):
code looks good, :+1: assuming tests pass. Thanks for sticking with this!

PR #1499, opened 2021-08-13
CA-1474 fix clone Workspace

gpolumbo-broad:
Ticket: [CA-1474](https://broadworkbench.atlassian.net/browse/CA-1474)
When creating/cloning a Workspace, the user can specify the bucket location/region.  When cloning a Workspace, if a user does not specify a bucket location, the bucket should be created in the same region as the Workspace being cloned.  To determine that region, we need to query Google and this query costs money, so we want to charge the Google Project for the target/cloned Workspace.  This introduces some timing issues due to when the bucket location is queried, when the bucket is created, and when we get a Google Project from RBS.  

TODO: Why didn't tests catch this bug?  We should have tests written that test this field on create/clone Workspace. https://broadworkbench.atlassian.net/browse/CA-1477

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -134,7 +134,7 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
         traceWithParent("withBillingProjectContext", s1)(s2 => withBillingProjectContext(workspaceRequest.namespace, s2) { billingProject =>
           for {
             workspace <- traceWithParent("withNewWorkspaceContext", s2) (s3 => dataSource.inTransactionWithAttrTempTable(Set(AttributeTempTableType.Workspace))({ dataAccess =>
-              withNewWorkspaceContext(workspaceRequest, billingProject, dataAccess, s3) { workspaceContext =>
+              withNewWorkspaceContext(workspaceRequest, billingProject, sourceBucketName = None, dataAccess, s3) { workspaceContext =>
                 DBIO.successful(workspaceContext)
               }
             }, TransactionIsolation.ReadCommitted)) // read committed to avoid deadlocks on workspace attr scratch table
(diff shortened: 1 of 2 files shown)


gpolumbo-broad:
Github won't let me thumb this because I created it originally.  But here's my text based thumb.

marctalbott (approved):
Greg's proxy thumb

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2175,6 +2182,20 @@ class WorkspaceService(protected val userInfo: UserInfo, val dataSource: SlickDa
...
 
+  // A new workspace request may specify the region where the bucket should be created. In the case of cloning a
+  // workspace, if no bucket location is provided, then the cloned workspace's bucket will be created in the same region
+  // as the source workspace's bucket. Rawls does not store bucket regions, so in order to get this information we need
+  // to query Google and this query costs money, so we need to make sure that the target Google Project is the one that
+  // gets charged. If neither a bucket location nor a source bucket name are provided, Future[None] is returned which
+  // will result in the default bucket location being used
+  private def determineWorkspaceBucketLocation(maybeBucketLocation: Option[String], maybeSourceBucketName: Option[String], googleProjectId: GoogleProjectId): Future[Option[String]] = {
+    (maybeBucketLocation, maybeSourceBucketName) match {
+      case (bucketLocation@Some(_), _) => Future(bucketLocation)
```
andy7i, line 2193: would 
```
case (Some(bucketLocation), _) => Future(bucketLocation)
```
not work here?

andy7i (approved):
👍 , 1 non-blocking question

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: Nope, the `@` symbol sets the variable name before it to be equal to whatever matches the pattern after it. So in the case of `bucketLocation@Some(_)`, `bucketLocation` is of type `Option[String]` which is the desired type. 

with
```
case (Some(bucketLocation), _) => Future(bucketLocation)
```

`bucketLocation` would be type `String` and we'd be returning `Future[String]` rather than `Future[Option[String]]`

andy7i: Ok that makes sense. So if I'm understanding it correctly, would this work then?

```
case (Some(bucketLocation), _) => Future(Option(bucketLocation))
```

marctalbott: Yes, that would work as well

PR #1500, opened 2021-08-20
[CA-1482] Move service perimeter update into withNewWorkspaceContext

marctalbott:
Ticket: [CA-1482](https://broadworkbench.atlassian.net/browse/CA-1482)
`cloneWorkspace` does not work at the moment for workspaces that are in a service perimeter. This is because we attempt to check the source workspace bucket's location using the destination workspace's Google project before the destination workspace's Google project has been added to the service perimeter. The service perimeter blocks this access as external access. To remedy this, we need to add the destination workspace's Google project to the service perimeter earlier in the creation/clone process. Unfortunately, this means the update perimeter GCP operation needs to reside within the `withNewWorkspaceContext` transaction. I'm not aware of a way that we can get around this without significant refactoring unfortunately.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -751,7 +751,9 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
           dataAccess.rawlsBillingProjectQuery.updateBillingProjects(Seq(billingProject.copy(servicePerimeter = Option(servicePerimeterName), googleProjectNumber = Option(googleProjectNumber))))
         }
 
-        _ <- servicePerimeterService.overwriteGoogleProjectsInPerimeter(servicePerimeterName)
+        _ <- dataSource.inTransaction { dataAccess =>
+          servicePerimeterService.overwriteGoogleProjectsInPerimeter(servicePerimeterName, dataAccess)
+        }
       } yield RequestComplete(StatusCodes.NoContent)
     }
   }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -751,7 +751,9 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
           dataAccess.rawlsBillingProjectQuery.updateBillingProjects(Seq(billingProject.copy(servicePerimeter = Option(servicePerimeterName), googleProjectNumber = Option(googleProjectNumber))))
         }
 
-        _ <- servicePerimeterService.overwriteGoogleProjectsInPerimeter(servicePerimeterName)
+        _ <- dataSource.inTransaction { dataAccess =>
+          servicePerimeterService.overwriteGoogleProjectsInPerimeter(servicePerimeterName, dataAccess)
+        }
```
andy7i, line 756: I like the idea of putting all of this into `dataSource.inTransaction`!

`retryUntilSuccessOrTimeout` is called in `overwriteGoogleProjectsInPerimeter` which may take a while. Could this result in some db locks?

andy7i (approved):
1 question but not blocking. 👍 

(Solidifying my own understanding:) We're in this pickle because the query to find the bucket location costs money and we want to charge that to the destination google project.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: Yeah, having it reside inside of a transaction is not ideal at all. Unfortunately, `withNewWorkspaceContext` as a whole resides within a transaction, and this now needs to happen inside of that function so I'm don't know that we have a better option. I suppose we could have two versions of `overwriteGoogleProjectsInPerimeter`, one that runs inside of a transaction for use in `withNewWorkspaceContext` and one that is identical to how it used to be for use in this case. Don't know that that's really worth it tbh, but open to feedback

andy7i: I think as long as perf tests are still ok, I'm ok with this.

gpolumbo-broad (approved):
Bums me out that we had to move this inside the transaction, but I totally understand why and given where we're at in terms of time, I think it's the right move :(

PR #1506, opened 2021-08-27
[CA-1480] Fix workflows in v1 workspaces

marctalbott:
Ticket: [CA-1480](https://broadworkbench.atlassian.net/browse/CA-1480)
We need to use different Cromwell backends depending on what kind of VPC network the workspace's Google project has.

See corresponding `firecloud-develop` PR: https://github.com/broadinstitute/firecloud-develop/pull/2642

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -316,8 +316,8 @@ object Boot extends IOApp with LazyLogging {
         conf.getBoolean("executionservice.useWorkflowCollectionField")
       val useWorkflowCollectionLabel =
         conf.getBoolean("executionservice.useWorkflowCollectionLabel")
-      val defaultNetworkBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultNetworkBackend"))
-      val highSecurityNetworkBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.highSecurityNetworkBackend"))
+      val defaultNetworkCromwellBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultNetworkBackend"))
+      val highSecurityNetworkCromwellBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.highSecurityNetworkBackend"))
 
       val wdlParsingConfig = WDLParserConfig(conf.getConfig("wdl-parsing"))
       def cromwellSwaggerClient = new CromwellSwaggerClient(wdlParsingConfig.serverBasePath)
@@ -473,8 +473,8 @@ object Boot extends IOApp with LazyLogging {
           requesterPaysRole,
           useWorkflowCollectionField,
           useWorkflowCollectionLabel,
-          defaultNetworkBackend,
-          highSecurityNetworkBackend,
+          defaultNetworkCromwellBackend,
+          highSecurityNetworkCromwellBackend,
           methodConfigResolver
         )
       } else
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -316,7 +316,8 @@ object Boot extends IOApp with LazyLogging {
         conf.getBoolean("executionservice.useWorkflowCollectionField")
       val useWorkflowCollectionLabel =
         conf.getBoolean("executionservice.useWorkflowCollectionLabel")
-      val defaultBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultBackend"))
+      val defaultNetworkBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultNetworkBackend"))
+      val highSecurityNetworkBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.highSecurityNetworkBackend"))
```
ehigham: Why are these values bound so far away from their use? Are you evaluating eagerly for effects? Can we not just inline the call to their use?

ehigham (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -316,7 +316,8 @@ object Boot extends IOApp with LazyLogging {
         conf.getBoolean("executionservice.useWorkflowCollectionField")
       val useWorkflowCollectionLabel =
         conf.getBoolean("executionservice.useWorkflowCollectionLabel")
-      val defaultBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultBackend"))
+      val defaultNetworkBackend: CromwellBackend = CromwellBackend(conf.getString("executionservice.defaultNetworkBackend"))
```
s-rubenstein: I like the name change here. Is there a reason we don't specify `Cromwell` in the variable name? Just thinking about clarity of purpose here, since it is specifically for Cromwell?

s-rubenstein (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
marctalbott: In this case, I was just following the convention that had already been set and didn't put any additional thought into it. That said, in my opinion there is value to isolating the config to Boot and subsequently passing those values around as the appropriate case classes (or more nested `Config` case classes) rather than passing strings around that can be used incorrectly.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
marctalbott: I don't know if there's any particular reason. It is common in Rawls to refer to the execution service rather than Cromwell, I suppose to keep the execution service abstract, but that obviously hasn't been followed here with the `CromwellBackend` type. I don't have a preference on the naming, I can change it if you prefer `defaultNetworkCromwellBackend` and `highSecurityNetworkCromwellBackend`

s-rubenstein: I think I would prefer specificity, especially because 'NetworkBackend' could be anything. I would also be okay with `ExecutionServiceNetworkBackend` but that starts to get long.

Would prefer either cromwell or executionService be included, whichever you prefer

marctalbott:
Successful swat run with this branch and the fc-dev branch linked in the PR description: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/53672/

PR #1335, opened 2020-12-21
CA-1095 disable requester pays when deleting workspace

dvoet:
- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [ ] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
  [alpha](https://console.cloud.google.com/sql/instances/terraform-r4caezzc35c4tb7pgdhwkmme4y/backups?project=broad-dsde-alpha&organizationId=548622027621)
  DBs in Google Cloud Console
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] Tell your tech lead (TL) that the PR exists if they want to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Assign to submitter** to finalize
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Squash commits and merge to develop
[...]

PR #1507, opened 2021-08-27
[CA-1489] Service perimeter project leak in workspace deletion  

andy7i:
CA-1489

Moves the workspace deletion to AFTER the google project is deleted. This means we have 2 DB transactions, but I believe this is necessary as we shouldn't delete the google project before aborting workflows (according to the Batch team: https://broadinstitute.slack.com/archives/C1EH66VCM/p1630088773005300). Otherwise, the initial commit (which was reverted) would be sufficient to prevent the service perimeter leak.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


andy7i:
jenkins retest

andy7i:
jenkins retest

andy7i:
jenkins retest -- orch PFBImportSpec 500s seem unrelated

andy7i:
jenkins retest -- both tests used the same fiab (fiab-gpolumbo-caring-chimp) which has issues standing up import service:
```
Resource not found (resource=import-service-notify-sub-fiab-gpolumbo-caring-chimp).","grpc_status":5}"
```
https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-stop/71652/artifact/logs/logs-20210831_125855/firecloud_import-service-app_1.log

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,12 +365,29 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  private def deleteWorkspaceTransaction(workspaceName: WorkspaceName, workspaceContext: Workspace) = {
+    dataSource.inTransaction { dataAccess =>
+      for {
+        // Delete components of the workspace
+        _ <- dataAccess.submissionQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
+        _ <- dataAccess.methodConfigurationQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
+        _ <- dataAccess.entityQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
+
+        // Delete the workspace
+        _ <- dataAccess.workspaceQuery.delete(workspaceName)
```
gpolumbo-broad: nit: I would make this call the last one in the transaction.  Sort of like, "delete all the stuff in the box, then delete the box".  As opposed to, "delete most of the stuff in the box, then delete the box, and then if there was anything that fell on the floor, delete that too."  

It's just a nit though, because the bigger question is really about when we call `deleteWorkspaceTransaction` in general compared to the calls to Sam and WSM etc.  But that's a bigger question maybe outside the scope of this ticket.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,12 +365,29 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+    }
+  }
+
   private def deleteWorkspace(workspaceName: WorkspaceName, workspaceContext: Workspace): Future[PerRequestMessage] = {
     //Attempt to abort any running workflows so they don't write any more to the bucket.
     //Notice that we're kicking off Futures to do the aborts concurrently, but we never collect their results!
     //This is because there's nothing we can do if Cromwell fails, so we might as well move on and let the
     //ExecutionContext run the futures whenever
-    val deletionFuture: Future[Seq[WorkflowRecord]] =
+    val abortWorkflowFuture: Future[Seq[WorkflowRecord]] =
```
gpolumbo-broad: Hooray for descriptive naming!  After reading it further though, might I suggest something like:  

```suggestion
    val gatherWorkflowsToAbort: Future[Seq[WorkflowRecord]] =
```

Optional:  I know there seems to be a convention in Rawls to use nested functions.  I personally feel like they complicate things.  I find shorter, well named functions easier to read/understand/follow.  So I might suggest extracting this out into its own private method, but totally up to you.  (for example, with the indenting and all the curly braces, I had a hard time telling in Github where the function ends)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -387,26 +404,14 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
             workflowsToAbort
           }
         }
       }
     for {
 
-      workflowsToAbort <- deletionFuture recoverWith {
+      workflowsToAbort <- abortWorkflowFuture recoverWith {
         case t:Throwable => {
           logger.warn(s"Unexpected failure deleting workspace (while running `deletionFuture`) for workspace `${workspaceName}`", t)
```
gpolumbo-broad: Update the log message, something like:

```suggestion
          logger.warn(s"Unexpected failure deleting workspace (while gathering workflows that need to be aborted) for workspace `${workspaceName}`", t)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,12 +365,29 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  }
+
   private def deleteWorkspace(workspaceName: WorkspaceName, workspaceContext: Workspace): Future[PerRequestMessage] = {
     //Attempt to abort any running workflows so they don't write any more to the bucket.
     //Notice that we're kicking off Futures to do the aborts concurrently, but we never collect their results!
     //This is because there's nothing we can do if Cromwell fails, so we might as well move on and let the
     //ExecutionContext run the futures whenever
-    val deletionFuture: Future[Seq[WorkflowRecord]] =
+    val abortWorkflowFuture: Future[Seq[WorkflowRecord]] =
       requesterPaysSetupService.revokeAllUsersFromWorkspace(workspaceContext).flatMap { _ =>
```
gpolumbo-broad: I know this is out of scope for this issue, but not sure why this needs to be in this method.  Consider pulling it out into its own line in the calling `for` comp.  (but 100% OK if you just want to leave it for now, this is not what this ticket is about, but if we're refactoring and cleaning up the delete flow, it's something we could do to improve overall quality)

gpolumbo-broad (changes requested):

gpolumbo-broad:
As for the multiple transactions, I think that's ok here.  One of the things we want to fix in Rawls is isolating DB transactions from external API calls, and in order to do that properly, we need to think pretty seriously about how to keep state consistent across services.  Again, that's a bigger discussion, but I don't think you've made things any worse here, having multiple transactions is not in and of itself a bad thing.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,6 +365,28 @@ class WorkspaceService(protected val userInfo: UserInfo,
     }
   }
 
+  private def gatherWorkflowsToAbortAndSetStatusToAborted(workspaceName: WorkspaceName, workspaceContext: Workspace) = {
```
gpolumbo-broad, line 368: Love it!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -387,33 +409,19 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
-                  dataAccess.workflowQuery.updateStatus(wf, WorkflowStatuses.Aborted) { status =>
-                    if (config.trackDetailedSubmissionMetrics) Option(workflowStatusCounter(workspaceSubmissionMetricBuilder(workspaceName, wf.submissionId))(status))
-                    else None
-                  }
-              }
-            }
-          } yield {
-            workflowsToAbort
-          }
+    val gatherWorkflowsFuture: Future[Seq[WorkflowRecord]] = gatherWorkflowsToAbortAndSetStatusToAborted(workspaceName, workspaceContext)
```
gpolumbo-broad: Was there a conscious reason why this call to `gatherWorkflowsToAbortAndSetStatusToAborted` was chosen to be run outside of the `for` comp?  I know this happened before the `for` comp before, but looking at this now, I probably would have opted to put this call inside now.  

Functionally, as written now, `gatherWorkflowsToAbortAndSetStatusToAborted` will be started and then processing will immediately proceed (without waiting for the future to complete) into the `for` comp and start `revokeAllUsersFromWorkspace`, so that those two futures are running in parallel.  By moving this line inside the `for` comp, you would be lining them up to run in sequence, which is how it functioned before anyway.  And if we really wanted to keep things ordered the same, the `revokeAllUsersFromWorkspace` should be above the call to `gatherWorkflowsToAbortAndSetStatusToAborted`.  

So I've talked myself into it.  I would say move `gatherWorkflowsToAbortAndSetStatusToAborted` inside the `for` comp and make it second call after `revokeAllUsersFromWorkspace`.   Unless, of course, you had a conscious reason why you have it as written.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -387,33 +409,19 @@ class WorkspaceService(protected val userInfo: UserInfo,
     //Notice that we're kicking off Futures to do the aborts concurrently, but we never collect their results!
     //This is because there's nothing we can do if Cromwell fails, so we might as well move on and let the
     //ExecutionContext run the futures whenever
```
gpolumbo-broad: nit: move this comment down by the line where we actually call `executionServiceCluster.abort`

andy7i:
jenkins retest -- unrelated error (for a 1-line logging change) -- 

```

18:09:13 [error] java.io.IOException: Server returned HTTP response code: 403 for URL: https://broadinstitute.jfrog.io/broadinstitute/libs-release-local;build.timestamp=1630534106/org/broadinstitute/dsde/rawls-model_2.12/0.1-9d51546b-SNAP/rawls-model_2.12-0.1-9d51546b-SNAP.pom
```

gpolumbo-broad (changes requested):
I personally am really digging this refactoring!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: yeah, the original code has `val deletionFuture` defined outside of the for comp, so I was just keeping it there.

I see your point about keeping `revokeAllUsersFromWorkspace` and `gatherWorkflowsToAbortAndSetStatusToAborted` in the same order. Looking at the git history though I don't think the order particularly matters
https://github.com/broadinstitute/rawls/pull/1335
https://broadworkbench.atlassian.net/browse/CA-1095

I can't start the for comp with an assignment (`=`), so I'll put it under `revokeAllUsersFromWorkspace` as you're suggesting

andy7i (commented):
@gpolumbo-broad thanks for the thorough review / ideas about refactoring!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,55 +365,65 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
-                  }
+  private def gatherWorkflowsToAbortAndSetStatusToAborted(workspaceName: WorkspaceName, workspaceContext: Workspace) = {
+    dataSource.inTransaction { dataAccess =>
+      for {
+        // Gather any active workflows with external ids
+        workflowsToAbort <- dataAccess.workflowQuery.findActiveWorkflowsWithExternalIds(workspaceContext)
+
+        //If a workflow is not done, automatically change its status to Aborted
+        _ <- dataAccess.workflowQuery.findWorkflowsByWorkspace(workspaceContext).result.map { recs =>
+          recs.collect {
```
ehigham: Would this benefit from splitting into distinct `filter` and `foreach` calls?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,55 +365,65 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+            case wf if !WorkflowStatuses.withName(wf.status).isDone =>
+              dataAccess.workflowQuery.updateStatus(wf, WorkflowStatuses.Aborted) { status =>
+                if (config.trackDetailedSubmissionMetrics) Option(workflowStatusCounter(workspaceSubmissionMetricBuilder(workspaceName, wf.submissionId))(status))
```
ehigham: You're not doing anything with the result of `collect` - are these functions (that all read like nouns) intended for side effects??

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,55 +365,65 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  }
 
-          } yield {
-            workflowsToAbort
-          }
+  private def deleteWorkspace(workspaceName: WorkspaceName, workspaceContext: Workspace): Future[PerRequestMessage] = {
+    for {
+      _ <- requesterPaysSetupService.revokeAllUsersFromWorkspace(workspaceContext) recoverWith {
+        case t:Throwable => {
+          logger.warn(s"Unexpected failure deleting workspace (while revoking 'requester pays' users) for workspace `${workspaceName}`", t)
```
ehigham, line 411: Can you also log the workspace name and/or context?

ehigham: Also the "unexpected..." stuff is rather verbose. Doesn't logging give us where the thing was logged? How about
`"Failed to revoke all users from workspace"`, then include which workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,55 +365,65 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
       }
 
-      // Abort running workflows
+      //Attempt to abort any running workflows so they don't write any more to the bucket.
+      //Notice that we're kicking off Futures to do the aborts concurrently, but we never collect their results!
+      //This is because there's nothing we can do if Cromwell fails, so we might as well move on and let the
+      //ExecutionContext run the futures whenever
       aborts = Future.traverse(workflowsToAbort) { wf => executionServiceCluster.abort(wf, userInfo) } recoverWith {
         case t:Throwable => {
           logger.warn(s"Unexpected failure deleting workspace (while aborting workflows) for workspace `${workspaceName}`", t)
```
ehigham, line 429: Why bother even logging this? Since we're doing this asynchronously and don't care about the outcome, why log it?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -422,9 +432,22 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+          logger.error(s"Unexpected failure deleting workspace (while deleting google project) for workspace `${workspaceName}`", t)
+          throw t
+        }
+      }
+
+      // Delete the workspace records in Rawls. Do this after deleting the google project to prevent service perimeter leaks.
+      _ <- deleteWorkspaceTransaction(workspaceName, workspaceContext) recoverWith {
+        case t:Throwable => {
+          logger.error(s"Unexpected failure deleting workspace (while deleting workspace in Rawls DB) for workspace `${workspaceName}`", t)
+          throw t
```
ehigham, line 446: And then it'll get caught and logged again? What's the point of all this try then log stuff? If the answer is to have context, we'll have that already from the stack traces.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: looks like they are used for metrics

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: `WorkspaceName` contains billing-project/workspace, which I think is all we need from here.

the reason for the verbose logging is for ease of search. We have log sinks to catch all of these errors starting with `Unexpected failure deleting` (CA-1470) to help with debugging for CA-1206.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: This probably is unnecessary but it was added in to try to log any issues in deleteWorkspace, again to help with debugging the root cause of CA-1206

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: I believe so, yeah. This is existing code (that I moved). There's a lot that could be refactored in deleteWorkspace in general. I'll make the change here though.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -365,55 +365,65 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  private def gatherWorkflowsToAbortAndSetStatusToAborted(workspaceName: WorkspaceName, workspaceContext: Workspace) = {
+    dataSource.inTransaction { dataAccess =>
+      for {
+        // Gather any active workflows with external ids
+        workflowsToAbort <- dataAccess.workflowQuery.findActiveWorkflowsWithExternalIds(workspaceContext)
+
+        //If a workflow is not done, automatically change its status to Aborted
+        _ <- dataAccess.workflowQuery.findWorkflowsByWorkspace(workspaceContext).result.map { workflowRecords =>
+          val notDoneWorkflowRecords = workflowRecords.filter(workflowRecord => !WorkflowStatuses.withName(workflowRecord.status).isDone)
+          notDoneWorkflowRecords.foreach { workflowRecord =>
```
ehigham: Thanks for doing this! Is it possible to chain these operations without the intermediate assignment though? Something like:
```scala
result.map {  rs => rs
  .filter { !WorkflowStatuses.withName(_.status).isDone }
  .foreach { r => .. }
}
```

(ideally, I'd eliminate the `rs` altogether)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ehigham: Thanks for the explanation.

ehigham (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ehigham: I'm not a massive fan of this amount of log-then-throw stuff (since it'll probably get logged again further up the stack). But I won't drag hold you up about it.

andy7i:
jenkins retest - a tag was pushed but then couldn't be found
```
14:43:41 Successfully tagged broadinstitute/rawls-tests:b8bf6abfef17
14:43:41 pushing broadinstitute/rawls:b8bf6abfef17...
14:43:41 The push refers to repository [docker.io/broadinstitute/rawls]
14:43:41 0d4adc09e36d: Preparing
14:43:41 8b90693284e9: Preparing
14:43:41 f0028f89bbf6: Preparing
14:43:41 36bf8e204893: Preparing
14:43:41 5945aff80f6f: Preparing
14:43:41 7555a8182c42: Preparing
14:43:41 7555a8182c42: Waiting
14:43:41 f0028f89bbf6: Layer already exists
14:43:41 5945aff80f6f: Layer already exists
14:43:41 36bf8e204893: Layer already exists
14:43:41 7555a8182c42: Layer already exists
14:43:43 8b90693284e9: Pushed
14:44:48 0d4adc09e36d: Pushed
14:44:50 b8bf6abfef17: digest: sha256:6d1d345cd9b9e2e22c0fa4154fb553f5a1f263694c71953716a46352be17e0de size: 1583
14:44:50 Error response from daemon: No such image: broadinstitute/rawls:b8bf6abfef17
14:44:50 Build step 'Conditional steps (multiple)' marked build as failure
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: i thought you might request that 🙃 
I made the change as requested, but I think it might be harder to read now. Thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
andy7i: yeah that's understandable. it helps a lot with viewing what is causing issues with workspace deletion at a glance though! I think it's the lesser evil right now rather than having to dive into logs for every workspace deletion error to figure out what went wrong. (looking through doug's log sinks, most of the errors logged there do not have the stack trace bubbled up)

andy7i:
jenkins retest -- merged dev into v2_workspaces to fix leo tests

andy7i:
jenkins retest - leo suite aborted

PR #1515, opened 2021-09-08
unit tests should use utf8 database [AS-927]

davidangb:
Part of AS-927 but does not complete that ticket.

Tell our unit tests to set `character_set_server=utf8` on their mysql instance.

This makes our unit tests consistent with our live environments, which is always a good thing. Additionally, that utf8 setting is a critical prerequisite for any SQL queries that attempt to override a collation, since not all collations work with all character sets. Specifically, this setting is required to fix AS-927. Without this setting, unit tests for AS-927 would fail, even though the runtime code would work correctly against our live Cloud SQL instances.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,22 +11,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-#    services:
-#      mysql:
-#        image: mysql/mysql-server:5.6
-#        env:
-#          MYSQL_ROOT_PASSWORD: rawls-test
-#          MYSQL_USER: rawls-test
-#          MYSQL_PASSWORD: rawls-test
-#          MYSQL_DATABASE: testdb
-#        options: >-
-#          --health-cmd="mysqladmin ping"
-#          --health-interval 10s
-#          --health-timeout 5s
-#          --health-retries 5
-#        ports:
-#          - 3306:3306
-
     steps:
 
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,9 @@ jobs:
 
     steps:
 
+      - name: Shutdown Ubuntu MySQL (SUDO)
+        run: sudo service mysql stop # Shutdown the Default MySQL, "sudo" is necessary, please not remove it
+
       # run this step first to give it time to start up
       - name: Set up MySQL
         uses: mirromutth/mysql-action@v1.1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
 
     services:
       mysql:
-        image: mysql/mysql-server:5.6
+        image: 'mysql/mysql-server:5.6 --character-set-server=utf8'
         env:
           MYSQL_ROOT_PASSWORD: rawls-test
           MYSQL_USER: rawls-test
@@ -24,7 +24,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-          --character-set-server=utf8
         ports:
           - 3306:3306
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,7 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+          --character-set-server=utf8
         ports:
           - 3306:3306
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,7 +60,7 @@ jobs:
           sbt clean coverage test coverageReport
           -J-Xmx3g
           -Denv.type=test
-          -Dmysql.host=localhost
+          -Dmysql.host=127.0.0.1
           -Dmysql.port=3306
           it:compile
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,22 +11,27 @@ jobs:
 
     runs-on: ubuntu-latest
 
+#    services:
+#      mysql:
+#        image: mysql/mysql-server:5.6
+#        env:
+#          MYSQL_ROOT_PASSWORD: rawls-test
+#          MYSQL_USER: rawls-test
+#          MYSQL_PASSWORD: rawls-test
+#          MYSQL_DATABASE: testdb
+#        options: >-
+#          --health-cmd="mysqladmin ping"
+#          --health-interval 10s
+#          --health-timeout 5s
+#          --health-retries 5
+#        ports:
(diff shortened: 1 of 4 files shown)


davidangb:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
```diff
@@ -987,7 +987,7 @@ class SubmissionApiServiceSpec extends ApiServiceSpec with TableDrivenPropertyCh
     }
   }
 
-  private val userComment1000character = RandomStringUtils.random(1000)
+  private val userComment1000character = RandomStringUtils.randomGraph(1000)
```
davidangb, line 990: Before this change, the test could fail with something like:
```
o.b.d.r.webservice.RawlsApiService$ - Incorrect string value: '\xF0\xA2\x8F\x86\xF0\x90...' for column 'USER_COMMENT' at row 1
```

The description of `random` says "Characters will be chosen from the set of all characters."

See https://stackoverflow.com/questions/10957238/incorrect-string-value-when-trying-to-insert-utf-8-into-mysql-via-jdbc … the "set of all characters" can, I assume, include 4-byte characters. I assume this would fail on production too.

TBH, I'm not sure how it was passing before!

davidangb:
@salonishah11 I tagged you for review due to the changes in `SubmissionApiServiceSpec`, please let me know if you don't have the bandwidth. Thank you!

.github/workflows/test.yml
```diff
@@ -11,22 +11,6 @@ jobs:
...
-          MYSQL_PASSWORD: rawls-test
-          MYSQL_DATABASE: testdb
-        options: >-
-          --health-cmd="mysqladmin ping"
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 3306:3306
-
```
davidangb, line 29: this method of starting mysql doesn't allow flags such as `--character-set-server=utf8` (or at least I couldn't figure it out) so I abandoned it in favor of reusing the `docker/run-mysql.sh` script.

.github/workflows/test.yml
```diff
@@ -59,7 +47,7 @@ jobs:
           -J-Xmx3g
           -Denv.type=test
           -Dmysql.host=localhost
-          -Dmysql.port=3306
+          -Dmysql.port=3310
```
davidangb, line 50: `docker/run-mysql.sh` uses 3310 instead of 3306, therefore this change

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessSpec.scala
```diff
@@ -70,4 +70,21 @@ class DataAccessSpec extends TestDriverComponentWithFlatSpecAndMatchers with Sca
...
+       A failure of this test likely means that our test environment is not set up correctly; the mysql
+       instance used by tests does not have the right setting. If the test-instance mysql is not set up
+       correctly, other tests can fail and/or return false positives.
+     */
+    val charsetLookup = runAndWait(sql"""SHOW VARIABLES WHERE Variable_name = 'character_set_server';""".as[(String, String)])
+    charsetLookup should have size 1
+    withClue("is the mysql against which these unit tests ran set up correctly with --character-set-server=utf8 or equivalent?") {
+      charsetLookup.head._2 shouldBe "utf8"
+    }
+  }
```
davidangb, line 89: I verified that this test fails ("latin1 was not equal to utf8") without the changes to `docker/run-mysql.sh`

docker/run-mysql.sh
```diff
@@ -10,7 +10,7 @@ start() {
 
     # start up mysql
     echo "starting up mysql container..."
-    docker run --name $CONTAINER -e MYSQL_ROOT_HOST='%' -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d -p 3310:3306 mysql/mysql-server:$MYSQL_VERSION
+    docker run --name $CONTAINER -e MYSQL_ROOT_HOST='%' -e MYSQL_ROOT_PASSWORD=rawls-test -e MYSQL_USER=rawls-test -e MYSQL_PASSWORD=rawls-test -e MYSQL_DATABASE=testdb -d -p 3310:3306 mysql/mysql-server:$MYSQL_VERSION --character-set-server=utf8
```
davidangb, line 13: this is the fundamental change in the PR: make sure our test mysql instance uses utf8 character set.

.github/workflows/test.yml
ungwudik: It looks like the port used in `run-mysql.sh` is `3310:3306` which is a little different from the one here `3306:3306`.  Does `run-mysql.sh` need to be updated, or is the discrepancy negligible?

.github/workflows/test.yml
ungwudik: ahh I think this explains my other comment?

davidangb:
one unit test failure in Jenkins: `should attach outputs and not deadlock with multiple submissions all updating the same entity at once`. I haven't seen this happen before, including all the other test runs on this PR. Transient? Jenkins retest.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
ungwudik: hmm did the test fail consistently before this change? And how does that consistency compare to test runs _after_ the change?

sagehen03 (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
davidangb: the test passed consistently before (else we would have blocked PRs), when it was inserting into a `latin1` collation. Now that it's inserting into `utf8` it's failing, which I believe is due to the 3-byte restriction for `utf8` collation. Something about the `latin1` collation must have allowed it, but we need `utf8` for other reasons.

.github/workflows/test.yml
davidangb: yes, I thought it was better to update the GHA tests to use the same port as `docker/run-mysql.sh` instead of the other way around, because the change is more isolated

ungwudik (approved):

salonishah11 (approved):
LGTM 👍

PR #1516, opened 2021-09-08
simplify filteredCount query for paginated entity results [AS-925]

davidangb:
AS-925

See also doc at https://docs.google.com/document/d/1NKvxZbZocV2At_uyuUKn3AFv6raoDlJ0ziGKXyscBes/edit?usp=sharing

This bug surfaced under the following conditions:
* the entity data contained multiple attributes whose names differed only in case; e.g. one entity had an attribute "foo" and another had an attribute "Foo"
* you performed an entity query that sorted by one of these columns, e.g. sorted by "Foo"

Under these conditions, the "filteredCount" value returned in the result metadata was incorrect.

In this PR, I have significantly simplified the SQL query used to generate the "filteredCount" value, and added unit tests.

OLD QUERY:
```sql
select count(1) from (
	select
		e.id,
		e.name,
		e.all_attribute_values,
		sort_a.list_length as sort_list_length,
		sort_a.value_string as sort_field_string,
		sort_a.value_number as sort_field_number
		 sort_a.value_boolean as sort_field_boolean,
		 sort_a.value_json as sort_field_json,
		 sort_e_ref.name as sort_field_ref
	from ENTITY e
		left outer join ENTITY_ATTRIBUTE sort_a on sort_a.owner_id = e.id and sort_a.name = ? and ifnull(sort_a.list_index, 0) = 0
		left outer join ENTITY sort_e_ref on sort_a.value_entity_ref = sort_e_ref.id 
	where
		e.deleted = 'false'
		and e.entity_type = ? and e.workspace_id = ? )
	pagination
	where
		concat(pagination.name, ' ', pagination.all_attribute_values) like ?
```

NEW QUERY:
```sql
select count(1) from ENTITY e
     where e.deleted = 0
     and e.entity_type = ?
     and e.workspace_id = ? and concat(e.name, ' ', e.all_attribute_values) like ?
```

TODO for a future ticket/PR: investigate the other uses of `activeActionForPagination` and `paginationSubquery` to look for any other faulty logic

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -307,19 +307,23 @@ trait EntityComponent {
           sql" limit #${entityQuery.pageSize} offset #${(entityQuery.page-1) * entityQuery.pageSize} ) p on p.id = e.id "
         )
 
-        for {
-          unfilteredCount <- findActiveEntityByType(workspaceContext.workspaceIdAsUUID, entityType).length.result
-          filteredCount <- if (entityQuery.filterTerms.isEmpty) {
-                            // if the query has no filter, then "filteredCount" and "unfilteredCount" will always be the same; no need to make another query
-                            DBIO.successful(Vector(unfilteredCount))
-                           } else {
-                            val filteredQuery =
-                              sql"""select count(1) from ENTITY e
+        def filteredCountQuery: ReadAction[Vector[Int]] = {
+          val filteredQuery =
+            sql"""select count(1) from ENTITY e
                                    where e.deleted = 0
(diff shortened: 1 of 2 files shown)


ungwudik (commented):
I don't understand the implications (if any) of the SQL query change, and I'd rather someone with experience there look. Everything else looks good to me.

sagehen03 (approved):
Thank you for the explanation and the much improved code.   If I understand correctly, `sort_a.name = ?` in one of the left joins of the original query is the culprit behind the weird behavior that the user reported?

davidangb:
@sagehen03 yup, exactly. Since that `name` column uses `utf8_general_ci`, the comparison is case-insensitive. `sort_a.name = 'case'` will return both `case` and `CASE`. There are other fixes for that case-sensitivity issue … but they're totally orthogonal, this code shouldn't need it at all.

ansingh7115 (approved):
I like the simplification of the query!

PR #1520, opened 2021-09-16
standalone entity cache table [AS-929]

davidangb:
AS-929: goal is to move all entity-cache writes away from the `WORKSPACE` table. We should now never need to lock any rows on `WORKSPACE` due to the entity cache.

In this PR:
* create a new `WORKSPACE_ENTITY_CACHE` table with workspace_id, entity_cache_last_updated, error_message columns
* copy all values from `WORKSPACE.entity_cache_last_updated` to `WORKSPACE_ENTITY_CACHE.entity_cache_last_updated`
* drop the `entity_cache_last_updated` column from `WORKSPACE`
* use the new `WORKSPACE_ENTITY_CACHE` table for all cache-timestamp reads and writes
* move SQL queries related to cache-timestamps out of `WorkspaceComponent` and into `EntityCacheComponent`. Because they're joins now, hand-roll them instead of using Slick
* in the case of a cache-update failure:
    * log the error message on the same line as the indication of failure, to make it easy to find in Kibana/Cloud Logging
    * write the error message and stack trace to the `WORKSPACE_ENTITY_CACHE` table for easy debugging after the fact

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
@@ -9,7 +9,7 @@
                              referencedTableName="WORKSPACE" referencedColumnNames="id"
                              foreignKeyName="FK_WEC_WORKSPACE" deleteCascade="true"/>
             </column>
-            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02">
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:01">
                 <constraints nullable="false"/>
             </column>
             <column name="error_message" type="TEXT">
@@ -29,6 +29,8 @@
         </rollback>
     </changeSet>
 
-    <!-- TODO: drop entity_cache_last_updated from WORKSPACE -->
+    <changeSet logicalFilePath="dummy" author="davidan" id="drop_WORKSPACE_entity_cache_last_updated">
+        <dropColumn tableName="WORKSPACE" columnName="entity_cache_last_updated"/>
+    </changeSet>
 
 </databaseChangeLog>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
@@ -9,7 +9,7 @@
                              referencedTableName="WORKSPACE" referencedColumnNames="id"
                              foreignKeyName="FK_WEC_WORKSPACE" deleteCascade="true"/>
             </column>
-            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:01">
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02">
                 <constraints nullable="false"/>
             </column>
             <column name="error_message" type="TEXT">
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="create_WORKSPACE_ENTITY_CACHE">
+        <createTable tableName="WORKSPACE_ENTITY_CACHE">
+            <column name="workspace_id" type="BINARY(16)">
+                <constraints primaryKey="true"
+                             nullable="false"
+                             referencedTableName="WORKSPACE" referencedColumnNames="id"
+                             foreignKeyName="FK_WEC_WORKSPACE" deleteCascade="true"/>
+            </column>
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02">
+                <constraints nullable="false"/>
+            </column>
+            <column name="error_message" type="TEXT">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_WORKSPACE_ENTITY_CACHE">
+        <sql>
+            INSERT INTO WORKSPACE_ENTITY_CACHE(workspace_id, entity_cache_last_updated)
+                SELECT id, entity_cache_last_updated
+                FROM WORKSPACE;
+        </sql>
+        <rollback>
+            DELETE FROM WORKSPACE_ENTITY_CACHE;
+        </rollback>
+    </changeSet>
+
+    <!-- TODO: drop entity_cache_last_updated from WORKSPACE -->
+
+</databaseChangeLog>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
@@ -34,7 +34,7 @@ trait EntityCacheComponent {
       // C. lastModified is before @param cooldownBound, meaning the workspace isn't likely actively being updated
       // D. Ordered by lastModified from oldest to newest. Meaning, return the workspace that was modified the longest ago
 
-      val baseQuery = sql"""SELECT * FROM (
+      val baseQuery = sql"""SELECT id, last_modified, entity_cache_last_updated FROM (
               |  SELECT w.id, w.last_modified, c.entity_cache_last_updated
               |    FROM WORKSPACE w LEFT OUTER JOIN WORKSPACE_ENTITY_CACHE c
               |    on w.id = c.workspace_id
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
@@ -39,7 +39,7 @@ trait EntityCacheComponent {
       |      FROM WORKSPACE w LEFT OUTER JOIN WORKSPACE_ENTITY_CACHE c
       |      on w.id = c.workspace_id
       |      where w.last_modified < $maxModifiedTime
-      |    ) outerJoin
+      |    ) workspacesAndCacheTimes
       |    WHERE entity_cache_last_updated IS NULL or ($minCacheTime < entity_cache_last_updated AND entity_cache_last_updated < last_modified)
         |  ORDER BY last_modified asc
         |  LIMIT 1;
@@ -50,9 +50,7 @@ trait EntityCacheComponent {
 
     // insert if not exist
     def updateCacheLastUpdated(workspaceId: UUID, timestamp: Timestamp, errorMessage: Option[String] = None): ReadWriteAction[Int] = {
-      val dbResult = entityCacheQuery.insertOrUpdate(EntityCacheRecord(workspaceId, timestamp, errorMessage))
-      logger.info(s"******* updated cache record for $workspaceId to $timestamp")
-      dbResult
+      entityCacheQuery.insertOrUpdate(EntityCacheRecord(workspaceId, timestamp, errorMessage))
     }
 
     def isEntityCacheCurrent(workspaceId: UUID): ReadAction[Boolean] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
@@ -37,7 +37,8 @@ trait EntityCacheComponent {
                                 |FROM WORKSPACE w LEFT OUTER JOIN WORKSPACE_ENTITY_CACHE c
                                 |    on w.id = c.workspace_id
                                 |where (c.entity_cache_last_updated > $minCacheTime or c.entity_cache_last_updated is null)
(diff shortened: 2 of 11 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
```diff
@@ -0,0 +1,73 @@
...
+      val baseQuery = sql"""SELECT * FROM (
+              |  SELECT w.id, w.last_modified, c.entity_cache_last_updated
+              |    FROM WORKSPACE w LEFT OUTER JOIN WORKSPACE_ENTITY_CACHE c
+              |    on w.id = c.workspace_id
+              |    where w.last_modified < $maxModifiedTime) workspacesAndCacheTimes
+              |  WHERE entity_cache_last_updated IS NULL
+              |    or ($minCacheTime < entity_cache_last_updated AND entity_cache_last_updated < last_modified)
+              |  ORDER BY last_modified asc
+              |  LIMIT 1;
+              |""".stripMargin.as[(UUID, Timestamp, Option[Timestamp])]
```
davidangb: Suggestions on how to optimize this SQL query?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -52,26 +57,25 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
...
-        dataAccess.workspaceQuery.findMostOutdatedEntityCacheAfter(minCacheTime, maxModifiedTime).flatMap {
-          case Some((workspaceId, lastModified)) =>
+        dataAccess.entityCacheQuery.findMostOutdatedEntityCacheAfter(MIN_CACHE_TIME, maxModifiedTime).flatMap {
+          case Some((workspaceId, lastModified, cacheLastUpdated)) =>
             rootSpan.putAttribute("workspaceId", OpenCensusAttributeValue.stringAttributeValue(workspaceId.toString))
             traceDBIOWithParent("updateStatisticsCache", rootSpan) { _ =>
               DBIO.from(updateStatisticsCache(workspaceId, lastModified).map { _ =>
-                logger.info(s"Updated entity cache for workspace $workspaceId. Cache was ${System.currentTimeMillis() - lastModified.getTime}ms out of date.")
+                val outDated = lastModified.getTime - cacheLastUpdated.getOrElse(MIN_CACHE_TIME).getTime
+                logger.info(s"Updated entity cache for workspace $workspaceId. Cache was ${outDated}ms out of date.")
```
davidangb, line 70: not super necessary for this PR, but tweaking this log entry seemed like a good idea

core/src/test/resources/logback-test.xml
```diff
@@ -30,15 +30,21 @@
 
     <!--
     <logger name="slick.jdbc.JdbcBackend.statement" level="DEBUG" additivity="false">
-        <appender-ref ref="file"/>
-        <appender-ref ref="console"/>
-    </logger>
+        <appender-ref ref="file"/>
+        <appender-ref ref="console"/>
+    </logger>
```
davidangb, line 35: these were non-standard `&ensp;` en size spaces which bugged me so I changed them

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -26,10 +25,14 @@ object EntityStatisticsCacheMonitor {
...
   case object ScheduleDelayedSweep extends EntityStatisticsCacheMessage
+
+  //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
+  val MIN_CACHE_TIME = new Timestamp(1000)
+
 }
 
 class EntityStatisticsCacheMonitorActor(val dataSource: SlickDataSource, val timeoutPerWorkspace: Duration, val standardPollInterval: FiniteDuration, val workspaceCooldown: FiniteDuration)(implicit val executionContext: ExecutionContext) extends Actor with EntityStatisticsCacheMonitor with LazyLogging {
-  import context._
+  import context.setReceiveTimeout
```
davidangb, line 35: `import context._` confused IntelliJ and made the `pipeTo` on line 45 appear in red; I've tweaked this line to remove IntelliJ warnings.

davidangb:
bah; pervasive swat failures in gpalloc and buffer service

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210915_create_entity_cache_table.xml
```diff
@@ -0,0 +1,36 @@
...
+        <createTable tableName="WORKSPACE_ENTITY_CACHE">
+            <column name="workspace_id" type="BINARY(16)">
+                <constraints primaryKey="true"
+                             nullable="false"
+                             referencedTableName="WORKSPACE" referencedColumnNames="id"
+                             foreignKeyName="FK_WEC_WORKSPACE" deleteCascade="true"/>
+            </column>
+            <column name="entity_cache_last_updated" type="TIMESTAMP(6)" defaultValueDate="1970-01-01 00:00:02">
+                <constraints nullable="false"/>
+            </column>
```
davidangb, line 14: in practice, the code should never rely on the default value here. Still, I've chosen a default value that is one second later than the magic "cache is in error state" value. This should make it such that any workspace that somehow uses the db default value will be considered out of date and get its cache populated in an upcoming sweep.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -53,6 +54,17 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+                "cache disabled at system level"
+              else if (!useCache)
+                "user request specified cache bypass"
+              else if (!isEntityCacheCurrent)
+                "cache is out of date"
+              else
+                "unknown reason - this should be unreachable"
+
+              logger.info(s"entity statistics cache: miss ($missReason) [${workspaceContext.workspaceIdAsUUID}]")
+
```
davidangb, line 67: the extra logging in this file isn't necessary for this ticket, but I think it will help us analyze performance going forward.

davidangb:
[last swat run](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/54489/) had one and only one error, which happened during test cleanup of `WorkspaceApiSpec.Rawls should allow project owners to create, clone, and delete workspaces`:
```
org.broadinstitute.dsde.workbench.service.RestException: {"causes":[],"message":"Sam call to HttpMethod(DELETE) /api/resources/v1/billing-project/gpalloc-qa-master-mecq2r6 failed with error HttpEntity.Strict(application/json,2389 bytes total)","source":"rawls","stackTrace":[],"statusCode":500}
```
this error is a known test-infrastructure error and does not affect runtime code; I'm calling that a passed run.

sagehen03:
It seems like much of the logic in this PR is identifying which caches need to be refreshed.  But what if instead of pulling this information we pushed it?  In other words, we could do something like add a trigger on `WORKSPACE.last_modified` that would then add a record to a `WS_CACHES_TO_UPDATE` table.  The code to refresh the caches would just read from that queue table taking the oldest record, and the code could identify caches that are out of date by checking whether there's a record for that workspace in the queue table.  I think that simplifies the query logic significantly and gives us an easy way to query and understand the state of the caching system.  I'm probably missing some considerations, but this re-design leapt to mind when I first started thinking about this code.  What do you think?

davidangb:
@sagehen03 I like the concept of what you're suggesting, but I'm also hesitant to add a trigger on `WORKSPACE.last_modified`. I want to avoid as much contention as possible on the `WORKSPACE` table, so I don't want to add any extra triggers that would add to the transaction that is writing to `WORKSPACE.last_modified`. I agree the current logic is a bit complex, but I like that it is isolated from workspace transactions - it simply reads the state of the workspaces at some point after they've been updated.

davidangb:
@sagehen03 also note that this PR is simply replicating the pre-existing logic but moving it into standalone tables - I'm not adding any extra logic. The only part I guess I'm adding is that WORKSPACE_ENTITY_CACHE may not have a row for the workspace, so I need to check for missing rows

sagehen03:
Fair enough. I'll let sleeping dogs lie after I say my piece that we know what workspaces' caches need refreshing as we update the `last_modified` column and we're choosing to let that information slip away and then reconstructing that info through the query in `findMostOutdatedEntityCacheAfter` which looks like new logic to me?  We also wouldn't necessarily have to use a trigger, we could also do it through code in a separate sql statement, but it's fair to say the code or trigger would impose additional db processing closer to the time of writes.

davidangb:
`findMostOutdatedEntityCacheAfter` moved from `WorkspaceComponent` to `EntityCacheComponent`, and I rewrote it in plain SQL instead of Slick to make the queries more readable and (hopefully) optimized. I totally agree with your assessment, and I'm open to a larger refactoring - how would you feel about getting this PR in as-is (modulo non-refactoring feedback) to relieve the failures, then think about a refactor?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  object entityCacheQuery extends TableQuery(new EntityCacheTable(_)) with RawSqlQuery with LazyLogging {
+
+    val driver: JdbcProfile = EntityCacheComponent.this.driver
+
+    def findMostOutdatedEntityCacheAfter(minCacheTime: Timestamp, maxModifiedTime: Timestamp): ReadAction[Option[(UUID, Timestamp, Option[Timestamp])]] = {
+      // Find the workspace that has the entity cache that is the most out of date:
+      // A. Workspace has a cacheLastUpdated date that is not current ("current" means equal to lastModified)
+      // B. cacheLastUpdated is after @param timestamp
```
sagehen03:
```suggestion
      // B. cacheLastUpdated is after @param minCacheTime
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
```diff
@@ -0,0 +1,73 @@
...
+      // C. lastModified is before @param cooldownBound, meaning the workspace isn't likely actively being updated
```
sagehen03:
```suggestion
      // C. lastModified is before @param maxModifiedTime, meaning the workspace isn't likely actively being updated
```

sagehen03:
> `findMostOutdatedEntityCacheAfter` moved from `WorkspaceComponent` to `EntityCacheComponent`, and I rewrote it in plain SQL instead of Slick to make the queries more readable and (hopefully) optimized. I totally agree with your assessment, and I'm open to a larger refactoring - how would you feel about getting this PR in as-is (modulo non-refactoring feedback) to relieve the failures, then think about a refactor?

Totally! If you got the impression that I was trying to be Gandalf and saying you shall not pass--that was not my intention.  I understand this is the most incremental approach and I respect your instinct to go that route here.  I just wanted to start a discussion based on my first impressions of this code.  It may be a minor point of semantics, but while the net result `findMostOutdatedEntityCacheAfter` in its new location should be the same as it was in `WorkspaceComponent` it now has to pull data from the new table so it's a new logic implementation--I'd say :).  I'm reviewing this PR as is; it's just taking me a while to get some of the things straight in my mental model.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -26,10 +25,14 @@ object EntityStatisticsCacheMonitor {
   sealed trait EntityStatisticsCacheMessage
   case object Sweep extends EntityStatisticsCacheMessage
   case object ScheduleDelayedSweep extends EntityStatisticsCacheMessage
+
+  //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
+  val MIN_CACHE_TIME = new Timestamp(1000)
```
sagehen03, line 30: Is this value exclusively used to identify workspaces whose caches didn't refresh due to error and should now be left alone?  If yes, my pretty strong preference would be not to overload the meaning of `entityCacheLastUpdated` in that fashion.  Let's just add an separate column to the `WORKSPACE_ENTITY_CACHE` to denote those workspaces whose caches aren't refreshing. That approach has the benefit of lowering the cognitive burden and more easily allowing for retries.

davidangb: it is, yes - how would you feel about changing that logic as part of AS-931 instead of in this PR?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityCacheComponent.scala
sagehen03: I don't believe this is necessarily any more performant, but for me it's easier to follow 
```
SELECT w.id, w.last_modified, c.entity_cache_last_updated
FROM WORKSPACE w LEFT OUTER JOIN WORKSPACE_ENTITY_CACHE c
    on w.id = c.workspace_id
where (c.entity_cache_last_updated > $minCacheTime or c.entity_cache_last_updated is null)
and w.last_modified < $maxModifiedTime and  c.entity_cache_last_updated < w.last_modified
order by w.last_modified asc limit 1
```
Probably not something to include in this PR, but the query is already doing the work of sorting all the caches that need refreshing and I think that sort is stable so you might as well just return all of them as a list and process them in scala at whatever speed you'd like and only run this query when the sweeper is out of caches to refresh.  That way we're asking less from the db.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
sagehen03: That's fine.

sagehen03 (approved):
As I've detailed in other comments, the code is more complicated and tougher to understand for the initiated than I think it could be, but I understand that's both a function of the existing code and some of the inherent challenges with our production db situation.  I don't see bugs with this code other than a couple of out of date comments.  Feel free to incorporate or memory hole my comments with regard to future Rawls cache work as you see fit.

davidangb:
@sagehen03
1. Good call on the outer join! I admit I originally tried the outer join but for some reason I didn't get it working and my unit tests were failing; your suggested SQL works and I've adopted it.
2. Your comments about changing business logic and simplifying are totally valid and worth pursuing; I'm just trying to keep the scope of this PR limited. If it wasn't clear, what I'm picturing is:
  * AS-929: move existing business logic to new db tables, get everything working as-is on the new schema
  * AS-931 (and beyond): update/improve business logic, starting from the new schema

MatthewBemis (approved):
I think the latest SQL change has caused a few unit test failures? I am going to leave a thumb pending that though, as everything else LGTM

sagehen03 (approved):

sagehen03:
> @sagehen03
> 
>     1. Good call on the outer join! I admit I originally tried the outer join but for some reason I didn't get it working and my unit tests were failing; your suggested SQL works and I've adopted it.
> 
>     2. Your comments about changing business logic and simplifying are totally valid and worth pursuing; I'm just trying to keep the scope of this PR limited. If it wasn't clear, what I'm picturing is:
> 
> 
>     * [AS-929](https://broadworkbench.atlassian.net/browse/AS-929): move existing business logic to new db tables, get everything working as-is on the new schema
> 
>     * [AS-931](https://broadworkbench.atlassian.net/browse/AS-931) (and beyond): update/improve business logic, starting from the new schema

Thanks I appreciate you clarifying the sequence of PRs/changes.  I didn't have that all that context in mind when I made my original comments.

PR #1523, opened 2021-09-17
honor attribute namespaces when sorting [SUP-523]

davidangb:
SUP-523: we were not including the attribute namespace in the consideration of sorting.

the new unit test - and the pre-existing `return sorted results on entity query for number field` unit test - both fail when the new fixture data is used but the runtime change is missing, and pass when the runtime change is present.


davidangb:
no space left on device; jenkins retest

ansingh7115 (approved):

PR #1530, opened 2021-09-23
64 shards and new naming for shards [AS-957, AS-958]

davidangb:
PR against feature branch:
* move from 32 to 64 shards
* move from `ENTITY_ATTRIBUTE_1_8b` naming scheme to `ENTITY_ATTRIBUTE_18_1b` naming scheme


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -3,38 +3,33 @@
 
     <!-- create db functions and stored procedures that will help with the db migration -->
     <changeSet logicalFilePath="dummy" author="davidan" id="create_shard_functions_and_procedures">
-        <sql stripComments="true" endDelimiter="/">
+        <sql stripComments="true" endDelimiter="~">
```
davidangb, line 6: because I needed to use `/` as the division operator when calculating the lower bound, I needed to change the delimiter

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210615_sharded_entity_tables.xml
```diff
@@ -3,38 +3,33 @@
 
     <!-- create db functions and stored procedures that will help with the db migration -->
     <changeSet logicalFilePath="dummy" author="davidan" id="create_shard_functions_and_procedures">
-        <sql stripComments="true" endDelimiter="/">
```
ansingh7115, line 6: Out of curiosity, how are you testing this changeset works? Are you cloning the dev db, or running it on a local db?

davidangb: Unit tests prove it's not entirely broken - but then I ran it against a clone of the dev db and verified that shards contain data and I can query against that data.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -185,24 +185,24 @@ trait AttributeComponent {
...
-        ENTITY_ATTRIBUTE_archive; we'd want to rename it to ENTITY_ATTRIBUTE_unsharded. We'd also need to update
-        liquibase to not copy all attribute rows into the shards; we'd only want to copy rows for any workspaces
-        that should use sharding.
+        Sharded workspaces would use a shard identifier such as "04_07", while unsharded workspaces
+        would use a shard identifier of "archive". This would translate into referencing tables named
+        ENTITY_ATTRIBUTE_04_07 or ENTITY_ATTRIBUTE_archive, respectively.
      */
 
-    // see the liquibase changeset "20210615_sharded_entity_tables.xml" for expected shard identifier values
+    val shardSize = 4 // range of characters in a shard, e.g. 4 = "00-03"
```
ansingh7115, line 193: Does this need to change to 5 or am I misunderstanding? (I've never actually sharded a db so I'm learning as I go!)

davidangb: this value is really only used in our custom code, it's used to figure out how to go from workspaceId -> shard name. Our algorithm looks at the first two characters of the workspaceId. A value of "4" here means that we include 4 consecutive values in each shard, like this:

shard 1:
* 00xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 01xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 02xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 03xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx

shard 2:
* 04xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 05xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 06xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
* 07xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx

ansingh7115: ahhh, got it!

ansingh7115 (approved):

PR #1532, opened 2021-09-24
separate create- and populate-shard liquibase changesets

davidangb:
This PR makes separate liquibase changesets for creating the shard tables vs. populating those shards.

This sets us up for incremental deployment, wherein we may want to create all the shards early on, but leave them empty and use some as-yet-unsettled strategy for populating them.

As of this PR, it's still an all-at-once migration; we populate all shards on deploy. But once we settle on a population strategy, it should be easy to edit the changesets to make them do what we need.

Last caveat: the rollbacks in each populate-shard changeset roll back by dropping and recreating the table, resetting it to empty. This may be too destructive for whatever population strategy we settle on; when we know what our population strategy is we must also reconsider those rollbacks.

I tested this by running a full migration against yet another clone of the dev db, it succeeded in ~33 minutes.

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
@@ -1,0 +1,331 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <!-- START populate each shard in its own changeset. If we hit an error with an
+            individual shard, we can restart the liquibase migration at that point,
+            and don't have to rollback any progress up to that changeset. -->
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_00_03">
+        <sql>call copyRowsToAttributeShard('00_03');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_00_03; call createEntityAttributeShard('00_03');</rollback>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_04_07">
+        <sql>call copyRowsToAttributeShard('04_07');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_04_07; call createEntityAttributeShard('04_07');</rollback>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_08_0b">
+        <sql>call copyRowsToAttributeShard('08_0b');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_08_0b; call createEntityAttributeShard('08_0b');</rollback>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_0c_0f">
+        <sql>call copyRowsToAttributeShard('0c_0f');</sql>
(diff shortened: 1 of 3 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
```diff
@@ -100,333 +92,146 @@
...
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_e0_e3;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_e4_e7;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_e8_eb;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_ec_ef;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_f0_f3;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_f4_f7;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_f8_fb;
+            DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_fc_ff;
+        </rollback>
     </changeSet>
```
davidangb, line 231: in my local test, this changeset to create all 64 shard tables took 11594ms

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
```diff
@@ -0,0 +1,331 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <!-- START populate each shard in its own changeset. If we hit an error with an
+            individual shard, we can restart the liquibase migration at that point,
+            and don't have to rollback any progress up to that changeset. -->
+    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_00_03">
+        <sql>call copyRowsToAttributeShard('00_03');</sql>
+        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_00_03; call createEntityAttributeShard('00_03');</rollback>
```
MatthewBemis, line 9: what happens if the rollback fails? does Liquibase just silently move on to the next changeset or does it throw a fatal exception that prevents booting?

MatthewBemis (approved):
lgtm, just one question

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
MatthewBemis: i think what I am worried about is that if the `copyRowsToAttributeShard` fails once, is it likely that it will fail a 2nd time and then boot will be prevented entirely? this would require us to either go in and make a manual database change to fix the data (if that's the problematic piece) or make a quick hotfix change to the migration to fix the issue

davidangb: A few points:
* note that the rollback calls `createEntityAttributeShard`, which already succeeded in a previous changeset - so it's not a guarantee that it'll work based on current db conditions, but it's some indication of confidence.
* I tried running unit tests with the liquibase `updateTestingRollback` [command](https://docs.liquibase.com/commands/community/updatetestingrollback.html). I could see that all the sharding-related changesets rolled back correctly. But unfortunately, the overall liquibase migration failed - there are some unrelated changesets that have [missing rollbacks](https://github.com/broadinstitute/rawls/blob/f83c7d2adec00e1d5759310429c36f60cfec9103/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210916_clone_workspace_file_transfer.xml#L27). It gave me confidence in *these* rollbacks, though
* overall good question! And I didn't put too much thought/effort into these "populate" changesets or their rollbacks, because I wasn't sure where we'd end up for a populate strategy. Once we settle on that, I think these need more attention.

MatthewBemis: Ugh. I'll admit, I embarrassingly misread the rollback statement. I think this is good as you have it. Thanks for clarifying :)

PR #1533, opened 2021-09-28
[PROD-577] Fix bug in CloneWorkspaceFileTransferMonitor

marctalbott:
Ticket: [PROD-577](https://broadworkbench.atlassian.net/browse/PROD-577)
* don't fast fail when copying files for multiple workspaces and one fails

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
@@ -39,7 +39,13 @@ class CloneWorkspaceFileTransferMonitorActor(val dataSource: SlickDataSource, va
         dataAccess.cloneWorkspaceFileTransferQuery.listPendingTransfers()
       }
       _ <- pendingTransfers.toList.traverse { pendingTransfer =>
-        IO.fromFuture(IO(copyBucketFiles(pendingTransfer)))
+        IO.fromFuture(IO(copyBucketFiles(pendingTransfer))).attempt.map {
+          case Left(e) => {
+            logger.warn(s"Failed to copy files from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName}", e)
+            List.empty
+          }
+          case Right(res) => res
+        }
       }.unsafeToFuture()
     } yield()
   }
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
@@ -195,4 +195,53 @@ class CloneWorkspaceFileTransferMonitorSpec(_system: ActorSystem) extends TestKi
       system.stop(actor)
     }
   }
+
+  it should "copy files for good workspaces even when problematic workspaces are failing" in {
+    withEmptyTestDatabase { dataSource: SlickDataSource =>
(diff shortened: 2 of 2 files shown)


gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
```diff
@@ -39,7 +39,13 @@ class CloneWorkspaceFileTransferMonitorActor(val dataSource: SlickDataSource, va
         dataAccess.cloneWorkspaceFileTransferQuery.listPendingTransfers()
       }
       _ <- pendingTransfers.toList.traverse { pendingTransfer =>
-        IO.fromFuture(IO(copyBucketFiles(pendingTransfer)))
+        IO.fromFuture(IO(copyBucketFiles(pendingTransfer))).attempt.map {
+          case Left(e) => {
+            logger.warn(s"Failed to copy files from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName}", e)
+            List.empty
```
jacmrob, line 45: why add an empty list?

andy7i, line 45: so if I'm understanding this, this catches the error (which is ok because this is an async call) so that it doesn't affect other file transfers. 
If the file transfer fails all retries, would users know other than noticing that files are missing?

andy7i (approved):
some questions but not blocking

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
marctalbott: Just for type purposes. The `Right` case returns `List[Option[StorageObject]]` or something like that, so I wanted the `Left` case to return a list as well

marctalbott: We haven't exposed that to users yet, but there's an additional field on the `WorkspaceDetails` object that indicates when the workspace's files were successfully copied. The information is there and could be shown in the UI, but it isn't yet

jacmrob (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
rtitle: Although the result is discarded anyway so might as well just return `()` :) Returning `List` is fine too though

rtitle: For the future would be good to figure out an error recovery process. E.g. maybe we try to copy N times, and report to the user somehow if we still couldn't copy the files. As written now it will continually retry forever and generate a lot of errors in Rawls log.  (Maybe we should even `DELETE` the offending row in `CLONE_WORKSPACE_FILE_TRANSFER` in the prod Rawls DB.)

rtitle (approved):
LGTM

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
marctalbott: Agreed, filed a ticket on the CI board https://broadworkbench.atlassian.net/browse/CA-1542. I am doing a bit more digging into the offending row in prod that's affecting users, but it does seem like we might want to delete that row manually

PR #1534, opened 2021-09-29
AS-951: Delete Entity Column API

ansingh7115:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -341,6 +341,13 @@ trait AttributeComponent {
       }
     }
 
+    def deleteAttributeColumn(workspaceNamespace: String, workspaceName: String, entityType: String, entityNamespace: String, attributeName: String) = {
+     sql"""delete ea from ENTITY_ATTRIBUTE ea
+               join ENTITY e on e.id = ea.owner_id
+               join WORKSPACE w on w.id = e.workspace_id
+           where w.namespace = ${workspaceNamespace} AND w.name = ${workspaceName} AND e.entity_type = ${entityType} AND ea.namespace = ${entityNamespace} AND ea.name = ${attributeName}""".as[Int]
+    }
```
davidangb, line 364: could you have this accept a workspace UUID instead of workspaceNamespace + workspaceName? In that case you could omit the join to the `WORKSPACE` table and the `where w.namespace = ${workspaceNamespace} AND w.name = ${workspaceName}`, replacing it with `where e.workspace_id = $workspaceId`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -341,6 +341,13 @@ trait AttributeComponent {
       }
     }
 
+    def deleteAttributeColumn(workspaceNamespace: String, workspaceName: String, entityType: String, entityNamespace: String, attributeName: String) = {
+     sql"""delete ea from ENTITY_ATTRIBUTE ea
```
davidangb: nit: I don't believe you need the `delete ea from …`, it should just be `delete from …`

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -103,6 +103,23 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
+      val deleteFuture = for {
+        entityProvider <- entityManager.resolveProviderFuture(entityRequestArguments)
+        _ <- entityProvider.deleteEntityAttribute(workspaceContext, entityType, entityAttributeNamespace, entityAttributeName)
+      } yield {
+        PerRequest.RequestComplete(StatusCodes.NoContent)
+      }
+
+      deleteFuture.recover {
+        case delEx: DeleteEntitiesConflictException => RequestComplete(StatusCodes.Conflict, delEx.referringEntities)
+      }.recover(bigQueryRecover)
```
davidangb: I don't think the `bigQueryRecover` is helpful here - that's meant to trap BigQuery-related exceptions

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -100,6 +100,18 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.entityQuery.checkAttributeNameExists(workspaceContext.workspaceIdAsUUID, entityType, entityAttributeNamespace, entityAttributeName) flatMap { attributeNameExists =>
+        if (attributeNameExists.toList.head)
+          dataAccess.entityQuery.deleteColumn(workspaceContext, entityType, entityAttributeNamespace, entityAttributeName)
+        else
+          throw new DataEntityException(message = s"Could not find column $entityAttributeName belonging to entity namespace $entityAttributeNamespace of entity type $entityType in workspace ${workspaceContext.namespace}/${workspaceContext.name}",
+            code = StatusCodes.BadRequest)
+      }
+    }
+  }
```
davidangb: do you need the `checkAttributeNameExists` at all? I believe `deleteColumn` returns the number of rows it actually deleted. Could you run the delete and inspect its results? If it deleted > 0 rows, all good; if it deleted 0 rows, then throw the `Could not find column …` exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -103,6 +103,23 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       }.recover(bigQueryRecover)
     }
 
+  def deleteEntityAttribute(workspaceName: WorkspaceName, entityType: String, entityAttributeNamespace: String, entityAttributeName: String, dataReference: Option[DataReferenceName], billingProject: Option[GoogleProjectId]): Future[PerRequestMessage] =
```
sagehen03: Still new to scala so likely I'm missing something, but why have the dataReference and billingProject params here?  It looks like there's only one call site for this method and we don't use those params there.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -92,6 +92,13 @@ trait EntityApiService extends UserInfoDirectives {
             }
           }
         } ~
+        path("workspaces" / Segment / Segment / "entities" / Segment / Segment / Segment ) { (workspaceNamespace, workspaceName, entityType, entityAttributeNamespace, entityAttributeName) =>
+          delete {
+            complete {
+              entityServiceConstructor(userInfo).deleteEntityAttribute(WorkspaceName(workspaceNamespace, workspaceName), entityType, entityAttributeNamespace, entityAttributeName, None, None)
+            }
+          }
+        } ~
```
davidangb: two comments on the API signature:
1) Instead of separate fields for the attribute namespace and attribute name, could you accept a delimited "namespace:name" string, and parse it via [`AttributeName.fromDelimitedName`](https://github.com/broadinstitute/rawls/blob/7cbf0de410a5deaf70bcdd89f0f0f1c5fe7fd7c0/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala#L103)? This would allow the common case of the default namespace to omit the namespace, e.g. "myattribute" and only require specifying a namespace for non-default, e.g. "pfb:projectname".
2) Could this signature be modified to allow deleting multiple columns in one API call? It feels somewhat limiting to support only one at a time.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
sagehen03: It may be just be a legacy name, but we're not calling Big Query I think?  It looks like we use `bigQueryRecover` in other MySQL operations so I'm guessing there's no issue here, but it looks strange to me.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -100,6 +100,18 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
     }
   }
 
+  override def deleteEntityAttribute(workspaceContext: Workspace, entityType: String, entityAttributeNamespace: String, entityAttributeName: String): Future[Vector[Int]] = {
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.entityQuery.checkAttributeNameExists(workspaceContext.workspaceIdAsUUID, entityType, entityAttributeNamespace, entityAttributeName) flatMap { attributeNameExists =>
```
sagehen03: What's the disadvantage of just running the delete without checking to see that the attribute exists?  Won't these requests be coming from the UI and therefore constrained to existing attributes?  It seems like we're doing extra work here, but I could be misunderstanding the use case.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -360,6 +360,13 @@ trait EntityComponent {
         val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { ref => sql"(${ref.entityType}, ${ref.entityName})" })
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
       }
+
+      def checkAttributeNameExists(workspaceId: UUID, entityType: String, attributeNamespace: String, attributeName: String) = {
```
sagehen03: I left a comment at the call site for this method with more detail, but I'm wondering why we need this code?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -341,6 +341,13 @@ trait AttributeComponent {
...
+               join ENTITY e on e.id = ea.owner_id
+               join WORKSPACE w on w.id = e.workspace_id
+           where w.namespace = ${workspaceNamespace} AND w.name = ${workspaceName} AND e.entity_type = ${entityType} AND ea.namespace = ${entityNamespace} AND ea.name = ${attributeName}""".as[Int]
```
sagehen03: Could we just use workspaceId instead of workspaceNamespace and workspaceName?

sagehen03 (commented):
I promise I didn't read David's review first.  My comments were all written before I saw his review :)

sagehen03 (approved):
In addition to running the unit tests, I tested this with swagger and curl on my thieagen anonymized workspace and it seems to run in 1-2s which seems pretty good to me.  I think there are some changes we can potentially make here to reduce the total amount of code and to simplify things, but I don't see any bugs or stop the release train type issues.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
ansingh7115: It was so that we can check if the attribute actually exists before deleting it. Our endpoints for deletion pass up an error response if you try to delete something that doesn't exist. However, per David An's message below, there's another way we could do this that might be easier.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
ansingh7115: So we wanted to make Entity Service be able to be generic across rawls entities and data repo snapshot data. So I added the `dataReference` and `billingProject` params in case we eventually wanted to implement this functionality for data repo snapshots too. But, that implementation would be so far off that it might not be useful to add this stuff in now.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
ansingh7115: I was trying to write this method similarly to `deleteEntities` above it so that we could eventually add a deleteAttribute feature for data repo snapshot tables too. But that might be overkill right now, so I could just simplify it and remove the bigquery related stuff from this method

ansingh7115: Answered in the following comment - https://github.com/broadinstitute/rawls/pull/1534/files#r719977997

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
ansingh7115: I was trying to make it match the behavior of other endpoints that delete entities, which fail if the entity does not exist. I'm going to change the code a bit, though, so we don't have to do this check and will just return an error if the number of rows we delete is zero.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
ansingh7115: This is a good point! I will do this instead

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
ansingh7115: I've made this delete multiple columns - but one thing I want to decide is the behavior if the some of the items in the list of column names are not found, but some are - what's the behavior? The way it's written now, the deletable columns will be deleted, while the columns that cannot be found will not, and a 204 is returned. we could instead make it fail all deletes if one delete fails.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1912,6 +1912,42 @@ paths:
...
+      description: Delete entity column
+      operationId: delete_entity_column
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespacePathParam'
+        - $ref: '#/components/parameters/workspaceNamePathParam'
+        - $ref: '#/components/parameters/entityTypePathParam'
+        - name: attributeNames
+          in: query
+          description: |
+             The attribute Names to delete. Required.
```
davidangb: nits: don't capitalize "Names"; add a comment that this accepts a comma-delimited list

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -341,6 +341,27 @@ trait AttributeComponent {
...
+
+      def deleteAttributeColumn(workspaceId: UUID, entityType: String, attributeNames: Set[AttributeName]) = {
+        val attributeNamesSql = reduceSqlActionsWithDelim(attributeNames.map { attName =>
+          sql"""(${attName.namespace},${attName.name})"""
+        }.toSeq)
+
+        val deleteQueryBase = sql"""delete ea from ENTITY_ATTRIBUTE ea
+               join ENTITY e on e.id = ea.owner_id
+               join WORKSPACE w on w.id = e.workspace_id
+           where w.id = ${workspaceId} AND e.entity_type = ${entityType} AND (ea.namespace, ea.name) IN """
```
davidangb: if you have the workspaceId, you don't need to join to the `WORKSPACE` table and can thus make it easier on MySQL: remove the join, and change to using `where e.workspace_id = ${workspaceId}`

Also, have you run an EXPLAIN on this query? I am curious if it uses indexes well. See https://broadworkbench.atlassian.net/browse/BW-688 for a related investigation … changing the SQL may make it much more efficient, index-wise.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -486,6 +493,10 @@ trait EntityComponent {
       }
     }
 
+    def checkAttributeNameExists(workspaceId: UUID, entityType: String, attributeNamespace: String, attributeName: String) = {
+      EntityAndAttributesRawSqlQuery.checkAttributeNameExists(workspaceId, entityType, attributeNamespace, attributeName)
+    }
```
davidangb: this - and `EntityAndAttributesRawSqlQuery.checkAttributeNameExists` - are now unused, right? If so, delete them

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/EntityProvider.scala
```diff
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.rawls.entities.base
 import org.broadinstitute.dsde.rawls.entities.base.ExpressionEvaluationSupport.LookupExpression
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.GatherInputsResult
 import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.EntityUpdateDefinition
-import org.broadinstitute.dsde.rawls.model.{AttributeEntityReference, AttributeValue, Entity, EntityQuery, EntityQueryResponse, EntityTypeMetadata, SubmissionValidationEntityInputs}
+import org.broadinstitute.dsde.rawls.model.{AttributeEntityReference, AttributeValue, Entity, EntityQuery, EntityQueryResponse, EntityTypeMetadata, SubmissionValidationEntityInputs, Workspace}
```
davidangb: nit: unused import, this file doesn't need to change at all

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoEntityProvider.scala
```diff
@@ -21,7 +21,7 @@ import org.broadinstitute.dsde.rawls.expressions.parser.antlr.{AntlrTerraExpress
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.GatherInputsResult
 import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.EntityUpdateDefinition
 import org.broadinstitute.dsde.rawls.model.deltalayer.v1.{DeltaInsert, InsertDestination, InsertSource}
-import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, AttributeEntityReference, AttributeNull, AttributeNumber, AttributeString, AttributeValue, AttributeValueList, AttributeValueRawJson, Entity, EntityQuery, EntityQueryResponse, EntityTypeMetadata, ErrorReport, GoogleProjectId, SubmissionValidationEntityInputs}
+import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, AttributeEntityReference, AttributeNull, AttributeNumber, AttributeString, AttributeValue, AttributeValueList, AttributeValueRawJson, Entity, EntityQuery, EntityQueryResponse, EntityTypeMetadata, ErrorReport, GoogleProjectId, SubmissionValidationEntityInputs, Workspace}
```
davidangb: nit: unused import, this file doesn't need to change at all

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -100,6 +100,7 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
     }
   }
 
+
```
davidangb: nit: only changing whitespace, this file doesn't need to change at all

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -92,6 +95,26 @@ trait EntityApiService extends UserInfoDirectives {
             }
           }
         } ~
+        path("workspaces" / Segment / Segment / "entities" / Segment ) { (workspaceNamespace, workspaceName, entityType) =>
```
davidangb: line 37 already has a `path("workspaces" / Segment / Segment / "entityQuery" / Segment) { (workspaceNamespace, workspaceName, entityType) =>` declaration, containing a `get`; could just just add the `delete` method there instead of re-declaring the path?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -92,6 +95,26 @@ trait EntityApiService extends UserInfoDirectives {
...
+                } else if (paramValues.isEmpty) {
+                  throw new RawlsExceptionWithErrorReport(ErrorReport(BadRequest, s"Parameter '$paramName' must be included.")(ErrorReportSource("rawls")))
+                } else {
+                  paramValues.head.split(',').toSet.map { (value: String) => AttributeName.fromDelimitedName(value.trim) }
+                }
+              }
+              complete {
+                entityServiceConstructor(userInfo).deleteEntityAttributes(WorkspaceName(workspaceNamespace, workspaceName), entityType, thing())
+              }
+            }
```
davidangb: how about reusing `WorkspaceFieldSpecs`? Looks like you copied some code from it, could you use it instead? Something like:
```
          delete {
            parameterSeq { allParams =>
              val fieldSpecs = WorkspaceFieldSpecs.fromQueryParams(allParams, "attributeNames")
              if (fieldSpecs.fields.isEmpty)
                throw new RawlsExceptionWithErrorReport(ErrorReport(BadRequest, s"Parameter 'attributeNames' must be included.")(ErrorReportSource("rawls")))
              val attrsToDelete = fieldSpecs.fields.flatMap(_.map(f => AttributeName.fromDelimitedName(f.trim))).toSet

              complete {
                entityServiceConstructor(userInfo).deleteEntityAttributes(WorkspaceName(workspaceNamespace, workspaceName), entityType, attrsToDelete)
              }
            }
          }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
davidangb: I think what you have is the right solution - don't throw an error unless nothing at all was deleted

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -1627,6 +1627,87 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+    Delete(s"${testData.workspace.path}/entities/type1?attributeNames=hi", httpJson(EntityDeleteRequest(e))) ~>
+      sealRoute(services.entityRoutes) ~>
+      check {
+        assertResult(StatusCodes.BadRequest) {
+          status
+        }
+      }
+  }
+
+
```
davidangb, line 1736: could you add a test to check deleting multiple columns at once?

davidangb (commented):
some more comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
ansingh7115: Here's what EXPLAIN yields: 
Screen Shot 2021-10-04 at 10 36 14 PM

davidangb: thank you!!

davidangb (approved):
:+1: assuming tests pass

PR #1535, opened 2021-09-30
[CA-1580] Refactor / serialize policy updates in create workspace

andy7i:
QA-1588

Run policy updates serially in create workspace to prevent akka subscription timeout.

---

I'm seeing this error in the call to syncPolicyToGoogle:

```
500 Internal Server Error entity: {"causes":[],"exceptionClass":"java.util.concurrent.TimeoutException","message":"Response entity was not subscribed after 1 second. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`. POST /api/google/v1/resource/workspace/58410ad7-4632-4fbb-8953-19abbe0ae1c5/owner/sync Empty -> 200 OK Chunked
```

the response entity is read immediately after the request is made: https://github.com/broadinstitute/rawls/blob/5a5582ec9ba0ed1cb751f2a0e20523e67abda[…]/scala/org/broadinstitute/dsde/rawls/util/HttpClientUtils.scala

 

One option is to increase the akka subscription timeout to prevent this issue: Substream Source cannot be materialized more than once when reading from s30 - NEW. I tested this out and while it does fix this issue, I'm a bit hesitant to do this as it is an app-wide setting. 

The other is to serially make policy sync calls. This may make createWorkspace even a little longer, but could also fix QA-1571: Sam call to HttpMethod(DELETE) /api/resources/v1/billing-project/abc failed IN PROGRESS as well.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2179,7 +2179,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
     )
   }
 
-  private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span) = {
+  private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span): Future[List[Any]] = {
     traceWithParent("traversePolicies", parentSpan) (s1 =>
       // This used to be Future.traverse but that runs all the futures eagerly and concurrently. This led to some perf
       // issues, hitting the akka subscription timeout. We switched to use cats traverse to make these calls serially.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2006,7 +2006,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
       _ = logger.info(s"Creating labels for ${googleProjectId}.")
       googleProjectLabels = gcsDAO.labelSafeMap(Map("workspaceNamespace" -> workspaceName.namespace, "workspaceName" -> workspaceName.name, "workspaceId" -> workspaceId), "")
       googleProjectName = gcsDAO.googleProjectNameSafeString(s"${workspaceName.namespace}--${workspaceName.name}")
-      _ = logger.info(s"Setting up project ${googleProjectId} in cloud resource manager.")
+      _ = logger.info(s"Setting up google project ${googleProjectId} in cloud resource manager.")
       googleProject <- traceWithParent("setUpProjectInCloudResourceManager", span)(_ => setUpProjectInCloudResourceManager(googleProjectId, googleProjectLabels, googleProjectName))
       _ = logger.info(s"Remove RBS SA from owner policy ${googleProjectId}.")
       googleProjectNumber = gcsDAO.getGoogleProjectNumber(googleProject)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2181,6 +2181,16 @@ class WorkspaceService(protected val userInfo: UserInfo,
   }
 
   private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span): Future[Unit] = {
+    // Seq[String] of the policy names on a Workspace that need to be sync'd with their corresponding policy group on
+    // GSuite. Only syncs policies with bucket access (which need a google group)
+    // If there is not an AuthDomain, then we do not need to sync the Project Owner policy. Instead of syncing an extra
+    // group, we will use the billing project admin policy email directly on workspace to help keep the number of google
+    // groups a user is in below the limit of 2000.
+    def policyNamesThatRequireSync(withAuthDomain: Boolean): Seq[String] = {
+      val policyNames = WorkspaceAccessLevels.all.map(_.toString) :+ SamWorkspacePolicyNames.canCompute.toString()
+      if (withAuthDomain) policyNames else policyNames.filterNot(_ == WorkspaceAccessLevels.ProjectOwner.toString)
+    }
+
     traceWithParent("traversePolicies", parentSpan) { s1 =>
       // This used to be Future.traverse but that runs all the futures eagerly and concurrently. This led to a perf
       // issue with hitting the akka subscription timeout. We switched to use cats traverse to make these calls serially.
@@ -2189,10 +2199,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
           // when there isn't an auth domain, we will use the billing project admin policy email directly on workspace
           // resources instead of synching an extra group. This helps to keep the number of google groups a user is in below
           // the limit of 2000
-          // only sync policies that have corresponding WorkspaceAccessLevels to google because only those are
-          // granted bucket access (and thus need a google group)
-          !(policyName.isProjectOwner && !workspaceRequest.hasAuthorizationDomain) &&
-            (WorkspaceAccessLevels.withPolicyName(policyName.value).isDefined || policyName.isCanCompute)) (
+          policyNamesThatRequireSync(workspaceRequest.hasAuthorizationDomain).contains(policyName.value)) (
           IO.fromFuture(IO(traceWithParent(s"syncPolicy-${policyName}", s1)(_ =>
             samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName))))
         )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.rawls.workspace
 
 import java.util.UUID
-
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.StatusCodes
@@ -9,6 +8,7 @@ import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.stream.Materializer
 import bio.terra.workspace.client.ApiException
 import cats.implicits._
+import cats.effect.{ContextShift, IO}
 import com.google.api.services.cloudresourcemanager.model.Project
 import com.google.api.services.storage.model.StorageObject
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -77,7 +77,8 @@ object WorkspaceService {
                   servicePerimeterService: ServicePerimeterService,
                   googleIamDao: GoogleIamDAO, terraBillingProjectOwnerRole: String, terraWorkspaceCanComputeRole: String)
                  (userInfo: UserInfo)
-                 (implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext): WorkspaceService = {
+                 (implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext,
+                  cs: ContextShift[IO]): WorkspaceService = {
```
andy7i, line 82: also not sure if using the implicit CS is the right thing to do here, or if we should create a new ContextShift

```
  implicit lazy val contextShift: ContextShift[IO] = cats.effect.IO.contextShift(executionContext)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2178,22 +2179,24 @@ class WorkspaceService(protected val userInfo: UserInfo,
     )
   }
 
-  private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span) = {
+  private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span): Future[List[Any]] = {
     traceWithParent("traversePolicies", parentSpan) (s1 =>
-      Future.traverse(policyEmailsByName.keys) { policyName =>
+      // This used to be Future.traverse but that runs all the futures eagerly and concurrently. This led to some perf
+      // issues, hitting the akka subscription timeout. We switched to use cats traverse to make these calls serially.
+      policyEmailsByName.keys.toList.traverse { policyName =>
```
andy7i: not sure if this is the right/best way to run serially

andy7i (commented):
looking for feedback on this approach

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2178,22 +2179,24 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
           // only sync policies that have corresponding WorkspaceAccessLevels to google because only those are
           // granted bucket access (and thus need a google group)
-          traceWithParent(s"syncPolicy-${policyName}", s1)(_ => samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName))
+          IO.fromFuture(IO(traceWithParent(s"syncPolicy-${policyName}", s1)(_ => samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName))))
         } else {
-          Future.successful(())
+          IO.fromFuture(IO(Future.successful(())))
         }
-      }
+      }.unsafeToFuture()
```
ehigham: I think all this can be simplified into one [Applicative[Future].whenA](https://typelevel.org/cats/api/cats/Applicative.html#whenA[A](cond:Boolean)(f:=%3EF[A]):F[Unit]) call (or whatever the logic is, it's too confusing)
```scala
// why do we not have these predicates???
def isProjectOwner(policy: SamResourcePolicyName): Boolean = 
  policy == SamWorkspacePolicyNames.projectOwner

def isCanCompute(policy: SamResourcePolicyName): Boolean = 
  policy == SamWorkspacePolicyNames.canCompute

def hasAuthorizationDomain(request: WorkspaceRequest): Boolean =
  request.authorizationDomain match {
    case Some(xs) => !xs.isEmpty
    case None => False
  }

...

Applicative[Future].whenA(
  !isProjectOwner(policyName) && 
  hasAuthorizationDomain(workspaceRequest) &&
  (WorkspaceAccessLevels.withPolicyName(policyName.value).isDefined ||  isCanCompute(policyName)),
  traceWithParent(s"syncPolicy-${policyName}", s1)(_ => 
    samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName)
  )
)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2005,7 +2006,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
       _ = logger.info(s"Creating labels for ${googleProjectId}.")
       googleProjectLabels = gcsDAO.labelSafeMap(Map("workspaceNamespace" -> workspaceName.namespace, "workspaceName" -> workspaceName.name, "workspaceId" -> workspaceId), "")
       googleProjectName = gcsDAO.googleProjectNameSafeString(s"${workspaceName.namespace}--${workspaceName.name}")
-      _ = logger.info(s"Setting up project in ${googleProjectId} cloud resource manager.")
+      _ = logger.info(s"Setting up project ${googleProjectId} in cloud resource manager.")
```
gpolumbo-broad:
```suggestion
      _ = logger.info(s"Setting up google project ${googleProjectId} in cloud resource manager.")
```

gpolumbo-broad (approved):
LGTM so long as we can get a double check on the `ContextShift` from an `IO` expert

andy7i:
confirmed with @Qi77Qi that we should use the default ContextShift.

ehigham (approved):
Looks great. Thank you for making that change. I wonder if those helper functions and predicates would be better placed elsewhere. For example, those predicates that test policy names could they live in the same namespace as where the policy names are defined? That might enable better reuse.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2178,23 +2180,36 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
-          // the limit of 2000
-          Future.successful(())
-        } else if (WorkspaceAccessLevels.withPolicyName(policyName.value).isDefined || policyName == SamWorkspacePolicyNames.canCompute) {
-          // only sync policies that have corresponding WorkspaceAccessLevels to google because only those are
-          // granted bucket access (and thus need a google group)
-          traceWithParent(s"syncPolicy-${policyName}", s1)(_ => samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName))
-        } else {
-          Future.successful(())
-        }
+  private def syncPolicies(workspaceId: String, policyEmailsByName: Map[SamResourcePolicyName, WorkbenchEmail], workspaceRequest: WorkspaceRequest, parentSpan: Span): Future[List[Any]] = {
```
ehigham: If you want to, you can simplify the return type that you. Just `Future[Unit]` using `traverse_` instead of `traverse`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2178,23 +2180,36 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+    // only sync policies that have corresponding WorkspaceAccessLevels to google because only those are
+    // granted bucket access (and thus need a google group)
+    def isCanCompute(policy: SamResourcePolicyName): Boolean =
+      policy == SamWorkspacePolicyNames.canCompute
+
+    traceWithParent("traversePolicies", parentSpan) { s1 =>
+      // This used to be Future.traverse but that runs all the futures eagerly and concurrently. This led to a perf
+      // issue with hitting the akka subscription timeout. We switched to use cats traverse to make these calls serially.
+      policyEmailsByName.keys.toList.traverse { policyName =>
+        Applicative[Future].whenA(
```
Qi77Qi: I thought you want this sync call to happen sequentially? in that case, you do want to convert these `Future` into `IO`, and run `traverse` (or `traverse_` if you're returning `IO[Unit]`) on`IO` instead of `Future`

andy7i:
update: ran the workspaces test against perf 3 times and the create workspace tests all passed. Kibana shows 0 instances of the error akka subscription timeout error from these runs.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
gpolumbo-broad: I thought you said you talked with Qi and were just going to use the implicit here?  Or am I misunderstanding?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2178,23 +2180,24 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
           // only sync policies that have corresponding WorkspaceAccessLevels to google because only those are
           // granted bucket access (and thus need a google group)
-          traceWithParent(s"syncPolicy-${policyName}", s1)(_ => samDAO.syncPolicyToGoogle(SamResourceTypeNames.workspace, workspaceId, policyName))
-        } else {
-          Future.successful(())
-        }
-      }
-    )
+          !(policyName.isProjectOwner && !workspaceRequest.hasAuthorizationDomain) &&
+            (WorkspaceAccessLevels.withPolicyName(policyName.value).isDefined || policyName.isCanCompute)) (
```
gpolumbo-broad: I'm not thrilled about this logic statement, it took me a long time to grok what this is trying to say which has me worried if we have it right in the first place and has me worried that this will be fragile/difficult to maintain.

Long story short, I wonder if we create a new enumeration/list of policy names that should be synced.  Something like: 

`// Seq[String] of the policy names on a Workspace that need to be sync'd with their corresponding policy group on GSuite
val policyNamesThatRequireSync: Seq[String] = WorkspaceAccessLevels.all.map(toString) :+ SamWorkspacePolicyNames.canCompute.toString()`

(might be able to pretty that up too)

And then if we apply DeMorgan's to the first predicate, I think the logic can be expressed more simply as:

`(workspaceRequest.hasAuthorizationDomain || !policyName.isProjectOwner) && policyNamesThatRequireSync.contains(policyName.value)`

Which I think boils down do the idea that if the policy is one a "sync'able" policy and the Workspace has an AuthDomain, then go ahead and sync it.  If the Workspace does not have an AuthDomain, then skip the `Project Owner` policy.  Which has me thinking instead we go to:

`// Seq[String] of the policy names on a Workspace that need to be sync'd with their corresponding policy group on GSuite.  If there is not an AuthDomain, then we do not need to sync the Project Owner policy
def policyNamesThatRequireSync(withAuthDomain: Boolean): Seq[String] = { 
    val policyNames = WorkspaceAccessLevels.all.map(toString) :+ SamWorkspacePolicyNames.canCompute.toString()
    if (withAuthDomain) policyNames else policyNames.filterNot(_ == WorkspaceAccessLevels.ProjectOwner.value)
}`

So finally, I think this logical expression I'm commenting on can be boiled down to:

`policyNamesThatRequireSync(workspaceRequest.hasAuthorizationDomain).contains(policyName.value)`

Sorry this is such a long comment, it took me several tries to think this all through and to try to express this logic more clearly.  I really really hope I've got it right, I'm not convinced I do, which is exactly the reason why I wanted to try to simplify the expression.

gpolumbo-broad (changes requested):

andy7i:
putting down this refactor and tracking it as part of CA-1580 since it is not urgent and not a team priority

PR #1536, opened 2021-09-30
entity streaming for download-all entity API [AS-192]

davidangb:
this changes the list-all-entities API to use streams, all the way from the db to the json response.

We should see much improved memory usage and object allocation, though I haven't quantified that.

We should also see much improved [time-to-first-byte](https://developer.mozilla.org/en-US/docs/Glossary/time_to_first_byte) (TTFB). This will help especially with proxy timeouts, which are sometimes caused because the TTFB exceeds the timeout; the server is so busy calculating stuff that it hasn't yet even sent a byte in response. By lowering the TTFB and streaming bytes as they're ready, we keep the proxy connection alive.

Measurement using curl against a **_2504-entity/7-attribute dataset_** running Rawls on my local machine:

  | Streaming (this PR) |   | Materializing (dev branch) |  
-- | -- | -- | -- | --
  | TTFB | Total | TTFB | Total
run 1 | 0.854295 | 8.258964 | 5.503287 | 5.517296
run 2 | 0.857463 | 6.071911 | 7.147478 | 7.164272
run 3 | 0.986192 | 7.673797 | 6.814422 | 6.831227
run 4 | 1.852225 | 8.882299 | 4.975124 | 4.994485
  |   |   |   |  
average | 1.13754375 | 7.72174275 | 6.11007775 | 6.12682

Measurement using curl against a **_32,865-entity/60-attribute dataset_** running Rawls on my local machine:

  | Streaming (this PR) |   | Materializing (dev branch) |  
-- | -- | -- | -- | --
  | TTFB | Total | TTFB | Total
run 1 | 2.318231 | 138.972411 | 91.426873 | 93.138935
run 2 | 2.243682 | 168.240942 | 89.526169 | 91.260541
run 3 | 1.919636 | 228.454679 | 88.052662 | 89.706733
run 4 | 2.246611 | 174.422509 | 118.243408 | 119.984261
  |   |   |   |  
average | 2.18204 | 177.5226353 | 96.812278 | 98.5226175

the TTFB is _much_ better, though it looks as if total request time increased. This could use some more perf tests (I wonder if our perf tests capture TTFB at all?)

EDIT 11/15/21: I ran curl against a **_32,865-entity/60-attribute dataset_** running Rawls on my local machine, comparing a batch size of 1000 to a batch size of 5000. The larger batch size performed much better in terms of the total request time:

   | batch size 1000 |   | batch size 5000 |  
-- | -- | -- | -- | --
   | TTFB | Total | TTFB | Total
run 1 | 1.106258 | 211.857628 | 2.104846 | 114.173619
run 2 | 2.381596 | 213.056431 | 1.754114 | 75.178991
run 3 | 2.129985 | 212.905949 | 1.936276 | 143.929869
run 4 | 2.439934 | 212.906833 | 1.737956 | 82.278875
run 5 | 2.329623 | 226.098492 | 1.765553 | 104.955654
   |   |   |   |  
avg | 2.0774792 | 215.3650666 | 1.883298 | 103.8903385

The total request time for batch size 5000 is in line with the materializing approach.


davidangb:
one single swat test failure on `AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain can be cloned and have a group added to the auth domain`:
```
org.broadinstitute.dsde.workbench.service.RestException: {
  "causes": [],
  "message": "You cannot delete this workspace: Resource workspace/daf7370a-e36a-4ddc-8de3-231dfd668be0 not found",
  "source": "FireCloud",
  "stackTrace": [],
  "timestamp": 1633023934242
}
```
I feel like that is unrelated. Jenkins retest.

davidangb:
and now, fiab-start failed. Jenkins retest.

davidangb:
sigh yet another swat error that I believe is unrelated, in `AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain when the user is in one of the groups when shared with them can be seen but is not accessible`:
```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 2445, groupName -> billing-project_gpalloc-qa-master-jgwy1g7_owner, still used in group(s): -> {workspace_9d331350-d4f9-443a-8e78-83edc81aec4c_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}
```

davidangb:
it's a new day, let's see if jenkins/swat is any more stable: jenkins retest

davidangb:
maybe the weekend is better: jenkins retest

davidangb:
I see the `Foreign Key Violation(s) while deleting group(s)` tracked in QA-1571, looks like this is a known issue

davidangb:
looks like swat failures are resolved: jenkins retest

davidangb:
swat tests failed, timed out rendering Leo configs. This should be transient. Jenkins retest.

davidangb:
one swat failure - timeout in `org.broadinstitute.dsde.test.api.MethodLaunchSpec.owner can abort a launched submission`, which should be totally unrelated.

core/src/main/resources/reference.conf
```diff
@@ -57,6 +57,15 @@ data-source {
...
 
+slick {
+  # batchSize is used for writes, to group inserts/updates
+  # this must be explicitly utilized via custom business logic
+  batchSize = 2000
+  # fetchSize is used during Slick streaming to set the size of pages
+  # this must be explicitly set via withStatementParameters
+  fetchSize = 5000
+}
+
```
davidangb, line 68: I'll need to add `fetchSize` (and associated comments) to firecloud-develop as well. At least for this PR they're just here in reference.conf

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -144,9 +145,11 @@ trait EntityApiService extends UserInfoDirectives {
         } ~
         path("workspaces" / Segment / Segment / "entities" / Segment) { (workspaceNamespace, workspaceName, entityType) =>
           get {
-            traceRequest { span =>
+            // if any other APIs adopt streaming, move this implicit val higher up in the EntityApiService trait
+            implicit val jsonStreamingSupport: JsonEntityStreamingSupport = EntityStreamingSupport.json()
+            traceRequest { _ =>
               complete {
-                entityServiceConstructor(userInfo).listEntities(WorkspaceName(workspaceNamespace, workspaceName), entityType, span)
+                entityServiceConstructor(userInfo).listEntities(WorkspaceName(workspaceNamespace, workspaceName), entityType)
```
davidangb, line 152: opencensus tracing doesn't support tracing inside streams, so for this API we only have the outer `traceRequest`; no need to pass the span down into `listEntities`

PR #1538, opened 2021-09-30
add unit tests for sharding [AS-943]

davidangb:
AS-943. This PR merges to the feature branch.

We have pretty good coverage, mostly in `EntityApiServiceSpec`, for entity-related functionality.

This PR adds tests specific to sharding: when adding/updating/deleting attributes, do the SQL row operations occur in the correct shard?


ansingh7115 (approved):
Test cases look good to me!

sagehen03 (approved):
Thanks for being a diligent unit tester!

PR #1539, opened 2021-10-01
field selection for entityQuery [AS-955]

davidangb:
adds a `?fields=` query parameter to the entityQuery API, so an end user can control which attributes are returned by the query.

If the user requests no fields, via `?fields=`, we skip all the extra SQL joins and make a more efficient query just to the `ENTITY` table. The initial use case for this is for the Terra UI to streamline its select-all API query.

The `entityType`, `name`, and `attributes` keys are always returned in the response payload. If the entities have no matching attributes, the response will have `attributes: {}`.

One potential improvement is to exclude `entityType` unless the user requests it. This would cause a lot of code change - I would like to save that for a future follow-on PR.

I've also identified a few places where we can improve the SQL statements underlying entityQuery. I'd also like to save those for future follow-on PRs, and am looking forward to tackling those.

_**REVIEWER:**_ ignore whitespace! Especially important for `EntityApiService`.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -229,6 +229,12 @@ trait EntityComponent {
           left outer join ENTITY_ATTRIBUTE a on a.owner_id = e.id and a.deleted = e.deleted
           left outer join ENTITY e_ref on a.value_entity_ref = e_ref.id"""
 
+      // NB: the "null" as the last column here is important! It allows this query to be translated into
+      // EntityAndAttributesResult objects; the null represents the lack of any attributes for this entity
+      val baseEntitySql =
+        s"""select e.id, e.name, e.entity_type, e.workspace_id, e.record_version, e.deleted, e.deleted_date, null
+          from ENTITY e"""
+
       // Active actions: only return entities and attributes with their deleted flag set to false
 
       def activeActionForType(workspaceContext: Workspace, entityType: String): ReadAction[Seq[EntityAndAttributesResult]] = {
@@ -302,7 +308,7 @@ trait EntityComponent {
           val fieldsOption = entityQuery.fields.fields.map { fieldList =>
             val attributeNameList: Set[AttributeName] = fieldList.map(AttributeName.fromDelimitedName)
             val attrClauses = attributeNameList.map(attrName =>
-              sql"a.namespace = ${attrName.namespace} AND a.name = ${attrName.name}"
+              sql"(a.namespace = ${attrName.namespace} AND a.name = ${attrName.name})"
             )
             concatSqlActions(sql"#$prefix ", reduceSqlActionsWithDelim(attrClauses.toSeq, sql" or "))
           }
@@ -321,8 +327,7 @@ trait EntityComponent {
           sql") pagination ",
           filterSql("where", "pagination"),
           order("pagination"),
-          sql" limit #${entityQuery.pageSize} offset #${(entityQuery.page-1) * entityQuery.pageSize} ) p on p.id = e.id ",
-          attrSelectionSql("where")
+          sql" limit #${entityQuery.pageSize} offset #${(entityQuery.page-1) * entityQuery.pageSize} ) p on p.id = e.id "
         )
 
         def filteredCountQuery: ReadAction[Vector[Int]] = {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -256,6 +256,9 @@ trait EntityComponent {
         * @return
         */
       private def paginationSubquery(workspaceId: UUID, entityType: String, sortFieldName: String) = {
+        /* TODO: can we eliminate the SELECT ... e.all_attribute_values by being smarter about our where clauses?
+            while not returned to the end user, it still causes mysql extra work
+         */
```
davidangb, line 263: I'd like to tackle this - and the other improvements called out as TODOs - in a separate, follow-on PR to avoid major scope-creep and risk of regression

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -316,6 +346,44 @@ trait EntityComponent {
...
+                             a.id, a.namespace, a.name, a.value_string, a.value_number, a.value_boolean, a.value_json, a.value_entity_ref, a.list_index, a.list_length, a.deleted, a.deleted_date,
+                             e_ref.id, e_ref.name, e_ref.entity_type, e_ref.workspace_id, e_ref.record_version, e_ref.deleted, e_ref.deleted_date
+                             from ENTITY e
+                             left outer join ENTITY_ATTRIBUTE a on a.owner_id = e.id and a.deleted = e.deleted """,
+              attrSelectionSql(" and "),
+              sql""" left outer join ENTITY e_ref on a.value_entity_ref = e_ref.id """,
+              paginationJoin, order("p")).as[EntityAndAttributesResult]
+          }
+        }
+
```
davidangb, line 392: this is a refactoring/optimization of the previous `concatSqlActions(sql"#$baseEntityAndAttributeSql", paginationJoin, order("p")).as[EntityAndAttributesResult]` as seen below … this is the only major refactoring I tackled in this PR.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -248,7 +248,10 @@ object SortDirections {
 
   def toSql(direction: SortDirection) = toString(direction)
 }
-case class EntityQuery(page: Int, pageSize: Int, sortField: String, sortDirection: SortDirections.SortDirection, filterTerms: Option[String])
+case class EntityQuery(page: Int, pageSize: Int,
+                       sortField: String, sortDirection: SortDirections.SortDirection,
+                       filterTerms: Option[String],
+                       fields: WorkspaceFieldSpecs = WorkspaceFieldSpecs())
```
davidangb, line 254: is it worth renaming `WorkspaceFieldSpecs` to something more generic? Or does that really matter? It's a small change but would touch a lot of code and cause noise in this PR.

davidangb:
one and only one failure in swatomation: `org.broadinstitute.dsde.test.api.orch.MethodApiSpec.For a method config that references a redacted method launching an analysis should not be possible`

```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 1380, groupName -> billing-project_gpalloc-qa-master-x4o8mhq_owner, still used in group(s): -> {workspace_cd015b18-30a4-4958-9cdc-0644c1c60d3b_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}
```

jenkins retest

davidangb:
I see the `Foreign Key Violation(s) while deleting group(s)` tracked in QA-1571, looks like this is a known issue

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
ansingh7115: I think that change will also impact the PR I currently have open. Maybe we change it once both are merged in a separate small PR?

davidangb: works for me! I'd prefer to do it in a standalone PR anyway to avoid the noise

ansingh7115 (approved):
Great test cases! Couldn't really find anything to nitpick :)
Provided that the new sql query is performant, LGTM. Have you tried testing this with lots of entities / attributes?

davidangb:
@ansingh7115 I haven't fully perf-tested this, but I ran it manually against the older anonymized dataset we have, and I've looked at the execution plans - all of those turned out ok!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -316,6 +347,47 @@ trait EntityComponent {
...
+              sql"""select e.id, e.name, e.entity_type, e.workspace_id, e.record_version, e.deleted, e.deleted_date, null
+                             from ENTITY e
+                             where e.deleted = 'false' and e.entity_type = $entityType and e.workspace_id = ${workspaceContext.workspaceIdAsUUID} """,
+              filterSql("and", "e"),
+              order("e"),
+              sql" limit #${entityQuery.pageSize} offset #${(entityQuery.page-1) * entityQuery.pageSize}").as[EntityAndAttributesResult]
+          } else {
+            // user is sorting by an attribute value, so we need the largest number of joins
+            // this query is very similar to baseEntityAndAttributeSql
+            /* TODO: include "where e.deleted = 'false' and e.entity_type = $entityType and e.workspace_id = $workspaceId"
```
MatthewBemis, line 375: Don't we need the clause to filter out the deleted entities? I've walked through the code but perhaps I'm missing something

MatthewBemis (commented):
I have 1 question, but will be a 👍  once I get clarity on that

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: in this branch of the code - the `else` clause at line 370 - the `e.deleted = 'false'` is included inside `paginationSubquery`, which is called from `paginationJoin`. So, it's functionally correct and returns only active entities. What I'm suggesting in this comment is to *also* include `e.deleted = 'false'` in the first select clause - at lines 380-384 - as an optimization for MySQL. It's possible that by filtering that first select clause, MySQL will have less internal work to do inside its query planner because there's less data. But it also may not help the planner, and might just be extra SQL; it needs some investigation.

MatthewBemis (approved):

PR #1541, opened 2021-10-05
[CA-1460] Delete Errored V1 Billing Projects

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1460

When a user creates a v1 billing account, rawls creates the resource for the google project in sam BEFORE it creates the google project proper. If the project creation fails, we delete the project from rawl’s database but we never tell sam.

When listing billing projects, we ask sam for all resources with resource type billing project. Since it doesn’t know that the billing project and google project don’t exist, it returns us all the resources AS IF it existed.

TLDR; we can’t trust that a resource exists just because sam says it does. This is seems reasonable because a user could stab us in the back and delete these resources in google and we’d be in the same situation.

Fix: Test that google project that backs the billing project exists before trying to delete it.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -25,7 +25,7 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.{BigQueryTableName, GoogleProject}
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success, Try}
+import scala.util.{Failure, Success}
 
 /**
  * Created by dvoet on 10/27/15.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -396,14 +396,18 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     def F = Applicative[Future]
 
     def deleteResourcesInGoogle(projectId: GoogleProjectId) =
-      deletePetsInProject(projectId, userInfoForSam) *>
-        F.whenA(deleteGoogleProjectWithGoogle)(gcsDAO.deleteV1Project(projectId))
+      for {
+        _ <- deletePetsInProject(projectId, userInfoForSam)
+        _ <- F.whenA(deleteGoogleProjectWithGoogle)(gcsDAO.deleteV1Project(projectId))
+      } yield ()
 
     val projectId = GoogleProjectId(projectName.value)
     samDAO.listResourceChildren(SamResourceTypeNames.billingProject, projectName.value, userInfoForSam) flatMap { resourceChildren =>
       F.whenA(resourceChildren contains SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))(
-        googleProjectExists(projectId).ifM(deleteResourcesInGoogle(projectId), F.unit) *>
-          samDAO.deleteResource(SamResourceTypeNames.googleProject, projectName.value, userInfoForSam)
+        for {
+          _ <- googleProjectExists(projectId).ifM(deleteResourcesInGoogle(projectId), F.unit)
+          _ <- samDAO.deleteResource(SamResourceTypeNames.googleProject, projectName.value, userInfoForSam)
+        } yield ()
       )
     }
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -2,10 +2,10 @@ package org.broadinstitute.dsde.rawls.user
 
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets.UTF_8
-
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
+import cats.Applicative
 import cats.implicits._
 import com.google.api.client.auth.oauth2.TokenResponseException
 import com.google.api.client.http.HttpResponseException
@@ -25,7 +25,7 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.{BigQueryTableName, GoogleProject}
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 /**
  * Created by dvoet on 10/27/15.
@@ -308,7 +308,7 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
   def deleteBillingProject(projectName: RawlsBillingProjectName): Future[PerRequestMessage] = {
     requireProjectAction[PerRequestMessage](projectName, SamBillingProjectActions.deleteBillingProject) {
       for {
-        _ <- validateNoWorkspaces(projectName)
+        _ <- failUnlessHasNoWorkspaces(projectName)
         _ <- deleteGoogleProjectIfChild(projectName, userInfo)
         _ <- unregisterBillingProjectWithUserInfo(projectName, userInfo)
       } yield RequestComplete(StatusCodes.NoContent)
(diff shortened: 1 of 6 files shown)


gpolumbo-broad:
Great description of the PR!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -377,33 +377,33 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+      .ifM(Future.unit, Future.failed(new RawlsExceptionWithErrorReport(ErrorReport(
+        StatusCodes.BadRequest,
+        "Project cannot be deleted because it contains workspaces."
+      ))))
 
   // TODO - once workspace migration is complete and there are no more v1 workspaces or v1 billing projects, we can remove this https://broadworkbench.atlassian.net/browse/CA-1118
   private def deleteGoogleProjectIfChild(projectName: RawlsBillingProjectName, userInfoForSam: UserInfo, deleteGoogleProjectWithGoogle: Boolean = true) = {
-    samDAO.listResourceChildren(SamResourceTypeNames.billingProject, projectName.value, userInfoForSam).flatMap { projectChildren =>
-      if (projectChildren.contains(SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))) {
+    def googleProjectExists(projectId: GoogleProjectId) =
```
gpolumbo-broad, line 389: Not sure if we've ever decided on a convention for declaring the return type, but I would opt to always include it.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -377,33 +377,33 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+      gcsDAO.getGoogleProject(projectId) transform {
```
gpolumbo-broad, line 390: TIL - [Future.transform](https://www.scala-lang.org/api/2.13.6/scala/concurrent/Future.html#transform[S](f:scala.util.Try[T]=%3Escala.util.Try[S])(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[S]) thanks!

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -24,7 +24,8 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
 
   protected def when500orGoogleError(throwable: Throwable): Boolean = {
     throwable match {
-      case t: HttpResponseException => t.getStatusCode/100 == 5 || t.getStatusCode/100 == 4
+      case t: HttpResponseException if t.getStatusCode/100 == 4 => t.getStatusCode != 404
```
andy7i: I haven't looked too deeply into this, but I'm worried that this changes the behavior for many calls to google, since `when500orGoogleError` is used in `retryWhen500orGoogleError` and `retryWithRecoverWhen500orGoogleError`, which are used 50+ times between the two. What are your thoughts on this?

ehigham: Yeah, this is a big change and I discussed it in slack in https://broadinstitute.slack.com/archives/C2B7JFJ4S/p1633451415146700.
The TLDR is I think we should give special handling to 404 errors and not just roll it up into default handling for "`GoogleError`s". If it's a case where 404s should be handled by retries, then that's easy enough to add back for those cases.
I COULD have added a new interface method for this but decided against that as:
- It's not true - 404s don't mean that the google project doesn't exist
- It would have incurred more tech debt

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -377,33 +377,33 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
+    samDAO.listResourceChildren(SamResourceTypeNames.billingProject, projectName.value, userInfoForSam) flatMap { resourceChildren =>
+      Applicative[Future].whenA(resourceChildren contains SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))(
         for {
-          _ <- deletePetsInProject(GoogleProjectId(projectName.value), userInfoForSam)
-          _ <- if (deleteGoogleProjectWithGoogle) {
-            gcsDAO.deleteV1Project(GoogleProjectId(projectName.value))
-          } else Future.successful(())
+          _ <- googleProjectExists(projectId).ifM(deletePetsInProject(projectId, userInfoForSam)
+            *> Applicative[Future].whenA(deleteGoogleProjectWithGoogle)(gcsDAO.deleteV1Project(projectId)),
+            Future.unit)
```
gpolumbo-broad: I'm on the fence as to requesting this line to be extracted into its own method.  It has a lot going on in one expression with a couple layers of nesting, which is elegant, but not necessarily the easiest to read at first glance (in my opinion).  As is, I think it is fine, but if I needed to modify this or add to it, the first thing I would do would be to extract it.  So I guess what I'm saying is that if you were on the fence about extracting a method here, I'd say do it.  If you're happy/prefer it as is, then it is fine with me to keep it as is.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -221,6 +223,39 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
     }
   }
 
+  it should "Successfully to delete a billing project when the google project does not exist" in {
```
gpolumbo-broad:
```suggestion
  it should "Successfully to delete a billing project when the google project does not exist on GCP" in {
```

Just want to be explicit because there are `google-project` Resources that exist in Sam too

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
ehigham: I would agree for exported functions for documentation purposes - for simple helpers and predicates I don't think it adds much.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
ehigham: I've tried to make it more readable. I think people are more familiar with `for` comprehensions rather than `*>` so I've reverted to that. Let me know what you think.

andy7i (approved):
I'm still a bit wary of changing the behavior of `when500orGoogleError`, since this could fail with new cloud resources and may be difficult to debug. Overall it's probably still a step in the right direction so 👍

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
ehigham: I've backed out of changing all the 404 error handling and just localised it to `getGoogleProject`. This should mitigate some of the risk.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -377,33 +376,39 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
...
-    samDAO.listResourceChildren(SamResourceTypeNames.billingProject, projectName.value, userInfoForSam).flatMap { projectChildren =>
-      if (projectChildren.contains(SamFullyQualifiedResourceId(projectName.value, SamResourceTypeNames.googleProject.value))) {
+    def googleProjectExists(projectId: GoogleProjectId) =
+      gcsDAO.getGoogleProject(projectId) transform {
+        case Success(_) => Success(true)
+        case Failure(e: HttpResponseException) if e.getStatusCode == 404 => Success(false)
+        case Failure(t) => Failure(t)
+      }
+
+    def F = Applicative[Future]
```
marctalbott, line 396: I have some minor reservations about introducing Applicative here. I'm not sure what benefits it provides over an if statement. Is there something I'm missing?

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
ehigham: In simple cases where we'd have one branch, I agree - `when` might be overkill/conceptual overhead. Here I've got 3 instances where the alternative in the `if` statement is to do nothing. I feel `when` improves readability in these places as it reduces noise and code bloat, especially when combined with effects. Thanks for the shout out on reabability - I'm always happy to have these conversations.

PR #1543, opened 2021-10-06
Improve error messages for case sensitivity errors [AS-979]

davidangb:
Note that I used the wrong jira ticket in the branch name. AS-979 is the right one.

This is an interim/iterative change. We have specific use cases that currently return very ugly errors to the end user. This PR makes those error messages more helpful and less scary. It does not fix the underlying problem; we need more thought and decisions on that front. But, let's not wait on those decisions to change the error messages.

Before:
Screenshot-case-before

After:
Screenshot-case-after


ungwudik (approved):
LGTM

PR #1544, opened 2021-10-06
AS-976 + AS-977: Support per-workspace sharding 

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -583,7 +584,7 @@ case class WorkspaceDetails(namespace: String,
                             billingAccount: Option[RawlsBillingAccountName],
                             billingAccountErrorMessage: Option[String] = None,
                             completedCloneWorkspaceFileTransfer: Option[DateTime]) {
-  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProject, googleProjectNumber, billingAccount, billingAccountErrorMessage, completedCloneWorkspaceFileTransfer)
+  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProject, googleProjectNumber, billingAccount, billingAccountErrorMessage, completedCloneWorkspaceFileTransfer, sharded = None)
```
MatthewBemis: `None` because this is a temporary thing and I don't want to introduce a new field to `WorkspaceDetails`. A user may end up consuming that for some reason and I want to avoid backwards compatibility concerns

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -191,7 +192,7 @@ object Workspace {
     val randomString = java.util.UUID.randomUUID().toString
     val googleProjectId = GoogleProjectId(randomString)
     val googleProjectNumber = GoogleProjectNumber(randomString)
-    new Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes, isLocked, WorkspaceVersions.V2, googleProjectId, Option(googleProjectNumber), None, None, Option(createdDate))
+    new Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes, isLocked, WorkspaceVersions.V2, googleProjectId, Option(googleProjectNumber), None, None, Option(createdDate), sharded = Option(true))
```
MatthewBemis: All new workspaces will be sharded, so default `sharded` to `true`

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -191,7 +191,7 @@ class AdminApiServiceSpec extends ApiServiceSpec {
       check {
         assertResult(StatusCodes.OK) { status }
         // TODO: why is this result returned out of order?
-        sortAndAssertWorkspaceResult(testData.allWorkspaces) { responseAs[Seq[WorkspaceDetails]].map(_.toWorkspace) }
+        sortAndAssertWorkspaceResult(testData.allWorkspaces.map(_.copy(sharded = None))) { responseAs[Seq[WorkspaceDetails]].map(_.toWorkspace) }
```
MatthewBemis: Since `sharded` defaults to `None` when converting from a `WorkspaceDetails` object to a `Workspace` object, this needed to be changed to just ignore that value entirely. I may have mentioned this elsewhere but the only way to avoid this would be to add `sharded` to the `WorkspaceDetails` object which I wanted to avoid

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2229,7 +2229,8 @@ class WorkspaceService(protected val userInfo: UserInfo,
       googleProjectNumber = googleProjectNumber,
       currentBillingAccountOnWorkspace,
       billingAccountErrorMessage = None,
-      completedCloneWorkspaceFileTransfer = completedCloneWorkspaceFileTransfer
+      completedCloneWorkspaceFileTransfer = completedCloneWorkspaceFileTransfer,
+      sharded = Some(true)
```
MatthewBemis: this creates a *new* workspace, so default it to `true`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -92,7 +92,8 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
         _ <- dataAccess.entityTypeStatisticsQuery.deleteAllForWorkspace(workspaceId)
         _ <- dataAccess.entityTypeStatisticsQuery.batchInsert(workspaceId, entityTypesWithCounts)
         //update entity attribute statistics
-        entityTypesWithAttrNames <- dataAccess.entityQuery.getAttrNamesAndEntityTypes(workspaceId)
+        sharded <- dataAccess.workspaceQuery.isWorkspaceSharded(workspaceId)
```
MatthewBemis: The earlier DB lookups only ever return the `workspaceId`, so this needs an explicit lookup for the `sharded` value

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -428,7 +428,10 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
 
     //yank out of Seq[(Long, Option[Entity])] and into Seq[(Long, Entity)]
     val entityIds = workflowsWithEntities.flatMap{ case (workflowRec, outputs) => workflowRec.workflowEntityId }
-    dataAccess.entityQuery.getEntities(workspaceId, entityIds).map(_.toMap)
+
+    dataAccess.workspaceQuery.isWorkspaceSharded(workspaceId).flatMap { sharded =>
```
MatthewBemis: after re-reading this, I am realizing that this one doesn't need to lookup the value. we've already got a `Workspace` object loaded in further up

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -411,8 +413,8 @@ trait EntityComponent {
       EntityRecordRawSqlQuery.activeActionForRefs(workspaceId, entities)
     }
 
-    private def findActiveAttributesByEntityId(workspaceId: UUID, entityId: Rep[Long]): EntityAttributeQuery = for {
-      entityAttrRec <- entityAttributeShardQuery(workspaceId) if entityAttrRec.ownerId === entityId && ! entityAttrRec.deleted
+    private def findActiveAttributesByEntityId(workspaceId: UUID, sharded: Option[Boolean], entityId: Rep[Long]): EntityAttributeQuery = for {
```
MatthewBemis: most of the changes in this file past this point are just changing signatures to weave through the new `sharded` value

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
```diff
@@ -41,9 +41,9 @@ trait DataAccess
     (0L to 15L) map { secondLong =>
       val first = firstLong.toHexString
       val second = secondLong.toHexString
-      determineShard(java.util.UUID.fromString(s"$first${second}000000-0000-0000-0000-000000000000")).toString
+      determineShard(java.util.UUID.fromString(s"$first${second}000000-0000-0000-0000-000000000000"), shardedOpt = Option(true)).toString
     }
-  }).toSet
+  }).toSet + determineShard(java.util.UUID.fromString(s"00000000-0000-0000-0000-000000000000"), shardedOpt = Option(false)).toString
```
MatthewBemis: make sure to add the `archived` EA table to the list of shards

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -180,38 +180,39 @@ trait AttributeComponent {
...
-        Sharded workspaces would use a shard identifier such as "04_07", while unsharded workspaces
-        would use a shard identifier of "archive". This would translate into referencing tables named
+  def determineShard(workspaceId: UUID, shardedOpt: Option[Boolean]): ShardId = {
+    /* Sharded workspaces use a shard identifier such as "04_07", while unsharded workspaces
+        use a shard identifier of "archived". This translates into referencing tables named
         ENTITY_ATTRIBUTE_04_07 or ENTITY_ATTRIBUTE_archive, respectively.
      */
 
-    val shardSize = 4 // range of characters in a shard, e.g. 4 = "00-03"
+    val sharded = shardedOpt.getOrElse(throw new RawlsException(s"Unexpected shard status for workspace $workspaceId. Shard value was null."))
```
MatthewBemis: this really shouldn't ever happen because Liquibase backfills. 

so why did I make this an Option? you'll see later on, but there is a point where `WorkspaceDetails` (the user facing object) gets converted into a `Workspace` object. I did not want to add the `sharded` value to the user-facing object, so when that conversion happens, it felt cleaner to set it to `None` rather than a potentially incorrect `false` or `true`. It does make the code _very slightly_ more confusing but I think it is worth the tradeoff to avoid exposing this value to the users

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211006_add_sharded_column.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="add_workspace_sharded_column" author="mbemis" logicalFilePath="dummy">
+        <addColumn tableName="WORKSPACE">
+            <column name="sharded" type="boolean" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
```
davidangb, line 10: could this be a number column, or even a string column, instead of boolean? That would give us some future-proofing.
* 1 = unsharded
* 2 = sharded, release 1.0
* 3 = sharded, release 2.0 (if we need this)
* 4 = db per workspace
* 5 = ???? future stuff ????

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -395,7 +395,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
         // Delete components of the workspace
         _ <- dataAccess.submissionQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
         _ <- dataAccess.methodConfigurationQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
-        _ <- dataAccess.entityQuery.deleteFromDb(workspaceContext.workspaceIdAsUUID)
+        _ <- dataAccess.entityQuery.deleteFromDb(workspaceContext)
```
davidangb, line 398: this would be so much easier if our foreign keys used cascading deletes! just thinking out loud, no action necessary for this PR

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -375,13 +375,13 @@ trait EntityComponent {
     private object EntityDependenciesDeletionQuery extends RawSqlQuery {
       val driver: JdbcProfile = EntityComponent.this.driver
 
-      def deleteAction(workspaceId: UUID): WriteAction[Int] = {
-        val shardId = determineShard(workspaceId)
+      def deleteAction(workspaceContext: Workspace): ReadWriteAction[Int] = {
+        val shardId = determineShard(workspaceContext.workspaceIdAsUUID, workspaceContext.sharded)
```
davidangb: why did this change from `WriteAction` to `ReadWriteAction`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -690,15 +690,15 @@ trait EntityComponent {
...
     }
 
     // perform actual deletion (not hiding) of all entities in a workspace
 
-    def deleteFromDb(workspaceId: UUID): WriteAction[Int] = {
-      EntityDependenciesDeletionQuery.deleteAction(workspaceId) andThen {
-        filter(_.workspaceId === workspaceId).delete
+    def deleteFromDb(workspaceContext: Workspace): ReadWriteAction[Int] = {
+      EntityDependenciesDeletionQuery.deleteAction(workspaceContext) andThen {
+        filter(_.workspaceId === workspaceContext.workspaceIdAsUUID).delete
```
davidangb, line 701: another change from `WriteAction` to `ReadWriteAction` - necessary?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
davidangb: This is initially scary to me, I worry that some code might call this and inherit a `true` when it shouldn't have a `true`. But I also see this is only called by unit tests … bleh. I feel wonky about this but I don't have a better suggestion.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
davidangb: Would it make this PR easier/lighter if you did include the new sharded field in `WorkspaceDetails`, but then changed [`WorkspaceDetailsFormat`](https://github.com/broadinstitute/rawls/blob/af4675058fb1a53ec0f672175a9dccf398da0fa6/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala#L892) such that it never serializes the "sharded" field to json, so users wouldn't see it?

davidangb (commented):
a few comments inline, nothing too big, I am very close to a :+1: on this!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: An earlier iteration of the code had `determineShard` returning a `ReadAction`, so I had to change this return type. Will revert

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
MatthewBemis: I like this idea and I've included it in a recent commit

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211006_add_sharded_column.xml
MatthewBemis: Love it, I've updated the code accordingly

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -637,8 +655,30 @@ object WorkspaceFieldNames {
 
 object WorkspaceDetails {
   def apply(workspace: Workspace, authorizationDomain: Set[ManagedGroupRef]): WorkspaceDetails = {
-    fromWorkspaceAndOptions(workspace, Option(authorizationDomain),true)
+    fromWorkspaceAndOptions(workspace, Option(authorizationDomain),true).copy(shardState = None)
```
MatthewBemis, line 659: Removes the `shardState` field when this `apply` is used

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -637,8 +655,30 @@ object WorkspaceFieldNames {
...
+  }
+
+  def applyOmitShardState(namespace: String,
```
MatthewBemis, line 662: for use by the json marshaller so we don't even return the internal `shardState` field

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -657,7 +697,8 @@ object WorkspaceDetails {
       workspace.googleProjectNumber,
       workspace.currentBillingAccountOnGoogleProject,
       workspace.billingAccountErrorMessage,
-      workspace.completedCloneWorkspaceFileTransfer
+      workspace.completedCloneWorkspaceFileTransfer,
+      Option(workspace.shardState)
```
MatthewBemis: This should actually be `None`

model/src/test/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModelSpec.scala
```diff
@@ -406,7 +406,7 @@ class WorkspaceModelSpec extends AnyFreeSpec with Matchers {
     "should introspect WorkspaceDetails correctly" in {
       val expected = List("namespace", "name", "workspaceId", "bucketName", "workflowCollectionName", "createdDate",
         "lastModified", "createdBy", "attributes", "isLocked", "authorizationDomain", "googleProject",
-        "googleProjectNumber", "workspaceVersion", "billingAccount", "billingAccountErrorMessage", "completedCloneWorkspaceFileTransfer")
+        "googleProjectNumber", "workspaceVersion", "billingAccount", "billingAccountErrorMessage", "completedCloneWorkspaceFileTransfer", "shardState")
```
MatthewBemis, line 409: Despite the `shardState` field never being returned, it's still technically there so I had to update these 2 tests

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -582,8 +599,9 @@ case class WorkspaceDetails(namespace: String,
                             googleProjectNumber: Option[GoogleProjectNumber],
                             billingAccount: Option[RawlsBillingAccountName],
                             billingAccountErrorMessage: Option[String] = None,
-                            completedCloneWorkspaceFileTransfer: Option[DateTime]) {
-  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProject, googleProjectNumber, billingAccount, billingAccountErrorMessage, completedCloneWorkspaceFileTransfer)
+                            completedCloneWorkspaceFileTransfer: Option[DateTime],
+                            shardState: Option[WorkspaceShardState]) {
+  def toWorkspace: Workspace = Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes.getOrElse(Map()), isLocked, workspaceVersion, googleProject, googleProjectNumber, billingAccount, billingAccountErrorMessage, completedCloneWorkspaceFileTransfer, shardState.getOrElse(WorkspaceShardStates.Unknown))
```
MatthewBemis, line 605: The `Unknown` state here has to do with converting back and forth between `Workspace` and `WorkspaceDetails`. If the shard state is present, then use it. But that information probably won't be there so default to `Unknown`. Note that this is only used by unit tests so this Unknown state should never bleed into reality

MatthewBemis: Also, if there are thoughts on how to better handle this- I would welcome that!

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -191,7 +193,7 @@ object Workspace {
     val randomString = java.util.UUID.randomUUID().toString
     val googleProjectId = GoogleProjectId(randomString)
     val googleProjectNumber = GoogleProjectNumber(randomString)
-    new Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes, isLocked, WorkspaceVersions.V2, googleProjectId, Option(googleProjectNumber), None, None, Option(createdDate))
+    new Workspace(namespace, name, workspaceId, bucketName, workflowCollectionName, createdDate, lastModified, createdBy, attributes, isLocked, WorkspaceVersions.V2, googleProjectId, Option(googleProjectNumber), None, None, Option(createdDate), shardState = WorkspaceShardStates.Sharded)
```
davidangb, line 197: 1) tiny nit: please update the code comment "defaults workspace version to v2 …" to include that this also default the workspace to sharded

2) Buuuut … if this constructor will be removed when PPW is complete, will it cause any problems for the sharding code?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211006_add_sharded_column.xml
sagehen03: I prefer to use enums or strings in this type of situation so that there's less indirection--and it looks like were' going with strings which works!

sagehen03 (approved):
I don't see any clear issues.  This is more code than I was expecting to allow us the flexibility of shard states--the delta here with David's original sharding branch is big.  However, it is what is and :crossed_fingers: that our tests have ensured that we got everything.

davidangb (approved):
one tiny comment inline. I'd love to be able to pull this code OUT at some point, but it may be that it has a significant shelf-life, will all depend on when/if we migrate 100% of the workspaces.

MatthewBemis:
@sagehen03 I should have given this background but me and David are both in agreement that this required more code than we expected. The original approach was to add a DB lookup for the `shardState` to `determineShard`, which would have kept the code very well-contained. The issue ended up being that the return type of `determineShard` would change from `ShardId` to `ReadAction[ShardId]`, which required changing every upstream consumer of that to deal with the added database action. There were some other limitations related to that which made construction of some of the `Query` objects pretty gnarly and/or impossible. Thus, I decided to include it in the `Workspace` object instead and pass that around everywhere. The one benefit of doing it this way is that it means we almost never need to do the extra query to grab the `shardState` because that work has already been done upstream

PR #1545, opened 2021-10-07
ensure workflow messages are returned [BW-845]

davidangb:
I helped debug some failed workflows … the user needed help because the workflow failure messages were not displayed in the Terra UI, nor were they returned from the get-submission API that powers the UI.

I got curious why they weren't returned, and here's the seed of a fix. This problem happens for any workflow whose `WORKFLOW.ENTITY_ID` column is null in the database. In the use case I debugged, the user hit an error in DRS resolution when trying to start the workflow.

The problem happens because if `WORKFLOW.ENTITY_ID` is null, `workflowRec.workflowEntityId.map` never processed its code, and thus `getWorkflowMessagesListResult` method never called a bunch of `r.<<` column extractors. Since these were never called, by the time it tried to extract the message, it was still on column 11, which is `ENTITY.name`, and `ENTITY.name` was null. So it thought it had no messages.

Screenshot

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
@@ -462,16 +462,9 @@ trait SubmissionComponent {
         val workflowRec = WorkflowRecord(r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<)
         val rootEntityTypeOption: Option[String] = r.<<
 
-        // val entityRec = workflowRec.workflowEntityId.map(EntityRecord(_, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
-        val entityRec = workflowRec.workflowEntityId match {
-          case Some(eid) => Option(EntityRecord(eid, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
-          case None =>
-            // consume the entity record columns, but ignore the result
-            val _ = EntityRecord(-1, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<)
-            None
-        }
-
         val messageOption: Option[String] = r.<<
+
+        val entityRec = workflowRec.workflowEntityId.map(EntityRecord(_, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
 
         val externalEntityOption = for {
           rootEntityType <- rootEntityTypeOption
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -461,7 +461,15 @@ trait SubmissionComponent {
...
-        val entityRec = workflowRec.workflowEntityId.map(EntityRecord(_, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
+
+        // val entityRec = workflowRec.workflowEntityId.map(EntityRecord(_, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
+        val entityRec = workflowRec.workflowEntityId match {
+          case Some(eid) => Option(EntityRecord(eid, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<))
+          case None =>
+            // consume the entity record columns, but ignore the result
+            val _ = EntityRecord(-1, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<)
+            None
+        }
```
davidangb: you could also solve this by re-ordering the column selection in WorkflowAndMessagesRawSqlQuery.action so the messages come before the entity, so it doesn't matter if you consume the entity columns or not

pshapiro4broad: Your analysis makes sense to me. I think changing the query to move the message before the entity record fields would be better than reading fields that aren't needed.

Looking at this code, all I can think about is how in ANSI/ISO C you can't rely on order of parameter evaluation and this kind of code would be totally broken there.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponentSpec.scala
```diff
@@ -239,6 +240,72 @@ class SubmissionComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+          None, 1, Option("unittestdefault"), None),
+        WorkflowRecord(22222, Option("workflow2"), sid, WorkflowStatuses.Failed.toString, Timestamp.from(java.time.Instant.now()),
+          None, 1, Option("unittestdefault"), None),
+      )
+      val saveWorkflows = workflowQuery ++= workflows
+      runAndWait(saveWorkflows)
+      val actualWorkflows = runAndWait(workflowQuery.findWorkflowsBySubmissionId(sid).result)
+      withClue("checking count of saved workflows from findWorkflowsBySubmissionId: ") {
+        actualWorkflows.size shouldBe 2
+      }
```
davidangb, line 276: I don't know what use cases cause this particular shape of data to land in the database - `WorkflowRecord.workflowEntityId` is `None` - and I couldn't easily find a higher-level method to insert that data, so I'm just inserting the raw rows.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponentSpec.scala
```diff
@@ -239,6 +240,72 @@ class SubmissionComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers
...
+        withClue(s"checking count of workflow failure messages returned from loadSubmission for ${w.workflowId}: ") {
+          w.messages.size shouldBe 1
+        }
+        withClue(s"checking content of workflow failure messages returned from loadSubmission for ${w.workflowId}: ") {
+          w.messages.head.value shouldBe s"failure message for ${w.workflowId.getOrElse("")}"
+        }
+      })
+    }
+  }
+
```
davidangb, line 308: I verified that this unit test fails without the changes in `SubmissionComponent`, but passes after those changes.

salonishah11 (approved):
LGTM 👍  
Thank you for fixing this!

cjllanwarne (approved):

PR #1546, opened 2021-10-08
Array with single element inside WDL Struct doesn't break workflows [BW-678]

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/BW-678

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
@@ -1,12 +1,17 @@
 package org.broadinstitute.dsde.rawls.entities.base
 
+import cromwell.client.model.ValueType.TypeNameEnum
+import cromwell.client.model.ValueTypeObjectFieldTypes
 import org.antlr.v4.runtime.tree.ParseTree
 import org.broadinstitute.dsde.rawls.entities.base.ExpressionEvaluationSupport.{EntityName, ExpressionAndResult, LookupExpression}
 import org.broadinstitute.dsde.rawls.expressions.JsonExpressionEvaluator
 import org.broadinstitute.dsde.rawls.expressions.parser.antlr.ReconstructExpressionVisitor
+import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver.MethodInput
 import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, AttributeNull, AttributeNumber, AttributeString, AttributeValue, AttributeValueRawJson}
-import spray.json.{JsArray, JsBoolean, JsNull, JsNumber, JsString, JsValue}
+import spray.json.{JsArray, JsBoolean, JsNull, JsNumber, JsObject, JsString, JsValue}
 
+import scala.collection.JavaConverters._
+import scala.collection.mutable
 import scala.util.{Success, Try}
 
 object InputExpressionReassembler {
@@ -32,7 +37,8 @@ object InputExpressionReassembler {
     */
   def constructFinalInputValues(lookupExpressionAndResults: Seq[ExpressionAndResult],
                                 parsedInputExpression: ParseTree,
-                                rootEntityNames: Option[Seq[EntityName]]): Map[EntityName, Try[Iterable[AttributeValue]]] = {
+                                rootEntityNames: Option[Seq[EntityName]],
+                                input: Option[MethodInput]): Map[EntityName, Try[Iterable[AttributeValue]]] = {
     // unpack the evaluated AttributeValue to JsValue and transform it into sequence of tuples with entity name as key
     val seqOfEntityToLookupExprAndValue: Seq[(EntityName, (LookupExpression, Try[JsValue]))] = unpackAndTransformEvaluatedOutput(lookupExpressionAndResults)
 
@@ -43,7 +49,7 @@ object InputExpressionReassembler {
     val mapOfEntityToEvaluatedExprMap: Map[EntityName, Try[Map[LookupExpression, JsValue]]] = convertEvaluatedExprSeqToMap(mapOfEntityToSeqOfEvaluatedExpr)
 
     // replace the value for evaluated attribute references in the input expression for each entity name
-    val mapOfEntityToInputExpr: Map[EntityName, Try[JsValue]] = reconstructInputExprForEachEntity(mapOfEntityToEvaluatedExprMap, parsedInputExpression, rootEntityNames)
+    val mapOfEntityToInputExpr: Map[EntityName, Try[JsValue]] = reconstructInputExprForEachEntity(mapOfEntityToEvaluatedExprMap, parsedInputExpression, rootEntityNames, input)
 
     /*
       For each entity name and it's generated expression call JsonExpressionEvaluator to reconstruct the desired return type
@@ -92,6 +98,29 @@ object InputExpressionReassembler {
        */
   private def unpackAndTransformEvaluatedOutput(seqOfTuple: Seq[ExpressionAndResult])
   : Seq[(EntityName, (LookupExpression, Try[JsValue]))] = {
+
+//    inputOption match {
+//      case Some(input) => input.workflowInput.getValueType.getTypeName match {
+//        case TypeNameEnum.OBJECT => ???
+//        case _ => ???
+//      }
+//      case None => ???
+//    }
+//
+//    def isObjectFieldArray(lookupExpr: LookupExpression) = {
+//      inputOption.map { input =>
+//        val abc: mutable.Seq[ValueTypeObjectFieldTypes] = input.workflowInput.getValueType.getObjectFieldTypes.asScala
+//        abc.collect {
+//          case x if x.getFieldName == lookupExpr =>
+//            x.getFieldType.getTypeName == TypeNameEnum.ARRAY
+//        }
+//      }.get.head
+//    }
+//
+//    val isInputRawJson = inputOption.exists(input => input.workflowInput.getValueType.getTypeName == TypeNameEnum.OBJECT)
+//
+//    if(isInputRawJson && isObjectFieldArray(lookupExpr))
+
     seqOfTuple.flatMap {
       case (lookupExpr, slickEvaluatedAttrValueMap) =>
         // returns  Map[EntityName, (LookupExpression, Try[JsValue])]
@@ -169,17 +198,50 @@ object InputExpressionReassembler {
    */
   private def reconstructInputExprForEachEntity(mapOfEntityToEvaluatedExprMap: Map[EntityName, Try[Map[LookupExpression, JsValue]]],
                                                 parsedTree: ParseTree,
-                                                rootEntityNames: Option[Seq[EntityName]]): Map[EntityName, Try[JsValue]] = {
+                                                rootEntityNames: Option[Seq[EntityName]],
+                                                inputOption: Option[MethodInput]): Map[EntityName, Try[JsValue]] = {
     // when there are no root entities handle as a single root entity with empty string for name
     rootEntityNames.getOrElse(Seq("")).map { entityName =>
       // in the case of literal JSON there are no LookupExpressions
       val evaluatedLookupMapTry = mapOfEntityToEvaluatedExprMap.getOrElse(entityName, Success(Map.empty[LookupExpression, JsValue]))
-      val inputExprWithEvaluatedRef = evaluatedLookupMapTry.map { lookupMap =>
+      val inputExprWithEvaluatedRef: Try[JsValue] = evaluatedLookupMapTry.map { lookupMap =>
         val visitor = new ReconstructExpressionVisitor(lookupMap)
         visitor.visit(parsedTree)
       }
 
-      entityName -> inputExprWithEvaluatedRef
+      val updatedMap: Try[JsValue] = inputOption match {
+        case Some(input) => input.workflowInput.getValueType.getTypeName match {
+          case TypeNameEnum.OBJECT => inputExprWithEvaluatedRef.map { evaluatedMap =>
+            val updatedObject: Map[String, JsValue] = evaluatedMap.asJsObject.fields.map { case (key, value) =>
+              val objectFields: mutable.Seq[ValueTypeObjectFieldTypes] = input.workflowInput.getValueType.getObjectFieldTypes.asScala
+              val isFieldAnArray: Boolean = objectFields.collect {
+                case x if x.getFieldName == key =>
+                  x.getFieldType.getTypeName == TypeNameEnum.ARRAY
+              }.head
+
+              val updatedValue: JsValue = if(isFieldAnArray){
+                value match {
+                  case v: JsBoolean => JsArray(v)
+                  case v: JsNumber => JsArray(v)
+                  case v: JsString => JsArray(v)
+                  case _ => value
+                }
+              }
+              else
+                value
+
+              key -> updatedValue
+            }
+            JsObject(updatedObject)
+          }
+          case _ =>
+            inputExprWithEvaluatedRef
+        }
+        case None =>
+          inputExprWithEvaluatedRef
+      }
+
+      entityName -> updatedMap
     }.toMap
   }
 }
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
```diff
@@ -169,17 +175,61 @@ object InputExpressionReassembler {
    */
   private def reconstructInputExprForEachEntity(mapOfEntityToEvaluatedExprMap: Map[EntityName, Try[Map[LookupExpression, JsValue]]],
                                                 parsedTree: ParseTree,
-                                                rootEntityNames: Option[Seq[EntityName]]): Map[EntityName, Try[JsValue]] = {
+                                                rootEntityNames: Option[Seq[EntityName]],
+                                                inputOption: Option[MethodInput]): Map[EntityName, Try[JsValue]] = {
+
+    def updateInputExprValueMapForObjectInput(inputExprWithEvaluatedRef: JsValue, objectFieldsFromInput: mutable.Seq[ValueTypeObjectFieldTypes]): JsObject = {
+      val updatedObject: Map[String, JsValue] = inputExprWithEvaluatedRef.asJsObject.fields.map { case (inputName, evaluatedValue) =>
```
cahrens, line 185: I'm finding this difficult to read/understand. Why does there need to be a case statement (don't we do this for all keys/values)?

salonishah11: Yes. We are iterating for all (key, value) pairs inside the map. Since the data type of fields inside a JsObject is Map[String, JsValue], we can use a case statement to identify the key and value separately. This can also be written as below, but using case statement and having explicit variable names makes it more readable. In my opinion, using `obj._1` and `obj._2` makes it harder to understand/read what is going on. We can always extract `obj._1` and `obj._2` into variables but case statement implicitly does that.
```
inputExprWithEvaluatedRef.asJsObject.fields.map { obj =>
// in this case obj._1 will be key of type String and obj._2 will be the value of type JsValue

val inputObjectField: Option[ValueTypeObjectFieldTypes] = objectFieldsFromInput.find(x => x.getFieldName == obj._1)
...
}
```

salonishah11:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
cahrens: Thanks! Doing operations through case statements still doesn't feel intuitive to me yet, but hopefully it will soon!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
```diff
@@ -169,17 +175,61 @@ object InputExpressionReassembler {
...
+        val inputObjectField: Option[ValueTypeObjectFieldTypes] = objectFieldsFromInput.find(x => x.getFieldName == inputName)
+
+        val updatedValue: JsValue = inputObjectField match {
+          case Some(objectField) => objectField.getFieldType.getTypeName match {
+            case TypeNameEnum.OBJECT =>
+              // if there are nested objects, recursively call this method to reach leaf fields
```
cahrens, line 191: Do you have a test case that covers this? I'm having trouble understanding the test cases.

salonishah11: Yes. [This](https://github.com/broadinstitute/rawls/blob/e7065edb153ba75f5fc47bdc436d12c67f11f15f/core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala#L121-L131) test case should cover it

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
```diff
@@ -169,17 +175,61 @@ object InputExpressionReassembler {
    */
   private def reconstructInputExprForEachEntity(mapOfEntityToEvaluatedExprMap: Map[EntityName, Try[Map[LookupExpression, JsValue]]],
                                                 parsedTree: ParseTree,
-                                                rootEntityNames: Option[Seq[EntityName]]): Map[EntityName, Try[JsValue]] = {
+                                                rootEntityNames: Option[Seq[EntityName]],
+                                                inputOption: Option[MethodInput]): Map[EntityName, Try[JsValue]] = {
+
+    def updateInputExprValueMapForObjectInput(inputExprWithEvaluatedRef: JsValue, objectFieldsFromInput: mutable.Seq[ValueTypeObjectFieldTypes]): JsObject = {
```
cjllanwarne, line 184: Docs please!

cjllanwarne: eg point out that this is specifically for inputs which are formatted like json objects

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
cjllanwarne: It looks like we have one test, but this kind of code is always potentially tricky. We should probably validate all possible input permutations (and possibly highlight known bugs with "we should fix this known problem" test cases.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
```diff
@@ -169,17 +175,61 @@ object InputExpressionReassembler {
...
       }
 
-      entityName -> inputExprWithEvaluatedRef
+      /* In SlickExpressionEvaluator, for attribute references that evaluate to AttributeValueList type, the elements inside the list are extracted and returned.
+         As a result, for attributes that contain only 1 element, we can no longer identify whether the original value was a list or a single AttributeValue.
+         Because of this type erasure, when we are reconstructing object type inputs, List/Array with single element get incorrectly reconstructed as single AttributeValue,
+         which leads to Cromwell failing while processing inputs (https://broadworkbench.atlassian.net/browse/BW-678).
+         To fix this, once the object type inputs have been reconstructed, we iterate over each field for object input and verify for Array type inputs that the
+         evaluated value is also an Array. If it isn't we convert it into Array and updated the evaluated object.
+         Note: This approach does not fix the problem if there is single element array inside an array. This would be fixed in https://broadworkbench.atlassian.net/browse/BW-846.
```
cjllanwarne, line 223: Or you could have an equivalent, recursive `updateInputExprValueMapForArrayInput`, presumably?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
cjllanwarne: And that it is called further down this method for generating the `updatedEvaluatedRefMap` if the input is an object, in order to array-ify any not-yet-detected array values

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
cjllanwarne: By permutations I mean things like:

Array types:
- Empty arrays
- Single element arrays
- Multi-element arrays

In places like:
- As top-level inputs
- Fields in objects
- Fields in nested objects
- Values in arrays
- Values in nested arrays

etc

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
salonishah11: Hmm, interesting point. Just so that I am understanding correctly, if the input type is an Array, we recursively call `updateInputExprValueMapForArrayInput` until the type inside an Array is not an Array. I think it would be possible since the fieldType of nested Array is an Array 🤔

cjllanwarne: I'm not sure, I was thinking you could just have an equivalent function to the one that looks inside objects for looking inside arrays... but I'm not sure. I wasn't suggesting _doing_ this, more that it's another option when we do come to look at this story again

salonishah11: Ah I see what you mean. Yeah looking inside Array inputs might be tricky especially for deeper nested arrays. I would hope if we look at this bug again, we would solve the problem for real by fixing the type erasure.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/InputExpressionReassembler.scala
salonishah11: Sounds good. I will add more permutations. I think multi-element array tests are already present. I will add the rest.

salonishah11: I am still working on the tests, but I was wondering how do we test the test cases that we know fail? For example, Single element array as values in array or values in nested arrays will fail. How do we assert that?

salonishah11: I have added test cases from the permutation list that was not previously present. Test cases for different array types as top-level inputs were already present.

cjllanwarne (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
```diff
@@ -82,6 +82,18 @@ class LocalEntityProviderSpec extends AnyWordSpecLike with Matchers with TestDri
         Map(sampleSet2.name -> Seq(SubmissionValidationValue(Some(AttributeValueEmptyList), None, intArrayNameWithWfName)))
     }
 
+    "resolve empty lists into empty Array in nested WDL Struct" in withConfigData {
+      val context = workspace
```
cahrens, line 86: Nit: you could refactor these test points to share the common code. The inputs and expected outputs vary, but everything else is the same.

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigTestSupport.scala
```diff
@@ -325,7 +439,8 @@ trait MethodConfigTestSupport {
       AttributeName.withDefaultNS("samples") -> AttributeEntityReferenceList(Seq(
         sampleGood.toReference,
         sampleGood2.toReference
-      ))
+      )),
+      AttributeName.withDefaultNS("rawJsonDoubleArray") -> AttributeValueRawJson( "[[0,1,2],[3,4,5]]".parseJson )
```
cahrens, line 443: Why did you add this?

salonishah11: This was added for test case where we have multi-element array inside an array

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProviderSpec.scala
salonishah11: Since this is test code, I am inclined to leave it like this as in my opinion time involved in refactoring outweighs the benefits

salonishah11: Having said that, if you do feel strongly, I can refactor them :)

cahrens: No, I don't feel strongly about it-- that's why I prefaced it with "Nit".

In Python unit tests, there was a nice way to annotate a test method as something that should run with a list of inputs/expected outputs, and the test harness then ran the test point with each set of inputs/outputs individually. It was a very convenient way to keep the tests concise and maintainable. Are you aware of anything like this in Scala?

salonishah11: I am not sure. I will look it up 👍

salonishah11: I was able to find [this](https://github.com/broadinstitute/rawls/blob/7bc2274032277fd9605495fbeb4c7a97daefc370/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala#L837-L872) pattern. Is this what you had in mind?

cahrens: Yes, but I was wondering if there were language constructs/libraries to make this easier. For example, the ddt library in Python as shown below.

Don't go searching if you don't know… I was just curious if you were aware of anything like this.

````python
    @data(
        ['Alewife', 'Alewife', [['Red Line']]],
        ['Shawmut', 'Quincy Adams', [['Red Line']]],
        ['Downtown Crossing', 'Alewife', [['Red Line']]],
        ['Downtown Crossing', 'Haymarket', [['Orange Line']]],
        ['Ashmont', 'Milton', [['Mattapan Trolley']]],
        ['Haymarket', 'Prudential', [['Green Line E']]],
        ['Park Street', 'North Station', [['Green Line C'], ['Green Line E']]],
        ['Ruggles', 'Wonderland', [['Orange Line', 'Blue Line']]],
        ['South Station', 'State', [['Red Line', 'Orange Line']]],
        ['Copley', 'Chinatown', [['Green Line C', 'Orange Line'],
                                 ['Green Line E', 'Orange Line']]],
        ['Fenway', 'North Station', [['Green Line D', 'Green Line C'],
                                     ['Green Line D', 'Green Line E']]],
        ['Mattapan', 'Roxbury Crossing', [['Mattapan Trolley', 'Red Line', 'Orange Line']]],
        ['Broadway', 'Aquarium', [['Red Line', 'Orange Line', 'Blue Line'],
                                  ['Red Line', 'Green Line C', 'Blue Line'],
                                  ['Red Line', 'Green Line D', 'Blue Line'],
                                  ['Red Line', 'Green Line E', 'Blue Line']]],
        ['Fenway', 'Assembly', [['Green Line D', 'Green Line C', 'Orange Line'],
                                ['Green Line D', 'Green Line E', 'Orange Line'],
                                ['Green Line D', 'Blue Line', 'Orange Line']]]
    )
    @unpack
    def test_get_route(self, source, destination, possible_routes):
        """
        Tests the get_route method (Question 3). If multiple routes are possible,
        makes sure that the route returned is in the list of possible routes.
        """
        route = self.subway.get_route(source, destination)
        self.assertIn(
            route, possible_routes,
            'Route incorrect for stops {source} -> {destination}. Expected one of {possible_routes}, but got {route}.'.format(
                source=source, destination=destination, possible_routes=possible_routes, route=route
            )
        )

salonishah11: No I am not aware of such a library. I can look into it outside of this PR when I have some time (out of curiosity).

cahrens: I've been Googling a bit and haven't found anything yet. It's clearly just "syntactic sugar", but I found that it reduced my tendency to copy/paste test code, and it makes the tests easier to understand because the team becomes familiar with a single syntax, vs. everyone rolling their own syntax for this sort of thing.

PR #1549, opened 2021-10-08
manage invalid entity statistic caches [AS-930]

davidangb:
**_I have created an alert that should ping the #dsp-appservices-notifications Slack channel if any statistics caches become invalid. The alert is viewable via cloud console [here](https://console.cloud.google.com/monitoring/alerting/policies/11623762382109243788?project=broad-dsde-prod) (requires @ firecloud.org account)_**

This adds two new admin APIs:
* list those workspaces which have an invalid entity statistics cache, along with the cache's error message
* reset the entity statistics cache for specific workspaces, so the cache can be rebuilt and become valid again

I tested this a bunch manually, as well as adding unit tests for the APIs.

IMPORTANT: this PR also carries a change for anyone running unit tests locally. The flag `-Duser.timezone=UTC` is now required in `SBT_OPTS` in order for unit tests to pass. The new unit tests rely on sql `Timestamp`s, and when the client (Rawls) generated a timestamp and sent it over JDBC, the timestamp would be converted based on the time zone difference between Rawls and the MySQL server. In live environments like dev and prod, as well as in Jenkins and Github Actions, both the client and the server were in UTC so no conversions occurred. But, if you're running unit tests locally on your laptop, you are likely in the Eastern timezone, and therefore timestamps would be offset by a few hours.

This specifically caused a problem with the `MIN_CACHE_TIME` value used by the entity statistics cache. When MySQL translated it, it resulted in a value of `1969-12-31 19:00:01.0` instead of the UTC `1970-01-01 00:00:01.0` … and that translated value is actually below the minimum allowable value for a timestamp column in MySQL. Thus, the translation caused rows to fail to insert.

I've addressed this issue by running the unit tests in UTC, via the addition of the `-Duser.timezone=UTC` flag. This seemed the most appropriate approach, since it mimics the behavior of our live envs; other rejected options include fussing with JDBC connection string options, but that would make test connections behave differently than live connections.

-----

See also #1551, which is an alternate approach to AS-930.

My take on how the two PRs compare:

| #1549 (this PR) - admin APIs | #1551 - helper scripts |
| ------ | ------- |
| unit testable, has tests | manual testing only |
| adds more runtime code to Rawls | standalone, doesn't make Rawls bigger |
| all invocations are logged | no audit logging |
| more of a black box, need Scala source to understand behavior | easier for a developer to read code |
| requires only a browser | requires source code to exist locally, docker, and VPN |
| may be overkill, since invalid caches should now be rare | easier to excise, if we eliminate invalid caches |
| behavior changes require releasing/deploying Rawls | behavior changes require a merge to github and/or tweaks to local code |
| does not require devs to write directly to the database | but it's ok for devs to write directly to the database |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,7 @@ jobs:
           -Denv.type=test
           -Dmysql.host=localhost
           -Dmysql.port=3310
+          -Duser.timezone=UTC
           it:compile
 
       # A known github bug results in 'annotations generated in a Github action during a step executed in
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
@@ -28,6 +28,8 @@ object EntityStatisticsCacheMonitor {
 
   //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
   val MIN_CACHE_TIME = new Timestamp(1000)
+  // when resetting an invalid cache so that it can be recalculated, use this timestamp
+  val RESET_CACHE_TIME = new Timestamp(2000)
 
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1882,7 +1882,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
           } yield {
             val errorMap = workspaceIdsAndErrors.toMap
             workspaces.map { w =>
-              (w, errorMap.getOrElse(w.workspaceIdAsUUID, ""))
+              (w, errorMap.get(w.workspaceIdAsUUID).flatten.getOrElse(""))
             }
           }
         }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1898,6 +1898,20 @@ class WorkspaceService(protected val userInfo: UserInfo,
     }
   }
 
+  def adminResetWorkspaceEntityCaches(ids: Seq[String]) = {
+    asFCAdmin {
+      // validate workspaceIds
+      val workspaceIds = ids.map(s => Try(UUID.fromString(s)).getOrElse(throw new RawlsException(s"all ids must be valid UUIDs. Found: $s")))
+      val updated = dataSource.inTransaction { dataAccess =>
+        dataAccess.entityCacheQuery.resetInvalidCaches(workspaceIds)
+      }
+      updated map { num =>
+        RequestComplete(StatusCodes.OK, Map("numReset" -> num))
+      }
+    }
+
+  }
+
   def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
     //don't do the sam REST call inside the db transaction.
     getWorkspaceContext(workspaceName) flatMap { workspaceContext =>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1872,6 +1872,32 @@ class WorkspaceService(protected val userInfo: UserInfo,
     }
   }
 
+  def adminListWorkspacesWithInvalidEntityCaches: Future[PerRequestMessage] = {
+    asFCAdmin {
+      for {
+        workspacesAndErrors <- dataSource.inTransaction { dataAccess =>
+          for {
+            workspaceIdsAndErrors <- dataAccess.entityCacheQuery.listInvalidCaches
+            workspaces <- dataAccess.workspaceQuery.listByIds(workspaceIdsAndErrors.map(_._1), Option(WorkspaceAttributeSpecs(all=false)))
+          } yield {
+            val errorMap = workspaceIdsAndErrors.toMap
+            workspaces.map { w =>
+              (w, errorMap.getOrElse(w.workspaceIdAsUUID, ""))
+            }
+          }
+        }
+        results <- Future.traverse(workspacesAndErrors) { workspaceAndError =>
+          val workspace = workspaceAndError._1
+          loadResourceAuthDomain(SamResourceTypeNames.workspace, workspace.workspaceId, userInfo).map(WorkspaceDetails(workspace, _)).map { details =>
+            WorkspaceDetailsAndCacheError(details, workspaceAndError._2)
+          }
+        }
+      } yield {
+        RequestComplete(StatusCodes.OK, results)
+      }
+    }
+  }
+
   def getBucketUsage(workspaceName: WorkspaceName): Future[PerRequestMessage] = {
     //don't do the sam REST call inside the db transaction.
     getWorkspaceContext(workspaceName) flatMap { workspaceContext =>
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
@@ -349,6 +349,8 @@ class AdminApiServiceSpec extends ApiServiceSpec {
     // in live environments, both server and client use UTC timezone; test behavior should mimic that.
     val min = MIN_CACHE_TIME
     min shouldBe new Timestamp(1000)
+    withClue("this failure may indicate a time zone mismatch between the mysql server and the client running the unit tests. " +
+      "Do you have -Duser.timezone=UTC set in your SBT_OPTS? Underlying test failure message: ")
     min.toString shouldBe "1970-01-01 00:00:01.0"
   }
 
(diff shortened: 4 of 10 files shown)


davidangb:
`java.net.UnknownHostException: broadinstitute.jfrog.io` in swat build, should be transient. Jenkins retest.

.github/workflows/test.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           -Denv.type=test
           -Dmysql.host=localhost
           -Dmysql.port=3310
+          -Duser.timezone=UTC
```
davidangb, line 51: Github Actions, as well as Jenkins, were passing without this flag. However, it felt correct to add the flag here and in `install.sh` below if we require it for locally-run unit tests, to be consistent.

sagehen03 (commented):
I checked the prod database and I didn't see any records in `WORKSPACE_ENTITY_CACHE` where the cache has failed to update aka `select count(*) from WORKSPACE_ENTITY_CACHE where error_message is not null` returns 0 so I'm going put forth what is likely one of the most unpopular questions in the PR process--are you sure we need to do this?  I understand that we aren't changing existing behavior and there are no callers yet to this code so the risk of this causing trouble in production is probably close to 0.  Nonetheless I think it's inescapable that more code means more places for bugs to lurk and the new change to set the timezone in the jvm args will likely flummox (hopefully briefly) a developer or two running tests locally.  I'm going to go ahead review the review code as if it is necessary, but my primary reaction to this code is to say that it doesn't feel like we need it.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -28,6 +28,8 @@ object EntityStatisticsCacheMonitor {
 
   //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
   val MIN_CACHE_TIME = new Timestamp(1000)
+  // when resetting an invalid cache so that it can be recalculated, use this timestamp
+  val RESET_CACHE_TIME = new Timestamp(2000)
```
sagehen03, line 32: I'm not a fan of magic values.  I understand that doing it this way means the background cache refresh operation code doesn't have to change, but I think we should be more explicit and use columns that specifically say the cache needs refreshing and in what order.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1872,6 +1872,46 @@ class WorkspaceService(protected val userInfo: UserInfo,
     }
   }
 
+  def adminListWorkspacesWithInvalidEntityCaches: Future[PerRequestMessage] = {
```
sagehen03, line 1875: It looks like we're making 2 distinct database calls here.. getting the invalid caches then hydrating those ids to workspaces.  I know it would involve coding up the sql, but I'm sure it would run faster if we made it one query and I don't know the particulars of what slick does for listByIds but if it's a where in (ids) you eliminate the minuscule risk of overflowing the where in clause.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1872,6 +1872,46 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+            workspaces.map { w =>
+              (w, errorMap.get(w.workspaceIdAsUUID).flatten.getOrElse(""))
+            }
+          }
+        }
+        results <- Future.traverse(workspacesAndErrors) { workspaceAndError =>
+          val workspace = workspaceAndError._1
+          loadResourceAuthDomain(SamResourceTypeNames.workspace, workspace.workspaceId, userInfo).map(WorkspaceDetails(workspace, _)).map { details =>
+            WorkspaceDetailsAndCacheError(details, workspaceAndError._2)
+          }
```
sagehen03, line 1893: For my own edification why are we calling sam here?  It seems like we have all the information that we'd need in Rawls' database.  I'm guessing it's to verify that the caller is an admin?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1872,6 +1872,46 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+        }
+      } yield {
+        RequestComplete(StatusCodes.OK, results)
+      }
+    }
+  }
+
+  def adminResetWorkspaceEntityCaches(ids: Seq[String]) = {
```
sagehen03, line 1901: It looks like if one of caches fails to reset, then the whole operation fails?  I'd be in favor of changing the POST to just accepting a single workspace ID at a time.  I think that makes error handling and messaging to the caller cleaner.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -273,4 +276,154 @@ class AdminApiServiceSpec extends ApiServiceSpec {
         }
       }
   }
+
+  // utility functions used by the entity statistics cache tests, following
+  def assertInvalidCacheCount(services: TestApiService, expectedCount: Int, expectedWorkspaces: Seq[Workspace] = Seq(),
```
sagehen03, line 281: I'm a huge fan of your tests--thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
davidangb: I agree - but as before, I'm hoping that work happens in AS-931. I realize this is adding even more tech debt for AS-931 to fix, but I wanted this PR to focus on the admin APIs and not on the cache schema.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: I agree this is less efficient. Since this is an admin API and my expectation is it will get called at most a few times per day, and only by DSP engineers, I opted away from efficiency and towards keeping the code modular and encapsulated: add one new method to identify the workspace ids that have invalid caches, otherwise reuse the pre-existing `workspaceQuery.listByIds` method without making any changes to it. Further, I expect that the number of invalid caches (i.e. the size of the IN clause) will be in single digits or possibly in the tens of ids … if we find that the number of invalid caches is larger than that, we have a much bigger problem than just this admin API

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: This is to populate the [auth domain](https://support.terra.bio/hc/en-us/articles/360026775691-Managing-data-privacy-and-access-with-Authorization-Domains) info for the workspace, so that when listing the workspace, it shows the auth domain. This isn't purely necessary … the result payload for this API is currently only for human consumption, so we can omit anything we want. That said, since this is an admin API I didn't care about the extra call, and if any of these workspaces DO have an auth domain, I thought it would be helpful to display it instead of a misleading empty `"authorizationDomain": []` just because we didn't populate it

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: good call on this - I admit that since this is an admin API I didn't put too much effort into making it resilient or super user-friendly, hoping that our internal engineers have more tolerance for manually retrying/debugging any problems

davidangb:
@sagehen03 you raise a valid point about whether or not the added code and tests (especially the need for the new `-D` flag is worth the benefit of these APIs. My personal opinion is that you answered the question yourself … when you say "I checked the prod database" I know that means you opened a mysql client directly to prod - potentially with a user credential that has write privilege on the db? - and that's exactly what this PR is meant to avoid. I have personally done the same at least 5 times over the last few months: logged into prod, checked invalid caches, and hand-wrote UPDATE statements to fix the invalid ones. That makes me so uncomfortable, I'd rather put that operation on rails, with authorization and logging, and avoid the risk of human error. It may be that #1520 eliminated invalid caches … but without these APIs the only way to know that is to log into the db and ask, which brings us back to needing these APIs.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
sagehen03: > @sagehen03 you raise a valid point about whether or not the added code and tests (especially the need for the new `-D` flag is worth the benefit of these APIs. My personal opinion is that you answered the question yourself … when you say "I checked the prod database" I know that means you opened a mysql client directly to prod - potentially with a user credential that has write privilege on the db? - and that's exactly what this PR is meant to avoid. I have personally done the same at least 5 times over the last few months: logged into prod, checked invalid caches, and hand-wrote UPDATE statements to fix the invalid ones. That makes me so uncomfortable, I'd rather put that operation on rails, with authorization and logging, and avoid the risk of human error. It may be that #1520 eliminated invalid caches … but without these APIs the only way to know that is to log into the db and ask, which brings us back to needing these APIs.

I disagree with your conclusion here.  First I think we're being excessively afraid of logging into the prod db.  mysql-connect.sh now uses a read only user by default, right?  Also it's unclear to me that we'll ever need to run mysql updates by hand for this problem again; I'd advocate for waiting to put forth a solution to that problem until it actually arises again.  And finally, we're only updating the ENTITY_WORKSPACE_CACHE table. You have to twist yourself into knots to envision a scenario where a bad update causes a significant prod incident.  What would that even look like?

aherbst-broad:

PR #1551, opened 2021-10-15
developer scripts for managing entity statistics caches [AS-930]

davidangb:
**_I have created an alert that should ping the #dsp-appservices-notifications Slack channel if any statistics caches become invalid. The alert is viewable via cloud console [here](https://console.cloud.google.com/monitoring/alerting/policies/11623762382109243788?project=broad-dsde-prod) (requires @ firecloud.org account)_**

This PR adds some bash/sql scripts to:
* list those workspaces which have an invalid entity statistics cache, along with the cache's error message
* reset all invalid entity statistics caches, so those caches can be rebuilt and become valid again

This is an alternate approach to AS-930, as compared to #1549.

My take on how the two PRs compare:

| #1549 - admin APIs | #1551 (this PR) - helper scripts |
| ------ | ------- |
| unit testable, has tests | manual testing only |
| adds more runtime code to Rawls | standalone, doesn't make Rawls bigger |
| all invocations are logged | no audit logging |
| more of a black box, need Scala source to understand behavior | easier for a developer to read code |
| requires only a browser | requires source code to exist locally, docker, and VPN |
| may be overkill, since invalid caches should now be rare | easier to excise, if we eliminate invalid caches |
| behavior changes require releasing/deploying Rawls | behavior changes require a merge to github and/or tweaks to local code |
| does not require devs to write directly to the database | but it's ok for devs to write directly to the database |


sagehen03 (approved):
Big :+1: for adding the slack alert.  I think that's the most important thing. I'm not familiar with that feature for gcloud and would like you to show me and others who may not be familiar how that's set up.  It looks like it's based on looking for certain log pattern?  Anyhow, I think having something that will proactively let us know if there are stale entity caches is the highest pri item related to these caches. 

 My preference would be to wait to add this bash code, the api end points, and the self healing caches until those alerts demonstrate that this will continue to be a problem going forward in the hopes that the alerts never go off and we don't need to add more code for this issue.  However, I understand we are of different mindsets on that so if you'd like for me to chose between the bash scripts and the api changes, I'd vote for these scripts.  I think keeping the ad hoc code out of the Rawls scala codebase is worth the other tradeoffs that you outline in this PR.

davidangb:
> I'm not familiar with that feature for gcloud and would like you to show me and others who may not be familiar how that's set up.

@sagehen03 I'm a few hours ahead of you, I learned this today while exploring the best way to create an alert.
1. In the Logs Explorer inside cloud console, perform the search you care about
2. From the Actions menu, choose "Create log alert"
3. Follow the log alert wizard prompts

aherbst-broad:

PR #1554, opened 2021-10-21
[qa-1571] upgrade workbench libs and migrate to cats 3

andy7i:
QA-1571

This PR (which includes significant contributions from the CI mob) follows up on these PRs:
* https://github.com/broadinstitute/workbench-libs/pull/770
* https://github.com/broadinstitute/rawls/pull/1547
* https://github.com/broadinstitute/firecloud-develop/pull/2685
* https://github.com/broadinstitute/rawls/pull/1558
And will be completed (hopefully) by this PR:
* https://github.com/broadinstitute/rawls/pull/1556

Changes in this PR
Upgrade to cats3 for workbench-libs compatibility
Updates workbench libs 
Optimize imports across rawls (except for XMLs, which we do not want to alter)

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [ ] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
  [alpha](https://console.cloud.google.com/sql/instances/terraform-r4caezzc35c4tb7pgdhwkmme4y/backups?project=broad-dsde-alpha&organizationId=548622027621)
  DBs in Google Cloud Console
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] Tell your tech lead (TL) that the PR exists if they want to look at it
[...]

PR #1558, opened 2021-10-25
remove unused val [QA-1571]

davidangb:
helps with QA-1571 but does not complete that ticket.

This PR removes an unused val. That val caused an upgrade problem in #1554 - since it's unused, let's just remove it instead of solving the upgrade problem. See https://github.com/broadinstitute/rawls/pull/1554#discussion_r734604892 for details.


andy7i (approved):
thanks for the help!

MatthewBemis (approved):

davidangb:
one failure in swat, on `org.broadinstitute.dsde.test.api.RawlsApiSpec.Rawls should be able to run sub-workflow tasks in a cloned workspace in non-US regions`, timeout looking for `""executionStatus":"Done"".`

PR #1560, opened 2021-10-26
discard/subscribe entity on ok status checks [QA-1558]

davidangb:
When checking `/status` on an external service, always consume/discard the response entity body. In theory, this should avoid `Response entity was not subscribed after 1 second.` errors from akka. Note that I do not expect this to solve the pervasive timeout errors we have seen in fiab tests, but this does address one single known sore spot.

Found during QA-1558, does not resolve that ticket.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ServiceDAOWithStatus.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ServiceDAOWithStatus.scala
@@ -15,7 +15,12 @@ trait ServiceDAOWithStatus {
     for {
       response <- http.singleRequest(Get(statusUrl))
       ok = response.status.isSuccess
-      message <- if (ok) Future.successful(None) else Unmarshal(response.entity).to[String].map(Option(_))
+      message <- if (ok) {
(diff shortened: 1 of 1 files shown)


davidangb:
one test failure in Rawls, `org.broadinstitute.dsde.test.api.AuthDomainMultiGroupApiSpec.A workspace with multiple groups in its auth domain when the user is in none of the groups when the user is a billing project owner can be seen but is not accessible`:
```
{"causes":[],"exceptionClass":"com.google.api.client.googleapis.json.GoogleJsonResponseException","message":"409 Conflict\nDELETE https://www.googleapis.com/admin/directory/v1/groups/policy-f527d978-c130-45a3-9049-b1af676cb4cf@quality.firecloud.org\n{\n  \"code\" : 409,\n  \"errors\" : [ {\n    \"domain\" : \"global\",\n    \"message\" : \"Conflicting requests. Please try again\",\n    \"reason\" : \"conflict\"\n  } ],\n  \"message\" : \"Conflicting requests. Please try again\"\n}","source":"sam", ...
```

Jenkins retest

marctalbott (approved):

MatthewBemis (approved):

PR #1561, opened 2021-10-27
AS-1012: Migrate CDC shard and 3 smallest shards [start of Phase 2]

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
@@ -5,259 +5,259 @@
             individual shard, we can restart the liquibase migration at that point,
             and don't have to rollback any progress up to that changeset. -->
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_00_03">-->
-<!--        <sql>call copyRowsToAttributeShard('00_03');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('00_03');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_04_07">-->
-<!--        <sql>call copyRowsToAttributeShard('04_07');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('04_07');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_08_0b">-->
-<!--        <sql>call copyRowsToAttributeShard('08_0b');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('08_0b');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_0c_0f">-->
-<!--        <sql>call copyRowsToAttributeShard('0c_0f');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('0c_0f');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_10_13">-->
-<!--        <sql>call copyRowsToAttributeShard('10_13');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('10_13');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_14_17">-->
-<!--        <sql>call copyRowsToAttributeShard('14_17');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('14_17');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_18_1b">-->
-<!--        <sql>call copyRowsToAttributeShard('18_1b');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('18_1b');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_1c_1f">-->
-<!--        <sql>call copyRowsToAttributeShard('1c_1f');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('1c_1f');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_20_23">-->
-<!--        <sql>call copyRowsToAttributeShard('20_23');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('20_23');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_24_27">-->
-<!--        <sql>call copyRowsToAttributeShard('24_27');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('24_27');</sql>-->
 <!--    </changeSet>-->
 
     <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_28_2b">
-        <sql>call copyRowsToAttributeShard('28_2b');</sql>
+        <sql>call populateShardAndMarkAsSharded('28_2b');</sql>
     </changeSet>
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_2c_2f">-->
-<!--        <sql>call copyRowsToAttributeShard('2c_2f');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('2c_2f');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_30_33">-->
-<!--        <sql>call copyRowsToAttributeShard('30_33');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('30_33');</sql>-->
 <!--    </changeSet>-->
 
 <!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_34_37">-->
-<!--        <sql>call copyRowsToAttributeShard('34_37');</sql>-->
+<!--        <sql>call populateShardAndMarkAsSharded('34_37');</sql>-->
 <!--    </changeSet>-->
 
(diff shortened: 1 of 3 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
```diff
@@ -4,325 +4,261 @@
     <!-- START populate each shard in its own changeset. If we hit an error with an
             individual shard, we can restart the liquibase migration at that point,
             and don't have to rollback any progress up to that changeset. -->
-    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_00_03">
-        <sql>call copyRowsToAttributeShard('00_03');</sql>
-        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_00_03; call createEntityAttributeShard('00_03');</rollback>
```
MatthewBemis, line 9: this rollback procedure is no longer valid, since it could end up deleting data that doesn't exist in the `archived` table

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
```diff
@@ -4,325 +4,261 @@
...
-    </changeSet>
+<!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_00_03">-->
```
MatthewBemis, line 7: I think a reasonable approach is to just uncomment shards once we want to migrate them

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
```diff
@@ -4,325 +4,261 @@
...
-    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_98_9b">
-        <sql>call copyRowsToAttributeShard('98_9b');</sql>
-        <rollback>DROP TABLE IF EXISTS ENTITY_ATTRIBUTE_98_9b; call createEntityAttributeShard('98_9b');</rollback>
-    </changeSet>
+<!--    <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_98_9b">-->
+<!--        <sql>call copyRowsToAttributeShard('98_9b');</sql>-->
+<!--    </changeSet>-->
 
     <changeSet logicalFilePath="dummy" author="davidan" id="populate_shard_9c_9f">
         <sql>call copyRowsToAttributeShard('9c_9f');</sql>
```
MatthewBemis: this is the important shard! the other 3 uncommented are the 3 smallest shards (as of 1-2 weeks ago, at least)

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
davidangb: we can't reuse the changeset though - each migration will need to be a new changeset. EDIT: duh, brainfart … these are all individual changesets. Ignore this comment!

davidangb (changes requested):
A few thoughts:
1. This is missing a critical step of updating the `shard_state` on individual workspaces to `Sharded`. Without that, we just copy the entity attribute rows, but the workspaces will keep using the big table.
2. I think that if we are migrating shards incrementally, we should delete the attribute rows from the big table after we successfully copy them into the shard table. Probably this should happen as part of the same transaction that copies them, so any transaction rollback will cover both operations. I could be swayed into the deletes happening in separate changesets, too - as long as we ensure that the copy operation succeeds before the delete operation runs.

davidangb:
I wonder if the right approach, to make future migrations easier, is:
1. create a new stored proc that updates all workspaces in a given shard to `Sharded`
2. create a new stored proc that deletes all rows in a given shard from the big attribute table
3. create a new "wrapper" stored proc that calls `copyRowsToAttributeShard`, then the two new procs from 1 and 2, and does all of this inside a single transaction
4. use the "wrapper" stored proc inside the liquibase changesets

MatthewBemis:
re: 1)....oops duh. quite the oversight.

re: 2) I thought we had decided as part of [the rollback process](https://docs.google.com/document/d/10zECWn-OiKtOgycm8uPChKPR5ABIBrQ8IuQh7Ibr0Us/edit#) that we would not be deleting the rows from the main table until later on? Or was this only the rollback process for Phase 1?

davidangb:
ah, I thought we hadn't settled on a decision about deleting rows from the main table … I was landing on a decision in this PR thread but I guess you're right! That's fine - though once we get another tranche or two of shards migrated, we might want to reconsider for perf reasons

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
```diff
@@ -239,4 +239,38 @@
...
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mbemis" id="create_shard_wrapper_procedure">
+        <sql stripComments="true" endDelimiter="~">
+            -- procedure to set recently migrated workspaces to Sharded state,
+            DROP PROCEDURE IF EXISTS populateShardAndMarkAsSharded ~
+            CREATE PROCEDURE populateShardAndMarkAsSharded(IN shardIdentifier CHAR(5))
+
+            BEGIN
+                DECLARE exit handler FOR SQLEXCEPTION
```
MatthewBemis, line 262: this exit handler stuff is a bit unfamiliar to me but based on research seemed like the appropriate way to guarantee that the two underlying `call`s were in the same txn and would be rolled back

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
```diff
@@ -239,4 +239,38 @@
         </rollback>
     </changeSet>
 
+    <changeSet logicalFilePath="dummy" author="mbemis" id="create_set_workspace_sharded_procedure">
+        <sql stripComments="true" endDelimiter="~">
+            -- procedure to set recently migrated workspaces to Sharded state,
+            DROP PROCEDURE IF EXISTS setWorkspacesAsSharded ~
+            CREATE PROCEDURE setWorkspacesAsSharded(IN shardIdentifier CHAR(5))
+            BEGIN
+                SET @sql_text:=CONCAT('update WORKSPACE w set shard_state="Sharded" where shardIdentifier(hex(w.id)) = ''',@tableSuffix,''';');
```
MatthewBemis: to reviewer: after reading this again i'm double checking this line right now

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
sagehen03: Good catch! Were you thinking about new workspaces starting to populate those sharded tables or something else?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
sagehen03: If you've tested this and feel confident in it, then by all means, but my opinion is it's overkill.  The migration will run single threaded when I believe nothing else will be accessing the database so a failure seems unlikely, and I believe copyRowsToAttributeShard is already transactional? So the scenario that I think this new code is guarding against is the workspaces for the shard being migrated but then not updating the migration flag which is a state that's pretty easy to remedy, my .02.

sagehen03 (approved):

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
```diff
@@ -239,4 +239,38 @@
...
+            -- procedure to set recently migrated workspaces to Sharded state,
+            DROP PROCEDURE IF EXISTS setWorkspacesAsSharded ~
+            CREATE PROCEDURE setWorkspacesAsSharded(IN shardIdentifier CHAR(5))
+            BEGIN
+                SET @sql_text:=CONCAT('update WORKSPACE w set shard_state="sharded" where shardIdentifier(hex(w.id)) = ''',shardIdentifier,''';');
+                PREPARE stmt from @sql_text;
+                EXECUTE stmt;
+            END ~
+        </sql>
+    </changeSet>
```
davidangb, line 253: nit: in Scala, we set new workspaces to a `shard_state` value of `Sharded` (capitalized). Could you make it consistent here? Else we'll end up with `Sharded` for newly-created workspaces and `sharded` for migrated workspaces. I know that Scala's `WorkspaceShardStates.withName` is case-insensitive but this will bug me looking at the db rows :)

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
davidangb: I also think the exception case is unlikely, but this exit handler looks correct to me. Since we'll never actually hit the exit handler I don't mind leaving it in!

davidangb (approved):
especially because you told me you tested this a few times against dev clones, I'm :+1:. There's one nit inline about "Sharded" vs "sharded" but I'll proactively approve because that's such a small change!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml
MatthewBemis: Oh ew. We have `unsharded` and `Sharded` in live envs right now too so that's slightly annoying. I'll at least make the sharded state consistent. More motivation to finish the migration so we can drop the column entirely :p

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_populate_sharded_entity_tables.xml
MatthewBemis: Yup that's exactly it!

MatthewBemis:
jenkins retest

(seems to be stuck)

PR #1563, opened 2021-10-27
improve error messages from Sam [QA-1500]

davidangb:
found during QA-1500, does not complete that ticket. QA-1500 had some pain to debug because Rawls was obscuring the actual error message fro mSam.

For any calls from Rawls -> Sam that use `doSuccessOrFailureRequest`, improve the error messages in cases where Sam does not return an `ErrorReport`.

Previous to this PR, those failures would result in an error message similar to
```
Sam call to HttpMethod(POST) /api/users/v1/invite/fake-email failed with error HttpEntity.Strict(application/octet-stream,18 bytes total)
```
which wasn't helpful. Now they will say
```
Sam call to HttpMethod(POST) /api/users/v1/invite/fake-email failed with error 'the-actual-error-from-Sam'
```

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -72,7 +72,7 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
                     case Success(stringErr) => stringErr
                     case Failure(_) => response.entity.toString
                   }
-                  throw new RawlsExceptionWithErrorReport(ErrorReport(f, s"Sam call to ${request.method} ${request.uri.path} failed with error $stringErrMsg"))
+                  throw new RawlsExceptionWithErrorReport(ErrorReport(f, s"Sam call to ${request.method} ${request.uri.path} failed with error '$stringErrMsg'"))
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -61,12 +61,19 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
...
               case Success(err) =>
                 logger.error(s"Sam call to ${request.method} ${request.uri.path} failed with error $err")
                 throw new RawlsExceptionWithErrorReport(err)
               case Failure(_) =>
-                throw new RawlsExceptionWithErrorReport(ErrorReport(f, s"Sam call to ${request.method} ${request.uri.path} failed with error ${response.entity}"))
+                // attempt to extract something useful from the response entity, even though it's not an ErrorReport
+                toFutureTry(Unmarshal(response.entity).to[String]) map { maybeString =>
+                  val stringErrMsg = maybeString match {
+                    case Success(stringErr) => stringErr
+                    case Failure(_) => response.entity.toString
```
davidangb, line 73: the only cases I can think of where we can't unmarshal the entity to a String involve chunked/streamed (i.e. non-strict) entities; in that case we should just bail and not wait for the entity to materialize.

MatthewBemis (approved):

ungwudik (approved):

PR #1567, opened 2021-11-02
opportunistically update entity stats cache [AS-1003]

davidangb:
If a workspace's cache is out of date, and a user requests entity metadata, then save that metadata to the cache once it's been calculated!

This aims to address workspaces whose cache is invalid. If a user has waited through the metadata calculation then let's not just throw away that metadata.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -83,6 +83,7 @@ trait EntityComponent {
     with WorkspaceComponent
     with AttributeComponent
     with EntityTypeStatisticsComponent
+    with EntityCacheComponent
     with EntityAttributeStatisticsComponent =>
 
   object entityQueryWithInlineAttributes extends TableQuery(new EntityTableWithInlineAttributes(_)) {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1014,4 +1015,28 @@ trait EntityComponent {
...
+        _ <- entityTypeStatisticsQuery.batchInsert(workspaceId, entityTypesWithCounts)
+        //update entity attribute statistics
+        _ <- entityAttributeStatisticsQuery.deleteAllForWorkspace(workspaceId)
+        _ <- entityAttributeStatisticsQuery.batchInsert(workspaceId, entityTypesWithAttrNames)
+        //update cache update date
+        numCachesUpdated <- entityCacheQuery.updateCacheLastUpdated(workspaceId, timestamp)
+      } yield numCachesUpdated
+    }
+
+  }
```
davidangb, line 1041: this logic was extracted from `EntityStatisticsCacheMonitor`

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -65,7 +66,17 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+                  val saveCacheAction = if (cacheEnabled && !isEntityCacheCurrent) {
+                    // if the entity cache is not current, AND we have the metadata result, save it to the cache!
+                    // the user has done us the favor of waiting for the result, let's take advantage of that result.
+                    // if saving the cache here fails, ignore the failure and still get the metadata to the user
+                    opportunisticSaveEntityCache(metadata, dataAccess)
+                  } else {
+                    DBIO.successful(())
+                  }
+                  saveCacheAction.map(_ => metadata)
+                }
```
davidangb, line 79: I considered moving this to a fire-and-forget operation. I decided against that because it adds code and testing complexity. This extra call will only happen when the cache is out of date, so it shouldn't be all that frequent.

MatthewBemis (approved):
i love this change

ansingh7115 (approved):

PR #1568, opened 2021-11-02
specify timeout for cache monitor calculating attribute names [AS-1003]

davidangb:
Changes in this PR:
* give `EntityComponent.getAttrNamesAndEntityTypes` the ability to time out its SQL statement, plus an argument to control the timeout length. This defaults to "0" which means no timeout. See `java.sql.Statement.setQueryTimeout` for implementation info.
* in `EntityStatisticsCacheMonitor`, allocate 80% of the per-workspace actor timeout to be spent in `EntityComponent.getAttrNamesAndEntityTypes`.

The end result of these changes is that - hopefully - when `EntityStatisticsCacheMonitor` is processing a workspace, it will hit a SQL timeout exception before the actor times out. In turn, this will mark the workspace as un-updatable. If the SQL statement didn't have a timeout, then when the actor times out we never mark the workspace as invalid.

This still has a case where if > 20% of the per-workspace timeout is already spent before  `EntityComponent.getAttrNamesAndEntityTypes` fires, the SQL timeout will be too long. Handling this case feels like a complexity:benefit loss, so I'm ignoring it.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
@@ -92,7 +92,8 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
   }
 
   def updateStatisticsCache(workspaceId: UUID, timestamp: Timestamp): Future[Unit] = {
-    // allow 80% of the per-workspace timeout to be spent calculating the attribute names
+    // allow 80% of the per-workspace timeout to be spent calculating the attribute names.
+    // note that other statements do not have timeouts and are unbounded.
     val attrNamesTimeout = (timeoutPerWorkspace * .8).toSeconds.toInt
 
     val updateFuture = dataSource.inTransaction { dataAccess =>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
@@ -58,6 +58,7 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
   val dataSource: SlickDataSource
   val standardPollInterval: FiniteDuration
   val workspaceCooldown: FiniteDuration
+  val timeoutPerWorkspace: Duration
 
   def sweep(): Future[EntityStatisticsCacheMessage] = {
     trace("EntityStatisticsCacheMonitor.sweep") { rootSpan =>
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -370,6 +372,27 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+        runAndWait(entityQuery.save(testData.workspace, entity))
+      }
+
+      // now attempt to calculate attr names and types, with a timeout of 1 second
+      intercept[MySQLTimeoutException] {
+        runAndWait(entityQuery.getAttrNamesAndEntityTypes(context.workspaceIdAsUUID, WorkspaceShardStates.Sharded, 1))
+      }
+    }
+  }
+
```
davidangb, line 403: I'm really struggling to write a stable unit test here. The goal is to test that the `entityQuery.getAttrNamesAndEntityTypes` SQL query properly times out and returns a timeout exception.

As written above, the test relies on stuffing a whole lot of data into the db and thus causing the associated SQL query to be slow. Of course, if this test happens to run on a fast computer, it'll pass - and that's no good. I also attempted to write a test that issues a table-locking query including a SQL `sleep`, then run the timeout assertion in parallel, hoping it would time out waiting for the lock to expire - I couldn't get that to work.

Should I keep working on the parallel-lock test implementation? Should I abandon this test entirely, since fundamentally it's testing Slick's timeout feature? The only Rawls code this really tests is that the timeout passed as argument to `getAttrNamesAndEntityTypes` is actually used.

jordansafer: Two thoughts- first, is it possible to write a query that runs forever? Maybe a full table scan is close enough that no DB will conceivably complete in a reasonable amount of time? Second, do we have a client side timeout on the request? Or is that what the Slick timeout is doing? One thought is to set the clientside timeout to be very short, to ensure the request times out for us if the DB response is too slow-

MatthewBemis: Can you specify a 0 second timeout or does that just get interpreted as "don't set a timeout"? I'm assuming the smallest unit of time supported is a second?

Failing that, I'm inclined to abandon it. IMO 100% test coverage is a mirage anyway so if it is not possible to make this work after putting an honest effort in, then I don't think it's worth it given that it would test very little of our own code

davidangb: @MatthewBemis 0 seconds does equate to unbounded, so that doesn't help. I wonder if I could try a timeout of 0.001 though … EDIT: nope, it expects an int

@jordansafer we could certainly write a query that runs for an arbitrarily long time, using SQL `sleep` … but the trick is that I'm trying to test _this particular_ query, because that's the one I've modified to have a timeout. The Slick timeout is a client-side timeout (where Rawls is the client and MySQL is the server)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -91,6 +92,9 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
   }
 
   def updateStatisticsCache(workspaceId: UUID, timestamp: Timestamp): Future[Unit] = {
+    // allow 80% of the per-workspace timeout to be spent calculating the attribute names
```
jordansafer: I think it might be worth including a comment that we aren't putting a timeout on the other 20%

jordansafer (commented):
Looks good!

jordansafer (approved):

MatthewBemis (approved):

PR #1566, opened 2021-11-02
EntityStatisticsCacheMonitor resiliency for timeouts/troublesome workspaces [AS-1003]

davidangb:
part of AS-1003, does not complete that ticket.

EntityStatisticsCacheMonitor now:
* handles `akka.actor.ReceiveTimeout` messages, by pausing operations for a cooldown period and then starting another sweep
* on each sweep, samples the next 10 workspaces that are candidates for cache updates, and updates one of them at random. This will hopefully prevent a single - or a few - problematic workspaces from blocking other cache updates.

PR #1569, opened 2021-11-05
cache monitor double processing [AS-1003]

davidangb:
followup to #1566. That other PR was incomplete - because my understanding was incomplete - in its handling of `ReceiveTimeout` messages. There are two cases of `ReceiveTimeout`:
1. The actor is hung/the query died and will never finish
2. The actor is just plain slow, it received a timeout but the query it's currently running will eventually finish

In the latter case, with #1566 alone, we can end up with the monitor sweeping twice/multiple times simultaneously. Because the `ReceiveTimeout` triggers a new sweep, AND the current long-running sweep will finish, both threads will result in new `Sweep` messages.

In this PR, we make sure to kill the current actor before starting a new sweep upon timeout.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
@@ -23,8 +23,10 @@ object EntityStatisticsCacheMonitor {
   }
 
   sealed trait EntityStatisticsCacheMessage
+  case object Start extends EntityStatisticsCacheMessage
   case object Sweep extends EntityStatisticsCacheMessage
   case object ScheduleDelayedSweep extends EntityStatisticsCacheMessage
+  case object DieAndRestart extends EntityStatisticsCacheMessage
 
   //Note: Ignored workspaces have a cacheLastUpdated timestamp of 1000ms after epoch
   val MIN_CACHE_TIME = new Timestamp(1000)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -38,16 +40,28 @@ class EntityStatisticsCacheMonitorActor(val dataSource: SlickDataSource, val tim
...
+      val newActor = context.system.actorOf(EntityStatisticsCacheMonitor.props(dataSource, timeoutPerWorkspace, standardPollInterval, workspaceCooldown))
+      if (sender != self) sender ! newActor // used by unit tests only
+      context.stop(self)
     case akka.actor.ReceiveTimeout =>
-      val pauseLength = standardPollInterval*2
+      // This actor timed out. We can't be certain if this actor's sweep exited abnormally, or if it is hung,
+      // or if it is benignly just taking a while. Because we can't be sure of its state, kill this actor
+      // and start up a new one using the same configuration. Pause before starting the new one
+      // in case this one is actually still processing.
+      val pauseLength = FiniteDuration(timeoutPerWorkspace.toSeconds, "seconds")*2
```
jordansafer: nit, using "seconds" vs TimeUnit.SECONDS

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -38,16 +40,28 @@ class EntityStatisticsCacheMonitorActor(val dataSource: SlickDataSource, val tim
 
   override def preStart(): Unit = {
     super.preStart()
-    self ! Sweep
+    self ! Start
   }
 
   override def receive = {
+    case Start => sender ! Sweep
```
jordansafer: Why do we need this additional Start state before the Sweep state?

jordansafer (commented):
Overall this looks good! One question from the previous review #1566 , when we randomly select a candidate to refresh, is it possible to set aside failing candidates for a future retry?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
davidangb: heh, good catch/good question. We don't really need it. I added it to make unit testing easier - in the unit tests, I really wanted a way to test if an Actor was responding, and the easiest way to do that was to have some case where if I sent the Actor a message, it would respond with a message. So I added the `Start` -> `Sweep` here, and use it at https://github.com/broadinstitute/rawls/pull/1569/files#diff-d88e8e2a2ea10f6bc99fc697520db131fa10a43e0591bde5357749caa0b1a2f7R216-R217.

jordansafer (approved):

MatthewBemis (approved):

PR #1572, opened 2021-11-10
upgrade to latest workbench-libs google2 [AS-1024]

davidangb:
AS-1024: blaze-client is out of date; it needs updating. it is a transitive dependency from workbench-google2. So, let's upgrade workbench-google2.

This PR also upgrades workbench-model, to keep it current alongside google2.

Finally, this adds an exclusion rule for scalacache to the google2 import. We use scalacache internal to Rawls - though it's a different version in Rawls, and we have compile problems when using the version included in google2. Google2's usage of scalacache appears completely limited to test (https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala), so it's safe to ignore.

I have verified that this branch fixes the `blaze-client` warning by running a manual scan using the [SourceClear CLI](https://help.veracode.com/r/t_sc_cli_agent) (I don't believe we have enough seats for everyone to install this)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -98,7 +98,7 @@ object Dependencies {
   val workbenchGoogle: ModuleID =       "org.broadinstitute.dsde.workbench" %% "workbench-google" % workbenchGoogleV
   val workbenchGoogleMocks: ModuleID =  "org.broadinstitute.dsde.workbench" %% "workbench-google" % workbenchGoogleV % "test" classifier "tests"
   val workbenchGoogle2: ModuleID =      "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll(excludeScalaCache)
-  val workbenchGoogle2Tests: ModuleID = "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V % "test" classifier "tests"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -20,6 +20,7 @@ object Dependencies {
 
   val excludeBouncyCastle =     ExclusionRule(organization = "org.bouncycastle", name = s"bcprov-jdk15on")
   val excludeProtobufJavalite = ExclusionRule(organization = "com.google.protobuf", name = "protobuf-javalite")
+  val excludeScalaCache =       ExclusionRule(organization = "com.github.cb372")
```
sagehen03, line 23: it looks like we're excluding "com.github.cb372" group artifacts?  If so, why?  If not what are we doing with this line along with lines 100 and 101?

davidangb: > this adds an exclusion rule for scalacache to the google2 import. We use scalacache internal to Rawls - though it's a different version in Rawls, and we have compile problems when using the version included in google2. Google2's usage of scalacache appears completely limited to test (https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala), so it's safe to ignore.

sagehen03 (approved):

MatthewBemis (approved):

PR #1579, opened 2021-11-30
improve error message for inconsistent attribute types [SUP-549]

davidangb:
related to SUP-549 but does not complete that ticket.

This PR improves the error message that results when something tries to insert/update an entity, and that entity contains an attribute value list, and that list contains disparate data types.

old error message:

> inconsistent attributes for list: List(AttributeNumber(9), AttributeString(oops), AttributeNumber(7), AttributeNumber(6))

new error message:

> inconsistent attributes for list: attribute lists must consist of a single data type. For attribute 'test', found types: [AttributeNumber, AttributeString]. Sample values for these types: [AttributeNumber(9), AttributeString(oops)]

(I left the "inconsistent attributes for list" verbiage in place to preserve history - you can search e.g. Slack for previous instances of this error)

Note that this PR *only* changes the error message. If this error condition occurs because a workflow is trying to write outputs, I have made no changes to how that workflow will retry forever.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -567,10 +567,16 @@ trait AttributeComponent {
       if(!attributes.isEmpty) {
         val headAttribute = attributes.head
         if (!attributes.forall(_.getClass == headAttribute.getClass)) {
-          // generate a helpful error message
+          // determine the different attribute types
           val typeNames = attributes.map(_.getClass.getSimpleName).distinct
+          // generate example values for those types
+          val exampleValues = typeNames.flatMap { typeName =>
+            attributes.find(x => x.getClass.getSimpleName == typeName).map(_.toString)
+          }
+          // generate a helpful error message
           val errMsg = s"inconsistent attributes for list: attribute lists must consist of a single data type. For attribute " +
-            s"'${toDelimitedName(attributeName)}', found types: [${typeNames.mkString(", ")}]"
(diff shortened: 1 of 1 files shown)


davidangb:
@aednichols could you take a look, or find someone on your team to look at this? It's small and doesn't address the bigger problem but is an incremental improvement.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -562,11 +563,21 @@ trait AttributeComponent {
...
+          // determine the different attribute types
+          val typeNames = attributes.map(_.getClass.getSimpleName).distinct
+          // generate example values for those types
+          val exampleValues = typeNames.flatMap { typeName =>
+            attributes.find(x => x.getClass.getSimpleName == typeName).map(_.toString)
+          }
+          // generate a helpful error message
+          val errMsg = s"inconsistent attributes for list: attribute lists must consist of a single data type. For attribute " +
+            s"'${toDelimitedName(attributeName)}', found types: [${typeNames.mkString(", ")}]. " +
+            s"Sample values for these types: [${exampleValues.mkString(", ")}]"
```
aednichols: Users have been known to place very long lists of `gs://` paths in an `Array`, might want to set a 100-char limit on the length of each element in the examples list.

( did not attempt compiling )
```suggestion
            s"Sample values for these types: [${exampleValues.map(_.take(100)).mkString(", ")}]"
```

aednichols (approved):

breilly2 (approved):

aednichols (approved):
TYVM!

PR #1583, opened 2021-12-07
Finite retries for workflow termination errors [BW-694]

aednichols:
Tested with a specially crafted WDL [0] on dev, with `out` assigned to `this.array_with_null` in the UI.

**Before**

Workflow `Succeeded`, submission stuck in `Submitted`, infinitely recurring `inconsistent attributes for list` errors in logs

**After**

Workflow `Failed` with `inconsistent attributes for list` message in UI, submission `Done`, no recurring log messages

Screen Shot 2021-12-07 at 3 20 46 PM

**Additional effect**

Workflow errors that are not explicitly converted to the non-retried `RawlsFatalExceptionWithErrorReport` form will only be retried 10 times per Rawls JVM lifetime, down from infinite retries with the previous implementation. This does still mean we will get 10 retries every time we deploy but it's a lot better than infinite and we can't fix everything at once.

[0]
```
version development

workflow null_item_in_output_array {

  input {
    Int? int_in = None
  }

  output {
    Array[String?] out = ["asdf", "wasd", "qwerty", int_in]
  }

}
```

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -296,7 +296,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }
 
     //all workflow records in this status response list
-    val workflowsWithStatuses = response.statusResponse.collect {
+    val allWorkflows = response.statusResponse.collect {
       case Success(Some((aWorkflow, _))) => aWorkflow
     }
 
@@ -304,6 +304,16 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     val workflowsWithOutputs = response.statusResponse.collect {
       case Success(Some((workflowRec, Some(outputs)))) =>
         (workflowRec, outputs)
+    }
+
+    def markWorkflowsFailed(workflows: Seq[WorkflowRecord], fatal: RawlsFatalExceptionWithErrorReport) = {
+      logger.error(s"Marking ${workflows.length} workflows as failed handling outputs in $submissionId with user-visible reason ${fatal.toString})")
+      datasource.inTransaction { dataAccess =>
+        DBIO.sequence(workflows map { workflowRecord =>
+          dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed) andThen
+            dataAccess.workflowQuery.saveMessages(Seq(AttributeString(fatal.toString)), workflowRecord.id)
+        })
+      }
     }
 
     // Attach the outputs in a txn of their own.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -296,7 +296,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }
 
     //all workflow records in this status response list
-    val workflowsWithStatuses = response.statusResponse.collect {
+    val allWorkflows = response.statusResponse.collect {
```
cahrens, line 299: You renamed the variable, but am I correct that this is still marking the same number of workflows as having failed as it did before?

aednichols: Yep, that's the intention!

davidangb (approved):
lgtm, thanks for pushing this through! This feels like a big win.

Is there a good way (even manual) to test the "stop monitoring workflow after 10 attempts" feature?

aednichols:
> Is there a good way (even manual) to test the "stop monitoring workflow after 10 attempts" feature?

I know a way, there are metrics on submission monitor restarts in HostedGraphite that currently climb into the 10s of thousands, I can deploy on dev and make sure they only reach 10

cahrens (approved):
Code LGTM, I have not tested it.

aednichols:
Discovered an issue testing on dev:

1. The supervisor [restarts](https://github.com/broadinstitute/rawls/pull/1583/files#diff-e1e653eb2e6bd59fb9e08aed18c81fd74d2c86c5856f56a7fa460073eda14b0dR212) the submission monitor actor 10 times as expected
2. Once retries end, no new actor starts for that submission
3. The set of monitored submissions is calculated by [examining the supervisor's children](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L174)
4. The submission we just gave up on is absent from the children but present in the DB, so it [appears in `unmonitoredSubmissions`](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L177-L178) as if monitoring had never started
5. We [spin up a new actor for it](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L180-L182) with a new complement of 10 retries

The clue was in the logs where the in the `error monitoring submission [...] after $count times` message, `$count` repeatedly cycles from 1 to 10 and resets.

I've got a couple ideas on how to fix this that I'll try out tomorrow.

PR #1584, opened 2021-12-08
[CA-1121]  Workspace Migration: Claim and Configure a Fresh Google Project for Workspace

ehigham:
Expose functionality to the v1 migration monitor so that it can claim and configure fresh Google Projects for workspaces.
Add google project id, number and configured timestamp to the database + extend serialisation boiler plate. Fun.
Tested by re-using the `WorkspaceApiSpec` and exposing a few methods down the inheritance tree.
Please can we stop using inheritance - it's horrible. 

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
@@ -8,7 +8,10 @@
             <column name="NEW_GOOGLE_PROJECT_ID" type="VARCHAR(254)">
                 <constraints nullable="true"/>
             </column>
-            <column name="NEW_GOOGLE_PROJECT_SETUP" type="DATETIME">
+            <column name="NEW_GOOGLE_PROJECT_NUMBER" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="NEW_GOOGLE_PROJECT_CONFIGURED" type="DATETIME">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="ci-mob" id="create_V1_MIGRATION_GOOGLE_PROJECT">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="NEW_GOOGLE_PROJECT_ID" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="NEW_GOOGLE_PROJECT_SETUP" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
@@ -6,10 +6,12 @@ import cats.effect.IO
 import cats.implicits.{catsSyntaxOptionId, toTraverseOps}
 import com.google.cloud.storage.Storage.BucketGetOption
 import org.apache.commons.lang3.SerializationException
+import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{ReadAction, ReadWriteAction, WriteAction}
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, Workspace}
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, RawlsBillingProject, Workspace}
 import org.broadinstitute.dsde.rawls.monitor.MigrationOutcome.{Failure, Success}
+import org.broadinstitute.dsde.rawls.workspace.WorkspaceService
 import org.broadinstitute.dsde.workbench.google2.GoogleStorageService
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
 
@@ -18,6 +20,7 @@ import java.time.LocalDateTime
 import java.util.UUID
 import scala.collection.JavaConversions._
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
 
 
 sealed trait MigrationOutcome
@@ -180,7 +183,7 @@ object V1WorkspaceMigrationMonitor
                              workspace: Workspace,
                              destGoogleProject: GoogleProject,
                              googleStorageService: GoogleStorageService[IO]
-                            ): IO[(GcsBucketName, ReadWriteAction[Unit])] = {
+                            ): IO[(GcsBucketName, WriteAction[Unit])] = {
     val tmpBucketName = GcsBucketName(
       "terra-workspace-migration-" + UUID.randomUUID.toString.replace("-", "")
     )
@@ -197,6 +200,25 @@ object V1WorkspaceMigrationMonitor
         .filter(_.id === migrationAttempt.id)
         .map(r => (r.tmpBucket, r.tmpBucketCreated))
         .update((tmpBucketName.value.some, Timestamp.valueOf(LocalDateTime.now).some))
+    ))
+  }
+
+  final def setupGoogleProject(migrationAttempt: V1WorkspaceMigrationAttempt,
+                               workspaceService: WorkspaceService,
+                               workspace: Workspace,
+                               billingProject: RawlsBillingProject
+                              ): Future[(GoogleProjectId, WriteAction[Unit])] = {
+    for {
+      (googleProjectId, _) <- workspaceService.setupGoogleProject(
+        billingProject,
+        workspace.currentBillingAccountOnGoogleProject.getOrElse(throw new RawlsException(s"""No billing account for workspace '${workspace.workspaceId}'""")),
+        workspace.workspaceId,
+        workspace.toWorkspaceName
+      )
+    } yield (googleProjectId, DBIO.seq(
+      migrations.filter(_.id === migrationAttempt.id)
+        .map(r => (r.newGoogleProjectId, r.newGoogleProjectSetup))
+        .update((googleProjectId.value.some, Timestamp.valueOf(LocalDateTime.now).some))
     ))
   }
 }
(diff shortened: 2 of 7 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
```diff
@@ -0,0 +1,19 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="ci-mob" id="create_V1_MIGRATION_GOOGLE_PROJECT">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="NEW_GOOGLE_PROJECT_ID" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="NEW_GOOGLE_PROJECT_NUMBER" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="NEW_GOOGLE_PROJECT_CONFIGURED" type="DATETIME">
```
gpolumbo-broad, line 14: My preference is something like:

```suggestion
            <column name="NEW_GOOGLE_PROJECT_CONFIGURED_DATE" type="DATETIME">
```

…which would of course trickle down into the code too.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
```diff
@@ -197,6 +213,24 @@ object V1WorkspaceMigrationMonitor
...
     ))
   }
+
+  final def claimAndConfigureNewGoogleProject(migrationAttempt: V1WorkspaceMigrationAttempt,
+                                              workspaceService: WorkspaceService,
+                                              workspace: Workspace,
+                                              billingProject: RawlsBillingProject
+                                             ): Future[(GoogleProjectId, GoogleProjectNumber, WriteAction[Unit])] =
+    for {
+      (googleProjectId, googleProjectNumber) <- workspaceService.setupGoogleProject(
```
marctalbott: There's a few more things in the ticket that aren't done by `setupGoogleProject`. If we don't want to do those things for this ticket, then we need to move them to an existing ticket or create a new one to track that work.

* Add the Google project id and the Google project number to the DB for this workspace record [in the `WORKSPACE` table] (I expect this just needs to be moved to a different ticket since we're saving the project id and project number in our migration history table)
* Add workspace's Google project to a service perimeter if workspace's Terra billing project is in a service perimeter
* Create Sam `google-project` resource with existing `workspace` resource as the parent

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2280,8 +2283,12 @@ class WorkspaceService(protected val userInfo: UserInfo,
                   _ <- createWorkflowCollectionFuture
                   _ <- syncPoliciesFuture
                 } yield()})
-              (googleProjectId, googleProjectNumber) <- traceDBIOWithParent("setupGoogleProject", parentSpan)(_ => DBIO.from(
-                setupGoogleProject(billingProject, billingAccount, workspaceId, workspaceName, policyEmailsByName, billingProjectOwnerPolicyEmail, parentSpan)))
+              (googleProjectId, googleProjectNumber) <- traceDBIOWithParent("setupGoogleProject", parentSpan)(span => DBIO.from(
+                for {
```
marctalbott: I'd prefer to avoid nested for loops if we can

marctalbott: Although now that I say that I see one _right_ above... My preference is still to avoid them, but I'll leave it up to you

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -79,7 +79,7 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
   // TODO: can we be any more targeted about inTransaction vs. inTransactionWithAttrTempTable here?
   // this is used by many tests to set up fixture data, which includes saving entities, therefore it needs
   // the temp table.
-  protected def runAndWait[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect], duration: Duration = 1 minutes): R = {
+  def runAndWait[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect], duration: Duration = 1 minutes): R = {
```
marctalbott, line 82: why is this not protected anymore?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
```diff
@@ -197,6 +213,24 @@ object V1WorkspaceMigrationMonitor
         .update((tmpBucketName.value.some, Timestamp.valueOf(LocalDateTime.now).some))
     ))
   }
+
+  final def claimAndConfigureNewGoogleProject(migrationAttempt: V1WorkspaceMigrationAttempt,
+                                              workspaceService: WorkspaceService,
+                                              workspace: Workspace,
+                                              billingProject: RawlsBillingProject
+                                             ): Future[(GoogleProjectId, GoogleProjectNumber, WriteAction[Unit])] =
```
marctalbott: Can we stick to either `IO` or `Future` in this class?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
```diff
@@ -82,4 +84,37 @@ class V1WorkspaceMigrationMonitorSpec
     }
   }
 
+  "claimAndConfigureGoogleProject" should "return a valid database operation" in {
+    val spec = new WorkspaceServiceSpec()
```
marctalbott, line 87: Instantiating a different test spec here doesn't feel quite right to me. Are we doing this because we need a `WorkspaceService` to test with?

marctalbott (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
ehigham: Because I need to use it in a test outside of the inheritance hierachy (hence my plea to stop using inheritance)

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
ehigham: Agreed, but I need the mocked workspace service and I really didn't want refactor this hierarchy in this change.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
```diff
@@ -82,4 +84,37 @@ class V1WorkspaceMigrationMonitorSpec
...
+              attempt,
+              services.workspaceService,
+              spec.testData.v1Workspace,
+              spec.testData.billingProject
+            )
+          },
+        5.seconds // really should be instant if we've mocked correctly?
+      )
+
+      spec.runAndWait(dbOp) shouldBe ()
```
gpolumbo-broad, line 115: Is there a reason why you decided not to write an assertion on the migration record that gets updated as part of `claimAndConfigureNewGoogleProject`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
ehigham: Yeah thanks. We won't be updating the workspace record until the very last moment. I thought we'd update the record and sam once the migration was done. Service perimeters are covered by a separate ticket

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
ehigham: Sadly this isn't possible owing to IO and Future not implementing the same set of type classes. I'll coerce `Future` into `IO` so there's some consistent return types at least

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211206_v1_migration_google_project.xml
ehigham: This is tricky because SQL `DATETIME` does not translate to Java's `DateTime` (instead it's a `Timestamp`) and therefore adding a type related suffix doesn't really translate. I don't particually like `DATE` either because it's not a date.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
ehigham: I dind't really want to expose database internals outside of the v1 workspace monitor. I certanly could but didn't think it was worth it at the time. I'll add it.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
```diff
@@ -94,26 +93,42 @@ class V1WorkspaceMigrationMonitorSpec
...
+            .map(r => (r.newGoogleProjectId, r.newGoogleProjectNumber, r.newGoogleProjectConfigured))
+            .result
+        }
+    })
+      .map(_.head)
+      .unsafeRunSync
+
+      projectId should be ('defined)
+      projectNumber should be ('defined)
+      projectConfigured should be ('defined)
```
gpolumbo-broad: I think this would work too:

```suggestion
      projectConfigured shouldBe defined
```

gpolumbo-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
ehigham: Nice! Thanks

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
ehigham: I tried to refactor this and nearly died. It's a lot of work for very little benefit.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitor.scala
marctalbott: The security perimeter ticket seems to be more focused on figuring out when the perimeter update has actually finished propagating. The service perimeter update action can still be added now IMO. If we don't want to add it now, then we should note in the security perimeter ticket that we also need to do the update.

Is there a ticket for updating the record and Sam when the migration is done? I don't want those steps to be lost in the shuffle.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/V1WorkspaceMigrationMonitorSpec.scala
marctalbott: Oof, I thought `TestApiService` and its related fixtures were in `TestDriverComponent`, not `WorkspaceServiceSpec`. Still seems like we must get around the need for a mocked workspace service in other tests, but I wouldn't be surprised at all if we didn't and this refactor isn't worth dying over

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
marctalbott: Can't say I've heard this plea, where can I hear more?

marctalbott (approved):

PR #1585, opened 2021-12-09
[CA-1628] Create Storage Transfer Job code

s-rubenstein:
RR: https://broadworkbench.atlassian.net/browse/CA-1628

In this change, @s-rubenstein and I have added functionality to create storage transfer servive jobs to transfer the contents of one workspace bucket to another.
To do this, we've added a new table called "PPW_STORAGE_TRANSFER_SERVICE_JOB" that tracks STS jobs related to workspace migration attempts. The idea is we create a job and write into the database and then poll sts for updates on each job (we could write a bespoke actor for this). When the job is done, we move to the next stage of migration.

This change seems a little large and most of the noise has been generated from moving things around. In this change:
- I decided that the "WorkspaceMigrationMonitor" was getting too large and moved the database-related types and functions into their own files, private to the newly created "migration" package. 
- I dropped the "V1" prefix because it's unlikely that we'll still be migrating v1 workspaces by the time v3 comes along
- I eliminated the use of behavioural inheritance in these types and hid the specifics of table "truncation" in the monitor. Inheritance is evil. Inheritance must die.

RR: https://broadworkbench.atlassian.net/browse/CA-1629
Add new timestamp to database to record when this occurs.
Add some new extension methods to better support chaining via the `.`
operator.

RR: https://broadworkbench.atlassian.net/browse/CA-1632
Add timestamp to migration table to record when the bucket was deleted.
Some minor fixes, including ensuring that buckets are deleted
recursively.

RR: https://broadworkbench.atlassian.net/browse/CA-1631
Re-use most of Sky's code for transferring to the temp bucket.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/Shared.scala
```diff
@@ -0,0 +1,33 @@
+package org.broadinstitute.dsde.rawls.monitor.migration
+
+import cats.implicits.{catsSyntaxOptionId, toTraverseOps}
+import org.broadinstitute.dsde.rawls.RawlsException
+
+object Shared {
```
marctalbott: rename to something more informative/specific. Maybe `MigrationHelpers`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -298,22 +174,44 @@ object V1WorkspaceMigrationMonitor
...
+  final def failNoTmpBucket[A](attempt: WorkspaceMigration): IO[A] =
     IO.raiseError(new IllegalStateException(s"""Temporary storage bucket was not created for migration "${attempt.id}.""""))
 
+
+  final def startBucketStorageTransferJob(migration: WorkspaceMigration,
+                                          originBucket: GcsBucketName,
+                                          destBucket: GcsBucketName,
+                                          projectToBill: GoogleProject,
+                                          sts: GoogleStorageTransferService[IO]
+                                          ): IO[(TransferJob, WriteAction[Unit])] =
```
marctalbott, line 271: I'm starting to think that this return type is not what we want. Returning the `DBIOAction` separately means we aren't guaranteed to write the result of the STS operation (or any other GCP operations) to the DB immediately after they complete -- it leaves it up to whoever is calling this fn to be diligent about running the `DBIOAction` right after and leaves open the possibility that they won't run it at all which we definitely don't want. The alternative is to wrap all calls to the `MigrationMonitor` in `inTransaction` blocks, but that's not ideal either since a lot of those calls involve issuing requests to GCP which would tie up DB connections. 

My proposed solution would be to run the `DBIOAction` in this function and have the return type just be `IO[TransferJob]`.

ehigham: I understand the concern but I worry that'll lead to mixing/nesting IO and database transactions and will cause all sorts of pain. At least this way it's harder to do that. While it maybe be up to the caller to get it right, we'll be the ones writing that code so it'll be fairly straightforward to prevent (ignoring return values usually indicates smell). I'll think about it, but I'm not going to do this just yet.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/Shared.scala
ehigham: Sure

s-rubenstein (commented):
Thumbs up for all Edmund's changes on top of mine! I am PR author so can't approve myself.

ehigham (approved):
Approval on behalf of @s-rubenstein

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala
```diff
@@ -0,0 +1,413 @@
...
+
+      // We need a temp bucket to transfer the workspace bucket contents into
+      val migration = runAndWait(
+        WorkspaceMigrationMonitor.workspaceMigrations
+          .filter(_.workspaceId === v1Workspace.workspaceIdAsUUID).result
+      )
+        .head
+        .copy(
+          newGoogleProjectId = GoogleProjectId(destProject).some,
+          tmpBucketName = GcsBucketName(sourceBucket).some
```
gpolumbo-broad: This might seem trivial, but can we just use another bucket as the `tmpBucket` here?  I know it technically works for the test but in reviewing it, I had to stop and try to figure out if there was a good reason why it was using `sourceBucket` and it doesn't seem like there is.

```suggestion
          tmpBucketName = GcsBucketName("foo-bar").some
```

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala
```diff
@@ -0,0 +1,413 @@
...
+      )
+
+      workspaceBucketDeleted shouldBe defined
+    }
+  }
+
+  "deleteTemporaryBucket" should "delete the temporary bucket and record when it was deleted" in {
+    val sourceProject = "general-dev-billing-account"
+    val sourceBucket = "az-leotest"
+    val destProject = "terra-dev-7af423b8"
```
gpolumbo-broad: I think the above "real" values are artifacts from when were were actually testing this thing against real buckets.  Now that this test is a proper unit test again, can we change these values to ones that are clearly "test" values?  I know it technically works, but I don't want future maintainers to come in here and think this test needs very specific values or projects to be used here.

gpolumbo-broad (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala
ehigham: Oh perhaps it shouldnt be...

ehigham: I've removed all the redefinitions - they weren't needed

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala
```diff
@@ -0,0 +1,389 @@
...
+      transferJobs.head.jobName.value shouldBe job.getName
+      transferJobs.head.originBucket.value shouldBe migration.tmpBucketName.get.value
+      transferJobs.head.destBucket.value shouldBe minimalTestData.v1Workspace.bucketName
+    }
+  }
+
+  "createFinalBucket" should "create a new bucket in the same region as the tmp workspace bucket" ignore {
+    val sourceProject = "general-dev-billing-account"
+    val sourceBucket = "az-leotest"
+    val destProject = "terra-dev-7af423b8"
```
gpolumbo-broad, line 330: Similar to the other change, can we remove these "real" values?  There's another occurrence around line 111 too.

gpolumbo-broad (changes requested):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala
ehigham: we actually use these when we run the test manually so not yet

gpolumbo-broad (approved):

PR #1590, opened 2021-12-16
DO NOT MERGE: parallel sharding migration [AJ-21]

davidangb:
DO NOT RUN THIS BRANCH WITHOUT UNDERSTANDING IT

DO NOT MERGE THIS BRANCH. It is intended to be run manually, not merged into the main codestream. I am leaving this PR as draft to prevent merging.

This branch migrates entity attribute data for all unsharded workspaces to shards. It migrates N shards in parallel via a Scala ExecutionContext. We do this parallel migration because it is significantly faster than a liquibase migration; liquibase is purely serial. See the parent ticket AJ-21 for context.

The settings (number of threads, order of shards) is configured for the dev environment. Next up is testing against a prod clone, after which some settings may change but I do not expect (I hope!) that the code will not change in any significant way.

There is a lot of `Await.result` and blocking in this PR. I am prioritizing easy-to-understand code over concurrency and thread-optimization; this branch should only be run by a single developer, and the code in question runs before Rawls even opens itself up for requests. Since we are not merging this code I am ok with "bad practice" patterns like `Await`.

If you want to see what a migration looks like using this branch, here is an example rawls log:
[migration2.log](https://github.com/broadinstitute/rawls/files/7842413/migration2.log)

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
@@ -4,11 +4,11 @@ import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.dataaccess.SlickDataSource
 import slick.jdbc.TransactionIsolation
 
-import java.util.concurrent.Executors
+import java.util.concurrent.{ConcurrentHashMap, Executors}
 import java.util.concurrent.atomic.AtomicInteger
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutorService, Future}
-
+import scala.collection.JavaConverters._
 
 class ParallelShardingMigration(slickDataSource: SlickDataSource) extends LazyLogging {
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
```diff
@@ -0,0 +1,243 @@
...
+
+  // this shenanigans around futures and blocking makes it easy to use the Future.traverse above
+  private def migrateShard(shardId: String, rowsToMigrate: Int): Future[Int] = {
+    Future(migrateShardImpl(shardId, rowsToMigrate))
+  }
+
+  private def migrateShardImpl(shardId: String, rowsToMigrate: Int): Int = {
+    val tick = System.currentTimeMillis()
+
+    shardsStarted.incrementAndGet()
```
jordansafer: Let's store the results of shardsStarted.incrementAndGet() for logging here and below to avoid have a confusing log from the increment-

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
```diff
@@ -0,0 +1,243 @@
...
+              AND w.shard_state = 'unsharded';""".as[Int]
+    runSql(rowsToMigrateSql.head)
+  }
+
+  private def runSql[T](sql: ReadWriteAction[T]) = {
+    // Serializable here ensures the before/after row counts are accurate; these really help
+    // with verifying success of shard copies.
+    // Since we intend this to run while Rawls as a whole is down, and no other db activity,
+    // it really shouldn't matter
+    val transaction = sql
```
jordansafer, line 214: Do we need any exception handling or retries on the sql queries? Or is this handled in the db facade? I'm thinking that a connection failure might be one possible source of a migration error

jordansafer (approved):
Looks great!

sagehen03 (approved):
Let's migrate! 🎲

sagehen03:
> The settings (number of threads, order of shards) is configured for the dev environment. Next up is testing against a prod clone, after which some settings may change but I do not expect (I hope!) that the code will not change in any significant way.

I think in prod I got peak efficiency around 20 concurrent migrations.  Going up to 24 migrations was about 15 minutes slower than 20.  (2:17 vs. 2:30). [slack](https://app.slack.com/client/T0CMFS7GX/G01UE377KJL/thread/G01UE377KJL-1633010421.187900)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
```diff
@@ -0,0 +1,244 @@
...
+
+    // now, process the shards in *shardId* alphabetical order. This
+    // means the large and small row counts should be pretty evenly distributed,
+    // and thus we are migrating small shards in parallel to
+    // large shards for balanced CPU utilization
+    //
+    // we could also simply use expectedCounts.toList, since that returns a list
+    // in "map order," which is based on the key hashes - also should be evenly distributed.
+    // but, alpha order is easier for us humans to read about in the logs
+    val orderedShards = expectedCounts.toList.sortBy(_._1)
```
sagehen03: I think the script will run faster if we reverse the order and start the biggest shards first.  In my testing the slowest shard took over 1.5 hours so if that shard does not get an available thread until way late in the process we'll end up with the job essentially running single-threaded for the last x minutes of execution.  In the grand scheme, it's not a big deal, but if we want to keep the most threads threads busiest for the biggest proportion of program execution, I think the better approach is to start with the biggest shards.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
davidangb: retries are not handled automatically - I will look to add them in!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
davidangb: fixed!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/ParallelShardingMigration.scala
davidangb: yeah - thread count and ordering are the two big places I'm looking to experiment against the real-scale prod clone. Starting with largest first sounds like a good idea, I will definitely evaluate that!

jordansafer (approved):

davidangb:
closing this PR, but leaving the branch in place.

PR #1598, opened 2022-01-06
remove already-migrated rows from archived table

davidangb:
do not merge yet, if we migrate all shards we don't need this. It's here as a fallback in case we need more time to roll out the all-shard migration.


davidangb:
closing; we won't need this

PR #1599, opened 2022-01-07
update shard procedures [AS-1053]

davidangb:
compared to the previous iteration of the `copyRowsToAttributeShard` stored procedure ([link](https://github.com/broadinstitute/rawls/blob/a4bdc468fad3ffb166c7e7291d01e5ba0e94ba2e/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210924_sharded_entity_tables.xml#L59-L86)), this PR changes:
1. only copy rows for workspaces that are currently unsharded; this avoids duplication for the workspaces we previously migrated
2. don't attempt to reuse the primary keys for attribute rows. These PKs are not user-facing and it doesn't matter if they change. If we try to reuse them, we run the (real) risk of id conflicts between rows we're migrating and rows already created inside shards.

I wish there was a good way to diff this PR against the previous liquibase changeset, I can't think of a way to do that in github web UI.

This change is a prereq to rolling out all workspaces to a sharded state.


sagehen03 (approved):
lgtm!

PR #1600, opened 2022-01-07
CA-1669 Do not update Billing Account information if the current value on Google does not match what Rawls thinks it should be

gpolumbo-broad:
As part of fixing PROD-625, we are going to manually update some Billing Account information in the Rawls db.  These changes will then be picked up by the `WorkspaceBillingAccountMonitor` which will actually try to call Google to set the desired Billing Accounts on the right Google Projects.  

One of the problems we caused before with the PPW release was that we overwrote a bunch of Billing Account information to the whatever value Rawls thought should be set.  However, users have historically been able to change the Billing Account on their Terra managed Google Projects.  This means that users could change the Billing Account to exactly what they want, but Rawls would not know about this change.  Similarly, Billing Account owners could disable billing on Terra managed projects and Rawls would not know about this change.

In an effort to avoid overwriting user's changes to their Google Projects Billing Account, this PR adds an additional check when Rawls is trying to update the Billing Account on a Terra managed Google Project.  This check compares the current Billing Account that is actually set on the Google Project to the Billing Account value that Rawls _thinks_ should be set on that Project.  If those values match, then Rawls can proceed with the update.  If those values don't match, then we have to assume that the user (or Billing Account owner) modified the Billing Account directly on Google and Rawls _should not_ change that value.   This check _can_ be overridden by providing a `force` flag.  This should be used with caution, but it was needed when creating new v2 Workspaces which use Google Projects allocated from RBS.  We `force` the overwrite at this time so that Rawls does not need to know the Billing Account set on RBS projects in its pool. 

Unfortunately, since the business logic for this change is buried in `HttpGoogleServicesDAO`, it is not unit testable because that class relies on real interactions with GCP.  Ideally, we would refactor that method and extract that logic someplace where it would be testable, but that effort would significantly expand the scope of this change and it sounds like we cannot afford to do that right now.

Additionally, we also modified a bunch of log statements to try and capture the `oldBillingAccount` value that is being overwritten in the `updateBillingAccount` request(s).  

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -725,7 +725,7 @@ class HttpGoogleServicesDAO(
           Option(RawlsBillingAccountName(projectBillingInfo.getBillingAccountName))
 
         // Check actual Billing Account value from google against the value that Rawls thinks it should be before the update
-        if (oldBillingAccount != currentBillingAccountFromGoogle) {
+        if (!force && (oldBillingAccount != currentBillingAccountFromGoogle)) {
           throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.PreconditionFailed,
             s"Could not update Billing Account on Google Project ID ${googleProjectId} to Billing Account ${newBillingAccount} because Billing Account in Rawls ${oldBillingAccount} did not equal current Billing Account in Google ${projectBillingInfo.getBillingAccountName}"))
         }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -688,7 +688,7 @@ class HttpGoogleServicesDAO(
     * Terra users should not have IAM permissions to change the Billing Account on their Terra managed Google Projects
     * directly on Google.  However, Billing Account Admins (who may know nothing about Terra) _can_ at least remove
     * their Billing Account from Terra managed Google Projects as a means of disabling billing.  See https://broadworkbench.atlassian.net/browse/CA-1585
-    * TODO: Refactor method.  A decent amount of business logic has crept into this method and is currently untested as a result.
+    * TODO: Refactor method: https://broadworkbench.atlassian.net/browse/CA-1673.  A decent amount of business logic has crept into this method and is currently untested as a result.
     *
     * @param googleProjectId - Google Project ID that the Billing Account will be changed on
     * @param newBillingAccount - The new value we want to set on the project
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -679,7 +679,12 @@ class HttpGoogleServicesDAO(
     }
   }
 
-  override def updateGoogleProjectBillingAccount(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): Future[ProjectBillingInfo] = {
+  // This should only work if the existing Billing Account value for the Google Project is what Terra thinks it should be
+  // If it is some other value, then we should _not_ update the Billing Account because it was updated on Google via
+  // some other means and we don't want to mess things up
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -698,13 +718,29 @@ class HttpGoogleServicesDAO(
...
       blocking {
         val projectBillingInfo = executeGoogleRequest(fetcher)
+
+        val currentBillingAccountFromGoogle = if (projectBillingInfo.getBillingAccountName == null || projectBillingInfo.getBillingAccountName.isBlank)
+          None
+        else
+          Option(RawlsBillingAccountName(projectBillingInfo.getBillingAccountName))
+
+        // Check actual Billing Account value from google against the value that Rawls thinks it should be before the update
+        if (oldBillingAccount != currentBillingAccountFromGoogle) {
```
s-rubenstein: Does this need a `&& !force`?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1312,12 +1312,13 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
     // whatever that Google Project is, we set the right Billing Account on it, which is the Billing Account specified
     // in the Billing Project
     val billingAccountNameCaptor = captor[RawlsBillingAccountName]
-    verify(services.gcsDAO).updateGoogleProjectBillingAccount(any[GoogleProjectId], Option(billingAccountNameCaptor.capture))
+    // TODO: CA-1669 See TODO from WorkspaceService.scala:1987
+    verify(services.gcsDAO).updateGoogleProjectBillingAccount(any[GoogleProjectId], Option(billingAccountNameCaptor.capture), ???)
```
s-rubenstein: Does this have an any syntax? I feel like we've tended to use that for matching more than `???`, but wanted to check.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -698,13 +718,29 @@ class HttpGoogleServicesDAO(
     retryWithRecoverWhen500orGoogleError(() => {
       blocking {
         val projectBillingInfo = executeGoogleRequest(fetcher)
+
+        val currentBillingAccountFromGoogle = if (projectBillingInfo.getBillingAccountName == null || projectBillingInfo.getBillingAccountName.isBlank)
+          None
+        else
+          Option(RawlsBillingAccountName(projectBillingInfo.getBillingAccountName))
```
ehigham, line 725: More succinctly:

```suggestion
        val currentBillingAccountFromGoogle = Option(projectBillingInfo.getBillingAccountName)
            .filter(!_.isBlank)
            .map(RawlsBillingAccountName)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountMonitor.scala
```diff
@@ -38,13 +38,13 @@ class WorkspaceBillingAccountMonitor(dataSource: SlickDataSource, gcsDAO: Google
...
-        case (googleProjectId, newBillingAccount) =>
-          IO.fromFuture(IO(updateGoogleAndDatabase(googleProjectId, newBillingAccount))).attempt.map {
+        case (googleProjectId, newBillingAccount, oldBillingAccount) =>
+          IO.fromFuture(IO(updateGoogleAndDatabase(googleProjectId, newBillingAccount, oldBillingAccount))).attempt.map {
             case Left(e) => {
               // We do not want to throw e here. traverse stops executing as soon as it encounters a Failure, but we
               // want to continue traversing the list to update the rest of the google project billing accounts even
               // if one of the update operations fails.
-              logger.warn(s"Failed to update billing account on $googleProjectId to $newBillingAccount", e)
+              logger.warn(s"Failed to update billing account from ${oldBillingAccount}to ${newBillingAccount} on project $googleProjectId", e)
```
ehigham, line 47: Missing space:

```suggestion
              logger.warn(s"Failed to update billing account from ${oldBillingAccount} to ${newBillingAccount} on project $googleProjectId", e)
```

ehigham (approved):

s-rubenstein (approved):
LGTM

PR #1603, opened 2022-01-14
[CA-1119] Migrate Workspace Bucket to New Google Project

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1119
Add functionality to incrementally advance the state of a migration via
composiong together actions that read migrations of a specified state,
perform some action, then write the updated state to the database.
Use ReaderT to inject depedencies into each step/handler.
Use OptionT to encode that there might not be a migration in the right
state to be handled by that handler.
Enable higher-kinded types, type-projection and better-monadic-for.

For anyone unfamiliar with monad transformers:

`ReaderT` and `OptionT` are what's called "Monad Transformers" - they wrap an inner monad and extend its effects with its own. We can "stack" monad transformers to build new, more capable monads with more interesting effects. Let's make this idea clear with an example that's motivated in this PR. Consider a system that fires nuclear weapons at designated targets when an order is given. In such a system, we'll read a command then fire the nuclear weapons and update the command if a command exists.

```scala
def getCommand: IO[Option[Command]] = ???

// https://youtu.be/t8xliaDUPwg?t=37
def fireTheNuclearWeapons(command: Command): IO[Outcome] = ???

def updateCommand(command: Command, outcome: Outcome): IO[Unit] = ???
```

Let's put this system to use:

```scala
def doomsdayMachine: IO[Unit] =
  for {
    commandOpt <- getCommand
    _ <- if (commandOpt.isEmpty) IO.unit else
     for {
       outcome <- fireTheNuclearWeapons(commandOpt.get)
       _ <- updateCommand(outcome, commandOpt.get)    
     } yield ()
  } yield ()
```
What's the problem? Well most of the logic in this function is nested an extra later down leading to clutter. We might need to perform more complicated logic involving optional values which would make this mess worse. Well, we could just write another function. But what about if we have other missiles that need to be fired other than nuclear ones? Then we'd just end up writing tons of functions and we'd end up with a mess.

In this particular case, we can simplify our code by noticing that our whole system conditionally executes if an order exists. This is what the `OptionT` monad transformer enables - it transforms the `IO` `bind` (or `flatMap`) operation from 
`IO[Option[A]] => (Option[A] => IO[Option[B]]) => IO[Option[B] `
 to 
`OptionT[IO, A] => (A => OptionT[IO, B]) => OptionT[IO, B]`
. What does that mean? In `IO`, the argument to `flatMap` (called a "Kleisli arrow") is always called regardless of the state of the `Option`. With `OptionT`, the arrow is only called if the optional value is Some(a). Modifying our function signatures:

```scala
def getCommand: OptionT[IO, Command] = ???

def fireTheNuclearWeapons(command: Command): OptionT[IO, Outcome] = ???

def updateCommand(command: Command, outcome: Outcome): OptionT[IO, Unit] = ???
```
Our system becomes:

```scala
def doomsdayMachine: OptionT[IO, Unit] =
  for {
    command <- getCommand
    outcome <- fireTheNuclearWeapons(command)
    _ <- updateCommand(outcome, command)    
  } yield ()
```

Beautiful.

Now let's say that our commands come from a cloud pubsub queue and `getCommand` and `updateCommand` need access to that pubsub queue. Well, we'll just pass them along as arguments:

```scala
def doomsdayMachine(messageQueue: PubSubQueue): OptionT[IO, Unit] =
  for {
    command <- getCommand(messageQueue)
    outcome <- fireTheNuclearWeapons(command)
    _ <- updateCommand(messageQueue, outcome, command)    
  } yield ()
```

Ok simple enough. But what about if we have other dependencies? Can we just pass these in too? Well yes, but then we get lots of noise from forwarding dependencies to functions, which may need to forward dependencies to the functions they call etc etc etc. Yuck. We can do one better. Enter the `ReaderT` monad transformer!!

The `ReaderT` provides a means of threading dependencies through your application without explicit handling. You can `ask` what the environment is and you can modify the environment locally (`local`) to update a dependency for an inner funciton call. `ReaderT[F[_], Env, A]` is simply a wrapper for the function `Env => F[A]`. We can "stack" the reader onto our existing monad stack to add dependency injection to our system:

```scala
def doomsdayMachine: ReaderT[OptionT[IO, *], PubSubQueue, Unit] =
  for {
    command <- getCommand
    outcome <- fireTheNuclearWeapons(command)
    _ <- updateCommand(outcome, command)    
  } yield ()
```
And we have returned to our original simple program. Monad transformers are awesome.

`ReaderT`: https://typelevel.org/cats/datatypes/kleisli.html

 `OptionT`: https://typelevel.org/cats/datatypes/optiont.html

This is all inspired by monad transformers in Haskell. If you're up to it, the following articles do a great job at explaining things:
https://wiki.haskell.org/IO_inside -- really helpful for understanding IO (and monads in general) in a pure functional language.
https://wiki.haskell.org/Monad_Transformers_Explained
https://wiki.haskell.org/All_About_Monads#The_Reader_monad

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
@@ -6,9 +6,9 @@
     <changeSet logicalFilePath="dummy" author="ehigham" id="add_UPDATED_TIMESTAMP_TO_WORKSPACE_MIGRATION">
         <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
             <column name="UPDATED"
-                    type="DATETIME"
-                    defaultValueComputed="CURRENT_TIMESTAMP"
-                    remarks="When the system scheduled migrating the WORKSPACE.">
+                    type="DATETIME(6)"
+                    defaultValueComputed="NOW(6)"
+                    remarks="When the system updated the migration attempt.">
                 <constraints nullable="false"/>
             </column>
         </addColumn>
@@ -17,26 +17,26 @@
                 BEFORE UPDATE ON V1_WORKSPACE_MIGRATION_HISTORY
                 FOR EACH ROW
             BEGIN
-                SET NEW.UPDATED = CURRENT_TIMESTAMP();
+                SET NEW.UPDATED = NOW(6);
             END ~
         </sql>
     </changeSet>
     <changeSet logicalFilePath="dummy" author="ehigham" id="add_WORKSPACE_BUCKET_TRANSFER_TIMESTAMPS">
         <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
-            <column name="WORKSPACE_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME">
+            <column name="WORKSPACE_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME(6)">
                 <constraints nullable="true"/>
             </column>
-            <column name="WORKSPACE_BUCKET_TRANSFERRED" type="DATETIME">
+            <column name="WORKSPACE_BUCKET_TRANSFERRED" type="DATETIME(6)">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
     </changeSet>
     <changeSet logicalFilePath="dummy" author="ehigham" id="add_TMP_BUCKET_TRANSFER_TIMESTAMPS">
         <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
-            <column name="TMP_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME">
+            <column name="TMP_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME(6)">
                 <constraints nullable="true"/>
             </column>
-            <column name="TMP_BUCKET_TRANSFERRED" type="DATETIME">
+            <column name="TMP_BUCKET_TRANSFERRED" type="DATETIME(6)">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="ehigham" id="add_UPDATED_TIMESTAMP_TO_WORKSPACE_MIGRATION">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="UPDATED"
+                    type="DATETIME"
+                    defaultValueComputed="CURRENT_TIMESTAMP"
+                    remarks="When the system scheduled migrating the WORKSPACE.">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <sql endDelimiter="~">
+            CREATE TRIGGER bump_migration_timestamp_on_update
+                BEFORE UPDATE ON V1_WORKSPACE_MIGRATION_HISTORY
+                FOR EACH ROW
+            BEGIN
+                SET NEW.UPDATED = CURRENT_TIMESTAMP();
+            END ~
+        </sql>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="ehigham" id="add_WORKSPACE_BUCKET_TRANSFER_TIMESTAMPS">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="WORKSPACE_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+            <column name="WORKSPACE_BUCKET_TRANSFERRED" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="ehigham" id="add_TMP_BUCKET_TRANSFER_TIMESTAMPS">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="TMP_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+            <column name="TMP_BUCKET_TRANSFERRED" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
@@ -74,14 +74,15 @@ object WorkspaceMigrationMonitor {
   type MigrateAction[T] = ReaderT[OptionT[IO, *], MigrationDeps, T]
 
   object MigrateAction {
+    // lookup a value in the environment
     final def asks[T](f: MigrationDeps => T): MigrateAction[T] =
       ReaderT.ask[OptionT[IO, *], MigrationDeps].map(f)
 
-
+    // lift an IO action into the context of a MigrateAction
     final def liftIO[A](ioa: IO[A]): MigrateAction[A] =
       ReaderT.liftF(OptionT.liftF(ioa))
 
-
+    // empty action
     final def unit: MigrateAction[Unit] =
       ReaderT.pure()
   }
@@ -95,7 +96,8 @@ object WorkspaceMigrationMonitor {
         .result
     }
 
-
+  // Read workspace migrations in various states, attempt to advance their state forward by one
+  // step and write the outcome of each step to the database.
   final def migrate: MigrateAction[Unit] =
     startMigration >>
       claimAndConfigureGoogleProject >>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
@@ -2,7 +2,8 @@ package org.broadinstitute.dsde.rawls.monitor.migration
 
 import akka.actor.typed.Behavior
 import akka.actor.typed.scaladsl.Behaviors
-import cats.data.ReaderT
+import cats.Traverse
+import cats.data.{OptionT, ReaderT}
 import cats.effect.IO
 import cats.implicits._
 import com.google.cloud.storage.Storage.BucketGetOption
@@ -10,14 +11,13 @@ import com.google.storagetransfer.v1.proto.TransferTypes.TransferJob
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, GoogleProjectNumber, RawlsBillingProject, Workspace}
-import org.broadinstitute.dsde.rawls.monitor.migration.MigrationStatus._
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, RawlsBillingProject, Workspace}
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils._
+import org.broadinstitute.dsde.rawls.monitor.migration.WorkspaceMigrationHistory._
 import org.broadinstitute.dsde.rawls.workspace.WorkspaceService
 import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.JobTransferSchedule
 import org.broadinstitute.dsde.workbench.google2.{GoogleStorageService, GoogleStorageTransferService}
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
-import slick.dbio.DBIOAction
 
 import java.sql.Timestamp
 import java.time.LocalDateTime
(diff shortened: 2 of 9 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
+      deleteWorkspaceBucket >>
+      createFinalWorkspaceBucket >>
+      issueTmpBucketTransferJob >>
+      deleteTemporaryBucket >>
+      MigrateAction.unit
+
+
+  final def startMigration: MigrateAction[Unit] =
+    for {
+      migration <- findMigration(_.started.isEmpty)
```
s-rubenstein, line 115: This works because there won't be any entry if it hasn't been scheduled, correct? TOL: since this is all asychronous, I am now wondering why we need the started field at all… is it just to see if the monitor saw the row? I am not entirely clear from a semantic standpoint what the difference between scheduled and started will fundamentally end up being.

Definitely beyond the scope of this PR though.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
+      tmpBucketName <- MigrateAction.liftIO(IO {
+        migration.tmpBucketName.getOrElse(throw noGoogleProjectError(migration))
+      })
 
-      _<- IO.raiseUnless(successOpt.contains(true)) {
-        noWorkspaceBucketError(GcsBucketName(workspace.bucketName))
+      workspace <- getWorkspace(migration.workspaceId)
+      transferJob <- startBucketTransferJob(migration, GcsBucketName(workspace.bucketName), tmpBucketName)
+      issued <- timestampNow
+      _ <- inTransaction { _ =>
```
s-rubenstein, line 204: Do we need to store a transferJob id or some sort of identifier? Or is that all findable down the road?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
-    } yield (GcsBucketName(workspace.bucketName), workspaceMigrations
-      .filter(_.id === migration.id)
-      .map(_.finalBucketCreated)
-      .update(created.some)
-      .ignore
-    )
+      workspace <- getWorkspace(migration.workspaceId)
+      transferJob <- startBucketTransferJob(migration, tmpBucketName, GcsBucketName(workspace.bucketName))
+      issued <- timestampNow
+      _ <- inTransaction { _ =>
```
s-rubenstein, line 284: Same question as above, do we need an identifier for the transfer job?

s-rubenstein (approved):
This looks very nice! I like this option a lot. A few things, but nothing I feel like I need another review on this for.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: If I understand your question correctly then yes - this will return a migration if it has been scheduled and it's not currently being processed anywhere else in the pipeline.
`started` is somewhat artificial but I wanted to include it so we could monitor when the actor started working on any particular migration. My main motivation is that it allows us to consider alternative scheduling strategies. For example, we may want to limit the number of concurrent migrations we run initially, just until we know things are more or less OK before we ramp up migration. We might want to prioritise work on certain workspaces. We might want to do something else entirely. The `started` entry is a simple flag that can be used to encode this and throttle the number of migrations that enter the pipeline at once.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: I was thinking of having another actor (managed by this one) handle managing the transfer jobs. When we want to start a job, we message that actor. That actor would then message this one when a particular job had finished with its outcome. The migration actor would then update the state of the migration to be `WorkspaceBucketTransferred`, `TmpBucketTransferred` or `Failed`. Kind of like an interrupt. This way, the migration actor doesn't need to know about managing sts jobs.

ehigham: To answer your question, yes, we can look this up.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
s-rubenstein: Right, but couldn't we just use whatever is the first step to notice that? (I guess nothing could stop us from adding steps before that one, but I just worry about us adding more complexity that we don't need)

ehigham:
Jenkins test failing with
```
org.broadinstitute.dsde.rawls.entities.datarepo.DataRepoBigQuerySanitizationSpec *** ABORTED ***
java.net.UnknownHostException: raw.githubusercontent.com
at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:220)
at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
at java.base/java.net.Socket.connect(Socket.java:609)
at java.base/sun.security.ssl.SSLSocketImpl.connect(SSLSocketImpl.java:299)
at java.base/sun.security.ssl.BaseSSLSocketImpl.connect(BaseSSLSocketImpl.java:173)
at java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:182)
at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:474)
at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:569)
at java.base/sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:266)
at java.base/sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:373)
```

This is unrelated to this PR

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
```diff
@@ -0,0 +1,44 @@
...
+                SET NEW.UPDATED = CURRENT_TIMESTAMP();
+            END ~
+        </sql>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="ehigham" id="add_WORKSPACE_BUCKET_TRANSFER_TIMESTAMPS">
+        <addColumn tableName="V1_WORKSPACE_MIGRATION_HISTORY">
+            <column name="WORKSPACE_BUCKET_TRANSFER_JOB_ISSUED" type="DATETIME">
+                <constraints nullable="true"/>
+            </column>
+            <column name="WORKSPACE_BUCKET_TRANSFERRED" type="DATETIME">
```
gpolumbo-broad: Perhaps a nit, but maybe we want this named something else like `WORKSPACE_BUCKET_TRANSFER_JOB_COMPLETE` or `WORKSPACE_BUCKET_TRANSFER_JOB_END` (if we do the latter, we might want to change the `ISSUED` date to `START` too).  Same naming convention would apply to the `TMP_BUCKET` too.

My reasoning is that I suspect the STS job can complete as either a success or a failure, and having this field named specifically for only successes may not be what we want.  If it _is_ what we want (meaning we only populate this field when the job finishes successfully, which is fine) then we may want another field or two for tracking when/why it failed?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
+                                 workspaceService: WorkspaceService,
+                                 storageService: GoogleStorageService[IO],
+                                 storageTransferService: GoogleStorageTransferService[IO])
+
 
+  type MigrateAction[T] = ReaderT[OptionT[IO, *], MigrationDeps, T]
 
-  final def timestampNow: IO[Timestamp] = IO.delay(Timestamp.valueOf(LocalDateTime.now))
+  object MigrateAction {
+    // lookup a value in the environment
```
gpolumbo-broad: What exactly does this comment mean?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: I prefer to have a step that has no real functionality to represent this rather than bundling that in with steps with migration logic.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
     for {
-      res <- IO.fromFuture(IO {
+      migration <- findMigration { m =>
+        m.started.isDefined && m.newGoogleProjectConfigured.isEmpty
+      }
+
+      workspace <- getWorkspace(migration.workspaceId)
+      currentBillingAccountOnGoogleProject <- MigrateAction.liftIO(IO {
+        workspace.currentBillingAccountOnGoogleProject.getOrElse(
+          throw new RawlsException(s"""No billing account for workspace '${workspace.workspaceId}'""")
```
gpolumbo-broad, line 135: This may need to go elsewhere, but here's my thought…

In addition to ensuring that the Workspace has its Billing Account specified, we should validate that Billing Account Values prior to migrating the workspace.  This validation should include:

1.  The Billing Account on the Workspace should match the Billing Account specified on its Billing Project (aka, Namespace)
2. The `billingAccountErrorMessage` field on this Workspace should be empty
3. The Billing Account on the the Workspace's Billing Project needs to be valid: https://github.com/broadinstitute/rawls/blob/9939bf75a0f8c044867e33928997b8b3a548fc8c/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala#L56

If any of the above checks fail, we should not progress with the migration of this Workspace.  

In terms of _where_ we do this checking, it seems reasonable to do so when first scheduling the migration.  But it also makes sense to do it here in case the migration was scheduled and then the Billing Accounts got messed up in some way?  I'm not sure the best answer here, this all only just occurred to me, so we can talk more.

gpolumbo-broad: Given the state of this PR, I'm 100% ok if we want to make these validations as their own follow-up change after this merges.  (we should just make a ticket if we decide to go that route)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
```diff
@@ -55,236 +57,392 @@ object WorkspaceMigrationMonitor {
...
-        projectToBill,
-        sts
-      )
+  final def findMigration(predicate: WorkspaceMigrationHistory => Rep[Boolean])
+  : MigrateAction[WorkspaceMigration] =
+    inTransaction { _ =>
+      workspaceMigrations
+        .filter(row => row.finished.isEmpty && predicate(row))
+        .sortBy(_.updated)
+        .take(1)
```
gpolumbo-broad, line 338: Are there any protections to prevent us from creating multiple migration rows for the same workspace?  Uniqueness constraints or something in queries to raise a fuss if we somehow have multiples?

gpolumbo-broad (commented):
Given how long this PR has been open and we're not going to be running this stuff yet, I'm ok if you merge now so long as we create Jira issues for any TODO items that arise from my comments because I don't think any of them fundamentally change what you're doing here, but they do add further checks and validations that I think are important.  In other words, they're things that can be added onto this code later.  I just don't want them falling through the cracks either, hence the desire to have tickets for them.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220112_add_updated_and_transfer_job_timestamps_to_workspace_migration.xml
ehigham: The state of these jobs is maintained in a different table. Getting more detailed information like when the job `FINISHED` or the job `Outcome` is a join on the migration id.
I'm using the `*_TRANSFERRED` timestamp as a positive case for advancing the migration. If it were just `JOB_END` I'd then need to do a case analysis on the outcome every time - this feels a bit clunky.
I wanted to avoid using `STARTED` because the transfer job doesn't necessarily start when you POST to google sts. Maybe not the most important detail, but that would be a nit for me lol

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: Ah... Maybe should explain `ReaderT` a bit more. Scala is a noisy language so I'm afraid the implicits and type variables might get in the way a bit...

Recall that `ReaderT[F, Env, A]` is a wrapper around the function, `run: Env => F[A]`, ie: 
```scala
case class ReaderT[F[_], Env, A](run: Env => F[A]) { ... }
```

By convention, the argument of the function `run` is referred to as the "environment". To `ask` a `ReaderT` is to return the environment:

```scala
def ask[F[_], Env](implicit F: Applicative[F]): ReaderT[F, Env, Env] = 
    ReaderT { env => F.pure(env) }
```

`asks` is a variation of `ask` where a selector is used to select which value to return from the environment (the argument to the reader)

```scala
def asks[F[_], Env, A](selector: Env => A)(implicit F: Applicative[F]): ReaderT[F, Env, Env] = 
    ReaderT { env => F.pure(selector(env)) }
```

I've implemented `asks` as cats doesn't seem to have it and it's often more useful than just `ask`

ehigham: Given that, howzabout  "lookup a value in the environment using `selector`"?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: Thanks for this. I'll address these comments in a follow-up PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: Yes - I've configured the table such that you can't have more than one active (ie non-finished) migration attempt (row) at a time. You CAN have as many finished attempts as you like. If you try to create concurrent migration attempts of the same workspace you'll get an error.

ehigham: https://github.com/broadinstitute/rawls/blob/b7252ec2d1826e0d28bc5378d2632c3d00315740/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20211115_v1_workspace_migration_history.xml#L56-L81

ehigham: https://github.com/broadinstitute/rawls/blob/b7252ec2d1826e0d28bc5378d2632c3d00315740/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationMonitorSpec.scala#L99-L105

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationMonitor.scala
ehigham: https://broadworkbench.atlassian.net/browse/CA-1708

ehigham: https://github.com/broadinstitute/rawls/pull/1606

gpolumbo-broad (approved):

PR #1604, opened 2022-01-17
Cromwell latency metric (BW-745).

cahrens:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
@@ -20,6 +20,7 @@ import spray.json.DefaultJsonProtocol._
 import spray.json._
 
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.postfixOps
@@ -411,6 +412,10 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
       // call to submitWorkflows returns a tuple:
       val executionServiceKey = workflowSubmitResult._1
       val executionServiceResults = workflowSubmitResult._2
+
+      // Emit metric for time until Cromwell has processed workflow.
+      val elapsedTime = System.currentTimeMillis() - submissionRec.submissionDate.getTime
+      workflowToCromwellLatency.update(elapsedTime, TimeUnit.MILLISECONDS)
 
       (executionServiceKey, workflowRecs.zip(executionServiceResults))
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
@@ -62,7 +62,6 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
   protected def workflowToCromwellLatency: Timer =
     ExpandedMetricBuilder
       .expand(WorkspaceMetricKey, "submission_to_cromwell")
-      .transient()
       .asTimer("latency")
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.rawls.metrics
 
-import nl.grons.metrics4.scala.Counter
+import nl.grons.metrics4.scala.{Counter, Timer}
 import org.broadinstitute.dsde.rawls.metrics.RawlsExpansion._
 import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
@@ -55,6 +55,15 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
       .expand(WorkflowStatusMetricKey, status)
       .transient()
       .asCounter("count")
+
+  /**
+    * A timer for capturing latency between initial Rawls submission and workflow processing in Cromwell.
+    */
+  protected def workflowToCromwellLatency: Timer =
+    ExpandedMetricBuilder
+      .expand(WorkspaceMetricKey, "submission_to_cromwell")
+      .transient()
+      .asTimer("latency")
 }
 
 object RawlsInstrumented {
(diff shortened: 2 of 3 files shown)


cahrens:
jenkins retest

cahrens:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
```diff
@@ -386,22 +386,34 @@ class SubmissionApiServiceSpec extends ApiServiceSpec with TableDrivenPropertyCh
...
+          assertResult(numSamples) {
+            submission.workflows.size
+          }
 
-        abortSubmission(services, wsName, submission.submissionId, false)
+          abortSubmission(services, wsName, submission.submissionId, false)
+        }
+      }
+    } { capturedMetrics =>
+      var counter: Int = 0
```
cahrens, line 408: This seemed to be the only existing unit test that hit the bit of code I was touching. I'm not sure if it is OK to just piggyback on the existing test instead of writing a dedicated one (I updated the description to include the metric testing), but to keep the test passing I at least need to include the "withStatsD" wrapper. And then it doesn't add any additional overhead to include the metric assert code here.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -55,6 +55,14 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
...
       .transient()
       .asCounter("count")
+
+  /**
+    * A timer for capturing latency between initial Rawls submission and workflow processing in Cromwell.
+    */
+  protected def workflowToCromwellLatency: Timer =
+    ExpandedMetricBuilder
+      .expand(WorkspaceMetricKey, "submission_to_cromwell")
+      .asTimer("latency")
```
cahrens, line 65: A timer allows us to show mean elapsed time, 95th percentile, and standard deviation. For example,

image

(the "transient" part has been removed in the code, but that hasn't been deployed to dev yet)

cjllanwarne (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -412,6 +413,10 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
       val executionServiceKey = workflowSubmitResult._1
       val executionServiceResults = workflowSubmitResult._2
 
+      // Emit metric for time until Cromwell has processed workflow.
+      val elapsedTime = System.currentTimeMillis() - submissionRec.submissionDate.getTime
+      workflowToCromwellLatency.update(elapsedTime, TimeUnit.MILLISECONDS)
```
breilly2, line 418: TOL: Anyone have thoughts about whether this should be in the `yield` or the last item in the `for` block? Or maybe even immediately after the `for ... yield`?

cahrens: Good questions. I don't think we want it _after_ the for loop because we are trying to capture the time for the individual workflows in batch. If I understand the yield correctly, it will be executed after each step in the for loop.

I think of the for comprehension as creating variables for subsequent steps, and this metric emitting is not relevant for future steps in the for comprehension. So I wouldn't be inclined to put it in the for unless my understanding of yield is incorrect, and the executing of the yield does not line up with the individual for loop executions.

salonishah11, line 418: Just to make sure I am understanding this correctly, this would be the time between when a "submission" is created and after all the workflows in submission have been submitted to Cromwell?

cahrens: Also take everything I said with a huge grain of salt, given my limited Scala knowledge. If the yield should only be doing things directly related to the return values, moving the metric emitting into the for comprehension would be better. I'm really hoping y'all tell me what the right thing is to do here.

cahrens: The intent was to capture the latency between initial Rawls submission and workflow submission to Cromwell, for each call to `executionServiceCluster.submitWorkflows`. So I guess that is a batch of workflows within the submission.

cahrens: FWIW, [here](https://github.com/broadinstitute/rawls/blob/f70467005d2e226ca62e2304cee9fa6619c4e7f5/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSupervisor.scala#L167) is another example where it looks like yield is being used to collect statistics instead of returning something.

If putting the metric emission into the for comprehension is the better thing to do, I need some help with the syntax.

salonishah11 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
cjllanwarne: > If I understand the yield correctly, it will be executed after each step in the for loop.

Actually it's the opposite. Each `->` in the `for` contents cause everything below it to iterate over every unpacked element. The yield is executed once and collects together all of the results.

So if you want one data point per batch, put it in the `yield`, but if you want one data point per workflow, put it at the end of the `for`.

If you want to put stuff at the end of the for it would be something like:

```
for {
  ...
  ... (existing content...)
  ...
  elapsedTime = System.currentTimeMillis() - submissionRec.submissionDate.getTime
  _ = workflowToCromwellLatency.update(elapsedTime, TimeUnit.MILLISECONDS)
}
```

cahrens: @cjllanwarne But the yield has to execute once for each call to `executionServiceCluster.submitWorkflows` since the yield depends on the result of that call (workflowSubmitResult). So I think it doesn't matter if the code is in the yield or in the for loop immediately after the call to `executionServiceCluster.submitWorkflows`. Am I missing something?

cjllanwarne: Oh good point, I think you're right. TBH I don't normally put anything other than the result value in the `yield` so I probably got it wrong

PR #1605, opened 2022-01-19
[CA-1702][CA-1704][CA-1708] Project-Per-Workspace Workspace Migration Actor

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1702

Add functionality to poll the google storage transfer service for status updates of running jobs. To do this, we 
- peek the database for the first non-finished job that was updated last (ie the longest time ago)
- touch that database reocrd so that future peeks will return another record first
- fetch the operations of the job from the storage transfer service and combine them into an `Outcome`
- write the outcome of the job

If the job finished, we
- update the status of the migration that job relates to
- set the relevent timestamp if the job succeeded or
- fail the migration if the job failed (we can consider retries another time)

Update `PPW_STORAGE_TRANSFER_SERVICE_JOB` to have an `UPDATED` column that automagically updates when you touch the record.

I've included a rough implementation of the Akka Actor as I used this to inform my implementation for polling.
The way it works is this: the system scheduler messages the actor to:
- run the migration pipeline every 10s
- update an active transfer job every 5s (two jobs are created per migration)

We can make these values configurable at another time,

To make this more testable, the migration pipeline runs in reverse. This allows me to set-up a record in the database, run `migrate` and make sure the correct step picked up the record and performed the right effects.

There are a number of cosmetic changes. Sorry for the noise - I try to fix as I go and tidy older code.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
@@ -4,8 +4,9 @@ import cats.effect.IO
 import cats.implicits._
 import cats.kernel.Semigroup
 import org.broadinstitute.dsde.rawls.RawlsException
-import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{Success, Failure}
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{Failure, Success}
 import slick.dbio.{DBIOAction, Effect, NoStream}
+import spray.json.{JsObject, JsString}
 
 import scala.concurrent.Future
 import scala.language.higherKinds
@@ -41,6 +42,23 @@ object MigrationUtils {
           case Failure(msgB) => Failure (msgA ++ "\n" ++ msgB)
         }
       }
+
+    implicit class IgnoreResultExtensionMethod[+R, +S <: NoStream, -E <: Effect](action: DBIOAction[R, S, E]) {
+      /** Ignore the result of the DBIOAction and return unit */
+      def ignore: DBIOAction[Unit, NoStream, E with Effect] = action >> DBIOAction.successful()
+    }
+
+
+    import slick.jdbc.MySQLProfile.api._
+    implicit class InsertExtensionMethod[E, T, C[_]](query: Query[E, T, C]) {
+      /** alias for `+=` supporting dot syntax */
+      def insert(value: T) = query += value
+    }
+
+
+    implicit class FutureToIO[+T](future: => Future[T]) {
+      def io: IO[T] = IO.fromFuture(IO(future))
+    }
   }
 
 
@@ -50,20 +68,10 @@ object MigrationUtils {
   }
 
 
-  implicit class IgnoreResultExtensionMethod[+R, +S <: NoStream, -E <: Effect](action: DBIOAction[R, S, E]) {
-    /** Ignore the result of the DBIOAction and return unit */
-    def ignore: DBIOAction[Unit, NoStream, E with Effect] = action >> DBIOAction.successful()
-  }
+  final case class MigrationException(message: String, data: Map[String, Any] = Map.empty, cause: Throwable = null)
+    extends RawlsException(
+      message = JsObject((data + ("message" -> message)).mapValues(s => new JsString(s.toString))).toString,
+      cause = cause
+    ) {}
 
-
-  import slick.jdbc.MySQLProfile.api._
-  implicit class InsertExtensionMethod[E, T, C[_]](query: Query[E, T, C]) {
-    /** alias for `+=` supporting dot syntax */
-    def insert(value: T) = query += value
-  }
-
-
-  implicit class FutureToIO[+T](future: => Future[T]) {
-    def io: IO[T] = IO.fromFuture(IO(future))
-  }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
@@ -2,7 +2,9 @@ package org.broadinstitute.dsde.rawls.monitor.migration
 
 import cats.effect.IO
 import cats.implicits._
+import cats.kernel.Semigroup
 import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{Success, Failure}
 import slick.dbio.{DBIOAction, Effect, NoStream}
 
 import scala.concurrent.Future
@@ -29,6 +31,16 @@ object MigrationUtils {
       case Success => ("Success".some, None)
       case Failure(msg) => ("Failure".some, msg.some)
     }
+  }
+
+  object Implicits {
+    implicit val semigroupOutcome: Semigroup[Outcome] = (a, b) => a match {
+        case Success => b
+        case Failure(msgA) => b match {
+          case Success => a
+          case Failure(msgB) => Failure (msgA ++ "\n" ++ msgB)
+        }
+      }
   }
 
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
@@ -19,14 +19,12 @@ object MigrationUtils {
 
     final case class Failure(message: String) extends Outcome
 
-    final def fromFields(outcome: Option[String], message: Option[String]): Either[String, Option[Outcome]] = {
-      type EitherStringT[T] = Either[String, T]
-      outcome.traverse[EitherStringT, Outcome] {
+    final def fromFields(outcome: Option[String], message: Option[String]): Either[String, Option[Outcome]] =
+      outcome.traverse[Either[String, *], Outcome] {
         case "Success" => Right(Success)
         case "Failure" => Right(Failure(message.getOrElse("")))
         case other => Left(s"""Failed to read outcome: unknown value -- "$other"""")
       }
-    }
 
     final def toTuple(outcome: Outcome): (Option[String], Option[String]) = outcome match {
       case Success => ("Success".some, None)
@@ -50,14 +48,14 @@ object MigrationUtils {
 
 
     import slick.jdbc.MySQLProfile.api._
-    implicit class InsertExtensionMethod[E, T, C[_]](query: Query[E, T, C]) {
+    implicit class InsertExtensionMethod[E, A, C[_]](query: Query[E, A, C]) {
       /** alias for `+=` supporting dot syntax */
-      def insert(value: T) = query += value
+      def insert(value: A) = query += value
     }
 
 
-    implicit class FutureToIO[+T](future: => Future[T]) {
-      def io: IO[T] = IO.fromFuture(IO(future))
+    implicit class FutureToIO[+A](future: => Future[A]) {
+      def io: IO[A] = IO.fromFuture(IO(future))
     }
   }
 
@@ -68,9 +66,11 @@ object MigrationUtils {
   }
 
 
-  final case class MigrationException(message: String, data: Map[String, Any] = Map.empty, cause: Throwable = null)
+  final case class WorkspaceMigrationException(message: String,
+                                               data: Map[String, Any] = Map.empty,
+                                               cause: Throwable = null)
     extends RawlsException(
-      message = JsObject((data + ("message" -> message)).mapValues(s => new JsString(s.toString))).toString,
+      message = JsObject((Map("message" -> message) ++ data).mapValues(s => new JsString(s.toString))).toString,
       cause = cause
     ) {}
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/PpwStorageTransferJob.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/PpwStorageTransferJob.scala
@@ -12,6 +12,7 @@ final case class PpwStorageTransferJob(id: Long,
                                        jobName: GoogleStorageTransferService.JobName,
                                        migrationId: Long,
                                        created: Timestamp,
+                                       updated: Timestamp,
                                        destBucket: GcsBucketName,
                                        originBucket: GcsBucketName,
                                        finished: Option[Timestamp],
@@ -20,16 +21,29 @@ final case class PpwStorageTransferJob(id: Long,
 
 private[migration]
 object PpwStorageTransferJob {
-  type RecordType = (Long, String, Long, Timestamp, String, String, Option[Timestamp], Option[String], Option[String])
+  type RecordType = (
+    Long,                 // id
+      String,             // jobName
+      Long,               // migrationId
+      Timestamp,          // created
+      Timestamp,          // updated
+      String,             // destBucket
+      String,             // originBucket
+      Option[Timestamp],  // finished
+      Option[String],     // outcome
+      Option[String]      // message
+    )
+
 
   def fromRecord(record: RecordType): Either[String, PpwStorageTransferJob] = record match {
-    case (id, jobName, migrationId, created, destBucket, originBucket, finished, outcome, message) =>
+    case (id, jobName, migrationId, created, updated, destBucket, originBucket, finished, outcome, message) =>
       Outcome.fromFields(outcome, message).map { outcome =>
         PpwStorageTransferJob(
           id,
           GoogleStorageTransferService.JobName(jobName),
           migrationId,
           created,
+          updated,
           GcsBucketName(destBucket),
           GcsBucketName(originBucket),
           finished,
@@ -45,6 +59,7 @@ object PpwStorageTransferJob {
       job.jobName.value,
       job.migrationId,
       job.created,
+      job.updated,
       job.destBucket.value,
       job.originBucket.value,
       job.finished,
@@ -53,6 +68,7 @@ object PpwStorageTransferJob {
     )
   }
 }
+
 
 private[migration]
 object PpwStorageTransferJobs {
@@ -66,6 +82,7 @@ object PpwStorageTransferJobs {
     def jobName = column[String]("JOB_NAME")
     def migrationId = column[Long]("MIGRATION_ID")
     def created = column[Timestamp]("CREATED")
+    def updated = column[Timestamp]("UPDATED")
     def destBucket = column[String]("DEST_BUCKET")
     def originBucket = column[String]("ORIGIN_BUCKET")
     def finished = column[Option[Timestamp]]("FINISHED")
@@ -73,7 +90,7 @@ object PpwStorageTransferJobs {
     def message = column[Option[String]]("MESSAGE")
 
     override def * =
-      (id, jobName, migrationId, created, destBucket, originBucket, finished, outcome, message) <>
+      (id, jobName, migrationId, created, updated, destBucket, originBucket, finished, outcome, message) <>
         (MigrationUtils.unsafeFromEither(PpwStorageTransferJob.fromRecord, _),
           PpwStorageTransferJob.toRecord(_: PpwStorageTransferJob).some)
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
@@ -114,7 +114,8 @@ object WorkspaceMigrationActor {
       deleteWorkspaceBucket,
       createFinalWorkspaceBucket,
       issueTmpBucketTransferJob,
-      deleteTemporaryBucket
+      deleteTemporaryBucket,
+      updateWorkspaceRecord
     )
       .reverse
       .traverse_(runStep)
@@ -418,6 +419,25 @@ object WorkspaceMigrationActor {
       }
 
     } yield transferJob
+
+  final def updateWorkspaceRecord: MigrateAction[Unit] =
+    for {
+      migration <- findMigration(_.tmpBucketDeleted.isDefined)
+      _ <- MigrateAction.liftIO {
+        IO.raiseWhen(migration.newGoogleProjectId.isEmpty || migration.newGoogleProjectNumber.isEmpty) {
+          noGoogleProjectError(migration)
+        }
+      }
+      workspace <- getWorkspace(migration.workspaceId)
+      _ <- inTransaction(dataAccess =>
+        dataAccess
+          .workspaceQuery
+          .filter(_.id === workspace.workspaceIdAsUUID)
+          .map(w => (w.googleProjectId, w.googleProjectNumber))
+          .update((migration.newGoogleProjectId.get.toString, migration.newGoogleProjectNumber.map(_.toString)))
+      )
+      _ <- migrationFinished(migration.id, Success)
+    } yield ()
 
 
   final def refreshTransferJobs: MigrateAction[(Long, Outcome)] =
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
@@ -8,10 +8,10 @@ import cats.implicits._
 import com.google.cloud.storage.Storage.BucketGetOption
 import com.google.longrunning.Operation
 import com.google.storagetransfer.v1.proto.TransferTypes.TransferJob
-import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, RawlsBillingProject, Workspace}
+import org.broadinstitute.dsde.rawls.model.{GoogleProjectId, RawlsBillingProject, RawlsBillingProjectName, Workspace}
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits._
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{Failure, Success, toTuple}
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils._
 import org.broadinstitute.dsde.rawls.monitor.migration.WorkspaceMigrationHistory._
@@ -92,8 +92,12 @@ object WorkspaceMigrationActor {
       ReaderT.liftF(OptionT.liftF(ioa))
 
     // modify the environment that action is evaluated in
-    final def local[A](f: MigrationDeps => MigrationDeps)(action: MigrateAction[A]): MigrateAction[A]
-      = ReaderT.local(f)(action)
+    final def local[A](f: MigrationDeps => MigrationDeps)(action: MigrateAction[A]): MigrateAction[A] =
+      ReaderT.local(f)(action)
+
+    // Raises the error when the condition is true, otherwise returns unit
+    final def raiseWhen(condition: Boolean)(t: => Throwable): MigrateAction[Unit] =
+      MigrateAction.liftIO(IO.raiseWhen(condition)(t))
 
     // empty action
     final def unit: MigrateAction[Unit] =
@@ -146,11 +150,62 @@ object WorkspaceMigrationActor {
       }
 
       workspace <- getWorkspace(migration.workspaceId)
-      currentBillingAccountOnGoogleProject <- MigrateAction.liftIO(IO {
+      errorDetails = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> workspace.workspaceId)
+      )
+
+      _ <- MigrateAction.raiseWhen(workspace.billingAccountErrorMessage.isDefined) {
+        MigrationException(
+          message = "Workspace migration failed: billing account error exists on workspace",
+          data = errorDetails + ("billingAccountErrorMessage" -> workspace.billingAccountErrorMessage.get)
+        )
+      }
+
+      workspaceBillingAccount <- MigrateAction.liftIO(IO {
         workspace.currentBillingAccountOnGoogleProject.getOrElse(
-          throw new RawlsException(s"""No billing account for workspace '${workspace.workspaceId}'""")
+          throw MigrationException(
+            message = "Workspace migration failed: no billing account on workspace",
+            data = errorDetails
+          )
         )
       })
+
+      // Safe to assume that this exists if the workspace exists
+      billingProject <- inTransactionT { _
+        .rawlsBillingProjectQuery
+        .load(RawlsBillingProjectName(workspace.namespace))
+      }
+
+      billingProjectBillingAccount <- MigrateAction.liftIO(IO {
+        billingProject.billingAccount.getOrElse(
+          throw MigrationException(
+            message = "Workspace migration failed: no billing account on billing project",
+            data = errorDetails + ("billingProject" -> billingProject.projectName)
+          )
+        )
+      })
+
+      _ <- MigrateAction.raiseWhen(billingProject.invalidBillingAccount) {
+        MigrationException(
+          message = "The migration failed: invalid billing account on billing project",
+          data = errorDetails ++ Map(
+            ("billingProject" -> billingProject.projectName),
+            ("billingProjectBillingAccount" -> billingProjectBillingAccount)
+          )
+        )
+      }
+
+      _ <- MigrateAction.raiseWhen(workspaceBillingAccount != billingProjectBillingAccount) {
+        MigrationException(
+          message = "Workspace migration failed: billing account on workspace differs from billing account on billing project",
+          data = errorDetails ++ Map(
+            ("workspaceBillingAccount" -> workspaceBillingAccount),
+            ("billingProject" -> billingProject.projectName),
+            ("billingProjectBillingAccount" -> billingProjectBillingAccount)
+          )
+        )
+      }
 
       (workspaceService, billingProject) <- MigrateAction.asks { env =>
         (env.workspaceService, env.billingProject)
@@ -159,7 +214,7 @@ object WorkspaceMigrationActor {
       (googleProjectId, googleProjectNumber) <- MigrateAction.liftIO {
         workspaceService.setupGoogleProject(
           billingProject,
-          currentBillingAccountOnGoogleProject,
+          workspaceBillingAccount,
           workspace.workspaceId,
           workspace.toWorkspaceName
         ).io
@@ -188,6 +243,7 @@ object WorkspaceMigrationActor {
       workspace <- getWorkspace(migration.workspaceId)
       tmpBucketName <- MigrateAction.liftIO(randomSuffix("terra-workspace-migration-"))
       _ <- createBucketInSameRegion(
+        migration,
         GcsBucketName(workspace.bucketName),
         googleProjectId,
         GcsBucketName(tmpBucketName)
@@ -199,7 +255,6 @@ object WorkspaceMigrationActor {
           .filter(_.id === migration.id)
           .map(r => (r.tmpBucket, r.tmpBucketCreated))
           .update((tmpBucketName.some, created.some))
-          .ignore
       }
     } yield ()
 
@@ -248,7 +303,7 @@ object WorkspaceMigrationActor {
           ).compile.last
 
           _ <- IO.raiseUnless(successOpt.contains(true)) {
-            noWorkspaceBucketError(GcsBucketName(workspace.bucketName))
+            noWorkspaceBucketError(migration, GcsBucketName(workspace.bucketName))
           }
         } yield ()
       }
@@ -259,7 +314,6 @@ object WorkspaceMigrationActor {
           .filter(_.id === migration.id)
           .map(_.workspaceBucketDeleted)
           .update(deleted.some)
-          .ignore
       }
     } yield ()
 
@@ -278,7 +332,13 @@ object WorkspaceMigrationActor {
       })
 
       workspace <- getWorkspace(migration.workspaceId)
-      _ <- createBucketInSameRegion(tmpBucketName, googleProjectId, GcsBucketName(workspace.bucketName))
+      _ <- createBucketInSameRegion(
+        migration,
+        tmpBucketName,
+        googleProjectId,
+        GcsBucketName(workspace.bucketName)
+      )
+
       created <- nowTimestamp
       _ <- inTransaction { _ =>
         workspaceMigrations
@@ -350,12 +410,12 @@ object WorkspaceMigrationActor {
           .filter(_.id === migration.id)
           .map(_.tmpBucketDeleted)
           .update(deleted.some)
-          .ignore
       }
     } yield ()
 
 
-  final def createBucketInSameRegion(sourceBucketName: GcsBucketName,
+  final def createBucketInSameRegion(migration: WorkspaceMigration,
+                                     sourceBucketName: GcsBucketName,
                                      destGoogleProject: GoogleProjectId,
                                      destBucketName: GcsBucketName)
   : MigrateAction[Unit] =
@@ -372,7 +432,7 @@ object WorkspaceMigrationActor {
             List(BucketGetOption.userProject(destGoogleProject.value))
           )
 
-          sourceBucket <- IO(sourceBucketOpt.getOrElse(throw noWorkspaceBucketError(sourceBucketName)))
+          sourceBucket <- IO(sourceBucketOpt.getOrElse(throw noWorkspaceBucketError(migration, sourceBucketName)))
 
           // todo: CA-1637 do we need to transfer the storage logs for this workspace? the logs are prefixed
           // with the ws bucket name, so we COULD do it, but do we HAVE to? it's a csv with the bucket
@@ -590,15 +650,36 @@ object WorkspaceMigrationActor {
   }
 
   final def noGoogleProjectError[A](migration: WorkspaceMigration): Throwable =
-    new IllegalStateException(s"""Google Project "${migration.newGoogleProjectId}" was not found for migration "${migration.id}".""")
+    MigrationException(
+      message = "Workspace migration failed: Google Project not found.",
+      data = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> migration.workspaceId),
+        ("googleProjectId" -> migration.newGoogleProjectId)
+      )
+    )
 
 
-  final def noWorkspaceBucketError[A](name: GcsBucketName): Throwable =
-    new IllegalStateException(s"""Failed to retrieve workspace bucket "${name}".""")
+  final def noWorkspaceBucketError[A](migration: WorkspaceMigration, bucket: GcsBucketName): Throwable =
+    MigrationException(
+      message = "Workspace migration failed: Workspace cloud bucket not found.",
+      data = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> migration.workspaceId),
+        ("workspaceBucket" -> bucket)
+      )
+    )
 
 
   final def noTmpBucketError[A](migration: WorkspaceMigration): Throwable =
-    new IllegalStateException(s"""Temporary storage bucket "${migration.tmpBucketName}" was not found for migration "${migration.id}".""")
+    MigrationException(
+      message = "Workspace migration failed: Temporary cloud storage bucket not found.",
+      data = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> migration.workspaceId),
+        ("tmpBucket" -> migration.tmpBucketName)
+      )
+    )
 
 
   sealed trait Message
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
@@ -21,7 +21,7 @@ import org.broadinstitute.dsde.workbench.google2.{GoogleStorageService, GoogleSt
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
 
 import java.sql.Timestamp
-import java.time.{Clock, Instant, LocalDateTime, ZoneOffset}
+import java.time.{Instant, LocalDateTime, ZoneOffset}
 import java.util.UUID
 import scala.collection.JavaConversions._
 import scala.concurrent.ExecutionContext.Implicits.global
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
@@ -80,7 +80,7 @@ object WorkspaceMigrationActor {
   object MigrateAction {
 
     // lookup a value in the environment using `selector`
-    final def asks[T](selector: MigrationDeps => T): MigrateAction[T] =
+    final def asks[A](selector: MigrationDeps => A): MigrateAction[A] =
       ReaderT.ask[OptionT[IO, *], MigrationDeps].map(selector)
 
     // create a MigrateAction that ignores its input and returns the OptionT
@@ -156,7 +156,7 @@ object WorkspaceMigrationActor {
       )
 
       _ <- MigrateAction.raiseWhen(workspace.billingAccountErrorMessage.isDefined) {
-        MigrationException(
+        WorkspaceMigrationException(
           message = "Workspace migration failed: billing account error exists on workspace",
           data = errorDetails + ("billingAccountErrorMessage" -> workspace.billingAccountErrorMessage.get)
         )
@@ -164,7 +164,7 @@ object WorkspaceMigrationActor {
 
       workspaceBillingAccount <- MigrateAction.liftIO(IO {
         workspace.currentBillingAccountOnGoogleProject.getOrElse(
-          throw MigrationException(
+          throw WorkspaceMigrationException(
             message = "Workspace migration failed: no billing account on workspace",
             data = errorDetails
           )
@@ -179,7 +179,7 @@ object WorkspaceMigrationActor {
 
       billingProjectBillingAccount <- MigrateAction.liftIO(IO {
         billingProject.billingAccount.getOrElse(
-          throw MigrationException(
+          throw WorkspaceMigrationException(
             message = "Workspace migration failed: no billing account on billing project",
             data = errorDetails + ("billingProject" -> billingProject.projectName)
           )
@@ -187,7 +187,7 @@ object WorkspaceMigrationActor {
       })
 
       _ <- MigrateAction.raiseWhen(billingProject.invalidBillingAccount) {
-        MigrationException(
+        WorkspaceMigrationException(
           message = "The migration failed: invalid billing account on billing project",
           data = errorDetails ++ Map(
             ("billingProject" -> billingProject.projectName),
@@ -197,7 +197,7 @@ object WorkspaceMigrationActor {
       }
 
       _ <- MigrateAction.raiseWhen(workspaceBillingAccount != billingProjectBillingAccount) {
-        MigrationException(
+        WorkspaceMigrationException(
           message = "Workspace migration failed: billing account on workspace differs from billing account on billing project",
           data = errorDetails ++ Map(
             ("workspaceBillingAccount" -> workspaceBillingAccount),
@@ -480,13 +480,12 @@ object WorkspaceMigrationActor {
 
     } yield transferJob
 
+
   final def updateWorkspaceRecord: MigrateAction[Unit] =
     for {
       migration <- findMigration(_.tmpBucketDeleted.isDefined)
-      _ <- MigrateAction.liftIO {
-        IO.raiseWhen(migration.newGoogleProjectId.isEmpty || migration.newGoogleProjectNumber.isEmpty) {
-          noGoogleProjectError(migration)
-        }
+      _ <- MigrateAction.raiseWhen(migration.newGoogleProjectId.isEmpty || migration.newGoogleProjectNumber.isEmpty) {
+        noGoogleProjectError(migration)
       }
       workspace <- getWorkspace(migration.workspaceId)
       _ <- inTransaction(dataAccess =>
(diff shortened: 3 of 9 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -88,32 +97,34 @@ object WorkspaceMigrationMonitor {
...
+      createFinalWorkspaceBucket,
+      issueTmpBucketTransferJob,
+      deleteTemporaryBucket
+    )
+      .reverse
+      .traverse_(runStep)
+
+
+  // Sequence the action and return an empty MigrateAction if the action succeeded
+  def runStep(action: MigrateAction[Unit]): MigrateAction[Unit] =
```
s-rubenstein, line 129: I like this optimization. Very clear.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -402,13 +410,148 @@ object WorkspaceMigrationMonitor {
...
+      (storageTransferService, billingProject) <- MigrateAction.asks { env =>
+        (env.storageTransferService, GoogleProject(env.billingProject.googleProjectId.value))
+      }
+
+      outcome <- MigrateAction.liftF {
+        OptionT {
+          storageTransferService
+            .listTransferOperations(record.jobName, billingProject)
+            .map(_.foldLeft(Option.empty[Outcome]) {
+              (outcome, operation) => outcome |+| getOperationOutcome(operation)
```
s-rubenstein: What is the `|+|` operator? Googling seemed to not pick it up super well

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -71,13 +75,18 @@ object WorkspaceMigrationMonitor {
                                  storageTransferService: GoogleStorageTransferService[IO])
 
 
-  type MigrateAction[T] = ReaderT[OptionT[IO, *], MigrationDeps, T]
+  type MigrateAction[A] = ReaderT[OptionT[IO, *], MigrationDeps, A]
```
s-rubenstein, line 78: What does the change from `T` to `A` do?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -188,7 +199,7 @@ object WorkspaceMigrationMonitor {
     } yield ()
 
 
-  final def issueWorkspaceBucketTransferJob: MigrateAction[TransferJob] =
+  final def issueWorkspaceBucketTransferJob: MigrateAction[Unit] =
```
s-rubenstein, line 259: Why do we want to change this from a TransferJob to a Unit? My preference is usually more specific return types, rather than less, is there a reason that this needs to return things other than a TransferJob? Does it?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -268,7 +284,7 @@ object WorkspaceMigrationMonitor {
     } yield ()
 
 
-  final def issueTmpBucketTransferJob: MigrateAction[TransferJob] =
+  final def issueTmpBucketTransferJob: MigrateAction[Unit] =
```
s-rubenstein, line 349: Same comment here about TransferJob=>Unit

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -402,13 +410,148 @@ object WorkspaceMigrationMonitor {
...
+                  ))
+                }
+            },
+          migrations
+            .filter(_.workspaceBucketTransferred.isDefined)
+            .map(_.tmpBucketTransferred)
+            .update(transferred.some),
+          migrations
+            .filter(_.workspaceBucketTransferred.isEmpty)
+            .map(_.workspaceBucketTransferred)
```
s-rubenstein, line 558: I noticed this has a different format than above, but is still approaching the same type of operation (bucket transferred). Is there a reason to not use the same `isDefined` or `isEmpty` for both of them? E.g.,
```
migrations
  .filter(_.workspaceBucketTransferred.isDefined)
  .map(_.tmpBucketTransferred)
  .update(transferred.some),
migrations
  .filter(_.finalBucketTransferred.isDefined) //Whatever this variable should be
  .map(_.workspaceBucketTransferred)
  .update(transferred.some)
```

Because they are so similar as it is, but the implementation is different, I am worried that future us is going to do the same thing I just did, and assume they are checking in the same way, when they arent.

s-rubenstein (commented):
I've been noticing a lot of code style changes along with the functional changes , which has been leading to large diffs that are hard for me personally to review. Especially because they aren't in the PR description, it leaves me somewhat struggling to figure out the intended goals of the change. If possible, in the future, would you be willing to lift those style changes into their own PRs, which can be reviewed individually?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: It's the semigroup infix combine operator
https://typelevel.org/cats/typeclasses/semigroup.html

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Nothing, it's a style change for consistency.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Because the return value is not used and the status of the transfer job is tested elsewhere. This makes it cleaner to compose together the steps in `migrate` because now I don't need to do `issueWorkspaceBucketTransferJob >> MigrateAction.unit`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: If the temp bucket has been transferred already then that's an error. What I'm trying to do is split the migrations by if they've had their workspace bucket transferred to the tmp bucket. To those that have, I'm setting the flag that the tmp bucket has been transferred. To those that haven't I'm setting it.

ehigham: I've added a comment to clarify

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Found another way of doing it :)

ehigham:
> I've been noticing a lot of code style changes along with the functional changes , which has been leading to large diffs that are hard for me personally to review. Especially because they aren't in the PR description, it leaves me somewhat struggling to figure out the intended goals of the change. If possible, in the future, would you be willing to lift those style changes into their own PRs, which can be reviewed individually?

Sorry about that. I've updated the description so hopefully it better outlines the main aspects of this PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
s-rubenstein: Works for me. Would it be worth changing the other `T` values as well then?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
s-rubenstein: Maybe its my experiences working in typeless languages, but even if the return type isn't being used, do you think it is more clear to make this return something else then if the output isn't being used? I am personally concerned about future usage and clarity -> I would prefer either we explicitly return something unusable (so that people looking at this function explicitly don't use the returned object), or we clarify the return type so that users know exactly what they are getting and don't end up with accidental type mismatches down the line.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Fix as we go I think :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: I don't know. I guess I'm just viewing each of these steps as just side effects and the migration "pipeline" as a sequencing of these side effects. Personally I feel that unused return values are a bit of a code smell. I'm not sure what you mean by returning something unusable. Could you clarify?

s-rubenstein: Discussed offline.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: On you mean in this PR? Yeah sure! (Seems I'm consistently inconsistent lol)

s-rubenstein: Yeah, I meant in this PR :D, since we're already changing some values.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
```diff
@@ -16,42 +19,59 @@ object MigrationUtils {
 
     final case class Failure(message: String) extends Outcome
 
-    final def fromFields(outcome: Option[String], message: Option[String]): Either[String, Option[Outcome]] = {
-      type EitherStringT[T] = Either[String, T]
-      outcome.traverse[EitherStringT, Outcome] {
+    final def fromFields(outcome: Option[String], message: Option[String]): Either[String, Option[Outcome]] =
+      outcome.traverse[Either[String, *], Outcome] {
```
andy7i, line 23: intellij isn't telling me what it is, but based on what I found it seems like `*` is a kind, like a String or List[A]. Is that correct?
beyond that, I'm having difficulty understanding how the traverse is creating the return type `Either[String, Option[Outcome]]`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
```diff
@@ -16,42 +19,59 @@ object MigrationUtils {
...
         case "Success" => Right(Success)
         case "Failure" => Right(Failure(message.getOrElse("")))
         case other => Left(s"""Failed to read outcome: unknown value -- "$other"""")
```
andy7i, line 26: do we need to return an Option[Outcome]? Otherwise I think we can simplify this code to something like:

```
    final def fromFields(outcome: Option[String], message: Option[String]): Either[String, Outcome] =
      outcome.getOrElse("") match {
        case "Success" => Right(Success)
        case "Failure" => Right(Failure(message.orEmpty))
        case other => Left(s"""Failed to read outcome: unknown value -- "$other"""")
      }
```

based on the usage of this function, I think we could make that work

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
```diff
@@ -16,42 +19,59 @@ object MigrationUtils {
...
+    case Right(r) => r
+    case Left(msg) => throw new RawlsException(msg)
   }
+
+
+  final case class WorkspaceMigrationException(message: String,
+                                               data: Map[String, Any] = Map.empty,
+                                               cause: Throwable = null)
+    extends RawlsException(
+      message = JsObject((Map("message" -> message) ++ data).mapValues(s => new JsString(s.toString))).toString,
```
andy7i, line 73: are we converting the message to a json format so it's easier for us to parse the logs?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/PpwStorageTransferJob.scala
```diff
@@ -20,16 +21,29 @@ final case class PpwStorageTransferJob(id: Long,
 
 private[migration]
 object PpwStorageTransferJob {
-  type RecordType = (Long, String, Long, Timestamp, String, String, Option[Timestamp], Option[String], Option[String])
+  type RecordType = (
```
andy7i, line 24: naive question -- could we use an apply function instead of defining `RecordType`? Or do we need `RecordType` to work with the DB more easily?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+
+object WorkspaceMigrationActor {
+
+  import slick.jdbc.MySQLProfile.api._
+
+  val workspaceMigrations = WorkspaceMigrationHistory.workspaceMigrations
+  val storageTransferJobs = PpwStorageTransferJobs.storageTransferJobs
+
+
+  final def truncate: WriteAction[Unit] =
```
andy7i, line 40: why is this named `truncate`? It seems like it is deleting rather than shortening something.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+    storageTransferJobs.delete >> workspaceMigrations.delete >> DBIO.successful()
+
+
+  final def isInQueueToMigrate(workspace: Workspace): ReadWriteAction[Boolean] =
```
andy7i, line 44: nitpicky -- could this (and `isMigrating` below) be `ReadAction`s?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+    final def asks[A](selector: MigrationDeps => A): MigrateAction[A] =
+      ReaderT.ask[OptionT[IO, *], MigrationDeps].map(selector)
+
+    // create a MigrateAction that ignores its input and returns the OptionT
+    final def liftF[A](optionT: OptionT[IO, A]): MigrateAction[A] =
+      ReaderT.liftF(optionT)
+
+    // lift an IO action into the context of a MigrateAction
+    final def liftIO[A](ioa: IO[A]): MigrateAction[A] =
+      ReaderT.liftF(OptionT.liftF(ioa))
```
s-rubenstein, line 92: Just a style thing, but I think this could call `liftF`? If I am understanding correctly?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+      )
+
+      _ <- MigrateAction.raiseWhen(workspace.billingAccountErrorMessage.isDefined) {
+        WorkspaceMigrationException(
+          message = "Workspace migration failed: billing account error exists on workspace",
+          data = errorDetails + ("billingAccountErrorMessage" -> workspace.billingAccountErrorMessage.get)
+        )
+      }
+
+      workspaceBillingAccount <- MigrateAction.liftIO(IO {
```
s-rubenstein, line 165: Why is this one not a `raiseWhen`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+        WorkspaceMigrationException(
+          message = "The migration failed: invalid billing account on billing project",
+          data = errorDetails ++ Map(
+            ("billingProject" -> billingProject.projectName),
+            ("billingProjectBillingAccount" -> billingProjectBillingAccount)
+          )
+        )
+      }
+
+      _ <- MigrateAction.raiseWhen(workspaceBillingAccount != billingProjectBillingAccount) {
```
s-rubenstein, line 199: We have a whole sequence of checks like this, which all follow the same format:
```
_ <- MigrateAction.raiseWhen(_predicate_) {
  WorkspaceMigrationException(
    message,
    data
  )
)
```
This feels like a map to me. Could we pull these out into a map of check->object that let us avoid these repeated raiseWhens? Then we can just iterate over the map and follow the same format. I think this might let us abstract the business logic from the implementation, which would make things cleaner.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+        workspaceMigrations
+          .filter(_.id === migrationId)
+          .map(m => (m.finished, m.outcome, m.message))
+          .update((finished.some, status, message))
+          .ignore
+      }
+    }
+
+
+  def getMigrations(workspaceUuid: UUID): MigrateAction[Seq[WorkspaceMigration]] =
```
s-rubenstein, line 578: I tried searching this method in IntelliJ, and there were no usages. Where is this used?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
```diff
@@ -0,0 +1,731 @@
...
+  // This is a horrible hack to avoid refactoring the tangled mess in the WorkspaceServiceSpec.
+  val spec = new WorkspaceServiceSpec()
+
+
+  def runMigrationTest(test: MigrateAction[Assertion]): Assertion =
+    spec.withTestDataServices { services =>
+      test.run {
+        MigrationDeps(
+          services.slickDataSource,
+          GoogleProject("fake-google-project"),
```
s-rubenstein, line 60: Before when we did this we read from a value in our test file. Could we do that here? I am always a little bit concerned when I see raw strings in test code that aren't parameterized, because it can lead to string proliferation across the tests (which becomes brittle)

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
```diff
@@ -0,0 +1,731 @@
...
+  case class MockStorageService() extends MockGoogleStorageService[IO] {
+    override def deleteBucket(googleProject: GoogleProject, bucketName: GcsBucketName, isRecursive: Boolean, bucketSourceOptions: List[Storage.BucketSourceOption], traceId: Option[TraceId], retryConfig: RetryConfig): fs2.Stream[IO, Boolean] =
+      fs2.Stream.emit(true)
+
+    override def insertBucket(googleProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]], labels: Map[String, String], traceId: Option[TraceId], bucketPolicyOnlyEnabled: Boolean, logBucket: Option[GcsBucketName], retryConfig: RetryConfig, location: Option[String]): fs2.Stream[IO, Unit] =
+      fs2.Stream.emit(())
+  }
+
+
+  def getAttempt(workspaceUuid: UUID): ReadWriteAction[Option[WorkspaceMigration]] =
```
s-rubenstein, line 98: Could we use a `getMigrationAttempt` method from the real actor for this?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
```diff
@@ -0,0 +1,731 @@
...
+    }
+
+
+  it should "error when the billing account on the workspace does not match the billing account on the billing project" in
+    assertThrows[WorkspaceMigrationException] {
+      runMigrationTest {
+        for {
+          now <- nowTimestamp
+          workspace = spec.testData.v1Workspace.copy(
+            currentBillingAccountOnGoogleProject = RawlsBillingAccountName("greg").some,
```
s-rubenstein, line 273: To ensure that this is different, could we base it on the other billing account? (maybe something like `spec.testData.v1Workspacec.billingAccount + "A"`?)

Just because nothing stops the billing account on the billing project from being `"greg"`

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
```diff
@@ -0,0 +1,731 @@
...
+        }
+      } yield migration.tmpBucketDeleted shouldBe defined
+    }
+
+
+  it should "update the Workspace record when the temporary bucket has been deleted" in
+    runMigrationTest {
+      for {
+        now <- nowTimestamp
+        googleProjectId = GoogleProjectId("whatever dude")
```
s-rubenstein, line 508: I like the test level variable abstraction! Nice.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
```diff
@@ -0,0 +1,731 @@
...
+      } yield {
+        transferJob.jobName.value shouldBe job.getName
+        transferJob.migrationId shouldBe migration.id
+        transferJob.originBucket shouldBe workspaceBucketName
+        transferJob.destBucket shouldBe tmpBucketName
+      }
+    }
+
+
+  "peekTransferJob" should "return the first active job that ws updated last and touch it" in
```
s-rubenstein, line 561: Nit: `ws` -> `was`

s-rubenstein (approved):
Compare with clipboard was a godsend for this one. I have a couple cleanup opportunities, but I think the error abstractions can be done in a follow-up and I don't like blocking on nits.

We have follow up conversations around large PRs, but that is future looking to me, not necessarily for this one?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: It is in a subsequent PR

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: No - `liftF` only lifts one level down. `liftF` (or `lift`) comes from `MonadTrans` which can be written in scala as:
```scala
trait MonadTrans[T[_, _], F[_]] {
   def liftF[A](fa: F[A]): T[F, A]
}
```
`liftIO` is a special kind of a monad transformer and has it's own type class because lifting IO is a common pattern:
```scala
trait LiftIO[F[_]] {
   def liftIO[A](ioa: IO[A]): F[A]
}
```
`liftIO` can lift an `IO` at any level (unlike `liftF` where it's one level). We're trying to "lift" an IO into `OptionT` and then into `ReaderT`, thus it needs to call `liftF` twice. Does that help?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Can you give a code example? I don't quite follow your suggestion.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Ah it used to be used in tests - intellij's dead code highlighting isn't working... Thank you

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
ehigham: I'll move this into a variable and name it

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
ehigham: The actor code doesn't really need/use this functionality

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceMigrationActorSpec.scala
ehigham: Sure!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
s-rubenstein: Got it! Thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
s-rubenstein: As I am typing out the code, I think this might be more complexity than its worth, but here goes.

So, for example. (And this is kind of pseudocode-y?)
```
migrationPrerequisites = List.of(
  {
    predicate: (workspace, billingProject) => getWorkspaceBillingAccount(workspace) != getBillingProjectBillingAccount(billingProject)
    message: "Workspace migration failed: billing account on workspace differs from billing account on billing project"
    detailKeys: List.of("migrationId", "workspaceId", "workspaceBillingAccount", "billingProject", "billingProjectBillingAccount")
  },
  {
    predicate: (workspace, billingProject) => billingProject.invalidBillingAccount
    message: "Workspace migration failed: no billing account on billing project"
    detailKeys: List.of("migrationId", "workspaceId", "billingProject")
  }
)
```

What I think this would do is abstract out the business logic into this list. Then, our error handling code turns into:
```
migrationPrerequisites.map(migrationPrerequisiteObject => MigrateAction.raiseWhen(migrationPrerequisiteObject) {
  WorkspaceMigrationException(message = migrationPrerequisiteObject.message, data = errorDetailsFullList.filter(errorDetailOption => migrationPrerequisiteObject.detailKeys.contains(errorDetailOption.key)))
}
```
I think the error keys and ensuring that the args to the predicate are always the same is where the complexity builds, but the thought is that the error object construction be done once, so if we change the way we want to throw errors, we only need to change the `migrationPrerequisites.map` code block once, as opposed to for each of the prerequisites. This would help avoid situations like the one where there was one set not in the `raiseWhen`, and I was confused, but its just that it was missed.

andy7i:
Is there some reasonable way to split up this PR? (I'm not sure if there is, but thought I'd ask to get your input)
And if not, since this PR covers more than just CA-1702, should we add those ticket #s into the title of the PR as well? (like `[CA-1702][CA-1704]` and so on.) I think it would help with tracking when those tickets got merged into dev/prod.

[CA-1702]: https://broadworkbench.atlassian.net/browse/CA-1702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CA-1704]: https://broadworkbench.atlassian.net/browse/CA-1704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+                                 storageTransferService: GoogleStorageTransferService[IO])
+
+
+  type MigrateAction[A] = ReaderT[OptionT[IO, *], MigrationDeps, A]
+
+  object MigrateAction {
+
+    // lookup a value in the environment using `selector`
+    final def asks[A](selector: MigrationDeps => A): MigrateAction[A] =
+      ReaderT.ask[OptionT[IO, *], MigrationDeps].map(selector)
```
andy7i, line 84: the use of Kleisli is really cool! it does feel a bit like black magic though (like `OptionT[IO, *]` or how the hell the return type is MigrateAction). Maybe I need to read up on category theory?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: I like the idea, I just think it's slightly over-engineered. As it currently stands it's simple and I'm hesitant to change that. What's the issue for you? I think the function is perhaps a bit long but I was on the fence about splitting it up because I wasn't sure it was worth it.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+  def getMigrations(workspaceUuid: UUID): MigrateAction[Seq[WorkspaceMigration]] =
+    inTransaction { _ =>
+      workspaceMigrations
+        .filter(_.workspaceId === workspaceUuid)
+        .sortBy(_.id.asc)
+        .result
+    }
+
+
+  final def findMigration(predicate: WorkspaceMigrationHistory => Rep[Boolean])
```
andy7i, line 587: Q: if findMigration does not return anything, do for comps stop?

ehigham:
> Is there some reasonable way to split up this PR? (I'm not sure if there is, but thought I'd ask to get your input) 

Can you look at individual commits? You don't need to review `c03fb7f` and `1d2380b`

[CA-1702]: https://broadworkbench.atlassian.net/browse/CA-1702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CA-1704]: https://broadworkbench.atlassian.net/browse/CA-1704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
s-rubenstein: I agree that its a little over-engineered. The predicate stuff was particularly hard to figure out in a way that as I was writing it up seemed rough. I was mostly just noticing that much of the structure seemed to echo other lines around it, so was curious if we could break it into a map or a function in order to avoid the code duplication, but reducing the code duplication isn't worth adding complexity in my book.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+  final def inTransaction[A](action: DataAccess => ReadWriteAction[A]): MigrateAction[A] =
+    for {
+      dataSource <- MigrateAction.asks(_.dataSource)
+      result <- MigrateAction.liftIO(dataSource.inTransaction(action).io)
+    } yield result
+
+
+  final def getWorkspace(workspaceId: UUID): MigrateAction[Workspace] =
+    inTransaction {
+      _.workspaceQuery.findByIdOrFail(workspaceId.toString)
```
andy7i, line 633: Q: `_` here is the MigrationDeps?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+
+      workspace <- getWorkspace(migration.workspaceId)
+      errorDetails = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> workspace.workspaceId)
+      )
+
+      _ <- MigrateAction.raiseWhen(workspace.billingAccountErrorMessage.isDefined) {
+        WorkspaceMigrationException(
+          message = "Workspace migration failed: billing account error exists on workspace",
```
andy7i, line 160: Q: if the workspace has a billing account error, does that get cleared if the billing account is successfully updated?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+
+  final def nowTimestamp: MigrateAction[Timestamp] =
+    MigrateAction.liftIO(IO {
+      Timestamp.valueOf(LocalDateTime.ofInstant(Instant.now, ZoneOffset.UTC))
+    })
+
+
+  final def randomSuffix(str: String): IO[String] = IO {
+    str ++ UUID.randomUUID.toString.replace("-", "")
+  }
```
andy7i, line 645: Why is it better to return this as an IO[String] rather than a String?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+                                     destBucketName: GcsBucketName)
+  : MigrateAction[Unit] =
+    for {
+      (storageService, googleProjectToBill) <- MigrateAction.asks { env =>
+        (env.storageService, env.googleProjectToBill)
+      }
+      _ <- MigrateAction.liftIO {
+        for {
+          sourceBucketOpt <- storageService.getBucket(
+            googleProjectToBill, // bill "requester-pays" requests to the migration project
```
andy7i, line 426: Q: is this only billing "requester-pays" requests to the `googleProjectToBill` project, or is it billing ALL of our bucket requests to this project?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+
+  final def issueWorkspaceBucketTransferJob: MigrateAction[Unit] =
+    for {
+      migration <- findMigration { m =>
+        m.tmpBucketCreated.isDefined && m.workspaceBucketTransferJobIssued.isEmpty
+      }
+
+      tmpBucketName <- MigrateAction.liftIO(IO {
+        migration.tmpBucketName.getOrElse(throw noGoogleProjectError(migration))
+      })
```
andy7i, line 267: could this work?

```
tmpBucketName = migration.tmpBucketName.getOrElse(throw noGoogleProjectError(migration))
```

or is it better to wrap it in an IO because otherwise it evaluates eagerly..?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+          ).compile.drain
+        } yield ()
+      }
+    } yield ()
+
+
+  final def startBucketTransferJob(migrationId: Long,
+                                   originBucket: GcsBucketName,
+                                   destBucket: GcsBucketName)
+  : MigrateAction[TransferJob] =
```
andy7i, line 452: do we want to return anything other than () here?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+    for {
+      (storageTransferService, googleProject) <- MigrateAction.asks { env =>
+        (env.storageTransferService, env.googleProjectToBill)
+      }
+      transferJob <- MigrateAction.liftIO {
+        for {
+          jobName <- randomSuffix("transferJobs/terra-workspace-migration-")
+          transferJob <- storageTransferService.createTransferJob(
+            jobName = GoogleStorageTransferService.JobName(jobName),
+            jobDescription = s"""Bucket transfer job from "${originBucket}" to "${destBucket}".""",
```
andy7i, line 462: I wonder if it would make debugging a bit easier if we also included the workspace name here

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+            JobTransferSchedule.Immediately
+          )
+        } yield transferJob
+      }
+
+      _ <- inTransaction { _ =>
+        storageTransferJobs
+          .map(job => (job.jobName, job.migrationId, job.destBucket, job.originBucket))
+          .insert((transferJob.getName, migrationId, destBucket.value, originBucket.value))
+      }
```
andy7i, line 475: similar thought -- I wonder if it would make debugging a bit easier if the workspace info is readily accessible in the db.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+  final def claimAndConfigureGoogleProject: MigrateAction[Unit] =
+    for {
+      migration <- findMigration { m =>
+        m.started.isDefined && m.newGoogleProjectConfigured.isEmpty
+      }
+
+      workspace <- getWorkspace(migration.workspaceId)
+      errorDetails = Map(
+        ("migrationId" -> migration.id),
+        ("workspaceId" -> workspace.workspaceId)
```
andy7i, line 155: for debugging, it might be useful to get the workspace name in the error message. (Since the workspaceId is a UUID, that means we would need to query the DB for the workspace name. Adding it here would save us a step. (a step I hope we never need to take))

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+          // if the workspace bucket has already been transferred then this job relates transferring
+          // the tmp -> final workspace bucket
+          migrations
+            .filter(_.workspaceBucketTransferred.isDefined)
+            .map(_.tmpBucketTransferred)
+            .update(transferred.some),
+          // if the workspace bucket has not been defined then this job relates to transferring
+          // old workspace bucket -> tmp bucket
+          migrations
+            .filter(_.workspaceBucketTransferred.isEmpty)
```
andy7i, line 557: Q: is this logic better than a simpler if/else?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+                )
+              )
+              .value
+              .void
+              .unsafeRunSync
+          } catch {
+            case failure: Throwable => context.executionContext.reportFailure(failure)
+          }
+        }
+        Behaviors.same
```
andy7i, line 720: this is some sort of optimization? do we need it?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+  // and integration tested as part of CA-1132.
+  // todo: finalise design and test in CA-1132
+  def apply(pollingInterval: FiniteDuration,
+            dataSource: SlickDataSource,
+            googleProjectToBill: GoogleProject,
+            workspaceService: WorkspaceService,
+            storageService: GoogleStorageService[IO],
+            storageTransferService: GoogleStorageTransferService[IO])
+  : Behavior[Message] =
+    Behaviors.setup { context =>
```
andy7i, line 697: Q: Behaviors?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+      migration <- findMigration { m =>
+        m.workspaceBucketDeleted.isDefined && m.finalBucketCreated.isEmpty
+      }
+
+      (googleProjectId, tmpBucketName) <- MigrateAction.liftIO(IO {
+        (
+          migration.newGoogleProjectId.getOrElse(throw noGoogleProjectError(migration)),
+          migration.tmpBucketName.getOrElse(throw noTmpBucketError(migration))
+        )
+      })
```
andy7i, line 329: this block is duplicated in `deleteTemporaryBucket`(). is it worth refactoring?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+      (googleProjectId, googleProjectNumber) <- MigrateAction.liftIO {
+        workspaceService.setupGoogleProject(
+          billingProject,
+          workspaceBillingAccount,
+          workspace.workspaceId,
+          workspace.toWorkspaceName
+        ).io
+      }
+
+      configured <- nowTimestamp
```
andy7i, line 220: pretty minor nitpick: feels like this could be named something more like "now" or "timestamp" and that might be a bit clearer below.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -0,0 +1,746 @@
...
+      _ <- inTransaction { _ =>
+        storageTransferJobs
+          .map(job => (job.jobName, job.migrationId, job.destBucket, job.originBucket))
+          .insert((transferJob.getName, migrationId, destBucket.value, originBucket.value))
+      }
+
+    } yield transferJob
+
+
+  final def updateWorkspaceRecord: MigrateAction[Unit] =
```
andy7i, line 480: minor: could we move `updateWorkspaceRecord` to after `deleteTemporaryBucket`?

andy7i (commented):
Really well organized methods / code. I learned a lot reading through this!

I ended up reading through all of the changes in WorkspaceMigrationActor to get more context and a better understanding overall of how the migration works. (I haven't gone through the tests yet.. and in the interest of time I might not look at them for this review)

I mostly just have a lot of questions. I'm sharing them here so they don't get lost.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: OK I tried many things - I tried a reduce, I tried to write simple header functions, I tired using `StateT` - it all leads to code explosion or increased complexity. I'm going to leave this.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
ehigham: This is a compiler extension I've enabled to work with partial type application. This stack overflow link explains it quite well:
https://stackoverflow.com/questions/5527684/partially-applying-type-parameters

andy7i:
do we know why the jenkins test startup failed?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
ehigham: I actually want it to be `Option` because callling `fromFields(None, None)` is a valid thing to do.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
ehigham: I'm converting to JSON because java and inheritance. One of my pet hates is string formatting in error messages - it's arbitrary and often doesn't provide enough information. Thus, I've included an extra `data` field to capture the extra data that helps provide more context to error messages. In JVM languages, however, often exceptions are caught by thier super-type (e. `Exception` or `Throwable`). The method `getMessage` is defined on `Throwable` to just return the `message` string and so if I don't serialise the extra map of data, then it'll be missed in any exception logging. I've got to choose some serialisation format so I chose JSON as that's how we're mandated to format logs.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/PpwStorageTransferJob.scala
ehigham: We need it to work with the database. See https://scala-slick.org/doc/3.2.1/gettingstarted.html for more info. In short, the database requires a tuple for its table type (via operator `*`).

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: It's called that to be consistent with the function that calls it for tests.
https://github.com/broadinstitute/rawls/blob/487e2b052cad4951050ba4dc18532e17d720e715/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala#L49-L83

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Yes they can!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: They're skipped, analogous to a for-comp over `Option`.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: https://github.com/broadinstitute/rawls/blob/487e2b052cad4951050ba4dc18532e17d720e715/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala#L587-L588

`inTransaction` takes a function from `DataAccess` to `ReadWriteAction[A]`. `_` is "placeholder syntax" for the argument to the function, so `DataAccess`. Here's a nice overview of underscore usage in scala (it's a lot lol):
https://stackoverflow.com/questions/8000903/what-are-all-the-uses-of-an-underscore-in-scala

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: I think so, at least that's how it seemed when i was doing some UI work for PPW.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Honestly it's pedentry. Random number generation is not pure (ie it has side effects) so I have deferred this in IO. I could have used `Random` from cats but I think that would just be overkill.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Here's where my knowledge of Scala breaks down and my head starts to hurt because we're mixing imperitive and functional code. It turns out that it's the same in a for-comp over a monad stack that includes `IO`: Expanding the operations:
```scala
ReaderT.flatMap(fa, f)
    = Reader { env => OptionT.flatMap(fa(env), a => f(a)(env).value) }
    = Reader { env => OptionT { IO.flatMap(fa(env).value, option => if (option.isEmpty) IO.pure(None) else (a => f(a)(env).value)(option.get)) } }
```
Then `IO.flatMap` does a try-catch. So yes. In this case, it's the same!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: It's returning `MigrateAction[TransferJob]` - this value is used in tests

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: I don't remember actually - we did this when we mobbed on this stuff originally to handle requester pays. I *think* it only applies for requester pays workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Oh good idea!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Yes we can!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Good idea

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: I can only point you to the akka doc. I have no clue. Many behaviors. Such confuse. https://doc.akka.io/docs/akka/current/typed/actors.html

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: No, I need to return a `Behavior` it seems and akka seems to indicate `same`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: No better or worse. Non-determinism is encoded in `Seq` here. it's actually been changed in #1611.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: Good idea.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: The way it's designed is that you can get that information by joining the migration table and the job table on migration id

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: You don't need to know about catagory theory lol (though it'll no doubt give a deeper understanding for this stuff).
Play around with some simple examples and get a practical intuition! Most of what seems like magic is simply syntactic sugar for lambdas.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
ehigham: https://github.com/broadinstitute/rawls/blob/aa92209cc57e155345d5ff904565978c8e82293e/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala#L562-L595

ehigham:
> do we know why the jenkins test startup failed?

Pass on re-run

andy7i (approved):
this is awesome work! thanks again for all your patience and help with going through everything

PR #1609, opened 2022-01-24
liquibase full sharding migration [AJ-19]

davidangb:
I am marking this PR as a draft to ensure it does not merge, but I would like code review on it.

We do not want to let this code reach production until *after* we have manually run the parallelized migration code. If this reaches production it will cause a ~7 hr liquibase migration, while the parallel code cuts that down significantly.

Once the parallel code has run, the migrations in this PR will be noops and fast. It is here for completeness, to ensure that any environment can be created/managed in a fully automated fashion and we don't _require_ the parallel code; we only _prefer_ the parallel code because it is faster.

We will need a future PR, sometime after all the migration completes, to remove the sharded vs. unsharded code paths in `AttributeComponent.determineShard()` and elsewhere.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
@@ -43,7 +43,7 @@ trait DataAccess
       val second = secondLong.toHexString
       determineShard(java.util.UUID.fromString(s"$first${second}000000-0000-0000-0000-000000000000"), shardState = WorkspaceShardStates.Sharded).toString
     }
-  }).toSet + determineShard(java.util.UUID.fromString(s"00000000-0000-0000-0000-000000000000"), shardState = WorkspaceShardStates.Unsharded).toString
+  }).toSet
 
   // only called from TestDriverComponent
   def truncateAll: WriteAction[Int] = {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
```diff
@@ -43,7 +43,7 @@ trait DataAccess
       val second = secondLong.toHexString
       determineShard(java.util.UUID.fromString(s"$first${second}000000-0000-0000-0000-000000000000"), shardState = WorkspaceShardStates.Sharded).toString
     }
-  }).toSet + determineShard(java.util.UUID.fromString(s"00000000-0000-0000-0000-000000000000"), shardState = WorkspaceShardStates.Unsharded).toString
+  }).toSet
```
davidangb, line 46: this prevents the `DataAccess` class - and thus Slick - from testing the existence of the `ENTITY_ATTRIBUTE_archived` table.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityShardingSpec.scala
```diff
@@ -72,7 +72,7 @@ class EntityShardingSpec extends AnyFlatSpec with Matchers
   // checks row counts for each shard, plus the _archived table
   def checkShardCounts(expected: Map[ShardId, Int] = Map()): Unit = {
     // default to 0 for all shards plus the archive table
-    val default: Map[ShardId, Int] = (allShards + "archived").map(_ -> 0).toMap
+    val default: Map[ShardId, Int] = (allShards).map(_ -> 0).toMap
```
davidangb, line 75: and this prevents this specific test class from trying to test the `ENTITY_ATTRIBUTE_archived` table

jordansafer (approved):

sagehen03 (approved):

PR #1619, opened 2022-02-01
[BW-278] Retry many transient reasons for submissions failing

cjllanwarne:
Note: Do not merge until [BW-1056](https://broadworkbench.atlassian.net/browse/BW-1056) is completed and a new Cromwell version is released.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
@@ -43,11 +43,14 @@ trait DsdeHttpDAO extends LazyLogging {
 
   protected def pipeline[A](implicit um: Unmarshaller[ResponseEntity, A]) = executeRequest[A] _
 
-  protected def when500(throwable: Throwable ): Boolean = {
+  protected def when5xx(throwable: Throwable ): Boolean = DsdeHttpDAO.when5xx(throwable)
+}
+
+object DsdeHttpDAO {
+  def when5xx(throwable: Throwable ): Boolean = {
     throwable match {
       case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(_.intValue/100 == 5)
       case _ => false
     }
   }
-
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
@@ -14,7 +14,7 @@ import org.broadinstitute.dsde.rawls.util.HttpClientUtils
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
- * Created by dvoet on 9/2/15.
+ * Wraps HTTP methods utility in a DAO-like wrapper. Helps interactions with HTTP endpoints within Terra (DSDE being the old name for DSP)
  */
 trait DsdeHttpDAO extends LazyLogging {
   protected implicit val system: ActorSystem
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.rawls.dataaccess
 import akka.actor.ActorSystem
 import akka.http.scaladsl.HttpExt
 import akka.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken}
-import akka.http.scaladsl.model.{HttpHeader, HttpRequest, ResponseEntity}
+import akka.http.scaladsl.model.{HttpHeader, HttpRequest, ResponseEntity, StatusCode}
 import akka.http.scaladsl.unmarshalling.Unmarshaller
 import akka.stream.Materializer
 import com.typesafe.scalalogging.LazyLogging
@@ -44,13 +44,17 @@ trait DsdeHttpDAO extends LazyLogging {
   protected def pipeline[A](implicit um: Unmarshaller[ResponseEntity, A]) = executeRequest[A] _
 
   protected def when5xx(throwable: Throwable ): Boolean = DsdeHttpDAO.when5xx(throwable)
+
+
 }
 
 object DsdeHttpDAO {
-  def when5xx(throwable: Throwable ): Boolean = {
-    throwable match {
-      case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(_.intValue/100 == 5)
-      case _ => false
-    }
+
+  private def statusCodePredicate(check: StatusCode => Boolean): Throwable => Boolean = {
+    case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(check)
+    case _ => false
   }
+
+  def when5xx: Throwable => Boolean = statusCodePredicate(_.intValue/100 == 5)
+  def whenUnauthorized: Throwable => Boolean = statusCodePredicate(_.intValue == 401)
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
@@ -52,7 +52,7 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
 
     val inputsJsonArray = inputs.mkString("[", ",", "]")
 
-    // We only map over inputs to make this list the same size. The input values themselves is ignored.
+    // Map over inputs to make this list the same size, but ihe inputs themselves are ignored.
     val requestedWorkflowIdsJsonArray = inputs.map(_ => s""""${UUID.randomUUID().toString}""""  ).mkString("[", ",", "]")
 
     val bodyParts = Seq(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
@@ -56,12 +56,12 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
 
     val formData = Multipart.FormData(bodyParts:_*)
 
-    pipeline[Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]]](userInfo) apply (Post(url, Marshal(formData).to[RequestEntity]))
+    retry(when5xx) { () => pipeline[Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]]](userInfo) apply (Post(url, Marshal(formData).to[RequestEntity])) }
   }
 
   override def status(id: String, userInfo: UserInfo): Future[ExecutionServiceStatus] = {
     val url = executionServiceURL + s"/api/workflows/v1/${id}/status"
-    retry(when500) { () => pipeline[ExecutionServiceStatus](userInfo) apply Get(url) }
+    retry(when5xx) { () => pipeline[ExecutionServiceStatus](userInfo) apply Get(url) }
   }
 
   // break out uri generation into a separate method so it's easily unit-testable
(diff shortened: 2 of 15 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
```diff
@@ -43,11 +43,18 @@ trait DsdeHttpDAO extends LazyLogging {
...
-    throwable match {
-      case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(_.intValue/100 == 5)
-      case _ => false
-    }
+  protected def when5xx(throwable: Throwable ): Boolean = DsdeHttpDAO.when5xx(throwable)
+
+
+}
+
+object DsdeHttpDAO {
```
cahrens, line 51: I don't know what "Dsde" and "DAO" stand for. Perhaps rename this with more a more explicit name, or add documentation?

cahrens: Sigh, I now see that this is in a file named "DsdeHttpDAO". So renaming is probably out of the question, but some documentation on the trait would be nice explaining what this is (and remove "Created by dvoet on 9/2/15.").

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
```diff
@@ -43,11 +43,18 @@ trait DsdeHttpDAO extends LazyLogging {
...
+
+  private def statusCodePredicate(check: StatusCode => Boolean): Throwable => Boolean = {
+    case t: RawlsExceptionWithErrorReport => t.errorReport.statusCode.exists(check)
+    case _ => false
   }
 
+  def when5xx: Throwable => Boolean = statusCodePredicate(_.intValue/100 == 5)
```
cahrens, line 58: So the implementation of this is actually the same as "when500", but the name has been changed to be more accurate?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -426,25 +426,25 @@ class HttpGoogleServicesDAO(
...
-  override def getBucket(bucketName: String, userProject: Option[GoogleProjectId])(implicit executionContext: ExecutionContext): Future[Option[Bucket]] = {
+  override def getBucket(bucketName: String, userProject: Option[GoogleProjectId])(implicit executionContext: ExecutionContext): Future[Either[String, Bucket]] = {
     implicit val service = GoogleInstrumentedService.Storage
-    retryWithRecoverWhen500orGoogleError(() => {
+    retryWithRecoverWhen500orGoogleError[Either[String, Bucket]](() => {
       val getter = getStorage(getBucketServiceAccountCredential).buckets().get(bucketName)
       userProject.map( p => getter.setUserProject(p.value) )
 
-      Option(executeGoogleRequest(getter))
+      Right(executeGoogleRequest(getter))
```
cahrens, line 435: So the Right returns the Bucket variable? But what if it's just a String (doesn't Either main it can be either a String or a Bucket?)?

cahrens: Oh, I guess this is setting the return value and saying it is a Bucket?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -426,25 +426,25 @@ class HttpGoogleServicesDAO(
...
     }) {
-      case _: HttpResponseException => None
+      case e: HttpResponseException => Left(s"HTTP ${e.getStatusCode}: ${e.getStatusMessage} (${e.getMessage})")
     }
   }
 
   override def getRegionForRegionalBucket(bucketName: String, userProject: Option[GoogleProjectId]): Future[Option[String]] = {
     getBucket(bucketName, userProject) map {
-      case Some(bucket) => bucket.getLocationType match {
+      case Right(bucket) => bucket.getLocationType match {
```
cahrens, line 443: Should the return type have changed to be String or Bucket?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -42,13 +43,7 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
...
   private def asRawlsSAPipeline[A](implicit um: Unmarshaller[ResponseEntity, A]) = executeRequestWithToken[A](OAuth2BearerToken(getServiceAccountAccessToken)) _
 
-  protected def when401or500(throwable: Throwable): Boolean = {
-    throwable match {
-      case t: RawlsExceptionWithErrorReport =>
-        t.errorReport.statusCode.exists(status => (status.intValue/100 == 5) || status.intValue == 401)
-      case _ => false
-    }
-  }
+  protected def when401or500: Predicate[Throwable] = anyOf(when5xx, DsdeHttpDAO.whenUnauthorized)
```
cahrens: Seems like this should be renamed `when401or5xx`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -248,7 +243,7 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
 
   override def getPetServiceAccountKeyForUser(googleProject: GoogleProjectId, userEmail: RawlsUserEmail): Future[String] = {
     val url = samServiceURL + s"/api/google/v1/petServiceAccount/${googleProject.value}/${URLEncoder.encode(userEmail.value, UTF_8.name)}"
-    retry(when401or500) { () => asRawlsSAPipeline[String] apply RequestBuilding.Get(url) }
+    retryUntilSuccessOrTimeout(when401or500)(interval = 5.seconds, timeout = 55.seconds) { () => asRawlsSAPipeline[String] apply RequestBuilding.Get(url) }
```
cahrens: How did you choose these values?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
```diff
@@ -215,8 +215,8 @@ class HealthMonitor private (val slickDataSource: SlickDataSource,
     // Note: call to `foldMap` depends on SubsystemStatusMonoid
     bucketsToCheck.toList.foldMap { bucket =>
       googleServicesDAO.getBucket(bucket, None).map {
-        case Some(_) => OkStatus
-        case None => failedStatus(s"Could not find bucket: $bucket")
+        case Right(_) => OkStatus
```
cahrens, line 218: Should the return type be changed? I'm sorry, but I think I'm still not understanding this Right vs. Left thing.

core/src/test/scala/org/broadinstitute/dsde/rawls/DsdeHttpDAOSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+  behavior of "when5xx"
+
+  it should "match 500" in {
+    val e = new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, ""))
+    DsdeHttpDAO.when5xx(e) should be(true)
+  }
+
+  it should "match 503" in {
+    val e = new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.ServiceUnavailable, ""))
+    DsdeHttpDAO.when5xx(e) should be(true)
```
cahrens: Would be good to have something that does not match.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
aednichols: DSDE is an [old organizational unit](https://docs.google.com/spreadsheets/d/1wn_XFAf54USob1UMqbk9SkxBgDXUV7UoY3fr24LVLeY/edit#gid=0) "Data Sciences and Data Engineering" that has since rolled up into DSP, that can definitely go away. [DAO](https://en.wikipedia.org/wiki/Data_access_object) is an at-least somewhat standard pattern.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -49,19 +50,25 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
       case wdl: WdlSource => Multipart.FormData.BodyPart("workflowSource", wdl.source)
     }
 
+    val inputsJsonArray = inputs.mkString("[", ",", "]")
+
+    // Map over inputs to make this list the same size, but ihe inputs themselves are ignored.
```
aednichols:
```suggestion
    // Map over inputs to make this list the same size, but the inputs themselves are ignored.
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -49,19 +50,25 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
...
+    val requestedWorkflowIdsJsonArray = inputs.map(_ => s""""${UUID.randomUUID().toString}""""  ).mkString("[", ",", "]")
```
aednichols, line 56: Since we're now submitting known workflow IDs to Cromwell, does it make sense to pre-populate database entries into the Rawls workflows table?

The code as-is will work fine, but it seems a bit odd that Rawls will generate the ID, send it to Cromwell, hear it echoed back, and then finally register it as a monitored workflow.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -49,19 +50,25 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
...
       wdlSourceOrUrl,
-      Multipart.FormData.BodyPart("workflowInputs", inputs.mkString("[", ",", "]"))
+      Multipart.FormData.BodyPart("workflowInputs", inputsJsonArray),
+      Multipart.FormData.BodyPart("requestedWorkflowId", requestedWorkflowIdsJsonArray)
     ) ++ options.map(Multipart.FormData.BodyPart("workflowOptions", _)) ++ labelsBodyPart ++ wfc
 
     val formData = Multipart.FormData(bodyParts:_*)
 
-    pipeline[Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]]](userInfo) apply (Post(url, Marshal(formData).to[RequestEntity]))
+    retry(anyOf(when5xx, DsdeHttpDAO.whenUnauthorized)) { () => pipeline[Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]]](userInfo) apply (Post(url, Marshal(formData).to[RequestEntity])) }
```
aednichols, line 66: Yessss

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
cjllanwarne: Yeah so this change is adding a [companion object](https://docs.scala-lang.org/overviews/scala-book/companion-objects.html) to this pre-existing `DsdeHttpDAO` class. I'll try to write my own personal interpretation of what this class is for, would welcome corrections though

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
cjllanwarne: Correct. It confused me with its original name

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cjllanwarne: `Either` is often used in a scala convention to mean "bad thing or good thing", with `Right` being the "right" thing and `Left` being the "wrong" thing. In this case, the Bucket is the "right" think and the `String` is the wrong thing - ie an error message.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cjllanwarne: This is converting `Either`ness into `Future`ness by unpacking the "right"ness of a bucket name into a successful future and the "wrong"ness of the string into an unsuccessful `Future`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
cjllanwarne: The previous retry tries up to 3 times at 100ms, 1 second and 3 seconds, but that wasn't wasn't spaced out enough to recover the occasional hiccups in this call. So I extended the timeout to just under a minute and increased the spacing to happen every 5 seconds.

So - slightly arbitrary but it made the transient errors go away.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala
cjllanwarne: The problem I had before was that the `Some` vs `None` was losing a lot of context. The `Either` allows me to carry around the context as a "wrong" (or `Left`) string rather than just the "didn't work" of the `None`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
cjllanwarne: I think... potentially yes that would be nice, with a few caveats -

1. I think users are used to thinking "workflow ID means it's been started" which we'd have to re-teach
2. The UI thinks "workflow ID means I can show the Job Manager/Workflow Dashboard link", which we'd want to change
3. More difficult than I was hoping for in this PR/

I can make a follow up ticket if you think it's worth it though?

aednichols: OK so long as it's what you have in mind I don't see a problem with it, and no need to add followup work (that we likely won't ever do)

aednichols (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DsdeHttpDAO.scala
cahrens: Thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cahrens: Got it, thanks!

cahrens:
There are some concerning test failures. Otherwise this LGTM.

cahrens (approved):
Approving conditionally based on tests passing after Cromwell merge

PR #1632, opened 2022-02-14
[CA-1776] Expose `GET` workspaces/{namespace}/{name}/migrate

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1776
Add endpoint to return a list of workspace migration attempt metadata.
These objects represent a simplified view of a migration attempt.
Add json support to utility classes.

Spec: https://broadworkbench.atlassian.net/wiki/spaces/CI/pages/2358149130/CA-1776+GET+Migration+Status

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
@@ -74,10 +74,20 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
       } ~
-      path("workspaces" / Segment / Segment / "migrate") { (workspaceNamespace, workspaceName) =>
+      path("workspaces" / Segment / Segment / "migrate") { (namespace, name) =>
+        val workspaceName = WorkspaceName(namespace, name)
+        get {
+          complete {
+            workspaceServiceConstructor(userInfo)
+              .getWorkspaceMigrationAttempts(workspaceName)
+              .map(ms => StatusCodes.OK -> ms)
+          }
+        } ~
         put {
           complete {
-            workspaceServiceConstructor(userInfo).migrateWorkspace(WorkspaceName(workspaceNamespace, workspaceName)).map(_ => StatusCodes.NoContent)
+            workspaceServiceConstructor(userInfo)
+              .migrateWorkspace(workspaceName)
+              .map(_ => StatusCodes.NoContent)
           }
         }
       } ~
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
@@ -74,7 +74,7 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
       } ~
-      path("workspaces" / Segment / Segment / "migrate") { (namespace, name) =>
+      path("workspaces" / Segment / Segment / "migrations") { (namespace, name) =>
         val workspaceName = WorkspaceName(namespace, name)
         get {
           complete {
@@ -83,7 +83,7 @@ trait WorkspaceApiService extends UserInfoDirectives {
               .map(ms => StatusCodes.OK -> ms)
           }
         } ~
-        put {
+        post {
           complete {
             workspaceServiceConstructor(userInfo)
               .migrateWorkspace(workspaceName)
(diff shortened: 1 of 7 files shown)


gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -74,10 +74,20 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
       } ~
-      path("workspaces" / Segment / Segment / "migrate") { (workspaceNamespace, workspaceName) =>
+      path("workspaces" / Segment / Segment / "migrate") { (namespace, name) =>
```
andy7i: the API being a `GET  ... migrate` is a bit weird to me. 

I wonder if there's something else we can do? Maybe renaming `migrate` to `migration`?

```suggestion
      path("workspaces" / Segment / Segment / "migration") { (namespace, name) =>
```

then it works with GET and PUT

andy7i (commented):
I'm assuming that nobody (or at least nobody improper) has the permission for `SamWorkspaceActions.migrate`

I didn't see any swagger documentation for the migrate APIs. Do we want to add it to swagger or keep the APIs unlisted?

ehigham:
> I'm assuming that nobody (or at least nobody improper) has the permission for `SamWorkspaceActions.migrate`
> 
> I didn't see any swagger documentation for the migrate APIs. Do we want to add it to swagger or keep the APIs unlisted?

Yeah, I have not included swagger doc deliberately as I want this api to be undocumented - as far as use cases are concerned, we're just using it for tests and we'll delete it when all workspaces are migrated. If there's a use case to expose it to users then we can consider that, I just don't want to commit to maintaining an API that's temporary in nature.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
ehigham: I agree in part but `PUT` migration seems a bit strange too. `POST` might be more appropriate in that case but is `POST` with an empty body ok?

andy7i: throwing out there that we could go with a more verbose option by creating an endpoint for `GET ... migrationStatus`.
I think since this is an undocumented API, sticking to one endpoint might make things easier for us. I think I still prefer `migration` over `migrate`, but I also don't feel super strongly about it.

andy7i (changes requested):
I just took a look at https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L130 and it looks like we grant `migrate` to all workspace owners. That doesn't seem right and I don't think we should expose this API until we figure this out

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
ehigham: I changed to POST and renamed to `migrations`

ehigham:
> I just took a look at https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L130 and it looks like we grant `migrate` to all workspace owners. That doesn't seem right and I don't think we should expose this API until we figure this out

Why does that not seem right? Why shouldn't anyone who can create a workspace migration attempt get these attempts?

ehigham:
> > I just took a look at https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L130 and it looks like we grant `migrate` to all workspace owners. That doesn't seem right and I don't think we should expose this API until we figure this out
> 
> Why does that not seem right? Why shouldn't anyone who can create a workspace migration attempt get these attempts?

As discussed in the mob - our plan is to remove `migrate` from workspace owners for the time being and grant it to admins so we can test the actor. I'll create a PR in sam for this. No further changes on this subject are required here.

andy7i (approved):
(After some discussion:) Looks like the PUT is already out there so it doesn't make sense to block this PR due to the permission. 

Let's create a ticket to figure out the right way we want "feature flag" this, but that shouldn't block this PR.

Side note: Did we get AppSec's blessing for removing the feature flag on these endpoints?

ehigham:
> (After some discussion:) Looks like the PUT is already out there so it doesn't make sense to block this PR due to the permission.
> 
> Let's create a ticket to figure out the right way we want "feature flag" this, but that shouldn't block this PR.

Created https://broadworkbench.atlassian.net/browse/CA-1782

> Side note: Did we get AppSec's blessing for removing the feature flag on these endpoints?

Removing the feature flag?

PR #1625, opened 2022-02-09
update mysql-connector-java [AJ-273]

davidangb:
Updates the version of `mysql-connector-java`, with corresponding syntax changes. This addresses proactive security vulnerability scan findings.

Note that AJ-273 also mentions `jersey-core-common`, I will handle that separately to keep PRs concise.

This PR uses connector version 8.0.28 against our 5.6 server. There was a question about compatibility; https://dev.mysql.com/downloads/connector/j/ states "MySQL Connector/J 8.0 is highly recommended for use with MySQL Server 8.0, 5.7 and 5.6. Please upgrade to MySQL Connector/J 8.0." 

The connector update required Scala syntax changes regarding exceptions. I used https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-exceptions-changes.html to guide the syntax changes.

PR #1634, opened 2022-02-15
update mysqlconnector [AJ-273]

davidangb:
See also: #1625 , #1633. The original PR #1625 to update `mysql-connector-java` passed all tests, including fiab/swatomation tests, but failed when deploying in argo/k8s. I reverted those changes in #1633 to unblock deploys.

The failure deploying in argo/k8s was due to a bug that happened while processing liquibase changesets. This bug only happens when processing changesets on top of an already-initialized database; it does not occur when processing changesets from scratch on an empty db. Thus, the bug did not previously appear in unit tests or fiabs, because those start from an empty db.

This PR:
* updates `mysql-connector-java` to a liquibase-compatible version
* updates the unit test framework to catch the bug in question

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import com.typesafe.config.ConfigFactory
+import com.typesafe.scalalogging.LazyLogging
 import nl.grons.metrics4.scala.{Counter, DefaultInstrumented, MetricName}
 import org.broadinstitute.dsde.rawls.TestExecutionContext
 import org.broadinstitute.dsde.rawls.config.WDLParserConfig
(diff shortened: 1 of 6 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -44,6 +45,9 @@ object DbResource {
 
   val dataSource = new SlickDataSource(dataConfig)(TestExecutionContext.testExecutionContext)
   dataSource.initWithLiquibase(liquibaseChangeLog, Map.empty)
+  logger.info("executing liquibase a second time to verify changesets work on an already-initialized database ...")
+  dataSource.initWithLiquibase(liquibaseChangeLog, Map.empty)
+
```
davidangb, line 50: before this change, the `mysql-connector-java`/`liquibase-core` bug did not show up in unit tests, because unit tests always used a fresh empty db, which was initialized once with liquibase. Now, we run liquibase twice - the second liquibase run simulates what happens in prod when we redeploy Rawls on top of the existing prod db.

sagehen03: I'm a little confused.  Now we're making the bug show up in tests, right?  But doesn't that mean that tests will fail?  What have we done to address the updated mysql-connector-java-liquibase unhappiness?

davidangb: the versions of `mysql-connector-java` and `liquibase-core` in this PR are compatible, so the bug will not appear. Note that this PR uses `mysql-connector-java` 8.0.22, where the original attempt used 8.0.28.

jordansafer (approved):

sagehen03 (approved):

PR #1636, opened 2022-02-17
always use entity metadata cache [AJ-310]

davidangb:
PR summary:
* define two workspace feature flags "alwaysCacheTypeCounts" and "alwaysCacheAttributes".
* if "alwaysCacheTypeCounts" is set for a workspace, we will always read the **_list of distinct types and their counts_** from cache when returning entity type statistics in metadata, even if the cache is out of date.
* if "alwaysCacheAttributes" is set for a workspace, we will always read the **_list of distinct attribute names for each type_** from cache when returning entity type statistics in metadata, even if the cache is out of date. This is the one that I expect to have the most benefit to end-user latency.
* additional tracing throughout the metadata API to assist with future performance debugging

This PR has one functional change above and beyond the feature-flagged "always cache" functionality: if cache is disabled at the system level via https://github.com/broadinstitute/rawls/blob/a5bb0ae387c54d40ae29c72680f4304dec01023d/core/src/main/resources/reference.conf#L83, we previously would _not_ opportunistically save metadata results to the cache. After this PR, we do. I'm fine with this change because I expect that we will never have cache disabled at the system level, and in fact we should probably just remove that setting in a future PR.

See also broadinstitute/firecloud-develop#2815

TODOs:
- [x] deal with the two now-unused methods (only used in unit tests). This could wait for a future PR if they are troublesome but let's not forget about them!
- [x] validate the new unit tests: are they asserting the right things? are they passing? do we need any more coverage?
- [x] test test test test, stand up the branch and set some flags and see how it's working

thoughts for later:
- [ ] maybe, add config to reference.conf to control default caching behavior in addition to the per-workspace flags
- [ ] if a metrics sink is easily available, use it instead of logging staleness

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
@@ -11,18 +11,23 @@ import java.sql.Timestamp
 import scala.concurrent.ExecutionContext
 import scala.util.{Failure, Success}
 
+/**
+  * helper methods that deal with entity type metadata and its cache
+  */
 trait EntityStatisticsCacheSupport extends LazyLogging {
 
   implicit protected val executionContext: ExecutionContext
-
-  val workspaceContext: Workspace
   protected val dataSource: SlickDataSource
+  val workspaceContext: Workspace
 
   import dataSource.dataAccess.driver.api._
 
-  /** convenience method for querying and then assembling an entity type metadata response
-    *
-    * see also getEntityTypeMetadata() in EntityComponent
+  private val FEATURE_ALWAYS_CACHE_TYPE_COUNTS = "alwaysCacheTypeCounts"
+  private val FEATURE_ALWAYS_CACHE_TYPE_ATTRIBUTES = "alwaysCacheAttributes"
+
+  /** convenience method for querying and then assembling an entity type metadata response.
+    * if this method retrieves metadata directly from entities and attributes (i.e. not from cache),
+    * it will update the cache with the results it found.
     * */
   def calculateMetadataResponse(dataAccess: DataAccess,
                                         countsFromCache: Boolean,
@@ -88,11 +93,11 @@ trait EntityStatisticsCacheSupport extends LazyLogging {
       // TODO: use actual feature flag methods once that PR merges. Will look something like:
       /*
       dataAccess.workspaceFeatureFlagQuery.listFlagsForWorkspace(workspaceContext.workspaceIdAsUUID,
-        List("alwaysCacheTypeCounts", "alwaysCacheAttributes")).flatMap { foundFlags =>
+        List(FEATURE_ALWAYS_CACHE_TYPE_COUNTS, FEATURE_ALWAYS_CACHE_TYPE_ATTRIBUTES)).flatMap { foundFlags =>
 
         val flagMap: Map[String, Boolean] = foundFlags.map { flag => (flag.flagname -> flag.enabled)}
-        val alwaysCacheTypeCounts = flagMap.getOrElse("alwaysCacheTypeCounts", false)
-        val alwaysCacheAttributes = flagMap.getOrElse("alwaysCacheAttributes", false)
+        val alwaysCacheTypeCounts = flagMap.getOrElse(FEATURE_ALWAYS_CACHE_TYPE_COUNTS, false)
+        val alwaysCacheAttributes = flagMap.getOrElse(FEATURE_ALWAYS_CACHE_TYPE_ATTRIBUTES, false)
         CacheFeatureFlags(alwaysCacheTypeCounts = alwaysCacheTypeCounts, alwaysCacheAttributes = alwaysCacheAttributes)
       }
       */
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
@@ -101,7 +101,7 @@ trait EntityStatisticsCacheSupport extends LazyLogging {
         CacheFeatureFlags(alwaysCacheTypeCounts = alwaysCacheTypeCounts, alwaysCacheAttributes = alwaysCacheAttributes)
       }
       */
-      DBIO.successful(CacheFeatureFlags(alwaysCacheTypeCounts = false, alwaysCacheAttributes = true))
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/util/OpenCensusDBIOUtils.scala
```diff
@@ -19,4 +19,16 @@ object OpenCensusDBIOUtils {
...
+    traceReadOnlyDBIOSpan(startSpanWithParent(name, parentSpan), failureStatus)(f)
+
+  def traceReadOnlyDBIO[T](name: String, failureStatus: Throwable => Status = (_: Throwable) => Status.UNKNOWN)(f: Span => ReadAction[T])(implicit executor: ExecutionContext): ReadAction[T] =
+    traceReadOnlyDBIOSpan(startSpan(name), failureStatus)(f)
+
+  private def traceReadOnlyDBIOSpan[T](span: Span, failureStatus: Throwable => Status)(f: Span => ReadAction[T])(implicit executor: ExecutionContext): ReadAction[T] =
+    f(span).cleanUp { _ =>
+      endSpan(span, Status.OK)
+      DBIO.successful(0)
+    }
```
davidangb, line 33: these mirror the "traceDBIO*" pre-existing methods, but return `ReadAction`s instead of `ReadWriteAction`s. This helps avoid changes in method signatures that want to add tracing

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -28,79 +27,58 @@ import scala.util.{Failure, Success, Try}
...
+              // cache is up to date - return cached
+              logger.info(s"entity statistics cache: hit [${workspaceContext.workspaceIdAsUUID}]")
+              calculateMetadataResponse(dataAccess, countsFromCache = true, attributesFromCache = true, rootSpan)
+            case Some(_) =>
+              // cache exists, but is out of date - check if this workspace has any always-cache feature flags set
+              cacheFeatureFlags(dataAccess, rootSpan).flatMap { flags =>
+                if (flags.alwaysCacheTypeCounts || flags.alwaysCacheAttributes) {
+                  rootSpan.putAttribute("alwaysCacheTypeCountsFeatureFlag", OpenCensusAttributeValue.booleanAttributeValue(flags.alwaysCacheTypeCounts))
+                  rootSpan.putAttribute("alwaysCacheAttributesFeatureFlag", OpenCensusAttributeValue.booleanAttributeValue(flags.alwaysCacheAttributes))
+                  logger.info(s"entity statistics cache: partial hit (alwaysCacheTypeCounts=${flags.alwaysCacheTypeCounts}, alwaysCacheAttributes=${flags.alwaysCacheAttributes}) [${workspaceContext.workspaceIdAsUUID}]")
```
davidangb: we are logging how stale the cache is inside of `EntityStatisticsCacheSupport .cacheStaleness()`. However, the place I really care about staleness is here, for the partial hits. When we intentionally serve stale data because one or more "alwaysCache*" feature flags are set, how stale is the data we are serving?

jordansafer:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
```diff
@@ -0,0 +1,145 @@
...
+  def calculateMetadataResponse(dataAccess: DataAccess,
+                                        countsFromCache: Boolean,
+                                        attributesFromCache: Boolean,
+                                        outerSpan: Span = null): ReadWriteAction[Map[String, EntityTypeMetadata]] = {
+    val typesAndCountsQ = typeCounts(dataAccess, countsFromCache, outerSpan)
+    val typesAndAttrsQ = typeAttributes(dataAccess, attributesFromCache, outerSpan)
+    traceDBIOWithParent[Map[String, EntityTypeMetadata]]("generateEntityMetadataMap", outerSpan) { innerSpan =>
+      dataAccess.entityQuery.generateEntityMetadataMap(typesAndCountsQ, typesAndAttrsQ).flatMap { metadata =>
+        // and opportunistically save, if we have bypassed cache for all components
+        val saveCacheAction = if (countsFromCache || attributesFromCache) {
```
sagehen03, line 43: It looks like we could turn on one of the cache feature flags and not the other and that could result in not updating the cache that was not set to always use the cached value even though we went through the trouble of running the query?  Is that correct?  If so it's probably not worth worrying about because the background cache updater should pick things up anyhow, but it's a little bit of an inconsistency and one more wrinkle to include when thinking about this process.

sagehen03 (approved):
I have one question/comment, but I think it's an edge case.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
davidangb: yes, your assessment is absolutely correct (and good catch reverse-engineering all that!). You already saw broadinstitute/firecloud-develop#2815, which should help reduce that length-of-staleness.

I'd prefer to save the intricacies of opportunistically updating half the cache for some other PR, when/if we realize we need it. I stared at it a bit here and didn't have a good solution for it. The big complexity I want to be careful about is that any code we add in this spot is already inside a ` dataSource.inTransaction` closure … so I want to make absolutely certain that any operations we add don't extend the db transaction longer than necessary.

TL;DR I think this is thorny and I'm putting it off

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
sagehen03: Thinking about this a little--still assuming I've got the right interpretation--couldn't we just change the logical or to an and to make the code opportunistically save any time we've recalculated the attribute or entity count parts of the cache?

davidangb: if only one flag was set, it means we pulled half the metadata from potentially-outdated cache, and the other half from "real". If we then saved this result back to the cache, we'd be marking the cache as updated - even though half of it might be outdated. I see what you're getting at - updating half the cache is still progress - but I am wary about then declaring the cache as up-to-date. I'm sticking to my stance of believing this needs further thought, not wanting to change it inside this PR.

PR #1638, opened 2022-02-22
workspace feature flags [AJ-309]

davidangb:
AJ-309: a generic feature-flag mechanism for workspaces, to make feature flagging easier.

Adds a table to the DB and some Scala/Slick helper functions.

Summary of new unit tests:
```
[info] WorkspaceFeatureFlagComponentSpec:
[info] WorkspaceFeatureFlagComponent
[info] - should save individual flags via save()
[info] - should error when inserting pre-existing flags via save()
[info] - should batch-save multiple flags via saveAll()
[info] - should error when inserting pre-existing flags via saveAll()
[info] - should return an empty list when listing all flags for a workspace if none exist
[info] - should return all flags when listing all flags for a workspace
[info] - should return an empty list when listing specific flags for a workspace if none exist
[info] - should return an empty list when listing specific flags for a workspace if those flags do not exist
[info] - should return the requested flags when listing specific flags for a workspace
[info] - should insert non-existent flags via saveOrUpdate
[info] - should update pre-existing flags via saveOrUpdate
[info] - should insert non-existent flags via saveOrUpdateAll
[info] - should update pre-existing flags via saveOrUpdateAll
[info] - should delete all flags for a workspace
[info] - should noop when deleting all flags for a workspace if none exist
[info] - should delete specific flags for a workspace
[info] Run completed in 31 seconds, 339 milliseconds.
[info] Total number of tests run: 16
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 16, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```


davidangb:
hm, looks like a transient failure in fiab-start:
```
15:35:37 ERROR: for cromiam-proxy  Cannot start service cromiam-proxy: driver failed programming external connectivity on endpoint firecloud_cromiam-proxy_1 (bc230a2b5b4672df8a8c58bb7919f38d8f8c44eac8a66ced5e7c59b21d6369b0): Error starting userland proxy: listen tcp4 0.0.0.0:33480: bind: address already in use
```
Jenkins retest.

sagehen03 (commented):
It seems like this would work.  What would a check for a feature flag look like?  `listFlagsForWorkspace(wid,  flagName).filterForEnabled()` ish?  I don't have super strong opinions but I'll throw out a few thoughts to see if any resonate with you.

- Do we need the enabled and config columns in the db?  Would it be enough just to check for the workspace and the feature flag name? Do we already have a scenario in which we need multiple values for a feature flag?  If we don't yet, I say let's wait until we do.
- In the past I've often seen magic string values read into enums, do you plan to do that?  It would give the benefit of type checking and avoiding spelling bugs.  
- I've also seen bit masks used as feature flags.  This has the advantage of only needing one row per workspace with the disadvantage of adding bit masks, bit shifting and so forth.
- It would be nice to combine this code with the code for optionally always fetching from cache in one PR.

jordansafer (commented):
I have a few high level thoughts/questions:
- Is there an easy way to enable/disable a flag for all workspaces? This is a really great functionality of flags, since we can quickly rollback a feature without any code change.
- Putting out a thought about using a JSON object store a workspace config with flags per workspace, and having a JSON model/object that defines what flags we expect to exist at any given time. If Rawls is also mysql5.6 (is it?), then we can't just do this-
- When would we use the saveAll/saveOrUpdateAll functionality for flags? I definitely was thinking about flags as per feature/per sub feature, is this something we would use to toggle flags across a large number of workspaces? (ie all CDC workspaces)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceFeatureFlagComponent.scala
```diff
@@ -0,0 +1,65 @@
...
+trait WorkspaceFeatureFlagComponent {
+  this: DriverComponent =>
+
+  import driver.api._
+
+  class WorkspaceFeatureFlagTable(tag: Tag) extends Table[WorkspaceFeatureFlagRecord](tag, "WORKSPACE_FEATURE_FLAG") {
+    def workspaceId = column[UUID]("workspace_id", O.PrimaryKey)
+    def flagName = column[String]("flagname", O.PrimaryKey, O.Length(100))
+    def enabled = column[Boolean]("enabled")
+    def config = column[Option[String]]("config", O.Length(4000))
```
jordansafer: Is the config column intended as a JSONish substitute?

davidangb:
@jordansafer @sagehen03 let's see if I can hit all the questions/comments:
1. I could easily drop the config column and only add it if/when we realize we need it, or need something like it. I'm guessing at future use cases and how they might want to store a value alongside the enable/disable boolean - something like "the entitystore-per-workspace flag is enabled with a value of 'postgres'" or "the entitystore-per-workspace flag is enabled with a value of 'elasticsearch'". But I definitely don't have clarity here.
2. re: "In the past I've often seen magic string values read into enums, do you plan to do that" … I understand about magic strings and enums, but how are you suggesting that fits into this framework? I'm not certain what use case you're asking about here. If this is about the config column, then sure - whoever implements a feature flag could parse the value of the config column into an enum.
3. re: bitmasks … I've also seen this and it's super helpful for high performance and/or large volume databases … at the expense of making it hard for humans to debug. I don't think we need to go there yet. We have 40k workspaces in prod, even if each of those had multiple flags set at once that's still not significant volume. My opinion is bitmasking is premature optimization but I'm open to being swayed if you think it's important.
4. re: combining this and the always-cache code into one PR … I could definitely do that if you think it's worthwhile, I usually opt for making smaller targeted PRs. Would this make it easier to understand how feature flags get used?
5. In the code so far, there is no way to enable/disable a flag for all workspaces. I envision that whoever/whatever team implements a flag for a given feature would set a default. I could also easily see follow-on PRs for additional functions or API endpoints like this.
6. re: using JSON to store flags - sure, that's an option especially if we upgrade the db
7. re: when to use saveAll/saveOrUpdateAll - I'm picturing two flags for the cache feature, "alwaysCacheAttributes" and "alwaysCacheCounts", I can imagine a future PR that turns on an API that allows an admin or an end user to set both flags in one API call. If this is premature I can take these methods out.

davidangb:
@sagehen03 if it is helpful, here is a draft of what it could look like for the metadata cache to call feature flags methods:
https://github.com/broadinstitute/rawls/blob/61e40acd381ccca32640b623bacfdd7452007f0c/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala#L88-L98

sagehen03:
> @sagehen03 if it is helpful, here is a draft of what it could look like for the metadata cache to call feature flags methods:
> 
> https://github.com/broadinstitute/rawls/blob/61e40acd381ccca32640b623bacfdd7452007f0c/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala#L88-L98

My preference would be to use enums so that misspelling "alwaysCacheTypeCounts" or "alwaysCacheAttributes" doesn't result in a bug.

davidangb:
@sagehen03 agreed - and sorry, that example is just a draft, it's not at all ready for review, I was just trying to illustrate how the feature flag methods would get called

sagehen03:
> @jordansafer @sagehen03 let's see if I can hit all the questions/comments:
> 
>     1. I could easily drop the config column and only add it if/when we realize we need it, or need something like it. I'm guessing at future use cases and how they might want to store a value alongside the enable/disable boolean - something like "the entitystore-per-workspace flag is enabled with a value of 'postgres'" or "the entitystore-per-workspace flag is enabled with a value of 'elasticsearch'". But I definitely don't have clarity here.
> 
>     2. re: "In the past I've often seen magic string values read into enums, do you plan to do that" … I understand about magic strings and enums, but how are you suggesting that fits into this framework? I'm not certain what use case you're asking about here. If this is about the config column, then sure - whoever implements a feature flag could parse the value of the config column into an enum.

> 
>     3. re: bitmasks … I've also seen this and it's super helpful for high performance and/or large volume databases … at the expense of making it hard for humans to debug. I don't think we need to go there yet. We have 40k workspaces in prod, even if each of those had multiple flags set at once that's still not significant volume. My opinion is bitmasking is premature optimization but I'm open to being swayed if you think it's important.
> 
>     4. re: combining this and the always-cache code into one PR … I could definitely do that if you think it's worthwhile, I usually opt for making smaller targeted PRs. Would this make it easier to understand how feature flags get used?
> 
>     5. In the code so far, there is no way to enable/disable a flag for all workspaces. I envision that whoever/whatever team implements a flag for a given feature would set a default. I could also easily see follow-on PRs for additional functions or API endpoints like this.
> 
>     6. re: using JSON to store flags - sure, that's an option especially if we upgrade the db
> 
>     7. re: when to use saveAll/saveOrUpdateAll - I'm picturing two flags for the cache feature, "alwaysCacheAttributes" and "alwaysCacheCounts", I can imagine a future PR that turns on an API that allows an admin or an end user to set both flags in one API call. If this is premature I can take these methods out.



> @jordansafer @sagehen03 let's see if I can hit all the questions/comments:
> 
>     1. I could easily drop the config column and only add it if/when we realize we need it, or need something like it. I'm guessing at future use cases and how they might want to store a value alongside the enable/disable boolean - something like "the entitystore-per-workspace flag is enabled with a value of 'postgres'" or "the entitystore-per-workspace flag is enabled with a value of 'elasticsearch'". But I definitely don't have clarity here.
I vote to wait until we need those columns.
> 
>     2. re: "In the past I've often seen magic string values read into enums, do you plan to do that" … I understand about magic strings and enums, but how are you suggesting that fits into this framework? I'm not certain what use case you're asking about here. If this is about the config column, then sure - whoever implements a feature flag could parse the value of the config column into an enum.
I am talking about the flagname column specifically but also the config value too I suppose.  Basically, I'm saying any time there's a new feature flag there should be a new enum entry, from your sample usage comment, it sounds like we're on the same page here.
> 
>     3. re: bitmasks … I've also seen this and it's super helpful for high performance and/or large volume databases … at the expense of making it hard for humans to debug. I don't think we need to go there yet. We have 40k workspaces in prod, even if each of those had multiple flags set at once that's still not significant volume. My opinion is bitmasking is premature optimization but I'm open to being swayed if you think it's important.
I don't feel strongly.  I think it depends on how we ultimately use this table, but I wanted to put this approach on your radar.
> 
>     4. re: combining this and the always-cache code into one PR … I could definitely do that if you think it's worthwhile, I usually opt for making smaller targeted PRs. Would this make it easier to understand how feature flags get used?
I would prefer to combine into one PR.
> 
>     5. In the code so far, there is no way to enable/disable a flag for all workspaces. I envision that whoever/whatever team implements a flag for a given feature would set a default. I could also easily see follow-on PRs for additional functions or API endpoints like this.
> 
>     6. re: using JSON to store flags - sure, that's an option especially if we upgrade the db
> 
>     7. re: when to use saveAll/saveOrUpdateAll - I'm picturing two flags for the cache feature, "alwaysCacheAttributes" and "alwaysCacheCounts", I can imagine a future PR that turns on an API that allows an admin or an end user to set both flags in one API call. If this is premature I can take these methods out.

jordansafer:
@davidangb, all sounds good, and +1 on keeping the PRs separate. Just re the enable-disable for all workspaces, the idea is that there is a global flag/flag entree, so we can disable/change without going through a deployment. +1 on having defaults for flags, but this would be separate, since we of course can't turn off the default :) . I think if we're putting together a flagging mechanism, we should keep global flagging in mind, unless we already have another mechanism to handle that?

For 7, I don't feel strongly, I think of the cost of ownership of new APIs vs flexibility, and it's not much of a concern in either direction-

davidangb:
@jordansafer @sagehen03 up for re-review, please and thank you!
* `enabled` and `config` db columns are now gone
* added `WorkspaceFeatureFlag` model class for strong typing; we're not tossing `String`s around everywhere
* changed signatures in `workspaceFeatureFlagQuery` so all methods are workspace-specific

sagehen03 (approved):

jordansafer (approved):

PR #1643, opened 2022-02-28
admin APIs for getting/setting feature flags [AJ-310]

davidangb:
Adds admin APIs for getting and setting feature flags for a given workspace.


jordansafer (approved):
Looks good! I was wondering about flag APIs during testing yesterday :)

PR #1645, opened 2022-03-01
SU-22: Fix error when deleting failed v1 billing projects with underlying Google projects that Rawls didn't create

MatthewBemis:
This typically happened when there was a naming collision, since GCP projects are a global namespace. Tested it with a few failed projects that b.adm.firec had access to

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -884,7 +884,7 @@ class HttpGoogleServicesDAO(
   override def getGoogleProject(googleProject: GoogleProjectId): Future[Project] = {
     implicit val service = GoogleInstrumentedService.Billing
     val cloudResManager = getCloudResourceManagerWithBillingServiceAccountCredential
-    retryExponentially(throwable => when500orExcludedGoogleError(throwable, Set(StatusCodes.NotFound, StatusCodes.Forbidden)))(() =>
+    retryExponentially(throwable => when500orNonExcludedGoogleError(throwable, Set(StatusCodes.NotFound, StatusCodes.Forbidden)))(() =>
       Future(blocking(executeGoogleRequest(cloudResManager.projects().get(googleProject.value))))
     )
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -884,7 +884,7 @@ class HttpGoogleServicesDAO(
   override def getGoogleProject(googleProject: GoogleProjectId): Future[Project] = {
     implicit val service = GoogleInstrumentedService.Billing
     val cloudResManager = getCloudResourceManagerWithBillingServiceAccountCredential
-    retryExponentially(when500orNon404GoogleError)(() =>
+    retryExponentially(throwable => when500orExcludedGoogleError(throwable, Set(StatusCodes.NotFound, StatusCodes.Forbidden)))(() =>
       Future(blocking(executeGoogleRequest(cloudResManager.projects().get(googleProject.value))))
     )
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -884,7 +884,8 @@ class HttpGoogleServicesDAO(
   override def getGoogleProject(googleProject: GoogleProjectId): Future[Project] = {
     implicit val service = GoogleInstrumentedService.Billing
     val cloudResManager = getCloudResourceManagerWithBillingServiceAccountCredential
-    retryExponentially(throwable => when500orNonExcludedGoogleError(throwable, Set(StatusCodes.NotFound, StatusCodes.Forbidden)))(() =>
+    val statusCodesExcludedFromRetry: Set[StatusCode] = Set(StatusCodes.NotFound, StatusCodes.Forbidden)
+    retryExponentially(throwable => when500orGoogleError(throwable) || whenGoogleStatusDoesntContain(throwable, statusCodesExcludedFromRetry))(() =>
       Future(blocking(executeGoogleRequest(cloudResManager.projects().get(googleProject.value))))
     )
   }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -371,7 +371,7 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
     def googleProjectExists(projectId: GoogleProjectId) =
```
nawatts: Would it be worth renaming this function? Since it now returns false for projects that do exist but can't be accessed? Though since it looks like it's local to `deleteGoogleProjectIfChild`, the potential for misunderstanding is probably low.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -40,6 +41,11 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
     case t => when500orGoogleError(t)
   }
 
+  def when500orNonExcludedGoogleError(throwable: Throwable, excludedStatusCodes: Set[StatusCode]): Boolean = throwable match {
```
nawatts: Would it read more clearly to inline this in `getGoogleProject` instead of creating a function? The double negative in the name is a bit difficult to parse. This function also has the potential to be confusing if someone passed 500 in the excluded status codes.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -884,7 +884,7 @@ class HttpGoogleServicesDAO(
   override def getGoogleProject(googleProject: GoogleProjectId): Future[Project] = {
     implicit val service = GoogleInstrumentedService.Billing
     val cloudResManager = getCloudResourceManagerWithBillingServiceAccountCredential
-    retryExponentially(when500orNon404GoogleError)(() =>
+    retryExponentially(throwable => when500orNonExcludedGoogleError(throwable, Set(StatusCodes.NotFound, StatusCodes.Forbidden)))(() =>
```
MatthewBemis: Background:

Terra billing projects used to be created with Google projects that had a matching name in GCP. But this namespace is global, so if someone else in the world had create a Google project called "hello-world", and then someone in Terra tried to create a project called "hello-world", it would fail. Users want to be able to delete these projects, because they've collected a bunch of them over the years and they serve no value. The existing code would attempt to get the Google project so it could delete it. But in this case, Rawls doesn't actually have access to that Google project -- because Rawls didn't create it. It would retry exponentially until the call timed out. This change quickly surfaces the 403 instead of retrying, and then other code has been changed to proceed with the rest of the billing project deletion if there's a 403.

nawatts (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: Good call, renamed it. Let me know if I made it worse 😂

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
MatthewBemis: Ah yeah, this and the other similar predicates are a bit messy. I will put some thought into ways to clean this up.

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -21,31 +22,45 @@ import scala.util.{Failure, Success, Try}
 trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInstrumented {
   implicit val executionContext: ExecutionContext
 
-  protected def when500orGoogleError(throwable: Throwable): Boolean = {
+  protected def when500or400orGoogleError(throwable: Throwable): Boolean = {
```
MatthewBemis, line 25: I renamed the existing `when500orGoogleError` to `when500or400orGoogleError` because the name was not accurate- it would retry both 5xx and 4xx status codes.

I ended up creating an actual `when500orGoogleError`, which does what it says

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -368,10 +368,10 @@ class UserService(protected val userInfo: UserInfo, val dataSource: SlickDataSou
 
   // TODO - once workspace migration is complete and there are no more v1 workspaces or v1 billing projects, we can remove this https://broadworkbench.atlassian.net/browse/CA-1118
   private def deleteGoogleProjectIfChild(projectName: RawlsBillingProjectName, userInfoForSam: UserInfo, deleteGoogleProjectWithGoogle: Boolean = true) = {
-    def googleProjectExists(projectId: GoogleProjectId) =
+    def rawlsCreatedGoogleProjectExists(projectId: GoogleProjectId) =
```
MatthewBemis, line 371: renamed because `googleProjectExists` is not entirely accurate with my change, which really just checks if a rawls-created google project exists

nawatts (approved):

google/src/main/scala/org/broadinstitute/dsde/rawls/google/GoogleUtilities.scala
```diff
@@ -21,31 +22,45 @@ import scala.util.{Failure, Success, Try}
...
   // into a default retrying handler. This ship has sailed - it's used everywhere and changing it
   // might cause a bunch of failures.
   def when500orNon404GoogleError(throwable: Throwable): Boolean = throwable match {
     case e: HttpResponseException if e.getStatusCode == 404 => false
-    case t => when500orGoogleError(t)
+    case t => when500or400orGoogleError(t)
+  }
+
+  //Predicate to exclude certain Google status codes from retries
+  def whenGoogleStatusDoesntContain(throwable: Throwable, excludedStatusCodes: Set[StatusCode]): Boolean = throwable match {
```
ehigham, line 53: Maybe you could curry this to avoid the call site looking grim: 
```scala 
def whenGoogleStatusDoesntContain(excludedStatusCodes: Set[StatusCode])(t: Throwable): Boolean
```

Then your call site is 
```scala 
retryExponentially(whenGoogleStatusCodeDoesntContain( ... )) { ... }
```

ehigham (approved):
Thank you. Appreciate the renaming - it makes it clearer.
One suggestion but I defer to you if you want to do it.

PR #1647, opened 2022-03-02
list-entity and entity-query safety limits [AJ-244]

davidangb:
Changes in this PR:
* enforce a hard cap on number of entities returned by the list-entities and entity-query APIs
* set the cap to 300,000. I specifically chose this number to not cause any errors to current users, after looking at the entity counts in our top workspaces and who is using the list-entities API.
* read the cap from configuration, so we could easily update it later by adding/updating a stanza in rawls.conf

the entity-query API simply looks at the pageSize parameter for the request and rejects if the pageSize is too high. This requires no database queries.

the list-entities API adds a `select count()` db query. If the result of this query is too high, it throws an error, else it continues on to the "real" query it already had in place. I verified by running Rawls locally that Slick generates this new query correctly. An example:
```sql
select count(1) from `ENTITY` where ((`entity_type` = 'mfjqsybnzsvcbtvozps') and (`workspace_id` = ?)) and (not `deleted`)
```

Since we increased the heap size for Rawls, we have been less in danger of OOMs (see graph below). So, it feels safe to set a high cap of 300,000 to protect us against pathological future cases, while still not interrupting current behaviors.

Screenshot (37)

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -86,6 +86,10 @@ entityStatisticsCache {
   workspaceCooldown = 4 minutes
 }
 
+entities {
+  pageSizeLimit = 300000
+}
+
 akka.http.host-connection-pool.max-open-requests = 16384
 akka.http.host-connection-pool.max-connections = 20
 akka.http.server.idle-timeout = 210 s
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -166,7 +166,7 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       metadataFuture.recover(bigQueryRecover)
     }
 
-  def listEntities(workspaceName: WorkspaceName, entityType: String, parentSpan: Span = null): Future[Seq[Entity]] = {
+  def listEntities(workspaceName: WorkspaceName, entityType: String, parentSpan: Span = null): Future[Seq[Entity]] =
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
       dataSource.inTransaction { dataAccess =>
         traceDBIOWithParent("countActiveEntitiesOfType", parentSpan) { countSpan =>
@@ -185,7 +185,6 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
         }
       }
     }
-  }
 
   def queryEntities(workspaceName: WorkspaceName, dataReference: Option[DataReferenceName], entityType: String, query: EntityQuery, billingProject: Option[GoogleProjectId], parentSpan: Span = null): Future[EntityQueryResponse] = {
     if (query.pageSize > pageSizeLimit) {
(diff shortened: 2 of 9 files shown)


core/src/main/resources/reference.conf
```diff
@@ -86,6 +86,10 @@ entityStatisticsCache {
   workspaceCooldown = 4 minutes
 }
 
+entities {
+  pageSizeLimit = 300000
+}
+
```
davidangb, line 92: if we want to change this later, we can copy this stanza into rawls.conf in firecloud-develop, instead of changing it here.

davidangb:
the `rawls-build` job was hung after 14 hours. Hopefully this is transient. Jenkins retest.

davidangb:
and, now we have a failure in fiab-start:
```
09:22:16 Registering user Vincent Crabbe
09:22:16 POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 503
09:22:16 [WARN] Retrying...
09:22:16 [WARN] Retried POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 409
09:22:16 [WARN] Retrying...
09:22:16 [WARN] Retried POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 409
09:22:16 [WARN] Retrying...
09:22:16 [WARN] Retried POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 409
09:22:16 [WARN] Retrying...
09:22:16 [WARN] Retried POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 409
09:22:16 [WARN] Retrying...
09:22:16 [WARN] Retried POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile: 409
09:22:16 [ERROR] POST https://firecloud-orchestration-fiab.dsde-qa.broadinstitute.org:23443/register/profile failed with status 409
09:22:16 {"causes":[{"causes":[],"message":"user vincent.crabbe@quality.firecloud.org already exists","source":"sam","stackTrace":[],"statusCode":409}],"message":"user vincent.crabbe@quality.firecloud.org already exists","source":"Sam","stackTrace":[],"statusCode":409}
```
one more time: Jenkins retest.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -166,18 +166,32 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       metadataFuture.recover(bigQueryRecover)
     }
 
-  def listEntities(workspaceName: WorkspaceName, entityType: String, parentSpan: Span = null): Future[Seq[Entity]] =
+  def listEntities(workspaceName: WorkspaceName, entityType: String, parentSpan: Span = null): Future[Seq[Entity]] = {
```
jordansafer: Nit: Do we need to add the braces `{}` around the function body here?

jordansafer (approved):
Looks good! I think 300k is very generous given that we want users to move to the paginated API. After we make this change, it would be interesting to see a metric on what the largest calls are to this API, and use this as a mechanism to migrate any holdout callers with big datasets.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: good catch! No, we do not need the braces. (I had previously made, then backed out, a change that required the braces … so they're just leftover cruft)

jordansafer (approved):

davidangb:
rawls-build hung again. huh? jenkins retest.

PR #1654, opened 2022-03-03
[WOR-175, WOR-176] SpendReportingService refactoring

marctalbott:
Ticket: [WOR-175](https://broadworkbench.atlassian.net/browse/WOR-175), [WOR-176](https://broadworkbench.atlassian.net/browse/WOR-176)
* `HttpGoogleBigQueryDAO` is a workbench-libs library that uses the older style of Google client library known as the [API Client Library](https://cloud.google.com/apis/docs/client-libraries-explained#google-api-client-libraries). Google has been slowly replacing these libraries with more idiomatic libraries known as [Cloud Client libraries](https://cloud.google.com/apis/docs/client-libraries-explained#cloud-client-libraries). In workbench-libs, we keep the Cloud Client libraries in the [`google2` directory](https://github.com/broadinstitute/workbench-libs/tree/develop/google2). Extracting the BigQuery results using the `HttpGoogleBigQueryDAO` was difficult and brittle, so I've refactored the `SpendReportingService` to use the `BigQueryService` which uses Google's Cloud Client library.
* Also included in this PR is an improvement to the error message that we return when our query to BigQuery returns no rows.
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -3,13 +3,15 @@ package org.broadinstitute.dsde.rawls.spendreporting
 import java.util.Currency
 
 import akka.http.scaladsl.model.StatusCodes
-import com.google.api.services.bigquery.model._
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+import com.google.cloud.bigquery.{Option => _, _}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.config.SpendReportingServiceConfig
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
 import org.broadinstitute.dsde.rawls.model._
-import org.broadinstitute.dsde.workbench.google.GoogleBigQueryDAO
+import org.broadinstitute.dsde.workbench.google2.GoogleBigQueryService
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.joda.time.format.ISODateTimeFormat
 import org.joda.time.{DateTime, Days}
@@ -19,14 +21,14 @@ import scala.jdk.CollectionConverters._
 import scala.math.BigDecimal.RoundingMode
 
 object SpendReportingService {
-  def constructor(dataSource: SlickDataSource, bigQueryDAO: GoogleBigQueryDAO, samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
+  def constructor(dataSource: SlickDataSource, bigQueryService: cats.effect.Resource[IO, GoogleBigQueryService[IO]], samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
                  (userInfo: UserInfo)
                  (implicit executionContext: ExecutionContext): SpendReportingService = {
-    new SpendReportingService(userInfo, dataSource, bigQueryDAO, samDAO, spendReportingServiceConfig)
+    new SpendReportingService(userInfo, dataSource, bigQueryService, samDAO, spendReportingServiceConfig)
   }
 }
 
-class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, bigQueryDAO: GoogleBigQueryDAO, samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
+class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, bigQueryService: cats.effect.Resource[IO, GoogleBigQueryService[IO]], samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
                            (implicit val executionContext: ExecutionContext) extends LazyLogging {
   private def requireProjectAction[T](projectName: RawlsBillingProjectName, action: SamResourceAction)(op: => Future[T]): Future[T] = {
     samDAO.userHasAction(SamResourceTypeNames.billingProject, projectName.value, action, userInfo).flatMap {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -19,14 +21,14 @@ import scala.jdk.CollectionConverters._
 import scala.math.BigDecimal.RoundingMode
 
 object SpendReportingService {
-  def constructor(dataSource: SlickDataSource, bigQueryDAO: GoogleBigQueryDAO, samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
+  def constructor(dataSource: SlickDataSource, bigQueryService: cats.effect.Resource[IO, GoogleBigQueryService[IO]], samDAO: SamDAO, spendReportingServiceConfig: SpendReportingServiceConfig)
```
cahrens, line 24: Why does this need to be wrapped with `cats.effect.Resource` (I don't know much about cats yet)?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -42,27 +44,27 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
   }
 
-  def extractSpendReportingResults(rows: List[TableRow], startTime: DateTime, endTime: DateTime): SpendReportingResults = {
+  def extractSpendReportingResults(rows: List[FieldValueList], startTime: DateTime, endTime: DateTime): SpendReportingResults = {
     val currency = getCurrency(rows)
 
     val dailySpend = rows.map { row =>
-      val rowCost = BigDecimal(row.getF.get(0).getV.toString).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
-      val rowCredits = BigDecimal(row.getF.get(1).getV.toString).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
+      val rowCost = BigDecimal(row.get("cost").getDoubleValue).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
```
cahrens, line 51: **SO** much nicer!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -64,8 +64,7 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
                       parsedStartDate,
                       parsedEndDate
                     ).map {
-                      case Some(spendReportResults) => StatusCodes.OK -> Option(spendReportResults)
-                      case None => StatusCodes.NotFound -> None
+                      sr => StatusCodes.OK -> sr
```
cahrens, line 67: Can you explain what this is doing?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -163,18 +165,19 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
-          result: GetQueryResultsResponse <- bigQueryDAO.getQueryResult(jobStatus)
+          queryJobConfiguration = QueryJobConfiguration
+            .newBuilder(query)
+            .addNamedParameter("billingAccountId", stringQueryParameterValue(spendExportConf.billingAccountId.withoutPrefix()))
+            .addNamedParameter("startDate", stringQueryParameterValue(dateTimeToISODateString(startDate)))
+            .addNamedParameter("endDate", stringQueryParameterValue(dateTimeToISODateString(endDate)))
+            .addNamedParameter("projects", stringArrayQueryParameterValue(workspaceProjects.map(_.value).toList))
+            .build()
+
+          result <- bigQueryService.use(_.query(queryJobConfiguration)).unsafeToFuture()
```
cahrens, line 176: Why `unsafeToFuture`?

cahrens (approved):
This API definitely seems better. I just have a few questions as a Scala newbie.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: This is how the BigQuery client library is implemented in `workbench-libs` and a `Resource` is basically a functional programming way of safely managing something that may need to be frequently acquired and then released or closed (like a `File`).

To actually make use of it, you can call `.use` on a `Resource`. `use` gives you a `GoogleBigQueryService` and then expects you to return an `IO[B]` where `B` is whatever you want it to be. So we can make use of our `Resource` as follows

``` 
bigQueryService.use { service: GoogleBigQueryService[IO] => 
    service.query() 
}
``` 

which will return whatever `service.query()` returns (in this case `IO[TableResult]`)

You can read more about cats resources here https://typelevel.org/cats-effect/docs/std/resource if you're interested!

marctalbott: Also, I find reading cats-effect stuff to be really difficult and unwelcoming, so definitely let me know if you'd like to talk through any of it since I've found that to be more helpful at times.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: `bigQueryService.use(_)` returns an `IO`, but this for comprehension requires that everything on the right side of the arrows be a `Future` (this is because the right side of the first line returns a `Future`). `unsafeToFuture` converts the `IO` to a `Future`. 

`unsafe` is the cats-effect way of indicating that whatever is in the `IO` is no longer contained within the "safe" confines of `IO`. In this case, that's because we're converting it to a `Future`. You'll also commonly see `unsafeRunSync` which is how you can force the evaluation of the `IO` and get its results. You can read more about it here: https://typelevel.org/cats-effect/docs/2.x/datatypes/io#introduction

And same offer as above, let me know if you want to talk through any cats-effect stuff at any point (or maybe cats-effect could be a good tech talk topic...)

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
marctalbott: On closer inspection, I think this is actually unnecessary and I'll remove it pending passing tests, but I'll explain why I thought it was needed in the first place.

`sr` is an instance of `SpendReportingResults`, and I wanted to make sure that the correct HTTP status code (200 in this case) was returned along with `sr`. `StatusCodes.{code} -> {result}` is one way of specifying the HTTP status code. However, it seems that Akka (or some other Scala magic, I'm not sure which) will default to a 200 and return your result without doing this.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
marctalbott: I forgot to remove this unnecessary code from this branch, but it'll be in the next spend reporting PR for per-workspace breakdowns

PR #1656, opened 2022-03-04
update WSM client to latest; deal with removed model classes [AJ-326]

davidangb:
https://github.com/DataBiosphere/terra-workspace-manager/pull/570 removed deprecated APIs from Workspace Manager. Along with those removals came removal of model classes inside the WSM client. Rawls was using those now-removed model classes in a bunch of places.

This PR:
* updates to the latest version of WSM client in both runtime and automation
* changes some code away from the removed model classes and onto supported model classes
* creates new, local to Rawls, case classes to represent some of the now-removed WSM models. These are necessary because while WSM dropped their deprecated APIs, Rawls still supports its "v1" snapshot APIs. In order to not change request/response payloads for those APIs, I have created replacement classes.

We should absolutely consider just removing those v1 APIs from Rawls, but we need to verify they are not used.

**EDIT:** I just checked, and it looks like the v1 APIs haven't been called in the last 90 days. I am going to work towards removing them, which should simplify this PR a lot.


davidangb:
closing, this is superseded by #1659

PR #1657, opened 2022-03-06
Remove v1 snapshot apis [AJ-328]

davidangb:
AJ-328

The v1 APIs were deprecated and superseded back in May 2021: https://github.com/broadinstitute/rawls/pull/1432.

Now, we're dropping them entirely. Dropping them eases the migration to the latest Workspace Manager client library, which has removed the model classes previously used by the v1 APIs.

I searched in Logit/Kibana and do not see ANY requests to the v1 APIs in the last 90 days. The query I'm using is:
```
request:/\/api\/workspaces\/[^\/]+\/[^\/]+\/snapshots.*/ AND NOT v2
```


jordansafer (approved):
Looks good!!

PR #1658, opened 2022-03-07
Filter out "pet service account not found" error before sending events to Sentry [BW-1125]

salonishah11:
This PR filters out the error at source and doesn't send it to Sentry. This way the user will still see/receive the error but Sentry won't receive it.

Closes https://broadworkbench.atlassian.net/browse/BW-1125

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala
@@ -20,6 +20,17 @@ class RawlsSentryEventFilterSpec extends AnyFreeSpec {
       }
     }
 
+    // https://broadworkbench.atlassian.net/browse/BW-1125
+    "should filter out 'pet service account not found' errors" in {
+      val e = evt("org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport: ErrorReport(rawls,HTTP error calling URI " +
+        "https://sam/api/google/petServiceAccount/abc@terra.iam.gserviceaccount.com. Response: {\"causes\":[]," +
+        "\"message\":\"pet service account not found\",\"source\":\"sam\",\"stackTrace\":[],\"statusCode\":404}," +
+        "Some(404 Not Found),List(),List(),None)")
+      assertResult(false) {
+        new RawlsSentryEventFilter().shouldSend(e)
+      }
+    }
+
     "should NOT filter any other messages" in {
       val e = evt("This is some other error message that should be sent to Sentry")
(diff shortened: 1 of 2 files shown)


salonishah11:
jenkins retest

salonishah11:
jenkins retest

aednichols (approved):
I don't have a problem with the code as it is now, but as discussed I am also open to mobbing on a solution where we refactor to make this case use something other than thrown exceptions to communicate failure.

core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala
```diff
@@ -20,6 +20,17 @@ class RawlsSentryEventFilterSpec extends AnyFreeSpec {
       }
     }
 
+    // https://broadworkbench.atlassian.net/browse/BW-1125
+    "should filter out 'pet service account not found' errors" in {
```
cjllanwarne, line 24: Do we want:
 - To test the other message type being checked for too? 
 - A "control" test where we make sure anything else gets through?
 - A test where an error is thrown to test the Try matching?

cjllanwarne (approved):
LGTM, one "TOL" level comment

core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala
salonishah11: So there is a [test to check the other message is being filtered out](https://github.com/broadinstitute/rawls/blob/9ec1589582550fcf352a2122c1addd69761f9f4a/core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala#L16-L21) and [test to make sure other messages are not being filtered out](https://github.com/broadinstitute/rawls/blob/9ec1589582550fcf352a2122c1addd69761f9f4a/core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala#L34-L39). Is this what you were thinking about? 

  >A test where an error is thrown to test the Try matching?
  
So a test where we would throw and error [here](https://github.com/broadinstitute/rawls/blob/9ec1589582550fcf352a2122c1addd69761f9f4a/core/src/test/scala/org/broadinstitute/dsde/rawls/util/RawlsSentryEventFilterSpec.scala#L30) instead of calling the function `shouldSend()`?

cjllanwarne: LGTM as is

PR #1659, opened 2022-03-07
update WSM client to latest; deal with removed model classes - take 2 [AJ-326]

davidangb:
https://github.com/DataBiosphere/terra-workspace-manager/pull/570 removed deprecated APIs from Workspace Manager. Along with those removals came removal of model classes inside the WSM client. Rawls was using those now-removed model classes in a bunch of places.

This PR:
* updates to the latest version of WSM client in both runtime and automation
* changes some code away from the removed model classes and onto supported model classes

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -146,12 +146,15 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
     SnapshotListResponse(refs)
   }
 
-  def updateSnapshot(workspaceName: WorkspaceName, snapshotId: String, updateInfo: UpdateDataReferenceRequestBody): Future[Unit] = {
+  def updateSnapshot(workspaceName: WorkspaceName, snapshotId: String, updateInfo: UpdateDataRepoSnapshotReferenceRequestBody): Future[Unit] = {
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       // check that snapshot exists before updating it. If the snapshot does not exist, the GET attempt will throw a 404
       workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
-      // build the update request body
+      // build the update request body, ignoring any changes to instanceName and snapshot, and requiring either name or description
+      if (Option(updateInfo.getName).isEmpty && Option(updateInfo.getDescription).isEmpty) {
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Either name or description is required."))
+      }
       val updateBody = new UpdateDataRepoSnapshotReferenceRequestBody()
       updateBody.setName(updateInfo.getName)
       updateBody.setDescription(updateInfo.getDescription)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiService.scala
@@ -45,9 +45,9 @@ trait SnapshotApiService extends UserInfoDirectives {
         }
       } ~
       patch {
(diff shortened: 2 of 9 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -146,12 +146,15 @@ class SnapshotService(protected val userInfo: UserInfo, val dataSource: SlickDat
...
+  def updateSnapshot(workspaceName: WorkspaceName, snapshotId: String, updateInfo: UpdateDataRepoSnapshotReferenceRequestBody): Future[Unit] = {
     val snapshotUuid = validateSnapshotId(snapshotId)
     getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))).map { workspaceContext =>
       // check that snapshot exists before updating it. If the snapshot does not exist, the GET attempt will throw a 404
       workspaceManagerDAO.getDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, userInfo.accessToken)
-      // build the update request body
+      // build the update request body, ignoring any changes to instanceName and snapshot, and requiring either name or description
+      if (Option(updateInfo.getName).isEmpty && Option(updateInfo.getDescription).isEmpty) {
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Either name or description is required."))
+      }
```
davidangb, line 157: `UpdateDataReferenceRequestBody`, which no longer exists in the WSM client, had only "name" and "description". `UpdateDataRepoSnapshotReferenceRequestBody`, which we're using now, has "name", "description", "instanceName", and "snapshot". I added this clause so our behavior is the same as previous. We don't have a use case currently for updating the instanceName or snapshot.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiService.scala
```diff
@@ -63,25 +63,6 @@ trait SnapshotApiService extends UserInfoDirectives {
...
-          complete {
-            snapshotServiceConstructor(userInfo).updateSnapshot(WorkspaceName(workspaceNamespace, workspaceName), snapshotId, updateDataReferenceRequestBody).map(_ => StatusCodes.NoContent)
-          }
-        }
-      } ~
-      delete {
-        complete {
-          snapshotServiceConstructor(userInfo).deleteSnapshot(WorkspaceName(workspaceNamespace, workspaceName), snapshotId).map(_ => StatusCodes.NoContent)
-        }
-      }
```
davidangb, line 84: apparently we defined these twice? As a copy/paste error? This block is exactly the same as lines 41-59, so I removed it.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala
```diff
@@ -20,77 +20,6 @@ object DataReferenceModelJsonSupport extends JsonSupport {
...
-    override def read(json: JsValue): DataReferenceList = {
-      json.asJsObject.getFields(RESOURCES) match {
-        case Seq(JsArray(resources)) =>
-          new DataReferenceList().resources(resources.map(_.convertTo[DataReferenceDescription]).asJava)
-        case _ => throw DeserializationException("DataReferenceList expected")
-      }
-    }
-  }
-
-
```
davidangb, line 93: the model classes referenced by these de/serializers are no longer used or no longer exist

model/src/test/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModelSpec.scala
```diff
@@ -29,31 +27,6 @@ class DataReferenceModelSpec extends AnyFreeSpec with Matchers {
...
-          ).asJava).toJson
-        }
-      }
-
-      "DataReferenceDescription with bad UUID's should fail" in {
-        assertThrows[DeserializationException] {
-          s"""{"referenceId": "abcd","name":"test-ref","workspaceId":"abcd","referenceType":"$DATA_REPO_SNAPSHOT","reference":{"instanceName":"test-instance","snapshot":"test-snapshot"},"cloningInstructions":"$NOTHING"}""".parseJson.convertTo[DataReferenceDescription]
-        }
-      }
-
```
davidangb, line 56: since the de/serializers no longer exist, these unit tests should also be deleted

sagehen03 (approved):

PR #1660, opened 2022-03-07
[WOR-150] per-workspace spend reporting breakdown

marctalbott:
Ticket: [WOR-150](https://broadworkbench.atlassian.net/browse/WOR-150)

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -499,7 +499,7 @@ paths:
             format: date
         - name: aggregationKey
           in: query
-          description: Key indicating how to aggregate spend data in results
+          description: Key indicating how to aggregate spend data in results. Will return only summary data (spendDetails is empty) if aggregationKey is not present
           required: false
           schema:
             type: string
@@ -6354,6 +6354,11 @@ components:
           type: string
           format: timestamp
           description: timestamp (UTC) marking the end of reported spend range (YYYY-MM-DDThh:mm:ss.fffZ)
+        googleProjectId:
+          type: string
+          description: if present, Google project of workspace where cost was generated
+        workspace:
+          $ref: '#/components/schemas/WorkspaceName'
     SpendReportConfiguration:
       required:
         - datasetGoogleProject
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -6310,12 +6310,9 @@ components:
       type: object
       properties:
         aggregationKey:
-          type: object
+          type: string
+          enum: [Total, Workspace]
           description: Key indicating how spendData has been aggregated. Ex. 'workspace' if all data in spendData is for a particular workspace
-          properties:
-            key:
-              type: string
-              description: key
         spendData:
           type: array
           items:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -497,6 +497,15 @@ paths:
           schema:
             type: string
             format: date
+        - name: aggregationKey
+          in: query
+          description: Key indicating how to aggregate spend data in results
+          required: false
+          schema:
+            type: string
+            enum:
+              - Daily
+              - Workspace
       responses:
         200:
           description: Success
@@ -6311,7 +6320,9 @@ components:
       properties:
         aggregationKey:
           type: string
-          enum: [Total, Workspace]
+          enum:
+            - Daily
+            - Workspace
           description: Key indicating how spendData has been aggregated. Ex. 'workspace' if all data in spendData is for a particular workspace
         spendData:
           type: array
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
@@ -1,9 +1,12 @@
 package org.broadinstitute.dsde.rawls.model
 
-import org.broadinstitute.dsde.workbench.model.google.{BigQueryDatasetName, GoogleProject}
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.model.SpendReportingAggregationKeys.SpendReportingAggregationKey
 import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport._
+import org.broadinstitute.dsde.workbench.model.google.{BigQueryDatasetName, GoogleProject}
 import org.joda.time.DateTime
 import spray.json.DefaultJsonProtocol._
+import spray.json.{DeserializationException, JsString, JsValue, RootJsonFormat}
 
 case class BillingProjectSpendConfiguration(datasetGoogleProject: GoogleProject, datasetName: BigQueryDatasetName)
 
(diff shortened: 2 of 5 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -497,6 +497,15 @@ paths:
           schema:
             type: string
             format: date
+        - name: aggregationKey
+          in: query
+          description: Key indicating how to aggregate spend data in results
+          required: false
```
cahrens, line 503: If the aggregation key is not present, will you only get the summary data?

marctalbott: At the moment it will default to `Daily`, but I could change it to only return summary data

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -143,44 +173,62 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
+                                                         |  SUM(cost) as cost,
+                                                         |  SUM(IFNULL((SELECT SUM(c.amount) FROM UNNEST(credits) c), 0)) as credits,
+                                                         |  currency,
+                                                         |  project.id as googleProjectId
+                                                         | FROM `$tableName`
+                                                         | WHERE billing_account_id = @billingAccountId
+                                                         | AND _PARTITIONTIME BETWEEN @startDate AND @endDate
+                                                         | AND project.id in UNNEST(@projects)
+                                                         | GROUP BY currency, googleProjectId
+                                                         |""".stripMargin
```
cahrens: It seems like the only difference here is the Group By parameter. Would it make sense to have the string as a variable (plugging in the parameter) to avoid duplication and ensure that they queries do not get out of sync?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -67,16 +99,33 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with TestDriverComponent
...
+
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(workspace1))
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(workspace2))
+
+    val reportingResults = Await.result(service.getSpendForBillingProject(testData.billingProject.projectName, DateTime.now().minusDays(1), DateTime.now(), SpendReportingAggregationKeys.Workspace), Duration.Inf)
+    reportingResults.spendSummary.cost shouldBe "0.10" // sum of costs in defaultTable
+    val workspaceAggregation = reportingResults.spendDetails.headOption.getOrElse(fail("workspace results not parsed correctly"))
+
+    workspaceAggregation.aggregationKey shouldBe SpendReportingAggregationKeys.Workspace
+    workspaceAggregation.spendData.headOption.getOrElse(fail("workspace results not parsed correctly")).googleProjectId shouldBe defined
```
cahrens, line 165: It seems like it might be worth testing that the cost information for each workspace is correct (and have non-zero costs for both so that the sum is actually the sum of both things).

I have the same comment for the date aggregation.

core/src/main/resources/swagger/api-docs.yaml
cahrens: I see my answer in the code below… Daily is the default value. If we want to stick with that, I think it should be documented here. Or I personally think it would be best to only return the summary data if an aggregationKey is not specified.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -16,18 +20,42 @@ case class SpendReportingForDateRange(
...
                                        startTime: DateTime,
-                                       endTime: DateTime
+                                       endTime: DateTime,
+                                       workspace: Option[WorkspaceName] = None,
+                                       googleProjectId: Option[GoogleProject] = None
                                      )
 
 // Key indicating how spendData has been aggregated. Ex. 'workspace' if all data in spendData is for a particular workspace
-case class SpendReportingAggregationKey(key: String)
+object SpendReportingAggregationKeys {
```
aherbst-broad, line 29: I like the switch to an enum here ❤️

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -44,9 +45,38 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
+
+    SpendReportingResults(Seq(spendAggregation), spendSummary)
+  }
+
+  private def extractWorkspaceSpendAggregation(rows: List[FieldValueList], currency: Currency, startTime: DateTime, endTime: DateTime, workspaceProjectsToNames: Map[GoogleProject, WorkspaceName]): SpendReportingAggregation = {
+    val workspaceSpend = rows.map { row =>
+      val rowCost = BigDecimal(row.get("cost").getDoubleValue).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
+      val rowCredits = BigDecimal(row.get("credits").getDoubleValue).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
+      val rowGoogleProjectId = GoogleProject(row.get("googleProjectId").getStringValue)
+      val rowWorkspaceName = workspaceProjectsToNames.getOrElse(rowGoogleProjectId, throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadGateway, s"unexpected project ${rowGoogleProjectId.value} returned by BigQuery")))
```
aherbst-broad, line 66: `BadGateway` seems like an odd error code if the workspace doesn't map to a project, a plain 500 might be more appropriate.

aherbst-broad (approved):
Looking good

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: The thinking here is that we've specified the set of Google projects that BigQuery should return (with `...AND project.id in UNNEST(@projects)...`), so if we hit an unexpected project here then BigQuery has returned some bad results. Same reasoning as the [502 if we get back different currencies](https://github.com/broadinstitute/rawls/pull/1660/files#diff-6bdd7e72eedead7aafdcd9b71fc64d8eaf45c390b2db2d004f768ea07a43fc53R121). 

That said, I'm good with switching this or both to a 500 if you think that's more clear/accurate

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -174,7 +174,7 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
       .build()
 }
 
-  def getSpendForBillingProject(billingProjectName: RawlsBillingProjectName, startDate: DateTime, endDate: DateTime, aggregationKey: Option[SpendReportingAggregationKey]): Future[SpendReportingResults] = {
+  def getSpendForBillingProject(billingProjectName: RawlsBillingProjectName, startDate: DateTime, endDate: DateTime, aggregationKey: Option[SpendReportingAggregationKey] = None): Future[SpendReportingResults] = {
```
cahrens, line 177: It might be good to have a test that you don't get the detailed results if the aggregation key is not present.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -202,33 +203,36 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
     }
   }
 
+  case class AggregationKeyQueryField(bigQueryFieldOpt: Option[String], aliasOpt: Option[String]) {
+    def aliased(): String = (bigQueryFieldOpt, aliasOpt) match {
+      case (Some(bq), Some(key)) => s""",
```
cahrens: Why is it formatted this way?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
cahrens: Thanks for making this change!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: This was from copying it out of the original multiline query, I'll remove the extra whitespace

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -21,42 +23,93 @@ import org.scalatestplus.mockito.MockitoSugar
...
-  )
-  val defaultFieldValues: List[FieldValueList] = List(
-    FieldValueList.of(firstRow.asJava, fields:_*),
-    FieldValueList.of(secondRow.asJava, fields:_*)
-  )
-  val defaultPage: PageImpl[FieldValueList] = new PageImpl[FieldValueList](null, null, defaultFieldValues.asJava)
-  val schema: Schema = Schema.of(fields:_*)
-  val defaultTableResult: TableResult = new TableResult(schema, 2, defaultPage)
+  object SpendReportingTestData {
+    object Daily {
```
marctalbott, line 31: Test data was starting to get unwieldy, so I've moved it into these objects. The test data in `Daily` is mostly the same as what existed before, but I've extracted a few values into variables (e.g. `firstRowCost` and `firstRowDate`) and added the `...CostRounded` variables to use in test assertions. 

The test data in `Workspace` has fields specific to workspace breakdowns, but is otherwise very similar to the `Daily` test data.

cahrens (approved):

PR #1662, opened 2022-03-09
get workspace by id [AJ-337]

davidangb:
Adds a new API to retrieve a workspace by its id. This API has the same features as - and at the implementation level, delegates to - the API to retrieve a workspace by its name.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
@@ -65,6 +65,15 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
     } ~
+      path("workspaces" / "tags") {
+        parameter('q.?) { queryString =>
+          get {
+            complete {
+              workspaceServiceConstructor(userInfo).getTags(queryString)
+            }
+          }
+        }
+      } ~
       path("workspaces" / Segment / Segment) { (workspaceNamespace, workspaceName) =>
         patch {
           entity(as[Array[AttributeUpdateOperation]]) { operations =>
@@ -207,15 +216,6 @@ trait WorkspaceApiService extends UserInfoDirectives {
         get {
           complete {
             workspaceServiceConstructor(userInfo).getBucketUsage(WorkspaceName(workspaceNamespace, workspaceName))
-          }
-        }
-      } ~
-      path("workspaces" / "tags") {
-        parameter('q.?) { queryString =>
-          get {
-            complete {
-              workspaceServiceConstructor(userInfo).getTags(queryString)
-            }
           }
         }
       } ~
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
@@ -75,7 +75,7 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
       } ~
-      path("workspaces" / Segment ) { workspaceId =>
+      path("workspaces" / "id" / Segment ) { workspaceId =>
         get {
           parameterSeq { allParams =>
             traceRequest { span =>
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -65,6 +66,27 @@ trait WorkspaceApiService extends UserInfoDirectives {
...
+          parameterSeq { allParams =>
+            traceRequest { span =>
+              complete {
+                workspaceServiceConstructor(userInfo).getWorkspaceById(workspaceId,
+                  WorkspaceFieldSpecs.fromQueryParams(allParams, "fields"), span)
+              }
+            }
+          }
+        }
+      } ~
```
davidangb, line 89: the `path("workspaces" / "tags")` had to move higher, so that it is considered first; otherwise, `path("workspaces" / Segment )` would match it.

As a weird complication of these APIs having the same shape: an end user could, in the swagger UI, enter "tags" as the UUID for the get-workspace-by-id API, and it will return the tags payload. Do we care about this?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -305,6 +305,26 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  def getWorkspaceById(workspaceId: String, params: WorkspaceFieldSpecs, parentSpan: Span = null): Future[JsObject] = {
+    val workspaceUuid = Try(UUID.fromString(workspaceId)) match {
+      case Success(uid) => uid
+      case Failure(_) => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, "invalid UUID"))
+    }
+    // retrieve the namespace/name for this workspace and then delegate to getWorkspace(WorkspaceName).
+    // note that this id -> namespace/name lookup does not enforce security; that's enforced in getWorkspace(WorkspaceName).
+    val workspaceRecords = dataSource.inTransaction { dataAccess =>
+      dataAccess.workspaceQuery.findByIdQuery(workspaceUuid).map(r => (r.namespace, r.name)).take(1).result
+    }
```
davidangb, line 317: the sql query generated by this is:
```sql
select `namespace`, `name` from `WORKSPACE` where `id` = ? limit 1
```

davidangb, line 317: the sql query generated by this is:
```sql
select `namespace`, `name` from `WORKSPACE` where `id` = ? limit 1
```

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
jordansafer: Agreed tags isn't actually a uuid, but I think we should avoid reusing the namespace. Since we're already using the namespace for workspace namespace, there's already potential for conflicts between 'tags' and the workspace namespace, and I think we should avoid the possibility of conflicts between uuid and namespace. The meaning I would infer from this exact path: "workspaces"/{variable} would be getting all workspaces for a workspace namespace, since we're already using workspaces/{namespace}/{workspaceName}. I would vote for something like workspaces/id/{workspaceId}

jordansafer (commented):
One minor comment on the API path discussion, otherwise all looks good!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
davidangb: yeah, unfortunately because `workspaces/{namespace}/{workspaceName}` exists, using `workspaces/id/{workspaceId}` translates to making the string "id" unusable as a namespace. blech, I will sleep on this!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -65,7 +66,35 @@ trait WorkspaceApiService extends UserInfoDirectives {
...
       path("workspaces" / Segment / Segment) { (workspaceNamespace, workspaceName) =>
+        /* we enforce a 6-character minimum for workspaceNamespace, as part of billing project creation.
+           the previous "mc", "tags", and "id" paths rely on this convention to avoid path-matching conflicts.
+           we might want to change the first Segment above to a regex a la """[^/.]{6,}""".r
+           but note that would be a behavior change: if a user entered fewer than 6 chars it would result in an
+           unmatched path rejection instead of the custom error handling inside WorkspaceService.
+        */
+
```
davidangb, line 97: additionally, we have some unit tests that use a namespace of "ns" - all those tests would need to be updated if we path-matched on a 6-char minimum here.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
davidangb: after some research and realizing that namespaces should always be 6+ characters, I agree with your suggestion and have changed the api signature to `/workspaces/id/{workspaceId}`!

jordansafer (approved):

aherbst-broad (approved):
lgtm

PR #1663, opened 2022-03-09
[SUP-449] Add an optional limit parameter to the get tags endpoint

nawatts:
Currently, users have to enter 3 characters for anything to show up in the tags menu on the workspaces list or workspace dashboard. We'd like to remove that restriction and show the most frequently used tags when the user focuses on the menu.  Currently, the only way to do that is to load all the tags in Terra and filter them client side. This adds an optional `limit` parameter to the get tags endpoint so that the UI can request only the top few tags.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -320,20 +320,27 @@ trait AttributeComponent {
       }
     }
 
-    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String]) = {
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None) = {
 
       val basicFilter = filter(rec =>
         rec.namespace === attrName.namespace &&
           rec.name === attrName.name &&
           rec.valueString.isDefined)
 
-      val res = (queryString match {
+      val baseQuery = (queryString match {
         case Some(query) => basicFilter.filter(_.valueString.like(s"%$query%"))
         case None => basicFilter
-      }).groupBy(_.valueString).map(queryThing =>
-        (queryThing._1, queryThing._2.length))
+      })
+        .groupBy(_.valueString)
+        .map(queryThing => (queryThing._1, queryThing._2.length))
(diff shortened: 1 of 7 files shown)


davidangb:
heads up that #1662 also has changes to the akka-http route definition for tags, I bet one of us will end up having to resolve conflicts

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -320,20 +320,27 @@ trait AttributeComponent {
...
-    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String]) = {
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None) = {
 
       val basicFilter = filter(rec =>
         rec.namespace === attrName.namespace &&
           rec.name === attrName.name &&
           rec.valueString.isDefined)
 
-      val res = (queryString match {
+      var res = (queryString match {
```
MatthewBemis: Instead of making this a `var` (mutable values are generally frowned upon in Scala), I would recommend building the query like:

```
      val baseQuery = (queryString match {
        case Some(query) => basicFilter.filter(_.valueString.like(s"%$query%"))
        case None => basicFilter
      }).groupBy(_.valueString).map(queryThing =>
        (queryThing._1, queryThing._2.length)
      ).sortBy(r => (r._2.desc, r._1))

      val res = limit match {
        case Some(limit) => baseQuery.take(limit)
        case None => baseQuery
      }

      res.map(x => (x._1.get, x._2))
```

MatthewBemis (commented):
Looks good to me, pending taking care of the `var`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
nawatts: Done in [e5c8eac](https://github.com/broadinstitute/rawls/pull/1663/commits/e5c8eac13721100800c30e99fd8964fd7c6eca6b)

MatthewBemis (approved):

aherbst-broad (approved):

nawatts:
Rebased to resolve merge conflicts.

PR #1664, opened 2022-03-10
SU-40: Add endpoint to check for active workspace file transfers

MatthewBemis:
A few users have reported that they've cloned a workspace, gone to the notebooks tab, and then been surprised to see that their notebooks weren't copied over. There is a delay, so we'd like to be able to show a notice to the users when their notebooks are still being copied.

Terra UI will call this endpoint to see if the transfer is still active and, if so, display a helpful message to the user.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3255,10 +3255,20 @@ paths:
           schema:
             type: string
       responses:
-        204:
+        200:
           description: Successful Request
-          content: {}
-
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/PendingCloneWorkspaceFileTransfer'
+        404:
+          description: Workspace Not Found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/workspaces/{workspaceNamespace}/{workspaceName}/lock:
     put:
       tags:
@@ -5955,6 +5965,24 @@ components:
           type: integer
           description: The current storage bucket usage in bytes
       description: ""
+    PendingCloneWorkspaceFileTransfer:
+      type: object
+      properties:
+        destWorkspaceId:
+          type: string
+          description: "The ID of the workspace that the files will be copied to"
+        sourceWorkspaceBucketName:
+          type: string
+          description: "The name of the bucket that the files will be copied from"
+        destWorkspaceBucketName:
+          type: string
+          description: "The name of the bucket that the files will be copied to"
+        copyFilesWithPrefix:
+          type: string
+          description: "The file prefix that was supplied when the transfer was initiated"
+        destWorkspaceGoogleProjectId:
+          type: string
+          description: "The Google project that the destination workspace belongs to"
     Attribute:
       type: object
       properties:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5979,7 +5979,7 @@ components:
           description: "The name of the bucket that the files will be copied to"
         copyFilesWithPrefix:
           type: string
-          description: "The file prefix that was supplied when the transfer was initiated"
+          description: "The file prefix indicating which files to copy to the destination workspace"
         destWorkspaceGoogleProjectId:
           type: string
           description: "The Google project that the destination workspace belongs to"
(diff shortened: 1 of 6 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3234,6 +3234,41 @@ paths:
...
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                type: array
```
MatthewBemis, line 3263: Note that this endpoint returns a list of `PendingCloneWorkspaceFileTransfer`s. 

In reality, it will currently only ever return 0 or 1 `PendingCloneWorkspaceFileTransfer`s. I personally find the list approach to be easier to consume. The other benefit here is that this is future-proofed for if/when we ever add file transfers as a general feature (meaning multiple active file transfers could be possible).

The alternative approach would be to return 204 in the case of none, 200 + JSON in the case of a pending transfer, and 404 otherwise. To reviewers: let me know which one you prefer and if you think I should keep it as-is or change it.

nawatts: Returning a list sounds good to me. The future proofing argument makes sense and checking the list's length seems cleaner than dealing with multiple different status codes.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -189,6 +189,13 @@ trait WorkspaceApiService extends UserInfoDirectives {
           }
         }
       } ~
+      path("workspaces" / Segment / Segment / "fileTransfers") { (workspaceNamespace, workspaceName) =>
+        get {
+          complete {
+            workspaceServiceConstructor(userInfo).listPendingFileTransfersForWorkspace(WorkspaceName(workspaceNamespace, workspaceName)).map(pendingTransfers => StatusCodes.OK -> pendingTransfers)
```
nawatts, line 195: Just for my own Scala education, what does `.map(pendingTransfers => StatusCodes.OK -> pendingTransfers)` do?

nawatts (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
MatthewBemis: That part is taking the result of the `listPendingFileTransfersForWorkspace` `Future`, unboxing the result contained in the `Future` (which I've named `pendingTransfers`), and then transforming the result into the `StatusCodes.OK -> pendingTransfers` pair, which is then re-boxed into the `Future` and fed to the outer `complete` akka-http directive, which returns the status code and marshals `pendingTransfers` into a JSON response. 

In this particular case, this extra step is not actually needed because `OK` is the default response code for a successful `Future`, so the result of `listPendingFileTransfersForWorkspace` would have been returned anyway with an OK status code. But I still like to do it because it makes the code more readable IMO. If I wanted this to return a status code other than OK, I'd definitely need to do this step.

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: I think this makes sense, and I like the future proofing.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3234,6 +3234,41 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/fileTransfers:
+    get:
+      tags:
+        - workspaces
+      summary: List pending file transfers for this workspace
```
aherbst-broad: It took me a minute to grok the semantics of this endpoint wrt incoming vs outgoing -- would it make sense to reword slightly to indicate that these are _incoming_ file transfers?

aherbst-broad (approved):
lgtm, thanks for doing this.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
nawatts: Thanks for the explanation.

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: I agree, it's too vague. I've updated the description in the latest commit

PR #1666, opened 2022-03-12
[Do not merge] [PROD-655] Cherry-pick Rawls Billing latency fix

jack-r-warren:
Context: https://broadinstitute.slack.com/archives/CA1481PRN/p1647095734770549?thread_ts=1647056957.328129&cid=CA1481PRN


jack-r-warren:
Do not merge this, because the commit involved has already been merged to dev! This PR is actually based on an older dev commit, so that this build would just have the fix without unrelated Rawls work. After Rawls is back on a normal version I expect us to close this PR and delete the branch.

PR #1667, opened 2022-03-17
[WOR-130] Spend Category breakdown, multiple aggregations, sub aggregation

marctalbott:
Ticket: [WOR-130](https://broadworkbench.atlassian.net/browse/WOR-130)
* Add `Category` as a new aggregation key to group spend by different types cloud costs (Storage, Compute, and Other)
* Support multiple aggregation keys at once
* Support sub-aggregation (e.g. break down cost by workspace and then further break down specific workspace cost by spend category)

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -31,18 +36,63 @@ object SpendReportingAggregationKeys {
     override def toString = getClass.getSimpleName.stripSuffix("$")
 
     override def withName(name: String): SpendReportingAggregationKey = SpendReportingAggregationKeys.withName(name)
+
```
marctalbott, line 39: Moved from `AggregationKeyQueryField` in `SpendReportingService`

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -60,7 +60,7 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with TestDriverComponent
       )
       val page: PageImpl[FieldValueList] = new PageImpl[FieldValueList](null, null, fieldValues.asJava)
       val schema: Schema = Schema.of(fields:_*)
-      val tableResult: TableResult = new TableResult(schema, 2, page)
+      val tableResult: TableResult = new TableResult(schema, fieldValues.length, page)
```
marctalbott: I'm getting a bit concerned about how big this test class is getting. A lot of fake test data and different assertions are required for the more complicated aggregations. We could work some magic with inheritance to slim this down, but this PR is already a somewhat significant refactor so I didn't want to do anything too drastic here. Let me know if you disagree and I can see about cutting this class down in size!

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -31,18 +36,63 @@ object SpendReportingAggregationKeys {
...
+    val bigQueryField: String
+    val bigQueryAlias: String
+
+    def bigQueryAliasClause(): String = s", $bigQueryField as $bigQueryAlias"
```
aherbst-broad, line 43: 👍 I like this approach

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -166,14 +281,177 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with TestDriverComponent
...
+  it should "support sub-aggregations" in withDefaultTestDatabase { dataSource: SlickDataSource =>
+    val service = createSpendReportingService(dataSource, tableResult = SpendReportingTestData.SubAggregation.tableResult)
+
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(SpendReportingTestData.Workspace.workspace1))
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(SpendReportingTestData.Workspace.workspace2))
+
+    val reportingResults = Await.result(service.getSpendForBillingProject(testData.billingProject.projectName, DateTime.now().minusDays(1), DateTime.now(), Set(SpendReportingAggregationKeyWithSub(SpendReportingAggregationKeys.Workspace, Option(SpendReportingAggregationKeys.Category)))), Duration.Inf)
+    val topLevelAggregation = reportingResults.spendDetails.headOption.getOrElse(fail("spend results not parsed correctly"))
+
+    withClue("total cost was incorrect"){
```
aherbst-broad, line 311: 👍 for breadcrumbs

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
aherbst-broad: This is ok with me, test cases for complex reporting logic like this can be somewhat gnarly.

aherbst-broad (approved):
lgtm

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -45,83 +45,125 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
-      val rowCost = BigDecimal(row.get("cost").getDoubleValue).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
-      val rowCredits = BigDecimal(row.get("credits").getDoubleValue).setScale(currency.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
-      SpendReportingForDateRange(rowCost.toString(),
-        rowCredits.toString(),
-        currency.getCurrencyCode,
-        DateTime.parse(row.get("date").getStringValue),
-        DateTime.parse(row.get("date").getStringValue).plusDays(1).minusMillis(1))
+  private def extractSpendAggregation(rows: List[FieldValueList], currency: Currency, aggregationKey: SpendReportingAggregationKeyWithSub, workspaceProjectsToNames: Map[GoogleProject, WorkspaceName] = Map.empty): SpendReportingAggregation = {
+    aggregationKey match {
+      case SpendReportingAggregationKeyWithSub(SpendReportingAggregationKeys.Category, subAggregationKey) => extractCategorySpendAggregation(rows, currency, subAggregationKey, workspaceProjectsToNames)
```
cahrens, line 80: What happens if the subaggregation key is the same as the aggregation key? Does the code handle that gracefully ? I can imagine that either the subaggregation key is ignored, or an error is thrown.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -31,7 +30,17 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
   val userServiceConstructor: UserInfo => UserService
   val spendReportingConstructor: UserInfo => SpendReportingService
 
-  implicit def spendReportingAggregationKeyUnmarshaller: Unmarshaller[String, SpendReportingAggregationKey] = Unmarshaller.strict(SpendReportingAggregationKeys.withName)
+  implicit def aggregationKeyParameterUnmarshaller: Unmarshaller[String, SpendReportingAggregationKeyWithSub] = Unmarshaller.strict { parameter =>
+    val delimitedParameter = parameter.split("~").toList
+
+    delimitedParameter match {
+      case key :: subKey :: Nil => SpendReportingAggregationKeyWithSub(SpendReportingAggregationKeys.withName(key), Option(SpendReportingAggregationKeys.withName(subKey)))
```
cahrens, line 37: If we are in the first case, we know `subKey` exists. So why is it wrapped with Option?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -57,25 +66,19 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
-                    case _: IllegalArgumentException => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid dates provided"))
-                  }.map { case (parsedStartDate, parsedEndDate) =>
-                    spendReportingConstructor(userInfo).getSpendForBillingProject(
-                      RawlsBillingProjectName(projectId),
-                      parsedStartDate,
-                      parsedEndDate.plusDays(1).minusMillis(1),
-                      aggregationKey
-                    )
-                  }
+                  spendReportingConstructor(userInfo).getSpendForBillingProject(
```
cahrens, line 76: It looks like the changes in this file aren't covered by tests. Are there unit tests that verify the routing based on URL that could be extended to cover this code?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -57,25 +66,19 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
           pathEndOrSingleSlash {
             get {
               parameters(
-                "startDate".as[String],
-                "endDate".as[String],
+                "startDate".as[DateTime],
```
cahrens, line 69: What happens now with invalid dates?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -166,14 +281,177 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with TestDriverComponent
...
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(SpendReportingTestData.Workspace.workspace1))
+    runAndWait(dataSource.dataAccess.workspaceQuery.createOrUpdate(SpendReportingTestData.Workspace.workspace2))
+
+    val reportingResults = Await.result(service.getSpendForBillingProject(testData.billingProject.projectName, DateTime.now().minusDays(1), DateTime.now(), Set(SpendReportingAggregationKeyWithSub(SpendReportingAggregationKeys.Workspace, Option(SpendReportingAggregationKeys.Category)), SpendReportingAggregationKeyWithSub(SpendReportingAggregationKeys.Category))), Duration.Inf)
+    withClue("total cost was incorrect") {
+      reportingResults.spendSummary.cost shouldBe SpendReportingTestData.SubAggregation.totalCostRounded.toString
+    }
+
+    reportingResults.spendDetails.map {
+      case SpendReportingAggregation(SpendReportingAggregationKeys.Workspace, spendDataForWorkspaces) => {
```
cahrens: This block of verification of Workspace~Category is the same is in the previous test point, correct? Could you pull it into a helper method and share it between the tests to remove the duplication? It also makes it easier to read the tests.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -166,14 +281,177 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with TestDriverComponent
...
+              } else {
+                fail(s"unexpected category found in spend results - $subAggregatedSpendData")
+              }
+            }
+          } else {
+            fail(s"unexpected workspace found in spend results - $spendData")
+          }
+        }
+      }
+      case SpendReportingAggregation(SpendReportingAggregationKeys.Category, spendData) => {
```
cahrens: I this this block of code is the same as an earlier test, where you verified the ability to aggregate by Category. Similar to my last comment, could you share the code?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
cahrens: Where we are asserting the same data in multiple test points (aggregation by `Compute`, aggregation by `Workspace~Compute`), I think it would really help to pull those blocks into helper methods that the individual test points can call. With well-named helper methods, the test points will become much easier to read, and it removes the code duplication.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -45,83 +45,125 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
+    SpendReportingAggregation(
+      SpendReportingAggregationKeys.Workspace, workspaceSpend
+    )
+  }
+
+  private def extractCategorySpendAggregation(rows: List[FieldValueList], currency: Currency, subAggregationKey: Option[SpendReportingAggregationKey] = None, workspaceProjectsToNames: Map[GoogleProject, WorkspaceName] = Map.empty): SpendReportingAggregation = {
+    val spendByCategory = rows.groupBy(row => TerraSpendCategories.categorize(row.get("service").getStringValue))
+    val categorySpend = spendByCategory.map { case (category, rowsForCategory) =>
+      val (cost, credits) = sumCostsAndCredits(rowsForCategory, currency)
+      val subAggregation = subAggregationKey.map { key =>
```
cahrens, line 131: It looks like subaggregation is only covered by tests for `Workspace~Compute`. I know setting up the test data is a pain, but I'm wondering if it would be possible to get better code coverage. I'm also having trouble understanding this code, so I'm not clear if it's obvious that all aggregation/subaggregation combinations will work.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -45,83 +45,125 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
+    )
+  }
 
-    Currency.getInstance(currencies.reduce { (x, y) =>
-      if (x.equals(y)) {
-        x
-      } else {
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadGateway, s"Inconsistent currencies found while aggregating spend data: $x and $y cannot be combined"))
+  private def extractWorkspaceSpendAggregation(rows: List[FieldValueList], currency: Currency, subAggregationKey: Option[SpendReportingAggregationKey] = None, workspaceProjectsToNames: Map[GoogleProject, WorkspaceName]): SpendReportingAggregation = {
+    val spendByGoogleProjectId = rows.groupBy(row => GoogleProject(row.get("googleProjectId").getStringValue))
```
cahrens: Add types to variable.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: It does handle it gracefully, although the report that's returned isn't very useful since the data in the sub-aggregation has already been aggregated in the same say. Here's a snippet of the response

```
        {
          "category": "Other",
          "cost": "0.07",
          "credits": "0.00",
          "currency": "USD",
          "subAggregation": {
            "aggregationKey": "Category",
            "spendData": [
              {
                "category": "Other",
                "cost": "0.07",
                "credits": "0.00",
                "currency": "USD"
              }
            ]
          }
        }

File: core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala (hunk 6 above)
Comment (marctalbott) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala:37]: `SpendReportingAggregationKeyWithSub` expects `subKey` to have the type `Option[SpendReportingAggregationKey]` because we may or may not have a sub-aggregation. So in this case, even though we know it exists, we still need to wrap it in `Option` to satisfy the class signature of `SpendReportingAggregationKeyWithSub`. 

We could use `Some` instead of `Option` here, but I like to default to `Option` over `Some` because `Option` is resilient to an unexpected `null` (even though a `null` is theoretically impossible here)

File: core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala (hunk 8 above)
Comment (marctalbott) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala:69]: It throws a 400. The error message is slightly different, but still informative

```
"400 Bad Request": "The query parameter 'endDate' was malformed:\nInvalid format: \"bad-date\""

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -538,7 +554,51 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
...
+    withApiServicesAndCustomSpendReporting(dataSource, mockSpendReportingService) { services =>
+      val project = createProject("project")
+
+      Get(s"/billing/v2/${project.projectName.value}/spendReport?startDate=2022-03-06&endDate=2022-03-07&aggregationKey=Workspace~Daily&aggregationKey=Category&aggregationKey=Category~Category") ~>
+        sealRoute(services.billingRoutesV2) ~>
+        check {
+          assertResult(StatusCodes.OK, responseAs[String]) {
+            status
+          }
+        }
```
cahrens, line 605: Is it possible to assert that spendDetails are present (the opposite of line 572)? Or is that pointless because it's just a mock anyway?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -550,6 +610,42 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
       }
   }
 
+  it should "400 when unsupported aggregation keys are given" in withEmptyDatabaseAndApiServices { services =>
```
cahrens, line 621: Thanks for adding these tests!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
cahrens: I see. I probably would have used `Some`, but I don't feel strongly about it.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
marctalbott: I think because they're both mocked, the `spendDetails` assertion isn't worth it here or even really as valuable as it looks on 572. I'm changing both tests a little to throw if `getSpendForBillingProject` isn't called correctly and then to also verify that the method was called exactly once which covers the behavior that we care most about IMO

cahrens: Sounds good!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -172,17 +215,31 @@ class SpendReportingService(userInfo: UserInfo, dataSource: SlickDataSource, big
...
+  def getSpendForBillingProject(billingProjectName: RawlsBillingProjectName, startDate: DateTime, endDate: DateTime, aggregationKeyParameters: Set[SpendReportingAggregationKeyWithSub] = Set.empty): Future[SpendReportingResults] = {
     validateReportParameters(startDate, endDate)
     requireAlphaUser() {
       requireProjectAction(billingProjectName, SamBillingProjectActions.readSpendReport) {
         for {
           spendExportConf <- getSpendExportConfiguration(billingProjectName)
           workspaceProjectsToNames <- getWorkspaceGoogleProjects(billingProjectName)
-
-          query = getQuery(aggregationKey, spendExportConf.spendExportTable.getOrElse(spendReportingServiceConfig.defaultTableName))
+          aggregationKeys = aggregationKeyParameters.flatMap(maybeKeys => Set(Option(maybeKeys.key), maybeKeys.subAggregationKey).flatten)
```
ungwudik, line 243: Nit: what does `maybeKeys` mean?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
ungwudik: +1 on this.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -538,7 +554,59 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
       }
   }
 
-  "GET /billing/v2/{projectName}/spendReport" should "400 when invalid dates are given" in withEmptyDatabaseAndApiServices { services =>
+  "GET /billing/v2/{projectName}/spendReport" should "200 and return only summary information when aggregation key is not present" in withEmptyTestDatabase { dataSource: SlickDataSource =>
+    val mockSpendReportingService = mock[SpendReportingService](RETURNS_SMART_NULLS)
+    // if the API service does not parse parameters and call the spendReportingService correctly, test will fail
+    when(mockSpendReportingService.getSpendForBillingProject(any[RawlsBillingProjectName], any[DateTime], any[DateTime], any[Set[SpendReportingAggregationKeyWithSub]]))
+      .thenReturn(Future.failed(new RawlsException("parameters were not parsed correctly")))
```
ungwudik, line 561: Would it be helpful to also log the said parameters that aren't parsed correctly?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: ah this could be clearer, I'll add an explanatory comment.

This line unboxes the `aggregationKeyParameters` (all of which are of type `SpendReportingAggregationKeyWithSub(key: SpendReportingAggregationKey, subKey: Option[SpendReportingAggregationKey])`) and makes a `Set` of all the keys. It's a way of handling the possible `None` values and generating a set of keys that can be easily handled by `getQuery`

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
marctalbott: Perhaps, but I'm not sure how to get those values through the mock

cahrens (approved):
LGTM. Thanks for all the test cleanup!

ungwudik (approved):
lgtm

PR #1668, opened 2022-03-18
preselect scopes in swagger [AJ-352]

davidangb:
in swagger-ui, preselect the "openid", "email", and "profile" scopes - these are the required/default scopes for most APIs.

This only affects swagger-ui, it does not affect behavior of the app itself.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SwaggerRoutes.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SwaggerRoutes.scala
@@ -60,6 +60,7 @@ trait SwaggerRoutes extends LazyLogging {
                                         |        clientSecret: "${swaggerConfig.realm}",
                                         |        realm: "${swaggerConfig.realm}",
                                         |        appName: "rawls",
+                                        |        scopes: ["openid", "email", "profile"],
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SwaggerRoutes.scala
```diff
@@ -60,6 +60,7 @@ trait SwaggerRoutes extends LazyLogging {
                                         |        clientSecret: "${swaggerConfig.realm}",
                                         |        realm: "${swaggerConfig.realm}",
                                         |        appName: "rawls",
+                                        |        scopes: ["openid", "email", "profile"],
```
jordansafer, line 63: How come we don't preselect billing?

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SwaggerRoutes.scala
davidangb: the billing scope is only necessary for a handful of the billing APIs, and as a principle of least privilege I'm defaulting to leaving that unchecked … most users/API calls won't need it.

PR #1689, opened 2022-04-08
SU-58: Make workspace tags private

MatthewBemis:
Currently, workspace tags are global: a user can see all of the tags on any workspace in the entire system. This was a deliberate product decision driven by hopes that tags would somehow inspire collaboration or give a more "open data" feeling. That motivation has never really come to fruition and product would like to alter the behavior of this endpoint to be more useful for users. There is also a potential security benefit here, i.e. we do not make it well known that tags are public and users may be putting private data into their tags.

This PR changes the behavior of the getTags endpoint by only returning tags for the workspaces that the user has access to. It does so by calling Sam to get the list of workspaceIds that the user has access to, and then limiting the DB query to those workspaceIds. 

https://broadworkbench.atlassian.net/browse/SU-58

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -326,12 +326,12 @@ trait AttributeComponent {
     }
 
     /**
-      * Returns all unique attributes values that meet the specified filters
+      * Returns all unique attribute values, with the specfied filters applied
       *
-      * @param attrName the name of the attribute to search for (note that this is a (namespace, name) pair)
-      * @param queryString the string value to search for, optional
+      * @param attrName the name of the attribute to filter for (note that this is a (namespace, name) pair)
+      * @param queryString the string value to filter for, optional
       * @param limit the maximum number of results to return, optional
-      * @param ownerIds the ownerIds (i.e. the entity ID or workspace ID) to limit the search to, optional. if not specified, will search the entire table
+      * @param ownerIds the ownerIds (i.e. the entity ID or workspace ID that the attribute belongs to) to filter for, optional
       * @return the attribute values that meet the specified filters
       */
     def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None, ownerIds: Option[Seq[OWNER_ID]] = None) = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -327,19 +327,18 @@ trait AttributeComponent {
 
     def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None, ownerIds: Option[Seq[OWNER_ID]] = None) = {
 
-      val basicFilter = filter(rec =>
+      val basicFilter = filter { rec =>
         rec.namespace === attrName.namespace &&
           rec.name === attrName.name &&
-          rec.valueString.isDefined)
-
-      val applyOwnerFilter = ownerIds match {
-        case Some(ids) => basicFilter.filter(_.ownerId inSetBind ids)
-        case None => basicFilter
-      }
+          rec.valueString.isDefined
+        }
+        .filterOpt(ownerIds) { case (table, ownerIds) =>
+          table.ownerId inSetBind ownerIds
+        }
 
       val baseQuery = (queryString match {
-        case Some(query) => applyOwnerFilter.filter(_.valueString.like(s"%$query%"))
-        case None => applyOwnerFilter
+        case Some(query) => basicFilter.filter(_.valueString.like(s"%$query%"))
+        case None => basicFilter
       })
         .groupBy(r => caseSensitiveCollate(r.valueString))
         .map(queryThing => (queryThing._1, queryThing._2.length))
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -325,16 +325,21 @@ trait AttributeComponent {
       qb.sqlBuilder += " collate utf8_bin"
     }
 
-    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None) = {
+    def findUniqueStringsByNameQuery(attrName: AttributeName, queryString: Option[String], limit: Option[Int] = None, ownerIds: Option[Seq[OWNER_ID]] = None) = {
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -325,16 +325,31 @@ trait AttributeComponent {
...
+      val basicFilter = filter { rec =>
         rec.namespace === attrName.namespace &&
           rec.name === attrName.name &&
-          rec.valueString.isDefined)
+          rec.valueString.isDefined
+      }
+
+      val withOptionalOwnerFilter =
+        basicFilter.filterOpt(ownerIds) { case (table, ownerIds) =>
+          table.ownerId inSetBind ownerIds
```
MatthewBemis, line 347: I need to double check this: potentially having >1000 UUIDs in an `in` clause might be an issue.

Batching might be the more scalable solution here.

MatthewBemis: OK yeah this definitely needs to batch

davidangb: `listWorkspaces` does the same thing without batching - many many ids inside the `in` clause. Are you sure you need to batch - and include all the headache of combining the batched results back into a useful result?

MatthewBemis: I actually just reached this same conclusion myself which is why I blew away the commit that had it partially implemented and resolved this! I tested with 2345 workspaceIds and it was totally fine

MatthewBemis: The reason I thought I needed to batch was because I saw this comment in another area that batches https://github.com/broadinstitute/rawls/blob/f536962c85911d50ae23246d3fd422608d6804e9/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L935

MatthewBemis:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
davidangb: oooh, fair point - though it looks like "The number of values in the IN() list is only limited by the [max_allowed_packet](https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_max_allowed_packet) value." according to https://dev.mysql.com/doc/refman/5.6/en/comparison-operators.html, which means we'd need ~4MB of workspaceIds to hit that limit

nawatts (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -161,8 +161,8 @@ trait WorkspaceComponent {
       }
     }
 
-    def getTags(queryString: Option[String], limit: Option[Int] = None): ReadAction[Seq[WorkspaceTag]] = {
-      val tags = workspaceAttributeQuery.findUniqueStringsByNameQuery(AttributeName.withTagsNS, queryString, limit).result
+    def getTags(queryString: Option[String], limit: Option[Int] = None, ownerIds: Option[Seq[UUID]] = None): ReadAction[Seq[WorkspaceTag]] = {
```
cahrens, line 164: Is there Swagger documentation to update for this change?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -651,10 +651,15 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
   }
 
-  def getTags(query: Option[String], limit: Option[Int] = None): Future[Seq[WorkspaceTag]] =
-    dataSource.inTransaction { dataAccess =>
-      dataAccess.workspaceQuery.getTags(query, limit)
-    }
+  def getTags(query: Option[String], limit: Option[Int] = None): Future[Seq[WorkspaceTag]] = {
+    for {
+      workspacesForUser <- samDAO.getPoliciesForType(SamResourceTypeNames.workspace, userInfo)
+      workspaceIdsForUser = workspacesForUser.map(resource => UUID.fromString(resource.resourceId)).toSeq
```
davidangb: you might want to add a `Try` around the `UUID.fromString` - see also https://github.com/broadinstitute/rawls/blob/0cbb762a1ce1f4493ee9cae04b628861fab52447/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L688.

At some point during corner-case testing, someone manually added non-UUID Sam resources of type "workspace". Any user with access to such a Sam resource would get a fatal error on `UUID.fromString` and this API will never complete.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
MatthewBemis: I've updated the description here: 0616372fb

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: Ugh, that's right. Handled in 062a04380

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -651,10 +651,18 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
-    dataSource.inTransaction { dataAccess =>
-      dataAccess.workspaceQuery.getTags(query, limit)
-    }
+  def getTags(query: Option[String], limit: Option[Int] = None): Future[Seq[WorkspaceTag]] = {
+    for {
+      workspacesForUser <- samDAO.getPoliciesForType(SamResourceTypeNames.workspace, userInfo)
+      //Filter out non-UUID workspaceIds, which are possible in Sam but not valid in Rawls
+      workspaceIdsForUser = workspacesForUser.map(resource => Try(UUID.fromString(resource.resourceId))).collect {
+        case Success(workspaceId) => workspaceId
+      }.toSeq
```
davidangb, line 660: slightly terser, your call on which to use:
```scala
workspaceIdsForUser = workspacesForUser.flatMap(resource => Try(UUID.fromString(resource.resourceId)).toOption).toSeq
```

davidangb (approved):
:+1: since I have been offering uncalled-for feedback on this PR :)

PR #1691, opened 2022-04-11
Create Azure relay on Workspace creation (WOR-136).

cahrens:
Note that the endpoint has not yet been added to Swagger-- that will be done shortly. For now, to test this you can use curl:

Example for running against local "front" rawls, with a test namespace/workspace name:
````
curl -X POST --location "https://local.broadinstitute.org:20443/api/workspaces/mc" \                                                                  
    -H "accept: application/json" \
    -H "Authorization: Bearer <<output of  'gcloud auth print-access-token'>>" \
    -H "Content-Type: application/json" \
    -H "User-Agent: Testing" \
    -d "{
          \"namespace\": \"car_testing_ns\",
          \"name\": \"car_18\",
          \"attributes\": {},
          \"cloudPlatform\": \"AZURE\", \"region\": \"eastus\"
        }"

````

Deletion of the relay will happen in WOR-228.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/MultiCloudWorkspaceSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/MultiCloudWorkspaceSpec.scala
@@ -21,7 +21,7 @@ class MultiCloudWorkspaceSpec extends AnyFreeSpecLike {
       val targetRawlsUrl = Uri(Rawls.url).withPath(Path(s"/api/workspaces/mc"))
       val testNs = "testing_ns_" + UUID.randomUUID().toString
       val name = "test_name" + UUID.randomUUID().toString
-      val payload = Map("namespace" -> testNs, "name" -> name, "attributes" -> Map.empty, "cloudPlatform" -> "AZURE")
+      val payload = Map("namespace" -> testNs, "name" -> name, "attributes" -> Map.empty, "cloudPlatform" -> "AZURE", "region" -> "eastus")
 
       val thrown = intercept[RestException](Rawls.postRequest(
         uri = targetRawlsUrl.toString(),
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -112,21 +112,22 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
     )
   }
 
-  def createControlledAzureRelay(workspaceId: UUID, region: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult = {
-    val jobControlId = UUID.randomUUID().toString // tried adding a prefix to this, but then got database exception about field too long
+  def createAzureRelay(workspaceId: UUID, region: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult = {
+    val jobControlId = UUID.randomUUID().toString
     getControlledAzureResourceApi(accessToken).createAzureRelayNamespace(
-      new CreateControlledAzureRelayNamespaceRequestBody().common(  // Got 500 without common fields
+      new CreateControlledAzureRelayNamespaceRequestBody().common(
         new ControlledResourceCommonFields().name(s"relay-rcf-${workspaceId}").
-          cloningInstructions(CloningInstructionsEnum.NOTHING).accessScope(AccessScope.SHARED_ACCESS).
+          cloningInstructions(CloningInstructionsEnum.NOTHING).
+          accessScope(AccessScope.SHARED_ACCESS).
           managedBy(ManagedBy.USER)
       ).azureRelayNamespace(
         new AzureRelayNamespaceCreationParameters().namespaceName(s"relay-ns-${workspaceId}").region(region)
-      ).jobControl(new JobControl().id(jobControlId)),  // got 500 without job control
+      ).jobControl(new JobControl().id(jobControlId)),
       workspaceId
     )
   }
 
-  def getControlledAzureRelayResult(workspaceId: UUID, jobControlId: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult ={
+  def getCreateAzureRelayResult(workspaceId: UUID, jobControlId: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult ={
     getControlledAzureResourceApi(accessToken).getCreateAzureRelayNamespaceResult(workspaceId, jobControlId)
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -127,7 +127,7 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
     )
   }
 
-  def getCreateAzureRelayResult(workspaceId: UUID, jobControlId: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult ={
+  def getCreateAzureRelayResult(workspaceId: UUID, jobControlId: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult = {
     getControlledAzureResourceApi(accessToken).getCreateAzureRelayNamespaceResult(workspaceId, jobControlId)
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -112,7 +112,21 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
     )
   }
 
-  def createControlledAzureRelay(workspaceId: UUID, requestInfo: CreateControlledAzureRelayNamespaceRequestBody, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult = {
-    getControlledAzureResourceApi(accessToken).createAzureRelayNamespace(requestInfo, workspaceId)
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala [hunk 1]
```diff
@@ -145,7 +145,8 @@ object WorkspaceVersions {
 case class MultiCloudWorkspaceRequest(namespace: String,
                                       name: String,
                                       attributes: AttributeMap,
-                                      cloudPlatform: WorkspaceCloudPlatform
+                                      cloudPlatform: WorkspaceCloudPlatform,
+                                      region: String
```
cahrens, line 149: I'm still feeling somewhat uncomfortable with `region` being supplied in the request to create a workspace. I would imagine that most people would want the relay to be in the same region as their managed resource group, and therefore I think we should get the region somehow from azure based on the information in the azureConfig.

Should I start a conversation about this with regard to workspace landing zones? We will need to know pretty soon what "region" selection we are going to present in the "new workspace" dialog (though we could hardcode a region in the short term I guess).

In any case though, I feel like we could change where `region` comes from in a future ticket/PR, and for now we probably just want to move this PR through to unblock future work.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -33,6 +33,10 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
     apiClientProvider.getWorkspaceApplicationApi(accessToken.token)
   }
 
+  private def getControlledAzureResourceApi(accessToken: OAuth2BearerToken) = {
```
aherbst-broad, line 36: There is an [HttpWorkspaceManagerDAOSpec ](https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAOSpec.scala) that we can use to get coverage on this method and the other new ones in this class.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -100,6 +102,16 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
         Future(workspaceManagerDAO.enableApplication(workspaceId, wsmConfig.leonardoWsmApplicationId, userInfo.accessToken))
       )
+      _ = logger.info(s"Creating Azure relay in WSM [workspaceId = ${workspaceId}]")
+      azureRelayCreateResult <- traceWithParent("createAzureRelayInWSM", parentSpan)(_ =>
+        Future(workspaceManagerDAO.createAzureRelay(workspaceId, workspaceRequest.region, userInfo.accessToken))
+      )
+      relayJobControlId = azureRelayCreateResult.getJobReport.getId
+      _ = logger.info(s"Polling on Azure relay in WSM [workspaceId = ${workspaceId}, jobControlId = ${relayJobControlId}]")
+      _ <- traceWithParent("pollGetAzureRelayCreationStatusInWSM", parentSpan)(_ =>
+        pollWMCreation(workspaceId, relayJobControlId, userInfo.accessToken, 5 seconds,
```
aherbst-broad, line 112: nitpick: any reason this polls every 5 seconds vs the 2 seconds during cloud context creation?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -160,8 +187,10 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
 }
 
-class CloudContextCreationFailureException(message: String,
-                                           val workspaceId: UUID,
-                                           val jobControlId: String) extends RawlsException(message)
+class WorkspaceManagerCreationFailureException(message: String,
+                                               val workspaceId: UUID,
+                                               val jobControlId: String) extends RawlsException(message)
+
+class WorkspaceManagerPollingOperationException(message: String, val status: StatusEnum) extends Exception(message)
```
cahrens, line 194: I added a message to this because the stack trace appears in the log on retries, and before the message was just "null" which made me think that I was hitting a NullPointer exception.

Here is how it looks now:

````
rawls [INFO] [13:36:43.092] [scala-execution-context-global-1231] o.b.d.r.w.MultiCloudWorkspaceService - retry-able operation failed: 10 retries remaining, retrying in 5 seconds
rawls org.broadinstitute.dsde.rawls.workspace.WorkspaceManagerPollingOperationException: Polling Azure relay [jobControlId = 5b03b18c-de55-40cf-b35e-bfbddc9bc56c] for status to be SUCCEEDED. Current status: RUNNING.
rawls 	at org.broadinstitute.dsde.rawls.workspace.MultiCloudWorkspaceService.getAzureRelayCreationStatus(MultiCloudWorkspaceService.scala:141)

File: core/src/test/resources/reference.conf [hunk 5]
```diff
@@ -134,7 +134,7 @@ multiCloudWorkspaces {
         tenantId = "fake_tenant"
     }
     workspaceManager {
-        cloudContextPollTimeoutSeconds = 30
+        pollTimeoutSeconds = 60
```
cahrens, line 137: I'm finishing (running locally) with about 15 seconds to spare. We may have to bump this higher to avoid sporadic failures.

cahrens:
Evidence of successful workspace creation and a relay actually existing:

image

image

automation/src/test/scala/org/broadinstitute/dsde/test/api/MultiCloudWorkspaceSpec.scala
```diff
@@ -31,4 +31,4 @@ class MultiCloudWorkspaceSpec extends AnyFreeSpecLike {
       thrown.fields("statusCode").convertTo[Int] shouldBe 501
     }
   }
```
cahrens, line 33: Some point soon I imagine we want to add an integration test that we can actually create an Azure workspace from rawls.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens, line 112: Polling here are 5 second interval instead of 2 because we expect this to take about 45 seconds.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aherbst-broad: Yeah, let's chat during today's tech talk, there some additional wrinkles w/WLZs that we need to consider. 

Agreed I think this is fine for now to block downstream UI work and integration.

aherbst-broad (approved):
Looks good

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: I think our comments crossed. Cloud context creation is pretty much instant, but this the relay takes about 45 seconds to create. I didn't see any reason to be polling every 2 seconds because it adds so much noise to the logs, but can change it to be consistent if we want.

automation/src/test/scala/org/broadinstitute/dsde/test/api/MultiCloudWorkspaceSpec.scala
cahrens: @aherbst-broad are you thinking we do this when the API to do this when we combine the APIs to create a workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -107,4 +111,23 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
   }
+
+  def createAzureRelay(workspaceId: UUID, region: String, accessToken: OAuth2BearerToken): CreateControlledAzureRelayNamespaceResult = {
+    val jobControlId = UUID.randomUUID().toString
+    getControlledAzureResourceApi(accessToken).createAzureRelayNamespace(
+      new CreateControlledAzureRelayNamespaceRequestBody().common(
+        new ControlledResourceCommonFields().name(s"relay-rcf-${workspaceId}").
+          cloningInstructions(CloningInstructionsEnum.NOTHING).
+          accessScope(AccessScope.SHARED_ACCESS).
+          managedBy(ManagedBy.USER)
```
marctalbott, line 122: I'm not very familiar with these settings, could you give a little background as to why we want these settings?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: If relay creation is going to take 45 seconds on its own, I expect there's a good chance workspace creation as a whole takes well over a minute which could cause problems for upstream services (AoU times out after 60 seconds, not sure if the Terra UI will do the same). Do we need to make this endpoint async?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
cahrens: I am not familiar with them either, but @aherbst-broad suggested them.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
cahrens: Added coverage in f3b873109639ff47b6e246d39d8d2fdde696b185.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: Agreed, we need to think about making the endpoint async. But I think that is out of the scope of this ticket. I will create a new one for it.

https://broadworkbench.atlassian.net/browse/WOR-259

marctalbott (approved):

PR #1694, opened 2022-04-13
[CA-1781] Sync Billing Account Changes with Google

ehigham:
DO NOT MERGE UNTIL https://broadworkbench.atlassian.net/browse/CA-1844 IS COMPLETE
Running CA-1844 after this is merged will potentially corrupt the state of billing accounts on workspaces and billing projects.

This change extends the `WorkspaceBillingAccountActor` (nee `WorkspaceBillingAccountMonitor`) code to update billing account changes on v1 billing projects even if there are no workspaces in the billing project.

Using the new `BillingAccountChanges` table, the actor
- gets the *latest* billing account change to all billing projects, ignoring the previous unsync'ed changed
- attempts to update the billing account on the billing project google project, if it exists
- gets all workspaces in the billiong project
- attempts to update the billing account on the workspace google project, if v2, and writes the state of the change to the workspace (ie billing account and billing account error message)

I've changed this file to use the more modern akka-actor-typed interface. I believe this has helped tremedously with testing behaviour directly, removing the need for polling.

In a FiaB, manually verified:
- changing billing accounts for empty v1 billing projects works
- billing accounts of v2 workspace google projects are updated.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+
+object WorkspaceBillingAccountActor {
+  sealed trait WorkspaceBillingAccountsMessage
+
+  case object UpdateBillingAccounts extends WorkspaceBillingAccountsMessage
+
+  def apply(dataSource: SlickDataSource,
+            gcsDAO: GoogleServicesDAO,
+            initialDelay: FiniteDuration,
+            pollInterval: FiniteDuration): Behavior[WorkspaceBillingAccountsMessage] =
```
rtitle, line 30: Typed actors do seem better. My akka-actor knowledge is rusty though. :)

fwiw, Leo migrated its async tasks from actors to fs2 streams a while back. See: https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/AsyncTaskProcessor.scala

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+    }
+}
+
+final case class WorkspaceBillingAccountActor(dataSource: SlickDataSource, gcsDAO: GoogleServicesDAO)
+  extends LazyLogging {
+
+  import dataSource.dataAccess._
+  import dataSource.dataAccess.driver.api._
+
+  /* Sync billing account changes to billing projects and their associated workspaces with Google
```
rtitle, line 53: I probably don't have all the context -- but what does a "billing account change" represent, and how is the table populated in Rawls?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+          "changeId" -> billingAccountChange.id,
+          "billingProject" -> billingAccountChange.billingProjectName.value,
+          "newBillingAccount" -> billingAccountChange.newBillingAccount.map(_.value)
+        )
+
+        billingProject <- loadBillingProject(billingAccountChange.billingProjectName)
+        billingProbeHasAccess <- billingProject.billingAccount.traverse(gcsDAO.testDMBillingAccountAccess(_).io)
+        billingProjectSyncAttempt <- syncBillingProjectWithGoogle(billingProject).attempt
+        syncOutcome = Outcome.fromEither(billingProjectSyncAttempt)
+        _ <- recordBillingProjectSyncOutcome(billingAccountChange, billingProbeHasAccess, syncOutcome)
```
rtitle: Should we record the outcome only after we've synced all the workspaces? (Since presumably it could still fail at the workspace stage.)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Yes - I've been interested in pursuing that at some point. Rawls has a lot of monitors/actors however, so wanted to get them into a more functional style before using fs2!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: When you update which google billing account a `RawlsBillingProject` uses, rawls only sets the new billing account if it's different. When the billing account is set, rawls creates a `BillingAccountChange` which is a record that describes what change to the billing account was made on the billing project, who made it, and when rawls applied this change in google.
When a `BillingProjectChange` is created, it goes into a queue that's consumed by the `WorkspaceBillingAccountActor` that pulls the latest changes per billing project and applies them to the resources in google.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Confusing v1/v2 billing project/workspace time!

v1 billing projects are backed by a google project. All v1 workspaces in a v1 billing project use that google project. When you sync a billing account change to a billing project, you're only syncing if it's a v1 billing project.
Since v2 workspaces use their own google project, any errors generated while updating the billing account on their google projects get written to the workspace record. Note that if rawls encounters a problem syncing the v1 billing project's billing account, then this error gets written to the v1 workspace's error field.

I believe the goal is to get rid of billing projects altogether as they currently stand so failures while setting the old v1 billing project's billing account are the only errors that get written to the billing project's message field (other than the legacy creation statuses/errors in creation).

rtitle: Thanks - this makes sense I think, didn't realize we captured errors at the workspace/googleProject level as well.

rtitle (approved):
LGTM - seems like it does the thing, and thanks for refactoring as you go. 👍 

Just curious how you test this stuff -- I see there are unit tests, but have you tried actually updating a real GCP billing account through Rawls (and is there any swatomation support for things like that)?

ehigham:
> LGTM - seems like it does the thing, and thanks for refactoring as you go. +1
> 
> Just curious how you test this stuff -- I see there are unit tests, but have you tried actually updating a real GCP billing account through Rawls (and is there any swatomation support for things like that)?

Yeah (I should have added this to the description!)
I've done some manual testing in a FiaB to make sure everything behaves as expected and that the logs are in order. Sadly we removed the ability to create v1 workspaces so I can't exercise that code in a FiaB manually :(

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActorSpec.scala
```diff
@@ -0,0 +1,467 @@
...
+        .unsafeRunSync()
+
+      runAndWait {
+        for {
+          lastChange <- BillingAccountChanges.getLastChange(testData.billingProject.projectName)
+          billingProject <- rawlsBillingProjectQuery.load(testData.billingProject.projectName)
+          workspace <- workspaceQuery.findByIdOrFail(testData.workspace.workspaceId)
+        } yield {
+          lastChange.value.googleSyncTime shouldBe defined
+          lastChange.value.outcome.value.isSuccess shouldBe true
```
andy7i: based on the definition of `override def setBillingAccountName`, I think isSuccess should be false

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -313,20 +320,6 @@ trait WorkspaceComponent {
       loadWorkspaces(workspaces)
     }
 
-    def listWorkspaceGoogleProjectsToUpdateWithNewBillingAccount(): ReadAction[Set[(GoogleProjectId, Option[RawlsBillingAccountName], Option[RawlsBillingAccountName])]] = {
```
andy7i, line 316: so happy we're removing this convoluted query!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+    }
+      .io
+      .map(_.toList)
+
+
+  private def updateBillingAccountOnGoogle(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): IO[Unit] =
+    for {
+      projectBillingInfo <- gcsDAO.getBillingInfoForGoogleProject(googleProjectId).io
+      currentBillingAccountOnGoogle = getBillingAccountOption(projectBillingInfo)
+      _ <- Applicative[IO].whenA(newBillingAccount != currentBillingAccountOnGoogle) {
```
andy7i: thoughts on logging `currentBillingAccountOnGoogle` before we make the change? Maybe I'm just stuck in the billing fiasco mindset of log _everything_ just in case / for more info

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+        info("Successfully updated Billing Account on Billing Project in Google", baseInfo :_*) else
+        warn("Failed to update Billing Account on Billing Project in Google", ("details" -> message) +: baseInfo :_*)
+
+      _ <- dataSource.inTransaction { _ =>
+        val thisChange = BillingAccountChanges.withId(change.id)
+        val thisBillingProject = rawlsBillingProjectQuery.withProjectName(change.billingProjectName)
+        DBIO.seq(
+          thisChange.setGoogleSyncTime(syncTime.some),
+          thisChange.setOutcome(outcome.some),
+          thisBillingProject.setInvalidBillingAccount(!billingProbeCanAccessBillingAccount.getOrElse(true)),
```
andy7i: Making sure I understand this part -- we expect that `billingProbeCanAccessBillingAccount` will be None if the new billing account is None? So that's why we do `.getOrElse(true)` here

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -0,0 +1,275 @@
...
+        "changeId" -> change.id,
+        "billingProject" -> change.billingProjectName.value,
+        "newBillingAccount" -> change.newBillingAccount.map(_.value)
+      )
+
+      (_, message) = Outcome.toTuple(outcome)
+      _ <- if (outcome.isSuccess)
+        info("Successfully updated Billing Account on Billing Project in Google", baseInfo :_*) else
+        warn("Failed to update Billing Account on Billing Project in Google", ("details" -> message) +: baseInfo :_*)
+
```
andy7i: When I first read this message, I thought this was recording the result of syncing all google projects (GP) in a billing project (BP) (including any v2 workspaces' GPs). This message is logged after syncing a v1 BP, but _before_ syncing v2 workspaces. 

If the intention was to print this message after all GPs (in a BP) are synced, I think we need to move this function so it happens after the workspace sync. Else maybe we should make the log say explicitly that this is for v1 BPs.

Q: If a v2 workspace GP sync fails, shouldn't that affect the Outcome of the BillingAccountChange?

andy7i (commented):
lots of great refactoring in this pr! (also, the slick refactor is still so nice to read!)
Mostly concerned about how we're recording the Outcome based on the BP GP sync but not workspace GP syncs. I will approve once that question is resolved!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: All the changes are recorded in the database and we're logging the record id. I think we logged everything because we didnt persist this information anywhere. Now we are writing to the DB i think we can get away with logging less. What do you think?

andy7i: I think we log the old billing account _according to Rawls_ (from the Rawls DB), but not the billing account _according to Google_ (on the actual Google Project), right?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActorSpec.scala
ehigham: Yeah you're right. I've updated the code so that it combines all errors and writes that as the `MESSAGE` field of the database. Now if an update fails, the failure message will reflect all the reasons why it failed.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Yeah it wasn't the clearest - I changed this to be a simple bool which will be false if the account is None or the billing probe can't access the accout.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Yep. Good point. I incorporaed your suggestion.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
```diff
@@ -0,0 +1,36 @@
...
+            for every change it makes. That was run before #1694 was merged. To prevent the new
+            `WorkspaceBillingAccountActor` from reading unfulfilled `BillingAccountChange`s that
+            might potentially undo the work of the script, we'll delete all `BillingAccountChange`s.
+            Note that these changes would have been applied by the `WorkspaceBillingAccountMonitor`
+            (the monitor the new actor replaces) - no data (other than audit history) will be lost.
+
+            This seemed like the simplest compromise as it would reduce the risk should newer,
+            less "soak-tested" code go awry.
+        </comment>
+        <delete tableName="BILLING_ACCOUNT_CHANGES"/>
```
andy7i: If we delete the table, do we need to re-create it?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
```diff
@@ -0,0 +1,53 @@
...
+            - a python script that aimed to re-synchronise billing project billing accounts in Rawls
+            with the values in Google as discrepancies arose post-"Project-per-Workspace" from
+            users making edits to the billing account associated with Google projects directly.
+
+            That script modifies the database directly and does not create `BillingAccountChange`s
+            for every change it makes. That was run before #1694 was merged. To prevent the new
+            `WorkspaceBillingAccountActor` from reading unfulfilled `BillingAccountChange`s that
+            might potentially undo the work of the script, we'll delete all `BillingAccountChange`s.
+            Note that these changes would have been applied by the `WorkspaceBillingAccountMonitor`
+            (the monitor the new actor replaces) - no data (other than audit history) will be lost.
```
andy7i: alternatively, what if we populate existing rows with a date in the past. Maybe like epoch time 0 or something. That way we don't delete any data.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
```diff
@@ -0,0 +1,53 @@
...
+            This seemed like the simplest compromise as it would reduce the risk should newer,
+            less "soak-tested" code go awry.
+        </comment>
+        <delete tableName="BILLING_ACCOUNT_CHANGES"/>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="ehigham" id="ALTER_BILLING_ACCOUNT_CHANGE_FK">
+        <comment>
+            The `FK_BILLING_ACCOUNT_CHANGES_PROJECT_NAME` foreign key constraint prevents users
+            from deleting their billing projects if they changed the billing account.
+        </comment>
```
andy7i: I like these sql comments you've added as breadcrumbs of knowledge / context.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -145,131 +155,145 @@ final case class WorkspaceBillingAccountActor(dataSource: SlickDataSource, gcsDA
...
+    } yield v2Outcome
 
 
-  private def listV2WorkspacesInProject(billingProject: RawlsBillingProject): IO[List[Workspace]] =
-    dataSource.inTransaction { _ =>
-      workspaceQuery
-        .withBillingProject(billingProject.projectName)
-        .withoutGoogleProjectId(billingProject.googleProjectId)
-        .read
+  private def setGoogleProjectBillingProject[F[_]](googleProjectId: GoogleProjectId)
```
andy7i:
```suggestion
  private def setGoogleProjectBillingAccount[F[_]](googleProjectId: GoogleProjectId)
```
and elsewhere

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActorSpec.scala
```diff
@@ -451,16 +496,24 @@ class WorkspaceBillingAccountActorSpec
...
+              msg should include(workspace.googleProjectId.value)
+            }
+          }
 
           billingProject.value.invalidBillingAccount shouldBe false
           billingProject.value.message shouldBe empty
 
-          workspace.billingAccountErrorMessage shouldBe defined
-          workspace.currentBillingAccountOnGoogleProject shouldBe testData.v1Workspace.currentBillingAccountOnGoogleProject
+          forAll(workspaces) { ws =>
```
andy7i, line 512: I think it would be interesting to see/test the behavior if 1 of the 2 workspaces failed. But I don't feel strongly about it.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
```diff
@@ -0,0 +1,53 @@
...
+            might potentially undo the work of the script, we'll delete all `BillingAccountChange`s.
+            Note that these changes would have been applied by the `WorkspaceBillingAccountMonitor`
+            (the monitor the new actor replaces) - no data (other than audit history) will be lost.
+
+            This seemed like the simplest compromise as it would reduce the risk should newer,
+            less "soak-tested" code go awry.
+        </comment>
+        <delete tableName="BILLING_ACCOUNT_CHANGES"/>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="ehigham" id="ALTER_BILLING_ACCOUNT_CHANGE_FK">
```
andy7i: maybe worth pulling out this bug fix as its own PR?

andy7i (commented):
overall looks good. Ideally I think we should pull out the bug fix into a separate PR, rename the function I commented on, and maybe consider setting a sync time instead of deleting the table.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
ehigham: We don't - we delete all rows in the table.
https://docs.liquibase.com/change-types/delete.html

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
ehigham: I think starting with a clean slate is the simplest thing to do. I don't much like the idea of setting a fake syched date because then you'd have to remember that some of the data in the table wasn't sync'ed by the actor and there was a special time after which the actor started and it's all a bit of a mess. 
I don't really consider this "losing data" as this is only an artefact of splitting PRs - it's meaningless without the actor.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Ah thanks

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220504_delete_billing_account_changes.xml
ehigham: https://broadworkbench.atlassian.net/browse/CA-1875 (https://github.com/broadinstitute/rawls/pull/1762)

andy7i (approved):
LGTM! thanks for all of the answers / changes / splitting out multiple PRs.

PR #1695, opened 2022-04-14
AJ-355 add rename entity type end point

sagehen03:
My `EntityService` and test code could likely use a more refined scala touch.

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -1416,7 +1416,7 @@ paths:
           description: Successful Request
           content: { }
         404:
-          description: Workspace or Entity not found
+          description: Workspace or Entity Type not found
           content:
             'application/json':
               schema:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -1460,7 +1460,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
         404:
-          description: Workspace or Entity not found
+          description: Workspace or Entity Type not found
           content:
             'application/json':
               schema:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -1394,7 +1394,7 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
       x-codegen-request-body-name: newEntityNameJson
-  /api/workspaces/{workspaceNamespace}/{workspaceName}/entityType/rename:
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameEntityType:
     post:
       tags:
         - entities
@@ -1405,7 +1405,7 @@ paths:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
         - $ref: '#/components/parameters/workspaceNamePathParam'
       requestBody:
-        description: Old entity type name and what you want to rename it
+        description: Old entity type name and the name you'd like to change it to
         content:
           'application/json':
             schema:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -1394,6 +1394,42 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
       x-codegen-request-body-name: newEntityNameJson
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entityType/rename:
+    post:
+      tags:
+        - entities
+      summary: rename entity type in a workspace
+      description: Rename an entity type
+      operationId: renameEntityType
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespacePathParam'
+        - $ref: '#/components/parameters/workspaceNamePathParam'
+      requestBody:
+        description: Old entity type name and what you want to rename it
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/EntityTypeRename'
+        required: true
+      responses:
+        204:
+          description: Successful Request
+          content: { }
+        404:
+          description: Workspace or Entity not found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        409:
+          description: New name for entity type already exists in workspace
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+      x-codegen-request-body-name: entityTypeRenameJson
   /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/{entityName}/evaluate:
     post:
       tags:
@@ -4191,6 +4227,18 @@ components:
           type: string
           description: The name of the entity
       description: ""
+    EntityTypeRename:
+      required:
+        - oldName
+        - newName
+      type: object
+      properties:
+        oldName:
+          type: string
+          description: the entity type name to change
+        newName:
+          type: string
+          description: the new entity type name
     EntityCopyDefinition:
       required:
         - destinationWorkspace
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -816,7 +816,8 @@ trait EntityComponent {
     }
 
     def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
-      ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName)
+      ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName) andThen
+        workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
     }
 
     def rename(workspaceContext: Workspace, entityType: String, oldName: String, newName: String): ReadWriteAction[Int] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -475,7 +475,7 @@ trait EntityComponent {
       def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
 
         sql"""select exists (select name from ENTITY
-               where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $entityType)
+               where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $entityType and deleted = 0)
           """.as[Boolean]
       }
     }
@@ -487,7 +487,7 @@ trait EntityComponent {
       def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): WriteAction[Int] = {
 
         sqlu"""update ENTITY set entity_type = $newName
-              where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $oldName
+              where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $oldName and deleted = 0
           """
       }
     }
@@ -811,8 +811,8 @@ trait EntityComponent {
       }
     }
 
-    def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
-      CheckForExistingEntityTypeQuery.doesEntityTypeAlreadyExist(workspaceContext, entityType)
+    def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Option[Boolean]] = {
+      uniqueResult(CheckForExistingEntityTypeQuery.doesEntityTypeAlreadyExist(workspaceContext, entityType))
     }
 
     def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -468,6 +468,30 @@ trait EntityComponent {
       }
     }
 
+    //noinspection SqlDialectInspection
+    private object CheckForExistingEntityTypeQuery extends RawSqlQuery {
+      val driver: JdbcProfile = EntityComponent.this.driver
+
+      def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
+
+        sql"""select exists (select name from ENTITY
+               where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $entityType)
+          """.as[Boolean]
+      }
+    }
+
+    //noinspection SqlDialectInspection
+    private object ChangeEntityTypeNameQuery extends RawSqlQuery {
+      val driver: JdbcProfile = EntityComponent.this.driver
+
+      def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): WriteAction[Int] = {
+
+        sqlu"""update ENTITY set entity_type = $newName
+              where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $oldName
+          """
+      }
+    }
+
     // Slick queries
 
     // Active queries: only return entities and attributes with their deleted flag set to false
@@ -785,6 +809,14 @@ trait EntityComponent {
       EntityDependenciesDeletionQuery.deleteAction(workspaceContext) andThen {
         filter(_.workspaceId === workspaceContext.workspaceIdAsUUID).delete
       }
+    }
+
+    def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
+      CheckForExistingEntityTypeQuery.doesEntityTypeAlreadyExist(workspaceContext, entityType)
+    }
+
+    def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
+      ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName)
     }
 
     def rename(workspaceContext: Workspace, entityType: String, oldName: String, newName: String): ReadWriteAction[Int] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -816,8 +816,12 @@ trait EntityComponent {
     }
 
     def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
-      ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName) andThen
-        workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
+      for {
+        numRowsRenamed <- ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName)
+        _ <- workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
+      } yield {
+        numRowsRenamed
+      }
     }
 
     def rename(workspaceContext: Workspace, entityType: String, oldName: String, newName: String): ReadWriteAction[Int] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -129,7 +129,14 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       dataSource.inTransaction { dataAccess =>
         dataAccess.entityQuery.doesEntityTypeAlreadyExist(workspaceContext, newName) flatMap {
           _.head match {
-            case false => dataAccess.entityQuery.changeEntityTypeName(workspaceContext, oldName, newName)
+            case false =>
+              dataAccess.entityQuery.doesEntityTypeAlreadyExist(workspaceContext, oldName) flatMap {
+                _.head match {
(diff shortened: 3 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -124,6 +124,25 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
+                _.head match {
+                  case false => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound,
+                    s"Can't find entity type ${oldName}"))
+                  case true => dataAccess.entityQuery.changeEntityTypeName(workspaceContext, oldName, newName)
+                }
+              }
+            case true => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict, s"${newName} already exists as an entity type"))
+          }
+        }
+      }
```
sagehen03: This is admittedly a mess.  I've tried simplifying to `if(_.head)..` but the compiler doesn't recognize the variable in that case.  I also originally had my `workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)` call immediately following the db call to change the entity type and that led to a 204 response every time but the db never updating.  Any tips, enlightenment especially appreciated here.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -787,6 +811,15 @@ trait EntityComponent {
       }
     }
 
+    def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
+      CheckForExistingEntityTypeQuery.doesEntityTypeAlreadyExist(workspaceContext, entityType)
+    }
```
davidangb, line 816: optionally, you could use `uniqueResult` here:
```scala
def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Option[Boolean]] = {
  uniqueResult(CheckForExistingEntityTypeQuery.doesEntityTypeAlreadyExist(workspaceContext, entityType))
}
```
I find it easier to think in terms of Options for a result like this than a Seq, where callers will always only be looking at Seq.head. But they're functionally equivalent; this is more of a code-style comment.

Thinking out loud: we could also add an `exactlyOneResult` helper parallel to `uniqueResult` - whereas `uniqueResult` expects 0 or 1 rows, `exactlyOneResult` would expect exactly 1.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -787,6 +811,15 @@ trait EntityComponent {
...
+
+    def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
+      ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName) andThen
+        workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
+    }
```
davidangb, line 825: by putting `workspaceQuery.updateLastModified` second here, the result of `changeEntityTypeName` will be the number of rows updated by  `workspaceQuery.updateLastModified`, not the number of rows updated by `ChangeEntityTypeNameQuery.changeEntityTypeName`. This seems wrong. One option is to swap the order of calls here, but that changes internal logic. To keep the order of operations the same but return the result of `ChangeEntityTypeNameQuery.changeEntityTypeName`, you could do:

```scala
def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): ReadWriteAction[Int] = {
  for {
    numRowsRenamed <- ChangeEntityTypeNameQuery.changeEntityTypeName(workspaceContext, oldName, newName)
    _ <- workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
  } yield {
    numRowsRenamed
  }
}
```
(for-yield constructs like this result in flatMaps, just like andThen results in a flatMap)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -124,6 +124,25 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
+    }
```
davidangb, line 159: you're looking for tips on how to clean up this code … my suggestion is to modularize the logic around validating the existence or non-existence of the oldName and the newName, by creating separate functions to handle those. With that in place, you could move to a for-yield instead of the flatMaps.

Here's an example of such a refactor. NOTE: this refactor assumes that `doesEntityTypeAlreadyExist` is rewritten using `uniqueResult` to return an `Option` instead of a `Seq` (see my previous comment):

```scala
  def renameEntityType(workspaceName: WorkspaceName, oldName: String, newName: String): Future[Int] = {

    import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction}

    def validateExistingType(dataAccess: DataAccess, workspaceContext: Workspace, oldName: String): ReadAction[Boolean] = {
      dataAccess.entityQuery.doesEntityTypeAlreadyExist(workspaceContext, oldName) map {
        case Some(true) => true
        case Some(false) => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.NotFound,
          s"Can't find entity type ${oldName}"))
        case None =>  throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.InternalServerError,
          s"Unexpected error; could not determine existence of entity type ${oldName}"))
      }
    }

    def validateNewType(dataAccess: DataAccess, workspaceContext: Workspace, newName: String): ReadAction[Boolean] = {
      dataAccess.entityQuery.doesEntityTypeAlreadyExist(workspaceContext, newName) map {
        case Some(true) => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict, s"${newName} already exists as an entity type"))
        case Some(false) => false
        case None =>  throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.InternalServerError,
          s"Unexpected error; could not determine existence of entity type ${newName}"))
      }
    }

    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
        for {
          _ <- validateNewType(dataAccess, workspaceContext, newName)
          _ <- validateExistingType(dataAccess, workspaceContext, oldName)
          renameResult <- dataAccess.entityQuery.changeEntityTypeName(workspaceContext, oldName, newName)
        } yield {
          renameResult
        }
      }
    }
  }
```

I haven't run this code through unit tests, but hopefully it illustrates the point

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: to get the `if` clauses working, grab the result of the `flatMap` and use that result instead of using `_`:
```scala
        dataAccess.entityQuery.doesEntityTypeAlreadyExist(workspaceContext, newName) flatMap { newNameExists =>
          if (newNameExists.head)
```
but also note that `head` will throw `NoSuchElementException` if `newNameExists` is empty

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -468,6 +468,30 @@ trait EntityComponent {
...
+    private object CheckForExistingEntityTypeQuery extends RawSqlQuery {
+      val driver: JdbcProfile = EntityComponent.this.driver
+
+      def doesEntityTypeAlreadyExist(workspaceContext: Workspace, entityType: String): ReadAction[Seq[Boolean]] = {
+
+        sql"""select exists (select name from ENTITY
+               where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $entityType)
+          """.as[Boolean]
+      }
+    }
```
davidangb, line 481: this should have an additional `and deleted=0` clause, I think. If the user has (soft-)deleted all entities of a given type, they should be allowed to rename to that type.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -468,6 +468,30 @@ trait EntityComponent {
...
+    private object ChangeEntityTypeNameQuery extends RawSqlQuery {
+      val driver: JdbcProfile = EntityComponent.this.driver
+
+      def changeEntityTypeName(workspaceContext: Workspace, oldName: String, newName: String): WriteAction[Int] = {
+
+        sqlu"""update ENTITY set entity_type = $newName
+              where workspace_id=${workspaceContext.workspaceIdAsUUID} and entity_type = $oldName
+          """
+      }
+    }
```
davidangb, line 493: Hm, what should *this* one do with soft-deleted entities? I think this should also include `and deleted=0` so it only renames active entities, but  I wonder if there's any reason to also rename all of them …

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -135,6 +135,13 @@ trait EntityApiService extends UserInfoDirectives {
             }
           }
         } ~
+        path("workspaces" / Segment / Segment / "entityType" / "rename") { (workspaceNamespace, workspaceName) =>
+          post {
+            entity(as[EntityTypeRename]) { rename =>
+              complete { entityServiceConstructor(userInfo).renameEntityType(WorkspaceName(workspaceNamespace, workspaceName), rename.oldName, rename.newName).map(_ => StatusCodes.NoContent) }
+            }
+          }
+        } ~
```
davidangb, line 144: style question: is it worth passing the entire `EntityTypeRename` object as an argument into `renameEntityType`, instead of passing the two Strings? That would allow us in the near future to add fields to the `EntityTypeRename` definition for merge options without having to change `renameEntityType`'s signature.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -159,6 +159,26 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+    withWorkspaceContext(testData.workspace) { context =>
+      val pair2EntityTypeExists = runAndWait(entityQuery.doesEntityTypeAlreadyExist(context, "Pair2")).head
+      assert(!pair2EntityTypeExists)
+      runAndWait(entityQuery.changeEntityTypeName(context, "Pair", "Pair2"))
+      val pair2EntityTypeExistsAfterUpdate = runAndWait(entityQuery.doesEntityTypeAlreadyExist(context, "Pair2")).head
+      assert(pair2EntityTypeExistsAfterUpdate)
+
+    }
+  }
+
```
davidangb, line 178: we've bemoaned how long Rawls unit tests take to run … a big contributor to that runtime is that each unit test truncates all tables and then repopulates with the exemplar data it needs for the test. `withDefaultTestDatabase` is a large exemplar dataset. Is there a smaller exemplar dataset you can use? Check out `withMinimalTestDatabase` instead (I don't know if it has any entity data), or `withConstantTestDatabase`, or maybe there's something else in there …

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityServiceSpec.scala
```diff
@@ -189,6 +189,23 @@ class EntityServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matcher
...
+    ex.errorReport.message shouldBe "Pair already exists as an entity type"
+  }
+
+  it should "rename an entity type as long as the selected name is not in use"  in withTestDataServices { services =>
+    val waitDuration = Duration(10, SECONDS)
+    assertResult(1) {Await.result(services.entityService.renameEntityType(testData.wsName, testData.pair1.entityType, "newPair"), waitDuration)}
+    // verify there are no longer any entities under the old entity name
+    val queryResult = Await.result(services.entityService.listEntities(testData.wsName, testData.pair1.entityType), waitDuration)
+    assert(queryResult.isEmpty)
+  }
```
davidangb, line 207: add a test for an attempt to rename an entity type that doesn't exist (i.e. "oldName" is incorrect")

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1394,6 +1394,42 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
       x-codegen-request-body-name: newEntityNameJson
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entityType/rename:
```
davidangb: Up for debate: what about making this 
`POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameType`?

It feels like this API should be parallel to 
`POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/delete`

alternately, it could be
`POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/rename`, where the oldName is specified in the path instead of the post body, but I like that less.

davidangb (commented):
first pass comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: I've changed to `uniqueResult`.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
sagehen03: I've incorporated this suggestion in its totality.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: Added.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: I added the deleted=0 for consistency, but it seems like this would only matter in an edge case where we resuscitate deleted entities for a customer.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
sagehen03: Seems reasonable, done.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
sagehen03: I changed one test (where we verify that an entity type does not exist) to `withMinimalTestDatabase` and that saves 800ms in my testing. Unfortunately, the minimal data set doesn't have entities so I left the other tests using the the default.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityServiceSpec.scala
sagehen03: Done

core/src/main/resources/swagger/api-docs.yaml
sagehen03: I went with `POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameEntityType`

sagehen03:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
davidangb: I know 800ms seems trivial, but we're currently at 1808 unit tests. 1808 * 800ms ≈ 24 minutes … gotta do everything we can!

sagehen03: Totally fair, I mentioned it because I was surprised it was that much.  Test went from 1.9s to 1.1s.

sagehen03:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: this is the only outstanding issue - should `changeEntityTypeName` return the count of rows that had their entity type changed, or should it always return a value of `1` for the count of workspaces that had their last-modified column changed?

davidangb (commented):
one remaining comment inline, otherwise :+1:!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: I've made the change to return the number of rows modified by the update ENTITY sql statement.

davidangb (approved):
proactive :+1: for the runtime changes; looks like there is still a unit test to fix

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -124,6 +124,41 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
       }
     }
 
+  def renameEntityType(workspaceName: WorkspaceName, renameInfo: EntityTypeRename): Future[Int] = {
+    import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction}
```
jordansafer, line 128: question, possibly to David since he suggested the syntax- what's our team approach to scope imports? Generally, I would look to avoid them for decluttering/readability, but I'm open to the idea that's helpful to import things right before they are used.

jordansafer (approved):

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1394,6 +1394,42 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
       x-codegen-request-body-name: newEntityNameJson
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameEntityType:
```
MatthewBemis, line 1397: This signature works for me so I'm happy to approve as-is, but I wanted to offer an alternative signature that is more RESTful and could avoid confusing bugs in the future:

PATCH:  `/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}`

And then a body that takes the new entity type name.

I think the ship already sailed on this long ago, but the API paths for entities are getting to be a bit crowded. I believe that REST principles state that the request verb and body should determine the action to perform, and the API path should determine the path to the resource that should be acted upon.

By adding actions to the API paths, it creates ambiguity, i.e. is a particular segment in the path telling us where to find the resource, or what action to perform? An example of the problem that this ambiguity creates can be found in our current APIs. See these two endpoints, one for getting a TSV for a certain entity type, and another for getting a single entity:

GET `/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv`
GET `/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/{entityName}`

Because the /tsv path is resolved first, it actually makes it impossible to call the second GET endpoint if your entity is named "tsv". No user has ever reported this bug, but it's still an issue and fixing it would be a bit of work. Preferably, the getTsv endpoint would have been something like `GET /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}&format=tsv`

sagehen03: You raise a number of valid points, Matt.  `PATCH` would be the right HTTP verb for this case according to Roy Fielding. Muddling the REST standard by putting `renameEntityType` in the url does at least help with the practical problem of avoiding url collisions in the future, but your suggested pattern would not collide with any existing urls at the moment.  I'm going to try making the PATCH and url pattern change and see if that's straightforward.

sagehen03: I went through the exercise of making the changes you suggest, @MatthewBemis .  You can see those changes over in #1699  Changing over to `PATCH` is quite straightforward, but while doing so I did notice/remember that switching to `PATCH` is at odds with the existing rename entity end point.  Is it worth being more RESTful with a new end point if that new end point is not consistent with an already published end point?  I'm not sure, but I'd lean towards probably not. 

Making the url change such that the entity type being changed is a path parameter was a bit less straightforward due to swagger.  Swaggers complains about this:     

PATCH: /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}
GET: /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}

It says that we have duplicate keys despite the urls associating with distinct verbs so I had to change the url to: 

PATCH: /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityTypeToRename}

(and add a new path param to the swagger).  I don't have a strong opinion on this, but if I had to pick I'd say consistency within Rawls' API (i.e. making rename entity type similar to rename entity) is a stronger impetus than adhering to REST standards.  @jordansafer @davidangb  what do you think?

jordansafer: For POST vs PATCH, I'm in favor of consistency. I think we could make a future ticket to clean up the APIs by making PATCH endpoints and deprecating the POST endpoints. For the URL change, I do think it's a good point that we're overloading with /entities/{entityType} and /entities/renameEntityType. I would vote to avoid the overload, and go with adding a path param. But that seems to require the PATCH change, in which case this could all be work for a future ticket.

sagehen03: I also see that we have a `PATCH` method for updating an entity's attributes so admittedly the consistency argument is not so cut and dry.  Rawls is consistently inconsistent perhaps, but I would say renaming an entity and renaming an entity type are closer relatives than renaming an entity type and updating an entity's attributes.

MatthewBemis: I think that API consistency is a compelling argument. And to be clear, I think this is a larger issue that goes far beyond the scope of this PR.

It's a bit of a moot point now, but the duplicate key error in the Swagger can be fixed by nesting both method types under the same API path. See [this example](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/swagger/api-docs.yaml#L145) where there is both a put and a delete listed underneath.

MatthewBemis (approved):
api looks good to me!

core/src/main/resources/swagger/api-docs.yaml
sagehen03: Cool, thank you for the swagger tip! I'm going to merge this as is for the sake of consistency with the rename entity end point with the acknowledgement that this API will surprise and frustrate a REST adherent :face_exhaling:

PR #1741, opened 2022-04-27
[CA-1781] Refactor billing project database operations to support query composition

ehigham:

Split from #1694 to refactor database oeprations to support composing database queries and operations. 
Using implicit classes to add "extension methods" to slick's `Query`, one can more easily chain together operations and re-use abstractions.

One change that was not unpicked is to only set the billing project's billing account when it's different from the current billing account.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
@@ -8,17 +8,73 @@ import org.broadinstitute.dsde.rawls.dataaccess.{GoogleApiTypes, GoogleOperation
 import org.broadinstitute.dsde.rawls.model.CreationStatuses.CreationStatus
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils
-import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits.{IgnoreResultExtensionMethod, InsertExtensionMethod}
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits.InsertExtensionMethod
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome
 import org.broadinstitute.dsde.workbench.model.google.{BigQueryDatasetName, BigQueryTableName, GoogleProject}
 
 import java.sql.Timestamp
 import java.time.Instant
-import scala.concurrent.ExecutionContext
 
-case class RawlsBillingProjectRecord(projectName: String, creationStatus: String, billingAccount: Option[String], message: Option[String], cromwellBackend: Option[String], servicePerimeter: Option[String], googleProjectNumber: Option[String], invalidBillingAccount: Boolean, spendReportDataset: Option[String], spendReportTable: Option[String], spendReportDatasetGoogleProject: Option[String])
+final case class RawlsBillingProjectRecord(projectName: String, creationStatus: String, billingAccount: Option[String], message: Option[String], cromwellBackend: Option[String], servicePerimeter: Option[String], googleProjectNumber: Option[String], invalidBillingAccount: Boolean, spendReportDataset: Option[String], spendReportTable: Option[String], spendReportDatasetGoogleProject: Option[String])
+
+object RawlsBillingProjectRecord {
+  def fromBillingProject(billingProject: RawlsBillingProject): RawlsBillingProjectRecord =
+    RawlsBillingProjectRecord(
+      billingProject.projectName.value,
+      billingProject.status.toString,
+      billingProject.billingAccount.map(_.value),
+      billingProject.message,
+      billingProject.cromwellBackend.map(_.value),
+      billingProject.servicePerimeter.map(_.value),
+      billingProject.googleProjectNumber.map(_.value),
+      billingProject.invalidBillingAccount,
+      billingProject.spendReportDataset.map(_.value),
+      billingProject.spendReportTable.map(_.value),
+      billingProject.spendReportDatasetGoogleProject.map(_.value)
+    )
+
+  def toBillingProject(projectRecord: RawlsBillingProjectRecord): RawlsBillingProject =
+    RawlsBillingProject(
+      RawlsBillingProjectName(projectRecord.projectName),
+      CreationStatuses.withName(projectRecord.creationStatus),
+      projectRecord.billingAccount.map(RawlsBillingAccountName),
+      projectRecord.message,
+      projectRecord.cromwellBackend.map(CromwellBackend),
+      projectRecord.servicePerimeter.map(ServicePerimeterName),
+      projectRecord.googleProjectNumber.map(GoogleProjectNumber),
+      projectRecord.invalidBillingAccount,
+      projectRecord.spendReportDataset.map(BigQueryDatasetName),
+      projectRecord.spendReportTable.map(BigQueryTableName),
+      projectRecord.spendReportDatasetGoogleProject.map(GoogleProject)
+    )
+
+  def toBillingProjectSpendExport(projectRecord: RawlsBillingProjectRecord): BillingProjectSpendExport = {
+    val table = for {
+      googleProjectId <- projectRecord.spendReportDatasetGoogleProject
+      datasetName <- projectRecord.spendReportDataset
+      tableName <- projectRecord.spendReportTable
+    } yield s"$googleProjectId.$datasetName.$tableName"
+
+    val billingAccount = RawlsBillingAccountName(projectRecord.billingAccount.getOrElse(
+      throw new RawlsException(s"billing account not set on project ${projectRecord.projectName}"))
+    )
+
+    BillingProjectSpendExport(RawlsBillingProjectName(projectRecord.projectName), billingAccount, table)
+  }
+
+}
 
 case class RawlsBillingProjectOperationRecord(projectName: String, operationName: GoogleOperationName, operationId: String, done: Boolean, errorMessage: Option[String], api: GoogleApiType)
+
+final case class BillingAccountChange(id: Long,
+                                      billingProjectName: RawlsBillingProjectName,
+                                      userId: RawlsUserSubjectId,
+                                      previousBillingAccount: Option[RawlsBillingAccountName],
+                                      newBillingAccount: Option[RawlsBillingAccountName],
+                                      created: Instant,
+                                      googleSyncTime: Option[Instant],
+                                      outcome: Option[Outcome]
+                                     )
 
 trait RawlsBillingProjectComponent {
   this: DriverComponent =>
@@ -48,7 +104,7 @@ trait RawlsBillingProjectComponent {
 
     def spendReportDatasetGoogleProject = column[Option[String]]("SPEND_REPORT_DATASET_GOOGLE_PROJECT", O.Length(1024))
 
-    def * = (projectName, creationStatus, billingAccount, message, cromwellBackend, servicePerimeter, googleProjectNumber, invalidBillingAccount, spendReportDataset, spendReportTable, spendReportDatasetGoogleProject) <> (RawlsBillingProjectRecord.tupled, RawlsBillingProjectRecord.unapply)
+    def * = (projectName, creationStatus, billingAccount, message, cromwellBackend, servicePerimeter, googleProjectNumber, invalidBillingAccount, spendReportDataset, spendReportTable, spendReportDatasetGoogleProject) <> ((RawlsBillingProjectRecord.apply _).tupled, RawlsBillingProjectRecord.unapply)
   }
 
   final class BillingAccountChanges(tag: Tag)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -8,18 +8,74 @@ import org.broadinstitute.dsde.rawls.dataaccess.{GoogleApiTypes, GoogleOperation
...
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits.InsertExtensionMethod
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome
 import org.broadinstitute.dsde.workbench.model.google.{BigQueryDatasetName, BigQueryTableName, GoogleProject}
 
 import java.sql.Timestamp
 import java.time.Instant
-import scala.concurrent.ExecutionContext
 
-case class RawlsBillingProjectRecord(projectName: String, creationStatus: String, billingAccount: Option[String], message: Option[String], cromwellBackend: Option[String], servicePerimeter: Option[String], googleProjectNumber: Option[String], invalidBillingAccount: Boolean, spendReportDataset: Option[String], spendReportTable: Option[String], spendReportDatasetGoogleProject: Option[String])
+final case class RawlsBillingProjectRecord(projectName: String, creationStatus: String, billingAccount: Option[String], message: Option[String], cromwellBackend: Option[String], servicePerimeter: Option[String], googleProjectNumber: Option[String], invalidBillingAccount: Boolean, spendReportDataset: Option[String], spendReportTable: Option[String], spendReportDatasetGoogleProject: Option[String])
```
rtitle, line 18: Don't need to update for this PR, but just a comment that you can make implicit `ColumnType`s to avoid using `String` for slick table mappings.

Example: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala#L86-L88

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -114,175 +170,183 @@ trait RawlsBillingProjectComponent {
...
+    def loadOperationsForProjects(projectNames: Seq[RawlsBillingProjectName], operationName: GoogleOperationName): ReadAction[Seq[RawlsBillingProjectOperationRecord]] =
       rawlsBillingProjectOperationQuery.filter(x => x.projectName.inSetBind(projectNames.map(_.value)) && x.operationName === operationName).result
-    }
+  }
 
-    private def marshalBillingProject(billingProject: RawlsBillingProject): RawlsBillingProjectRecord = {
-      RawlsBillingProjectRecord(billingProject.projectName.value, billingProject.status.toString, billingProject.billingAccount.map(_.value), billingProject.message, billingProject.cromwellBackend.map(_.value), billingProject.servicePerimeter.map(_.value), billingProject.googleProjectNumber.map(_.value), billingProject.invalidBillingAccount, billingProject.spendReportDataset.map(_.value), billingProject.spendReportTable.map(_.value), billingProject.spendReportDatasetGoogleProject.map(_.value))
-    }
+  implicit class RawlsBillingProjectExtensions(query: RawlsBillingProjectQuery) {
```
rtitle, line 270: Cool idea with these extension methods

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -326,36 +390,63 @@ trait RawlsBillingProjectComponent {
...
+    // filters
+    def withId(id: Long): BillingAccountChangeQuery =
+      query.filter(_.id === id)
+
+    def withProjectName(billingProjectName: RawlsBillingProjectName): BillingAccountChangeQuery =
+      query.filter(_.billingProjectName === billingProjectName.value)
+
+    /* SELECT *
+     * FROM BILLING_ACCOUNT_CHANGES BAC,
+     * (  SELECT BILLING_PROJECT_NAME, MAX(ID) AS MAXID
```
rtitle, line 422: Was wondering if there's a timestamp column we could use instead of `max(id)` (since this seems to care about most recent records).

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
```diff
@@ -96,9 +117,43 @@ object MigrationUtils {
...
 
-    implicit class ToJsonOps(data: Map[String, Any]) {
-      def toJson: JsValue = JsObject(data.view.mapValues(s => new JsString(s.toString)).toMap)
+    implicit class ToJsonOps(data: Iterable[(String, Any)]) {
+      def toJson: JsValue = JsObject(
+        data.map(Arrow[Function].second((s: Any) => new JsString(s.toString))).toMap
+      )
     }
+
+    implicit def monadErrorDBIOAction[E <: Effect]
```
rtitle, line 126: Noting this also exists :) https://github.com/RMSone/slick-cats

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -652,18 +654,21 @@ class UserService(protected val userInfo: UserInfo,
...
     dataSource.inTransaction { dataAccess =>
-      for {
-        _ <- dataAccess.rawlsBillingProjectQuery.updateBillingAccount(billingProjectName, maybeBillingAccountName, userInfo.userSubjectId)
-        // Since the billing account has been updated, any existing spend configuration is now out of date
-        _ <- dataAccess.rawlsBillingProjectQuery.clearBillingProjectSpendConfiguration(billingProjectName)
-        // if any workspaces failed to be updated last time, clear out the error message so the monitor will pick them up and try to update them again
-        _ <- dataAccess.workspaceQuery.deleteAllWorkspaceBillingAccountErrorMessagesInBillingProject(billingProjectName)
-        maybeBillingProject <- dataAccess.rawlsBillingProjectQuery.load(billingProjectName)
-      } yield maybeBillingProject
+      val F = Applicative[ReadWriteAction]
```
rtitle, line 659: I see what this is doing but with the Applicative syntax it feels less readable. Is there is a reason it can't be expressed as a for-comp as before?

rtitle (approved):
The changes generally look sane and the slick refactor seems like a nice improvement.

Changing all the slick/SQL in `RawlsBillingProjectComponent` makes me a little nervous, but I'm also not deeply familiar with it -- do  we feel like we have adequate test coverage of that stuff to be confident there are no regressions?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
ehigham: Nice!!! Thank you

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
ehigham: How would we use the timestamp other than max?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
ehigham: It does! But only defines for DBIO rather than the DBIOAction trait

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponentSpec.scala
```diff
@@ -58,43 +58,45 @@ class RawlsBillingProjectComponentSpec extends TestDriverComponentWithFlatSpecAn
...
+    withDefaultTestDatabase {
+      intercept[SQLException] {
+        runAndWait {
+          for {
+            _ <- rawlsBillingProjectQuery.updateBillingAccount(
+              testData.testProject1Name,
+              testData.billingProject.billingAccount,
+              testData.userOwner.userSubjectId
+            )
+          } yield fail("Should not allow billing accounts to be set to the same value.")
```
andy7i, line 79: I'm curious about why we need this `fail`. Do we need it for the SQLException to bubble up correctly?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
ehigham: Thanks but I don't think I can't claim the credit - I think this is how slick implements many of the operations on `Query`!

ehigham:
> The changes generally look sane and the slick refactor seems like a nice improvement.
> 
> Changing all the slick/SQL in `RawlsBillingProjectComponent` makes me a little nervous, but I'm also not deeply familiar with it -- do we feel like we have adequate test coverage of that stuff to be confident there are no regressions?

I think so. Most of the changes to the `RawlsBillingProjectComponent` are to pure functions where I've subusituted things like `filter(_.blah = blah.value)` - I think these are relatively safe.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
rtitle: Yeah I mean you'd do max(timestamp) as well… was just thinking if there's a situation where max(id) doesn't return the latest record (by timestamp) but I guess not. This seems fine

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -114,175 +170,183 @@ trait RawlsBillingProjectComponent {
...
+                          userSubjectId: RawlsUserSubjectId): ReadWriteAction[Int] =
+      for {
+        billingProjects <- query.read
+        count <- query.map(_.billingAccount).update(billingAccount.map(_.value))
+        // Record each billing account change
+        // - so the `WorkspaceBillingAccountActor` can synchronise the changes with google
+        // - to keep an audit log of billing account changes
+        _ <- DBIO.sequence(billingProjects.map { project =>
+          BillingAccountChanges.create(
+            project.projectName,
```
andy7i, line 337: I had some concerns that if this step failed (after the billing project table was updated), we could end up with billing projects that were updated but not logged in the BillingAccountChanges table. But since this is used within a transaction, we should be ok.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
andy7i: Could I get a quick explanation about what this is (is it Cats-ifying DBIOAction?) / why we need it (this sort of implicit is required for every type/supertype, right? and it sounds like one does not exist for DBIOAction?) ?

andy7i (approved):
the rawlsBillingProjectQuery refactor is quite _slick_!! I love how the code is much more readable.

I had a couple questions to help me better understand some things but not blocking.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponentSpec.scala
ehigham: It's more for documentation in code - ie execution should not arrive at this point. I didn't want some obscure error like "it did not throw" in the test output.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
ehigham: Yeah, this is the compromise we made by not doing this as a database hook. As you say, since this is in a transaction it should be fine. Then again, there's nothing stopping someone updating the billing project table directly.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MigrationUtils.scala
ehigham: I'll try :) Yes you're right - this is an "adapter" so that slick's DBIOActions can work with more general cats code. In cats-land, the type class implementation gets passed implicitly (hence this def needs to be implicit)
My goal was to support Applicative-style programming but I thought I'd do a little more while i was there! Why? so we can traverse data structures and sequence effects and not have to write if-then-else statements in for-comps.

andy7i: really interesting stuff. thanks for the explanation!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
ehigham: It could absolutely but I couldn't really find a satisfying way of doing this. The key difference here is that the original for-comp is being executed conditioanlly and wrapping that with another for-comp triggers a few irrational hates of mine!
- checking that an optional is defined in multiple places
- unnecesary re-load of object
- for-comp with two statements (dunno why, I just prefer bind/flatMap)

```scala
for {
  projectOpt <- load(billingProjectName)
  _ <- when(projectOpt.exists(_.billingAccount != newBillingAccount)) {  // <- check optional.isDefined here
   // original for-comp here
}
} yield projectOpt.map(_.copy(billingAccount = newBillingAccount)) // <- check optional.isDefined here
```

Does it feel sufficiently less-readable to you? If so, I'll get over myself and just do it haha!

PR #1747, opened 2022-04-28
[AJ-358] Add API to update attribute names

jordansafer:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

https://broadworkbench.atlassian.net/browse/AJ-358

Adding API to update Attribute names. The new API is:
PATCH /api/workspaces/{wsNamespace}/{wsName}/entityTypes/{entityType}/attributes/{attributeName}

The API supports renaming an attribute column. It:
1. Checks that the attribute exists
2. Checks that there are no conflicts with the new attribute name
3. Renames all attributes for the entityType to the new attributeName
4. Invalidates the entity statistics cache

Testing:
- Unit testing, Functional testing & local testing against a dev workspace


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -803,6 +804,22 @@ trait EntityComponent {
...
+
+    def renameAttribute(workspaceContext: Workspace,
+                         entityType: String,
+                         oldAttributeName: AttributeName,
+                         newAttributeName: AttributeName): ReadWriteAction[Int] = {
+      for {
+        numRowsRenamed <- entityAttributeShardQuery(workspaceContext).AttributeColumnQueries.renameAttribute(workspaceContext, entityType, oldAttributeName, newAttributeName)
+        _ <- workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
+      } yield numRowsRenamed
+    }
```
davidangb: how about putting these inside `AttributeComponent` instead of `EntityComponent`? These feel like they should be parallel to `deleteAttributes()`:
https://github.com/broadinstitute/rawls/blob/4247d58764d24e2b30acc5350d7fd3e67002b399/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala#L401-L403

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -803,6 +804,22 @@ trait EntityComponent {
         entityAttributeShardQuery(workspaceContext).deleteAttributes(workspaceContext, entityType, attributesNames)
     }
 
+    def doesAttributeNameAlreadyExist(workspaceContext: Workspace,
+                                      entityType: String,
+                                      newAttributeName: AttributeName): ReadAction[Option[Boolean]] = {
```
davidangb: nit: change argument `newAttributeName` to `attributeName` - this method doesn't care if it's looking up a "new" or "old" attribute

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -281,6 +282,44 @@ class EntityService(protected val userInfo: UserInfo, val dataSource: SlickDataS
...
+    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write, Some(WorkspaceAttributeSpecs(all = false))) flatMap { workspaceContext =>
+
+      def validateNewAttributeName(dataAccess: DataAccess,
+                                   workspaceContext: Workspace,
+                                   entityType: String,
+                                   attributeName: AttributeName): ReadAction[Boolean] = {
+        dataAccess.entityQuery.doesAttributeNameAlreadyExist(workspaceContext, entityType, attributeName) map {
+          case Some(false) => false
+          case Some(true) => throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Conflict,
+            s"${attributeName} already exists as an attribute name"))
```
davidangb: nit: use `${AttributeName.toDelimitedName(attributeName)}` in the error message instead of `${attributeName}` … the `toDelimitedName` will give a more user-friendly representation

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -1141,6 +1141,23 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+  it should "change the attribute name" in withDefaultTestDatabase {
+    withWorkspaceContext(testData.workspace) { context =>
+      val rowsUpdated = runAndWait(entityQuery.renameAttribute(context, "Pair",
+        AttributeName.withDefaultNS("case"), AttributeName.withDefaultNS("case2")))
+      assert(rowsUpdated == 2)
+      val exists = runAndWait(entityQuery.doesAttributeNameAlreadyExist(context, "Pair", AttributeName.withDefaultNS("case2"))).get
+      assert(exists)
+    }
+  }
+
```
davidangb: I would love to see a test that involves changing a non-default-namespace attribute name - possibly a test that changes both the namespace and the name.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -1141,6 +1141,23 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
     }
   }
 
+  it should "return false if the attribute name does not exist" in withMinimalTestDatabase { _ =>
+    withWorkspaceContext(testData.workspace) { context =>
+      val exists = runAndWait(entityQuery.doesAttributeNameAlreadyExist(context, "Pair", AttributeName.withDefaultNS("case2"))).get
+      assert(!exists)
+    }
+  }
```
davidangb: nit: these tests will result in output of:
```
[info] EntityComponentSpec:
[info] EntityComponent
[info] - should return false if the attribute name does not exist
[info] - should change the attribute name
```
which isn't too helpful for anyone reading the test results. Either add more info into the test name itself, like:
```
it should "return false from doesAttributeNameAlreadyExist if the attribute name does not exist" in withMinimalTestDatabase { _ =>
```
or get all Scalatest on it and do something like:
```
behavior of "doesAttributeNameAlreadyExist()"

it should "return false if the attribute name does not exist" in withMinimalTestDatabase { _ =>
```
but note that if you set `behavior of ` it will affect ALL tests underneath this one, so that's a pain in the butt

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityServiceSpec.scala
```diff
@@ -262,6 +262,36 @@ class EntityServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matcher
...
+  it should "rename an attribute name as long as the selected name is not in use"  in withTestDataServices { services =>
+    val oldAttributeName = AttributeName.withDefaultNS("case")
+    val waitDuration = Duration(10, SECONDS)
+    assertResult(2) {Await.result(services.entityService.renameAttribute(testData.wsName, testData.pair1.entityType,
+      oldAttributeName, AttributeRename(AttributeName.withDefaultNS("newAttributeName"))), waitDuration)}
+
+    // verify there are no longer any attributes under the old attribute name
+    val queryResult = Await.result(services.entityService.listEntities(testData.wsName, testData.pair1.entityType), waitDuration)
+    queryResult map { entity =>
+      assert(!entity.attributes.keySet.contains(oldAttributeName))
```
davidangb: nit: if this test fails, it will result in an error message of "true was not false" or something unhelpful like that, because you're just asserting a boolean value. To get better error messages, try something like:
```scala
      withClue(s"entity ${entity.entityType}/${entity.name} attributes ") {
        entity.attributes.keySet should not contain oldAttributeName
      }
```
using [matchers](https://www.scalatest.org/user_guide/using_matchers) like this helps with test debugging after the fact!

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1868,144 +1868,53 @@ paths:
...
-          content:
-            'application/json':
-              schema:
-                $ref: '#/components/schemas/ValidatedMethodConfiguration'
+        204:
+          description: Attribute name updated
+          content: { }
         403:
-          description: Invalid attribute namespace(s) in outputs
+          description: User does not have permission to update Attributes in the namespace
```
davidangb: nit: remove "in the namespace" here, that will be confusing to most users. How about "User does not have permission to update attribute names"?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1868,144 +1868,53 @@ paths:
...
-          "methodUri": "dockstore://path/version"
-        }
-        ```
-        The system is specified to check for a URI first before falling back to the legacy fields. Unsupported repos will return a 400 Bad Request.
-      operationId: create_method_configuration
+        - entities
+      summary: Change attribute name for an entity type
+      description: Change the attribute name for an entity type in a workspace.
+        The attribute must already exist, and the new name cannot exist. If there
+        is a conflict (between the new name and another name), the update will fail,
```
davidangb: nit: how about "If the new name does already exist, the update will fail …."

davidangb (commented):
some comments inline!

As far as API signature, the choice of
```
PATCH /api/workspaces/{workspaceNamespace}/{workspaceName}/entityTypes/{entityType}/attributes/{attributeName}
```
looks quite valid on its own, but is inconsistent with
```
POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameEntityType
```

I'd really like these two APIs to be consistent. I don't think it's too late to change renameEntityType, or we could change this PR … thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jordansafer: That makes sense! I was missing the fact that I should invalidate the statistics cache from EntityComponent, while keeping all the attribute logic in AttirbuteComponent-

sagehen03 (approved):
I don't see any errors.  Agree with David on adding a reason/explanation around failed assertions.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jordansafer: I made this change, but looking more closely at this, I would push back, I think this feels like an unneeded layer of boilerplate, and I would push to get rid of this for the deleteAttributes API too (effectively we're creating wrapper functions that don't perform any action)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -1141,6 +1141,33 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+
+  it should "change the attribute name and namespace when renameAttribute is called with an attribute with a new namespace" in withDefaultTestDatabase {
+    withWorkspaceContext(testData.workspace) { context =>
+      val rowsUpdated = runAndWait(entityQuery.renameAttribute(context, "Pair",
+        AttributeName.withDefaultNS("case"), AttributeName.fromDelimitedName("new_namespace:new_name")))
+      assert(rowsUpdated == 2)
+      val exists = runAndWait(entityQuery.doesAttributeNameAlreadyExist(context, "Pair", AttributeName.fromDelimitedName("new_namespace:new_name"))).get
+      assert(exists)
+    }
+  }
```
davidangb: oooh this raises a good point: we currently have a list of allowed namespaces, and reject any others. I question whether or not we want to keep that allow-list but that's out of scope for this PR … which means you'll need to validate the new namespace before proceeding with any renaming. `withAttributeNamespaceCheck` should handle this for you: https://github.com/broadinstitute/rawls/blob/8de18c92666a364ad2866714833057068bded6d1/core/src/main/scala/org/broadinstitute/dsde/rawls/util/AttributeSupport.scala#L23

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I see your point about extra boilerplate. The `renameAttribute` wrapper adds value - in that it updates the workspace's last-modified - but the `doesAttributeNameAlreadyExist` wrapper doesn't add anything. I'd be fine with deleting that.

There's no particular reason to only update the workspace's last-modified from EntityComponent as opposed to AttirbuteComponent.

In this PR as-is, we have the wrapper functions in BOTH EntityComponent and AttirbuteComponent, and the renameAttribute() implementation is different between the two; we definitely need to choose one. My suggestions is still to put the full renameAttribute() in AttributeComponent, and either put doesAttributeNameAlreadyExist() in AttributeComponent or delete it entirely. Maybe this needs a f2f chat, I'm not 100% sure I understand all of your reasoning.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -190,6 +190,16 @@ trait EntityApiService extends UserInfoDirectives {
...
+        } ~
+        path("workspaces" / Segment / Segment / "entityTypes" / Segment / "renameAttribute") { (workspaceNamespace, workspaceName, entityType) =>
+          post {
+            entity(as[AttributeRename]) { attributeRenameRequest =>
+              complete {
+                entityServiceConstructor(userInfo).renameAttribute(WorkspaceName(workspaceNamespace, workspaceName),
+                  entityType, attributeRenameRequest).map(_ => StatusCodes.NoContent)
+              }
+            }
+          }
```
davidangb, line 202: I'm gonna sleep on this API signature. IMHO it's still inconsistent with rename-entity-type, so maybe we should just move that other API …

davidangb (commented):
A few more pieces of feedback, I want to think a bit more about the API signature and we may need to chat in person, but want to get these comments out without waiting for either of those!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jordansafer: Ah, I see where we're disconnected. You initially referenced the "deleteAttributes" method, which consists of a "deleteAttributes" wrapper function in the EntityComponent that invalidates the EntityStatistics cache, and has another wrapper in the AttributeComponent, which doesn't add logic (the excerpt you have linked is the wrapper in AttributeComponent which doesn't add any logic). 

I was a bit confused about the need for the additional wrapper, but I followed the "deleteAttributes" pattern and updated the PR so each function had 2 wrappers, one in EntityComponent, and one in AttributeComponent. 

I added the comment pushing back because I wasn't feeling good about all the empty logic wrappers in the AttributeComponent. I'm a big fan of doing away with the wrappers, except the initial 2 wrappers from the first revision of the PR. This will be inconsistent with the DeleteAttributes pattern, which is invalidating the EntityStatistics cache in the EntityComponent, but I'll just move that over.

Happy to talk more about this face to face-

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
jordansafer: I think what you're saying is that the "entityTypes"/ \<entityType\> part of the path is inconsistent with the "entities" / \<entityType\> part of the other API. I'm strongly in favor of the "entityTypes" approach since we're not overloading the "entities" / \<entityType\> / \<entityName\> path, and thus not introducing potential edge cases about what an entity name can be. I'll update the other API signature to match-

As far as POST over PATCH, passing both old & new attributes in the request body, the "verb" at the end of the path- I believe both signatures are consistent in all of those regards.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
jordansafer: Adding this-

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1394,7 +1394,7 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
       x-codegen-request-body-name: newEntityNameJson
-  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/renameEntityType:
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/renameEntityType:
```
davidangb: if we're gonna move this one, how about moving it to
```
/api/workspaces/{workspaceNamespace}/{workspaceName}/entityTypes/{entityType}/renameEntityType
```
to make its URL exactly parallel to `renameAttribute`? I know that would require changing the payload of that other API as well.

davidangb (commented):
one more question/thought about API signature consistency, otherwise :+1:

jordansafer:
Jenkins error: "sbt.librarymanagement.ResolveException: Error downloading com.readytalk:metrics-statsd-common:4.2.0" 
Looks like this dependency isn't in our repositories. This looks like a recurring issue :/

jordansafer:
jenkins retest

davidangb (approved):
:+1: once tests pass, thanks for slogging all the way through this!

PR #1753, opened 2022-05-02
[IA-3326] Add role/lifesciences.workflowsRunner to canCompute policy

gpcarr:
Part of enabling Nextflow in terra

see Jira ticket: https://broadworkbench.atlassian.net/browse/IA-3326

Tested by running rawls locally:
Screen Shot 2022-05-13 at 10 21 21 AM

Screen Shot 2022-05-13 at 10 11 05 AM

Tested together with https://github.com/broadinstitute/sam/pull/711 

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2158,9 +2158,9 @@ class WorkspaceService(protected val userInfo: UserInfo,
   // todo: update this line as part of https://broadworkbench.atlassian.net/browse/CA-1220
   // This is done sequentially intentionally in order to avoid conflict exceptions as a result of concurrent IAM updates.
     List(
-      billingProjectOwnerPolicyEmail -> Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole),
-      policyEmailsByName(SamWorkspacePolicyNames.owner) -> Set(terraWorkspaceCanComputeRole),
-      policyEmailsByName(SamWorkspacePolicyNames.canCompute) -> Set(terraWorkspaceCanComputeRole)
+      billingProjectOwnerPolicyEmail -> Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole, "roles/lifesciences.workflowsRunner"),
+      policyEmailsByName(SamWorkspacePolicyNames.owner) -> Set(terraWorkspaceCanComputeRole, "roles/lifesciences.workflowsRunner"),
+      policyEmailsByName(SamWorkspacePolicyNames.canCompute) -> Set(terraWorkspaceCanComputeRole, "roles/lifesciences.workflowsRunner")
     )
       .traverse_ { case (email, roles) =>
         googleIamDao.addIamRoles(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -66,7 +66,7 @@ object WorkspaceService {
                   config: WorkspaceServiceConfig, requesterPaysSetupService: RequesterPaysSetupService,
                   entityManager: EntityManager, resourceBufferService: ResourceBufferService, resourceBufferSaEmail: String,
                   servicePerimeterService: ServicePerimeterService,
-                  googleIamDao: GoogleIamDAO, terraBillingProjectOwnerRole: String, terraWorkspaceCanComputeRole: String)
+                  googleIamDao: GoogleIamDAO, terraBillingProjectOwnerRole: String, terraWorkspaceCanComputeRole: String, terraWorkspaceNextflowRole: String)
                  (userInfo: UserInfo)
                  (implicit materializer: Materializer, executionContext: ExecutionContext): WorkspaceService = {
 
@@ -77,7 +77,7 @@ object WorkspaceService {
       genomicsServiceConstructor, maxActiveWorkflowsTotal,
       maxActiveWorkflowsPerUser, workbenchMetricBaseName, submissionCostService,
       config, requesterPaysSetupService, resourceBufferService, resourceBufferSaEmail, servicePerimeterService,
-      googleIamDao, terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole)
+      googleIamDao, terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole, terraWorkspaceNextflowRole)
   }
 
   val SECURITY_LABEL_KEY = "security"
@@ -139,7 +139,8 @@ class WorkspaceService(protected val userInfo: UserInfo,
                        servicePerimeterService: ServicePerimeterService,
                        googleIamDao: GoogleIamDAO,
                        terraBillingProjectOwnerRole: String,
-                       terraWorkspaceCanComputeRole: String)
+                       terraWorkspaceCanComputeRole: String,
+                       terraWorkspaceNextflowRole: String)
                       (implicit protected val executionContext: ExecutionContext) extends RoleSupport
   with LibraryPermissionsSupport
   with FutureSupport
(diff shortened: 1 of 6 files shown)


Qi77Qi (approved):
seems reasonable to me

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2160,7 +2160,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
     List(
       billingProjectOwnerPolicyEmail -> Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole),
       policyEmailsByName(SamWorkspacePolicyNames.owner) -> Set(terraWorkspaceCanComputeRole),
-      policyEmailsByName(SamWorkspacePolicyNames.canCompute) -> Set(terraWorkspaceCanComputeRole)
+      policyEmailsByName(SamWorkspacePolicyNames.canCompute) -> Set(terraWorkspaceCanComputeRole, "roles/lifesciences.workflowsRunner")
```
gpcarr: I only added `roles/lifesciences.workflowsRunner` to the `canCompute` policy. I would like to understand the difference between these three policies better (`SamWorkspacePolicyNames.owner`, `SamWorkspacePolicyNames.canCompute` and `billingProjectOwnerPolicyEmail`). Would I need to add it to `billingProjectOwnerPolicyEmail` as well? cc @rtitle @dvoet

rtitle: So the keys in this map represent [Sam roles](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L120) on the workspace resource: billing project owner, workspace owner, can-compute user. The values in this map represent the GCP IAM roles that we set in the workspace's googleProject. This is essentially saying we give project owners, workspace owners, and workspace "can-compute" users the "canCompute" custom role in the workspace's googleProject.

Based on this I think we would want to add `roles/lifesciences.workflowsRunner` to all 3 of these, assuming we want to give all 3 Sam roles the ability to launch PAPI jobs in the workspace.

Also I'm wondering if we want to put this behind a boolean config in Rawls conf. The reason to add this to Rawls (as opposed to the custom GCP role) is the flexibility to turn if off/on for specific workspaces. Maybe it would be good to start with a `addLifeSciencesWorkflowsRunnerRole` flag for now, and we could fine-tune later if necessary.

gpcarr: that makes sense and based on what the 3 sam roles are, i agree that we would want to add to all three.

Also I'm wondering if we want to put this behind a boolean config in Rawls conf. The reason to add this to Rawls (as opposed to the custom GCP role) is the flexibility to turn if off/on for specific workspaces. Maybe it would be good to start with a addLifeSciencesWorkflowsRunnerRole flag for now, and we could fine-tune later if necessary.

Not entirely sure I'm following. Adding the role to the Rawls conf would mean removing it from where I'm currently adding it now so that it doesn't always get added. What I'm not sure about is how to do this conditionally based on the Rawls config

Qi77Qi: Idk if using a conf flag will help much tho, cuz conf flag can only be flipped on/off for all in prod. This condition won't be the same as criteria for enabling it for some workspaces and not others (just curious what would this criteria be?)...

rtitle: I guess I don't feel strongly on the config flag if it will just be on/off for now. Maybe just add a comment in the Rawls code for now, explaining that we're setting the lifesciences role separate from the custom role to give us flexibility in case we need to fine-tune it in the future depending on data policies. (For example if a dataset has a regionality requirement then we would _not_ want to grant this, as it would allow users to launch compute in any region.)

aherbst-broad: FWIW there are workspace level [flags](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceFeatureFlagComponent.scala) now.

aherbst-broad (approved):

rtitle (approved):
The change LGTM, would still like a comment explaining why the lifesciences role are outside the can-compute policy. Something like:
```
// Add lifesciences.workflowsRunner outside of the canCompute policy to give the flexibility 
// to fine-tune which workspaces it's added to if needed. This role gives the user the ability 
// to launch compute in any region, which may be counter to some data regionality policies.
```

gpcarr:
> The change LGTM, would still like a comment explaining why the lifesciences role are outside the can-compute policy. Something like:
> 
> ```
> // Add lifesciences.workflowsRunner outside of the canCompute policy to give the flexibility 
> // to fine-tune which workspaces it's added to if needed. This role gives the user the ability 
> // to launch compute in any region, which may be counter to some data regionality policies.
> ```

meant to add a comment, will do that now!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2158,9 +2158,9 @@ class WorkspaceService(protected val userInfo: UserInfo,
   // todo: update this line as part of https://broadworkbench.atlassian.net/browse/CA-1220
   // This is done sequentially intentionally in order to avoid conflict exceptions as a result of concurrent IAM updates.
     List(
-      billingProjectOwnerPolicyEmail -> Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole),
-      policyEmailsByName(SamWorkspacePolicyNames.owner) -> Set(terraWorkspaceCanComputeRole),
-      policyEmailsByName(SamWorkspacePolicyNames.canCompute) -> Set(terraWorkspaceCanComputeRole)
+      billingProjectOwnerPolicyEmail -> Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole, "roles/lifesciences.workflowsRunner"),
```
andy7i: Do we think this is the only role that NextFlow will need (and we won't need to add additional roles or remove this role in the future)?

If yes:
I think the current implementation is ok.

If no or not sure:
I don't have a lot of context around NextFlow, but if we expect that things may change, then it will be easier manage this permission (and other potential permissions for NextFlow) via a custom IAM role.

The benefits of using a custom IAM role, assuming we change roles for Nextflow in the future:
* Eliminates the need for a migration to update each workspace. Instead, we just need to update the custom IAM role once (per Org, for example test.firecloud.org). Adding or removing roles is much simpler this way.
* If there are multiple roles added for NextFlow, they are conveniently grouped within the custom IAM role.

andy7i (commented):
+1 to Rob's comment about adding a comment

looking for thoughts on using a custom IAM role

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
rtitle: That's a good point @andy7i. To my knowledge `lifesciences.workflowsRunner` is all we expect to need for NextFlow, but I'll let @gpcarr confirm.

gpcarr: We talked offline yesterday and looks like it will be best to add a custom role. I merged these two PRs doing that and am working on integrating the new nextflow custom role with this rawls PR and that'll be the last step:
- https://github.com/broadinstitute/terraform-terra/pull/162 merged and ran the script in all three orgs with @andy7i 
- https://github.com/broadinstitute/firecloud-develop/pull/2924 merged and pulls new nextflow custom role into rawls conf

gpcarr:
@rtitle @andy7i @Qi77Qi looking for re-review since the implementation was changed! i updated the description with how this was tested and the relevant terraform-terra and firecloud-develop PRs that were merged to get this config in rawls are here: 
- https://github.com/broadinstitute/terraform-terra/pull/164 (we only ran the [create custom iam role for org script](https://github.com/broadinstitute/terraform-terra/blob/master/scripts/project-setup-scripts/create-custom-iam-roles-for-projects.sh) so this will still only be added to new projects after this rawls PR is merged) 
- https://github.com/broadinstitute/firecloud-develop/pull/2924

Qi77Qi (approved):
change lgtm...why did we add a custom nextflow role btw?

gpcarr:
> change lgtm...why did we add a custom nextflow role btw?

for a few reasons
- it eliminates the need for a migration to update each workspace. Instead, we just need to update the custom IAM role once (per Org, for example test.firecloud.org). Adding or removing roles is much simpler this way
- There are[ two roles](https://github.com/broadinstitute/terraform-terra/blob/master/scripts/project-setup-scripts/terra-workspace-nextflow-role.yaml) we're adding for NextFlow and they are conveniently grouped within the custom IAM role. 
-  we aren't completely certain on how nextflow use cases might change or how we may iterate on nextflow in the future so i thought this was the best path forward in case we end up needing to make any changes to the roles needed for nextflow

gpcarr:
jenkins retest

andy7i (approved):
LGTM 👍

gpcarr:
jenkins retest

PR #1761, opened 2022-05-05
 AJ-353: Try to improve workspace clone db operations

sagehen03:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


sagehen03:
I haven't been able to verify that these changes reduce memory consumption, but they _should_ since cloning entities should no longer involve bringing entities into Rawls' heap.  If anyone has thoughts on how best to do a before/after comparison for memory, let me know.  (I tried some log lines using Runtime.total|freeMemory()) I timed cloning a largish workspace (600k entity attribute records and 50k entity records) and with these changes it went from 149s to 23s.

Also, these changes just cover copying entity, entity attribute, and method config records.  Tests passed for me and it didn't look like the previous db code was doing anything else, but scala consistently surprises me so experienced folks please make sure I didn't miss any additional data that needs to come over to the new workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -797,17 +797,11 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
-                            dataAccess.entityQuery.copyAllEntities(sourceWorkspaceContext, destWorkspaceContext) andThen
-                              dataAccess.methodConfigurationQuery.listActive(sourceWorkspaceContext).flatMap { methodConfigShorts =>
-                                val inserts = methodConfigShorts.map { methodConfigShort =>
-                                  dataAccess.methodConfigurationQuery.get(sourceWorkspaceContext, methodConfigShort.namespace, methodConfigShort.name).flatMap { methodConfig =>
-                                    dataAccess.methodConfigurationQuery.create(destWorkspaceContext, methodConfig.get)
-                                  }
+                              dataAccess.entityQuery.cloneEntitiesToNewWorkspace(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID) andThen  {
+                                dataAccess.methodConfigurationQuery.clone(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID) flatMap {
+                                  _ => DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
```
sagehen03: Not totally sure what I did here.  This was a try until it seems to be working. Experienced scala eyes please scrutinize.

sagehen03:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -159,6 +159,23 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
     }
   }
 
+  it should "copy entities and attributes to new workspace for clone" in withDefaultTestDatabase {
```
sagehen03: Not sure whether we should keep this test.  I wrote this to verify my code in isolation initially, but WorkspaceServiceSpec might already cover things?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -480,6 +480,41 @@ trait EntityComponent {
...
+                       old_entity.list_index, new_entity.id, old_entity.list_length, old_entity.namespace,
+                       old_entity.VALUE_JSON, old_entity.deleted, old_entity.deleted_date
+                from ENTITY new_entity
+                JOIN (select ea.*, e.name as entity_name, e.entity_type from ENTITY_ATTRIBUTE_#$sourceShardId ea join ENTITY e
+                    on ea.owner_id = e.id where e.workspace_id = $clonedWorkspaceId and e.deleted = 0 and ea.deleted = 0) old_entity
+                on new_entity.name = old_entity.entity_name and new_entity.entity_type = old_entity.entity_type
+                where new_entity.workspace_id = $newWorkspaceId;
+          """
+      }
+    }
```
davidangb, line 562: in both of these queries, you could skip selecting `deleted, deleted_date` from the source tables, and hardcode inserts of `0, null` into those columns in the destination tables. Since you have `where deleted=0` in the where clause this should always be safe.

That said, this _might_ be a micro-optimization, but if we're working on a large number of rows, it could cut down the size of data mysql needs to process.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
davidangb: I poked around a bit and what I see is the important test is further down, here in `EntityComponentSpec`: "clone all entities from a workspace containing cycles". Note that test uses `copyAllEntities()` so you'd have to update it, unless you take over that method name.

If you decide to keep this test too, it would be good to verify the contents of the entities against each other, instead of just checking the counts. Doing that would make the "We should clone attributes as well as entities" assertion obselete.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -480,6 +480,41 @@ trait EntityComponent {
...
+                SELECT old_entity.name, old_entity.value_string,
+                       old_entity.value_number, old_entity.value_boolean, old_entity.value_entity_ref,
+                       old_entity.list_index, new_entity.id, old_entity.list_length, old_entity.namespace,
+                       old_entity.VALUE_JSON, old_entity.deleted, old_entity.deleted_date
+                from ENTITY new_entity
+                JOIN (select ea.*, e.name as entity_name, e.entity_type from ENTITY_ATTRIBUTE_#$sourceShardId ea join ENTITY e
+                    on ea.owner_id = e.id where e.workspace_id = $clonedWorkspaceId and e.deleted = 0 and ea.deleted = 0) old_entity
+                on new_entity.name = old_entity.entity_name and new_entity.entity_type = old_entity.entity_type
+                where new_entity.workspace_id = $newWorkspaceId;
+          """
```
davidangb, line 559: you'll need to do a similar mapping of old-entity-id to new-entity-id inside the attributes' `value_entity_ref` column. Else, any entity references that get copied will still be pointing at the old entities in the source workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: What you have here looks pretty good. Keeping this syntax, my only suggestion is to replace the last `flatMap` with another `andThen` for consistency, as below. This is probably the best approach to take, since this area of the code relies on lots of indenting.
```scala                              
 dataAccess.entityQuery.cloneEntitiesToNewWorkspace(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID) andThen  {
     dataAccess.methodConfigurationQuery.clone(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID) andThen {
         DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
     }
 }
```

But personally, I don't like all the indenting and I find `for … yield` easier to read. Here's what that code might look like:
```scala
 for {
     _ <- dataAccess.entityQuery.cloneEntitiesToNewWorkspace(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID)
     _ <- dataAccess.methodConfigurationQuery.clone(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID)
 } yield {
     DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
 }
```

a final option would be to use `DIBIO.seq`, like so:
```scala              
 val copyOperations =  DBIO.seq(dataAccess.entityQuery.cloneEntitiesToNewWorkspace(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID),
     dataAccess.methodConfigurationQuery.clone(sourceWorkspaceContext.workspaceIdAsUUID, destWorkspaceContext.workspaceIdAsUUID))
                            
 copyOperations andThen
     DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
```

these are all pretty much equivalent. There's possibly some miniscule advantage to the `DBIO.seq` approach because that doesn't bother storing the result of any of its actions, whereas the `andThen`s in the first approach store a result … but the result sets are super small anyway. I'd pick whichever you think makes the code most readable.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -797,17 +797,11 @@ class WorkspaceService(protected val userInfo: UserInfo,
                         val newAttrs = sourceWorkspaceContext.attributes ++ destWorkspaceRequest.attributes
                         traceDBIOWithParent("withNewWorkspaceContext (cloneWorkspace)", parentSpan) { s1 =>
                           withNewWorkspaceContext(destWorkspaceRequest.copy(authorizationDomain = Option(newAuthDomain), attributes = newAttrs), destBillingProject, sourceBucketNameOption, dataAccess, s1) { destWorkspaceContext =>
-                            dataAccess.entityQuery.copyAllEntities(sourceWorkspaceContext, destWorkspaceContext) andThen
```
davidangb, line 801: with this PR, `entityQuery.copyAllEntities` is no longer used (except in a unit test, see other comment). Could you delete it? Alternately, keep `copyAllEntities` but replace its implementation with your implementation.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/MethodConfigurationComponent.scala
```diff
@@ -188,12 +188,28 @@ trait MethodConfigurationComponent {
...
+      val driver: JdbcProfile = MethodConfigurationComponent.this.driver
+
+      def cloneMethodConfigs(srcWsId: UUID, destWsId: UUID): WriteAction[Int] = {
+
+        sqlu"""insert into METHOD_CONFIG (NAMESPACE, NAME, WORKSPACE_ID, ROOT_ENTITY_TYPE, METHOD_URI,
+               DELETED, METHOD_CONFIG_VERSION, DATA_REFERENCE_NAME) select NAMESPACE, NAME,
+               $destWsId, ROOT_ENTITY_TYPE, METHOD_URI, 0, METHOD_CONFIG_VERSION,
+               DATA_REFERENCE_NAME from METHOD_CONFIG where WORKSPACE_ID=$srcWsId and deleted = 0"""
+      }
+    }
```
davidangb: Awesome on seeing the opportunity here and forging ahead with optimizing this too!

Unfortunately I don't think this query is sufficient. It is failing to copy rows from the `METHOD_CONFIG_INPUT` and `METHOD_CONFIG_OUTPUT` tables.

My suggestion is to remove the method config optimization from this PR and tackle it in a follow-on, to unblock merging of the entity optimizations - but if you'd rather tackle it all in one, great.

Have you looked at test coverage for this? (I don't know much about method config coverage)

davidangb (commented):
see comments inline.

I'd love to follow this up with a future PR that applies the same technique to the `[/api/workspaces/entities/copy](https://rawls.dsde-dev.broadinstitute.org/#/entities/copy_entities)` API too - I think it could reuse a lot of the same code.

davidangb:
Oh also - doc on combining db actions: https://scala-slick.org/doc/3.3.3/dbio.html#sequential-execution

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
sagehen03: Cool, I appreciate these explanations!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: Good catch! ty

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -643,7 +643,7 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
           val c3_updated = Entity("c3", "samples", Map(AttributeName.withDefaultNS("foo") -> AttributeString("x"), AttributeName.withDefaultNS("bar") -> AttributeNumber(3), AttributeName.withDefaultNS("cycle3") -> AttributeEntityReference("samples", "c1")))
 
           runAndWait(entityQuery.save(originalContext, c3_updated))
-          runAndWait(entityQuery.copyAllEntities(originalContext, cloneContext))
+          runAndWait(entityQuery.cloneEntitiesToNewWorkspace(originalContext.workspaceIdAsUUID, cloneContext.workspaceIdAsUUID))
```
davidangb, line 645: looking at this test with a close eye, I do NOT believe this test actually verifies that the newly cloned entity references are actually referencing the new workspace's entities vs. the old workspace's entities. This is a fault of the original code, not your PR … but it also means we can't count on this test to verify cloning of references.

The objects actually being compared are `Set[Entity]`; those contain `AttributeEntityReference` and `AttributeEntityReference` only contains an entityType and an entityName, but it doesn't verify which workspace it's referencing. Similarly, I think that the code in `listActiveEntities()` doesn't limit references to referencing their own workspace.

Could you drop to plain sql to verify these rows, or if you'd rather use Slick (I bet you don't), might need to use lower-level methods on `entityAttributeShardQuery` to verify.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: Phew this is a thorny query! I know you said you tried a few different approaches and this is the most efficient one. I'll toss out one approach - if you tried it and eliminated it, all good. What about creating a temporary table with two columns: oldEntityId, newEntityId (based on your existing self-join on the entity type&name), then using that temp table to fill in the proper ids for the main INSERT INTO … SELECT statement, both for ENTITY_ATTRIBUTE.owner_id and for  ENTITY_ATTRIBUTE. value_entity_ref ?

sagehen03: I've changed the query and I think it's a little bit clearer.  I took your suggestion about creating a dedicated `existing entity id -> new cloned entity id` table and the above is what ultimately fell out of it.  Initially I tried using `create temporary table` as you were suggesting but you [can't reference a temporary table multiple times under different aliases](https://dev.mysql.com/doc/refman/8.0/en/temporary-table-problems.html) and we need to join back to that mapping twice to change both the `owner_id` and `value_entity_ref` so I'd have to create two temporary tables and that all is more verbose than what I ended up with.  

The derived tables `owner_entity` and `referenced_entity` are the same sql, but the same issue arises where I can't join to owner_entity twice for example.  I experimented a bit with trying to extract 
```
select old_entity.id as old_id, new_entity.id as new_id from ENTITY old_entity
                               join ENTITY new_entity on new_entity.entity_type = old_entity.entity_type
                               and new_entity.name = old_entity.name
                               and new_entity.workspace_id = $newWorkspaceId
                               and old_entity.workspace_id = $clonedWorkspaceId
```
to a variable but the parameters in that sql string complicates things.  I think we could probably do something similar to https://github.com/broadinstitute/rawls/blob/e1348ab636351f3f2799c73b6de6b3d0ab40e60a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L229 but I think that approach requires something like `UNHEX(${clonedWorkspaceId.toPrettyString})` so as to get a sql string that no longer has parameters.  That didn't seem like a worthwhile tradeoff to me.  I also tried escaping the $ in the extracted string with $$ but that didn't work for me either. @davidangb if you know of a clean way to make that sql a bit DRYer that might be worthwhile, but I also like the fact that a developer can more quickly see the query being used.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -15,6 +15,7 @@ import java.util.UUID
  */
 class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers with RawlsTestUtils {
   import driver.api._
+  implicit object SetUUIDParameter extends SetParameter[UUID] { def apply(v: UUID, pp: PositionedParameters) { pp.setBytes(uuidColumnType.toBytes(v)) } }
```
davidangb: instead of duplicating this via copy/paste, how about adding ` with RawSqlQuery` to the `EntityComponentSpec` definition at line 16? (this would allow you to remove the extra imports at line 9 too)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: To DRY this and store the subquery in a reusable variable, I think you'd need to rely on `concatSqlActions`. There are a bunch of examples of this in the codebase, you'd end up with something like:
```scala
        val entityIdMappingOldToNew = sql"""(select old_entity.id as old_id, new_entity.id as new_id from ENTITY old_entity
                                               join ENTITY new_entity on new_entity.entity_type = old_entity.entity_type
                                               and new_entity.name = old_entity.name
                                               and new_entity.workspace_id = $newWorkspaceId
                                               and old_entity.workspace_id = $clonedWorkspaceId)"""

        concatSqlActions(
          sql"""insert into ENTITY_ATTRIBUTE_#$destShardId (name, value_string, value_number, value_boolean, value_entity_ref,
                list_index, owner_id, list_length, namespace, VALUE_JSON, deleted, deleted_date)
                select ea.name, value_string, value_number, value_boolean, referenced_entity.new_id as value_entity_ref, list_index, owner_entity.new_id as owner_id,
                list_length, namespace, VALUE_JSON, 0, null from ENTITY_ATTRIBUTE_#$sourceShardId ea
                    join """, entityIdMappingOldToNew, sql""" owner_entity
                    on ea.owner_id = owner_entity.old_id
                    left join """, entityIdMappingOldToNew, sql""" referenced_entity
                    on ea.value_entity_ref = referenced_entity.old_id
                join ENTITY e on e.id = ea.owner_id where ea.deleted = false and e.deleted = false and e.workspace_id = $clonedWorkspaceId
          """
        ).as[Int].head
```

which is arguably not any easier to read. Maybe just add code comments explaining the situation? I am fine with either implementation, what you have in place right now looks pretty good and readable.

davidangb (commented):
minor comment inline about the `SetUUIDParameter`. Looking good  - I will :+: after rebase and making sure it passes tests!

jordansafer:
> I haven't been able to verify that these changes reduce memory consumption, but they _should_ since cloning entities should no longer involve bringing entities into Rawls' heap. If anyone has thoughts on how best to do a before/after comparison for memory, let me know. (I tried some log lines using Runtime.total|freeMemory()) I timed cloning a largish workspace (600k entity attribute records and 50k entity records) and with these changes it went from 149s to 23s.

Is the test in a local rawls? I would go for a JVM profiler like JVisualVM, but I haven't tested with Scala (I think it should work fine though).

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jordansafer: +1 to concatSqlActions. That, plus the comments makes it quite a bit easier to read-

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -807,6 +851,15 @@ trait EntityComponent {
       }
     }
 
+    def cloneEntitiesToNewWorkspace(sourceWs: UUID, destWs: UUID): WriteAction[Int] = {
+      for {
+        entitiesCopied <- CopyEntitiesQuery.copyEntities(sourceWs, destWs)
+        _ <- CopyEntityAttributesQuery.copyAllAttributes(sourceWs, destWs)
+      } yield {
+        entitiesCopied
```
jordansafer: Could we return the counts for both entitiesCopied and attributesCopied? It could be as a pair (entitiesCopied, attributesCopied)

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -652,6 +651,21 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
...
+                        join ENTITY e_ref on ea.value_entity_ref = e_ref.id
+                        where ea.value_entity_ref is not null and e_ref.workspace_id != $destWsId and e.workspace_id = $destWsId"""
+          assertResult(0, "cloned entity references should only point to entities within the same workspace") {
+            runAndWait(badEntityReferenceCount.as[Int].head)
+          }
+
+          val expectedEntityReferenceCount = sql"""select count(*) from ENTITY e join ENTITY_ATTRIBUTE_#$destShardId ea on e.id = ea.owner_id
+                        join ENTITY e_ref on ea.value_entity_ref = e_ref.id
+                        where ea.value_entity_ref is not null and e_ref.workspace_id = $destWsId and e.workspace_id = $destWsId"""
+          assertResult(3, "cloned entity references should only point to entities within the same workspace") {
```
jordansafer, line 667: nit: I think it would be helpful if this error message was slightly different to highlight the success case. How about something like "all entity references should be cloned to the destination workspace. The entity reference count in the destination workspace does not match the initial count from the source workspace". Err my error is a little wordy, just the "all entity references should be clone to the destination workspace" would be good

jordansafer (approved):

davidangb (approved):

PR #1776, opened 2022-05-10
[SU-46] Delete entity type endpoint

MatthewBemis:
This new endpoint allows for entire tables to be deleted, assuming that there are no references to the entities in that table.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -836,9 +836,9 @@ trait EntityComponent {
 
     // "deletes" entities of a certain type by hiding and renaming them
     def hideType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Int] = {
-      workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID) andThen
-        EntityAndAttributesRawSqlQuery.batchHideAttributesOfType(workspaceContext, entityType) andThen
-        EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum)
+      EntityAndAttributesRawSqlQuery.batchHideAttributesOfType(workspaceContext, entityType) andThen
+        EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum) andThen
+        workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
     }
 
     // perform actual deletion (not hiding) of all entities in a workspace
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -184,6 +184,14 @@ trait EntityComponent {
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
       }
 
+      def batchHideType(workspaceId: UUID, entityType: String): ReadWriteAction[Seq[Int]] = {
+        val renameSuffix = "_" + getSufficientlyRandomSuffix(1000000000) // 1 billion
+        val deletedDate = new Timestamp(new Date().getTime)
+        // issue bulk rename/hide for all entities of the specified type
+        val typeUpdate = sql"""update ENTITY set deleted=1, deleted_date=$deletedDate, name=CONCAT(name, $renameSuffix) where deleted=0 AND workspace_id=$workspaceId and entity_type=$entityType"""
+        typeUpdate.as[Int]
+      }
+
       def activeActionForRefs(workspaceId: UUID, entities: Set[AttributeEntityReference]): ReadAction[Seq[AttributeEntityReference]] = {
         if( entities.isEmpty ) {
           DBIO.successful(Seq.empty[AttributeEntityReference])
@@ -449,6 +457,19 @@ trait EntityComponent {
                 where e.workspace_id=${workspaceContext.workspaceIdAsUUID} and ea.deleted=0 and (e.entity_type, e.name) in ("""
         val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { ref => sql"(${ref.entityType}, ${ref.entityName})" })
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
+      }
+
+      def batchHideForType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Seq[Int]] = {
+        val shardId = determineShard(workspaceContext.workspaceIdAsUUID)
+        // get unique suffix for renaming
+        val renameSuffix = "_" + getSufficientlyRandomSuffix(1000000000) // 1 billion
+        val deletedDate = new Timestamp(new Date().getTime)
+        // issue bulk rename/hide for all entity attributes, given an entity type
+        val baseUpdate =
+          sql"""update ENTITY_ATTRIBUTE_#$shardId ea join ENTITY e on ea.owner_id = e.id
+                set ea.deleted=1, ea.deleted_date=$deletedDate, ea.name=CONCAT(ea.name, $renameSuffix)
+                where e.workspace_id=${workspaceContext.workspaceIdAsUUID} and ea.deleted=0 and e.entity_type=$entityType"""
+        baseUpdate.as[Int]
       }
     }
 
@@ -803,6 +824,13 @@ trait EntityComponent {
       workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID) andThen
         EntityAndAttributesRawSqlQuery.batchHide(workspaceContext, entRefs) andThen
         EntityRecordRawSqlQuery.batchHide(workspaceContext.workspaceIdAsUUID, entRefs).map(res => res.sum)
+    }
+
+    // "deletes" entities of a certain type by hiding and renaming them
+    def hideType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Int] = {
+      workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID) andThen
+        EntityAndAttributesRawSqlQuery.batchHideForType(workspaceContext, entityType) andThen
+        EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum)
     }
 
     // perform actual deletion (not hiding) of all entities in a workspace
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -459,7 +459,7 @@ trait EntityComponent {
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
       }
 
-      def batchHideForType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Seq[Int]] = {
+      def batchHideAttributesOfType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Seq[Int]] = {
         val shardId = determineShard(workspaceContext.workspaceIdAsUUID)
         // get unique suffix for renaming
         val renameSuffix = "_" + getSufficientlyRandomSuffix(1000000000) // 1 billion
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -938,6 +944,19 @@ trait EntityComponent {
         entityAction map { _.toSet map { e: Entity => e.toReference } }
       }
     }
+
+    def countReferringEntities(context: Workspace, entityIds: Set[Long]): ReadAction[Long] = {
```
MatthewBemis: My original implementation for this endpoint piggy-backed off of what the existing deleteEntities endpoint was doing. However, I realized quickly that it didn't perform well. When I looked into why- it was doing recursive queries to get the entire entity tree for each of the entities being deleted. I don't think that this is necessary, so I short-circuited it and am only checking to see if there are any direct references to the entities being deleted

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -741,6 +741,79 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+    val (entityCountAfter, attributeCountAfter) = countEntitiesAttrs(testData.workspace)
+    val (activeEntityCountAfter, activeAttributeCountAfter) = countActiveEntitiesAttrs(testData.workspace)
+
+    //Ensure that the entity and attribute count (including "deleted" entities is the same before and after
+    assertResult(entityCountBefore)(entityCountAfter)
+    assertResult(attributeCountBefore)(attributeCountAfter)
+
+    //Ensure that we have marked the appropriate number of entities and attributes as "deleted"
+    assertResult(activeEntityCountBefore - entities.size)(activeEntityCountAfter)
+    assertResult(activeAttributeCountBefore - entities.flatMap(_.attributes).size)(activeAttributeCountAfter)
```
nawatts, line 780: I could be misunderstanding, but it looks like this only works because each entity has one attribute. Should this sum up the number of attributes for each entity?

nawatts (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/EntityProvider.scala
```diff
@@ -21,6 +21,8 @@ trait EntityProvider {
 
   def deleteEntities(entityRefs: Seq[AttributeEntityReference]): Future[Int]
 
+  def deleteEntitiesOfType(entityType: String): Future[Int]
+
```
davidangb, line 25: This looks like it returns 0 if there are no conflicts, or the numConflicts if conflicts exist.

I suggest instead it should return the number of rows deleted if delete succeeds, or throw DeleteEntitiesOfTypeConflictException(numConflicts) if conflicts exist.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -938,6 +944,19 @@ trait EntityComponent {
...
+        val query = entityAttributeShardQuery(context) filter (_.valueEntityRef inSetBind idBatch) join
+          this on { (attr, ent) => attr.ownerId === ent.id && ! ent.deleted } map { case (attr, entity) => (attr.valueEntityRef.get, entity)}
+        query.result.map(_.toSet)
+      }
+
+      val batchedEntityIds: Iterable[Set[Long]] = entityIds.grouped(1000).toList
+      val batchActions: Iterable[ReadAction[Set[(Long, EntityRecord)]]] = batchedEntityIds map findReferringEntities
+
+      DBIO.sequence(batchActions).map(_.flatten.size)
+    }
```
davidangb: 1) Does this include any self-references? Will the user's request fail if sample1 references sample2, even though they're trying to delete all samples? I think we should ignore all references within the type being deleted.

2) I think you could optimize the SQL in these two methods a bunch more. Currently it looks like two queries, where the second query has a potentially large `IN` clause. You should be able to do this in one query, something like this to find all references to `sample` entities:
```sql
select count(ea.id)
from ENTITY doing_reference, ENTITY being_referenced, ENTITY_ATTRIBUTE_00_03 ea
where ea.value_entity_ref = being_referenced.id
and ea.owner_id = doing_reference.id
and being_referenced.entity_type = 'sample'
and doing_reference.entity_type != 'sample'
and ea.deleted = 0
and doing_reference.deleted = 0;
```
(note this ignores references within the same type)

You'll have to verify this implementation but hopefully it illustrates what I'm suggesting! Obviously this would change the implementation in `LocalEntityProvider` too.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
```diff
@@ -110,6 +110,25 @@ class LocalEntityProvider(workspace: Workspace, implicit protected val dataSourc
...
+      traceDBIO("LocalEntityProvider.deleteEntitiesOfType") { rootSpan =>
+        rootSpan.putAttribute("workspaceId", OpenCensusAttributeValue.stringAttributeValue(workspaceContext.workspaceId))
+        rootSpan.putAttribute("entityType", OpenCensusAttributeValue.stringAttributeValue(entityType))
+
+        dataAccess.entityQuery.getActiveIdsForType(workspaceContext.workspaceIdAsUUID, entityType) flatMap { idsForType =>
+          dataAccess.entityQuery.countReferringEntities(workspace, idsForType.keys.toSet) flatMap { referringEntities =>
+            if (referringEntities != 0)
+              throw new DeleteEntitiesOfTypeConflictException(referringEntities)
+            else {
+              dataAccess.entityQuery.hide(workspaceContext, idsForType.values.toSeq)
```
davidangb: to do this right - and be efficient about user requests for deleting large tables - we need a new `hide()` method. The current `hide()` method accepts a potentially quite long list of ids and uses `WHERE … IN (many ids)`, which is super inefficient for this use case. Instead it should use a `WHERE entity_type='whatever' and deleted = 0`, which would be much cleaner.

davidangb (commented):
sorry I'm making this bigger on you! Comments inline.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/EntityProvider.scala
MatthewBemis: True, that is more in line with current behavior of similar code. I'll change that!

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
MatthewBemis: Another way to write this would be `entities.map(_.attributes).flatten.size`

`entities.map(_.attributes)` would return a `Seq[Map[AttributeName, Attribute]]`, which would be a list of all of the attribute lists for all of the `entities`. Calling `.flatten` on that then makes it a `Seq[(AttributeName, Attribute)]`, which is the list of all attributes across all of the `entities`, which we can then take the size of to get the total count of attributes. 

I used `.flatMap` which combines the `.map` and subsequent `.flatten` into one step

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityProvider.scala
MatthewBemis: Addressed in f1011c1650d87c6502526a89de774cfef3df3544

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/EntityProvider.scala
MatthewBemis: Handled in d013e26ffccfae0c550950e2dd94dcc1e183429a

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: This implementation seems to work when scoped to a particular workspace_id!

MatthewBemis:
@davidangb this should be all set for re-review now

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -799,6 +834,13 @@ trait EntityComponent {
         EntityRecordRawSqlQuery.batchHide(workspaceContext.workspaceIdAsUUID, entRefs).map(res => res.sum)
     }
 
+    // "deletes" entities of a certain type by hiding and renaming them
+    def hideType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Int] = {
+      workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID) andThen
+        EntityAndAttributesRawSqlQuery.batchHideForType(workspaceContext, entityType) andThen
+        EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum)
+    }
```
davidangb, line 846: I recognize you are following the pattern here - there are lots of other places where we update the workspace's lastModified and then perform other writes. Thinking about this pattern - I suspect it would be best to update the workspace's lastModified as the final step, after all other processing has completed.

1. In general, the other sql actions are more likely to have errors and so we should try them first; if they fail we wouldn't even attempt the lastModified update and there's one less thing to roll back
2. In general, the other sql actions will be bigger, so we shouldn't hold the lock on WORKSPACE while the other actions run; we should grab the WORKSPACE lock as the last thing

What's your take on this?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -450,6 +458,33 @@ trait EntityComponent {
         val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { ref => sql"(${ref.entityType}, ${ref.entityName})" })
         concatSqlActions(baseUpdate, entityTypeNameTuples, sql")").as[Int]
       }
+
+      def batchHideForType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Seq[Int]] = {
```
davidangb: nit: `batchHideAttributesOfType`? `batchHideAttributes`? As-is, `batchHideForType` and `batchHideType` feel too confusingly similar

davidangb (commented):
one nit, one brainstorm …

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: Renamed in 043714445336abf20e2b4c591ae74182077e2a2f

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: I think that argument makes sense- this is a bit of a heavier operation than most others so it's important to be more mindful about how the locking is handled. I changed the order of the sql actions in 311e53f700255c92bb69d160e8585e37151b346f

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -799,6 +834,13 @@ trait EntityComponent {
         EntityRecordRawSqlQuery.batchHide(workspaceContext.workspaceIdAsUUID, entRefs).map(res => res.sum)
     }
 
+    // "deletes" entities of a certain type by hiding and renaming them
+    def hideType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Int] = {
+      EntityAndAttributesRawSqlQuery.batchHideAttributesOfType(workspaceContext, entityType) andThen
+        EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum) andThen
+        workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
+    }
```
davidangb:
```suggestion
    def hideType(workspaceContext: Workspace, entityType: String): ReadWriteAction[Int] = {
      for {
        _ <- EntityAndAttributesRawSqlQuery.batchHideAttributesOfType(workspaceContext, entityType)
        numEntitiesHidden <- EntityRecordRawSqlQuery.batchHideType(workspaceContext.workspaceIdAsUUID, entityType).map(res => res.sum)
        - <- workspaceQuery.updateLastModified(workspaceContext.workspaceIdAsUUID)
      } yield numEntitiesHidden
    }
```
as written, this method will now always return "1" on success - the number of rows changed by `workspaceQuery.updateLastModified` … I think you need to switch to the for/yield and control which value is returned

davidangb (approved):
thank you, one more change, giving a proactive :+1: if that's the only thing you change!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
MatthewBemis: gahhh nice catch! i'll handle this momentarily

MatthewBemis: Fixed in 81f402ae4ad34408cf08a18cb5448f8a6fbf0cec . I also commit the unit test that I used to verify that it returned what I expected

davidangb (approved):

PR #1805, opened 2022-05-16
AJ-356 case sensitive entity types [AJ-356]

davidangb:
In this PR:
* Change the collation of the `ENTITY.entity_type` db column to `utf8_bin`, to support case-sensitivity
* Modify the custom user-facing error message about case-sensitive entity types and names to only be about names
* Write all the unit tests that I could think of
* Remove the two obsolete unit tests in `LocalEntityProviderSpec`
* Update one unit test in `ExpressionEvaluatorTest` that broke now that types are case-sensitive, and added new unit tests in `CaseSensitivitySpec` to ensure good coverage of that feature

DO NOT MERGE this PR until UI features are solidly in place to support renaming of entity types. Once this merges, if a workspace contains case-divergent entity types they will be visible as dirty data, and we'll need rename features to support users in their self-service cleanup.

TODOs:
- [x] investigate the TODO in `CaseSensitivitySpec` around entity metadata cache staleness
- [x] manual testing through the UI against a db clone


core/src/test/scala/org/broadinstitute/dsde/rawls/expressions/ExpressionEvaluatorTest.scala
```diff
@@ -418,7 +418,7 @@ class ExpressionEvaluatorTest extends AnyFunSuite with TestDriverComponent {
       }
 
       assertResult(Map("sample1" -> TrySuccess(Seq(AttributeString("sample1"))))) {
-        runAndWait(evalFinalAttribute(workspaceContext, "Sample", "sample1", "this.sample_id"))
+        runAndWait(evalFinalAttribute(workspaceContext, "Sample", "sample1", "this.Sample_id"))
```
davidangb, line 421: this pre-existing test started to fail with the new db collation. It used an entity type of "Sample" (sentence case) but an expression of "this.sample_id" (lower case). With case-insensitive matching, this passed, but now fails. I've updated the test to be accurate, but this behavior change will need to be publicized.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -0,0 +1,503 @@
...
+
+          runAndWait(entityCacheQuery.entityCacheStaleness(testWorkspace.workspace.workspaceIdAsUUID)) should contain (0)
+        }
+
+        exemplarTypes foreach { typeUnderTest =>
+          s"should only rename target type [$typeUnderTest]" in withTestDataServices { services =>
+            // save exemplar data
+            runAndWait(entityQuery.save(testWorkspace.workspace, exemplarData))
+            // rename type
+            services.entityService.renameEntityType(testWorkspace.workspace.toWorkspaceName, typeUnderTest, EntityTypeRename("my-new-name")).futureValue
```
jordansafer: One test I'm wondering about: If we rename some entity type to "cAT", will this succeed? (It should, since we only have "cat", "Cat", and "CAT". Or will the rename API reject the argument as a conflict?

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -0,0 +1,503 @@
...
+
+            // get actual entity types from the db
+            val actualEntityTypes = getAllEntityTypes(testWorkspace.workspace)
+
+            actualEntityTypes shouldBe exemplarTypes - typeUnderTest
+          }
+        }
+
+        exemplarTypes foreach { typeUnderTest =>
+          s"should only delete an entire column from target type [$typeUnderTest]" in withTestDataServices { services =>
```
jordansafer: I think the goal of this test is to make sure than all cases of an attributed get deleted. nit: could we make the documentation for this test more detailed?

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -0,0 +1,503 @@
...
+
+            actualEntities.count(_.name == "my-new-name") shouldBe 1
+            actualEntities.count(_.name == "001") shouldBe exemplarTypes.size
+            actualEntities.count(_.name == "002") shouldBe exemplarTypes.size
+            actualEntities.count(_.name == "003") shouldBe exemplarTypes.size - 1
+          }
+        }
+
+        exemplarTypes foreach { typeUnderTest =>
+          s"should respect case when creating an entity reference [$typeUnderTest]" in withTestDataServices { _ =>
```
jordansafer, line 363: What should we expect for an existing~, case-insensitive entity reference? Wondering if there's a good way we could test this. Ie entity reference value is "Cat", which is the same as "cat" the old way. Then we enable case sensitivity, and the entityId for the attribute is unchanged. So we now expect that "Cat" will be pointing to "cat" I think-

jordansafer (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
davidangb: this should pass. I've updated the test to rename to a case-divergent type to prove this in practice!

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
davidangb: updated!

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
davidangb: you're correct. Case-sensitivity comes into play at the time of creation of the reference.

Before this PR:
* if an entity exists with type "cat" and name "001", and a user creates an entity reference to "CAT/001", creating the reference will succeed.

After this PR:
* any references that have already been created will remain in place, even if they relied on case-insensitivity to succeed
* any new attempts to create a case-insensitive reference will fail; "CAT/001" -> "cat/001" will throw an error at creation time.

davidangb:
`gcloud crashed` during rawls-build in swatomation. Jenkins retest.

jordansafer (approved):

PR #1814, opened 2022-05-17
AJ-418 (Make copy entities more efficient)

sagehen03:
This change set makes use of the work in #1761 to copy entities to new workspace.  It also corrects an existing flaw in copy and delete entities that resulted in `where field in (<single elemement>)` SQL when we actually intended `where field in (<batch size elements>).` The latter fix applies to delete entities too.


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DriverComponent.scala
```diff
@@ -38,10 +38,6 @@ trait DriverComponent extends StringValidationUtils {
     }
   }
 
-  def createBatches[T](items: Set[T], batchSize: Int = 1000): Iterable[Set[T]] = {
-    items.zipWithIndex.groupBy(_._2 % batchSize).values.map(_.map(_._1))
-  }
-
```
sagehen03, line 44: This function should use int division rather than modulo to achieve its intent, but `grouped` does the intended thing without writing any code so I'm replaced invocations of this function with `grouped`

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -187,7 +187,7 @@ trait EntityApiService extends UserInfoDirectives {
                 entity(as[EntityCopyDefinition]) { copyDefinition =>
                   traceRequest { span =>
                     complete {
-                      entityServiceConstructor(userInfo).copyEntities(copyDefinition, request.uri, linkExistingEntitiesBool).map { response =>
+                      entityServiceConstructor(userInfo).copyEntities(copyDefinition, request.uri, linkExistingEntitiesBool, span).map { response =>
```
sagehen03, line 190: Semi-tangential to the ticket, but this change should allow devs to find tracing for copy entities starting from the API url.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,7 +553,16 @@ trait EntityComponent {
                     on ea.value_entity_ref = referenced_entity.old_id
                 join ENTITY e on e.id = ea.owner_id where ea.deleted = false and e.deleted = false and e.workspace_id = $clonedWorkspaceId
           """
+        addEntitiesToWhereIfNeeded(entityRefs, baseInsert).head
+      }
 
+      private def addEntitiesToWhereIfNeeded(entityRefs: Set[AttributeEntityReference], baseInsert: SQLActionBuilder) = {
+        if (entityRefs.nonEmpty) {
+          val entityTypeNameTuples = reduceSqlActionsWithDelim(entityRefs.map { ref => sql"(${ref.entityType}, ${ref.entityName})" }.toSeq)
+          concatSqlActions(baseInsert, sql" and (e.entity_type, e.name) in (", entityTypeNameTuples, sql")").as[Int]
```
davidangb: could you check the explain-plan on this and see if is properly using indexes? Remember #1578 - if a similar change makes the query index-happy, it should be a win!

sagehen03: Surprisingly, it looks like the queries are using proper indexes.  For reference.. 
```
CREATE TABLE `ENTITY` (
  `name` varchar(254) NOT NULL,
  `entity_type` varchar(254) NOT NULL,
  `workspace_id` binary(16) NOT NULL,
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `record_version` bigint(20) NOT NULL,
  `all_attribute_values` text,
  `deleted` bit(1) DEFAULT b'0',
  `deleted_date` timestamp(6) NULL DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `idx_entity_type_name` (`workspace_id`,`entity_type`,`name`),
  CONSTRAINT `FK_ENTITY_WORKSPACE` FOREIGN KEY (`workspace_id`) REFERENCES `WORKSPACE` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION
) ENGINE=InnoDB AUTO_INCREMENT=28400068 DEFAULT CHARSET=utf8
```

`select id, name, entity_type, workspace_id, record_version, deleted, deleted_date from ENTITY where workspace_id = ? and (entity_type, name) in ((?, ?),(?, ?)))`

```
explain select id, name, entity_type, workspace_id, record_version, deleted, deleted_date from ENTITY where workspace_id = UNHEX('1FFB8836A7C341B99177B7DFC57A4CAC') and (entity_type, name) in (('analysis_file', '07d32415-2d4a-4759-bad7-400e3001c667'), ('analysis_file', '0f5e26a7-a7a8-4945-be1a-84aed70c809b'));
+----+-------------+--------+------+----------------------+----------------------+---------+-------+--------+------------------------------------+
| id | select_type | table  | type | possible_keys        | key                  | key_len | ref   | rows   | Extra                              |
+----+-------------+--------+------+----------------------+----------------------+---------+-------+--------+------------------------------------+
|  1 | SIMPLE      | ENTITY | ref  | idx_entity_type_name | idx_entity_type_name | 16      | const | 102948 | Using index condition; Using where |
+----+-------------+--------+------+----------------------+----------------------+---------+-------+--------+------------------------------------+

```
The query is using the idx_entity_type_name index, but the number of rows searched is high which seems to indicate it's not using the entity_type or name pieces of the index.  

```
explain select id, name, entity_type, workspace_id, record_version, deleted, deleted_date from ENTITY where workspace_id = UNHEX('1FFB8836A7C341B99177B7DFC57A4CAC') and ((entity_type = 'analysis_file' and name = '0f5e26a7-a7a8-4945-be1a-84aed70c809b') or (entity_type = 'analysis_file' and name = '51aab564-90d2-4fdc-ac2d-2808c593700d'));
+----+-------------+--------+-------+----------------------+----------------------+---------+------+------+-----------------------+
| id | select_type | table  | type  | possible_keys        | key                  | key_len | ref  | rows | Extra                 |
+----+-------------+--------+-------+----------------------+----------------------+---------+------+------+-----------------------+
|  1 | SIMPLE      | ENTITY | range | idx_entity_type_name | idx_entity_type_name | 1544    | NULL |    2 | Using index condition |
+----+-------------+--------+-------+----------------------+----------------------+---------+------+------+-----------------------+

```

Converting to the above query structure results in a better explain.  ~~I will switch the code to generate something like this but I think it will mean ditching slick for these queries.  Let me know if you have opinions on this tradeoff @davidangb or others?~~  Looks like we're already using a raw query, nvm.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -516,27 +516,23 @@ trait EntityComponent {
...
-                       0, null from ENTITY where workspace_id = $clonedWorkspaceId and deleted = 0
+        val baseInsert = sql"""insert into ENTITY (name, entity_type, workspace_id, record_version, all_attribute_values, deleted, deleted_date)
+                select name, entity_type, $newWorkspaceId, record_version, all_attribute_values, 0, null from ENTITY e where workspace_id = $clonedWorkspaceId and deleted = 0
           """
+        addEntitiesToWhereIfNeeded(entityRefs, baseInsert).head
       }
-    }
-
-    private object CopyEntityAttributesQuery extends RawSqlQuery {
-      val driver: JdbcProfile = EntityComponent.this.driver
```
davidangb, line 532: nice - I like removing the extra `CopyEntityAttributesQuery` object and consolidating inside `CopyEntitiesQuery`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -557,7 +553,16 @@ trait EntityComponent {
...
           """
+        addEntitiesToWhereIfNeeded(entityRefs, baseInsert).head
+      }
 
+      private def addEntitiesToWhereIfNeeded(entityRefs: Set[AttributeEntityReference], baseInsert: SQLActionBuilder) = {
+        if (entityRefs.nonEmpty) {
+          val entityTypeNameTuples = reduceSqlActionsWithDelim(entityRefs.map { ref => sql"(${ref.entityType}, ${ref.entityName})" }.toSeq)
+          concatSqlActions(baseInsert, sql" and (e.entity_type, e.name) in (", entityTypeNameTuples, sql")").as[Int]
+        } else {
+          concatSqlActions(baseInsert).as[Int]
```
davidangb: concatSqlActions is extraneous here, could be just:
```suggestion
          baseInsert.as[Int]
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -901,10 +906,14 @@ trait EntityComponent {
       EntityAndAttributesRawSqlQuery.countReferencesToType(workspaceContext, entityType).map(_.sum)
     }
 
-    def cloneEntitiesToNewWorkspace(sourceWs: UUID, destWs: UUID): WriteAction[(Int, Int)] = {
+    def cloneWorkspace(sourceWs: UUID, destWs: UUID): WriteAction[(Int, Int)] = {
+      copyEntitiesToNewWorkspace(sourceWs, destWs)
+    }
```
davidangb: 1) I think this is only called from one place, could you delete this wrapper method and have the calling code call `copyEntitiesToNewWorkspace` directly?

2) I believe the `private def copyEntities` at line 1003 is now unused - delete that?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -932,8 +941,10 @@ trait EntityComponent {
     // copy entities from one workspace to another, checking for conflicts first
 
     def checkAndCopyEntities(sourceWorkspaceContext: Workspace, destWorkspaceContext: Workspace, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean, parentSpan: Span = null): ReadWriteAction[EntityCopyResponse] = {
+      val batchSize = 500
```
davidangb: 1) do you have any sense of the impact of raising or lowering this number on performance?
2) Note that here in the code you already have access to a `batchSize` variable from the `DriverComponent` trait, which is populated from the `slick.batchSize` number in rawls.conf - currently set to 2000 [here](https://github.com/broadinstitute/firecloud-develop/blob/7757fd0e50ee26b8b9b3b70cce278eba2440079c/base-configs/rawls/rawls.conf.ctmpl#L497). Does it make any sense to use that instead?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -970,10 +981,13 @@ trait EntityComponent {
                 val allConflictRefs = softConflicts.flatMap(_.path)
                 val entitiesToCopy = allEntityRefs diff allConflictRefs
 
+                val entityRefsToCopy = entitiesToCopy.map(e => AttributeEntityReference(e.entityType, e.entityName)).toSet
+                val entitiesToCopyChunks = entityRefsToCopy.grouped(batchSize)
```
davidangb: `entitiesToCopy` already contains `AttributeEntityReference`s, no need for that first map:
```suggestion
                val entitiesToCopyChunks = entitiesToCopy.toSet.grouped(batchSize)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -901,10 +906,14 @@ trait EntityComponent {
...
+
+    def copyEntitiesToNewWorkspace(sourceWs: UUID, destWs: UUID, entityRefs: Set[AttributeEntityReference] = Set()): WriteAction[(Int, Int)] = {
       for {
-        entitiesCopiedCount <- CopyEntitiesQuery.copyEntities(sourceWs, destWs)
-        attributesCopiedCount <- CopyEntityAttributesQuery.copyAllAttributes(sourceWs, destWs)
+        entitiesCopiedCount <- CopyEntitiesQuery.copyEntities(sourceWs, destWs, entityRefs)
+        attributesCopiedCount <- CopyEntitiesQuery.copyAttributes(sourceWs, destWs, entityRefs)
       } yield {
         (entitiesCopiedCount, attributesCopiedCount)
       }
```
davidangb, line 930: I'm wondering if it makes sense to move the batching of `entityRefs` into here, instead of doing it inside `checkAndCopyEntities`. As-is, this PR is fine (and I wouldn't block its merging based on this comment), but in the future if other code needs to call `copyEntitiesToNewWorkspace`, those code locations would also have to implement their own batching before calling this. Additionally, if you moved the batching here, you have the opportunity to execute all the batches and then sum their results to return an accurate `(Int, Int)` return count. What do you think?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -970,10 +981,13 @@ trait EntityComponent {
...
 
+                val entityRefsToCopy = entitiesToCopy.map(e => AttributeEntityReference(e.entityType, e.entityName)).toSet
+                val entitiesToCopyChunks = entityRefsToCopy.grouped(batchSize)
                 for {
-                  entities <- traceDBIOWithParent("getActiveEntities", s2)(_ => getActiveEntities(sourceWorkspaceContext, entitiesToCopy))
-                  _ <- traceDBIOWithParent("copyEntities", s2)(_ => copyEntities(destWorkspaceContext, entities.toSet, allConflictRefs))
-                } yield EntityCopyResponse(entities.map(_.toReference).toSeq, Seq.empty, Seq.empty)
+                  _ <- traceDBIOWithParent("copyEntities", s2)(_ => DBIO.sequence(entitiesToCopyChunks map {chunk => copyEntitiesToNewWorkspace(sourceWorkspaceContext.workspaceIdAsUUID,
+                    destWorkspaceContext.workspaceIdAsUUID, chunk)}))
+                  _ <- workspaceQuery.updateLastModified(destWorkspaceContext.workspaceIdAsUUID)
```
davidangb, line 990: huh, the workspace 's lastModified was never updated when copying entities? Ouch - good catch!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: Initially I thought higher than 500 might run the risk of hitting SQL length limits, but that was just how I've usually operated with batching for `where.. in ()` if 2000 is working for Rawls everywhere else it makes sense to switch.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: Good catch.  I know it's our Scala idiom but I miss seeing the type info.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: Sorry for missing the deletion of `copyEntities` that's a good one!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: I hear you on moving the batching into `copyEntitiesToNewWorkspace.`  It would help callers avoid bugs, but it complicates sharing that code between cloning workspaces and copying entities--at least it does for my rudimentary scala brain.  Here's a first stab at what that code would look like and I'm genuinely unsure about the tradeoffs.  
```
def copyEntitiesToNewWorkspace(sourceWs: UUID, destWs: UUID, entityRefs: Set[AttributeEntityReference] = Set()): WriteAction[(Int, Int)] = {

def copyChunkOfEntitiesOrAllEntities(chunk: Set[AttributeEntityReference] = Set()) = {
        for {
          entitiesCopiedCount <- CopyEntitiesQuery.copyEntities(sourceWs, destWs, chunk)
          attributesCopiedCount <- CopyEntitiesQuery.copyAttributes(sourceWs, destWs, chunk)
        } yield {
          (entitiesCopiedCount, attributesCopiedCount)
        }
      }
      val copyActions = if(entityRefs.isEmpty) Iterator(copyChunkOfEntitiesOrAllEntities()) else entityRefs.grouped(batchSize) map { chunk =>
        copyChunkOfEntitiesOrAllEntities(chunk)
      }
      val allCopies = DBIO.sequence(copyActions)

      allCopies.map { copyActionResults: Iterator[(Int, Int)] =>
        (copyActionResults.map(_._1).sum, copyActionResults.map(_._2).sum)
      }
}
```
I'm wondering if there's a better way to get code reuse than using an empty set as a default parameter.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I think the `DriverComponent.batchSize` is only used in a few places, but it does seem stable so far. `EntityRecordRawSqlQuery.action` is probably the most similar usage.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I prefer seeing type info too, I am not attached at all to omitting it all the time.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I think the problem you're running into here is that `Set.grouped` will return an empty list if the set is empty, correct? And you always need at least one chunk on which to operate, if if that single chunk is the empty set?

So how about ensuring `chunks` has the shape you want, something like:
```scala
val chunks: Iterator[Set[AttributeEntityReference]] = if (entityRefs.nonEmpty) {
  entityRefs.grouped(batchSize)
} else {
  Iterator(Set.empty[AttributeEntityReference])
}
```
or perhaps even better:
```scala
val chunks: Iterator[Set[AttributeEntityReference]] = if (entityRefs.size > batchSize) {
  entityRefs.grouped(batchSize)
} else {
  Iterator(entityRefs)
}
```
and then you can proceed to map over `chunks`, knowing it's correct

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -168,8 +168,8 @@ trait EntityComponent {
         if( entities.isEmpty ) {
           DBIO.successful(Seq.empty[EntityRecord])
         } else {
-          val baseSelect = sql"select id, name, entity_type, workspace_id, record_version, deleted, deleted_date from ENTITY where workspace_id = $workspaceId and (entity_type, name) in ("
-          val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { entity => sql"(${entity.entityType}, ${entity.entityName})" }.toSeq)
+          val baseSelect = sql"select id, name, entity_type, workspace_id, record_version, deleted, deleted_date from ENTITY where workspace_id = $workspaceId and ("
+          val entityTypeNameTuples = reduceSqlActionsWithDelim(entities.map { entity => sql"(entity_type = ${entity.entityType} and name = ${entity.entityName})" }.toSeq, sql" OR ")
```
davidangb, line 172: this could use `addEntitiesToWhereIfNeeded` too - not important enough to hold this PR, your call

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I still think you could do away with `cloneWorkspace` entirely, and have `WorkspaceService` call `copyEntitiesToNewWorkspace` directly.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -901,12 +906,31 @@ trait EntityComponent {
...
+        }
+      }
+
+      val chunks: Iterator[Set[AttributeEntityReference]] = if (entityRefs.size > batchSize) {
+        entityRefs.grouped(batchSize)
+      } else {
+        Iterator(entityRefs)
+      }
+
+      val allCopies = DBIO.sequence(chunks.map(chunk => copyChunkOfEntitiesOrAllEntities(chunk)))
```
davidangb: Scala idiomaticness: when the item inside the map (in this case, `chunk`) is the only argument to the right hand side of the map, you can omit it entirely and have shorthand:
```scala
val allCopies = DBIO.sequence(chunks.map(copyChunkOfEntitiesOrAllEntities))
```
or 
```scala
val allCopies = DBIO.sequence(chunks map copyChunkOfEntitiesOrAllEntities)
```
I mention this just to highlight Scala syntaxes - if you prefer the existing code, leave it! The existing code is more explicit and can be easier to understand.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -932,11 +956,9 @@ trait EntityComponent {
...
 
     def checkAndCopyEntities(sourceWorkspaceContext: Workspace, destWorkspaceContext: Workspace, entityType: String, entityNames: Seq[String], linkExistingEntities: Boolean, parentSpan: Span = null): ReadWriteAction[EntityCopyResponse] = {
+
       def getHardConflicts(workspaceId: UUID, entityRefs: Seq[AttributeEntityReference]) = {
-        val batchActions = createBatches(entityRefs.toSet).map(batch => getEntityRecords(workspaceId, batch))
-        DBIO.sequence(batchActions).map(_.flatten.toSeq).map { recs =>
-          recs.map(_.toReference)
-        }
+       getEntityRecords(workspaceId, entityRefs.toSet)
       }
```
davidangb: looking at this even closer, the only place that `getHardConflicts` is called only uses the entity type and entity name from the retrieved records. So, `getEntityRecords` is overkill; it returns columns we don't need. `getHardConflicts` could call `activeActionForRefs` instead, which only returns the type&name columns. What's your take on this? Is it worth changing?

davidangb (approved):
some comments inline - not enough to block the PR, but I'd love to hear your thoughts on them. Giving this a :+1:

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
sagehen03: I looked into this a bit and it seems like I would need relocate `addEntitiesToWhereIfNeeded` somewhere else so it could be called from the `EntityRecordRawSqlQuery` object and then I would also need generify the function to return Int in the existing cases and EntityRecord in this new case?  Probably a good exercise for me.. I'll look into it :)

sagehen03: I took a stab at it and I got a compilation error about missing an implicit.  I think this might be best saved for a 1:1 so I'm going to punt on this change for now.

sagehen03:
jenkins retest

PR #1827, opened 2022-05-18
AJ 394

calypsomatic:
See AJ-394 for description of how to reproduce issue.  Unacked messages can be seen on google cloud console by going to Pub/Sub -> Subscriptions -> rawls-async-import-topic-sub-[ENV] (or rather, hopefully won't be seen).  

This PR adds a recover after creating `importFuture` to acknowledge the pubsub message in case anything fails.  Note that this does not cover the situation where `parseMessage` throws an error.  

Tests are added to check failure when finding the workspace, getting the pet user info, and fetching the import status.   To do this I added some mocked objects, which resulted in somewhat duplicated code (`setUpMockImportService/setUpMockSamDAO/setUpMockEntityManager` (pre-existing method, used as a model).  Would love input on whether refactoring this is necessary or desirable, or really anything else.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -276,7 +276,8 @@ class AvroUpsertMonitorActor(
 
     //Make sure message is acknowledged in the case of any failure while trying to construct importFuture
     importFuture.map(_ => ImportComplete)  recover {
-      case _ =>
+      case t =>
+        logger.error(s"unexpected error in importFuture for ${attributes.importId}: ${t.getMessage}", t)
         acknowledgeMessage(message.ackId)
     } pipeTo self
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -274,7 +274,12 @@ class AvroUpsertMonitorActor(
       }
     } yield ()
 
-    importFuture.map(_ => ImportComplete) pipeTo self
+    //Make sure message is acknowledged in the case of any failure while trying to construct importFuture
+    importFuture.map(_ => ImportComplete)  recover {
+      case Throwable =>
+        acknowledgeMessage(message.ackId)
+    } pipeTo self
+
   }
 
   private def publishMessageToUpdateImportStatus(importId: UUID, currentImportStatus: Option[ImportStatus], newImportStatus: ImportStatus, errorMessage: Option[String]) = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -276,7 +276,7 @@ class AvroUpsertMonitorActor(
 
     //Make sure message is acknowledged in the case of any failure while trying to construct importFuture
     importFuture.map(_ => ImportComplete)  recover {
-      case Throwable =>
+      case _ =>
         acknowledgeMessage(message.ackId)
     } pipeTo self
 
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
@@ -143,7 +143,7 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
     setUpPubSub(services)
 
     val mockImportServiceDAO =  mock[ImportServiceDAO]
-    when(mockImportServiceDAO.getImportStatus(failImportStatusUUID, importStatusFailingWorkspace, userInfo)).thenReturn(Future.failed(new Exception("USer not found")))
+    when(mockImportServiceDAO.getImportStatus(failImportStatusUUID, workspaceName, userInfo)).thenReturn(Future.failed(new Exception("User not found")))
 
     // Start the monitor
     system.actorOf(AvroUpsertMonitorSupervisor.props(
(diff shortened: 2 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -274,7 +274,12 @@ class AvroUpsertMonitorActor(
       }
     } yield ()
 
-    importFuture.map(_ => ImportComplete) pipeTo self
+    //Make sure message is acknowledged in the case of any failure while trying to construct importFuture
+    importFuture.map(_ => ImportComplete)  recover {
+      case _ =>
+        acknowledgeMessage(message.ackId)
```
davidangb: 1) To assist with future debugging, how about logging the error here? Something like:
```suggestion
      case t =>
        logger.error(s"unexpected error in importFuture for ${attributes.importId}: ${t.getMessage}", t)
        acknowledgeMessage(message.ackId)
```

Unless you're aware of some other logging statement that would consistently log these errors.

2) I realize this wasn't covered in the original Jira ticket - do you know what state this leaves the import job in? Is it stuck in "Upserting" until we expire it after 36 hours? We might want to add a `publishMessageToUpdateImportStatus(attributes.importId, None, ImportStatuses.Error, Option(s"${t.message}"))` here - though that would need more unit tests of course. We could do that in a follow-on PR.

3) After digging deeper into the existing code, I realize that there are some use cases in which this will be a double-ack (or a double-log, if you add the logging). Since `initUpsert()` is contained within `importFuture`, and `initUpsert()` has its own try/catch, some exceptions within `initUpsert` would first trigger `initUpsert`'s catch clause, then trigger this new recover clause. We'd ack the message twice … which I also think is harmless, so should cause no problem. Just something to keep an eye on in case it has unpredicted behavior.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -105,6 +108,7 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
         readTopicCreate <- services.gpsDAO.createTopic(importReadPubSubTopic)
         writeTopicCreate <- services.gpsDAO.createTopic(importWritePubSubTopic)
         subscriptionCreate <- services.gpsDAO.createSubscription(importWritePubSubTopic, importWriteSubscriptionName)
+//        subscriptionCreate <- services.gpsDAO.createSubscription(importReadPubSubTopic, importReadSubscriptionName)
```
davidangb: nit: if this is really not necessary, delete instead of commenting out

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -443,6 +493,102 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+
+    // check if correct message was posted on request topic. This will start the upsert attempt.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      assert(services.gpsDAO.receivedMessage(importReadPubSubTopic, importId1.toString))
+    }
+
+    // upsert will fail; check that a pubsub message was acked.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      assert(!services.gpsDAO.acks.isEmpty)
+    }
```
davidangb, line 529: 1) Since there's no good way to verify that we've acked the correct message - we don't have access to the ackId - how about at least adding an assertion earlier in the test to verify that `acks` is empty? Something like:
```scala
    // acks should be empty at this point
    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
      services.gpsDAO.acks shouldBe empty
    }
```
before publishing the "start import" message to the request topic. (this applies to all the new tests in this PR)

2) To be Scala/Scalatest idiomatic, check out [matchers](https://www.scalatest.org/user_guide/using_matchers#checkingForEmptiness). You could change `assert(!services.gpsDAO.acks.isEmpty)` to `services.gpsDAO.acks should not be empty`. This is functionally equivalent and doesn't make a difference to test behavior itself, but might be easier to read and more consistent with other tests.

davidangb (commented):
see comments inline!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
calypsomatic: Whoops!  Missed that. Thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
calypsomatic: For the case where a workspace isn't found, it does update the import status, but it wouldn't for errors in `getPetServiceAccountUserInfo` or `getImportStatus`.  I could move the `publishMessageToUpdateImportStatus` for missing workspaces to the recover so that it covers those cases.  Same with logging.

calypsomatic: I've added the logging statement, but left the `publishMessageToUpdateImportStatus` as is so I can add it on a different commit, or we could make it a separate ticket.  Which seems better?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
davidangb: whoops, this second assertion, after the upsert job runs, needs to be:
```scala
services.gpsDAO.acks should not be empty
```

davidangb (changes requested):
one required change in the tests!

I say let's handle publishing the message to change the job's status in a separate PR/separate ticket, so as to not block this one any longer than necessary.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
calypsomatic: It should be empty before publishmessage and after receive message.  But are you saying I should check that it's *not* empty inbetween?

calypsomatic: Ok, did it and double-checked that it fails without the code change.

davidangb (approved):
:+1: once this passes tests!

PR #1830, opened 2022-05-20
AJ-443 failing async imports do not always update import status to error

calypsomatic:
AJ-443, follow up to AJ-394.  Simply moves call to `publishMessageToUpdateImportStatus` which happens only in the case of a non-existent workspace to the `recover` block so that import status is updated for any kind of failure.  Adds unit tests in the case of non-existent workspace, `getPetServiceAccountUserInfo` throwing an exception ,and `getImportStatus` throwing an exception.


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -278,6 +277,7 @@ class AvroUpsertMonitorActor(
     importFuture.map(_ => ImportComplete)  recover {
       case t =>
         logger.error(s"unexpected error in importFuture for ${attributes.importId}: ${t.getMessage}", t)
+        publishMessageToUpdateImportStatus(attributes.importId, None, ImportStatuses.Error, Option(t.getMessage))
```
davidangb: nit: since this error message is user-facing, we probably shouldn't rely just on the exception's message. How about:
```
        publishMessageToUpdateImportStatus(attributes.importId, None, ImportStatuses.Error,
          Option(s"Failed to import data. The underlying error message is: ${t.getMessage}"))
```

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -602,6 +602,127 @@ class AvroUpsertMonitorSpec(_system: ActorSystem) extends ApiServiceSpec with Mo
...
+    // Store upsert json file
+    Await.result(googleStorage.createBlob(bucketName, GcsBlobName(importId1.toString), contents.getBytes()).compile.drain.unsafeToFuture(), Duration.apply(10, TimeUnit.SECONDS))
+
+    // Make sure the file saved properly
+    Await.result(googleStorage.unsafeGetBlobBody(bucketName, GcsBlobName(importId1.toString)).unsafeToFuture(), Duration.apply(10, TimeUnit.SECONDS))
+
+    // acks should be empty at this point
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      services.gpsDAO.acks shouldBe empty
+    }
```
davidangb: here - and in the test just below this one - the assertion on empty acks is extraneous, you could pull it out just to clean up a little code

davidangb (approved):
very minor comments, this looks good!

I'm giving this a proactive :+1: assuming you'll address the feedback … if you end up changing something other than responding to my comments, request a new review please!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
calypsomatic: Yep, missed those.  Good catch!

PR #1831, opened 2022-05-20
[CA-1895] Fix BillingAccountChange sync failure for v2 billing projects

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1895
This change stops the `WorkspaceBillingAccountActor` attempting to sync a BillingAccountChange on the billing project when the billing project name is not a valid google project id. Note that v2 workspaces are unaffected.

I'm not sure how else to validate that a google project id is valid other than with a regular expression. Suggestions welcome.
A better change to make would be to change the `googleProjectId` method on `RawlsBillingProject` to return an `Option[GoogleProjectId]` as v2 billing projects do not have a google project associated with them. That's a far larger change, however, and I think this needs to be fixed sooner.

I had to change the billing project name to be a valid google project in the test data. It turns out that those test data objects are mixed and matched so I had to change all occurences of `myNamespace` to `my-namspace`.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
```diff
@@ -111,15 +111,19 @@ final case class WorkspaceBillingAccountActor(dataSource: SlickDataSource, gcsDA
   private def syncBillingProjectWithGoogle[F[_]](billingProject: RawlsBillingProject)
                                                 (implicit R: Ask[F, BillingAccountChange], M: Monad[F], L: LiftIO[F])
   : F[Unit] =
-    for {
-      // only v1 billing projects are backed by google projects
-      isV1BillingProject <- L.liftIO {
-        gcsDAO.rawlsCreatedGoogleProjectExists(billingProject.googleProjectId).io
```
andy7i, line 117: I'm trying to better understand the problem here. I think this is how this error could occur. Is it one of these scenarios?:
1. V2 billing projects should have an empty value in `billingProject.googleProjectId`, right?
1. So this error should only occur for V1 billing projects
2. On V1 billing projects, `billingProject.googleProjectId` should be the same as the billing project name. 
     1. Which means that some v1 billing projects have the wrong value filled in for the `billingProject.googleProjectId` field. (Not sure how this would happen)
     4. OR some projects were created without billing project name validation. But I think that would mean that the billing project creation would fail when it tried to create the google project with an illegal name.
5.  OR `billingProject.googleProjectId` is different from the billing project name. (Again, not sure how this would happen)

is this due to 3.ii. ? If so, shouldn't these be errored billing projects? / should we filter out errored billing projects (as well)?

andy7i (commented):
(not sure I understand the problem well enough to properly review the solution)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: Almost!!! This only happens for v2 billing projects because their name doesn't have to be a valid google project id. When we ask google about a google project with an invalid id we get a 400 back and then this actor marks the attaempt to sync the billing project as failed. Note that billing account changes cannot be applied to v1 billing projects that have not been completed successfully.

`billingProject.googleProjectId` *always* returns the billing project name wrapped in a `GoogleProjectId` (really this should be `Option[GoogleProjectId]` but unfortunately we have no good way of telling that an arbitrary billing project is v1 or v2).

Does that help?

andy7i: I see, so we're storing the name of the billing project as the `googleProjectId` even for v2 billing projects.

I think we could still run into this issue even with this fix, if a v2 billing project has a name that is a valid v1 billing project name?

Could we check for / handle the (400) error we expect?

ehigham: We're only storing the name of the billing project. `googleProjectId` is a method that returns the name of the billing project wrapped in a `GoogleProjectId`.
I don't think checking for a 400 error is the right thing to because you can get 400s in more ways than an invalid project id (for example, future changes to that dao).

andy7i: oh, I didn't realize (or forgot) that's what `googleProjectId` did.

what are your thoughts on this:

> I think we could still run into this issue even with this fix, if a v2 billing project has a name that is a valid v1 billing project name?

I can see this change being worth merging because it probably covers most of the projects. 

I'm not sure if this will make it into the monolith release though, since it is scheduled to go out at 4pm.

Alternatively, we could revert the change, but ideally we won't need to..

andy7i (approved):
I don't want to block this (partial?) fix so here's a 👍  as long as we have a plan for fixing the remaining projects.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/WorkspaceBillingAccountActor.scala
ehigham: If the billing project name is a valid google project id then google will give us a 404 or 403 when we try to access a google project with the same name.

ehigham: If we wanted to make doubly-sure that it's a google project attributed to a billing project that that particular rawls instance created then we need to ask sam. I don't think we can't do that yet as we don't have the admin apis. For example, if you create a billing project in a fiab called `broad-dsde-dev`, without asking Sam, rawls would try to change the billing acount on that google project.

ehigham: I've just realised that this is a potential security vulnerability. We should back out the original change.

ehigham:
Closing. Consider using the Sam admin APIs (when merged) to test that the billing account name does indeed corresponnd to a googpe project resource whose parent is a billing project.

PR #1846, opened 2022-06-01
[AJ-450] Implement metrics for entity/attribute count in cloned workspaces

vitzlo:
This PR sends data regarding cloned workspaces to a histogram to be viewed in Grafana. Possible uses of the data include average, max, and percentile-based entity/attribute count for the cloned workspaces.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
@@ -5,4 +5,4 @@ import slick.jdbc.JdbcProfile
 import scala.concurrent.ExecutionContext
 
 class DataAccessComponent(val driver: JdbcProfile, val batchSize: Int)(implicit val executionContext: ExecutionContext)
-extends DriverComponent with DataAccess
+extends DriverComponent with DataAccess // TODO: resolve metric base name
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
@@ -5,4 +5,4 @@ import slick.jdbc.JdbcProfile
 import scala.concurrent.ExecutionContext
 
 class DataAccessComponent(val driver: JdbcProfile, val batchSize: Int)(implicit val executionContext: ExecutionContext)
-extends DriverComponent with DataAccess // TODO: resolve metric base name
+extends DriverComponent with DataAccess
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
-import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
@@ -78,7 +77,7 @@ class EntityTableWithInlineAttributes(tag: Tag) extends EntityTableBase[EntityRe
 }
 
 //noinspection TypeAnnotation
-trait EntityComponent extends RawlsInstrumented {
+trait EntityComponent {
   this: DriverComponent
     with WorkspaceComponent
     with AttributeComponent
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -536,6 +538,17 @@ trait EntityComponent {
...
+        var entityCount = 0
+        workspaceQuery.findByIdOrFail(clonedWorkspaceId.toString).map(w => {
+          listEntities(w).map(e => {
+            entityCount += 1
+            e
+          })
+          clonedWorkspaceEntityHistogram += entityCount
+          clonedWorkspaceAttributeHistogram += w.attributes.size
+        })
+
```
davidangb: check out `cloneEntitiesToNewWorkspace` at line 917 - I think it already returns/calculates all of the counts you need, without involving an extra database query!

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
```diff
@@ -5,4 +5,4 @@ import slick.jdbc.JdbcProfile
 import scala.concurrent.ExecutionContext
 
 class DataAccessComponent(val driver: JdbcProfile, val batchSize: Int)(implicit val executionContext: ExecutionContext)
-extends DriverComponent with DataAccess
+extends DriverComponent with DataAccess // TODO: resolve metric base name
```
davidangb: nit: is this comment still valid/necessary?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -2,6 +2,8 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
+import nl.grons.metrics4.scala.Gauge
```
davidangb: nit: unused import for `Gauge`, delete this line

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.rawls.metrics
 
-import nl.grons.metrics4.scala.{Counter, Timer}
+import nl.grons.metrics4.scala.{Counter, Gauge, Histogram, Timer}
```
davidangb: nit: unused import for `Gauge`

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -71,6 +71,22 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
...
+      .expand(WorkflowStatusMetricKey, "cloned_ws_entities")
+      .asHistogram("count")
+
+  /**
+    * A gauge to track the number of attributes in each cloned workspace.
+    */
+  protected def clonedWorkspaceAttributeHistogram: Histogram =
+    ExpandedMetricBuilder
+      .expand(WorkflowStatusMetricKey, "cloned_ws_attributes")
+      .asHistogram("count")
```
davidangb, line 89: 1. Update the comments to refer to histograms, not gauges
2. I think these should use `WorkspaceMetricKey` instead of `WorkflowStatusMetricKey`, right?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -536,6 +538,12 @@ trait EntityComponent {
 
         val sourceShardId = determineShard(clonedWorkspaceId)
         val destShardId = determineShard(newWorkspaceId)
+
+        cloneEntitiesToNewWorkspace(clonedWorkspaceId, newWorkspaceId).map({p =>
+          clonedWorkspaceEntityHistogram += p._1
+          clonedWorkspaceAttributeHistogram += p._2
+        })
```
davidangb: oooh calling `cloneEntitiesToNewWorkspace` a second time here would cause problems. `cloneEntitiesToNewWorkspace` actually performs the database inserts - so adding it here would mean that for every workspace-clone operation the user requests, we copy every single entity and attribute TWICE - obviously we don't want that.

I should have been clearer about my suggestion earlier … I think that instead of updating the histogram values here, you should delete this block of code and move the
```
          clonedWorkspaceEntityHistogram += p._1
          clonedWorkspaceAttributeHistogram += p._2
```
lines inside the `yield` clause of `cloneEntitiesToNewWorkspace` (changing `p._1` to `entitiesCopiedCount` and `p._2` to `attributesCopiedCount`). So, you're updating the histograms right at the code location where the copying happens.

davidangb (changes requested):
see important change request inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessComponent.scala
davidangb: this comment is definitely not necessary now :)

davidangb (approved):

PR #1850, opened 2022-06-02
[AJ-360] Make EntityAttributes case sensitive

jordansafer:
Changes:
- Removed case-ignoring calls for EntityAttributes: toLowerCase and equalsIgnoreCase
- (Removed some equalsIgnoreCase calls that impacted entityType joins for expression evaluation)
- Added a liquibase change to make the indices for namespace and name case sensitive
- Added tests for creating, deleting, and updating EntityAttributes

Notes:
- How do we to handle filtering? Intuitively, case-insensitive filtering may be expected behavior and useful (especially in the event of typos). We could follow the pattern use for tags and make a case insensitive query for filtering
- Workspace Attributes, specifically tags, appear to be case insensitive, and this is unchanged by this PR. I expect Submission Attributes are also case insensitive
- ~~Rollback configuration for the liquibase needs to be added. Working on adding it to this PR~~
- Fix tests in EntityComponentSpec and SubmissionMonitorSpec

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
```diff
@@ -507,12 +507,12 @@ trait DataRepoBigQuerySupport extends LazyLogging {
         // the given fromTable may be either the from or the to in the relationship model
         // depending on the direction the relationship is being traversed
         val entityJoin = fromTable.name match {
-          case forward if forward.equalsIgnoreCase(relationshipModel.getFrom.getTable) =>
+          case forward if forward.equals(relationshipModel.getFrom.getTable) =>
             val fromColumn = EntityColumn(snapshotModel, fromTable, relationshipModel.getFrom.getColumn)
```
jordansafer: Do we want to be case insensitive on expressions? Note to self, this needs to be tested

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityExpressionQueries.scala
```diff
@@ -85,8 +85,9 @@ trait LocalEntityExpressionQueries {
         attribute <- entityAttributeShardQuery(context.workspaceContext) if entity.id === attribute.ownerId && attribute.name === attrName.name && attribute.namespace === attrName.namespace
       } yield (rootEntityName, entity.name, attribute)
 
+      // TODO discuss this change
       val attributeForNameQuery =
-        if (attrName.namespace.equalsIgnoreCase(AttributeName.defaultNamespace) && attrName.name.toLowerCase.endsWith(Attributable.entityIdAttributeSuffix)) {
```
jordansafer: This change is making the reserved names case sensitive. For example with 'name' vs 'NAME', only 'name' will now be reserved.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -463,6 +471,205 @@ class CaseSensitivitySpec extends AnyFreeSpec with Matchers with TestDriverCompo
...
+        provider.batchUpdateEntities(updateDefinition).futureValue
+
+        // get our entity
+        val entity = provider.getEntity("cat", "005").futureValue
+
+        // make sure all attributes are created
+        exemplarAttributeNames.size should equal(entity.attributes.size)
+        exemplarAttributesMap.keys.foreach { attr => entity.attributes.get(attr).get shouldBe AttributeString(s"${toDelimitedName(attr)}: new-attribute") }
+      }
+    }
```
jordansafer, line 706: Need testing for expression evaluation is also needed on the attribute side-

davidangb:
re: "How do we to handle filtering?" … I think this should just work? Are you seeing a problem in practice here? From code inspection:
* when creating the `all_attribute_values` string, we `.toLowerCase` it: https://github.com/broadinstitute/rawls/blob/4507529d1896658ebccb8b5f6bff966185a34d03/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L109
* when executing the filter, we `.toLowerCase` the search criteria: https://github.com/broadinstitute/rawls/blob/4507529d1896658ebccb8b5f6bff966185a34d03/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L317

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -106,7 +106,7 @@ trait EntityComponent {
 
     private def createAllAttributesString(entity: Entity): Option[String] = {
       val maxLength = EntityComponent.allAttributeValuesColumnSize
-      Option(s"${entity.name} ${collectAttributeStrings(entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]), List(), maxLength)}".toLowerCase.take(maxLength))
+      Option(s"${entity.name} ${collectAttributeStrings(entity.attributes.values.filterNot(_.isInstanceOf[AttributeList[_]]), List(), maxLength)}".take(maxLength))
```
davidangb: ah - related to my earlier comment, I think we want to leave the `.toLowerCase` here, to allow for case-insensitive filtering

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1086,7 +1087,7 @@ trait EntityComponent {
     // Utility methods
 
     private def validateEntity(entity: Entity): Unit = {
-      if (entity.entityType.equalsIgnoreCase(Attributable.workspaceEntityType)) {
+      if (entity.entityType.equals(Attributable.workspaceEntityType)) {
```
davidangb: I think we should also keep this `equalsIgnoreCase`. Since we disallow "workspace" as an entity type we should also disallow "Workspace" and "WORKSPACE", etc.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1107,7 +1108,7 @@ trait EntityComponent {
             (entityType, EntityTypeMetadata(
               typesAndCounts.getOrElse(entityType, 0),
               entityType + Attributable.entityIdAttributeSuffix,
-              typesAndAttrs.getOrElse(entityType, Seq()).map(AttributeName.toDelimitedName).sortBy(_.toLowerCase)))
+              typesAndAttrs.getOrElse(entityType, Seq()).map(AttributeName.toDelimitedName).sorted))
```
davidangb: I *think* this is the right thing to do, but we might want to check with UI folks here too. This will enforce case-sensitive default ordering of columns in the data table. Certainly the UI could work around that if it wants case-insensitive sorting, but that's an explicit change we'd have to make

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/datarepo/DataRepoBigQuerySupport.scala
davidangb: oooh, I don't think we should change anything about BigQuery (snapshot-by-reference) handling of case - at least yet. We haven't really tested that and don't know what the implications are

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/LocalEntityExpressionQueries.scala
davidangb: I _think_ this means that current expressions of like `this.sample_ID` will no longer work. I suspect we want these to be case-insensitive but agree with your comment - we should discuss and loop in a few others

core/src/main/scala/org/broadinstitute/dsde/rawls/expressions/parser/antlr/DataRepoEvaluateToAttributeVisitor.scala
```diff
@@ -23,7 +23,8 @@ class DataRepoEvaluateToAttributeVisitor() extends TerraExpressionBaseVisitor[Se
 
   override def visitEntityLookup(ctx: EntityLookupContext): Seq[ParsedEntityLookupExpression] = {
     val relations = ctx.relation().asScala
-    Seq(ParsedEntityLookupExpression(relations.map(_.attributeName().getText).toList, ctx.attributeName().getText.toLowerCase, ctx.getText))
+    // TODO discuss this change
+    Seq(ParsedEntityLookupExpression(relations.map(_.attributeName().getText).toList, ctx.attributeName().getText, ctx.getText))
```
davidangb: suggest we don't make changes to snapshot-by-reference handling

model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala
```diff
@@ -76,8 +76,9 @@ trait StringValidationUtils {
   }
 
   def validateAttributeName(an: AttributeName, entityType: String): Unit = {
-    if (Attributable.reservedAttributeNames.exists(_.equalsIgnoreCase(an.name)) ||
-      AttributeName.withDefaultNS(entityType + Attributable.entityIdAttributeSuffix).equalsIgnoreCase(an)) {
+    // TODO discuss, we are allowing case insensitive matches of reserved names
+    if (Attributable.reservedAttributeNames.exists(_.equals(an.name)) ||
+      AttributeName.withDefaultNS(entityType + Attributable.entityIdAttributeSuffix).equals(an)) {
```
davidangb: I need to sit on this one too. I am leaning towards NOT allowing this - let's release case-sensitivity in all other places but keep reserved names case-insensitive. We could always make name reservations sensitive later, but can't go in the other direction.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -463,6 +471,205 @@ class CaseSensitivitySpec extends AnyFreeSpec with Matchers with TestDriverCompo
...
+    "for attribute names" - {
+      "should return all attributes in uncached metadata requests" in withTestDataServices { _ =>
+        // save case insensitive attribute data
+        runAndWait(entityQuery.save(testWorkspace.workspace, caseInsensitiveAttributeData))
+        // get provider
+        val provider = new LocalEntityProvider(testWorkspace.workspace, slickDataSource, false, "metricsBaseName")
+        // get metadata
+        val metadata = provider.entityTypeMetadata(false).futureValue
+        metadata("cat").attributeNames.size shouldEqual exemplarAttributeNames.size
+        metadata("cat").attributeNames should contain allElementsOf exemplarAttributeNames
```
davidangb: I think here you want to use `should contain theSameElementsAs` instead of `should contain allElementsOf` … else the test will pass even if the left-hand side contains extra values (this applies to all the other tests too)

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -463,6 +471,205 @@ class CaseSensitivitySpec extends AnyFreeSpec with Matchers with TestDriverCompo
...
+          val updatedEntity = services.entityService.updateEntity(testWorkspace.wsName, "cat", "005", attributeUpdates).futureValue
+
+          // make sure all attributes are returned, and only the updated attribute is updated
+          updatedEntity.attributes.size shouldBe exemplarAttributeNames.size
+          updatedEntity.attributes.foreach { attributeToCheck =>
+            if (attributeToCheck._1 == attributeToUpdate)
+              attributeToCheck._2 shouldBe AttributeString("new-value")
+            else
+              attributeToCheck._2 shouldBe AttributeString("bar")
+          }
```
davidangb, line 629: Scala idioms: you should be able to use a `case` partial function here to ease tuple handling and avoid the `._1` and `._2`:
```scala
          updatedEntity.attributes.foreach {
            case (attributeName, attributeValue) =>
              if (attributeName == attributeToUpdate)
                attributeValue shouldBe AttributeString("new-value")
              else
                attributeValue shouldBe AttributeString("bar")
          }
```
no need to change the PR, this is just for your knowledge

davidangb (commented):
some comments inline, mostly looking good!

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -21,6 +21,8 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.time.{Millis, Span}
 
 import scala.concurrent.ExecutionContext
+import org.broadinstitute.dsde.rawls.model.AttributeName.toDelimitedName
+import slick.jdbc.MySQLProfile.api._
```
davidangb: the Rawls convention is to `import driver.api._` as needed, at the class level e.g. https://github.com/broadinstitute/rawls/blob/f704fc8e693f0c14bcaba78d9e9b7b53e77daa7c/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccessSpec.scala#L11
or even for an individual test, e.g.
https://github.com/broadinstitute/rawls/blob/b844baa8e6179489dccfe5d6046ca28736c68821/core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala#L378

I suspect this is functionally the same as your import - and I honestly don't know if one is better than the other - but suggest sticking with the convention for consistency

davidangb (commented):
one small comment about importing `driver.api._`, otherwise looking good!

sagehen03 (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
jordansafer: Fixed! Moved this to the test level-

davidangb (approved):
:+1:, once tests pass

PR #1851, opened 2022-06-03
[WOR-192] Add a default bucket location and set it to us-central1

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/browse/WOR-192)
We want to default new workspaces to the `us-central1` GCS region going forward given upcoming pricing changes for the US multi-region.

## This PR
* Adds a config value for the default GCS region and threads it through to the workspace creation codepath
* Updates tests to reflect the new config value

----------

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -28,6 +28,7 @@ health-monitor-dispatcher {
 }
 
 gcs {
+  defaultLocation = "us-central1"
   bucketLogsMaxAge = "180"
   google-api-uri = "https://www.googleapis.com"
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2503,8 +2503,8 @@ class WorkspaceService(protected val userInfo: UserInfo,
   // workspace, if no bucket location is provided, then the cloned workspace's bucket will be created in the same region
   // as the source workspace's bucket. Rawls does not store bucket regions, so in order to get this information we need
   // to query Google and this query costs money, so we need to make sure that the target Google Project is the one that
-  // gets charged. If neither a bucket location nor a source bucket name are provided, Future[None] is returned which
-  // will result in the default bucket location being used
+  // gets charged. If neither a bucket location nor a source bucket name are provided, a default bucket location from
+  // the rawls configuration will be used
   private def determineWorkspaceBucketLocation(maybeBucketLocation: Option[String], maybeSourceBucketName: Option[String], googleProjectId: GoogleProjectId): Future[Option[String]] = {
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2509,7 +2509,7 @@ class WorkspaceService(protected val userInfo: UserInfo,
     (maybeBucketLocation, maybeSourceBucketName) match {
       case (bucketLocation@Some(_), _) => Future(bucketLocation)
       case (None, Some(sourceBucketName)) => gcsDAO.getRegionForRegionalBucket(sourceBucketName, Option(googleProjectId))
-      case (None, None) => Future(None)
+      case (None, None) => Future(Some(config.defaultLocation))
```
aherbst-broad, line 2512: Previously passing `None` meant we were using the default of `US`.

core/src/main/resources/reference.conf
```diff
@@ -28,6 +28,7 @@ health-monitor-dispatcher {
 }
 
 gcs {
+  defaultLocation = "us-central1"
```
cahrens, line 31: This is just the reference conf file-- doesn't the actual rawls conf template in firecloud-develop need to be updated also?

cahrens: (there's also a reference.conf in a test directory… would it make sense to define the defaultLocation there as well?)

aherbst-broad: I don't think we need to update the firecloud-develop file unless there are environment or run-context specific overrides we want for this value (which we don't).

Will do re: test directory.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: nit: can we update the comment above as well to reflect this?

core/src/main/resources/reference.conf
marctalbott: There are quite a few config values that aren't environment or run-context specific in the base-configs for Rawls and other services. I like that this is closer to the code, but it can also be tricky to find config values when they're defined in multiple places

marctalbott (approved):

core/src/main/resources/reference.conf
cahrens: I was confused by the name "reference.conf". I thought this is just an example conf, and what firecloud-develop produces is what we actually  use.

cahrens (approved):

PR #1856, opened 2022-06-06
auto-cast numbers and booleans to strings when populating submissions [AJ-457]

davidangb:
For local (MySQL) entities only - i.e. not snapshot-by-reference - provide some auto-casting of types when populating input values for Cromwell submissions.

If:
* any given method input is defined as a string, an optional string, or an array of strings
* the input expression evaluates to an AttributeNumber, AttributeBoolean, or AttributeValueList containing AttributeNumbers, AttributeBooleans

then:
* grab the value inside the AttributeNumber/AttributeBoolean, call toString on it, and return it as an AttributeString.

This is in support of AJ-339: we want to change TSV uploads to detect numbers and booleans instead of always blindly creating strings. However, some users have existing workflows with inputs defined as strings, since TSV uploads historically have created strings. These workflows will break for newly-uploaded TSVs if those TSVs create numbers/booleans. Hence, we will auto-cast; this should allow those workflows to continue functioning.


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
```diff
@@ -24,9 +24,42 @@ trait ExpressionEvaluationSupport {
...
+      case n:AttributeNumber => AttributeString(n.value.toString) // n.value.bigDecimal.toPlainString to avoid scientific notation
+      case b:AttributeBoolean => AttributeString(b.value.toString)
+      case l:AttributeValueList =>
+        val convertedValues = l.list.map(maybeConvertToString) collect {
+          case av:AttributeValue => av
+        }
+        AttributeValueList(convertedValues)
+      case _ => rawValue
+    }
+  }
```
davidangb, line 49: reviewer: what do you think about scientific notation? For very large or very small `AttributeNumber` values, the `toString` will generate scientific notation. Should we leave that as-is since it is the Java default? Or should we use `toPlainString` even if that results in e.g. many many zeros after a decimal point?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
```diff
@@ -24,9 +24,42 @@ trait ExpressionEvaluationSupport {
...
+  protected def isStringInputType(input: MethodInput): Boolean = {
+    val valuetype = input.workflowInput.getValueType
+
+    // TODO: map types? tuple types?
+    valuetype.getTypeName == TypeNameEnum.STRING ||
+      (valuetype.getTypeName == TypeNameEnum.OPTIONAL &&
+        valuetype.getOptionalType.getTypeName == TypeNameEnum.STRING) ||
+      (valuetype.getTypeName == TypeNameEnum.ARRAY &&
+        valuetype.getArrayType.getTypeName == TypeNameEnum.STRING)
+  }
```
davidangb, line 35: I have reached out to Batch to ask about map and tuple types, perhaps we need to handle those too.

jordansafer:
I like this change in that I'm confident it won't break things (we're only casting when things are incompatible). I'm not confident about the direction we are going. If we're autocasting to Strings, this takes away from  the value of providing types in the first place, adding a very common case where types become unenforced. I don't know that we'll have a point where we can stop autocasting, since it would depend on old workflows updating types, which may depend on old datasets updating to be strongly typed.

TLDR: This change is safe, and is a one-way door at the point where we start adding types at import-time. We need it to unblock us for adding said types, so I want to make sure this call out is included & was considered, since it will be hard to reverse this change once we've committed to it.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
```diff
@@ -24,9 +24,42 @@ trait ExpressionEvaluationSupport {
...
+
+  protected def maybeConvertToString(rawValue: Attribute): Attribute = {
+    // cast numbers and booleans to strings; leave other types alone
+    rawValue match {
+      case n:AttributeNumber => AttributeString(n.value.toString) // n.value.bigDecimal.toPlainString to avoid scientific notation
+      case b:AttributeBoolean => AttributeString(b.value.toString)
+      case l:AttributeValueList =>
+        val convertedValues = l.list.map(maybeConvertToString) collect {
+          case av:AttributeValue => av
```
jordansafer, line 44: what does the `case av => av` do? Does it unbox?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
jordansafer: I think we want `toPlainString` in order to preserve the number. I'm imagining that if a user uploading a very long number as a TSV (in the old/current Terra), we'd import it as a string and pass along a very long string to the workflow. So I think we'd break the workflow if we introduced scientific notation when casting

jordansafer: Thinking more about this, Batch may be able to parse the scientific notation to a number just as well. But I'm not confident about this, so vote is to keep things the same (toPlainString)

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupportSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  // define a map of (input attribute value) -> (expected result from castToString)
+  val stringConversions = Map(
+    AttributeNumber(123) -> AttributeString("123"),
+    AttributeNumber(123.4567) -> AttributeString("123.4567"),
+    AttributeNumber(Double.MaxValue) -> AttributeString("1.7976931348623157E+308"), // should we disable scientific notation??
+    AttributeNumber(Double.MinValue) -> AttributeString("-1.7976931348623157E+308"), // should we disable scientific notation??
+    AttributeNumber(Double.MinPositiveValue) -> AttributeString("4.9E-324"), // should we disable scientific notation??
+    AttributeBoolean(true) -> AttributeString("true"),
+    AttributeBoolean(false) -> AttributeString("false"),
```
jordansafer, line 91: technically we could cause a case change: User uploads `"False"` -> converted to boolean `false`  -> converted to string `"false"`. This should be fine since workflows likely are converting the string back to a boolean before checking against it

jordansafer (approved):
Everything looks good!! Added a couple minor comments and thoughts, thank you for doing this!!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
davidangb: `l.list.map(maybeConvertToString)` returns a `Seq[Attribute]`, but we know from business logic that all of those `Attribute`s will be `AttributeValue`s. So, I'm using the `collect { case av:AttributeValue` as a way of safely casting `Seq[Attribute]` to `Seq[AttributeValue]`. The latter is what we need to pass into `AttributeValueList(…)`.

If that's what you mean by unboxing then you're totally correct!

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupportSpec.scala
davidangb: This will also depend on how lenient we are during TSV parsing. The previous PR (https://github.com/broadinstitute/firecloud-orchestration/pull/929/files#diff-acb77f680e9d2503b8afd1827a0355015e278ef1313339a6acdfa621788b5f3bR154) allowed both "false" and "False" as boolean strings, so your comment holds, but we could change the TSV parsing when we roll it out again. That said, I hope this is a non-issue!

davidangb: Adding on: as soon as we parse the TSV into a boolean, we've lost its original string case - so users could upload "FaLsE" in their TSV but they'd see "false" in the UI, even if they never submit a workflow

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
aednichols: I don't think it's _too_ likely that users are turning strings into native composite types in their WDLs.

When I imagine them using ints or bools as strings, I think stuff like this:

`reticulate_splines` = `"true"`
`memory_gb` = `"4"`
```
command <<<
    java -jar my_tool.jar -Xmx~{memory_gb}g -Dreticulate_splines=~{reticulate_splines}
>>>
```

Which is to say, placing ints or bools in places where they would get cast to strings anyway.

If they were going to write a WDL that takes in a map as a string, and then really uses it as a map, they would need to introduce an explicit [`read_map()`](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#mapstring-string-read_mapstringfile) call that is surely more effort than having the input be a `Map` in the first place.

That said, this is only in my imagination, and any firsthand experience with how users interact with this behavior is helpful!

aednichols: Also, the Terra data table isn't going to automatically try to cast stringified maps or tuples to actual maps or tuples, is it?

i.e. `"{\"a\": \"b\"}"` to `{"a": "b"}`

If no, then there is no need to _reverse_ that conversion for compatibility on the backend, when we go to submit a workflow.

davidangb: no, the data table doesn't do any de-stringification. Thank you for the feedback here! I'm going to leave the PR as-is … and of course if we hear complaints/requests later, we can always make another PR.

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
```diff
@@ -24,9 +24,42 @@ trait ExpressionEvaluationSupport {
...
+        }
+        AttributeValueList(convertedValues)
+      case _ => rawValue
+    }
+  }
+
+  // TODO: unit tests
```
calypsomatic: Are there more unit tests to do than the ones included in this PR?

davidangb: ah good catch! I believe the unit tests in this PR cover the functionality, this `TODO` is left over. I will remove the comment.

calypsomatic (approved):

davidangb:
fiab-start failed:
```
12:05:59 [WARN] Timed out rendering configs for leonardo.  Retrying...
12:06:04 [ERROR] Ran out of retries without successful config render for leonardo.
```
Jenkins retest.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/base/ExpressionEvaluationSupport.scala
aednichols: Sounds good! And, thanks for this valuable work.

PR #1867, opened 2022-06-14
[AJ-465] Implement new round of Rawls metrics

vitzlo:
This branch adds instrumentation to the Rawls codebase for tracking the umber of workspaces created, number of cloned workspaces, and cache writing types (on demand vs. auto). Will coincide with a PR for the [terra-helmfile](https://github.com/broadinstitute/terra-helmfile) repository.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
+import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
@@ -77,7 +78,7 @@ class EntityTableWithInlineAttributes(tag: Tag) extends EntityTableBase[EntityRe
 }
 
 //noinspection TypeAnnotation
-trait EntityComponent {
+trait EntityComponent extends RawlsInstrumented {
   this: DriverComponent
     with WorkspaceComponent
     with AttributeComponent
@@ -1147,6 +1148,8 @@ trait EntityComponent {
                         entityTypesWithCounts: Map[String, Int],
                         entityTypesWithAttrNames: Map[String, Seq[AttributeName]],
                         timestamp: Timestamp) = {
+      entityCacheSaveCounter.inc()
+
       // TODO: beware contention on the approach of delete-all and batch-insert all below
       // if we see contention we could move to encoding the entire metadata object as json
       // and storing in a single column on WORKSPACE_ENTITY_CACHE
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
-import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
@@ -78,7 +77,7 @@ class EntityTableWithInlineAttributes(tag: Tag) extends EntityTableBase[EntityRe
 }
 
 //noinspection TypeAnnotation
-trait EntityComponent extends RawlsInstrumented {
+trait EntityComponent {
   this: DriverComponent
     with WorkspaceComponent
     with AttributeComponent
@@ -906,9 +905,6 @@ trait EntityComponent extends RawlsInstrumented {
         entitiesCopiedCount <- CopyEntitiesQuery.copyEntities(sourceWs, destWs)
         attributesCopiedCount <- CopyEntityAttributesQuery.copyAllAttributes(sourceWs, destWs)
       } yield {
-        clonedWorkspaceEntityHistogram += entitiesCopiedCount
-        clonedWorkspaceAttributeHistogram += attributesCopiedCount
-
         (entitiesCopiedCount, attributesCopiedCount)
       }
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -2,13 +2,11 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
-import nl.grons.metrics4.scala.Gauge
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
-import org.broadinstitute.dsde.rawls.model.AttributeName.toDelimitedName
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -2,11 +2,11 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
 import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
+import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
```
jordansafer: Are we using this import here?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
```diff
@@ -61,6 +61,7 @@ trait EntityStatisticsCacheSupport extends LazyLogging with RawlsInstrumented {
       case (typeName, typeMetadata) => typeName -> typeMetadata.attributeNames.map(AttributeName.fromDelimitedName)
     }
     val timestamp: Timestamp = new Timestamp(workspaceContext.lastModified.getMillis)
+    opportunisticEntityCacheSaveCounter.inc()
```
jordansafer: q: Do we want to record the metric in the Success case on line 69, so we're reporting when we try & succeed at saving? Or do we want to record the metric every time we try to opportunistically save (even when it fails)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -83,6 +88,7 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
         )
       )
     } else {
+      createdWorkspaceCounter.inc()
```
jordansafer: same question as above: do we want to record the metric when we try, vs when the call succeeds. In this case, I think the 2 numbers are most likely pretty similar, so it may not matter much

jordansafer (commented):
Looking pretty good! A couple small comments and otherwise waiting on tests-

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
vitzlo: Upon initial thought I think succeeded cache saves would hold more value, but that opinion is subject to change if other people dissent. Regardless though, there needs to be consistency with the `entityCacheSaveCounter` counter. The plan was for the two counters to not overlap/get tripped by the same cache save request, but also for the union of the two to cover every cache update.

So more broadly either exactly one of counters moves to another location in the code to keep the aforementioned consistency or we instead implement a counter that covers everything and another one that resides in a subset of the other counter, which one sounds better do you think?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
vitzlo: Deleted, ty

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/local/EntityStatisticsCacheSupport.scala
jordansafer: +1 for keeping the current counters and setting it up so exactly one counter gets tripped-

vitzlo: Got it, I'll move the incrementing to line 69. Relating to that though, is the counter for non-opportunistic saves in `EntityStatisticsCacheMonitor.scala` sufficient enough to uphold the consistency discussed in your first comment?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
vitzlo: In line with the other discussion I think succeeded calls should be tracked, but for this case I'm not entirely sure where to place the incrementation. I moved the line to `WorkspaceService.scala` in my most recent PR but there's probably a better place for it, if you have any input on it

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala
```diff
@@ -113,6 +115,7 @@ trait EntityStatisticsCacheMonitor extends LazyLogging {
     // allow 80% of the per-workspace timeout to be spent calculating the attribute names.
     // note that other statements do not have timeouts and are unbounded.
     val attrNamesTimeout = (timeoutPerWorkspace * .8).toSeconds.toInt
+    entityCacheSaveCounter.inc()
```
davidangb: following on to previous comments, you could move this down into the `yield` block to ensure it only gets incremented on a successful save.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -98,6 +102,7 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
       throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.NotImplemented, "MC workspaces are not enabled"))
     }
 
+    createdMultiCloudWorkspaceCounter.inc()
     traceWithParent("createMultiCloudWorkspace", parentSpan)(s1 =>
         createWorkspace(workspaceRequest, s1)
```
davidangb: to only increment the metric on successful workspace creation, you could move the `inc()` into the result of `createWorkspace()`, like so:
```
    traceWithParent("createMultiCloudWorkspace", parentSpan) { s1 =>
      createWorkspace(workspaceRequest, s1) andThen {
        case scala.util.Success(_) => createdMultiCloudWorkspaceCounter.inc()
      }
    }
```
or
```
    traceWithParent("createMultiCloudWorkspace", parentSpan) { s1 =>
      createWorkspace(workspaceRequest, s1) map { ws =>
        createdMultiCloudWorkspaceCounter.inc()
        ws
      }
    }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -163,12 +163,13 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+  def createWorkspace(workspaceRequest: WorkspaceRequest, parentSpan: Span = null): Future[Workspace] = {
+    createdWorkspaceCounter.inc()
+    traceWithParent("withAttributeNamespaceCheck", parentSpan)(s1 => withAttributeNamespaceCheck(workspaceRequest) {
       traceWithParent("withWorkspaceBucketRegionCheck", s1)(s2 => withWorkspaceBucketRegionCheck(workspaceRequest.bucketLocation) {
         traceWithParent("withBillingProjectContext", s1)(s2 => withBillingProjectContext(workspaceRequest.namespace, s2) { billingProject =>
           for {
-            workspace <- traceWithParent("withNewWorkspaceContext", s2) (s3 => dataSource.inTransactionWithAttrTempTable(Set(AttributeTempTableType.Workspace))({ dataAccess =>
+            workspace <- traceWithParent("withNewWorkspaceContext", s2)(s3 => dataSource.inTransactionWithAttrTempTable(Set(AttributeTempTableType.Workspace))({ dataAccess =>
               withNewWorkspaceContext(workspaceRequest, billingProject, sourceBucketName = None, dataAccess, s3) { workspaceContext =>
                 DBIO.successful(workspaceContext)
```
davidangb, line 174: and, you could move the `createdWorkspaceCounter.inc()` into here, just above the `DBIO.successful`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -832,7 +835,8 @@ class WorkspaceService(protected val userInfo: UserInfo,
                                 }
                                 DBIO.seq(inserts: _*)
                               } andThen {
-                              DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
+                                DBIO.successful((sourceWorkspaceContext, destWorkspaceContext))
```
davidangb, line 838: move the `clonedWorkspaceCounter.inc()` here, above the DBIO.successful

davidangb (commented):
comments inline, mostly about how to only increment counters on success

davidangb (approved):

Qi77Qi:
fyi there's https://github.com/broadinstitute/workbench-libs/blob/develop/openTelemetry/src/main/scala/org/broadinstitute/dsde/workbench/openTelemetry/OpenTelemetryMetrics.scala#L19-L33 for adding metrics

PR #1870, opened 2022-06-15
[CA-1781] Sync Workspace and Billing Project Billing Account Changes with Google

ehigham:
This change extends the BillingAccountChangeSynchronizer (nee WorkspaceBillingAccountMonitor) code to update billing account changes on v1 billing projects even if there are no workspaces in the billing project. Most of this has been reviewed before in https://github.com/broadinstitute/rawls/pull/1694, only updated to use Sam's admin APIs.

Using the new BillingAccountChanges table, the actor

- gets the latest billing account change to all billing projects, ignoring the previous unsync'ed changed
- attempts to update the billing account on the billing project google project, if sam says it's a child resource of the billing project
- gets all workspaces in the billiong project
  - attempts to update the billing account on the workspace google project if v2
  - writes the state of the change to the workspace (ie billing account and billing account error message)

I've changed this file to use the more modern akka-actor-typed interface. I believe this has helped tremedously with testing behaviour directly, removing the need for polling.

In a FiaB, manually verified:
- changing billing accounts for empty v1 billing projects works
- billing accounts of v2 workspace google projects are updated.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -472,22 +463,38 @@ trait WorkspaceComponent {
...
+
+    def withoutGoogleProjectId(googleProjectId: GoogleProjectId): WorkspaceQueryType =
+      query.filter(_.googleProjectId =!= googleProjectId.value)
+
+    // setters
+    def setCurrentBillingAccountOnGoogleProject(billingAccount: Option[RawlsBillingAccountName]): WriteAction[Int] =
+      query.map(_.currentBillingAccountOnGoogleProject).update(billingAccount.map(_.value))
+
+    def setBillingAccountErrorMessage(message: Option[String]): WriteAction[Int] =
+      query.map(_.billingAccountErrorMessage).update(message)
```
gpolumbo-broad, line 489: I really like the renaming here, 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -0,0 +1,312 @@
...
+    inTransaction {
+      BillingAccountChanges
+        .latestChanges
+        .unsynced
+        .take(1)
+        .result
+    }.map(_.headOption)
+
+
+  private def syncBillingAccountChange[F[_]](implicit R: Ask[F, BillingAccountChange], M: MonadThrow[F], L: LiftIO[F])
```
gpolumbo-broad, line 86: This is some signature.  Why does the `BillingAccountChange` (and/or the `Ask`?) need to be implicit?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -0,0 +1,312 @@
...
+                                                           (implicit R: Ask[F, BillingAccountChange], M: Monad[F], L: LiftIO[F])
+  : F[Unit] =
+    for {
+      // v1 billing projects are backed by google projects but we have no quick way of telling if
+      // an arbitrary billing project is v1 or not. We don't want to blindly set the billing account
+      // on any google project with an id equal to the billing project name though.
+      // We can use Sam's resource Admin APIs to list the child resources of the billing project. If
+      // a google project resource is listed then we can be sure that this is a v1 billing project.
+      // What follows assumes that the billing project exists and was created successfully.
+      isV1BillingProject <- L.liftIO {
```
gpolumbo-broad: This `for` comp could be extracted into its own method somewhere that we might be able to reuse it elsewhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -313,20 +320,6 @@ trait WorkspaceComponent {
...
-          (workspace.currentBillingAccountOnGoogleProject =!= billingProject.billingAccount ||
-            workspace.currentBillingAccountOnGoogleProject.isEmpty =!= billingProject.billingAccount.isEmpty)
-      } yield (workspace.googleProjectId, billingProject.billingAccount, workspace.currentBillingAccountOnGoogleProject)
-      query.result.map(results => results.map {
-        case (googleProjectId, newBillingAccount, fromWorkspaceBillingAccount) =>
-          (GoogleProjectId(googleProjectId), newBillingAccount.map(RawlsBillingAccountName), fromWorkspaceBillingAccount.map(RawlsBillingAccountName))
-      }.toSet)
-    }
-
     def updateWorkspaceBillingAccount(googleProjectId: GoogleProjectId, newBillingAccount: Option[RawlsBillingAccountName]): WriteAction[Int] = {
```
marctalbott: can we remove this and `updateWorkspaceBillingAccountErrorMessages` now that we've added `setCurrentBillingAccountOnGoogleProject` and `setBillingAccountErrorMessage`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -0,0 +1,312 @@
...
+
+
+  private def writeUpdateBillingProjectOutcome[F[_]](billingProbeCanAccessBillingAccount: Boolean, outcome: Outcome)
+                                                    (implicit R: Ask[F, BillingAccountChange], M: Monad[F], L: LiftIO[F])
+  : F[Unit] =
+    for {
+      change <- R.ask
+      (_, message) = Outcome.toTuple(outcome)
+      _ <- if (outcome.isSuccess)
+        info("Successfully updated Billing Account on Billing Project in Google") else
```
marctalbott, line 162: Can we log the name of the billing project in both of these messages?

gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
ehigham: Good observation -  the `BillingAccountChange` is effectively passed around implictly for the only reason that it removes one argument everywhere. This allows me to log information without being explict about all the information I'm logging (see the logContext funciton below).
`Ask` is a generalisation of `Reader` that you met in the `WorkspaceMigrationActor`. It abstracts dependecy injection such that I can pass in a `Reader[BillingAccountChange]` OR i can use a function that suppplies a `BilllingAccountChange` or I supply any structure that allows me to fetch a billing account change.
It's pretty neat and while I haven't made the most of what `Ask` has to offer, it allows me to abstract the concrete type I flatmap over. I wish I had used this in the migration code tbh.
That's rather a lot - I'd be glad to discuss it futher in a call :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
ehigham: Indeed it could but I wasn't sure what else would benefit from this (other than our past-lives CI scripts). Would you prefer that I do this? Where should it live?

gpolumbo-broad: I have a preference for smaller methods and breaking out logical chunks into their own methods so they can be named something descriptive. I subjectively find all of that easier to reason over and I know others find it easier to read if it's just laid out in one place.  Therefore, if you're back in this code for something else to get your requisite thumbs on this PR, then I'd prefer to see it extracted.  However, if there are no other review comments or reasons to make another commit, I do not think you need to come in and refactor this.  I have already approved the PR, so pending any major changes from other reviewers, you're good to go in my book.

ehigham: Sure thing, I'll move it into it's own definition :)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
ehigham: I thought I'd have a second attempt at answering this because you've asked one of those simple questions that has no simple answer. I'll try to be brief but the scope of this question spans many topics including what cats is trying to bring to scala - ie type classes. I know I'm going to make a mess of this...

The `BillingAccountChange` is not implict. What this is saying that the implict parameter `R` supplies the `Ask[_, BillingAccountChange]` type class instance for the type `F`. I've described `Ask` above but the cat's documentation does a good job at explaining it I think https://typelevel.org/cats-mtl/mtl-classes/ask.html. TL;DR is it allows me to read a `BillingAccountChange` object within the context of `F`.

So... What's a type class?
I'll give a hand-wavey description of type classes (and how they differ from object-oriented classes). I thoroughly recommend the following sources for this topic to understand what type classes *really* are:
https://youtu.be/re96UgMk6GQ?t=2583
https://homepages.inf.ed.ac.uk/wadler/papers/class-letter/class-letter.txt

The problem: How can we generalise over the types of object certain operations apply to? 

Consider sorting a list of objects. How can we say "this function is defined for all orderable objects"? How do we encode a generic "compare" operation?
An object-oriented approach might be to define the interface `Ordered` and to have all orderable types implement that interface. In Scala, we might write
```scala
// interface
trait Ordered[A] {
  def compare(another: A): Int
}

// sub-type
class MyType extends Ordered[MyType] {
  override def compare(another: MyType): Int = ???
}

def sort[A: Ordered](xs: List[A]): List[A] = ???
```
This approach is called "sub-typing" - the generic operation `compare` is defined on a type if that type is a sub-type of (extends) the interface that defines that operation. Generic operations are defined internally to the type (ie within the definition of that type) and are often used in a hierarchy of types. Which definition to call is resolved at runtime via https://en.wikipedia.org/wiki/Dynamic_dispatch.

Purely functional languages like Haskell were designed without type hierarchies. Instead, they encode a generic operation by defining a class of types (think "set") that declare that operation and/or state certain laws that those class of types hsould obey. We then write instances of that type class that define the specific implementation for a type. Instead of saying that type X "is a" Y, we say type X is an "instance" or "member of" class of types Y. In Scala, we could write: 
```scala
case class MyType { .. }

// type class
trait Orderable[A] {
  def compare(a: A, b A): Int
}

// type class instance (or how it's a member of the class of types)
implict object orderableMyType extends new Orderable[MyType] {
  override def compare(a: MyType, b: MyType) = ???
}

def sort[A](xs: List[A])(implict O: Orderable[MyType]): List[A] = ???
```
There are a few important observations here, one is common to type classes and the others are specific to Scala's (or cat's) implementation of type-classes:
1. Type classes provide ad-hoc polymorphism - support for generic operations does not need to be baked into the type's definition (unlike subtyping).
2. Type classes in Scala are implemented by subtyping because Scala is an object-oriented language and uses dynamic dispatch to resolve method calls.
3. Type class support in scala is supplied by an implict parameter on which generic type-class operations are dispatched, ie 
```scala
// subtyping
def greaterThan[A: Ordered](a: A, b: A): Boolean = 
  a.compare(b) > 0

// type class
def greaterThan[A](a: A, b: A)(implict O: Orderable[A]): Boolean =
  O.compare(a, b) > 0
```

Key take aways:
Type classes provide a means of defining generic operations and ad-hoc implementations without modifying the type's definition.
Types do not implement type classes in their definition. Instead an instance of that type class for that type is defined externally.
In Scala, type class instances are often defined as `implict` and passed as implict parameters to a function.

marctalbott (approved):
I confess that I don't fully understand the value that much of the Cats/type class stuff brings to the table, but it all _looks_ okay. I'll check out some of the resources you linked and might be reaching out for a code walk sometime in the future

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
ehigham: Yep we do. See `logContext` for fields we log.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
ehigham: Yes we can :) Thanks!

PR #1882, opened 2022-06-21
[PPWM-58] Ensure log bucket is empty if no storage logs found

ehigham:
RR: https://broadworkbench.atlassian.net/browse/PPWM-58

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -370,11 +370,9 @@ class HttpGoogleServicesDAO(
         case (None, _) =>
           // No storage logs, so make sure that the bucket is actually empty
           val fetcher = getStorage(getBucketServiceAccountCredential).objects.list(bucketName).setMaxResults(1L)
-          retryWhen500orGoogleError(() => {
-            Option(executeGoogleRequest(fetcher).getItems).exists(!_.isEmpty)
-          }) flatMap { notEmpty =>
-            if (notEmpty) Future.failed(new GoogleStorageLogException("Not Available"))
-            else Future.successful(BucketUsageResponse(BigInt(0), Option(DateTime.now())))
+          retryWhen500orGoogleError(() => Option(executeGoogleRequest(fetcher).getItems)) flatMap {
+            case Some(items) if !items.isEmpty => Future.failed(new GoogleStorageLogException("Not Available"))
+            case _ => Future.successful(BucketUsageResponse(BigInt(0), Option(DateTime.now())))
           }
         case (_, Some(nextPageToken)) => recurse(Option(nextPageToken))
         case (Some(items), None) =>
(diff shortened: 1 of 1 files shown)


dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -371,10 +371,10 @@ class HttpGoogleServicesDAO(
...
           val fetcher = getStorage(getBucketServiceAccountCredential).objects.list(bucketName).setMaxResults(1L)
           retryWhen500orGoogleError(() => {
-            Option(executeGoogleRequest(fetcher).getItems)
-          }) flatMap {
-            case None => Future.successful(BucketUsageResponse(BigInt(0), Option(DateTime.now())))
-            case Some(_) => Future.failed(new GoogleStorageLogException("Not Available"))
+            Option(executeGoogleRequest(fetcher).getItems).exists(!_.isEmpty)
+          }) flatMap { notEmpty =>
+            if (notEmpty) Future.failed(new GoogleStorageLogException("Not Available"))
+            else Future.successful(BucketUsageResponse(BigInt(0), Option(DateTime.now())))
```
marctalbott: Maybe my brain is just slow at 5pm after a long weekend, but this took me a second to reason through. Would it be simpler to have something like this?
```suggestion
            Option(executeGoogleRequest(fetcher).getItems).exists(_.isEmpty)
          }) flatMap { isEmpty =>
            if (isEmpty) Future.successful(BucketUsageResponse(BigInt(0), Option(DateTime.now())))
            else Future.failed(new GoogleStorageLogException("Not Available"))
```

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
ehigham: I don't think so - wouldn't that return `Future.failed blah` if `getItems()` returned `null`?

marctalbott: if `.getItems()` was null, then `Option(executeGoogleRequest(fetcher).getItems)` would be `None` and `.exists` would return false

ehigham: Maybe I'm being thick...

ehigham: Right yeah - so `isEmpty` would be false and we'd return `Future.failed`?

marctalbott: Yeah that's right. It's definitely an aesthetic change more than anything, if you think the current way you have it is more clear then I'm good with that too

ehigham: Do we want to fail if `getItems()` returns null? It didn't before so that's what's confusing me...

marctalbott: Ohhh okay yeah my bad, I did not think it through all the way. Forget this suggestion 🙃

ehigham: I take your point about not being obvious. I'll change it back to use a pattern match!

PR #1894, opened 2022-06-27
Aj 433 by-reference snapshots do not show datarepo_row_id columns, if the snapshot has a primary key

calypsomatic:
AJ-433
Although all snapshots have a `datarepo_row_id`, if it is not named the primary key for a by-reference snapshot, then the entity metadata is unaware of it without making a call to BigQuery.  This adds `datarepo_row_id` to the attributes in the case that it is neither the primary key nor otherwise listed with the attributes.  Test is updated to reflect this.

Note that the second half of AJ-433, duplicate id columns, is treated in the front end.  

Question: since this issue is mainly driven by displays in terra-ui, will this break anything for anyone using firecloud or API only?  Where else does this show up?  Reviewers, please alert me to these concerns!


davidangb (approved):
code looks good, :+1:

I have not tested this myself in conjunction with the UI. I assume you have and that it fixes the missing `datarepo_row_id` column problem!

davidangb:
> will this break anything for anyone using firecloud or API only?

The only case I can imagine this causing a problem is if the `datarepo_row_id` column does NOT exist in the referenced snapshot. If that column will always be present, it should always be safe to include `datarepo_row_id` in the entity-type-metadata response.

PR #1896, opened 2022-06-28
[SU-18] Return submissionRoot in getSubmission and listSubmission responses

MatthewBemis:
Updated Jul 14:

Submission execution currently takes place in the root of the bucket. The consequence of this is that after a user runs a bunch of submissions in their workspace, the root of their bucket becomes cluttered with UUID directories, reducing its usefulness for other files. This has been a frequent point of negative user feedback over the past several years. The end state that we want to be in is that instead of dumping the individual submission directories in the root of the bucket, we dump it in a path like `gs://{workspaceBucketName}/submissions/{submissionId}`.

One big issue is that switching to the `/submissions` path is not backwards compatible. Terra UI and FireCoud UI both make assumptions about the path corresponding to the submission, but once we start writing to `/submissions`, that assumption will no longer be valid. Old submissions will live at the root, new submissions will live under `/submissions`- and it will be impossible to differentiate between them, leading to broken links.

The purpose of this PR is to enable backwards compatibility before switching to `/submissions`. Backwards compatibility will be accomplished by including the `executionPath` in the response for the `listSubmissions` and `getSubmission` endpoints. Terra UI and FireCloud UI will use that field to create the various links in the UI rather than constructing links based off of an assumption.

After Terra UI and FireCloud UI are migrated to use the new field, then we will follow it up with a quick PR to change the execution path for newly-launched submissions.

----------

Old description:

Currently submission outputs are written to the root of the workspace bucket. Users have requested that it be possible to write these outputs to a different directory, to avoid over-cluttering the root of their bucket. This PR allows for a user to specify a full GCS path in their workspace bucket to write the Cromwell outputs.

There are some backwards compatibility concerns with this feature that require it to be rolled out in multiple phases:

Currently Terra UI and FireCloud UI (and possibly other UIs) make assumptions about where these outputs live in the bucket, i.e. `gs://{workspaceBucket}/{submissionId}`. This assumed bucket path is used by the UIs to construct a number of links to the folder in the GCS console. This pattern will no longer hold. And even if we implemented the simpler version of this by writing outputs to a predictable subdirectory within the bucket (i.e. `/submissions/{submissionId}`), that would still break backwards compatibility for old submissions. The UI needs to be able to query Rawls to determine where the outputs live, and construct the links based on that.

Rawls will still output to the root of the workspace bucket by default. After releasing this code, Terra UI will be updated to query Rawls for the `outputPath` and build its external GCS links based on that. At this point, we can also update Terra UI to allow users to specify their desired path at submission launch time.

Because Rawls will still output to the original location by default, FireCloud UI does not need to be updated. There is an edge case where if a user launches a submission with a custom path in Terra UI and tries to view it in FireCloud UI, then the GCS links will be broken. I believe that the overlap between FireCloud UI users and Terra UI users is non-existent so I don't think it's worth it to protect against this. We've already ensured basic backwards compatibility with FireCloud UI.

After all of the above is taken care of, we can switch the default path to be something like `gs://{bucketName}/submissions/{submissionId}`, if we'd like

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220627_add_submission_output_path.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet id="add_submission_output_path_column" author="mbemis" logicalFilePath="dummy">
+        <addColumn tableName="SUBMISSION">
+            <column name="OUTPUT_PATH" type="TEXT">
+                <constraints nullable="false" />
```
MatthewBemis, line 6: todo: this needs to backfill

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,37 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
     } yield {
       SubmissionReport(submissionRequest, submission.submissionId, submission.submissionDate, userInfo.userEmail.value, submission.status, header, submissionParameters.filter(_.inputResolutions.forall(_.error.isEmpty)))
     }
   }
 
   private def prepareSubmission(workspaceName: WorkspaceName, submissionRequest: SubmissionRequest):
-  Future[(Workspace, Stream[SubmissionValidationEntityInputs], Option[WorkflowFailureMode], SubmissionValidationHeader)] = {
+  Future[(Workspace, UUID, Stream[SubmissionValidationEntityInputs], Option[WorkflowFailureMode], SubmissionValidationHeader, String)] = {
+
+    val submissionId: UUID = UUID.randomUUID()
```
MatthewBemis, line 1575: this used to be done in `saveSubmission` but I had to push it up in order to construct the executionPath below

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220627_add_submission_output_path.xml
```diff
@@ -0,0 +1,33 @@
...
+                "/",
+                LOWER(
+                    CONCAT(
+                        SUBSTR(HEX(SUBMISSION.ID), 1, 8), '-',
+                        SUBSTR(HEX(SUBMISSION.ID), 9, 4), '-',
+                        SUBSTR(HEX(SUBMISSION.ID), 13, 4), '-',
+                        SUBSTR(HEX(SUBMISSION.ID), 17, 4), '-',
+                        SUBSTR(HEX(SUBMISSION.ID), 21)
+                    )
+                )
```
MatthewBemis: submission IDs are stored as UUID binary, not strings, so this unfortunate step is necessary to convert them into UUID strings

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,37 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+
     for {
+      workspaceContext <- getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write)
+
       _ <- requireComputePermission(workspaceName)
 
+      outputPath = submissionRequest.outputPath match {
+        case None => s"gs://${workspaceContext.bucketName}/${submissionId.toString}"
+        case Some(path) => {
+          if(path.startsWith(s"gs://${workspaceContext.bucketName}")) {
```
nawatts: Should this condition include a trailing slash after the bucket name?
```scala
if(path.startsWith(s"gs://${workspaceContext.bucketName}/"))
```

As is, wouldn't this accept paths where the bucket name began with the workspace's bucket name?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,37 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+            s"${path}/${submissionId.toString}"
```
nawatts: Should possible trailing slashes be removed from `path` before formatting this?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
nawatts: Also, there are some characters that Google advises against using in object names. Should we further restrict what's accepted for output paths?
https://cloud.google.com/storage/docs/naming-objects#objectnames

The upload page in Terra UI restricts folder names: https://github.com/DataBiosphere/terra-ui/blob/d4533d8e90a2bc1f947da99c9bf2339222b32191/src/pages/Upload.js#L391-L397

nawatts (commented):
A few questions about validating output paths.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: That first point is a good catch.

For the second point- I'll add some input validation

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: You're right- that's a common failure mode waiting to happen

breilly2:
> I believe that the overlap between FireCloud UI users and Terra UI users is non-existent

I think there was a time when people would use the FireCloud UI instead of Job Manager for various reasons. I don't know if that's still the case.

Speaking of Job Manager, it's possible that there are assumptions about output locations being made there as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -235,20 +235,20 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
     }
 
     ExecutionServiceWorkflowOptions(
-      s"gs://${workspace.bucketName}/${submissionId}",
+      jes_gcs_root = submission.outputPath,
```
breilly2: `jes_gcs_root = ` looks odd to me here. Is it needed?

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
```diff
@@ -250,14 +250,14 @@ class WorkflowSubmissionSpec(_system: ActorSystem) extends TestKit(_system) with
       assertResult(
         Some(
           ExecutionServiceWorkflowOptions(
-            jes_gcs_root = s"gs://${testData.workspace.bucketName}/${testData.submission1.submissionId}",
+            jes_gcs_root = s"gs://${testData.workspace.bucketName}/submissions/${testData.submission1.submissionId}",
```
breilly2: This looks like to me like a change in default behavior. Where is this extra `submissions/` path coming from?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
```diff
@@ -1110,6 +1111,61 @@ class SubmissionApiServiceSpec extends ApiServiceSpec with TableDrivenPropertyCh
     }
   }
 
+  it should "return 400 Bad Request if the outputPath is an external bucket" in {
```
breilly2: Should the API just take the path part and prepend the workspace bucket instead of making the caller prepend the only valid bucket?

breilly2 (commented):
A few initial comments/questions.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
MatthewBemis: No, I can remove it

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionSpec.scala
MatthewBemis: This isn't a behavioral change, I just changed the test data to flex it a bit: https://github.com/broadinstitute/rawls/pull/1896/files#diff-e6dd2e193c5a6757c3575da466015485f9ff152b74db02ffc76a0b5f21bcdf10R154

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiServiceSpec.scala
MatthewBemis: There may be a future where we allow for external buckets, but that was out of scope for this ticket. I wanted to future-proof it for that case

breilly2: Okay. The worry that I had in particular was possibly being able to save outputs from an auth-domain protected workspace to a different bucket with different permissions. There's not much we can do to prevent people from going outside of Terra and explicitly copying data out of one bucket and into another, but any features we build in should respect the access control rules that we have in place.

MatthewBemis: For sure- we could disable external bucket outputs in the case of AD'd workspaces, if we ever get there

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
MatthewBemis: Removed in 63bc53605211aa3dabdf92fba86a0a3c13d79570

MatthewBemis:
@breilly2 I did check JM originally and it did not seem to be making assumptions about GCS paths- it was working with full GCS paths that it was pulling directly from the Cromwell task metadata

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,41 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+          parseGcsPath(path) match {
+            case Left(error) => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath was invalid. ${error.value}"))
+            case Right(GcsPath(GcsBucketName(bucketName), GcsObjectName(objectName, _))) =>
+              if(!bucketName.equalsIgnoreCase(s"${workspaceContext.bucketName}")) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath must be within the workspace bucket gs://${workspaceContext.bucketName}"))
+              if(!objectName.matches(gcsObjectNameRegex.regex)) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath was invalid. See https://cloud.google.com/storage/docs/naming-objects for naming guidelines."))
+
+              val strippedPath = path.stripSuffix("/")
+              s"${strippedPath}/${submissionId.toString}"
+          }
+        }
```
MatthewBemis: @nawatts I've added some validation here. It's not 100% rock-solid, (i.e. it doesn't protect against extremely niche things like the acme-challenge case and may be slightly too restrictive with it's character set), but it should catch all of the discouraged/banned characters that would result in actual errors on Google's side

This block also takes into account your feedback about ensuring that the bucket name specified doesn't just begin with the bucket name, and also it strips off trailing slashes to avoid confusion.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,41 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+        case None => s"gs://${workspaceContext.bucketName}/${submissionId.toString}"
+        case Some(path) => {
+          parseGcsPath(path) match {
+            case Left(error) => throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath was invalid. ${error.value}"))
+            case Right(GcsPath(GcsBucketName(bucketName), GcsObjectName(objectName, _))) =>
+              if(!bucketName.equalsIgnoreCase(s"${workspaceContext.bucketName}")) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath must be within the workspace bucket gs://${workspaceContext.bucketName}"))
+              if(objectName.nonEmpty && !objectName.matches(gcsObjectNameRegex.regex)) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, """The specified outputPath was invalid. Path within bucket must be between 1 and 1024 characters and only contain alphanumeric characters or characters in the following list: !@$%^&(){}|<>/,.=_-."""))
+
+              val strippedPath = path.stripSuffix("/")
+              s"${strippedPath}/${submissionId.toString}"
```
nawatts: Based on the API doc, I would have assumed that workflow outputs would be written to the path specified, without a submission ID folder. But adding the submission ID folder does make it difficult to overwrite outputs from previous submissions.

nawatts (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
MatthewBemis: That's a good point. I will note in the API doc that the outputs will be written to a unique submission-specific folder within the specified path

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1555,22 +1555,42 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+              if(!bucketName.equalsIgnoreCase(s"${workspaceContext.bucketName}")) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"The specified outputPath must be within the workspace bucket gs://${workspaceContext.bucketName}"))
+              if(objectName.nonEmpty && !objectName.matches(gcsObjectNameRegex.regex)) throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, """The specified outputPath was invalid. Path within bucket must be between 1 and 1024 characters and only contain alphanumeric characters or characters in the following list: !@$%^&(){}|<>/,.=_-."""))
+
+              s"${pathFormatted}/${submissionId.toString}"
+          }
+        }
+      }
+
       // getWorkflowFailureMode early because it does validation and better to error early
       workflowFailureMode <- getWorkflowFailureMode(submissionRequest)
```
breilly2, line 1581: This comment seems a little out of date now. Could this still be the first thing we do? If not, I think what you've added is also in the spirit of validation, but that validation also now needs `getWorkspaceContextAndPermissions`. The comment might need to moved, be rephrased, or removed altogether?

MatthewBemis: I've moved it back up near the top of the `for` in 293e12f41f7c73e9021b41fe834596d91fcf58bd

cjllanwarne:
Probably more relevant than the "firecloud vs Terra UI" divide is the upcoming "Terra UI vs CBAS UI" divide. We should remember to make notes of all of these changes and find solutions for them in CBAS UI too.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -235,20 +235,20 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
     }
 
     ExecutionServiceWorkflowOptions(
-      s"gs://${workspace.bucketName}/${submissionId}",
+      submission.outputPath,
```
cjllanwarne: I think this is the gcs root you're changing, not (just) the outputs path. So I think what this will do is let users change their workflow execution root, rather than let them move outputs files away from the other junk generated by workflow execution

cjllanwarne (commented):
Apologies if I've missed some context and/or got my analysis of the changes wrong 😄.

I agree that rawls should be involved in taking a UI-requested outputs directory and passing that along to Cromwell, but I think Cromwell would be the more natural place to:

- Store and retrieve "outputs directory" information.
- Do the copying of files to and from the specified output directory, on workflow completion
- Calculate a default output directory if none is provided.

To do this in Rawls, I think you'll need to change workflow options so as to tell Cromwell to copy outputs around after workflow completion. This PR currently (seems to?) change only where Cromwell sets the root working directory for the entire workflow to.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
MatthewBemis: This seems like a terminology mistake on my part. I was treating the phrases "outputs" and "everything written by Cromwell during execution" as the same thing, but in Cromwell-world it seems like those two are distinct and separate terminology? What we are most interested in doing is allowing the user to specify a path for execution so they can avoid cluttering their workspace bucket in Terra, it's less important to specify a path to copy the "good" files to, leaving the junk behind. I do think exposing more of the Cromwell workflow options in Terra UI would be useful in the future, but I wouldn't want to add that to the scope of this PR. Terra UI also currently refers to the execution path as the "outputs" path, so we may need to change that to be more accurate as well.

To be more accurate with terminology, how does it sound if we reframe this PR as "Allow custom execution path for submissions"? If that works for you, I will go in and update the code accordingly to reflect that.

cjllanwarne: Aha, yes, and that makes a lot more sense. Yes, "execution path", "execution root", "root execution directory", "submission execution directory" or even "submission directory" would all be fine options

MatthewBemis:
jenkins retest - import service never booted fully

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
MatthewBemis: Done! Renamed everything from outputPath -> executionPath in https://github.com/broadinstitute/rawls/pull/1896/commits/87f5d429e93a2ce59ef26bdfe8ad80c17c7b679b

MatthewBemis:
hmmm import service failed to start again. seems like it got hung up on creating the db tables. one more retest before i debug more thoroughly...

jenkins retest

MatthewBemis:
jenkins retest - a docker pull failed during start

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3813,6 +3813,9 @@ components:
           description: Whether or not to delete intermediate output files when the
             workflow completes. See [Cromwell docs](https://cromwell.readthedocs.io/en/stable/wf_options/Google#google-pipelines-api-workflow-options)
             for more information.
+        executionPath:
+          type: string
+          description: The full GCS path indicating where execution took place.
```
cjllanwarne: Nit: It's really more of a "root" than a "path". Actual executions will use paths that are generated by appending to this stub.

cjllanwarne: Just to confirm: this is only used in responses, not requests?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -217,7 +217,7 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
   }
 
   def buildWorkflowOpts(workspace: WorkspaceRecord,
-                        submissionId: UUID,
+                        submission: SubmissionRecord,
```
cjllanwarne, line 220: The changes in this file don't seem necessary any more since only the ID is ever ready anyway, but I guess will make the next PR easier? 

My only question is - I assume this doesn't force rawls to bring anything more back from the DB per request? Just changing what we're passing around references to internally?

cjllanwarne (approved):

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: Yes only in responses- it's left out of the `SubmissionRequest` case class

MatthewBemis: Discussed via Slack- gonna change the terminology to be `submissionRoot` throughout

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
MatthewBemis: Yeah precisely- we're not making any additional queries or anything. Passing around the `SubmissionRecord` means we don't need to hardcode the `submissionRoot` here either so the code is a bit more resilient to future change (I assume that's what you were alluding to with your point about making the next PR easier)

PR #1904, opened 2022-07-05
[WOR-325] Don't reveal workspace namespace/name in getWorkspaceById

blakery:
Fixes [WOR-325](https://broadworkbench.atlassian.net/browse/WOR-325)


blakery:
Merging develop into the branch created a horrific diff, even with the squashed commit.

I'll create a different PR with a better diff.

PR #1907, opened 2022-07-07
[AJ-483] Fix potential faults in RawlsInstrumented objects

vitzlo:
Adds new key/value pair to metrics currently in the "Workspaces" section of Grafana to avoid potential conflicts with the naming. Also separates histogram metrics into three different transmitted values.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
@@ -21,6 +21,8 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
   final val SubsystemMetricKey         = "subsystem"
   final val WorkflowStatusMetricKey    = "workflowStatus"
   final val WorkspaceMetricKey         = "workspace"
+  final val StatsBoardMetricKey        = "statsBoard"
+  final val RawlsBoardMetricValue      = "rawlsBoard"
 
   /**
     * An ExpandedMetricBuilder for a WorkspaceName.
@@ -69,6 +71,7 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
     */
   protected def opportunisticEntityCacheSaveCounter: Counter =
     ExpandedMetricBuilder
+      .expand(StatsBoardMetricKey, RawlsBoardMetricValue)
       .expand(WorkspaceMetricKey, "opportunistic_entity_cache_save")
       .transient()
       .asCounter("count")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -21,6 +21,8 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
   final val SubsystemMetricKey         = "subsystem"
   final val WorkflowStatusMetricKey    = "workflowStatus"
   final val WorkspaceMetricKey         = "workspace"
+  final val StatsBoardMetricKey        = "statsBoard"
+  final val RawlsBoardMetricValue      = "rawlsBoard"
```
jordansafer: I don't think we should tie the metric paths to the dashboard they appear in, since we could always change dashboards. I would vote to group them logically based on function, so something like "workspaceAggregates" as a metric key

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
```diff
@@ -69,6 +71,7 @@ trait RawlsInstrumented extends WorkbenchInstrumented {
     */
   protected def opportunisticEntityCacheSaveCounter: Counter =
     ExpandedMetricBuilder
+      .expand(StatsBoardMetricKey, RawlsBoardMetricValue)
       .expand(WorkspaceMetricKey, "opportunistic_entity_cache_save")
```
jordansafer: My understanding was that we would replace "workspace/opportunistic_entity_cache_save" with "<new metric key>/opportunistic_entity_cache_save". Looks like we haven't replaced it here?

jordansafer (commented):
A couple minor naming/pathing comments

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/RawlsInstrumented.scala
vitzlo: Currently still looking for utility within the metric builder object to support it, empty strings wouldn't suffice I'm pretty sure but can you check me on that? [This line](https://github.com/broadinstitute/rawls/blob/4ef04331af8005419c6e6857f5cff2b37c59e97e/metrics/src/main/scala/org/broadinstitute/dsde/rawls/metrics/WorkbenchInstrumented.scala#L44) and its corresponding logic

jordansafer (approved):

PR #1919, opened 2022-07-11
[PPWM-71] Lock Workspace when Migration Starts

ehigham:
RR: https://broadworkbench.atlassian.net/browse/PPWM-71
Workspaces will be locked by the actor allowing us to schedule a whole
bunch without needing them to be locked when they're scheduled.
Ensure no active submissions when picking a workspace to migrate.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -483,6 +483,9 @@ trait WorkspaceComponent {
 
     def setBillingAccountErrorMessage(message: Option[String]): WriteAction[Int] =
       query.map(_.billingAccountErrorMessage).update(message)
+
+    def setIsLocked(isLocked: Boolean): WriteAction[Int] =
+      query.map(_.isLocked).update(isLocked)
   }
 
   private def groupByWorkspaceId(runningSubmissions: Seq[(UUID, Int)]): Map[UUID, Int] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigration.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigration.scala
@@ -5,7 +5,7 @@ import cats.MonadThrow
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.model.WorkspaceVersions.V1
-import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectId, GoogleProjectNumber, Workspace}
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, GoogleProjectId, GoogleProjectNumber, SubmissionStatuses, Workspace}
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits._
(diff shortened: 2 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -483,6 +483,9 @@ trait WorkspaceComponent {
 
     def setBillingAccountErrorMessage(message: Option[String]): WriteAction[Int] =
       query.map(_.billingAccountErrorMessage).update(message)
+
+    def setIsLocked(isLocked: Boolean): WriteAction[Int] =
```
dvoet, line 487: there is already a `lock` function, can you use that?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigration.scala
```diff
@@ -257,13 +257,22 @@ trait WorkspaceMigrationHistory extends RawSqlQuery {
...
+    def getNumActiveMigrations: ReadWriteAction[Int] =
+      sql"select count(*) from #$tableName where #$startedCol is not null and #$finishedCol is null".as[Int].head
+
+    def nextMigration: ReadWriteAction[Option[(Long, UUID, Boolean)]] =
+      sql"""
+        select m.#$idCol, m.#$workspaceIdCol, w.is_locked from #$tableName m
+        join (select id, is_locked from WORKSPACE) as w on (w.id = m.#$workspaceIdCol)
+        where m.#$startedCol is null and not exists (
+            select workspace_id, status from SUBMISSION
+            where status in (#${SubmissionStatuses.activeStatuses.map("'" + _ + "'").mkString(",")})
```
dvoet: String concatenation like this in sql make me nervous. I know in this particular instance it is safe but I still don't like to have it in code for other people to copy.

So either figure out how to do this with bind parameters or leave a comment saying why it is safe to do it in this case but is generally not a good practice

dvoet (approved):
2 little things then 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigration.scala
ehigham: Could you explain why it's bad practice? I guess I'm not sure what to write in such a comment?
Furthermore, I'm not sure I can do this with inSetBind as we're not using that dsl for the migration table? Is that incorrect?

dvoet: At first glance this looks like it could be vulnerable to sql injection. But we know `SubmissionStatuses. activeStatuses` is not user input so it is safe. But some future engineer might go hunting for how one does an in clause with raw sql and come across this, copy it and might think it is ok to do `userInput.map(...`. Or an auditor might find it and think it looks fishy (I don't know if auditors actually look that closely).

ehigham: Fair enough, I wasn't sure that you were referring to using raw parameters. I'll try to get it to work with bound parameters and failing that, add a comment as you suggest.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
ehigham: That takes the workspace name and I have it's UUID. I prefer to use primary key operations where possible.

PR #1863, opened 2022-06-13
Update grpc-core, grpc-netty-shaded from 1.33.1 to 1.47.0

broadbot:
Updates 
* [io.grpc:grpc-core](https://github.com/grpc/grpc-java)
* [io.grpc:grpc-netty-shaded](https://github.com/grpc/grpc-java)

 from 1.33.1 to 1.47.0.
[GitHub Release Notes](https://github.com/grpc/grpc-java/releases/tag/v1.47.0) - [Version Diff](https://github.com/grpc/grpc-java/compare/v1.33.1...v1.47.0)

I'll automatically update this PR to resolve conflicts as long as you don't change it yourself.

If you'd like to skip this version, you can just close this PR. If you have any feedback, just mention me in the comments below.

Configure Scala Steward for your repository with a [`.scala-steward.conf`](https://github.com/scala-steward-org/scala-steward/blob/17235ca1e60ff61e35fe7b7cf51163a15ffc435d/docs/repo-specific-configuration.md) file.

Have a fantastic day writing Scala!

<details>
<summary>Ignore future updates</summary>

Add this to your `.scala-steward.conf` file to ignore future updates of this dependency:
```
updates.ignore = [ { groupId = "io.grpc" } ]
```
</details>

labels: library-update, early-semver-minor, semver-spec-minor, commit-count:1

PR #1920, opened 2022-07-12
dont override google2 deps [AJ-490]

davidangb:
We have explicit dependencies for `io.grpc:grpc-core`, `io.grpc:grpc-netty-shaded`, `com.google.cloud:google-cloud-core-grpc`, and `com.google.auto.value:auto-value-annotations` configured as part of our `google2Dependencies `. Why do we need these listed explicitly? Everything seems to work without them, as workbench-libs already specifies what it needs. Perhaps an earlier version of workbench-libs didn't specify them correctly?

Looks like this was added in #1235 and #1259 - both by me!! So now let's pull them out.

The changes in this PR make other Scala Steward PRs irrelevant:
* obsoletes #1863
* supersedes #1900

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -54,12 +54,6 @@ object Dependencies {
   val googleDeploymentManager: ModuleID = "com.google.apis"   % "google-api-services-deploymentmanager" % ("v2beta-rev20210311-" + googleV)
   val googleGuava: ModuleID =             "com.google.guava"  % "guava" % "31.1-jre"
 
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -190,11 +184,7 @@ object Dependencies {
   // google2 lib requires specific versions of rpc libs:
```
calypsomatic: Presumably this comment is wrong then?  google2 lib no longer requires specific versions?

calypsomatic (approved):

aednichols (approved):

PR #1842, opened 2022-05-30
Update google-auth-library-oauth2-http from 0.9.1 to 0.27.0

broadbot:
Updates [com.google.auth:google-auth-library-oauth2-http](https://github.com/googleapis/google-auth-library-java) from 0.9.1 to 0.27.0.
[GitHub Release Notes](https://github.com/googleapis/google-auth-library-java/releases/tag/v0.27.0) - [Version Diff](https://github.com/googleapis/google-auth-library-java/compare/v0.9.1...v0.27.0)

I'll automatically update this PR to resolve conflicts as long as you don't change it yourself.

If you'd like to skip this version, you can just close this PR. If you have any feedback, just mention me in the comments below.

Configure Scala Steward for your repository with a [`.scala-steward.conf`](https://github.com/scala-steward-org/scala-steward/blob/17235ca1e60ff61e35fe7b7cf51163a15ffc435d/docs/repo-specific-configuration.md) file.

Have a fantastic day writing Scala!

<details>
<summary>Files still referring to the old version number</summary>

The following files still refer to the old version number (0.9.1).
You might want to review and update them manually.
```
project/plugins.sbt
```
</details>
<details>
<summary>Ignore future updates</summary>

Add this to your `.scala-steward.conf` file to ignore future updates of this dependency:
```
updates.ignore = [ { groupId = "com.google.auth", artifactId = "google-auth-library-oauth2-http" } ]
```
</details>

labels: library-update, early-semver-major, semver-spec-minor, old-version-remains, commit-count:1

PR #1921, opened 2022-07-12
remove google oauth dependencies [AJ-490]

davidangb:
* switch `JacksonFactory` to `GsonFactory`, since `JacksonFactory` [is deprecated](https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.json.jackson2.JacksonFactory).
* remove the unnecessary `google-api-services-oauth2` and `google-auth-library-oauth2-http` libraries. The only functionality these offered was pulling in `JacksonFactory`, which we no longer need. This supersedes #1842 .

NOT in this PR, considering for a future PR: align the direct `com.google.apis` dependencies in Rawls and the same libraries pulled in as transitive dependencies from workbench-libs' workbench-google.


davidangb:
jenkins retest

calypsomatic (approved):

PR #1923, opened 2022-07-13
use dependencyOverrides for overriding transitive deps [AJ-491]

davidangb:
For those transitive libraries that we need to override - such as when a static-analysis tool identifies a security vulnerability in a transitive dep - use sbt's `dependencyOverrides` ([doc](https://www.scala-sbt.org/1.x/docs/Library-Management.html#Overriding+a+version)) instead of adding as direct dependencies.

Hopefully this makes the code clearer, and in the corner case that the transitive dependency is no longer necessary, it prevents that dep from being included at all.


davidangb:
jenkins retest

calypsomatic (approved):

PR #1924, opened 2022-07-14
update data repo client [AJ-476]

davidangb:
We were woefully out of date on our `datarepo-client` dependency. This PR updates it to the latest version.

This update:
* requires a few small syntax compatibility changes
* allows us to remove the explicit `jersey-hk2` dependency, since that is now properly pulled in transitively from `datarepo-client`

Note that #1393, from a ways back, created a unit test to validate that the proper libraries are in place for data repo client connectivity.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/datarepo/HttpDataRepoDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/datarepo/HttpDataRepoDAO.scala
@@ -24,6 +24,7 @@ class HttpDataRepoDAO(dataRepoInstanceName: String, dataRepoInstanceBasePath: St
   override def getInstanceName: String = dataRepoInstanceName
 
   override def getSnapshot(snapshotId: UUID, accessToken: OAuth2BearerToken): SnapshotModel = {
-    getRepositoryApi(accessToken).retrieveSnapshot(snapshotId.toString)
+    // future enhancement: allow callers to specify the list of SnapshotRetrieveIncludeModel to retrieve
+    getRepositoryApi(accessToken).retrieveSnapshot(snapshotId, java.util.Collections.emptyList())
   }
 }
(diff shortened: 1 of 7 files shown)


project/Dependencies.scala
```diff
@@ -136,7 +135,6 @@ object Dependencies {
   // One reason to specify an override here is to avoid static-analysis security warnings.
   val transitiveDependencyOverrides = Seq(
     "commons-codec"                 % "commons-codec"         % "1.15",
-    "org.apache.httpcomponents"     % "httpclient"            % "4.5.13",
```
davidangb, line 139: this is a bonus fix unrelated to the `datarepo-client` update. We are already using `httpclient` version 4.5.13, so this override is extraneous. Removed.

project/Dependencies.scala
```diff
@@ -136,8 +135,7 @@ object Dependencies {
...
-    "org.glassfish.jersey.core"     % "jersey-client"         % "2.36"
+    "org.glassfish.jersey.core"     % "jersey-client"         % "2.36" // scala-steward:off (must match TDR)
```
davidangb, line 138: data repo client fails with jersey-client 3.04 - we need to stay on the 2.x branch here.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/datarepo/HttpDataRepoDAO.scala
```diff
@@ -24,6 +24,7 @@ class HttpDataRepoDAO(dataRepoInstanceName: String, dataRepoInstanceBasePath: St
   override def getInstanceName: String = dataRepoInstanceName
 
   override def getSnapshot(snapshotId: UUID, accessToken: OAuth2BearerToken): SnapshotModel = {
-    getRepositoryApi(accessToken).retrieveSnapshot(snapshotId.toString)
+    // future enhancement: allow callers to specify the list of SnapshotRetrieveIncludeModel to retrieve
+    getRepositoryApi(accessToken).retrieveSnapshot(snapshotId, java.util.Collections.emptyList())
```
davidangb, line 28: in the updated dependency, `retrieveSnapshot` now requires the second argument.

Additionally, the `snapshotId` changed from a `String` to a `UUID`, which drives syntax changes in other files below.

jordansafer (approved):

calypsomatic (approved):

PR #1939, opened 2022-07-21
clean up metrics dependencies [AJ-501]

davidangb:
Supersedes #1902. While looking at #1902, I questioned how the metrics libraries were being used, and saw that we could clean up a little bit.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -53,8 +53,12 @@ object Dependencies {
   val googleDeploymentManager: ModuleID = "com.google.apis"   % "google-api-services-deploymentmanager" % ("v2beta-rev20210311-" + googleV)
   val googleGuava: ModuleID =             "com.google.guava"  % "guava" % "31.1-jre"
 
+  // metrics4-scala and metrics3-statsd are pulled in by workbench-metrics, which is pulled in by
+  // workbench-google (workbenchGoogle variable in this file). Thus, anything that depends on workbench-google, such as
+  // rawlsCoreDependencies, does not need these. As of this writing, metrics4-scala and metrics3-statsd are only
+  // needed by the metrics subproject of Rawls.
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -53,8 +53,12 @@ object Dependencies {
   val googleDeploymentManager: ModuleID = "com.google.apis"   % "google-api-services-deploymentmanager" % ("v2beta-rev20210311-" + googleV)
   val googleGuava: ModuleID =             "com.google.guava"  % "guava" % "31.1-jre"
 
+  // metrics4-scala and metrics3-statsd are pulled in by workbench-metrics, which is pulled in by
+  // workbench-google (workbenchGoogle variable in this file). Thus, anything that depends on workbench-google, such as
+  // rawlsCoreDependencies, does not need these. As of this writing, metrics4-scala and metrics3-statsd are only
+  // needed by the metrics subproject of Rawls.
   // metrics-scala transitively pulls in io.dropwizard.metrics:metrics-core
-  val metricsScala: ModuleID =       "nl.grons"              %% "metrics4-scala"    % "4.1.19"
+  val metricsScala: ModuleID =       "nl.grons"              %% "metrics4-scala"    % "4.2.9"
```
aherbst-broad, line 61: If the above comment holds true should the declaration of `metricsDependencies` at L150 get dropped too? It appears unused.

aherbst-broad (approved):
I think this is fine since we're pulling in the libs transitively as you mention. Had a question about removing a dead dependency declaration.

project/Dependencies.scala
davidangb: Thank you for the review! `metricsDependencies` is still used here, fyi:
https://github.com/broadinstitute/rawls/blob/cd1125e28f6f5bec64ce6852911c7efff8cd42c0/project/Settings.scala#L118-L121

jordansafer (approved):

PR #1940, opened 2022-07-21
[AJ-395] Add a new API to Rawls for Retrying a submission with no config changes

vitzlo:
[Jira Ticket](https://broadworkbench.atlassian.net/browse/AJ-395?atlOrigin=eyJpIjoiOGNmMTYzY2Q4MTc3NGVmYmFmMTFmOGU3NjNiY2QyZmMiLCJwIjoiaiJ9)

[Design Doc](https://docs.google.com/document/d/12VO29hL1qViHOICicBtNhb_IcBnd6iEyRaQ9GLVMNlc/edit)

This ticket creates the API to retry the failed workflows via a new submission after a job is submitted via snapshot by reference.


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
```diff
@@ -44,6 +50,18 @@ trait SubmissionApiService extends UserInfoDirectives {
           }
         }
       } ~
+      path("workspaces" / Segment / Segment / "retrySubmission" / Segment) { (workspaceNamespace, workspaceName, submissionId) => // keeping all logic here and slowly moving out
```
davidangb: could this be `/api/workspaces/{workspaceNamespace}/{workspaceName}/submissions/{submissionId}/retry` instead?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
```diff
@@ -44,6 +50,18 @@ trait SubmissionApiService extends UserInfoDirectives {
...
+        post {
+          val submissionStatus = workspaceServiceConstructor(userInfo).getSubmissionStatus(WorkspaceName(workspaceNamespace, workspaceName), submissionId)
+          submissionStatus.onComplete {
```
davidangb: when/if this code moves into `WorkspaceService`, look at using the `withSubmission` helper method

davidangb (commented):
two details inline. Initial thoughts are that this looks good. Logic should move into `WorkspaceService` as we discussed, probably into a new method `retrySubmission` that is parallel to `createSubmission`.

Also make sure not to commit the `node_modules` or `yarn.lock`!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
vitzlo: I'm game for this change, I'll leave this up for discussion for a little while longer so everyone can give their two cents. I like this change given the other endpoints with existing similar paths and the prerequisite nature of the new endpoint coinciding with the nesting.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -25,6 +26,7 @@ case class SubmissionRequest(
   entityType: Option[String],
   entityName: Option[String],
   expression: Option[String],
+  retryType: RetryStatus,
```
jordansafer: I don't think we want to modify SubmissionRequest, since it's used in other cases where we don't have retry arguments. We'll need a separate request class

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -506,6 +508,32 @@ object WorkflowStatuses {
...
+        case RetryBoth => workflows.filter(wf => wf.status.equals(Failed) || wf.status.equals(Aborted))
+      }
+    }
+    override def toString = getClass.getSimpleName.stripSuffix("$")
+    override def withName(name: String) = SubmissionRetryStatuses.withName(name)
+  }
+
+  def withName(name: String): RetryStatus = {
+    name match {
+      case "Failed" => RetryFailed
```
jordansafer, line 550: minor nit that it will help with debugability if enum strings match the value names. For example `case "Failed" => Failed`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1569,6 +1570,34 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+      case Success(status) =>
+        val newSubmission = Submission(submissionId = submissionId.toString,
+          submissionDate = DateTime.now(),
+          submitter = WorkbenchEmail(userInfo.userEmail.value),
+          methodConfigurationNamespace = submissionRequest.methodConfigurationNamespace,
+          methodConfigurationName = submissionRequest.methodConfigurationName,
+          submissionEntity = submissionEntityOpt,
+          workflows = submissionRequest.retryType.filterWorkflow(status.workflows),
+          status = SubmissionStatuses.Submitted,
+          useCallCache = submissionRequest.useCallCache,
```
jordansafer: For the MVP, we want to take all these values from the previous submission, not from the request. As a future change, we want to support the user optionally overriding the previous submission values in the request.

So currently, we just want to use the values from previous submission.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
```diff
@@ -44,6 +46,13 @@ trait SubmissionApiService extends UserInfoDirectives {
           }
         }
       } ~
+      path("workspaces" / Segment / Segment / "submission" / Segment / "retry") { (workspaceNamespace, workspaceName, submissionId) =>
```
jordansafer: We'll need a swagger definition for the new API

jordansafer (changes requested):
This looks great! I left some comments, and generally I think this is the right direction.

I know we also talked about permissions, just making a note reminder to myself to keep an eye out for this-

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1268,6 +1268,45 @@ paths:
...
+        - $ref: '#/components/parameters/workspaceNamespacePathParam'
+        - $ref: '#/components/parameters/workspaceNamePathParam'
+        - $ref: '#/components/parameters/submissionIdPathParam'
+      responses:
+        201:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/SubmissionReport'
```
davidangb: this will need a new `RetriedSubmissionReport` model

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1582,6 +1582,42 @@ class WorkspaceService(protected val userInfo: UserInfo,
       }
     }
 
+  def retrySubmission(workspaceName: WorkspaceName, submissionRetry: SubmissionRetry, submissionId: String): Future[RetriedSubmissionReport] = {
+    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write) flatMap { workspaceContext =>
+      dataSource.inTransaction { dataAccess =>
+        withSubmission(workspaceContext, submissionId, dataAccess) { submission =>
+          val newSubmission = Submission(submissionId = submissionId,
```
davidangb: I think this is creating a new submission with the same UUID as the old submission. I'm surprised this isn't failing any tests!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1582,6 +1582,42 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+            submissionEntity = submission.submissionEntity,
+            workflows = submissionRetry.retryType.filterWorkflows(submission.workflows),
+            status = submission.status,
+            useCallCache = submission.useCallCache,
+            deleteIntermediateOutputFiles = submission.deleteIntermediateOutputFiles,
+            useReferenceDisks = submission.useReferenceDisks,
+            memoryRetryMultiplier = submission.memoryRetryMultiplier,
+            workflowFailureMode = submission.workflowFailureMode,
+            externalEntityInfo = submission.externalEntityInfo,
+            userComment = submission.userComment)
```
davidangb: you can simplify this block by using `copy` (note I've also changed the submissionId):
```suggestion
          val newSubmission = submission.copy(submissionId = UUID.randomUUID().toString,
            submissionDate = DateTime.now(),
            workflows = submissionRetry.retryType.filterWorkflows(submission.workflows))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1582,6 +1582,42 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+
+          implicit val subStatusCounter = submissionStatusCounter(workspaceMetricBuilder(workspaceContext.toWorkspaceName))
+          implicit val wfStatusCounter = (status: WorkflowStatus) =>
+            if (config.trackDetailedSubmissionMetrics) Option(workflowStatusCounter(workspaceSubmissionMetricBuilder(workspaceContext.toWorkspaceName, UUID.fromString(submissionId)))(status))
+            else None
+
+          for {
+            retriedSub <- dataAccess.submissionQuery.create(workspaceContext, submission)
```
davidangb: could you break out the `subStatusCounter`, `wfStatusCounter`, and `dataAccess.submissionQuery.create` into a shared method used by both `retrySubmission` and `saveSubmission`? At minimum, add a code comment to both places that if we change how we capture metrics, we need to change it in both places

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -509,6 +528,32 @@ object WorkflowStatuses {
   case object Unknown extends WorkflowStatus
 }
 
+object SubmissionRetryStatuses {
+  sealed trait RetryStatus extends RawlsEnumeration[RetryStatus] {
+    def filterWorkflows(workflows: Seq[Workflow]) = {
+      this match {
+        case RetryFailed => workflows.filter(wf => wf.status.equals(Failed))
+        case RetryAborted => workflows.filter(wf => wf.status.equals(Aborted))
+        case RetryBoth => workflows.filter(wf => wf.status.equals(Failed) || wf.status.equals(Aborted))
```
davidangb: nit: change this to `RetryFailedAndAborted` or similar - "both" is not clear without knowing that the other options are failed and aborted

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -229,6 +230,14 @@ case class SubmissionReport(
   workflows: Seq[SubmissionValidationEntityInputs]
 )
 
+case class RetriedSubmissionReport(
+  originalSubmissionId: String,
+  submissionId: String,
+  submissionDate: DateTime,
+  submitter: String,
+  status: SubmissionStatus,
+)
```
davidangb, line 245: I suggest adding to this:
* the retry type (failed, aborted, failed+aborted)
* list of workflows, similar to `SubmissionReport.workflows`

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -424,8 +435,12 @@ trait ExecutionJsonSupport extends JsonSupport {
 
   implicit val SubmissionFormat = jsonFormat17(Submission)
 
+  implicit val SubmissionRetryFormation = jsonFormat1(SubmissionRetry)
```
davidangb: nit: `Format`, not `Formation`

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1268,6 +1268,45 @@ paths:
...
+    post:
+      tags:
+        - submissions
+      summary: Retry certain workflows in previous submission
+      description: Retry aborted or failed workflows from the previous submission and omit the remaining workflows
+      operationId: retrySubmission
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespacePathParam'
+        - $ref: '#/components/parameters/workspaceNamePathParam'
+        - $ref: '#/components/parameters/submissionIdPathParam'
```
davidangb, line 1281: this should accept a `SubmissionRetry` in its request body

davidangb (commented):
initial comments inline. Could you add some unit tests for this?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
```diff
@@ -15,6 +16,7 @@ import spray.json.{JsString, PrettyPrinter}
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success}
```
davidangb: nit: both of the import changes - `RawlsExceptionWithErrorReport`, `Failure`, `Success` - are unused

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1582,6 +1582,33 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+          val newSubmissionRoot = s"gs://${workspaceContext.bucketName}/submissions/${newSubmissionId}"
+          val filterWorkFlows = submissionRetry.retryType.filterWorkflows(submission.workflows)
+          val filteredAndResetWorkflows = filterWorkFlows.map(wf => wf.copy(workflowId = None,
+            status = WorkflowStatuses.Queued,
+            statusLastChangedDate = DateTime.now))
+          val newSubmission = submission.copy(submissionId = newSubmissionId.toString,
+            submissionDate = DateTime.now(),
+            submissionRoot = newSubmissionRoot,
+            workflows = filteredAndResetWorkflows,
+            status = SubmissionStatuses.Submitted)
```
davidangb: also override the `submitter` to be the person submitting the retry; I think you have easy access to the current user in `userInfo.userEmail`

davidangb: I think it would be really valuable to store the original submissionId somewhere in the new submission. Without doing that, there's no way to connect the retry to the original except from the `RetriedSubmissionReport` … if another user of the workspace goes to the Job History tab and looks at the retry, that user can't see how the two are connected.

Perhaps the ideal way to do that would be to add a new field to the `Submission` class - this would require a db schema change and carry a whole lot of code changes for compatibility. In the short term, how about overriding the `userComment` to be something like `s"retry of submission $submissionId with retry type ${submissionRetry.retryType}"`? Another potential option is to set the `Submission`'s `submissionEntity` value to the old submissionId - but that would also require a lot of cascading code changes. What do you think?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1582,6 +1582,33 @@ class WorkspaceService(protected val userInfo: UserInfo,
       }
     }
 
+  def retrySubmission(workspaceName: WorkspaceName, submissionRetry: SubmissionRetry, submissionId: String): Future[RetriedSubmissionReport] = {
+    getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.write) flatMap { workspaceContext =>
+      dataSource.inTransaction { dataAccess =>
+        withSubmission(workspaceContext, submissionId, dataAccess) { submission =>
+          val newSubmissionId = UUID.randomUUID()
+          val newSubmissionRoot = s"gs://${workspaceContext.bucketName}/submissions/${newSubmissionId}"
+          val filterWorkFlows = submissionRetry.retryType.filterWorkflows(submission.workflows)
```
davidangb, line 2157: if this results in 0 workflows, I think this should short-circuit and not even create the new submission, something like:
```scala
if (filterWorkFlows.isEmpty) {
    throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "no workflows to retry"))
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1714,16 +1741,20 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
   }
 
+  def logAndCreateDbSubmission(workspaceContext: Workspace, submissionId: UUID, submission: Submission, dataAccess: DataAccess): ReadWriteAction[Submission] = {
+    // implicitly passed to SubmissionComponent.create
+    implicit val subStatusCounter = submissionStatusCounter(workspaceMetricBuilder(workspaceContext.toWorkspaceName))
+    implicit val wfStatusCounter = (status: WorkflowStatus) =>
+      if (config.trackDetailedSubmissionMetrics) Option(workflowStatusCounter(workspaceSubmissionMetricBuilder(workspaceContext.toWorkspaceName, submissionId))(status))
+      else None
+
+    dataAccess.submissionQuery.create(workspaceContext, submission)
```
davidangb, line 2417: This will throw an ugly and nigh-indecipherable stack trace if the method configuration associated with the old submission no longer exists.

I believe the offending code is the `configId.get` at 
https://github.com/broadinstitute/rawls/blob/40d4931dec2f82e6dba744ad200969b63a7bb7c0/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala#L197, because the `methodConfigurationQuery.findActiveByName` found zero results.

Unfortunately I think this is going to be a fairly common use case.
Screenshot 8-22-2022 at 03 57 PM

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -1258,6 +1259,13 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system)
     }
   }
 
+  "Retrysubmission" should "succeed" in withSubmissionTestWorkspaceService { workspaceService =>
```
davidangb: nit: "Retrysubmission" -> "Retry submission"

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionSpec.scala
```diff
@@ -1258,6 +1259,13 @@ class SubmissionSpec(_system: ActorSystem) extends TestKit(_system)
...
+    val req = workspaceService.retrySubmission(subTestData.wsName, SubmissionRetry(RetryAborted), subTestData.submissionTestAbortTwoGoodWorkflows.submissionId)
+    val report = Await.result(req, Duration.Inf)
+    assertResult(subTestData.submissionTestAbortTwoGoodWorkflows.submissionId, "Retried submission should reference original") { report.originalSubmissionId }
+    assert(report.submissionId != report.originalSubmissionId, "We should generate a new submission id")
```
davidangb, line 1563: could you add something like
```
report.workflows should have size 2
```
to assert that the logic is correct and it's actually finding two workflows to retry?

Also assuming you make changes to override the submitter and userComment, validate those changes here too

davidangb (commented):
see inline, this is very close!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
sagehen03: I would vote for leaving this as is.  If this becomes a widely used feature and users ask for this linkage that you're suggesting, then let's implement it then.  If you would really like to include it, then I'm happy to turn this ticket over to you.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
sagehen03: I'm not sure if you're asking for changes, but if you are..I don't know this functionality well, but I think the config.get issue is not a problem for regular (not retried) submissions because if the method config/workflow no longer exists then the UI doesn't let you create submission because you need to go through a workflow?  If all that is correct, then I'd say let's do something similar in the UI and hide the re-try failure/aborted buttons if the workflow doesn't exist.  Would that work?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SubmissionApiService.scala
davidangb: nit: `Failure` and `Success` are still unused, remove them?

davidangb (commented):
one minor unaddressed nit inline on unused imports.

On the topic of missing method configs: I agree that we should handle this primarily at the UX level - but since so many of the users asking for retries rely on scripts and automation, I also think we should improve the error messaging for this use case. #2011 should do it; if you don't want to include it in this PR, I can own merging that enhancement after this one goes in.

On linking the new submission to the one it retries - I do personally think it's necessary. That said, we should get product input and it can happen in a follow-on PR, and I won't block this one on it; we just may need to keep this feature a little quiet until we have that, assuming product deems it necessary.

sagehen03:
> one minor unaddressed nit inline on unused imports.
> 
> On the topic of missing method configs: I agree that we should handle this primarily at the UX level - but since so many of the users asking for retries rely on scripts and automation, I also think we should improve the error messaging for this use case. #2011 should do it; if you don't want to include it in this PR, I can own merging that enhancement after this one goes in.
> 
> On linking the new submission to the one it retries - I do personally think it's necessary. That said, we should get product input and it can happen in a follow-on PR, and I won't block this one on it; we just may need to keep this feature a little quiet until we have that, assuming product deems it necessary.

Ok, I fixed the unused import and changed my IntelliJ settings so as to hopefully catch that sort of thing in the future.  One thought I had on linking the original and the retry is we could prepend or append the original submission id in the user comment.  It's kind of ugly but would give the user some visibility I think and would take a less than an hour to change and add a test for.

jordansafer (approved):
LGTM- Can you make another ticket for adding parent submission id as a root entity type?

davidangb (approved):
:+1:, assuming a follow-on PR to update the user comment

davidangb (approved):
:+1: assuming tests pass

PR #1946, opened 2022-07-26
sam client lib and cross service tracing - a taste

dvoet:
I am looking for feedback before I embark on a sizable refactoring. The main goal is to enable cross service tracing between rawls and sam. An intermediate step is to replace rawls' hand rolled sam client implementation with the generated client.

The first choice is do we want to replace rawls' sam client. This would also include stopping use of resource/v1 api in sam which is good. A follow on choice is do we want to use the generated sam model classes in rawls or wrap them in the existing hand rolled case classes which are better types and use scala collections.

Another major change is to pass a `RawlsRequestContext` object from the routes to internal service instances (e.g. `UserService`) instead of `UserInfo`. `RawlsRequestContext` contains `UserInfo` and an optional tracing span. Eventually I hope we work this into our logging calls as well.

Most of the changes in the routes are auto formatting - suggest ignoring whitespace.

~There are 2 java files included in the PR which were copied from terra common lib because I could not get the dependency to work. Hopefully we can figure that our eventually and not copy the code. These are the smarts that know how to propagate trace information in our generated client libs.~

Here is an example trace (ignore the actual timings and alignment, this was made from a local rawls to a dev sam so it takes longer and there is probably clock skew)

image
(the last 3 spans are in sam)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -210,6 +210,23 @@ paths:
...
+        200:
+          description: OK
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/RawlsBillingProjectResponse'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
```
dvoet, line 229: this was just missing

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -295,7 +295,7 @@ object Boot extends IOApp with LazyLogging {
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
       val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(samDAO, multiCloudWorkspaceConfig)
 
-      val userServiceConstructor: (UserInfo) => UserService =
+      val userServiceConstructor: (RawlsRequestContext) => UserService =
```
dvoet, line 298: switch from `UserInfo` to `RawlsRequestContext`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -41,6 +49,27 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
...
   protected val statusUrl = samServiceURL + "/status"
 
+  private val okHttpClient = new ApiClient().getHttpClient
+
+  protected def getApiClient(ctx: RawlsRequestContext): ApiClient = {
+
+    val okHttpClientWithTracingBuilder = okHttpClient.newBuilder
+    ctx.traceSpan.foreach(span => okHttpClientWithTracingBuilder.
+      addInterceptor(new SpanSettingInterceptor(span)).
+      addInterceptor(new OkHttpClientTracingInterceptor(Tracing.getTracer)))
```
dvoet, line 59: sets up the tracing

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -75,11 +104,47 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 
-  override def getPolicySyncStatus(resourceTypeName: SamResourceTypeName, resourceId: String, policyName: SamResourcePolicyName, userInfo: UserInfo): Future[SamPolicySyncStatus] = {
-    val url = samServiceURL + s"/api/google/v1/resource/${resourceTypeName.value}/$resourceId/$policyName/sync"
+  private class SamApiCallback[T](functionName: String) extends ApiCallback[T] {
+    private val promise = Promise[T]()
+
+    override def onFailure(e: ApiException, statusCode: Int, responseHeaders: util.Map[String, util.List[String]]): Unit = {
```
dvoet, line 110: this was copied from `doSuccessOrFailureRequest` on line 77

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -75,11 +104,47 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
...
+
+    def future: Future[T] = promise.future
+  }
+
+  override def getPolicySyncStatus(resourceTypeName: SamResourceTypeName, resourceId: String, policyName: SamResourcePolicyName, ctx: RawlsRequestContext): Future[SyncStatus] = {
     retry(when401or5xx) { () =>
-      pipeline[SamPolicySyncStatus](userInfo) apply RequestBuilding.Get(url)
+      val callback = new SamApiCallback[SyncStatus]("syncStatusAsync")
+      googleApi(ctx).syncStatusAsync(resourceTypeName.value, resourceId, policyName.value, callback)
+      callback.future
```
dvoet, line 147: This is what an async call using the sam client lib looks like. We could also do this using a `Future` and a synchronous call but then we would probably want a different execution context, etc. The sam api client has threading stuff built in so it feels better to use that.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -310,3 +378,11 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 }
+
+class SpanSettingInterceptor(span: Span) extends Interceptor {
+  override def intercept(chain: Interceptor.Chain): Response = {
+    Using(Tracing.getTracer.withSpan(span)) { _ =>
+      chain.proceed(chain.request())
```
dvoet, line 385: The tracing library we use was built for java and passed the trace information through thread local variables. That does not work in scala using `Futures` or other delayed execution since we are switching threads all the time. So this interceptor contains the tracing span to propagate and calls `withSpan` when we have gotten to the right thread.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -175,28 +178,28 @@ class UserService(protected val userInfo: UserInfo,
...
 
-  private def constructBillingProjectResponses(samUserResources: Seq[SamUserResource], billingProjectsInRawlsDB: Seq[RawlsBillingProject]): List[RawlsBillingProjectResponse] = {
+  private def constructBillingProjectResponses(samUserResources: Seq[UserResourcesResponse], billingProjectsInRawlsDB: Seq[RawlsBillingProject]): List[RawlsBillingProjectResponse] = {
     val projectsByName = billingProjectsInRawlsDB.map(p => p.projectName.value -> p).toMap
     samUserResources.toList
       .flatMap { samUserResource =>
-        val projectRoles = samRolesToProjectRoles(samUserResource.direct.roles ++ samUserResource.inherited.roles)
-        projectsByName.get(samUserResource.resourceId).map(makeBillingProjectResponse(projectRoles, _))
+        val projectRoles = samRolesToProjectRoles(samUserResource.getDirect.getRoles.asScala.toSet ++ samUserResource.getInherited.getRoles.asScala.toSet)
+        projectsByName.get(samUserResource.getResourceId).map(makeBillingProjectResponse(projectRoles, _))
```
dvoet, line 195: here is an example where switching from the hand-rolled to the generated sam model classes is a little uncomfortable. These were really generated for Java so we have to use getXXX and convert any java collections. WDYT? The alternative is to build a mapping between the generated classes and the hand rolled classes. That's boiler plate stuff but it will also shield the rest of the code from changes like this.

drewherbst: I think we should use the sam model classes. We have 2 other examples of using similarly auto-generated model classes in Rawls (TDR + workspace manager, and soon, billing profile manager). Unless we plan on adapting those as well, we should push for consistency and use the client generated model classes for all of these services.

drewherbst:
Thanks for doing this, I am 💯 on replacing the hand rolled SAM client w/the real client lib. Left a comment inline about the model classes.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -210,6 +210,23 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
   /api/billing/v2:
+    get:
+      tags:
+        - billing_v2
+      summary: list billing project in rawls
+      description: list new billing projects in rawls
+      operationId: lsitBillingProjectsV2
```
rtitle, line 218: typo, `listBillingProjectsV2`

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -75,11 +104,47 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 
-  override def getPolicySyncStatus(resourceTypeName: SamResourceTypeName, resourceId: String, policyName: SamResourcePolicyName, userInfo: UserInfo): Future[SamPolicySyncStatus] = {
-    val url = samServiceURL + s"/api/google/v1/resource/${resourceTypeName.value}/$resourceId/$policyName/sync"
+  private class SamApiCallback[T](functionName: String) extends ApiCallback[T] {
+    private val promise = Promise[T]()
+
+    override def onFailure(e: ApiException, statusCode: Int, responseHeaders: util.Map[String, util.List[String]]): Unit = {
+      val response = e.getResponseBody
```
rtitle, line 111: Why is L111-130 needed? We're constructing a `Future[Try[...]]` from the response body but nothing is using the result… we're just completing the promise with the original `ApiException`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -75,11 +104,47 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 
-  override def getPolicySyncStatus(resourceTypeName: SamResourceTypeName, resourceId: String, policyName: SamResourcePolicyName, userInfo: UserInfo): Future[SamPolicySyncStatus] = {
-    val url = samServiceURL + s"/api/google/v1/resource/${resourceTypeName.value}/$resourceId/$policyName/sync"
+  private class SamApiCallback[T](functionName: String) extends ApiCallback[T] {
+    private val promise = Promise[T]()
```
rtitle, line 108: FWIW cats-effect `IO.async` has similar functionality to `Promise` where you can complete it from a callback… but Scala Promise/Future is fine here I think if that's the pattern in Rawls.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
rtitle: +1 just using the generated models for simplicity sake (at the cost of some extra Java conversions).

rtitle:
Big 👍 on using the Sam client in Rawls and setting up distributed tracing. Some nits above :)

project/Dependencies.scala
```diff
@@ -122,6 +122,8 @@ object Dependencies {
   val workspaceManager = excludeJakarta("bio.terra" % "workspace-manager-client" % "0.254.311-SNAPSHOT")
   val dataRepo = excludeJakarta("bio.terra" % "datarepo-client" % "1.379.0-SNAPSHOT")
   val resourceBufferService = excludeJakarta("bio.terra" % "terra-resource-buffer-client" % "0.4.3-SNAPSHOT")
+  val sam: ModuleID = excludeJakarta("org.broadinstitute.dsde.workbench" %% "sam-client" % "0.1-1607de4")
+  val terraCommonLib: ModuleID = ("bio.terra" % "terra-common-lib" % "0.0.62-SNAPSHOT" classifier "plain").exclude("org.broadinstitute.dsde.workbench", "sam-client_2.12")
```
rtitle, line 126: I also had to exclude `sam-client_2.12` from TCL in WSM. We should upgrade TCL to use the latest sam-client.

I made https://broadworkbench.atlassian.net/browse/TOAZ-181

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
dvoet: I don't want to bite off switching this from Future to IO

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
dvoet: that's an oversight, it should be completing the promise with a `RawlsExceptionWithErrorReport` so that the retry logic works the same

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -75,11 +104,47 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
...
+          toFutureTry(Unmarshal(response).to[String]) map { maybeString =>
+            val stringErrMsg = maybeString match {
+              case Success(stringErr) => stringErr
+              case Failure(_) => response
+            }
+            throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCode.int2StatusCode(statusCode), s"Sam call to $functionName failed with error '$stringErrMsg'"))
+          }
+      }
 
+      val x = Unmarshal(e.getResponseBody).to[ErrorReport]
```
davidangb, line 130: is this `val x` line unused?

aherbst-broad:

PR #1949, opened 2022-07-27
[SUP-708] Support firecloudHasAccess as optional query parameter to list billingAccounts endpoint

cahrens:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3337,6 +3337,15 @@ paths:
       summary: list billing accounts for a user
       description: list billing accounts for a user
       operationId: listUserBillingAccounts
+      parameters:
```
cahrens, line 3340: Do I need to update orchestration for this too? Or can we just wait and see if anyone ever requests the ability to pass this parameter through orchestration?

https://github.com/broadinstitute/firecloud-orchestration/blob/219bdb1a7c4c85df5350f27be9292bb66233a355/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala#L105

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -151,7 +151,6 @@ object Boot extends IOApp with LazyLogging {
 
     initAppDependencies[IO](conf, appName, metricsPrefix).use { appDependencies =>
       val gcsDAO = new HttpGoogleServicesDAO(
-        false,
```
cahrens, line 154: All of these variables were flagged as unused in IntelliJ.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -652,7 +646,11 @@ class HttpGoogleServicesDAO(
       }
 
       res <- allProcessedChunks.map(_.flatten).unsafeToFuture()
-    } yield res
+    } yield {
+        res collect {
```
cahrens: I spent a long time trying to figure out how to do this filtering above near line 643, where the `RawlsBillingAccount` is actually created. I was stumped (the Future seemed to complicate things), but would love to hear from Scala experts if there is a better way to do this.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -652,7 +646,11 @@ class HttpGoogleServicesDAO(
...
+          case account if (firecloudHasAccess.isEmpty || account.firecloudHasAccess == firecloudHasAccess.getOrElse(true)) => account
```
cahrens: Default value in `firecloudHasAccess.getOrElse(true)` won't be used because of the earlier `isEmpty` check. Is there a clearer way to do the comparison?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleServicesDAO.scala
```diff
@@ -14,27 +14,21 @@ import org.broadinstitute.dsde.rawls.model._
 import scala.collection.mutable
 import scala.concurrent._
 
-class MockBillingHttpGoogleServicesDAO( useServiceAccountForBuckets: Boolean,
+class MockHttpGoogleServicesDAO(
```
cahrens, line 17: This mock was previously unused.

core/src/main/resources/swagger/api-docs.yaml
marctalbott: It's generally good practice to update orchestration when updating Rawls to keep them as in-sync as possible. Since it's a passthrough in Orch, the change should just be to Orch's swagger page to reflect the downstream Rawls change

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
blakery: `} yield res.filter(account => firecloudHasAccess.map(access => access == account.firecloudHasAccess).getOrElse(true))`

The filter will give you all elements of the list where the passed function evaluates to true 

This should also answer your question below.

`.map(f)` on an optional will give you Some(the result of calling f with the value) if there's a value, or give you None if the original optional is none.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleServicesDAO.scala
```diff
@@ -64,10 +53,13 @@ class MockBillingHttpGoogleServicesDAO( useServiceAccountForBuckets: Boolean,
     accessContextManagerDAO = new MockGoogleAccessContextManagerDAO,
     resourceBufferJsonFile = resourceBufferJsonFile)(system, materializer, executionContext, timer) {
 
-  private var token: String = null
+  private val token: String = null
```
blakery: This could easily be removed, and the one place it's used replaced with a null literal (or don't even set it).

It's a private val, so it can never be overridden or changed.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleServicesDAO.scala
blakery, line 17: Personally, I'm suspicious of these sorts of project-wide 'mocks'.
I generally prefer to create a mock where I need it, and specify the methods I want behavior for.
eg: 
`val m = mock[ClassToBeMocked]`

and if I wanted the mock to  return a  specific value on a method, it would just be:
doReturn(the value).when(m.method(argument matchers))

It's lightweight, flexible, and the behavior that's being mocked is clear where it's being used.

core/src/main/resources/swagger/api-docs.yaml
blakery: It might be less overhead to do it now, since there's already a ticket open, and everything is fresh in mind.

blakery (approved):

core/src/main/resources/swagger/api-docs.yaml
cahrens: I'm just not sure if there is any need to expose it through orchestration, and I'd like to avoid the extra work of making changes in another repository unless there is a good reason to do so. The API has been this way for a long time with no complaints, and I don't even know if anyone is actually using the orchestration endpoint.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cahrens: Apparently this can be further simplified to `res.filter(account => firecloudHasAccess.forall(access => access == account.firecloudHasAccess))`. The `forAll` returns true if the Option is empty.

(I personally find the `collect` the easiest to read/understand, but thankfully I have unit tests that explain the desired behavior)

marctalbott: I believe you could use a collect up where the `RawlsBillingAccount` is actually created -- something like this might work

```
testDMBillingAccountAccess(acctName) collect {
            case firecloudHasAccount if (firecloudHasAccess.isEmpty || firecloudHasAccount == firecloudHasAccess.getOrElse(true)) => RawlsBillingAccount(acctName, firecloudHasAccount, acct.getDisplayName)
          }
```

marctalbott (approved):
Looks good, left another alternative for the filtering issue you ran into. The existing testing around this is messy, but it looks like you've made the best of the situation

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleServicesDAO.scala
cahrens: @blakery and I fought the good fight on trying to create a mock in the test itself instead of instantiating MockBillingHttpGoogleServicesDAO. We failed because of the way HttpGoogleServicesDAO is written, so I am going to stick with this method of mocking for the test.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cahrens: @marctalbott I tried that, but the Future complicates things.

```
Future.collect partial function is not defined at: false
scala.concurrent.Future$$anon$1: Future.collect partial function is not defined at: false

core/src/main/resources/swagger/api-docs.yaml
cahrens: Created https://broadworkbench.atlassian.net/browse/SUP-711 for this.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
cahrens: I pushed up the syntax of using `forall`.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/MockHttpGoogleServicesDAO.scala
cahrens: I removed it.

cahrens:
jenkins retest

PR #1953, opened 2022-07-29
[WOR-431] Refactor V2 billing project creation code

aherbst-broad:
## Context
We need to be able to create azure billing profiles in billing profile manager as part of azure billing project creation. This PR is the first step in the refactoring necessary to introduce that functionality. This step is focused solely on refactoring of existing functionality with an eye toward review-ability and more easily digestible PRs; it does **not** introduce any new azure/bpm related code.

## This PR
* Extracts the V2 billing project creation code path to a separate class for improved testability. The current coverage is achieved via heavyweight integration tests on the V2 billing endpoints. I've dubbed this class `BillingProjectOrchestrator`.
* Introduces a billing project repository, again for improved testability. The intention of this class is to isolate all slick DB calls for billing project operations and allow the orchestration layer to test it's business logic without the somewhat elaborate mocking needed for slick data sources or spinning up a real DB connection.

## Next Steps
* Introduce the Azure managed application billing profile creation code path. This will be implemented as a follow-on PR, currently [up in draft](https://github.com/broadinstitute/rawls/pull/1961)
------------
- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [ ] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
[...]

PR #1961, opened 2022-08-01
[WOR-431] Optionally create billing profile records and link for azure billing projects

aherbst-broad:
## Context
(This is the follow up to #1953) 
We need to be able to create azure billing profiles in billing profile manager as part of azure billing project creation. This PR is the second step in the refactoring necessary to introduce that functionality. This step introduces the code path necessary to 
determine if an incoming billing project creation request is eligible for linkage with a new bililng profile in billing profile manager (i.e, azure billing project creation requests)

## This PR
* Introduces a `BillingProjectCreator` trait, of which there are two implementations;
   * `BpmBillingProjectCreator`, which will handle the validation and creation of azure billing projects 
   *  `GoogleBillingProjectCreator` which will handle the current GCP codepath
* Updates the `BillingProfileManagerDAO` class with the needed client methods to validate and create azure billing profiles

## TODO
- [x] This PR is dependent on changes in [this upstream BPM PR](https://github.com/DataBiosphere/terra-billing-profile-manager/pull/15), that will need to merge first. 
- [x] Add tracing + profiling. There is a forthcoming pr that is going to modify much about how that works, so I am [deferring](https://broadworkbench.atlassian.net/browse/WOR-457) implementing that on this code until it merges 

-----------------
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

[WOR-431]: https://broadworkbench.atlassian.net/browse/WOR-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -214,7 +214,7 @@ paths:
       tags:
         - billing_v2
       summary: create billing project in rawls
-      description: create new billing project in rawls
+      description: create new billing project in rawls; the client must provide either Azure managed app coordinates or a Google billing account ID. Providing both will result in an error.
       operationId: createBillingProjectFullV2
       requestBody:
         description: create project request
@@ -5338,7 +5338,7 @@ components:
           description: the name of the project to create
         billingAccount:
           type: string
-          description: the id of the billing account to use, must start with 'billingAccounts/'
+          description: the id of the Google billing account to use, must start with 'billingAccounts/'
         servicePerimeter:
           type: string
           description: The fully qualified name of the GCP service perimeter to put
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3337,6 +3337,15 @@ paths:
       summary: list billing accounts for a user
       description: list billing accounts for a user
       operationId: listUserBillingAccounts
+      parameters:
+        - name: firecloudHasAccess
+          in: query
+          description: |
+            Whether the Firecloud service has been given access to the billing accounts. Optional.
+          required: false
+          example: true
+          schema:
+            type: boolean
       responses:
         200:
           description: Successful Request
@@ -5699,8 +5708,7 @@ components:
       description: List of workspace names to migrate
       type: array
       items:
-        type: string
-        pattern: "^{namespace}/{name}$"
+        $ref: '#/components/schemas/WorkspaceName'
   parameters:
     billingProjectIdPathParam:
       name: projectId
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5532,10 +5532,13 @@ components:
       properties:
         tenantId:
           type: string
+          description: UUID of the Azure tenant for this managed application deployment
         subscriptionId:
           type: string
+          description: UUID of the Azure subscription for this managed application deployment.
         managedResourceGroupId:
           type: string
+          description: ID of the Azure managed resource group for this managed application deployment.
     UpdateRawlsBillingAccountRequest:
       required:
         - billingAccount
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5330,7 +5330,6 @@ components:
       description: ""
     CreateRawlsV2BillingProjectFullRequest:
       required:
-        - billingAccount
         - projectName
       type: object
       properties:
@@ -5346,6 +5345,8 @@ components:
             this project into in the form accessPolicies/[POLICY NUMBER]/servicePerimeters/[NAME].
             Caller must have the add_project action for this service perimeter in
             Sam.
+        managedAppCoordinates:
+          $ref: '#/components/schemas/AzureManagedAppCoordinates'
       description: ""
     SpendReport:
       required:
@@ -5540,7 +5541,6 @@ components:
           description: UUID of the Azure subscription for this managed application deployment.
         managedResourceGroupId:
           type: string
-          format: uuid
           description: ID of the Azure managed resource group for this managed application deployment.
     UpdateRawlsBillingAccountRequest:
       required:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5532,12 +5532,15 @@ components:
       properties:
         tenantId:
           type: string
+          format: uuid
           description: UUID of the Azure tenant for this managed application deployment
         subscriptionId:
           type: string
+          format: uuid
           description: UUID of the Azure subscription for this managed application deployment.
         managedResourceGroupId:
           type: string
+          format: uuid
           description: ID of the Azure managed resource group for this managed application deployment.
     UpdateRawlsBillingAccountRequest:
       required:
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -289,7 +289,7 @@ object Boot extends IOApp with LazyLogging {
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
       val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
         samDAO,
-        new HttpBillingProfileManagerClientProvider(conf.getString("billingProfileManager.baseUrl")),
+        new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl")),
         multiCloudWorkspaceConfig
       )
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -14,7 +14,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import io.sentry.{Hint, Sentry, SentryEvent, SentryOptions}
 import net.ceedubs.ficus.Ficus._
-import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, BillingProjectOrchestrator, BillingRepository, HttpBillingProfileManagerClientProvider}
+import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, BillingProjectOrchestrator, BillingRepository, BpmBillingProjectCreator, GoogleBillingProjectCreator, HttpBillingProfileManagerClientProvider}
 import org.broadinstitute.dsde.rawls.config._
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.HttpDataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.martha.MarthaResolver
@@ -459,8 +459,14 @@ object Boot extends IOApp with LazyLogging {
         spendReportingServiceConfig
       )
 
+      val billingRepository = new BillingRepository(slickDataSource)
       val billingProjectOrchestratorConstructor: (UserInfo) => BillingProjectOrchestrator =
-        BillingProjectOrchestrator.constructor(samDAO, gcsDAO, new BillingRepository(slickDataSource), billingProfileManagerDAO)
+        BillingProjectOrchestrator.constructor(
+          samDAO,
+          billingRepository,
+          new GoogleBillingProjectCreator(samDAO, gcsDAO),
+          new BpmBillingProjectCreator(billingRepository , billingProfileManagerDAO)
+        )
 
       val service = new RawlsApiServiceImpl(
         multiCloudWorkspaceServiceConstructor,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -14,7 +14,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import io.sentry.{Hint, Sentry, SentryEvent, SentryOptions}
 import net.ceedubs.ficus.Ficus._
-import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, HttpBillingProfileManagerClientProvider}
+import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, BillingProjectOrchestrator, BillingRepository, HttpBillingProfileManagerClientProvider}
 import org.broadinstitute.dsde.rawls.config._
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.HttpDataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.martha.MarthaResolver
@@ -459,6 +459,9 @@ object Boot extends IOApp with LazyLogging {
         spendReportingServiceConfig
       )
 
+      val billingProjectOrchestratorConstructor: (UserInfo) => BillingProjectOrchestrator =
+        BillingProjectOrchestrator.constructor(samDAO, gcsDAO, new BillingRepository(slickDataSource), billingProfileManagerDAO)
+
       val service = new RawlsApiServiceImpl(
         multiCloudWorkspaceServiceConstructor,
         workspaceServiceConstructor,
@@ -467,6 +470,7 @@ object Boot extends IOApp with LazyLogging {
         genomicsServiceConstructor,
         snapshotServiceConstructor,
         spendReportingServiceConstructor,
+        billingProjectOrchestratorConstructor,
         statusServiceConstructor,
         shardedExecutionServiceCluster,
         ApplicationVersion(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -312,6 +312,7 @@ object Boot extends IOApp with LazyLogging {
       val submissionCostService: SubmissionCostService =
         SubmissionCostService.constructor(
           gcsConfig.getString("billingExportTableName"),
+          gcsConfig.getString("billingExportDatePartitionColumn"),
           gcsConfig.getString("serviceProject"),
           gcsConfig.getInt("billingSearchWindowDays"),
           bigQueryDAO
@@ -449,6 +450,7 @@ object Boot extends IOApp with LazyLogging {
       val spendReportingBigQueryService = appDependencies.bigQueryServiceFactory.getServiceFromJson(gcsConfig.getString("bigQueryJson"), GoogleProject(gcsConfig.getString("serviceProject")))
       val spendReportingServiceConfig = SpendReportingServiceConfig(
         gcsConfig.getString("billingExportTableName"),
+        gcsConfig.getString("billingExportTimePartitionColumn"),
         gcsConfig.getConfig("spendReporting").getInt("maxDateRange")
       )
 
@@ -465,8 +467,7 @@ object Boot extends IOApp with LazyLogging {
           samDAO,
           billingRepository,
           new GoogleBillingProjectCreator(samDAO, gcsDAO),
-          new BpmBillingProjectCreator(billingRepository , billingProfileManagerDAO)
-        )
+          new BpmBillingProjectCreator(billingRepository, billingProfileManagerDAO))
 
       val service = new RawlsApiServiceImpl(
         multiCloudWorkspaceServiceConstructor,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -14,7 +14,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import io.sentry.{Hint, Sentry, SentryEvent, SentryOptions}
 import net.ceedubs.ficus.Ficus._
-import org.broadinstitute.dsde.rawls.billing.BillingProfileManagerDAOImpl
+import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, HttpBillingProfileManagerClientProvider}
 import org.broadinstitute.dsde.rawls.config._
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.HttpDataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.martha.MarthaResolver
@@ -287,7 +287,11 @@ object Boot extends IOApp with LazyLogging {
       val servicePerimeterService = new ServicePerimeterService(slickDataSource, gcsDAO, servicePerimeterConfig)
 
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
-      val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(samDAO, multiCloudWorkspaceConfig)
+      val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
+        samDAO,
+        new HttpBillingProfileManagerClientProvider(conf.getString("billingProfileManager.baseUrl")),
+        multiCloudWorkspaceConfig
+      )
 
       val userServiceConstructor: (UserInfo) => UserService =
         UserService.constructor(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -14,7 +14,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigObject}
 import com.typesafe.scalalogging.LazyLogging
 import io.sentry.{Hint, Sentry, SentryEvent, SentryOptions}
 import net.ceedubs.ficus.Ficus._
-import org.broadinstitute.dsde.rawls.billing.BillingProfileManagerDAOImpl
+import org.broadinstitute.dsde.rawls.billing.{BillingProfileManagerDAOImpl, BillingProjectOrchestrator, BillingRepository}
 import org.broadinstitute.dsde.rawls.config._
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.HttpDataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.martha.MarthaResolver
@@ -455,6 +455,9 @@ object Boot extends IOApp with LazyLogging {
         spendReportingServiceConfig
       )
 
+      val billingProjectOrchestratorConstructor: (UserInfo) => BillingProjectOrchestrator =
+        BillingProjectOrchestrator.constructor(samDAO, gcsDAO, new BillingRepository(slickDataSource))
+
       val service = new RawlsApiServiceImpl(
         multiCloudWorkspaceServiceConstructor,
         workspaceServiceConstructor,
@@ -463,6 +466,7 @@ object Boot extends IOApp with LazyLogging {
         genomicsServiceConstructor,
         snapshotServiceConstructor,
         spendReportingServiceConstructor,
+        billingProjectOrchestratorConstructor,
         statusServiceConstructor,
         shardedExecutionServiceCluster,
         ApplicationVersion(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -151,7 +151,6 @@ object Boot extends IOApp with LazyLogging {
 
     initAppDependencies[IO](conf, appName, metricsPrefix).use { appDependencies =>
       val gcsDAO = new HttpGoogleServicesDAO(
-        false,
         clientSecrets,
         clientEmail,
         gcsConfig.getString("subEmail"),
@@ -160,18 +159,13 @@ object Boot extends IOApp with LazyLogging {
         dmConfig.orgID,
         gcsConfig.getString("groupsPrefix"),
         gcsConfig.getString("appName"),
-        gcsConfig.getInt("deletedBucketCheckSeconds"),
         serviceProject,
         gcsConfig.getString("billingPemEmail"),
         gcsConfig.getString("pathToBillingPem"),
         gcsConfig.getString("billingEmail"),
         gcsConfig.getString("billingGroupEmail"),
-        gcsConfig.getStringList("billingGroupEmailAliases").asScala.toList,
         dmConfig.billingProbeEmail,
-        gcsConfig.getInt("bucketLogsMaxAge"),
         googleStorageService = appDependencies.googleStorageService,
-        googleServiceHttp = appDependencies.googleServiceHttp,
-        topicAdmin = appDependencies.topicAdmin,
         workbenchMetricBaseName = metricsPrefix,
         proxyNamePrefix = gcsConfig.getStringOr("proxyNamePrefix", ""),
         deploymentMgrProject = dmConfig.projectID,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
@@ -16,10 +16,10 @@ trait BillingProfileManagerClientProvider {
 }
 
 
-class HttpBillingProfileManagerClientProvider(baseBpmUrl: String) extends BillingProfileManagerClientProvider {
+class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extends BillingProfileManagerClientProvider {
   override def getApiClient(accessToken: String): ApiClient = {
     val client: ApiClient = new ApiClient()
-    client.setBasePath(baseBpmUrl)
+    client.setBasePath(basePath)
     client.setAccessToken(accessToken)
 
     client
@@ -31,5 +31,9 @@ class HttpBillingProfileManagerClientProvider(baseBpmUrl: String) extends Billin
 
   override def getProfileApi(accessToken: String): ProfileApi = {
     new ProfileApi(getApiClient(accessToken))
+  }
+
+  private def basePath = {
+    baseBpmUrl.getOrElse(throw new NotImplementedError("Billing profile manager path not configured"))
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
@@ -1,0 +1,31 @@
+package org.broadinstitute.dsde.rawls.billing
+
+import bio.terra.profile.api.{AzureApi, ProfileApi}
+import bio.terra.profile.client.ApiClient
+
+trait BillingProfileManagerClientProvider {
+  def getApiClient(accessToken: String): ApiClient
+
+  def getAzureApi(accessToken: String): AzureApi
+
+  def getProfileApi(accessToken: String): ProfileApi
+}
+
+
+class HttpBillingProfileManagerClientProvider(baseBpmUrl: String) extends BillingProfileManagerClientProvider {
+  override def getApiClient(accessToken: String): ApiClient = {
+    val client: ApiClient = new ApiClient()
+    client.setBasePath(baseBpmUrl)
+    client.setAccessToken(accessToken)
+
+    client
+  }
+
+  override def getAzureApi(accessToken: String): AzureApi = {
+    new AzureApi(getApiClient(accessToken))
+  }
+
+  override def getProfileApi(accessToken: String): ProfileApi = {
+    new ProfileApi(getApiClient(accessToken))
+  }
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
@@ -3,6 +3,10 @@ package org.broadinstitute.dsde.rawls.billing
 import bio.terra.profile.api.{AzureApi, ProfileApi}
 import bio.terra.profile.client.ApiClient
 
+/**
+ * Implementors of this trait know how to instantiate billing profile manager client
+ * classes for use with the billing profile manager service.
+ */
 trait BillingProfileManagerClientProvider {
   def getApiClient(accessToken: String): ApiClient
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -1,23 +1,113 @@
 package org.broadinstitute.dsde.rawls.billing
 
+import scala.jdk.CollectionConverters._
+import akka.http.scaladsl.model.StatusCodes
+import bio.terra.profile.api.{AzureApi, ProfileApi}
+import bio.terra.profile.client.ApiClient
+import bio.terra.profile.model.{AzureManagedAppModel, AzureManagedAppsResponseModel, CloudPlatform, CreateProfileRequest, ProfileModel}
 import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.config.{AzureConfig, MultiCloudWorkspaceConfig}
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
-import org.broadinstitute.dsde.rawls.model.{CreationStatuses, RawlsBillingProject, RawlsBillingProjectName, RawlsBillingProjectResponse, SamResourceAction, SamResourceTypeNames, SamUserResource, UserInfo, AzureManagedAppCoordinates}
+import org.broadinstitute.dsde.rawls.model.{AzureManagedAppCoordinates, CreationStatuses, ErrorReport, RawlsBillingAccountName, RawlsBillingProject, RawlsBillingProjectName, RawlsBillingProjectResponse, SamResourceAction, SamResourceTypeNames, SamUserResource, UserInfo}
 
-import scala.concurrent.{ExecutionContext, Future}
+import java.util.UUID
+import scala.concurrent.{ExecutionContext, Future, blocking}
+import scala.util.{Failure, Success, Try}
 
 trait BillingProfileManagerDAO {
   def listBillingProfiles(userInfo: UserInfo, samUserResources: Seq[SamUserResource])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+
+  def createBillingProfile(displayName: String, billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates], userInfo: UserInfo): Future[ProfileModel]
 }
 
+
+trait BillingProfileManagerClientProvider {
+  def getApiClient(accessToken: String): ApiClient
+
+  def getAzureApi(accessToken: String): AzureApi
+
+  def getProfileApi(accessToken: String): ProfileApi
+}
+
+class HttpBillingProfileManagerClientProvider(baseBpmUrl: String) extends BillingProfileManagerClientProvider {
+  override def getApiClient(accessToken: String): ApiClient = {
+    val client: ApiClient = new ApiClient()
+    client.setBasePath(baseBpmUrl)
+    client.setAccessToken(accessToken)
+
+    client
+  }
+
+  override def getAzureApi(accessToken: String): AzureApi = {
+    new AzureApi(getApiClient(accessToken))
+  }
+
+  override def getProfileApi(accessToken: String): ProfileApi = {
+    new ProfileApi(getApiClient(accessToken))
+  }
+}
+
+class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
 
 /**
  * Facade over the billing profile manager service. This service will eventually by the source of truth for
  * billing profiles in the Terra system. For now, we are using this to layer in "external" billing profiles
  * for the purposes of testing Azure workspaces.
  */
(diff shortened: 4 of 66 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -0,0 +1,31 @@
+package org.broadinstitute.dsde.rawls.billing
+
+import bio.terra.profile.api.{AzureApi, ProfileApi}
+import bio.terra.profile.client.ApiClient
+
+trait BillingProfileManagerClientProvider {
```
aherbst-broad, line 10: This is extracting the HttpClient creation bits from the client DAO for better testability.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -1,30 +1,79 @@
...
+/**
+ * Common interface for Billing Profile Manager operations
+ */
 trait BillingProfileManagerDAO {
-  def listBillingProfiles(userInfo: UserInfo, samUserResources: Seq[SamUserResource])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+  def createBillingProfile(displayName: String, billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates], userInfo: UserInfo): Future[ProfileModel]
+
+  def listBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+
+  def listManagedApps(subscriptionId: UUID, userInfo: UserInfo): Future[Seq[AzureManagedAppModel]]
```
aherbst-broad, line 22: Minor change on this method signature, I put the userInfo last to maintain consistency with the rest of the methods on this trait.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectCreator.scala
```diff
@@ -0,0 +1,56 @@
...
+ * a) the creation request is valid for the given cloud provider
+ * b) external state is valid after rawls internal state is updated (i.e, syncing groups, etc.)
+ */
+trait BillingProjectCreator {
+  def validateBillingProjectCreationRequest(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Unit]
+  def postCreationSteps(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Unit]
+}
+
+
+class GoogleBillingProjectCreator(samDAO: SamDAO, gcsDAO: GoogleServicesDAO) extends BillingProjectCreator with StringValidationUtils {
```
aherbst-broad: This is the "legacy" code path for google project creation, refactored out of the `BillingProjectOrchestrator` class.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -1,33 +1,48 @@
...
       _ <- validateBillingProjectName(createProjectRequest.projectName.value)
-      _ <- ServicePerimeterService.checkServicePerimeterAccess(samDAO, createProjectRequest.servicePerimeter, userInfo)
-      hasAccess <- gcsDAO.testBillingAccountAccess(createProjectRequest.billingAccount, userInfo)
-      _ = if (!hasAccess) {
-        throw new GoogleBillingAccountAccessException(ErrorReport(StatusCodes.BadRequest, "Billing account does not exist, user does not have access, or Terra does not have access"))
-      }
+
+      _ <- billingProjectCreator.validateBillingProjectCreationRequest(createProjectRequest, userInfo)
       result <- createV2BillingProjectInternal(createProjectRequest, userInfo)
+      _ <- billingProjectCreator.postCreationSteps(createProjectRequest, userInfo)
```
aherbst-broad: Add a rollback step if this fails

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -1,33 +1,48 @@
...
+                                 googleBillingProjectCreator: BillingProjectCreator,
+                                 bpmBillingProjectCreator: BillingProjectCreator)
+                                (implicit val executionContext: ExecutionContext) extends StringValidationUtils {
   implicit val errorReportSource = ErrorReportSource("rawls")
 
+  /**
+   * Creates a "v2" billing project, using either Azure managed app coordinates or a Google Billing Account
+   */
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
+    val billingProjectCreator = createProjectRequest.billingInfo match {
```
aherbst-broad, line 36: Extract this to a factory/provider

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -139,8 +140,24 @@ case class CreateRawlsBillingProjectFullRequest(
 // perimeter, we found that they needed the extra security.
 case class CreateRawlsV2BillingProjectFullRequest(
   projectName: RawlsBillingProjectName,
-  billingAccount: RawlsBillingAccountName,
-  servicePerimeter: Option[ServicePerimeterName])
+  billingAccount: Option[RawlsBillingAccountName],
+  servicePerimeter: Option[ServicePerimeterName],
+  managedAppCoordinates: Option[AzureManagedAppCoordinates]) {
+
+  def billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates] = {
```
aherbst-broad, line 147: Maybe a type alias

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -333,8 +333,8 @@ class WorkspaceService(protected val userInfo: UserInfo,
 
           Option(wsmInfo.getAzureContext) match {
             case Some(azureContext) => Some(AzureManagedAppCoordinates(
-              azureContext.getTenantId,
-              azureContext.getSubscriptionId,
+              UUID.fromString(azureContext.getTenantId),
+              UUID.fromString(azureContext.getSubscriptionId),
```
aherbst-broad, line 337: This may need a swagger update

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -53,88 +59,6 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
-    }
-
-    assertResult(Some(StatusCodes.Forbidden)) {
-      ex.errorReport.statusCode
-    }
-  }
-
   behavior of "billing project creation"
 
   it should "create a billing project record when provided a valid request" in {
```
aherbst-broad, line 62: Add verifies on tests that are consuming a Future.successful.()

aherbst-broad: Investigate .thenAnswer()

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -781,13 +782,7 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
       val externalProject =  billingProjectFromName(UUID.randomUUID().toString)
       runAndWait(rawlsBillingProjectQuery.create(ownerProject))
       val samDAO = mock[SamDAO](RETURNS_SMART_NULLS)
```
aherbst-broad, line 784: Use `RETURN_SMART_NULLS`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: After further consideration, this feels like a nice-to-have and I am going to defer that extraction unless folks feel strongly otherwise.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: Implemented in [this commit](https://github.com/broadinstitute/rawls/pull/1961/commits/ba8df16cab10823c9f6f1e92a36ac02416aa8d23)

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -287,7 +287,11 @@ object Boot extends IOApp with LazyLogging {
       val servicePerimeterService = new ServicePerimeterService(slickDataSource, gcsDAO, servicePerimeterConfig)
 
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
-      val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(samDAO, multiCloudWorkspaceConfig)
+      val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
+        samDAO,
+        new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl")),
```
aherbst-broad, line 292: Config will be rendered in https://github.com/broadinstitute/firecloud-develop/pull/3063. I've made this an optional to get swatomation passing.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -1,34 +1,59 @@
...
-        throw new GoogleBillingAccountAccessException(ErrorReport(StatusCodes.BadRequest, "Billing account does not exist, user does not have access, or Terra does not have access"))
+      _ = logger.info(s"Validating billing project creation request [name=${billingProjectName.value}]")
+      _ <- billingProjectCreator.validateBillingProjectCreationRequest(createProjectRequest, userInfo)
+      _ = logger.info(s"Creating billing project record [name=${billingProjectName}]")
+      _ <- createV2BillingProjectInternal(createProjectRequest, userInfo)
+      _ = logger.info(s"Created billing project record, running post-creation steps [name=${billingProjectName.value}]")
+      result <- billingProjectCreator.postCreationSteps(createProjectRequest, userInfo).recoverWith {
+        case t: Throwable =>
+          logger.error(s"Error in post-creation steps for billing project [name=${billingProjectName.value}]")
+          billingRepository.deleteBillingProject(createProjectRequest.projectName).map(_ => throw t)
```
marctalbott: We'll still leak the `billing-project` resource in Sam, which will prevent the user from re-creating a billing project with the same name. I don't know how much effort we want to put into flawless rollbacks, but wanted to note this as a possibility.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectCreator.scala
```diff
@@ -0,0 +1,55 @@
...
+ * a) the creation request is valid for the given cloud provider
+ * b) external state is valid after rawls internal state is updated (i.e, syncing groups, etc.)
+ */
+trait BillingProjectCreator {
+  def validateBillingProjectCreationRequest(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo): Future[Unit]
+  def postCreationSteps(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo): Future[Unit]
+}
+
+
+class GoogleBillingProjectCreator(samDAO: SamDAO, gcsDAO: GoogleServicesDAO)(implicit executionContext: ExecutionContext) extends BillingProjectCreator {
```
marctalbott: Would it be worth pulling this out into a separate file? Even when we add support for Google billing profiles to BPM, we'll still need different validation and post creation steps for Azure and Google in Rawls, right? Or will these Google steps get moved into `BpmBillingProjectCreator` and we'll branch there depending on if it's Google or Azure?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -1,30 +1,79 @@
...
 }
 
 
+class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
```
marctalbott, line 26: Out of curiosity, why extend the `RawlsExceptionWithErrorReport` with these rather than just use the `RawlsExceptionWithErrorReport`? To make it easier to identify the issue when it crops up? I suppose it's a bit easier to test that the correct exception is getting thrown as well

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreatorSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+    )
+    when(
+      samDAO.syncPolicyToGoogle(
+        ArgumentMatchers.eq(SamResourceTypeNames.billingProject),
+        ArgumentMatchers.eq(createRequest.projectName.value),
+        ArgumentMatchers.eq(SamBillingProjectPolicyNames.owner))
+    ).thenReturn(Future.successful(Map(WorkbenchEmail(userInfo.userEmail.value) -> Seq())))
+    val gbp = new GoogleBillingProjectCreator(samDAO, mock[GoogleServicesDAO])
+
+    Await.result(gbp.postCreationSteps(createRequest, userInfo), Duration.Inf)
```
marctalbott, line 103: Is the assertion missing in this test or is `postCreationSteps` not throwing a sufficient assertion? Could add a `verify` that `syncPolicyToGoogle` gets called

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
aherbst-broad: Yeah, that's why, I want to ensure we are propagating the correct exception. There are so many places in our code that raise RawlsException that I wouldn't have confidence that the exception we were intercepting in the test was the correct culprit. We'd probably need to resort to string/message matching which I'm pretty 👎 on.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreatorSpec.scala
aherbst-broad: Good catch, will add that verification.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: Ah good note. I'll take a look at this.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectCreator.scala
aherbst-broad: I think we'll likely have a google flavor and an azure flavor of creator, though the billing project creation code may be reused a bit. I've moved this to a separate file.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -781,13 +782,7 @@ class UserServiceSpec extends AnyFlatSpecLike with TestDriverComponent with Mock
...
-      val bpmDAO = new BillingProfileManagerDAO {
-        override def listBillingProfiles(userInfo: UserInfo, samUserResources: Seq[SamUserResource])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
-          Future.successful(Seq(
-            externalProject
-          ))
-        }
-      }
+      val bpmDAO = mock[BillingProfileManagerDAO](RETURNS_SMART_NULLS)
```
blakery, line 785: How did I go this long without knowing about RETURNS_SMART_NULLS and the related mockito configs...

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
blakery: looks like the type should still be a string, but you an specify uuid as a _format_ option

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: It also makes for better pattern matching if we're ever handling the exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+    for {
+      apps <- blocking {
+        billingProfileManagerDAO.listManagedApps(azureManagedAppCoordinates.subscriptionId, userInfo)
+      }
+      _ = apps.find(app => app.getSubscriptionId == azureManagedAppCoordinates.subscriptionId &&
+        app.getManagedResourceGroupId == azureManagedAppCoordinates.managedResourceGroupId &&
+        app.getTenantId == azureManagedAppCoordinates.tenantId
+      ) match {
+        case None => throw new ManagedAppNotFoundException(ErrorReport(StatusCodes.Forbidden, "Managed application not found"))
+        case Some(a) => a
```
blakery: Nit: This structure is a bit weird: Since the value is being unwrapped and returned, it looks like it actually matters. But the only part of this that actually matters, is that it's not None.
It might a bit more clear to match with _, then return Unit.

blakery (approved):
I think most of what I had for this was covered Friday.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
aherbst-broad: I know, it's super useful, Marc pointed that out during our codewalk

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: @marctalbott updated, please take a look.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5330,7 +5330,6 @@ components:
       description: ""
     CreateRawlsV2BillingProjectFullRequest:
       required:
-        - billingAccount
```
cahrens, line 5333: If this optional argument is present it should represent a Google billing project, correct? If so, I think it would be good to clarify that in the documentation below (line 5341).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -0,0 +1,39 @@
...
+import bio.terra.profile.client.ApiClient
+
+/**
+ * Implementors of this trait know how to instantiate billing profile manager client
+ * classes for use with the billing profile manager service.
+ */
+trait BillingProfileManagerClientProvider {
+  def getApiClient(accessToken: String): ApiClient
+
+  def getAzureApi(accessToken: String): AzureApi
```
cahrens, line 13: My first thought when reading this… why is there an Azure Api and not a GCP Api? What exactly is the ApiClient and the ProfileApi… is one of those really GCP-specific? Perhaps some documentation would help me understand the role of these 3 client classes, but for now I will continue reading and see if it all becomes clear.

cahrens: Also, it appears that `getApiClient` is only used by the other 2 methods. So does it need to be exposed as a public part of the interface?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -1,30 +1,79 @@
...
  * billing profiles in the Terra system. For now, we are using this to layer in "external" billing profiles
  * for the purposes of testing Azure workspaces.
  */
-class BillingProfileManagerDAOImpl(samDAO: SamDAO, config: MultiCloudWorkspaceConfig) extends BillingProfileManagerDAO with LazyLogging {
+class BillingProfileManagerDAOImpl(samDAO: SamDAO,
+                                   apiClientProvider: BillingProfileManagerClientProvider,
+                                   config: MultiCloudWorkspaceConfig) extends BillingProfileManagerDAO with LazyLogging {
+
+
+  override def listManagedApps(subscriptionId: UUID, userInfo: UserInfo): Future[Seq[AzureManagedAppModel]] = {
```
cahrens, line 38: I would be tempted to make this method `listAzureManagedApps` for clarity, though perhaps the return type is sufficient for providing that contextual clue.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
cahrens: Having gotten further, I am now reminded that BPM does not support GCP yet. So presumably someday there may be a `getGcpApi` method?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def getProfileApi(accessToken: String): ProfileApi
```
cahrens, line 15: I would suggest calling this `getBillingProfileApi`. When I first read this trait, I was confused as to what profile referred to (given that we also have other types of profiles in Terra).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: Actually, this is interesting because someday (I believe), the BillingProfileManager will also support GCP, and yet GCP will never have managed apps. That increases my discomfort with this method.

cahrens: I wonder if it would be clearer if this method took `AzureManagedAppCoordinates` as an argument instead of `subscriptionId`. I get that it only needs the `subscriptionId`, but as-is the caller has to determine where to get the `subscriptionId` from.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -1,30 +1,79 @@
...
+    val azureApi = apiClientProvider.getAzureApi(userInfo.accessToken.token)
+
+    val result = azureApi.getManagedAppDeployments(subscriptionId).getManagedApps.asScala.toList
+    Future.successful(result)
+  }
+
+  override def createBillingProfile(displayName: String,
+                                    billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates],
+                                    userInfo: UserInfo): Future[ProfileModel] = {
+    val azureManagedAppCoordinates = billingInfo match {
```
cahrens, line 48: @aherbst-broad what are your thoughts about when/how we add test coverage of this method?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * This class knows how to validate Rawls billing project requests and instantiate linked billing profiles in the
+ * billing profile manager service.
+ */
+class BpmBillingProjectCreator(billingRepository: BillingRepository,
+                               billingProfileManagerDAO: BillingProfileManagerDAO)
+                              (implicit val executionContext: ExecutionContext) extends BillingProjectCreator {
+
+  override def validateBillingProjectCreationRequest(createProjectRequest: CreateRawlsV2BillingProjectFullRequest,
```
cahrens, line 23: From my Java memories, I would expect this method to either return a boolean (indicating whether it is valid), or throw a checked exception. I see that this is using exceptions to indicate if something is invalid. In Scala are you not encouraged to declare checked exceptions that callers should be prepared to handle (vs. undeclared runtime errors that are indication of code errors)?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -1,34 +1,72 @@
...
+      _ = logger.info(s"Creating billing project record [name=${billingProjectName}]")
+      _ <- createV2BillingProjectInternal(createProjectRequest, userInfo)
+
+      _ = logger.info(s"Created billing project record, running post-creation steps [name=${billingProjectName.value}]")
+      result <- billingProjectCreator.postCreationSteps(createProjectRequest, userInfo).recoverWith {
+        case t: Throwable =>
+          // rollback any resources we created
+          logger.error(s"Error in post-creation steps for billing project [name=${billingProjectName.value}]")
+          for {
+            _ <-billingRepository.deleteBillingProject(createProjectRequest.projectName).recover {
```
cahrens: These are rolling back things created in `createV2BillingProjectInternal`. I wonder if it's worth putting these into a method called `rollbackCreateV2BillingProjectInternal` for clarity. I really like how the rest of `createBillingProjectV2` reads, but when I got here I was like… wait, why does this code know what  `createV2BillingProjectInternal` did?

cahrens: `createV2BillingProjectInternal` could possibly even return a rollback function that you would call here.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import java.net.URLEncoder
+import scala.concurrent.{ExecutionContext, Future}
+import java.nio.charset.StandardCharsets.UTF_8
+
+class GoogleBillingProjectCreator(samDAO: SamDAO, gcsDAO: GoogleServicesDAO)(implicit executionContext: ExecutionContext) extends BillingProjectCreator {
+  implicit val errorReportSource: ErrorReportSource = ErrorReportSource("rawls")
+
+  override def validateBillingProjectCreationRequest(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo): Future[Unit] = {
+    for {
```
cahrens, line 23: Scala question… why is this written as a for/yield instead of simply 3 sequential steps?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
```diff
@@ -0,0 +1,42 @@
...
+      _ = if (!hasAccess) {
+        throw new GoogleBillingAccountAccessException(ErrorReport(StatusCodes.BadRequest, "Billing account does not exist, user does not have access, or Terra does not have access"))
+      }
+    } yield {}
+  }
+
+
+  override def postCreationSteps(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo): Future[Unit] = {
+    for {
+      _ <- syncBillingProjectOwnerPolicyToGoogleAndGetEmail(samDAO, createProjectRequest.projectName)
```
cahrens, line 35: Again, why a for/yield here?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
```diff
@@ -3,13 +3,20 @@ package org.broadinstitute.dsde.rawls.billing
...
 
+import java.util.UUID
 import scala.concurrent.Future
 
 /**
  * Data access for rawls billing projects
  */
 class BillingRepository(dataSource: SlickDataSource) {
 
+  def setBillingProfileId(projectName: RawlsBillingProjectName, billingProfileId: UUID): Future[Int] = {
```
blakery, line 14: Should we maybe have a sub-package for DAOs?
I'm not adamant about it or anything, but it might make things feel a little more organized.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
```diff
@@ -0,0 +1,42 @@
...
+    } yield {}
+  }
+}
+
+object GoogleBillingProjectCreator {
+  def checkServicePerimeterAccess(servicePerimeterOption: Option[ServicePerimeterName], samDAO: SamDAO, userInfo: UserInfo)(implicit executionContext: ExecutionContext): Future[Unit] = {
+    servicePerimeterOption.map { servicePerimeter =>
```
cahrens: This used to call `ServicePerimeterService.checkServicePerimeterAccess(samDAO, createProjectRequest.servicePerimeter, userInfo)`. Why did you decide to implement something custom here?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    for {
+      apps <- blocking {
+        billingProfileManagerDAO.listManagedApps(azureManagedAppCoordinates.subscriptionId, userInfo)
+      }
+      _ = apps.find(app => app.getSubscriptionId == azureManagedAppCoordinates.subscriptionId &&
+        app.getManagedResourceGroupId == azureManagedAppCoordinates.managedResourceGroupId &&
+        app.getTenantId == azureManagedAppCoordinates.tenantId
+      ) match {
+        case None => throw new ManagedAppNotFoundException(ErrorReport(StatusCodes.Forbidden, "Managed application not found"))
```
cahrens: Will it be obvious from the exception stacktrace what the managedAppCoordinates were?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+        billingProfileManagerDAO.createBillingProfile(createProjectRequest.projectName.value, createProjectRequest.billingInfo, userInfo)
+      }
+      _ <- billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
+    } yield {}
+  }
+
+}
+
+
+class ManagedApplicationAccessException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
```
cahrens: I don't think anything uses `ManagedApplicationAccessException`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+
+class InvalidCreationRequest(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
```
cahrens: I don't think anything uses `InvalidCreationRequest`.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -139,8 +140,24 @@ case class CreateRawlsBillingProjectFullRequest(
...
+    if (billingAccount.isDefined && managedAppCoordinates.isDefined) {
+      throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Invalid billing project creation request"))
+    }
+    if (billingAccount.isEmpty && managedAppCoordinates.isEmpty) {
+      throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Invalid billing project creation request"))
```
cahrens: Perhaps add detail that both things cannot be empty.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -139,8 +140,24 @@ case class CreateRawlsBillingProjectFullRequest(
...
   projectName: RawlsBillingProjectName,
-  billingAccount: RawlsBillingAccountName,
-  servicePerimeter: Option[ServicePerimeterName])
+  billingAccount: Option[RawlsBillingAccountName],
+  servicePerimeter: Option[ServicePerimeterName],
+  managedAppCoordinates: Option[AzureManagedAppCoordinates]) {
+
+  def billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates] = {
+    if (billingAccount.isDefined && managedAppCoordinates.isDefined) {
+      throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Invalid billing project creation request"))
```
cahrens: Perhaps add detail that both things cannot be specified.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
blakery: I don't think it's _automatically_ an indication of a code error if an exception propagates up. It kind of just depends on how your service is structured. 
In this service, I think RawlsExceptionWithErrorReport is intended to be an exception that is caught higher up, and not necessarily handled by the caller, in order to propagate to the user. 

You raise a good point about NotImplementedError though. 
IMO, it's probably fine if you don't expect to hit it (making it more of a sanity check sort of thing), but if it's more validation that might be too low level to throw as-is.

ManagedAppNotFoundException extends RawlsExceptionWithErrorReport so as far as I can tell, should be fine to throw bringing explicit attention to it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -1,34 +1,72 @@
...
+   */
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
+    val billingProjectCreator = createProjectRequest.billingInfo match {
+      case Left(_) => googleBillingProjectCreator
+      case Right(_) => bpmBillingProjectCreator
+    }
+    val billingProjectName = createProjectRequest.projectName
+
     for {
       _ <- validateBillingProjectName(createProjectRequest.projectName.value)
```
blakery, line 43: What is the value of the for comprehension if none of the values are carried through, and we aren't returning anything? Wouldn't just a foreach be more clear for that?

core/src/main/resources/swagger/api-docs.yaml
cahrens: I guess it is technically a Rawls billingAccount, but it is only used for GCP?

cahrens:
@aherbst-broad I'm done reviewing!

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: It's a google billing account, and it's now optional given the presence of the managedAppCoordinates arg (technically an `Either` relationship). I'll update the docs 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
aherbst-broad: Yes, this is a structure dictated by the auto generated billing profile manager client JAR. These `AzureApi`, `FooApi` types directly correspond to the controllers/sections on the BPM swagger (https://bpm.dsde-dev.broadinstitute.org/swagger-ui.html).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
aherbst-broad: The type as exposed by BPM is the `ProfileApi`, we should keep those in sync.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: Yeah, that works, I'll move this over there.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
aherbst-broad: Let's try it here for now, it's easy enough to move if need be.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+                              (implicit val executionContext: ExecutionContext) extends BillingProjectCreator {
+
+  override def validateBillingProjectCreationRequest(createProjectRequest: CreateRawlsV2BillingProjectFullRequest,
+                                                     userInfo: UserInfo): Future[Unit] = {
+    val azureManagedAppCoordinates = createProjectRequest.billingInfo match {
+      case Left(_) => throw new NotImplementedError("Google billing accounts not supported in billing profiles")
+      case Right(coords) => coords
+    }
+
+    for {
```
cahrens, line 30: Do we need the for/yield here?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -0,0 +1,57 @@
...
+      }
+    } yield {}
+  }
+
+  /**
+   * Creates a billing profile with the given billing creation info and links the previously created billing project
+   * with it
+   */
+  override def postCreationSteps(createProjectRequest: CreateRawlsV2BillingProjectFullRequest, userInfo: UserInfo): Future[Unit] = {
+    for {
```
cahrens, line 53: For/yield question again.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
aherbst-broad: This method is intended to serve two purposes:
1. For the context of this PR, enumerate the managed apps and ensure one exists prior to going ahead with profile creation. BPM does this as I look at the create [profile code there](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/54ae94586a2c1d7aed7166114a9ad4f83475d09e/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java#L24), so I think we may want to forgo the call to this method altogether [here](https://github.com/broadinstitute/rawls/blob/6e84bd7ac99efe20bd84fdb5ae88c7ae3c97c134/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala#L27)

_but_

2. We also need this method to populate the future managed apps dropdown in the UI. The user will provide a subscription ID and we will need to call this method to enumerate the managed apps in the subscription. At that time in the profile creation lifecycle, we will not have the tenant ID nor the managed resource group ID so I think signature needs to remain as-is.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
aherbst-broad: Good eye, will add some coverage.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: Scala does not have checked exceptions, they are all runtime (at least in scala 2, I am unsure about 3). 

Re: `NotImplementedError` I want some kind of precondition here as a sanity check as @blakery mentions; I am fine raising some other exception, but the intention is for this to be a "this should never happen" assertion.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: OK, makes sense about the signature.

If you don't call this method in the validate step, then it seems that you are at risk of finding out later that the managed app actually doesn't exist, requiring the execution of the rollback code. If that's true, I would stick with doing this in validate.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: Good call, I'll include those in the message.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: Yup, these are both vestigial, will remove.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: Yup, these are both vestigial, will remove.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
cahrens: I can see not mentioning `NotImplementedError`, but I think it should be documented in some way how the method indicates that the request isn't valid. Ideally I would expect that to be with "throws ManagedAppNotFoundException", but since that isn't possible, documentation seems like it would be helpful. Another alternative would be to follow this pattern:

```
More generally, Either[E,V] has been used as a return type with the convention that returning Left[E] holds an error E, such as an exception, while Right[V] returns a successful value of type V. Try[V] works similarly.
```

It looks like checked exceptions are being considered for Scala (which is where I also got ^)-- https://medium.com/scala-3/safer-exceptions-in-scala-3-12c47100f050.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: So this is syntactic sugar over flatmapping. That is, we need to express a dependency between the two futures:
1) The BPM operation to create the profile
2) The repo operation to update the billing profile ID which needs the UUID returned from step 1 to establish the linkage. 

Desguared, we could do something like:

```
billingProfileManagerDAO.createBillingProfile(createProjectRequest.projectName.value, createProjectRequest.billingInfo, userInfo).flatMap {
      profileModel => billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
    }
```

I think the for comprehension is slightly more readable, but the desugared approach is fine with me as well if you feel strongly about it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: ~As mentioned in a previous comment, I think this check is redundant given what we're doing in BPM, I'll remove this. See my subsequent comment re: sequencing futures in a for comprehension.~ Actually, this should remain so we don't have to go through the rollback process if the managed app does not exist.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
cahrens: I don't feel strongly about it if there's a reason… I just couldn't figure out what the reason was. I am now reading through https://docs.scala-lang.org/overviews/core/futures.html.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
cahrens: So this is again because of Futures (`checkServicePerimeterAccess` returns a `Future[Unit]`)?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
aherbst-broad: Ugh, that's a mistake, I think an artifact from a rebase from a related branch. I will remove this dupe method.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectCreator.scala
aherbst-broad: That's correct.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
aherbst-broad: I like the idea of returning a semantically clearer `Either[E, V]`. That being said, most of the consuming code for this class is expecting a failed future rather than a successful future. 

Also given the general approach in Rawls seems to be to throw and fail the future, I am inclined to stick with the current approach for consistency and update the docs.

cahrens: Sure, that's fine… I've come around to it now.

cahrens (approved):

PR #860, opened 2018-02-26
Scala212

dvoet:
migration to scala 2.12 and akka http

main approach for REST api servicing code: 
* stop using actors
* all the messages have changed into functions usually returning Future[PerRequestMessage] which are completed directly in the routes
* most of PerRequest is gone, instead is a ToResponseMarshaller that knows how to handle PerRequestMessage
* This way most of the changes are formulaic: calling per request, creating a new actor, creating and sending a message replaced with object instantiation and direct function call wrapped in a "complete" directive.
* Handling Location headers is different and does not use PerRequestMessage

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check that the **Product Owner** has signed off on any user-facing changes
- [ ] **Submitter**: Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Database checks:
  * If PR includes new or changed db queries, include the explain plans in the description
  * Make sure liquibase is updated if appropriate
  * If doing a migration, take a backup of the
  [dev](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/backups?project=broad-dsde-dev&organizationId=548622027621)
  and
  [alpha](https://console.cloud.google.com/sql/instances/terraform-r4caezzc35c4tb7pgdhwkmme4y/backups?project=broad-dsde-alpha&organizationId=548622027621)
  DBs in Google Cloud Console
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] Tell your tech lead (TL) that the PR exists if they want to look at it
- [ ] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
[...]

PR #1964, opened 2022-08-02
WOR-408: add metrics around api endpoints

blakery:
This covers the rawls source portion of [WOR-408](https://broadworkbench.atlassian.net/browse/WOR-408)

All endpoints are wrapped in a instrumentation function that reports request times and error rates to prometheus via the statsd-export container.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
@@ -96,16 +96,14 @@ trait RawlsApiService //(val workspaceServiceConstructor: UserInfo => WorkspaceS
   implicit val executionContext: ExecutionContext
   implicit val materializer: Materializer
 
-  val baseApiRoutes = workspaceRoutes ~ entityRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutesV2 ~ billingRoutes ~ notificationsRoutes ~ servicePerimeterRoutes ~ snapshotRoutes
+  val baseApiRoutes = workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutesV2 ~ billingRoutes ~ notificationsRoutes ~ servicePerimeterRoutes ~ snapshotRoutes
 
-  val instrumentedRoutes = instrumentRequest(Route.seal(baseApiRoutes))
+  val instrumentedRoutes = instrumentRequest(baseApiRoutes)
 
   def apiRoutes =
     options(complete(OK)) ~
     withExecutionContext(ExecutionContext.global) { //Serve real work off the global EC to free up the dispatcher to run more routes, including status
-      // tests for methodConfigRoutes fail when sealed, so they don't work when instrumented either
-      // the methodConfigRoutes tests also fail when some routes are before it in the order here
-      methodConfigRoutes ~ instrumentedRoutes
+      instrumentedRoutes
     }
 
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -96,14 +96,19 @@ trait RawlsApiService //(val workspaceServiceConstructor: UserInfo => WorkspaceS
...
+
+  val instrumentedRoutes = instrumentRequest(Route.seal(baseApiRoutes))
 
   def apiRoutes =
     options(complete(OK)) ~
     withExecutionContext(ExecutionContext.global) { //Serve real work off the global EC to free up the dispatcher to run more routes, including status
-      baseApiRoutes
+      // tests for methodConfigRoutes fail when sealed, so they don't work when instrumented either
+      // the methodConfigRoutes tests also fail when some routes are before it in the order here
+      methodConfigRoutes ~ instrumentedRoutes
```
blakery: The fact that methodConfigRoutes seems to behave differently based on order is concerning to me. I think it would imply ambiguous routes.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -98,12 +98,15 @@ trait RawlsApiService //(val workspaceServiceConstructor: UserInfo => WorkspaceS
 
   val baseApiRoutes = workspaceRoutes ~ entityRoutes ~ methodConfigRoutes ~ submissionRoutes ~ adminRoutes ~ userRoutes ~ billingRoutesV2 ~ billingRoutes ~ notificationsRoutes ~ servicePerimeterRoutes ~ snapshotRoutes
 
+  val instrumentedRoutes = instrumentRequest(baseApiRoutes)
+
   def apiRoutes =
     options(complete(OK)) ~
     withExecutionContext(ExecutionContext.global) { //Serve real work off the global EC to free up the dispatcher to run more routes, including status
-      baseApiRoutes
+      instrumentedRoutes
```
cahrens, line 106: `instrumentedRoutes` was used briefly for non-test code in 2017, then removed [here](https://github.com/broadinstitute/rawls/pull/860/files#diff-7dbb8ca023296657cb3f8bdc7e7fcd59737d95cff157d294f177817a476490d5L39). @dvoet, any concerns about putting this back in?

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
msilva-broad, line 106: poking into instrumentRequest a bit, I see that it calls mapResponse... noob question... will this work for error/exception cases as well as successful response cases?

msilva-broad: also, is there a unit test to assert the expected logging?

blakery: > will this work for error/exception cases as well as successful response cases?
Short answer: 
It does work for both cases. 

Long answer:
A route can have 1 of 3 outcomes:
1. The route returns successfully (an httpResponse where status == 2XX) 
2. The route returns failure (an httpResponse where status == 4-5XX)
3. The route returns rejected (the route cannot process that request - and the next route gets a try at it)
If the route is 'sealed', number \# 3 is mapped to \# 2, and the result is always guaranteed to be a completed http response. 
When run from a full application, so long as  the correct handlers are in scope, the routes are implicitly sealed.
So yes, this function gets the http response that is the result of the call, whether success or failure.

> also, is there a unit test to assert the expected logging?

Unit tests... not really...
Integrated checks.... there's quite a few that were already there.
I can write some new ones that are proper unit tests if we think it's worth the time to fight with the existing infrastructure (and ideally break up the existing tests into more digestible pieces).

msilva-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
msilva-broad: ok great.

Unit test question was more informational.  Not a hard blocker since the state of things sounds like a massive lift.   It wouldn't be fair to ask for such a heavy lift to make the overarching codebase more testable RIGHT now since the IRWG is actively exploring best practices and example of doing that :)

PR #1965, opened 2022-08-02
add entity_type_id to ENTITY table [AJ-540]

davidangb:
Do not merge until after the scheduled manual db-migration downtime.

Is there any drawback to proactively adding a `entity_type_id` column to the `ENTITY` table, without updating any of the Slick/Scala code that uses the table? Tests on this PR say no.

We added this column in the prod db during the scheduled case-sensitivity downtime, but have not added it anywhere else; let’s do that to keep lower envs (and local envs) in sync with prod.


jordansafer (approved):

calypsomatic (approved):
LGTM, 👍 assuming tests pass.

PR #1969, opened 2022-08-03
WOR-455 Update listWorkspaces 

ghost:
https://broadworkbench.atlassian.net/browse/WOR-455 

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -776,8 +776,12 @@ class WorkspaceService(protected val userInfo: UserInfo,
                 true
               } catch {
                 case e: ApiException =>
-                  logger.warn(s"MC Workspace ${workspace.name} (${workspace.workspaceIdAsUUID}) does not exist in the current WSM instance. ")
-                  false
+                  if (e.getCode == StatusCodes.NotFound.intValue) {
+                    logger.warn(s"MC Workspace ${workspace.name} (${workspace.workspaceIdAsUUID}) does not exist in the current WSM instance. ")
+                    false
+                 } else {
+                    true
+                  }
               }
             case _ => true
           }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -765,38 +765,38 @@ class WorkspaceService(protected val userInfo: UserInfo,
               )
             }
           }
-
-        workspaces.mapFilter { workspace =>
-          try {
-            val cloudPlatform = workspace.workspaceType match {
-              case WorkspaceType.McWorkspace => Option(workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, userInfo.accessToken)) match {
-                case Some(mcWorkspace) if (mcWorkspace.getAzureContext != null) =>  Option(WorkspaceCloudPlatform.Azure)
-                case Some(mcWorkspace) if (mcWorkspace.getGcpContext != null) => Option(WorkspaceCloudPlatform.Gcp)
-                case _ =>  throw new RawlsException(s"unexpected state, no cloud context found for workspace ${workspace.workspaceId}")
+          workspaces.mapFilter { workspace =>
+            try {
+              val cloudPlatform = workspace.workspaceType match {
+                case WorkspaceType.McWorkspace => Option(workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, ctx)) match {
+                  case Some(mcWorkspace) if (mcWorkspace.getAzureContext != null) => Option(WorkspaceCloudPlatform.Azure)
+                  case Some(mcWorkspace) if (mcWorkspace.getGcpContext != null) => Option(WorkspaceCloudPlatform.Gcp)
+                  case _ => throw new RawlsException(s"unexpected state, no cloud context found for workspace ${workspace.workspaceId}")
+                }
+                case WorkspaceType.RawlsWorkspace => Option(WorkspaceCloudPlatform.Gcp)
               }
-              case WorkspaceType.RawlsWorkspace => Option(WorkspaceCloudPlatform.Gcp)
-            }
-            val wsId = UUID.fromString(workspace.workspaceId)
-            val workspacePolicy = policiesByWorkspaceId(workspace.workspaceId)
-            val accessLevel = if (workspacePolicy.missingAuthDomainGroups.nonEmpty) WorkspaceAccessLevels.NoAccess else WorkspaceAccessLevels.withPolicyName(workspacePolicy.accessPolicyName.value).getOrElse(WorkspaceAccessLevels.NoAccess)
-            // remove attributes if they were not requested
-            val workspaceDetails = WorkspaceDetails.fromWorkspaceAndOptions(workspace, Option(workspacePolicy.authDomainGroups.map(groupName => ManagedGroupRef(RawlsGroupName(groupName.value)))), attributesEnabled, cloudPlatform)
-            // remove submission stats if they were not requested
-            val submissionStats: Option[WorkspaceSubmissionStats] = if (submissionStatsEnabled) {
-              Option(submissionSummaryStats(wsId))
-            } else {
-              None
-            }
-            Option(WorkspaceListResponse(accessLevel, workspaceDetails, submissionStats, workspacePolicy.public))
-
-          } catch {
-            case ex: ApiException if ex.getCode == StatusCodes.NotFound.intValue => {
-              logger.warn(s"MC Workspace ${workspace.name} (${workspace.workspaceIdAsUUID}) does not exist in the current WSM instance. ")
-              None
+              val wsId = UUID.fromString(workspace.workspaceId)
+              val workspacePolicy = policiesByWorkspaceId(workspace.workspaceId)
+              val accessLevel = if (workspacePolicy.missingAuthDomainGroups.nonEmpty) WorkspaceAccessLevels.NoAccess else WorkspaceAccessLevels.withPolicyName(workspacePolicy.accessPolicyName.value).getOrElse(WorkspaceAccessLevels.NoAccess)
+              // remove attributes if they were not requested
+              val workspaceDetails = WorkspaceDetails.fromWorkspaceAndOptions(workspace, Option(workspacePolicy.authDomainGroups.map(groupName => ManagedGroupRef(RawlsGroupName(groupName.value)))), attributesEnabled, cloudPlatform)
+              // remove submission stats if they were not requested
+              val submissionStats: Option[WorkspaceSubmissionStats] = if (submissionStatsEnabled) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -765,7 +765,25 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+        // Internal folks may create MCWorkspaces in local WorkspaceManager instances, and those will not
+        // be reachable when running against the dev environment.
+        val reachableWorkspaces = workspaces.filter { workspace =>
+          workspace.workspaceType match {
+            case WorkspaceType.McWorkspace =>
+              try {
+                workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, userInfo.accessToken)
+                true
+              } catch {
+                case e: ApiException =>
```
aherbst-broad: Check the status code here, we should really only be catching and ignoring if it's 404.

aherbst-broad (commented):
Thanks for doing this! Same comment re: status codes as before.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -765,7 +765,29 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
           }
-          workspaces.map { workspace =>
+
+        // Internal folks may create MCWorkspaces in local WorkspaceManager instances, and those will not
+        // be reachable when running against the dev environment.
+        val reachableWorkspaces = workspaces.filter { workspace =>
+          workspace.workspaceType match {
+            case WorkspaceType.McWorkspace =>
+              try {
+                workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, userInfo.accessToken)
```
cahrens: Any opinions about avoiding 2 calls to `getWorkspace` per McWorkspace? It would require either storing the first result from `getWorkspace` or doing a combination of filter/map. I couldn't figure out how to do the exception handling with a collect, but I thought Scala experts might have some ideas.

On the other hand, I would expect that the getWorkspace call is O(n), and thus calling it twice shouldn't introduce much inefficiency. @aherbst-broad am I incorrect about the cost of calling getWorkspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -765,7 +765,29 @@ class WorkspaceService(protected val userInfo: UserInfo,
...
+                true
+              } catch {
+                case e: ApiException =>
+                  if (e.getCode == StatusCodes.NotFound.intValue) {
```
cahrens: You can put this into line 778.

`case ex: ApiException if ex.getStatusCode == StatusCodes. NotFound.intValue => `

cahrens:
Does anyone know how `adminListWorkspacesWithAttribute` is used? I'm wondering if we need to worry that it won't have the new cloudContext value. Looking at the code, I can't figure out what the point of the method is.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
blakery: list.mapFilter() uses a function that returns an option, and filters out all None values.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -2073,7 +2073,39 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
...
+        _ <- slickDataSource.dataAccess.workspaceQuery.createOrUpdate(googleWorkspace)
+      } yield()
+    }
+
+    when (service.workspaceManagerDAO.getWorkspace(azureWorkspace.workspaceIdAsUUID, services.userInfo1.accessToken)).thenAnswer(
+      _ => throw new ApiException(StatusCodes.NotFound.intValue, "not found"))
+    when (service.workspaceManagerDAO.getWorkspace(googleWorkspace.workspaceIdAsUUID, services.userInfo1.accessToken)).thenReturn(
+      new WorkspaceDescription().gcpContext(new GcpContext())
+    )
+    when (service.samDAO.getPoliciesForType(SamResourceTypeNames.workspace, services.userInfo1)).thenReturn(
```
blakery: Nit/Tip:
It generally works better with mockito to use argument matchers, like this: `when (service.samDAO.getPoliciesForType(eq(SamResourceTypeNames.workspace), eq(services.userInfo1)))`

That lets you do things like mix and match: `when (service.samDAO.getPoliciesForType(eq(SamResourceTypeNames.workspace), any()))`
and still works if it isn't the exact object that was passed.

What's here works fine as-is, but I figured this might be useful info generally.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -2073,7 +2073,39 @@ class WorkspaceServiceSpec extends AnyFlatSpec with ScalatestRouteTest with Matc
...
+    val expected = List((googleWorkspaceDetails.workspaceId, googleWorkspaceDetails.cloudPlatform))
+
+    runAndWait {
+      for {
+        _ <- slickDataSource.dataAccess.workspaceQuery.createOrUpdate(azureWorkspace)
+        _ <- slickDataSource.dataAccess.workspaceQuery.createOrUpdate(googleWorkspace)
+      } yield()
+    }
+
+    when (service.workspaceManagerDAO.getWorkspace(azureWorkspace.workspaceIdAsUUID, services.userInfo1.accessToken)).thenAnswer(
```
blakery: General question:
I'm slightly confused about what is mocked, and what is going to actual services/data sources.
It looks like there's mocking here to handle the WorkspaceDescription, but then the workspaces themselves are created using an actual (connected) slickDataSource?
Why not mock that as well?

I realize that this is consistent with other tests here, and I don't necessarily expect any changes from this: I'm just trying to get context, and maybe provoke a little re-evaluation of our usage patterns.

blakery (approved):
None of my questions are anything that must be addressed - they're more just conversational points.
I'm also assuming the current (2 call) implementation of the getWorkspace calls aren't problematic.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Ohhh, that sounds perfect. @yunip20, can you give that a try?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -765,7 +765,9 @@ class WorkspaceService(protected val userInfo: UserInfo,
               )
             }
           }
-          workspaces.map { workspace =>
+
+        workspaces.mapFilter { workspace =>
```
cahrens: The indenting has gotten off on this code (all the way down to line 809). Can you please fix that?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -786,20 +787,26 @@ class WorkspaceService(protected val userInfo: UserInfo,
             } else {
               None
             }
+            Option(WorkspaceListResponse(accessLevel, workspaceDetails, submissionStats, workspacePolicy.public))
 
-            WorkspaceListResponse(accessLevel, workspaceDetails, submissionStats, workspacePolicy.public)
+          } catch {
+            case ex: ApiException if ex.getCode == StatusCodes.NotFound.intValue => {
```
cahrens: Can you add a comment about why we are doing this? 

Also, we only intend to catch the exception in the case where the workspace is a MCWorkspace, so can you check that as well at line 793?

cahrens: Something along the lines of

```
case ex: ApiException if (WorkspaceType.McWorkspace == workspace.workspaceType) && (ex.getCode == StatusCodes.NotFound.intValue) => {

Comment (ghost) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala]: To avoid multiple calls to `getWorkspace`, we use `list.mapFilter()` which returns an `Option` and filters out all `None` values. 
The try catch block catches any ApiExceptions (specifically 404 errors) and returns `None`. `list.mapFilter()` will then filter out those workspaces and only return `WorkspaceListResponse` for workspaces that do not throw those errors.

Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala]: Sorry, I meant a comment in the code for future people who might be confused about why we are handling the 404 error in this way. Something like:

```
          // Internal folks may create MCWorkspaces in local WorkspaceManager instances, and those will not
          // be reachable when running against the dev environment.

ghost: Got it, sorry I'll add that in the code!

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -767,29 +767,36 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+              val submissionStats: Option[WorkspaceSubmissionStats] = if (submissionStatsEnabled) {
+                Option(submissionSummaryStats(wsId))
+              } else {
+                None
+              }
+              Option(WorkspaceListResponse(accessLevel, workspaceDetails, submissionStats, workspacePolicy.public))
+            } catch {
+              // Internal folks may create MCWorkspaces in local WorkspaceManager instances, and those will not
+              // be reachable when running against the dev environment.
+              case ex: ApiException if ex.getCode == StatusCodes.NotFound.intValue => {
```
cahrens: @yunip20 you lost the check for whether it is a MC workspace.

cahrens:
@yunip20 unfortunately the code change that required you to rebase got reverted, so now you have merge conflicts again. :(

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: https://github.com/broadinstitute/rawls/pull/1969#discussion_r938122447

cahrens (approved):

cahrens:
jenkins restart

cahrens:
jenkins restart

cahrens:
jenkins restart

PR #1994, opened 2022-08-18
First pass at including billing profiles from BPM (WOR-400).

cahrens:
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -287,9 +287,10 @@ object Boot extends IOApp with LazyLogging {
       val servicePerimeterService = new ServicePerimeterService(slickDataSource, gcsDAO, servicePerimeterConfig)
 
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
+      val billingProfileManagerClientProvider = new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl"))
       val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
         samDAO,
-        new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl")),
+        billingProfileManagerClientProvider,
         multiCloudWorkspaceConfig
       )
 
@@ -398,6 +399,7 @@ object Boot extends IOApp with LazyLogging {
       val multiCloudWorkspaceServiceConstructor: (UserInfo) => MultiCloudWorkspaceService = MultiCloudWorkspaceService.constructor(
         slickDataSource,
         workspaceManagerDAO,
+        billingProfileManagerClientProvider,
         samDAO,
         multiCloudWorkspaceConfig,
         metricsPrefix
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -287,10 +287,9 @@ object Boot extends IOApp with LazyLogging {
       val servicePerimeterService = new ServicePerimeterService(slickDataSource, gcsDAO, servicePerimeterConfig)
 
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
-      val billingProfileManagerClientProvider = new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl"))
       val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
         samDAO,
-        billingProfileManagerClientProvider,
+        new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl")),
         multiCloudWorkspaceConfig
       )
 
@@ -399,7 +398,7 @@ object Boot extends IOApp with LazyLogging {
       val multiCloudWorkspaceServiceConstructor: (UserInfo) => MultiCloudWorkspaceService = MultiCloudWorkspaceService.constructor(
         slickDataSource,
         workspaceManagerDAO,
-        billingProfileManagerClientProvider,
+        billingProfileManagerDAO,
         samDAO,
         multiCloudWorkspaceConfig,
         metricsPrefix
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -17,7 +17,10 @@ import scala.jdk.CollectionConverters._
 trait BillingProfileManagerDAO {
   def createBillingProfile(displayName: String, billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates], userInfo: UserInfo): Future[ProfileModel]
 
-  def listBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+  /***
+    * Uses BPM to retrieve billing profile information for each of the RawlsBillingProjects (via their billingProjectIDs).
+    */
+  def populateBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
 
   def listManagedApps(subscriptionId: UUID, userInfo: UserInfo): Future[Seq[AzureManagedAppModel]]
 }
@@ -69,11 +72,9 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
 
 
   /**
-   * Fetches the billing profiles to which the user has access.
-   *
-   * This method only returns Azure billing profiles for now
+   * Fetches the details of the Azure billing profiles, based on the input rawlsBillingProjects.
    */
-  def listBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
+  def populateBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
     if (!config.multiCloudWorkspacesEnabled) {
       return Future.successful(Seq())
     }
@@ -86,7 +87,7 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
     }
 
     for {
-      billingProfiles <- getAllBillingProfiles(azureConfig, userInfo)
+      billingProfiles <- getAllBillingProfiles(azureConfig, userInfo, rawlsBillingProjects)
     } yield {
       billingProfiles.filter {
         bp => samUserResources.map(_.resourceId).contains(bp.projectName.value)
@@ -94,7 +95,7 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
     }
   }
 
-  private def getAllBillingProfiles(azureConfig: AzureConfig, userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
+  private def getAllBillingProfiles(azureConfig: AzureConfig, userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
     // NB until the BPM is live, we are returning a hardcoded
     // Azure billing profile, with access enforced by SAM
     samDAO.userHasAction(
@@ -104,23 +105,51 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
       userInfo
     ).flatMap {
       case true =>
-        Future.successful(
-          Seq(
+        val profileApi = apiClientProvider.getProfileApi(userInfo.accessToken.token)
+        val fullBillingProjects = rawlsBillingProjects map {
+          project =>
+            // TODO: throw exception if ID doesn't exist (already filtering for this, so is it necessary?)
+            val profileModel = profileApi.getProfile(UUID.fromString(project.billingProfileId.get))
             RawlsBillingProject(
-              RawlsBillingProjectName(azureConfig.billingProjectName),
-              CreationStatuses.Ready,
-              None,
-              None,
+              project.projectName,
+              project.status,
+              project.billingAccount,
+              project.message,
+              project.cromwellBackend,
+              project.servicePerimeter,
+              project.googleProjectNumber,
+              project.invalidBillingAccount,
+              project.spendReportDataset,
+              project.spendReportTable,
+              project.spendReportDatasetGoogleProject,
               azureManagedAppCoordinates = Some(
                 AzureManagedAppCoordinates(
-                  UUID.fromString(azureConfig.azureTenantId),
-                  UUID.fromString(azureConfig.azureSubscriptionId),
-                  azureConfig.azureResourceGroupId
+                  profileModel.getTenantId,
+                  profileModel.getSubscriptionId,
+                  profileModel.getManagedResourceGroupId
                 )
+              ),
+              billingProfileId = project.billingProfileId
+            )
+        }
+
+        // Will remove after users can create Azure-backed Billing Accounts via Terra.
+        val temporaryAccount = Seq(RawlsBillingProject(
+            RawlsBillingProjectName(azureConfig.billingProjectName),
+            CreationStatuses.Ready,
+            None,
+            None,
+            azureManagedAppCoordinates = Some(
+              AzureManagedAppCoordinates(
+                UUID.fromString(azureConfig.azureTenantId),
+                UUID.fromString(azureConfig.azureSubscriptionId),
+                azureConfig.azureResourceGroupId
               )
             )
           )
         )
+        Future.successful(fullBillingProjects ++ temporaryAccount)
+
       case false =>
         Future.successful(Seq.empty)
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -134,7 +134,7 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
         }
 
         // Will remove after users can create Azure-backed Billing Accounts via Terra.
-        val temporaryAccount = Seq(RawlsBillingProject(
+        val temporaryAzureBillingProject = Seq(RawlsBillingProject(
             RawlsBillingProjectName(azureConfig.billingProjectName),
             CreationStatuses.Ready,
             None,
@@ -148,7 +148,7 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
             )
           )
         )
-        Future.successful(fullBillingProjects ++ temporaryAccount)
+        Future.successful(fullBillingProjects ++ temporaryAzureBillingProject)
 
       case false =>
         Future.successful(Seq.empty)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -23,6 +23,8 @@ trait BillingProfileManagerDAO {
   def populateBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
 
   def listManagedApps(subscriptionId: UUID, userInfo: UserInfo): Future[Seq[AzureManagedAppModel]]
+
+  def getBillingProfile(billingProfileId: UUID, userInfo: UserInfo): ProfileModel
 }
 
 
@@ -93,6 +95,10 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
         bp => samUserResources.map(_.resourceId).contains(bp.projectName.value)
       }
     }
+  }
+
+  def getBillingProfile(billingProfileId: UUID, userInfo: UserInfo): ProfileModel = {
+    apiClientProvider.getProfileApi(userInfo.accessToken.token).getProfile(billingProfileId)
   }
 
   private def getAllBillingProfiles(azureConfig: AzureConfig, userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.rawls.billing
 import bio.terra.profile.model.{AzureManagedAppModel, CloudPlatform, CreateProfileRequest, ProfileModel}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
-import org.broadinstitute.dsde.rawls.config.{AzureConfig, MultiCloudWorkspaceConfig}
+import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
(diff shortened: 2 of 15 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -398,6 +399,7 @@ object Boot extends IOApp with LazyLogging {
       val multiCloudWorkspaceServiceConstructor: (UserInfo) => MultiCloudWorkspaceService = MultiCloudWorkspaceService.constructor(
         slickDataSource,
         workspaceManagerDAO,
+        billingProfileManagerClientProvider,
```
cahrens, line 402: I wasn't sure if we want to pass `billingProfileManagerDAO` instead of just the client. I need the client to be able to look up the profile during workspace creation.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -17,7 +17,10 @@ import scala.jdk.CollectionConverters._
 trait BillingProfileManagerDAO {
   def createBillingProfile(displayName: String, billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates], userInfo: UserInfo): Future[ProfileModel]
 
-  def listBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+  /***
+    * Uses BPM to retrieve billing profile information for each of the RawlsBillingProjects (via their billingProjectIDs).
+    */
+  def populateBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
```
cahrens: This method was originally documented as returning all the profiles that a user has access to. However, I really only need to look up the profiles for the Rawls projects the user has access to, using the stored billingProjectIDs.

I suppose the alternative is to return all profiles, and then match them up with the rawlsBillingProjects after the fact. But it's easier to include the fake profile with this implementation.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -104,23 +105,51 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
     ).flatMap {
       case true =>
-        Future.successful(
-          Seq(
+        val profileApi = apiClientProvider.getProfileApi(userInfo.accessToken.token)
+        val fullBillingProjects = rawlsBillingProjects map {
+          project =>
+            // TODO: throw exception if ID doesn't exist (already filtering for this, so is it necessary?)
+            val profileModel = profileApi.getProfile(UUID.fromString(project.billingProfileId.get))
             RawlsBillingProject(
```
cahrens: Is there an easier way to clone a `RawlsBillingProject` in order to add the `azureManagedAppCoordinates` information?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -75,7 +75,8 @@ case class RawlsBillingProjectResponse(projectName: RawlsBillingProjectName,
                                        roles: Set[ProjectRoles.ProjectRole],
                                        status: CreationStatuses.CreationStatus,
                                        message: Option[String],
-                                       managedAppCoordinates: Option[AzureManagedAppCoordinates]
+                                       managedAppCoordinates: Option[AzureManagedAppCoordinates], // TODO: do we even want to return this?
```
cahrens, line 78: The ticket requested that we change the API to provide the cloud platform of a billing project in a clearer way than looking for the presence of `managedAppCoordinates`. I am wondering if we even want to return `managedAppCoordinates` at all, as we don't pass it to the server when using the billing project.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -75,7 +75,8 @@ case class RawlsBillingProjectResponse(projectName: RawlsBillingProjectName,
...
+                                       cloudPlatform: String
```
cahrens, line 79: My first thought was to make this of type `bio.terra.profile.model.CloudPlatform` (which is an enum), but I had trouble with the JSON serialization format declaration for that.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -161,7 +162,7 @@ class UserService(protected val userInfo: UserInfo,
     }
   }
 
-  private def makeBillingProjectResponse(projectRoles: Set[ProjectRole], billingProject: RawlsBillingProject) =
+  def makeBillingProjectResponse(projectRoles: Set[ProjectRole], billingProject: RawlsBillingProject) =
```
cahrens: I wanted to be able to use this for tests (note though that I haven't attempted to get tests passing yet).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -112,14 +147,12 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
   private def createWorkspace(workspaceRequest: MultiCloudWorkspaceRequest, parentSpan: Span): Future[Workspace] = {
-    val azureConfig = multiCloudWorkspaceConfig.azureConfig.getOrElse(throw new RawlsException("Azure config not present"))
     val wsmConfig = multiCloudWorkspaceConfig.workspaceManager.getOrElse(throw new RawlsException("WSM app config not present"))
 
-    // TODO these will come from the spend profile service in the future
-    val spendProfileId = azureConfig.spendProfileId
-    val azureTenantId = azureConfig.azureTenantId
-    val azureSubscriptionId = azureConfig.azureSubscriptionId
-    val azureResourceGroupId = azureConfig.azureResourceGroupId
+    val spendProfileId = workspaceRequest.billingProfileID
```
cahrens, line 152: Hopefully `spendProfileId` is the same thing as the billingProfileID.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2380,6 +2380,18 @@ class WorkspaceService(protected val userInfo: UserInfo,
     } yield result
   }
 
+  def getBillingProfileId(billingProjectName: String): Future[Option[String]] = {
```
cahrens: I thought of using `withBillingProjectContext`, which is at line 2367. However, that doesn't work on Azure billing projects because of `failUnlessBillingAccountHasAccess(billingProject, parentSpan)`.

I do wonder though if we want to thread a span through all of this for logging.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -949,7 +951,9 @@ class WorkspaceJsonSupport extends JsonSupport {
 
   implicit val workspaceCloudPlatformFormat = rawlsEnumerationFormat(WorkspaceCloudPlatform.withName)
 
-  implicit val MultiCloudWorkspaceRequestFormat = jsonFormat5(MultiCloudWorkspaceRequest)
+  implicit val AzureManagedAppCoordinatesFormat = jsonFormat3(AzureManagedAppCoordinates)
```
cahrens, line 954: Renamed for clarity.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -112,14 +147,12 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
   private def createWorkspace(workspaceRequest: MultiCloudWorkspaceRequest, parentSpan: Span): Future[Workspace] = {
-    val azureConfig = multiCloudWorkspaceConfig.azureConfig.getOrElse(throw new RawlsException("Azure config not present"))
     val wsmConfig = multiCloudWorkspaceConfig.workspaceManager.getOrElse(throw new RawlsException("WSM app config not present"))
 
-    // TODO these will come from the spend profile service in the future
-    val spendProfileId = azureConfig.spendProfileId
-    val azureTenantId = azureConfig.azureTenantId
-    val azureSubscriptionId = azureConfig.azureSubscriptionId
-    val azureResourceGroupId = azureConfig.azureResourceGroupId
+    val spendProfileId = workspaceRequest.billingProfileId
```
cahrens, line 152: Hopefully spendProfileId is the same thing as the billingProfileId.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -287,9 +287,10 @@ object Boot extends IOApp with LazyLogging {
       val servicePerimeterService = new ServicePerimeterService(slickDataSource, gcsDAO, servicePerimeterConfig)
 
       val multiCloudWorkspaceConfig = MultiCloudWorkspaceConfig.apply(conf)
+      val billingProfileManagerClientProvider = new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl"))
       val billingProfileManagerDAO = new BillingProfileManagerDAOImpl(
         samDAO,
-        new HttpBillingProfileManagerClientProvider(conf.getStringOption("billingProfileManager.baseUrl")),
+        billingProfileManagerClientProvider,
```
cahrens: I wasn't sure if we want to pass billingProfileManagerDAO instead of just the client. I need the client to be able to look up the profile during workspace creation.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
aherbst-broad: Case classes have a `copy` method that may be suitable.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
aherbst-broad: Yeah if it's not needed for UI consumption I am +1 on removing it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aherbst-broad: It should be.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
aherbst-broad: We should pass the DAO rather than the raw client and add a `getProfile` method, to avoid breaking the abstraction.  That is, BillingProfileManagerDAO should really be the only part of our code that knows about the HTTP client providers.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -86,15 +87,15 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
+      billingProfiles <- getAllBillingProfiles(azureConfig, userInfo, rawlsBillingProjects)
     } yield {
       billingProfiles.filter {
         bp => samUserResources.map(_.resourceId).contains(bp.projectName.value)
       }
     }
   }
 
-  private def getAllBillingProfiles(azureConfig: AzureConfig, userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
+  private def getAllBillingProfiles(azureConfig: AzureConfig, userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
```
aherbst-broad: I think I fouled up the abstraction here during the initial implementation of this DAO. I think this class's sole concern should be BPM model objects (i.e, Profile) and should have little (no?) knowledge of `RawlsBililngProject`s. I think this stitching might better live in the BillingProjectOrchestrator class.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: Agreed about adding a span. Maybe we make the `failUnlessBillingAccountHasAccess` resilient to non-gcp projects? I'm fine scoping that out for this effort though.

aherbst-broad (commented):
Looking good, had some initial comments.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: I think most people are too quick to declare class members private, for exactly this reason.

cahrens: I ended up moving it to the case class so you can call it statically.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
cahrens: This is the only place it's used in terra-ui, and it would be good to replace it with the more explicit cloudContext variable. We will have to temporarily make terra-ui handle either one though.

```
const isAzureBillingProject = project => {
    if (project === undefined) {
      project = _.find({ projectName: namespace }, billingProjects)
    }
    // Azure billing projects have `managedAppCoordinates` defined.
    return !!project?.managedAppCoordinates
  }

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala (hunk 8 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala]: We could potentially change `withBillingProjectContext` to only call `failUnlessBillingAccountHasAccess` if `billingProfileId` is not defined.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala (hunk 2 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala]: @aherbst-broad what was your thought about this? Do you think we should keep this method as returning all billing profiles the user has access to?

File: core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala (hunk 12 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala]: I'm trying to understand the scope of what you think might move into `BillingProjectOrchestrator`, which currently only is used for billing project creation. You are suggesting moving `listBillingProjectsV2` to the orchestrator?

File: core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala (hunk 11 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala]: That is what I suspected. I have pushed up the change, with a question.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala [hunk 13]
```diff
@@ -17,9 +17,14 @@ import scala.jdk.CollectionConverters._
...
 
-  def listBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+  /***
+    * Uses BPM to retrieve billing profile information for each of the RawlsBillingProjects (via their billingProjectIDs).
+    */
+  def populateBillingProfiles(samUserResources: Seq[SamUserResource], userInfo: UserInfo, rawlsBillingProjects: Seq[RawlsBillingProject])(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
 
   def listManagedApps(subscriptionId: UUID, userInfo: UserInfo): Future[Seq[AzureManagedAppModel]]
+
+  def getBillingProfile(billingProfileId: UUID, userInfo: UserInfo): ProfileModel
```
cahrens, line 22: I noticed that `createBillingProfile` returns `Future[ProfileModel]`, but I don't understand why. I would imagine that we want both methods to return the same thing, but adding in the `Future` seems unnecessarily complex to me.

@aherbst-broad what was your thinking with wrapping the return value in a `Future.successful`?

cahrens: It also doesn't look like `listManagedApps` actually needs to return a Future… I feel like I'm missing something.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: @aherbst-broad I went back to returning all billing profileModels the user has access to in the commit I just pushed up.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: Yep, copy was what I needed.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: @aherbst-broad I made an attempt at switching the DAO to return BPM model objects. The only issue I ran into is that for the hard-coded Azure billing project, I really need to return a RawlsBillingProject because I need to include the Rawls billing project name in the return value. Therefore I created a specific method for getting that hard-coded account, and we will just delete that method when we no longer need the hack.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -85,44 +87,62 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-              None,
-              None,
-              azureManagedAppCoordinates = Some(
-                AzureManagedAppCoordinates(
-                  UUID.fromString(azureConfig.azureTenantId),
-                  UUID.fromString(azureConfig.azureSubscriptionId),
-                  azureConfig.azureResourceGroupId
-                )
+        val profileApi = apiClientProvider.getProfileApi(userInfo.accessToken.token)
+        Future.successful(profileApi.listProfiles(0, 1000).getItems.asScala.toSeq)
```
cahrens, line 99: Note that we have provide a limit to the number returned. To make the code robust we should check the number returned and continue to fetch profiles until we get them all. Not bothering with that though until we decide if we want to go with this approach (as opposed to fetching each profile individually by its `billlingProfileId`).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -85,44 +87,62 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-        )
-      case false =>
-        Future.successful(Seq.empty)
+          )))
+        case false =>
+          Future.successful(Seq.empty)
+      }
+    } yield {
+      billingProjects.filter {
+        bp => samUserResources.map(_.resourceId).contains(bp.projectName.value)
```
cahrens, line 144: This filtering was in the original `listBillingProfiles` method, but I believe it is only needed for the hardcoded account because of the code at https://github.com/broadinstitute/rawls/pull/1994/files#diff-727d553ef55a7aabb5b0d0bca2e6a6fe44422feef4c7ad5ebb9c86bcd92b6572L179.

It also seems good that with the exception of this temporary hack method, we are no longer passing Sam resources into the DAO methods.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: (I didn't move anything to `BillingProjectOrchestrator` though)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -76,19 +80,50 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
+            UUID.fromString(azureConfig.azureTenantId),
+            UUID.fromString(azureConfig.azureSubscriptionId),
+            azureConfig.azureResourceGroupId
+          ),
+          azureConfig.spendProfileId
         )
       )
     } else {
-      workspaceService.createWorkspace(workspaceRequest, parentSpan)
+      traceWithParent("withBillingProjectContext", parentSpan)(
```
cahrens, line 102: @aherbst-broad switched to using ` workspaceService.withBillingProjectContext` and passed in `parentSpan`. Please check that I did the span stuff correctly.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -76,19 +80,50 @@ class MultiCloudWorkspaceService(userInfo: UserInfo,
...
+                  workspaceRequest.attributes,
+                  WorkspaceCloudPlatform.Azure,
+                  azureConfig.defaultRegion,
+                  AzureManagedAppCoordinates(
+                    profileModel.getTenantId,
+                    profileModel.getSubscriptionId,
+                    profileModel.getManagedResourceGroupId
+                  ),
+                  billingProfileID
+                ), childSpan
```
cahrens, line 122: Passed `childSpan` here.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2375,7 +2375,11 @@ class WorkspaceService(protected val userInfo: UserInfo,
 
       billingProject = maybeBillingProject.getOrElse(throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.BadRequest, s"Billing Project ${billingProjectName} does not exist")))
       _ <- failUnlessBillingProjectReady(billingProject)
-      _ <- failUnlessBillingAccountHasAccess(billingProject, parentSpan)
+      _ <- billingProject.billingProfileId match {
+        // If `billingProfileId` is defined, this project is backed by a BPM record, and we don't need to check access.
+        case Some(_) => Future.successful()
+        case _ => failUnlessBillingAccountHasAccess(billingProject, parentSpan)
```
cahrens, line 2381: Exposed this method and modified it to work with both GCP and Azure billing projects.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -85,44 +87,62 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-                  UUID.fromString(azureConfig.azureSubscriptionId),
-                  azureConfig.azureResourceGroupId
-                )
+        val profileApi = apiClientProvider.getProfileApi(userInfo.accessToken.token)
+        Future.successful(profileApi.listProfiles(0, 1000).getItems.asScala.toSeq)
+      case _ => Future.successful(Seq())
+    }
+  }
+
+  // Code copied from above, but method will be deleted.
```
cahrens, line 104: I have created https://broadworkbench.atlassian.net/browse/WOR-486 to delete all this code associated with the hardcoded account.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: There really isn't a reason to use a future, either here, or in `createBillingProfile`. Both of them are just waiting synchronously, and then wrapping the result in a future.

If we were concerned about the execution time, we could return something like: 
`Future(apiClientProvider.getProfileApi(userInfo.accessToken.token).getProfile(billingProfileId))`
Which would return the result of `apiClientProvider.getProfileApi(userInfo.accessToken.token).getProfile(billingProfileId)` asynchronously.

But that only makes sense if you're also doing something else while waiting for the operation to complete (such as making multiple async calls, and waiting for them all), or returning a future from the call site (where you would use `.map` or `.flatMap` for further operations on the result.

In this case, there's no reason to use a Future that I can see.
It's plausible that `createBillingProfile` was doing something in the past that merited it, but it isn't anymore. I suggest we remove the future from both.

aherbst-broad (commented):
Left some comments last week, please re-ping when ready for further review.

cahrens:
Closing in favor of #2010

PR #1997, opened 2022-08-19
Aj 522 remove empty columns

calypsomatic:
[AJ-522](https://broadworkbench.atlassian.net/browse/AJ-522)
This PR adds a `removeEmptyColumns` flag to the `Submission` object that defaults to `true` (although I have set it to `false` in most of the tests to avoid breaking them).  When this flag is `true`, empty outputs generated by a submission are removed, so that no new columns are introduced unless they would not be empty.  
A follow-on PR will add the option to toggle this flag in the UI.


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220817_add_remove_empty_columns.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet id="add_empty_columns" author="bmlrga " logicalFilePath="dummy">
```
davidangb: bmlrga?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220817_add_remove_empty_columns.xml
```diff
@@ -0,0 +1,10 @@
...
+        <addColumn tableName="SUBMISSION">
+            <column name="REMOVE_EMPTY_COLUMNS" type="BOOLEAN" defaultValue="true"/>
```
davidangb: 1. why not make this nullable? That would allow all the old submissions to be null, indicating that they ran before this flag was available.
2. My $.02 is to name this something like "ignore empty optional outputs" - I'm a bit hesitant to limit it to columns. This would also come into play if the output was writing to workspace attributes.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4044,6 +4044,9 @@ components:
         userComment:
           type: string
           description: Comment for the submission (max length 1000 characters)
+        removeEmptyColumns:
+          type: boolean
+          description: Whether or not to create output columns if they would be empty.  Defaults to true (no columns created).
```
davidangb: add a `default:` key to specify the default.

I defer to Product on this, but I suggest making the default the opposite - by default, it SHOULD create the empty columns. This will preserve current behavior - and therefore risk no compatibility issues.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -391,6 +394,14 @@ trait SubmissionComponent {
       query.result.map(_.toMap)
     }
 
+    def getEmptyColumnParam(submissionId: UUID) = {
```
davidangb: I think you should put the `ReadAction[Boolean]` result type back in - my suggestion to remove that was just to ease debugging while we were pairing on this

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -391,6 +394,14 @@ trait SubmissionComponent {
...
+      val query = for {
+        submission <- submissionQuery if submission.id === submissionId
+      } yield submission.removeEmptyColumns
```
davidangb: do you know what SQL this ultimately generates? If not, that's a good debugging tool to learn (see https://scala-slick.org/doc/3.3.3/config.html and https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/logback.xml#L102).

Another way to write this is `submissionQuery.filter(_.id === submissionId).map(_.removeEmptyColumns)`, in case it ends up generating more optimal sql.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -521,7 +538,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       }.toSeq
 
       if (parsedExpressions.forall(_.isSuccess)) {
-        val boundExpressions: Seq[BoundOutputExpression] = parsedExpressions.collect { case Success(boe @ BoundOutputExpression(target, name, attr)) => boe }
+        val boundExpressions: Seq[BoundOutputExpression] = parsedExpressions.collect { case Success(boe @ BoundOutputExpression(target, name, attr)) if !(attributeIsEmpty(attr) && removeEmptyColumns) => boe }
```
davidangb: I'm struggling with this. I don't think this is right, but I don't know what to suggest that is better. I will continue to look at the code and try to come up with a more productive comment.

As it stands, this is looking at the output's *value* to determine whether or not it is empty. In some of those cases like AttributeString(""), it feels like there are valid use cases for saving those as values.

This logic should be looking at the WDL output's definition, to verify if that output is marked as optional. Only if the output is marked as optional should the create/not-create logic come into play. But I don't know where to access that info … I will keep looking.

davidangb (commented):
some interim comments, I want to keep digging on this to understand it better but also want to get you early feedback!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220817_add_remove_empty_columns.xml
calypsomatic: 😕 no idea what happened there!

core/src/main/resources/swagger/api-docs.yaml
calypsomatic: I was going with the ticket's description that in the UI the checkbox should be checked, but now that you mention it there's no reason why the backend should have that as the default.  Will ask.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
calypsomatic: I did struggle to find where/how to identify the empty columns, so I'm very open to suggestions.  Can also spend some time poking around on my own.

davidangb: here is some starter code for inspecting the output definitions: https://github.com/broadinstitute/rawls/blob/874b2e1bc6d95f90e1cf0db7569cfbaae9da34bf/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/MethodConfigResolver.scala#L59-L69 - `parseWDL` interprets the WDL and reads its outputs, `WorkflowDescription.getOutputs` will return a list of `ToolOutputParameter`, and `ToolOutputParameter` has a `getValueType` which will indicate if it is optional or not. Having said that, `parseWDL` is a heavyweight method that makes a network call to Cromiam, so we need to be careful with it.

jordansafer:
Note Jenkins tests for Rawls are consistently failing, under investigation and this may be related to rolling out BEEs

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -432,6 +435,12 @@ trait SubmissionComponent {
       query.result.map(_.toMap)
     }
 
+    def getEmptyOutputParam(submissionId: UUID): ReadAction[Boolean] = {
+      val query = submissionQuery.filter(_.id === submissionId).map(_.ignoreEmptyOutputs)
+
+      query.result.head
+    }
```
davidangb, line 442: for older submissions that have a null for the `IGNORE_EMPTY_OUTPUTS` column, I _think_ this will throw an error trying to cast null into Boolean. Even if it doesn't, consider using the `uniqueResult` helper function plus a `getOrElse` to set the default:

```scala
    def getEmptyOutputParam(submissionId: UUID): ReadAction[Boolean] =
      uniqueResult[Boolean](submissionQuery.filter(_.id === submissionId).map(_.ignoreEmptyOutputs))
        .map(_.getOrElse(false))
```
or, return an `Option[Boolean]` and let calling code figure out what to do if it returns null/None:
```scala
    def getEmptyOutputParam(submissionId: UUID): ReadAction[Option[Boolean]] =
      uniqueResult[Boolean](submissionQuery.filter(_.id === submissionId).map(_.ignoreEmptyOutputs))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -603,10 +613,18 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }
   }
 
+  private def attributeIsEmpty(attribute: Attribute): Boolean =
+    attribute match {
+      case AttributeNull          => true
+      case AttributeString(value) => value == ""
+      case _                      => false
+    }
```
davidangb, line 621: to be Scala-idiomatic, include the "if" clause in the `case` definition:
```scala
  private def attributeIsEmpty(attribute: Attribute): Boolean =
    attribute match {
      case AttributeNull                         => true
      case AttributeString(value) if value == "" => true
      case _                                     => false
    }
```
I'm personally fine with either approach, just bringing it up for discussion/your choice

davidangb (commented):
2 comments inline, this is really close! Some additional thoughts:
* this will definitely need good/detailed release notes
* I still wonder if we need separate settings for "ignore nulls" and "ignore empty strings" … and I realize that would be a significant refactor. I think we should ship this as-is and see if we get any of that feedback … so, this is just me thinking out loud.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
jordansafer: Makes sense! Using uniqueResult

jordansafer:
Jenkins retest

davidangb (approved):
:+1: once tests pass!

jordansafer:
jenkins retest

jordansafer:
jenkins retest

jordansafer (approved):
Approving on behalf of @calypsomatic because the github UI isn't allowing the initial author to review after the PR work changed hands-

jordansafer:
jenkins retest

PR #2001, opened 2022-08-24
only send 5xx WSM ApiException errors to Sentry [AJ-560]

davidangb:
When Rawls encounters an `ApiException` from talking to Workspace Manager, we previously sent all exceptions to Sentry.

This caused spam - for instance, if a caller repeatedly attempted to add the same snapshot reference and got a 409, we sent all those 409s to Sentry.

**_REVIEWER:_** ignore whitespace and/or review commit-by-commit; there's a lot of lines changed purely due to `scalafmt`. The first commit has the single functional change.


aherbst-broad (approved):
lgtm

calypsomatic (approved):

PR #2002, opened 2022-08-25
aj-525 clearer entity name validation message

calypsomatic:
AJ-525 Error message is unclear when importing a snapshot with invalid entity name. This gives a more specific error message (although it is still a little ugly, wrapped in `ErrorReport`).


model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala
```diff
@@ -10,79 +10,97 @@ trait StringValidationUtils {
...
 
   private lazy val entityNameRegex = "[A-z0-9\\._-]+".r
-  def validateEntityName(s: String): Unit = {
-    if(! entityNameRegex.pattern.matcher(s).matches) {
-      val msg = s"Invalid input: $s. Input may only contain alphanumeric characters, underscores, dashes, and periods."
-      throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = msg, statusCode = StatusCodes.BadRequest))
+  def validateEntityName(s: String): Unit =
+    if (!entityNameRegex.pattern.matcher(s).matches) {
+      val msg =
+        s"Invalid name for Terra table: $s. Input may only contain alphanumeric characters, underscores, dashes, and periods."
```
calypsomatic: This is the only actual change; the rest is just whitespace formatting.

model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala
```diff
@@ -10,79 +10,97 @@ trait StringValidationUtils {
...
 
   private lazy val entityNameRegex = "[A-z0-9\\._-]+".r
-  def validateEntityName(s: String): Unit = {
-    if(! entityNameRegex.pattern.matcher(s).matches) {
-      val msg = s"Invalid input: $s. Input may only contain alphanumeric characters, underscores, dashes, and periods."
-      throw new RawlsExceptionWithErrorReport(errorReport = ErrorReport(message = msg, statusCode = StatusCodes.BadRequest))
+  def validateEntityName(s: String): Unit =
+    if (!entityNameRegex.pattern.matcher(s).matches) {
+      val msg =
+        s"Invalid entity name: $s. Input may only contain alphanumeric characters, underscores, dashes, and periods. If importing a snapshot, making an API call, uploading a TSV, or adding a new row from the Terra UI, please modify the impacted name(s) and try again."
```
davidangb: you don't have to hold the PR for this, but for the future: use three double quotes to support multiline strings for readability, like:
```scala
      val msg =
        s"""Invalid entity name: $s. Input may only contain alphanumeric characters, underscores, dashes, and periods. 
           |If importing a snapshot, making an API call, uploading a TSV, or adding a new row from the Terra UI, 
           |please modify the impacted name(s) and try again.""".stripMargin
```

davidangb (approved):

model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala
calypsomatic: Thank you!  That was bothering me.  I'm gonna fix it.

PR #2004, opened 2022-08-26
result of scalafmtAll without HEAD filter [WOR-180]

davidangb:
Following up on #2003, which was insufficient … this PR runs `scalafmt` on ALL files, ignoring the `"diff-ref=HEAD^"` filter in `Settings.scala`. This should bring the entire codebase up to spec with the formatting rules, preventing formatting changes from sneaking into other PRs.


calypsomatic (approved):
I would rather have one throwaway PR like this, even if it messes up history, to enable easier PRs in the future - otherwise every time someone submits a PR that changes an unformatted file, their changes would be lost in the formatting changes.

blakery (approved):
Might as well rip the band-aid off

PR #2010, opened 2022-08-29
[WOR-400] Include billing projects from BPM.

blakery:
from https://github.com/broadinstitute/rawls/pull/1994/files

This required a fair number of manual fixes, where actual underlying code had changed, as well as formatting.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.rawls.billing
 import bio.terra.profile.model.{AzureManagedAppModel, CloudPlatform, CreateProfileRequest, ProfileModel}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
-import org.broadinstitute.dsde.rawls.config.{AzureConfig, MultiCloudWorkspaceConfig}
+import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
 import org.broadinstitute.dsde.rawls.model.{
   AzureManagedAppCoordinates,
@@ -39,7 +39,9 @@ trait BillingProfileManagerDAO {
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]]
 
   // This is a temporary method that will be deleted once users can create their own Azure-backed billing projects in Terra.
-  def getHardcodedAzureBillingProject(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]]
+  def getHardcodedAzureBillingProject(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit
+    ec: ExecutionContext
+  ): Future[Seq[RawlsBillingProject]]
 
 }
 
@@ -50,9 +52,10 @@ class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExcepti
  * billing profiles in the Terra system. For now, we are using this to layer in "external" billing profiles
  * for the purposes of testing Azure workspaces.
  */
-class BillingProfileManagerDAOImpl(samDAO: SamDAO,
-                                   apiClientProvider: BillingProfileManagerClientProvider,
-                                   config: MultiCloudWorkspaceConfig
+class BillingProfileManagerDAOImpl(
+  samDAO: SamDAO,
+  apiClientProvider: BillingProfileManagerClientProvider,
+  config: MultiCloudWorkspaceConfig
 ) extends BillingProfileManagerDAO
     with LazyLogging {
 
@@ -63,9 +66,10 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
     Future.successful(result)
   }
 
-  override def createBillingProfile(displayName: String,
-                                    billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates],
-                                    ctx: RawlsRequestContext
+  override def createBillingProfile(
+    displayName: String,
+    billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates],
+    ctx: RawlsRequestContext
   ): Future[ProfileModel] = {
     val azureManagedAppCoordinates = billingInfo match {
       case Left(_)       => throw new NotImplementedError("Google billing accounts not supported in billing profiles")
@@ -89,11 +93,8 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
     Future.successful(createdProfile)
   }
 
-
-  def getBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): ProfileModel = {
+  def getBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): ProfileModel =
     apiClientProvider.getProfileApi(ctx).getProfile(billingProfileId)
-  }
-
 
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]] = {
 
@@ -124,7 +125,9 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
       }
   }
 
-  def getHardcodedAzureBillingProject(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit ec: ExecutionContext): Future[Seq[RawlsBillingProject]] = {
+  def getHardcodedAzureBillingProject(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit
+    ec: ExecutionContext
+  ): Future[Seq[RawlsBillingProject]] = {
     if (!config.multiCloudWorkspacesEnabled) {
       return Future.successful(Seq())
     }
@@ -137,32 +140,38 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
     }
 
     for {
-      billingProjects <- samDAO.userHasAction(
-        SamResourceTypeNames.managedGroup,
-        azureConfig.alphaFeatureGroup,
-        SamResourceAction("use"),
-        userInfo
-      ).flatMap {
-        case true =>
-          // Will remove after users can create Azure-backed Billing Accounts via Terra.
-          Future.successful(Seq(RawlsBillingProject(
-            RawlsBillingProjectName(azureConfig.billingProjectName),
-            CreationStatuses.Ready,
-            None,
-            None,
-            azureManagedAppCoordinates = Some(
-              AzureManagedAppCoordinates(
-                UUID.fromString(azureConfig.azureTenantId),
-                UUID.fromString(azureConfig.azureSubscriptionId),
-                azureConfig.azureResourceGroupId
-              )))))
-        case false =>
-          Future.successful(Seq.empty)
-      }
-    } yield {
-      billingProjects.filter {
-        bp => samUserResources.map(_.resourceId).contains(bp.projectName.value)
-      }
+      billingProjects <- samDAO
+        .userHasAction(
+          SamResourceTypeNames.managedGroup,
+          azureConfig.alphaFeatureGroup,
+          SamResourceAction("use"),
+          userInfo
+        )
+        .flatMap {
+          case true =>
+            // Will remove after users can create Azure-backed Billing Accounts via Terra.
+            Future.successful(
+              Seq(
+                RawlsBillingProject(
+                  RawlsBillingProjectName(azureConfig.billingProjectName),
+                  CreationStatuses.Ready,
+                  None,
+                  None,
+                  azureManagedAppCoordinates = Some(
+                    AzureManagedAppCoordinates(
+                      UUID.fromString(azureConfig.azureTenantId),
+                      UUID.fromString(azureConfig.azureSubscriptionId),
+                      azureConfig.azureResourceGroupId
+                    )
+                  )
+                )
+              )
+            )
+          case false =>
+            Future.successful(Seq.empty)
+        }
+    } yield billingProjects.filter { bp =>
+      samUserResources.map(_.resourceId).contains(bp.projectName.value)
     }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -19,6 +19,7 @@ import org.broadinstitute.dsde.rawls.model.{
 }
 
 import java.util.UUID
+import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 
@@ -108,6 +109,17 @@ class BillingProfileManagerDAOImpl(
       case Some(value) => value
     }
 
+    @tailrec
+    def callListProfiles(accumulator: Seq[ProfileModel] = Seq.empty, offset: Int = 0): Seq[ProfileModel] = {
+      val profileApi = apiClientProvider.getProfileApi(ctx)
+      val response = profileApi.listProfiles(offset, 1000)
+      if (response.getTotal >= 1000) {
+        callListProfiles(accumulator ++ response.getItems.asScala.toSeq, offset + response.getTotal)
+      } else {
+        accumulator ++ response.getItems.asScala.toSeq
+      }
+    }
+
     // NB until the BPM is live, we want to ensure user is in the alpha group
     samDAO
       .userHasAction(
@@ -117,10 +129,8 @@ class BillingProfileManagerDAOImpl(
         ctx.userInfo
       )
       .flatMap {
-        case true =>
-          val profileApi = apiClientProvider.getProfileApi(ctx)
-          Future.successful(profileApi.listProfiles(0, 1000).getItems.asScala.toSeq)
-        case _ => Future.successful(Seq())
+        case true => Future.successful(callListProfiles())
+        case _    => Future.successful(Seq())
       }
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -45,7 +45,13 @@ trait BillingProfileManagerDAO {
 
 }
 
+
 class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
+
+
+object BillingProfileManagerDAO {
+  val BillingProfileRequestBatchSize = 1000
+}
 
 /**
  * Facade over the billing profile manager service. This service will eventually by the source of truth for
@@ -96,6 +102,7 @@ class BillingProfileManagerDAOImpl(
   def getBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Option[ProfileModel] =
     Option(apiClientProvider.getProfileApi(ctx).getProfile(billingProfileId))
 
+
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]] = {
 
     if (!config.multiCloudWorkspacesEnabled) {
@@ -109,11 +116,13 @@ class BillingProfileManagerDAOImpl(
       case Some(value) => value
     }
 
+    val profileApi = apiClientProvider.getProfileApi(ctx)
+
     @tailrec
     def callListProfiles(accumulator: Seq[ProfileModel] = Seq.empty, offset: Int = 0): Seq[ProfileModel] = {
-      val profileApi = apiClientProvider.getProfileApi(ctx)
-      val response = profileApi.listProfiles(offset, 1000)
-      if (response.getTotal >= 1000) {
+      val response = profileApi.listProfiles(offset, BillingProfileManagerDAO.BillingProfileRequestBatchSize)
+      // if we got the entire batch size, call next batch
+      if (response.getTotal >= BillingProfileManagerDAO.BillingProfileRequestBatchSize) {
         callListProfiles(accumulator ++ response.getItems.asScala.toSeq, offset + response.getTotal)
       } else {
         accumulator ++ response.getItems.asScala.toSeq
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -32,11 +32,17 @@ trait BillingProfileManagerDAO {
                            ctx: RawlsRequestContext
   ): Future[ProfileModel]
 
-  def listBillingProfiles(samUserResources: Seq[SamUserResource], ctx: RawlsRequestContext)(implicit
+  def listManagedApps(subscriptionId: UUID, ctx: RawlsRequestContext): Future[Seq[AzureManagedAppModel]]
+
+  def getBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): ProfileModel
```
cahrens: Note that I did not make this method return a Future-- see comment in https://github.com/broadinstitute/rawls/pull/1994/files#r949578236

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -125,24 +119,60 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-                CreationStatuses.Ready,
-                None,
-                None,
-                azureManagedAppCoordinates = Some(
-                  AzureManagedAppCoordinates(
-                    UUID.fromString(azureConfig.azureTenantId),
-                    UUID.fromString(azureConfig.azureSubscriptionId),
-                    azureConfig.azureResourceGroupId
+          val profileApi = apiClientProvider.getProfileApi(ctx)
+          Future.successful(profileApi.listProfiles(0, 1000).getItems.asScala.toSeq)
```
cahrens: Note that we have provide a limit to the number returned. To make the code robust we should check the number returned and continue to fetch profiles until we get them all. I didn't bother with that until we decide if we want to go with this approach (as opposed to fetching each profile individually by its `billlingProfileId`). But I don't think we should submit this with the arbitrary limit.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -125,24 +119,60 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
+        case _ => Future.successful(Seq())
+      }
+  }
+
+  def getHardcodedAzureBillingProject(samUserResources: Seq[SamUserResource], userInfo: UserInfo)(implicit
```
cahrens: I have created https://broadworkbench.atlassian.net/browse/WOR-486 to delete all this code associated with the hardcoded account.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -125,24 +119,60 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
             )
-          )
-        case false =>
-          Future.successful(Seq.empty)
-      }
+          case false =>
+            Future.successful(Seq.empty)
+        }
+    } yield billingProjects.filter { bp =>
+      samUserResources.map(_.resourceId).contains(bp.projectName.value)
```
cahrens: This filtering was in the original listBillingProfiles method, but I believe it is only needed for the hardcoded account because of the code at https://github.com/broadinstitute/rawls/pull/1994/files#diff-727d553ef55a7aabb5b0d0bca2e6a6fe44422feef4c7ad5ebb9c86bcd92b6572L179.

It also seems good that with the exception of this temporary hack method, we are no longer passing Sam resources into the DAO methods.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -62,36 +64,40 @@ object ManagedGroup {
...
+case class RawlsBillingProjectResponse(
+  projectName: RawlsBillingProjectName,
+  billingAccount: Option[RawlsBillingAccountName],
+  servicePerimeter: Option[ServicePerimeterName],
+  invalidBillingAccount: Boolean,
+  roles: Set[ProjectRoles.ProjectRole],
+  status: CreationStatuses.CreationStatus,
+  message: Option[String],
+  managedAppCoordinates: Option[AzureManagedAppCoordinates], // TODO: do we even want to return this?
+  cloudPlatform: String
```
cahrens, line 101: My first thought was to make this of type bio.terra.profile.model.CloudPlatform (which is an enum), but I had trouble with the JSON serialization format declaration for that.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -62,36 +64,40 @@ object ManagedGroup {
...
-                                       managedAppCoordinates: Option[AzureManagedAppCoordinates]
+case class RawlsBillingProjectResponse(
+  projectName: RawlsBillingProjectName,
+  billingAccount: Option[RawlsBillingAccountName],
+  servicePerimeter: Option[ServicePerimeterName],
+  invalidBillingAccount: Boolean,
+  roles: Set[ProjectRoles.ProjectRole],
+  status: CreationStatuses.CreationStatus,
+  message: Option[String],
+  managedAppCoordinates: Option[AzureManagedAppCoordinates], // TODO: do we even want to return this?
```
cahrens: We do wish to remove the managedAppCoordinates, but terra-ui will need to be updated first to use the cloud context instead. See https://github.com/broadinstitute/rawls/pull/1994#discussion_r949397934

We probably don't want to check in the code with a TODO. I think we can create a new ticket for the work to modify terra-ui and then remove `managedAppCoordinates`, and remove the TODO comment here which was just present to start a conversation.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -84,20 +89,53 @@ class MultiCloudWorkspaceService(ctx: RawlsRequestContext,
...
+            UUID.fromString(azureConfig.azureSubscriptionId),
+            azureConfig.azureResourceGroupId
+          ),
+          azureConfig.spendProfileId
         ),
         parentContext
       )
     } else {
-      workspaceService.createWorkspace(workspaceRequest, parentContext)
+      traceWithParent("withBillingProjectContext", ctx)(childSpan =>
```
cahrens, line 110: @aherbst-broad switched to using workspaceService.withBillingProjectContext and passed in parentSpan. Please check that I did the span stuff correctly.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -84,20 +89,53 @@ class MultiCloudWorkspaceService(ctx: RawlsRequestContext,
...
+        workspaceService.withBillingProjectContext(workspaceRequest.namespace, childSpan) { billingProject =>
+          billingProject.billingProfileId match {
+            case None                   => workspaceService.createWorkspace(workspaceRequest, ctx)
+            case Some(billingProfileID) =>
+              // TODO: error handling of no billing profile model exists
```
cahrens: The error handling still needs to be written, and the TODO removed.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -205,30 +222,45 @@ class UserService(protected val ctx: RawlsRequestContext,
...
       projectsInDB <- dataSource.inTransaction { dataAccess =>
         dataAccess.rawlsBillingProjectQuery.getBillingProjects(projectNames)
       }
-      bpmProfiles <- billingProfileManagerDAO.listBillingProfiles(samUserResources, ctx)
-    } yield constructBillingProjectResponses(samUserResources, projectsInDB ++ bpmProfiles)
+      // For projects in projectsInDB that have a billingProfileId, look up their managed app coordinates from BPM
+      (bpmProjects, legacyProjects) = projectsInDB.partition(_.billingProfileId.isDefined)
+      billingProfileModels <- billingProfileManagerDAO.getAllBillingProfiles(ctx)
+      populatedBpmProjects = bpmProjects.flatMap { bpmProject =>
+        // TODO error handling if no billing profileModel exists with the given ID.
```
cahrens: The error handling still needs to be done, and the TODO removed.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3128,7 +3128,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     * @tparam T
     * @return
     */
-  private def withBillingProjectContext[T](billingProjectName: String, parentContext: RawlsRequestContext)(
+  def withBillingProjectContext[T](billingProjectName: String, parentContext: RawlsRequestContext)(
```
cahrens, line 3181: Exposed this method and modified it to work with both GCP and Azure billing projects.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: > Note that I did not make this method return a Future

Which I think was correct

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -104,18 +109,18 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-    for {
-      billingProfiles <- getAllBillingProfiles(azureConfig, ctx)
-    } yield billingProfiles.filter { bp =>
-      samUserResources.map(_.resourceId).contains(bp.projectName.value)
+    @tailrec
+    def callListProfiles(accumulator: Seq[ProfileModel] = Seq.empty, offset: Int = 0): Seq[ProfileModel] = {
+      val profileApi = apiClientProvider.getProfileApi(ctx)
+      val response = profileApi.listProfiles(offset, 1000)
+      if (response.getTotal >= 1000) {
+        callListProfiles(accumulator ++ response.getItems.asScala.toSeq, offset + response.getTotal)
```
aherbst-broad, line 126: Can we get coverage on this recursive call?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -26,10 +26,11 @@ object Attributable {
   // if updating these, also update their use in SlickExpressionParsing
   val entityIdAttributeSuffix = "_id"
   val workspaceEntityType = "workspace"
-  val workspaceIdAttribute = workspaceEntityType + entityIdAttributeSuffix
```
blakery, line 29: The reason the diff on this file is so large, is mostly adding type notation to a bunch of publicly available implicit values.
There's a few other automatically applied style changes as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -62,45 +65,63 @@ object ManagedGroup {
...
+  servicePerimeter: Option[ServicePerimeterName],
+  invalidBillingAccount: Boolean,
+  roles: Set[ProjectRoles.ProjectRole],
+  status: CreationStatuses.CreationStatus,
+  message: Option[String],
+  managedAppCoordinates: Option[AzureManagedAppCoordinates], // remove after ui is updated  to use cloud context
+  cloudPlatform: String
 )
 
+object RawlsBillingProjectResponse {
```
blakery, line 104: Using an apply method like this seemed a lot more clear than using UserAuth. makeBillingProjectResponse.

I can remove it if anyone disagrees, or switch all the usages of makeBillingProjectResponse to this, if we want.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -213,55 +237,75 @@ class UserAuthJsonSupport extends JsonSupport {
 
     override def read(json: JsValue): ProjectRole = json match {
       case JsString(name) => ProjectRoles.withName(name)
-      case _              => throw new DeserializationException("could not deserialize project role")
+      case _              => throw DeserializationException("could not deserialize project role")
     }
   }
 
-  implicit val servicePerimeterNameFormat = ValueObjectFormat(ServicePerimeterName)
+  implicit val servicePerimeterNameFormat: ValueObjectFormat[ServicePerimeterName] = ValueObjectFormat(
```
blakery, line 244: This file also had a bunch of type annotations for public values added, contributing to the diff size.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -124,25 +129,57 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
...
-                  AzureManagedAppCoordinates(
-                    UUID.fromString(azureConfig.azureTenantId),
-                    UUID.fromString(azureConfig.azureSubscriptionId),
-                    azureConfig.azureResourceGroupId
+        case true => Future.successful(callListProfiles())
+        case _    => Future.successful(Seq())
+      }
+  }
+
+  def getHardcodedAzureBillingProject(samUserResourceIds: Set[String], userInfo: UserInfo)(implicit
```
blakery, line 146: This was only using the resourceIds from Sam, so I thought it might be more clear, or make mocking a bit easier to consume just the ids. 

It turned out not to matter much.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -205,38 +222,39 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-                                               billingProjectsInRawlsDB: Seq[RawlsBillingProject]
-  ): List[RawlsBillingProjectResponse] = {
-    val projectsByName = billingProjectsInRawlsDB.map(p => p.projectName.value -> p).toMap
-    samUserResources.toList
-      .flatMap { samUserResource =>
-        val projectRoles = samRolesToProjectRoles(samUserResource.direct.roles ++ samUserResource.inherited.roles)
-        projectsByName.get(samUserResource.resourceId).map(makeBillingProjectResponse(projectRoles, _))
-      }
-      .sortBy(_.projectName.value)
+  def listBillingProjectsV2(): Future[List[RawlsBillingProjectResponse]] = for {
```
blakery, line 225: I completely refactored this, when I added error handling for when a billing profile isn't found.

I started by nesting constructBillingProjectResponses inside listBillingProjectsV2, since it was only used there, but then I realized the resources it used were available for the entire sequence of transformations, if I just put the call to sam at the top.

This let me transform more directly to the RawlsBillingProjectResponse, and indicate the lack of a billing profile there.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -104,18 +109,18 @@ class BillingProfileManagerDAOImpl(samDAO: SamDAO,
       case Some(value) => value
     }
 
-    for {
-      billingProfiles <- getAllBillingProfiles(azureConfig, ctx)
-    } yield billingProfiles.filter { bp =>
-      samUserResources.map(_.resourceId).contains(bp.projectName.value)
+    @tailrec
```
blakery, line 121: This will be optimized by the compiler regardless of the annotation -  it just makes it so we'll get a compiler warning if it's not a proper tail call.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -205,38 +222,39 @@ class UserService(protected val ctx: RawlsRequestContext,
...
+    samUserResources <- samDAO.listUserResources(SamResourceTypeNames.billingProject, ctx.userInfo)
```
aherbst-broad, line 226: I think this may be a good candidate for extraction to the `BillingProjectOrchestrator` class and unit level coverage, given the amount of business logic in play here. Thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aherbst-broad: This looks ok 👍 , we can verify the traces are going through once we merge.

aherbst-broad (approved):
This looks reasonable. I tested this out locally against GCP + the hardcoded project and it worked well.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: Already added it

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: This method should be going away before long, so I don't think it matters.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
cahrens: I don't think we want both methods, though I don't have a preference about which one you keep.

cahrens: Currently this method is only used for the implicit parsing of json?

cahrens: And why does it need to be `apply` there vs. just `RawlsBillingProjectResponse`?

https://github.com/broadinstitute/rawls/pull/2010/files#diff-a257c53a0296aad7d58c01a06ae5a075382008742e4fc3bca818a0d172705814R307

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: I do think the billing _package_ is a good place for this.

Not sure about `BillingProjectOrchestrator` though. It seems to be more about managing multiple-step creation, and being able to roll things back, etc. I feel like this doesn't really fit, and risks bloating that class.

I'll poke around and see if there's somewhere where it fits.

blakery: I see why you were thinking about putting it there, now.... there isn't really a different billing specific service...

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -225,18 +247,22 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Test
...
+      namespace,
+      "fake_name",
+      Map.empty,
+      WorkspaceCloudPlatform.Azure,
+      "fake_region",
+      AzureManagedAppCoordinates(tenantId, subscriptionId, "fake_mrg_id"),
+      "fake_billingProjectId"
     )
-
+//case class AzureManagedAppCoordinates(tenantId: UUID, subscriptionId: UUID, managedResourceGroupId: String)
```
cahrens, line 265: Delete this?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -940,22 +945,24 @@ case class AttributeString(value: String) extends AttributeValue
...
+case class AttributeEntityReference(entityType: String, entityName: String) extends AttributeListElementable
 
 object AttributeStringifier {
   def apply(attribute: Attribute): String =
     attribute match {
       case AttributeNull                     => ""
       case AttributeString(value)            => value
       case AttributeNumber(value)            => value.toString()
-      case AttributeBoolean(value)           => value.toString()
+      case AttributeBoolean(value)           => value.toString
```
cahrens, line 963: This change confuses me. Can you explain it?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: I agree that `BillingProjectOrchestrator` didn't feel like the right place to me, or if it is, the documentation/concept of it needs to change. But I do think we should put it someplace where we can get some test coverage… the code is certainly non-trivial.

I would advocate for doing this work in a new ticket so we can close this one out this sprint and continue making progress on putting the frontend and backend together.

cahrens (approved):
I had a couple questions and would like us to consolidate the apply/create response code. But otherwise LGTM, and I don't necessarily need to re-review.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: I think the correct thing is to create something like a `BillingQueryService` (under a separate ticket). It would have the additional effect of breaking up resource dependencies.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
blakery: Functionally, there's no change. 
Scala _allows_ the omission of parentheses for zero argument methods.
The convention however, is to only omit parentheses for purely functional operations, with no side effects.

This was where I let my IDE auto-apply some style inspections on the file. I think the reason it didn't remove the parentheses from the others, is that this was a built-in object library object the IDE was confident it could remove them here, but was more conservative on the others since they were things like `BigDecimal`s.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
blakery: The short answer is to conform to other usages of the class.

The longer answer has to do with the details and purpose of case classes:

The idea of a case class is that it is to represent immutable data in a concise way, provide some conveniences around it automatically, and allow it to be easily de-structured in match statements.

When you define something as a case class, one thing the compiler does is define a default apply method with the class arguments. That's why you can just use `RawlsBillingProjectResponse(... args)` instead of the `new` keyword. 

What we're doing here is essentially overloading the `constructor`, but since it's a case class it makes for more consistent syntax to use `apply`, so you don't have to use a mix of `new` and the apply method (which would be _really_ confusing).

blakery:
> I had a couple questions and would like us to consolidate the apply/create response code

I didn't see that until after I hit merge.

PR #2013, opened 2022-08-30
Better error message for method-config-doesnt-exist when creating a submission [AJ-395]

davidangb:
When creating a submission, provide a more helpful error message if the method config for that submission cannot be found. The retry-submission feature in AJ-395 makes it significantly more likely for end users to run into this case: they could be retrying an older submission whose config has already been superseded.

Before:
Screenshot 8-29-2022 at 04 13 PM

After:
Screenshot 8-29-2022 at 04 15 PM

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
@@ -1,8 +1,9 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
+import akka.http.scaladsl.model.StatusCodes
 import cats.implicits._
 import nl.grons.metrics4.scala.Counter
-import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented._
 import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -216,6 +217,14 @@ trait SubmissionComponent {
...
           ) flatMap { entityId =>
             configIdAction flatMap { configId =>
+              if (configId.isEmpty) {
+                throw new RawlsExceptionWithErrorReport(
+                  ErrorReport(
+                    StatusCodes.BadRequest,
+                    s"Can't find this submission's method config ${submission.methodConfigurationNamespace}/${submission.methodConfigurationName}."
+                  )
+                )
+              }
```
davidangb, line 227: without this block, the `configId.get` at line 232 would fail instead, leading to a `NoSuchElement` error which is not helpful to the end user. See also https://github.com/broadinstitute/rawls/pull/1940#discussion_r951857339

sagehen03 (approved):
that seems easy enough!

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -435,7 +435,6 @@ trait ExecutionJsonSupport extends JsonSupport {
 
   implicit val SubmissionFormat = jsonFormat17(Submission)
 
-
```
davidangb, line 438: All of the changes in this file are `scalafmt`. I don't know why the formatting in this file passed in #1940, but is being flagged in this PR which I rebase-merged on top of #1940. But here it is, formatting is updated.

davidangb:
fiab-start I/O timeouts, sigh. Jenkins retest.

jordansafer (approved):

PR #2025, opened 2022-09-09
[WOR-492] Check that a user is enabled before getting a workspace context

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/browse/WOR-492)

We are varying the error messaging for disabled users when fetching a workspace depending on whether the workspace exists.

## This PR
* Adds a check with SAM to determine if the user is disabled when getting a workspace. If so, the get() call fails out with the same error message, regardless of whether the workspace exists. 

## TODO
- [x] Update tests

---------
  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.model
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model._
 import spray.json.DefaultJsonProtocol._
+import spray.json.RootJsonFormat
 
 /*
   Resource type names
@@ -154,6 +155,7 @@ case class SamUserResource(resourceId: String,
                            authDomainGroups: Set[WorkbenchGroupName],
                            missingAuthDomainGroups: Set[WorkbenchGroupName]
 )
+case class SamUserStatusResponse(userSubjectId: String, userEmail: String, enabled: Boolean)
 
 object SamModelJsonSupport extends JsonSupport {
   implicit val SamFullyQualifiesResourceIdFormat = jsonFormat2(SamFullyQualifiedResourceId)
@@ -173,4 +175,5 @@ object SamModelJsonSupport extends JsonSupport {
 
   implicit val SamRolesAndActionsFormat = jsonFormat2(SamRolesAndActions)
   implicit val SamUserResourceFormat = jsonFormat6(SamUserResource)
+  implicit val SamUserStatusResponseFormat = jsonFormat3(SamUserStatusResponse)
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.rawls.model
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model._
 import spray.json.DefaultJsonProtocol._
-import spray.json.RootJsonFormat
 
 /*
   Resource type names
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
@@ -31,7 +31,7 @@ case class RawlsGroupMemberList(
 case class RawlsUserInfo(user: RawlsUser, billingProjects: Seq[RawlsBillingProjectName])
 case class RawlsUserInfoList(userInfoList: Seq[RawlsUserInfo])
 
-case class RawlsUser(userSubjectId: RawlsUserSubjectId, userEmail: RawlsUserEmail)
+case class RawlsUser(userSubjectId: RawlsUserSubjectId, userEmail: RawlsUserEmail, enabled: Boolean = true)
 
 object RawlsUser {
   implicit def toRef(u: RawlsUser): RawlsUserRef = RawlsUserRef(u.userSubjectId)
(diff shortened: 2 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -31,7 +31,7 @@ case class RawlsGroupMemberList(
 case class RawlsUserInfo(user: RawlsUser, billingProjects: Seq[RawlsBillingProjectName])
 case class RawlsUserInfoList(userInfoList: Seq[RawlsUserInfo])
 
-case class RawlsUser(userSubjectId: RawlsUserSubjectId, userEmail: RawlsUserEmail)
+case class RawlsUser(userSubjectId: RawlsUserSubjectId, userEmail: RawlsUserEmail, enabled: Boolean = true)
```
marctalbott: It feels safer to me to have this default to false. How painful would the refactor be to have this default to false instead of true? Maybe the `apply` in the object sets `enabled=true` if that's only being used in tests

aherbst-broad: Yeah, this is a good point. I think tacking on `enabled` to RawlsUser when it's possible that there are call sites that will not know the enabled/disabled status is a code smell. I've introduced a new `SamUserStatusResponse` model in this [commit](https://github.com/broadinstitute/rawls/pull/2025/commits/aea707f038a4e68a4233fb527885d64e40aae91a) that will avoid this confusion.

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
```diff
@@ -13,8 +13,10 @@ class MockSamDAO(dataSource: SlickDataSource)(implicit executionContext: Executi
 
   override def registerUser(userInfo: UserInfo): Future[Option[RawlsUser]] = ???
 
-  override def getUserStatus(userInfo: UserInfo): Future[Option[RawlsUser]] =
-    Future.successful(Option(RawlsUser(userInfo)))
+  override def getUserStatus(userInfo: UserInfo): Future[Option[SamUserStatusResponse]] =
```
aherbst-broad, line 16: Re-using `RawlsUser` as the return type was forcing me to tack on an `enabled` val to that case class, with a default of `true` (see @marctalbott 's point earlier in the PR). Since there are only two call sites for this method, shifting to a first-class SAM model seems semantically clearer.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -39,6 +32,13 @@ trait WorkspaceSupport {
   import dataSource.dataAccess.driver.api._
 
   // Access/permission helpers
+  def userEnabledCheck(userInfo: UserInfo): Future[Unit] =
```
aherbst-broad, line 35: This method and the related call on L189 are the nut of the fix.

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.model
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model._
 import spray.json.DefaultJsonProtocol._
+import spray.json.RootJsonFormat
```
davidangb: nit: unused import

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -39,6 +32,13 @@ trait WorkspaceSupport {
...
+    samDAO.getUserStatus(userInfo) flatMap {
+      case Some(user) =>
+        if (user.enabled) Future.successful()
+        else Future.failed(new UserDisabledException(StatusCodes.Unauthorized, "Unauthorized"))
+      case None => Future.failed(new UserDisabledException(StatusCodes.Unauthorized, "Unauthorized"))
```
davidangb: optional, I wouldn't block the PR on this … if you felt like consolidating:
```scala
      case Some(user) if (user.enabled) => Future.successful()
      case _ => Future.failed(new UserDisabledException(StatusCodes.Unauthorized, "Unauthorized"))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -186,6 +186,7 @@ trait WorkspaceSupport {
                                         attributeSpecs: Option[WorkspaceAttributeSpecs] = None
   ): Future[Workspace] =
     for {
+      _ <- userEnabledCheck(ctx.userInfo)
```
davidangb, line 187: thinking out loud: I poked at this a bit to find a good way to avoid always making this Sam call. The best I could come up with was adding a recover/catch and only making this extra Sam call on an error, which would eliminate the extra Sam call from the typical/successful path. The code gets more complex with that. Again, not gonna block the PR on this, not requesting changes, just talking.

davidangb (approved):
one nit and two TOLs inline, giving this a :+1: and your call on whether or not to change anything based on my feedback.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
aherbst-broad: Yeah, that's a good point. I'm inclined to favor a simpler/more obvious codepath for now; we can optimize down the line if need be.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
aherbst-broad: Good call, implemented.

davidangb (approved):

PR #2027, opened 2022-09-09
refactoring spendreporting for WOR-430

blakery:
When implementing WOR-430, it was very difficult to test changes, or even see what was going on in the service.
This is the resulting refactor.

In unit tests:
* remove all behind the scenes test setup using mixins
* remove all calls to real databases (all resources are now properly mocked)
* change called service methods to target tested behavior  more narrowly

in service:
* move functions that don't use any class resources to the companion object to allow for testing without mocking resources
* refactor function boundaries to better isolate side effects and improve testability

I also added metrics to the spend reporting service before deciding to turn this into a separate PR.
I'm leaving it in, since it is pretty isolated, and won't break anything existing as is.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -41,16 +41,13 @@ object SpendReportingService {
     val currency = allRows.map(_.get("currency").getStringValue).distinct match {
       case head :: List() => Currency.getInstance(head)
       case head :: tail =>
-        throw new RawlsExceptionWithErrorReport(
-          ErrorReport(
-            StatusCodes.BadGateway, // todo: Probably the wrong status code
-            s"Inconsistent currencies found while aggregating spend data: $head and ${tail.head} cannot be combined"
-          )
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.BadGateway, // todo: Probably the wrong status code
+          s"Inconsistent currencies found while aggregating spend data: $head and ${tail.head} cannot be combined"
         )
       case List() =>
-        throw new RawlsExceptionWithErrorReport(
-          ErrorReport(StatusCodes.NotFound, "No currencies found while aggregating spend data")
-        )
+        throw RawlsExceptionWithErrorReport(StatusCodes.NotFound, "No currencies found while aggregating spend data")
+
     }
 
     def sum(rows: List[FieldValueList], field: String): String = rows
@@ -80,8 +77,9 @@ object SpendReportingService {
       .map { case (googleProjectId, projectRows) =>
         val workspaceName = workspaceProjectsToNames.getOrElse(
           googleProjectId,
-          throw new RawlsExceptionWithErrorReport(
-            ErrorReport(StatusCodes.BadGateway, s"unexpected project ${googleProjectId.value} returned by BigQuery")
+          throw RawlsExceptionWithErrorReport(
+            StatusCodes.BadGateway,
+            s"unexpected project ${googleProjectId.value} returned by BigQuery"
           )
         )
         SpendReportingForDateRange(
@@ -152,11 +150,9 @@ class SpendReportingService(
     samDAO.userHasAction(SamResourceTypeNames.billingProject, projectName.value, action, ctx.userInfo).flatMap {
       case true => op
       case false =>
-        throw new RawlsExceptionWithErrorReport(
-          errorReport = ErrorReport(
-            StatusCodes.Forbidden,
-            s"${ctx.userInfo.userEmail.value} cannot perform ${action.value} on project ${projectName.value}"
-          )
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.Forbidden,
+          s"${ctx.userInfo.userEmail.value} cannot perform ${action.value} on project ${projectName.value}"
         )
     }
 
@@ -170,57 +166,49 @@ class SpendReportingService(
     .flatMap {
       case true => op
       case false =>
-        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, "This API is not live yet."))
+        throw RawlsExceptionWithErrorReport(StatusCodes.Forbidden, "This API is not live yet.")
     }
 
   private def dateTimeToISODateString(dt: DateTime): String = dt.toString(ISODateTimeFormat.date())
 
-  def getSpendExportConfiguration(billingProjectName: RawlsBillingProjectName): Future[BillingProjectSpendExport] =
+  def getSpendExportConfiguration(projectName: RawlsBillingProjectName): Future[BillingProjectSpendExport] =
     dataSource
       .inTransaction { dataAccess =>
-        dataAccess.rawlsBillingProjectQuery.getBillingProjectSpendConfiguration(billingProjectName)
+        dataAccess.rawlsBillingProjectQuery.getBillingProjectSpendConfiguration(projectName)
       }
       .recover { case _: RawlsException =>
-        throw new RawlsExceptionWithErrorReport(
-          ErrorReport(StatusCodes.BadRequest,
-                      s"billing account not found on billing project ${billingProjectName.value}"
-          )
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.BadRequest,
+          s"billing account not found on billing project ${projectName.value}"
         )
       }
       .map(
         _.getOrElse(
-          throw new RawlsExceptionWithErrorReport(
-            ErrorReport(StatusCodes.NotFound, s"billing project ${billingProjectName.value} not found")
-          )
+          throw RawlsExceptionWithErrorReport(StatusCodes.NotFound, s"billing project ${projectName.value} not found")
         )
       )
 
-  def getWorkspaceGoogleProjects(
-    billingProjectName: RawlsBillingProjectName
-  ): Future[Map[GoogleProject, WorkspaceName]] = dataSource
-    .inTransaction { dataAccess =>
-      dataAccess.workspaceQuery.listWithBillingProject(billingProjectName)
-    }
-    .map { workspaces =>
-      workspaces.collect {
-        case workspace if workspace.workspaceVersion == WorkspaceVersions.V2 =>
-          GoogleProject(workspace.googleProjectId.value) -> workspace.toWorkspaceName
-      }.toMap
-    }
+  def getWorkspaceGoogleProjects(projectName: RawlsBillingProjectName): Future[Map[GoogleProject, WorkspaceName]] =
+    dataSource
+      .inTransaction { dataAccess =>
+        dataAccess.workspaceQuery.listWithBillingProject(projectName)
+      }
+      .map { workspaces =>
+        workspaces.collect {
+          case workspace if workspace.workspaceVersion == WorkspaceVersions.V2 =>
+            GoogleProject(workspace.googleProjectId.value) -> workspace.toWorkspaceName
+        }.toMap
+      }
 
   def validateReportParameters(startDate: DateTime, endDate: DateTime): Unit = if (startDate.isAfter(endDate)) {
-    throw new RawlsExceptionWithErrorReport(
-      ErrorReport(
-        StatusCodes.BadRequest,
-        s"start date ${dateTimeToISODateString(startDate)} must be before end date ${dateTimeToISODateString(endDate)}"
-      )
+    throw RawlsExceptionWithErrorReport(
+      StatusCodes.BadRequest,
+      s"start date ${dateTimeToISODateString(startDate)} must be before end date ${dateTimeToISODateString(endDate)}"
     )
   } else if (Days.daysBetween(startDate, endDate).getDays > spendReportingServiceConfig.maxDateRange) {
-    throw new RawlsExceptionWithErrorReport(
-      ErrorReport(
-        StatusCodes.BadRequest,
-        s"provided dates exceed maximum report date range of ${spendReportingServiceConfig.maxDateRange} days"
-      )
+    throw RawlsExceptionWithErrorReport(
+      StatusCodes.BadRequest,
+      s"provided dates exceed maximum report date range of ${spendReportingServiceConfig.maxDateRange} days"
     )
   }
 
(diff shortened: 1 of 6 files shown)


blakery:
A lot of the hideousness of the diff is because I merged in develop at the start of this.

model/src/main/scala/org/broadinstitute/dsde/rawls/RawlsException.scala
```diff
@@ -1,11 +1,25 @@
...
 
-import org.broadinstitute.dsde.rawls.model.ErrorReport
+import akka.http.scaladsl.model.StatusCode
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
 
 class RawlsException(message: String = null, cause: Throwable = null) extends Exception(message, cause)
 
 class RawlsExceptionWithErrorReport(val errorReport: ErrorReport) extends RawlsException(errorReport.toString)
 
+object RawlsExceptionWithErrorReport {
```
blakery, line 10: just adding some quality of life improvements, so we don't need to explicitly create an error report every time

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
-      case true => op
-      case false =>
-        Future.failed(
-          new RawlsExceptionWithErrorReport(
-            errorReport = ErrorReport(
-              StatusCodes.Forbidden,
-              s"${ctx.userInfo.userEmail.value} cannot perform ${action.value} on project ${projectName.value}"
-            )
-          )
+  def extractSpendReportingResults(
```
blakery, line 33: I moved this to the companion object to:
1. Make it more clear that it was a pure function that didn't rely on any class resources
2. Simplify unit testing a bit by removing the need to instantiate the service with all the mocked resources

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
-      }
-      SpendReportingForDateRange(
-        cost.toString,
-        credits.toString,
-        currency.getCurrencyCode,
-        category = Option(category),
-        subAggregation = subAggregation
-      )
-    }.toList
+  def getWorkspaceGoogleProjects(projectName: RawlsBillingProjectName): Future[Map[String, WorkspaceName]] =
```
blakery, line 186: It's arguable that it would have been better to leave the map key as `GoogleProject` instead of a `String`.
I was getting frustrated with all the verbose types floating around, and the bloat that was added by the having to constantly unbox something that was essentially a renamed String. 
If it wasn't already a separate case class, I'd probably just add `type GoogleProject = String`, which would give the clarity of what we were tossing around, without needing to have the (mental) overhead of the separate class.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -411,188 +315,140 @@ class SpendReportingServiceSpec
...
     }
 
-  private def verifyWorkspaceCategorySubAggregation(topLevelAggregation: SpendReportingAggregation) = {
+  def verifyWorkspaceCategorySubAggregation(topLevelAggregation: SpendReportingAggregation): Unit = {
     topLevelAggregation.aggregationKey shouldBe SpendReportingAggregationKeys.Workspace
 
-    topLevelAggregation.spendData.map { spendData =>
-      val workspaceGoogleProject = spendData.googleProjectId.getOrElse(fail("spend results not parsed correctly")).value
-      val subAggregation = spendData.subAggregation.getOrElse(fail("spend results not parsed correctly"))
+    topLevelAggregation.spendData.foreach { spendData =>
```
cahrens, line 425: For my own edification, why is foreach better than map here?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -601,231 +457,330 @@ class SpendReportingServiceSpec
...
-      runAndWait(
-        dataSource.dataAccess.rawlsBillingProjectQuery.create(
-          RawlsBillingProject(projectName, CreationStatuses.Ready, billingAccount = None, None)
-        )
-      )
+  it should "throw an exception if the billing project does not have a linked billing account" in {
+    val samDAO = mock[SamDAO]
+    when(samDAO.userHasAction(any(), any(), any(), any())).thenReturn(Future.successful(true))
+    val dataSource = mock[SlickDataSource]
+    when(dataSource.inTransaction[Option[BillingProjectSpendExport]](any(), any()))
```
cahrens, line 600: I'm unclear how this maps to the billing project not having a linked billing account.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -601,231 +457,330 @@ class SpendReportingServiceSpec
...
+    val service = new SpendReportingService(
+      testContext,
+      mock[SlickDataSource],
+      Resource.pure[IO, GoogleBigQueryService[IO]](mock[GoogleBigQueryService[IO]]),
+      mock[SamDAO],
+      spendReportingServiceConfig
+    )
+    val startDate = DateTime.now().minusDays(spendReportingServiceConfig.maxDateRange)
+    val endDate = DateTime.now()
+    service.validateReportParameters(startDate, endDate)
```
cahrens, line 670: Is it worth testing what the validation does return?

cahrens: Or is the thing here that the start and end dates are within range? Perhaps just updating the test description would clarify this.

cahrens:
@blakery you can rebase to pick up the integration test fixes.

cahrens (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
cahrens: Is it just that we don't actually want to return anything, and map returns a sequential?

blakery: map is for when you want to do something with the result. 
foreach is to execute for side effects

In this case, what we get back from the execution is just a list of type Unit placeholder type, which is something along the lines of void in other languages. 
So there isn't actually anything we can do with the result anyway - this just makes the result agree how we're using it.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
+    end: DateTime,
+    workspaceProjectsToNames: Map[String, WorkspaceName],
+    aggregations: Set[SpendReportingAggregationKeyWithSub]
+  ): SpendReportingResults = {
+
+    val currency = allRows.map(_.get("currency").getStringValue).distinct match {
+      case head :: List() => Currency.getInstance(head)
+      case head :: tail =>
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.BadGateway, // todo: Probably the wrong status code
```
aherbst-broad, line 45: Plain 500 probably works here.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
blakery: `getSpendExportConfiguration` is calling `RawlsBillingProject.getBillingProjectSpendConfiguration`, which calls

```scala
  def toBillingProjectSpendExport(projectRecord: RawlsBillingProjectRecord): BillingProjectSpendExport = {
    val table = for {
      googleProjectId <- projectRecord.spendReportDatasetGoogleProject
      datasetName <- projectRecord.spendReportDataset
      tableName <- projectRecord.spendReportTable
    } yield s"$googleProjectId.$datasetName.$tableName"

    val billingAccount = RawlsBillingAccountName(
      projectRecord.billingAccount.getOrElse(
        throw new RawlsException(s"billing account not set on project ${projectRecord.projectName}")
      )
    )

    BillingProjectSpendExport(RawlsBillingProjectName(projectRecord.projectName), billingAccount, table)
  }
```

So if the billing account isn't set, it throws the exception, which `getSpendExportConfiguration` is looking for.
It isn't completely clear to me why this is divided up where it is,  but from a behavior perspective, this does mock how the underlying call behaves.

I'm not completely certain how `RawlsExceptionWithErrorReport`s are handled when the ErrorReport doesn't have a status code set (I would assume it would default to 400). If that is the case, we could just not catch the exception in `getSpendExportConfiguration`, and let it bubble up, which would give the same behavior, and let us eliminate this test entirely. That actually sounds like a good idea....

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
blakery: On a success, it doesn't return anything (or rather, it returns Unit, which is the same thing).
So no exception == validation passed.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -601,231 +457,330 @@ class SpendReportingServiceSpec
...
-          Duration.Inf
-        )
-      }
-      e.errorReport.statusCode shouldBe Option(StatusCodes.BadRequest)
-  }
+  it should "throw an exception if the billing project cannot be found" in {
+    val samDAO = mock[SamDAO]
+    when(samDAO.userHasAction(any(), any(), any(), any())).thenReturn(Future.successful(true))
+    val dataSource = mock[SlickDataSource]
+    when(dataSource.inTransaction[Option[BillingProjectSpendExport]](any(), any())).thenReturn(Future.successful(None))
```
aherbst-broad, line 579: Out of scope for this PR, but I'd like to start pushing more toward the repository pattern to avoid this slick datasource mocking (i.e., similar to what I've done [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala)).

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
aherbst-broad: I _think_ we default to 500 judging by the default exception handler code [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala#L48)

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
+    with RawlsInstrumented {
+
   /**
-    * Ensure that BigQuery results only include one type of currency and return that currency.
+    * Base name for all metrics. This will be prepended to all generated metric names.
+    * Example: dev.firecloud.rawls
     */
-  private def getCurrency(rows: List[FieldValueList]): Currency = {
-    val currencies = rows.map(_.get("currency").getStringValue)
+  override val workbenchMetricBaseName: String = spendReportingServiceConfig.workbenchMetricBaseName
```
aherbst-broad, line 143: Just curious, are we doing anything with this at the moment or is this pure wireup for a subsequent PR?

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
cahrens: Right, the test description confused me because the critical point here is that the start and end dates in this test are in range. Perhaps

_should "not throw an exception when validating start and end dates representing a valid range"_

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
+      _.collect {
+        case w if w.workspaceVersion == WorkspaceVersions.V2 => w.googleProjectId.value -> w.toWorkspaceName
+      }.toMap
+    }
 
-    SpendReportingAggregation(
-      SpendReportingAggregationKeys.Category,
-      categorySpend
+  def validateReportParameters(startDate: DateTime, endDate: DateTime): Unit = if (startDate.isAfter(endDate)) {
+    throw RawlsExceptionWithErrorReport(
```
aherbst-broad, line 194: Out of scope as well but: We should think about pursuing more specific exception types for this class as well, rather than reusing RawlsExceptionWithErrorReport everywhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
-            .addNamedParameter("billingAccountId",
-                               stringQueryParameterValue(spendExportConf.billingAccountId.withoutPrefix())
-            )
-            .addNamedParameter("startDate", stringQueryParameterValue(dateTimeToISODateString(startDate)))
-            .addNamedParameter("endDate", stringQueryParameterValue(dateTimeToISODateString(endDate)))
-            .addNamedParameter("projects",
-                               stringArrayQueryParameterValue(workspaceProjectsToNames.keySet.map(_.value).toList)
-            )
-            .build()
+          query = getQuery(aggregationKeys, spendExportConf)
```
aherbst-broad, line 269: I like the refactoring here

aherbst-broad (approved):
Looks good.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
blakery: this is just setting up metrics for future changes

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
blakery: I would think 400, since retrying won't fix it, and it's _technically_ user data

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: I'd vote to leave this as a `GoogleProject`. Dealing with the unboxing is a small inconvenience compared to the benefit to readability by having a meaningful return type.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
marctalbott: It's data that Google generates iirc, so it's not something that users really have any control over as far as I know. I'd vote 500

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -23,377 +22,271 @@ import scala.jdk.CollectionConverters._
...
+      }
+      .toList
+
+    def byWorkspace(rows: List[FieldValueList], subKey: SubKey): List[SpendReportingForDateRange] = rows
+      .groupBy(row => row.get("googleProjectId").getStringValue)
+      .map { case (googleProjectId, projectRows) =>
+        val workspaceName = workspaceProjectsToNames.getOrElse(
+          googleProjectId,
+          throw RawlsExceptionWithErrorReport(
+            StatusCodes.BadGateway,
```
marctalbott, line 80: If we're changing the inconsistent currency exception, we should do the same to this one

marctalbott (commented):
saw that you merged it so sending over the comments I had. Nothing was blocking anyways

PR #2040, opened 2022-09-15
Cleaner async import error messages

davidangb:
Previously, we stringified a whole bunch of ErrorReports. Now, we extract their messages and compile them, and make sure that's what we send in the response to a user.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -332,10 +332,14 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
                 )
               }
             case Failure(t) =>
+              val errMsg = t match {
+                case errRpt: RawlsExceptionWithErrorReport => errRpt.errorReport.message
+                case x                                     => x.getMessage
+              }
               publishMessageToUpdateImportStatus(attributes.importId,
                                                  Option(status),
                                                  ImportStatuses.Error,
-                                                 Option(t.getMessage)
+                                                 Option(errMsg)
               )
           }
         case Some(_) =>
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -622,6 +622,71 @@ class AvroUpsertMonitorSpec(_system: ActorSystem)
       }
   }
```
davidangb, line 624: the below test is copied from the pre-existing "bubble up useful error message if upserts result in partial failure" test just above it, changed only to ensure that all entities fail to import and to expect a slightly different error message in the end.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -513,8 +517,8 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
         throw new RawlsExceptionWithErrorReport(
           RawlsErrorReport(
             StatusCodes.BadRequest,
-            s"All entities failed to update. There were ${failureReports.size} errors in total$additionalErrorString",
-            failureReportsForCaller
+            s"All entities failed to update. There were ${failureReports.size} errors in total$additionalErrorString" +
+              s" Error messages: ${stringMessageFromFailures(failureReportsForCaller, 100)}"
```
davidangb, line 521: we used `stringMessageFromFailures` elsewhere, such as in the partial-success case. But here in the total-failure case, we didn't use it … now we do. This fixes the nested stringification.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -332,10 +332,14 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
                 )
               }
             case Failure(t) =>
+              val errMsg = t match {
+                case errRpt: RawlsExceptionWithErrorReport => errRpt.errorReport.message
+                case x                                     => x.getMessage
+              }
```
davidangb, line 338: this fixes the outermost stringification in the case we have a `RawlsExceptionWithErrorReport`.

calypsomatic:
Are there other areas of the code other than `AvroUpsertMonitor` that need to have their error reports better stringified, and/or is the ticket only focused on this one spot?

calypsomatic (approved):

davidangb:
we discussed in person - this PR is focused just on this one spot. If/when we discover additional error messages that need to be cleaned up, let's create new PRs for those.

sagehen03 (approved):

PR #2046, opened 2022-09-19
Add and remove users from BPM projects (WOR-387).

cahrens:
When add or remove member on a billing project is called (v2 version of the API), we will update the BPM billing profile record if the project is backed by a BPM record.

If testing this out with a locally running terra-ui, make sure you are up-to-date with dev.

This PR also deprecates the v1 version of the billing APIs (Orchestration changes to follow).

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -78,7 +78,8 @@ paths:
       tags:
         - billing
       summary: create billing project in rawls and google
-      description: create new billing project in rawls and google
+      description: Create new billing project in rawls and google. Deprecated, use v2 version.
+      deprecated: true
       operationId: createBillingProjectFull
       requestBody:
         description: create project request
@@ -121,7 +122,8 @@ paths:
       tags:
         - billing
       summary: list members of billing project
-      description: list members of billing project. owners can view all members, users will only be able to view owners
+      description: List members of billing project. Owners can view all members, users will only be able to view owners. Deprecated, use v2 version.
+      deprecated: true
       operationId: listBillingProjectMembers
       parameters:
         - $ref: '#/components/parameters/billingProjectIdPathParam'
@@ -147,7 +149,8 @@ paths:
       tags:
         - billing
       summary: add user or group to billing project the caller owns
-      description: add user or group to billing project the caller owns
+      description: Add user or group to billing project the caller owns. Deprecated, use v2 version.
+      deprecated: true
       operationId: addUserToBillingProject
       parameters:
         - $ref: '#/components/parameters/billingProjectIdPathParam'
@@ -180,7 +183,8 @@ paths:
       tags:
         - billing
       summary: remove user or group from billing project the caller owns
-      description: remove user or group from billing project the caller owns
+      description: Remove user or group from billing project the caller owns. Deprecated, use v2 version.
+      deprecated: true
       operationId: removeUserFromBillingProject
       parameters:
         - $ref: '#/components/parameters/billingProjectIdPathParam'
@@ -210,6 +214,30 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
   /api/billing/v2:
+    get:
+      tags:
+        - billing_v2
+      summary: list billing projects for a user
+      description: list billing projects for a user
+      operationId: listUserBillingProjectsV2
+      responses:
+        200:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                type: array
+                description: list of billing projects for this user
+                items:
+                  $ref: '#/components/schemas/RawlsBillingProjectResponse'
+        404:
+          description: User not found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
     post:
       tags:
         - billing_v2
@@ -3436,7 +3464,8 @@ paths:
         - billing
         - users
       summary: list billing projects for a user
-      description: list billing projects for a user
+      description: List billing projects for a user. Deprecated, use v2 version.
+      deprecated: true
       operationId: listUserBillingProjects
       responses:
         200:
@@ -3462,7 +3491,8 @@ paths:
         - billing
         - users
       summary: billing project status
-      description: billing project status
+      description: Billing project status. Deprecated, use v2 API.
+      deprecated: true
       operationId: billingProjectStatus
       parameters:
         - name: projectName
@@ -3489,7 +3519,8 @@ paths:
         - billing
         - users
       summary: delete billing project
-      description: delete billing project
+      description: Delete billing project. Deprecated, use v2 version.
+      deprecated: true
       operationId: deleteBillingProject
       parameters:
         - name: projectName
@@ -5757,6 +5788,11 @@ components:
           description: informational message about the project
         azureManagedAppCoordinates:
           $ref: '#/components/schemas/AzureManagedAppCoordinates'
+        cloudPlatform:
+          type: string
+          enum:
+            - GCP
+            - AZURE
     AzureManagedAppCoordinates:
       required:
         - tenantId
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -6,7 +6,19 @@ import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
 import org.broadinstitute.dsde.rawls.model.ProjectRoles.ProjectRole
-import org.broadinstitute.dsde.rawls.model.{AzureManagedAppCoordinates, CreationStatuses, ErrorReport, ProjectRoles, RawlsBillingAccountName, RawlsBillingProject, RawlsBillingProjectName, RawlsRequestContext, SamResourceAction, SamResourceTypeNames, UserInfo}
+import org.broadinstitute.dsde.rawls.model.{
+  AzureManagedAppCoordinates,
+  CreationStatuses,
+  ErrorReport,
+  ProjectRoles,
+  RawlsBillingAccountName,
+  RawlsBillingProject,
+  RawlsBillingProjectName,
+  RawlsRequestContext,
+  SamResourceAction,
+  SamResourceTypeNames,
+  UserInfo
+}
 
 import java.util.UUID
 import scala.annotation.tailrec
@@ -72,12 +84,12 @@ class BillingProfileManagerDAOImpl(
   }
 
   override def createBillingProfile(
-                                     displayName: String,
-                                     billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates],
-                                     ctx: RawlsRequestContext
-                                   ): ProfileModel = {
+    displayName: String,
+    billingInfo: Either[RawlsBillingAccountName, AzureManagedAppCoordinates],
+    ctx: RawlsRequestContext
+  ): ProfileModel = {
     val azureManagedAppCoordinates = billingInfo match {
-      case Left(_) => throw new NotImplementedError("Google billing accounts not supported in billing profiles")
+      case Left(_)       => throw new NotImplementedError("Google billing accounts not supported in billing profiles")
       case Right(coords) => coords
     }
 
@@ -190,7 +202,7 @@ class BillingProfileManagerDAOImpl(
   private def getProfileApiPolicy(samRole: ProjectRole): String =
     samRole match {
       case ProjectRoles.Owner => "owner"
-      case ProjectRoles.User => "user"
+      case ProjectRoles.User  => "user"
     }
 
   def addProfilePolicyMember(billingProfileId: UUID,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -1,12 +1,30 @@
 package org.broadinstitute.dsde.rawls.billing
 
-import bio.terra.profile.model.{AzureManagedAppModel, CloudPlatform, CreateProfileRequest, PolicyMemberRequest, ProfileModel}
+import bio.terra.profile.model.{
+  AzureManagedAppModel,
+  CloudPlatform,
+  CreateProfileRequest,
+  PolicyMemberRequest,
+  ProfileModel
+}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
 import org.broadinstitute.dsde.rawls.model.ProjectRoles.ProjectRole
-import org.broadinstitute.dsde.rawls.model.{AzureManagedAppCoordinates, CreationStatuses, ErrorReport, ProjectRoles, RawlsBillingAccountName, RawlsBillingProject, RawlsBillingProjectName, RawlsRequestContext, SamResourceAction, SamResourceTypeNames, UserInfo}
+import org.broadinstitute.dsde.rawls.model.{
+  AzureManagedAppCoordinates,
+  CreationStatuses,
+  ErrorReport,
+  ProjectRoles,
+  RawlsBillingAccountName,
+  RawlsBillingProject,
+  RawlsBillingProjectName,
+  RawlsRequestContext,
+  SamResourceAction,
+  SamResourceTypeNames,
+  UserInfo
+}
 
 import java.util.UUID
 import scala.annotation.tailrec
@@ -33,9 +51,17 @@ trait BillingProfileManagerDAO {
     ec: ExecutionContext
   ): Future[Seq[RawlsBillingProject]]
 
-  def addProfilePolicyMember(billingProfileId: UUID, role: ProjectRole, memberEmail: String, ctx: RawlsRequestContext): Unit
+  def addProfilePolicyMember(billingProfileId: UUID,
+                             role: ProjectRole,
+                             memberEmail: String,
+                             ctx: RawlsRequestContext
+  ): Unit
 
-  def deleteProfilePolicyMember(billingProfileId: UUID, role: ProjectRole, memberEmail: String, ctx: RawlsRequestContext): Unit
+  def deleteProfilePolicyMember(billingProfileId: UUID,
+                                role: ProjectRole,
+                                memberEmail: String,
+                                ctx: RawlsRequestContext
+  ): Unit
 
 }
 
@@ -182,22 +208,35 @@ class BillingProfileManagerDAOImpl(
     } yield billingProjects.filter(bp => samUserResourceIds.contains(bp.projectName.value))
   }
 
-  private def getProfileApiPolicy(samRole: ProjectRole): String = {
-     samRole match {
+  private def getProfileApiPolicy(samRole: ProjectRole): String =
+    samRole match {
       case ProjectRoles.Owner => "owner"
-      case ProjectRoles.User => "user"
+      case ProjectRoles.User  => "user"
     }
-  }
 
-  def addProfilePolicyMember(billingProfileId: UUID, role: ProjectRole, memberEmail: String, ctx: RawlsRequestContext): Unit = {
-    apiClientProvider.getProfileApi(ctx).addProfilePolicyMember(
-      new PolicyMemberRequest().email(memberEmail), billingProfileId, getProfileApiPolicy(role)
-    )
-  }
+  def addProfilePolicyMember(billingProfileId: UUID,
+                             role: ProjectRole,
+                             memberEmail: String,
+                             ctx: RawlsRequestContext
+  ): Unit =
+    apiClientProvider
+      .getProfileApi(ctx)
+      .addProfilePolicyMember(
+        new PolicyMemberRequest().email(memberEmail),
+        billingProfileId,
+        getProfileApiPolicy(role)
+      )
 
-  def deleteProfilePolicyMember(billingProfileId: UUID, role: ProjectRole, memberEmail: String, ctx: RawlsRequestContext): Unit = {
-    apiClientProvider.getProfileApi(ctx).deleteProfilePolicyMember(
-      billingProfileId, getProfileApiPolicy(role), memberEmail
-    )
-  }
+  def deleteProfilePolicyMember(billingProfileId: UUID,
+                                role: ProjectRole,
+                                memberEmail: String,
+                                ctx: RawlsRequestContext
+  ): Unit =
+    apiClientProvider
+      .getProfileApi(ctx)
+      .deleteProfilePolicyMember(
+        billingProfileId,
+        getProfileApiPolicy(role),
+        memberEmail
+      )
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -1,30 +1,12 @@
 package org.broadinstitute.dsde.rawls.billing
 
-import bio.terra.profile.model.{
-  AzureManagedAppModel,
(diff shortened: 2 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -188,4 +208,35 @@ class BillingProfileManagerDAOImpl(
     } yield billingProjects.filter(bp => samUserResourceIds.contains(bp.projectName.value))
   }
 
+  private def getProfileApiPolicy(samRole: ProjectRole): String =
+    samRole match {
```
cahrens, line 203: I thought this was better than lower-casing ProjectRoles.Owner.toString and ProjectRoles.User.toString, as it doesn't rely on knowledge of what those values happen to be.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
       }
     }
 
+  private def getLegacyBillingPolicies(samRole: ProjectRole): Seq[SamResourcePolicyName] =
+    samRole match {
+      case ProjectRoles.Owner => Seq(SamBillingProjectPolicyNames.owner)
+      case ProjectRoles.User =>
+        Seq(SamBillingProjectPolicyNames.workspaceCreator, SamBillingProjectPolicyNames.canComputeUser)
```
cahrens, line 660: I'm considering this "legacy" based on https://github.com/broadinstitute/rawls/pull/2046/files#diff-727d553ef55a7aabb5b0d0bca2e6a6fe44422feef4c7ad5ebb9c86bcd92b6572R688

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
   def addUserToBillingProject(projectName: RawlsBillingProjectName,
                               projectAccessUpdate: ProjectAccessUpdate
   ): Future[Unit] =
     requireProjectAction(projectName, SamBillingProjectActions.alterPolicies) {
-      val policies = projectAccessUpdate.role match {
-        case ProjectRoles.Owner => Seq(SamBillingProjectPolicyNames.owner)
-        case ProjectRoles.User =>
-          Seq(SamBillingProjectPolicyNames.workspaceCreator, SamBillingProjectPolicyNames.canComputeUser)
+      val policies = getLegacyBillingPolicies(projectAccessUpdate.role)
+      addUserToBillingProjectInner(projectName, projectAccessUpdate, policies)
```
cahrens, line 673: I recommend hiding whitespace differences so this refactoring lines up better.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -167,19 +165,19 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
                   put {
                     complete {
                       userServiceConstructor(ctx)
-                        .addUserToBillingProject(RawlsBillingProjectName(projectId),
-                                                 ProjectAccessUpdate(userEmail, ProjectRoles.withName(workbenchRole))
+                        .addUserToBillingProjectV2(RawlsBillingProjectName(projectId),
```
cahrens, line 168: Following the pattern of v2 methods handling BPM. Also this now lines up with the operationId values in the [Swagger documentation](https://github.com/broadinstitute/rawls/blob/5bf74086a612d272b64dd8f0a882540453d79f8b/core/src/main/resources/swagger/api-docs.yaml#L522).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
+              )
+              Future.successful(())
+            } else {
+              Future.failed(regrets)
+            }
+          }
       }
+    } yield {}
 
+  def addUserToBillingProjectV2(projectName: RawlsBillingProjectName,
```
cahrens, line 702: Does orchestration need to be updated to point to these new method? The orchestration Swagger already says that the `operationID` for the v2 version is `addUserToBillingProjectV2` (which didn't previously exist). I don't understand though how the dispatching of calls happens-- do we need to change orchestration to actually dispatch to these new methods?

aherbst-broad: The operationId refers to [this](https://github.com/broadinstitute/rawls/blob/8552f25e1bcf0b68c0d77ede2d3758609717a382/core/src/main/resources/swagger/api-docs.yaml#L522-L586) endpoint in the _rawls_ swagger. We should verify if firecloud-orch is indeed calling this endpoint.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-                Future.failed(regrets)
-              }
-            }
+            Seq(getV2BillingPolicy(projectAccessUpdate.role))
         }
+        _ <- addUserToBillingProjectInner(projectName, projectAccessUpdate, policies)
       } yield {}
     }
 
+  def getBillingProfileId(projectName: RawlsBillingProjectName): Future[Option[String]] =
```
aherbst-broad: This would be a good method to put on the [billing project repository.](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala)

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-                                projectName.value,
-                                policy,
-                                projectAccessUpdate.email,
-                                ctx.userInfo
-          )
-          .recover {
-            case e: RawlsExceptionWithErrorReport if e.errorReport.statusCode.contains(StatusCodes.BadRequest) =>
-              throw new RawlsExceptionWithErrorReport(e.errorReport.copy(statusCode = Some(StatusCodes.NotFound)))
-          }
+        billingProjectId <- getBillingProfileId(projectName)
```
aherbst-broad: I think this var should be called `billingProfileId`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -43,6 +51,18 @@ trait BillingProfileManagerDAO {
     ec: ExecutionContext
   ): Future[Seq[RawlsBillingProject]]
 
+  def addProfilePolicyMember(billingProfileId: UUID,
+                             role: ProjectRole,
+                             memberEmail: String,
+                             ctx: RawlsRequestContext
+  ): Unit
```
aherbst-broad, line 52: So this DAO has half of its methods returning Future's and half returning bare values. I think we should covert all of them to bare values for consistency. Thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -188,4 +208,35 @@ class BillingProfileManagerDAOImpl(
     } yield billingProjects.filter(bp => samUserResourceIds.contains(bp.projectName.value))
   }
 
+  private def getProfileApiPolicy(samRole: ProjectRole): String =
+    samRole match {
+      case ProjectRoles.Owner => "owner"
+      case ProjectRoles.User  => "user"
+    }
+
+  def addProfilePolicyMember(billingProfileId: UUID,
```
aherbst-broad, line 208: It'd be good to get coverage on these methods in the related [spec](https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala)

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
+              Future.failed(regrets)
+            }
+          }
       }
+    } yield {}
 
+  def addUserToBillingProjectV2(projectName: RawlsBillingProjectName,
+                                projectAccessUpdate: ProjectAccessUpdate
+  ): Future[Unit] =
+    requireProjectAction(projectName, SamBillingProjectActions.alterPolicies) {
```
aherbst-broad, line 705: I'm not seeing the reason for this method to exist; it looks like it's the same behavior as `addUserToBillingProject` with the difference being that it's checking for a related billing profile ; if present it's adding the user to the billing profile and using the newer policies for the call to `adduserToBillingProjectInner`. 

Why wouldn't we want this behavior on both the v2 and v1 apis as we have it today? I admit it's odd that we have an existing v2 API that's calling the same codepath as the v1 api.

aherbst-broad (commented):
Looking good, had a couple of comments.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: I thought our general philosophy was that only the v2 APIs supported BPM. For example, we defined a `listBillingProjectsV2` method that gets things from BPM, and left the existing `listBillingProjects` method alone.

I'm also trying to line up with the operationId defined in Swagger by naming the method `addUserToBillingProjectV2`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: Yes, it should be consistent-- this is just me not understanding Scala. I will try to sort it out.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: I added coverage for the new methods. The old ones are covered in integration tests (which we can't do yet for Azure).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Oh right, orchestration calls the Rawls REST API.

This is what orch is calling. I don't see that we call the v2 version of _anything_, and I don't see that there are pass throughs for much of what is in the Swagger related to billing. Is the Swagger just all bogus (@aherbst-broad , @marctalbott )?

```
  private def editBillingMembershipURL(projectId: String, role: ProjectRole, email: String) = {
    FireCloudConfig.Rawls.authUrl + s"/billing/$projectId/${role.toString}/${java.net.URLEncoder.encode(email, "UTF-8")}"
  }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: I agree that bare values are the way to go: the futures are only useful if you're doing something while it's executing.

cahrens: I converted all of them except 2-- `getAllBillingProfiles` and `getHardcodedAzureBillingProject` both rely on `samDAO.userHasAction`, which returns a Future. However, once we can delete the alpha group gating that can be removed (and the hardcoded method is getting deleted).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
     requireProjectAction(projectName, SamBillingProjectActions.alterPolicies) {
-      val policies = projectAccessUpdate.role match {
-        case ProjectRoles.Owner => Seq(SamBillingProjectPolicyNames.owner)
-        case ProjectRoles.User =>
-          Seq(SamBillingProjectPolicyNames.workspaceCreator, SamBillingProjectPolicyNames.canComputeUser)
+      val policies = getLegacyBillingPolicies(projectAccessUpdate.role)
+      addUserToBillingProjectInner(projectName, projectAccessUpdate, policies)
+    }
+
+  private def addUserToBillingProjectInner(projectName: RawlsBillingProjectName,
```
blakery, line 676: If you find it useful, you can nest this inside the main function.
Personally, I tend to prefer nesting when I'm breaking the function up for reasons that don't involve re-use, since it makes it clear where the logic belongs, and you still have the context for values that are just pass-through.

Totally optional though.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -651,60 +651,126 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-        case ProjectRoles.User  => SamBillingProjectPolicyNames.workspaceCreator
-      }
+      removeUserFromBillingProjectInner(projectName, projectAccessUpdate)
+    }
+
+  private def removeUserFromBillingProjectInner(projectName: RawlsBillingProjectName,
+                                                projectAccessUpdate: ProjectAccessUpdate
+  ): Future[Unit] =
+    for {
+      _ <- samDAO
```
blakery: If you aren't doing anything with the results, and it's a single async operation, the for construct feels a bit gratuitous. 
If you remove the for, and do nothing else, it's exactly the same functionality.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: This is used by 2 functions-- the v1 and v2 wrappers.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Thanks, done.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Definitely, thanks.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: Done.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreator.scala
```diff
@@ -60,15 +53,15 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
-      profileModel <- blocking {
-        billingProfileManagerDAO.createBillingProfile(createProjectRequest.projectName.value,
-                                                      createProjectRequest.billingInfo,
-                                                      ctx
-        )
-      }
       _ <- billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
-    } yield {}
-
+    } yield { }
```
cahrens, line 65: Have the `yield` here because `setBillingProfileId` returns Future[Int]. We could change the return value for this method, but folks aren't using the return value and `validateBillingProjectCreationRequest` is also returning Future[Unit].

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -1315,4 +1318,324 @@ class UserServiceSpec
       result should contain theSameElementsAs expected
     }
   }
+
+  behavior of "addUserToBillingProjectV2"
+
+  it should "update billing profile record when a user is added to a billing project" in {
+    val billingProfileId = UUID.randomUUID()
```
cahrens: Since these test points no longer require `withMinimalTestDatabase`, I might want to move them into a separate test class so they don't have the setup overhead.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: @aherbst-broad I gave this a shot in https://github.com/broadinstitute/rawls/pull/2046/commits/7d22a12f5fb2e29e6024a4b25a7a784db2d0e6ba -- let me know what you think.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: We have decided to keep calls to BPM in the v2 versions only. I still need to see if I can mark the v1 methods as deprecated in the Swagger.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: This is causing a lot of tests to bomb because I need to provide an implementation for `getBillingProfileId`. If you think this is worth pursuing, I will probably need some help tomorrow figuring out how best to do this.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreatorSpec.scala
```diff
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.rawls.billing
 
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import bio.terra.profile.model.{AzureManagedAppModel, ProfileModel}
-import cromwell.client.ApiException
```
cahrens, line 5: I don't think this was the ApiException we intended to throw!

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Actually, I realize now that my approach of passing BillingRegistry into UserService is overkill given the current situation where UserService directly uses the `slickDataSource` for all other database access. I think I'm going to revert these changes and instead simply create an instance of BillingRegistry in UserService with the `slickDataSource` reference it already has. That will drastically reduce the changes, and once/if UserService has migrated to using BillingRegistry for everything we can start passing it into UserService in place of the `slickDataSource`. Sound good to you, @aherbst-broad?

aherbst-broad: @cahrens ah, yeah, that makes sense. Thanks for the explanation. (assuming you mean BillingRepository instead of BillingRegistry).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: (Will discuss at tech talk)

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -78,7 +78,8 @@ paths:
       tags:
         - billing
       summary: create billing project in rawls and google
-      description: create new billing project in rawls and google
+      description: Create new billing project in rawls and google. Deprecated, use v2 version.
+      deprecated: true
```
cahrens, line 82: Deprecated everything except listBillingAccounts, which does not  have a v2 version.

image

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -210,6 +214,30 @@ paths:
         500:
           $ref: '#/components/responses/RawlsInternalError'
   /api/billing/v2:
+    get:
+      tags:
```
cahrens, line 218: Was missing. We will want to update orchestration Swagger also (both with deprecation and this). Waiting until after the tech talk to better understand how orchestration swagger and dispatching works.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3462,7 +3491,8 @@ paths:
         - billing
         - users
       summary: billing project status
-      description: billing project status
+      description: Billing project status. Deprecated, use v2 API.
```
cahrens, line 3494: We don't have a specific "status" endpoint in v2, but status is included in the billing project response.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Have marked endpoints deprecated in Swagger.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -129,6 +129,8 @@ class UserService(protected val ctx: RawlsRequestContext,
 
   import dataSource.dataAccess.driver.api._
 
+  val billingRepository = new BillingRepository(dataSource)
```
aherbst-broad, line 132: Minor nit, instead of a direct instantiation, this could added as a parameter to the (admittedly gargantuan) constructor. We are instantiating a billing repository instance during boot [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala#L497) which could likely be threaded through the UserService ctor [call](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala#L312)

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreatorSpec.scala
aherbst-broad: Good catch

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
blakery: you could wait in the method
I don't think it's important though, since it's going away.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: That also allows for better mocking in tests. Right now you'd need to.... override the class? I don't think `spy` would even work, since it's a `val`, not a function.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: Ignore me, then :-P

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: This is the approach I did initially in https://github.com/broadinstitute/rawls/commit/7d22a12f5fb2e29e6024a4b25a7a784db2d0e6ba (discussed in https://github.com/broadinstitute/rawls/pull/2046#discussion_r976415528). However, it led to lots of changes because every place UserService is instantiated (tests), I had to pass a BillingRepository. And I didn't see that it added any value given that UserService still takes slickDataSource as an input.

I think we eventually want to move to passing BillingRepository instead of slickDataSource, but we are a ways off from that.

cahrens: And yes, passing it into `UserService` did allow for mocking instead of using the minimal test database. I ended up backing that test change out, but do recognize that's the direction we ultimately want to go.

@blakery if you want to change to passing `BillingRepository` into `UserService` as part of your ticket, go for it!

cahrens:
jenkins retest

core/src/main/resources/swagger/api-docs.yaml
cahrens: Created https://broadworkbench.atlassian.net/browse/WOR-521 for orchestration Swagger updates.

PR #2047, opened 2022-09-23
[WOR-518] Update getBillingProject to return full cloud platform details

blakery:
[WOR-518](https://broadworkbench.atlassian.net/browse/WOR-518)

Adds billing profile lookup + cloud mapping to getBillingProject.

  * To Demo flag is set
  * Add notes on how to QA
  and
  DBs in Google Cloud Console
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.rawls.model
 
-import bio.terra.profile.model.{ProfileModel, CloudPlatform => BPMCloudPlatform}
+import bio.terra.profile.model.{CloudPlatform => BPMCloudPlatform, ProfileModel}
 import bio.terra.workspace.model.{CloudPlatform => WSMCloudPlatform}
 import bio.terra.datarepo.model.{CloudPlatform => DRCloudPlatform}
 
@@ -25,9 +25,8 @@ object CloudPlatform extends Enumeration {
     case DRCloudPlatform.AZURE => AZURE
   }
 
-
   def apply(profile: ProfileModel): CloudPlatform = profile.getCloudPlatform match {
-    case BPMCloudPlatform.GCP => GCP
+    case BPMCloudPlatform.GCP   => GCP
     case BPMCloudPlatform.AZURE => AZURE
     case _ => UNKNOWN // BPM requires cloud platform to be set, so we shouldn't need to rely on the app coordinates
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
@@ -1,0 +1,27 @@
+package org.broadinstitute.dsde.rawls.model
+import bio.terra.profile.model.{CloudPlatform => BPMCloudPlatform}
+import bio.terra.workspace.model.{CloudPlatform => WSMCloudPlatform}
+import bio.terra.datarepo.model.{CloudPlatform => DRCloudPlatform}
+
+object CloudPlatform extends Enumeration {
+  type CloudPlatform = Value
+  val GCP: CloudPlatform = Value("GCP")
+  val AZURE: CloudPlatform = Value("AZURE")
+  val UNKNOWN: CloudPlatform = Value("UNKNOWN")
+
+  def apply(platform: BPMCloudPlatform): CloudPlatform = platform match {
+    case BPMCloudPlatform.GCP   => GCP
+    case BPMCloudPlatform.AZURE => AZURE
+  }
+
+  def apply(platform: WSMCloudPlatform): CloudPlatform = platform match {
+    case WSMCloudPlatform.GCP   => GCP
+    case WSMCloudPlatform.AZURE => AZURE
+  }
+
+  def apply(platform: DRCloudPlatform): CloudPlatform = platform match {
+    case DRCloudPlatform.GCP   => GCP
+    case DRCloudPlatform.AZURE => AZURE
+  }
+
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
@@ -1,5 +1,5 @@
 package org.broadinstitute.dsde.rawls.model
-import bio.terra.profile.model.{CloudPlatform => BPMCloudPlatform}
+import bio.terra.profile.model.{ProfileModel, CloudPlatform => BPMCloudPlatform}
 import bio.terra.workspace.model.{CloudPlatform => WSMCloudPlatform}
 import bio.terra.datarepo.model.{CloudPlatform => DRCloudPlatform}
 
@@ -22,6 +22,13 @@ object CloudPlatform extends Enumeration {
   def apply(platform: DRCloudPlatform): CloudPlatform = platform match {
     case DRCloudPlatform.GCP   => GCP
     case DRCloudPlatform.AZURE => AZURE
+  }
+
+
+  def apply(profile: ProfileModel): CloudPlatform = profile.getCloudPlatform match {
+    case BPMCloudPlatform.GCP => GCP
+    case BPMCloudPlatform.AZURE => AZURE
+    case _ => UNKNOWN // BPM requires cloud platform to be set, so we shouldn't need to rely on the app coordinates
   }
 
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
@@ -1,4 +1,5 @@
 package org.broadinstitute.dsde.rawls.model
+
 import bio.terra.profile.model.{ProfileModel, CloudPlatform => BPMCloudPlatform}
 import bio.terra.workspace.model.{CloudPlatform => WSMCloudPlatform}
 import bio.terra.datarepo.model.{CloudPlatform => DRCloudPlatform}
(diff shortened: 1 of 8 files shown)


cahrens:
@blakery I know the title of the ticket doesn't reflect it, but the intention was that the GET for a single billing project returns the same thing as the list method. That means that the managed app coordinates need to be included also (as you will see in the current terra-ui workaround which is to be removed as part of the ticket).

Currently with list I see this as an example:

```
  {
    "cloudPlatform": "AZURE",
    "invalidBillingAccount": false,
    "managedAppCoordinates": {
      "managedResourceGroupId": "mrg-aherbst-2022020-20220815170937",
      "subscriptionId": "62b22893-6bc1-46d9-8a90-806bb3cce3c9",
      "tenantId": "fad90753-2022-4456-9b0a-c7e5b934e408"
    },
    "projectName": "CARAzureRawls",
    "roles": [
      "Owner"
    ],
    "status": "Ready"
  },
```

vs. GET:
```
{
  "cloudPlatform": "GCP",
  "invalidBillingAccount": false,
  "projectName": "CARAzureRawls",
  "roles": [
    "Owner"
  ],
  "status": "Ready"
}
```

I think the GCP project responses do match.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -111,7 +115,7 @@ object RawlsBillingProjectResponse {
     project.status,
     project.message,
     project.azureManagedAppCoordinates,
-    project.azureManagedAppCoordinates.map(_ => CloudPlatform.AZURE).getOrElse(CloudPlatform.GCP).toString
+    platform.map(_.toString).getOrElse("Unknown")
```
cahrens: Can you update the enum in the Swagger documentation to indicate that "Unknown" is a possibility? We may want all caps for consistency. It might be best to make a CloudPlatform.UNKNOWN type.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -84,53 +86,40 @@ object UserService {
...
+  samDAO: SamDAO,
+  bqServiceFactory: GoogleBigQueryServiceFactory,
+  bigQueryCredentialJson: String,
+  requesterPaysRole: String,
+  protected val dmConfig: DeploymentManagerConfig,
+  protected val projectTemplate: ProjectTemplate,
+  servicePerimeterService: ServicePerimeterService,
+  adminRegisterBillingAccountId: RawlsBillingAccountName,
+  billingProfileManagerDAO: BillingProfileManagerDAO,
+  val billingRepository: BillingRepository // = new BillingRepository(dataSource)
```
cahrens: Remove comment?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -214,73 +203,76 @@ class UserService(protected val ctx: RawlsRequestContext,
...
+      case _ if roles.isEmpty => None
+      case (None, _)          => Some(RawlsBillingProjectResponse(roles, project, Some(CloudPlatform.GCP)))
+      case (Some(_), Some(p)) =>
+        val c = AzureManagedAppCoordinates(p.getTenantId, p.getSubscriptionId, p.getManagedResourceGroupId)
+        val copy = project.copy(azureManagedAppCoordinates = Some(c))
+        Some(RawlsBillingProjectResponse(roles, copy, Some(CloudPlatform.AZURE)))
+      case (Some(id), None) =>
+        val message = Some(s"Unable to find billing profile in Billing Profile Manager for billing profile id: $id")
+        val copy = project.copy(message = message, status = CreationStatuses.Error)
+        Some(RawlsBillingProjectResponse(roles, copy, None))
```
cahrens: Pass `CloudPlatform.UNKNOWN` here?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -214,73 +203,76 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-          case None =>
-            project.copy(
-              message = Some(s"Unable to find billing profile in Billing Profile Manager for billing profile id: $id"),
-              status = CreationStatuses.Error
-            )
-        }
+    // For projects in projectsInDB that have a billingProfileId, look up their managed app coordinates from BPM
+    val billingProfile = project.billingProfileId.flatMap(id => billingProfiles.find(_.getId == UUID.fromString(id)))
+    (project.billingProfileId, billingProfile) match {
+      case (None, _) => RawlsBillingProjectResponse(roles, project, Some(CloudPlatform.GCP))
```
cahrens: It feels like this block of code should be shared with 217 above (especially since the bug was caused by the fact that they diverged).

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -1173,147 +1175,265 @@ class UserServiceSpec
...
+    ).thenReturn(Future.successful(Seq(hardcodedExternalProject)))
 
-      val userService = getUserService(dataSource, samDAO, billingProfileManagerDAO = bpmDAO)
+    val userService = getUserService(samDAO = samDAO, bpmDAO = bpmDAO, billingRepository = Some(repository))
 
-      val result = Await.result(userService.listBillingProjectsV2(), Duration.Inf)
+    val expected = Seq(
+      RawlsBillingProjectResponse(Set(ProjectRoles.Owner), ownerProject, CloudPlatform.GCP),
+      RawlsBillingProjectResponse(Set(ProjectRoles.User), billingProfileBackedProject, CloudPlatform.AZURE),
+      RawlsBillingProjectResponse(Set(ProjectRoles.User), hardcodedExternalProject, CloudPlatform.GCP)
```
cahrens, line 1394: Nit: the hardcoded project really shouldn't be GCP. But it's going away soon, so it doesn't really matter.

cahrens (approved):

cahrens:
@blakery don't forget the follow-up terra-ui PR to remove the workaround.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -214,73 +204,75 @@ class UserService(protected val ctx: RawlsRequestContext,
...
-            project.copy(azureManagedAppCoordinates = Some(c))
-          case None =>
-            project.copy(
-              message = Some(s"Unable to find billing profile in Billing Profile Manager for billing profile id: $id"),
-              status = CreationStatuses.Error
-            )
-        }
+      case (Some(_), Some(p)) =>
+        val c = AzureManagedAppCoordinates(p.getTenantId, p.getSubscriptionId, p.getManagedResourceGroupId)
+        RawlsBillingProjectResponse(roles, project.copy(azureManagedAppCoordinates = Some(c)), CloudPlatform.AZURE)
```
cahrens: @blakery thinking about this more, we probably should determine `AZURE` vs. `GCP` here based on the existence of the managed app coordinates.

blakery: I'm not sure there's a difference, since the managed app coordinates are directly derived from the billing profile.

cahrens: @blakery Right, but the managed app coordinates would exist for GCP billing profiles (when we start using them)-- they are specific to Azure. So what I'm suggesting is that when we are in the case of having a billing profile, we set the cloud platform to Azure vs. GCP based on the existence of managed app coordinates.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
```diff
@@ -0,0 +1,27 @@
+package org.broadinstitute.dsde.rawls.model
+import bio.terra.profile.model.{CloudPlatform => BPMCloudPlatform}
+import bio.terra.workspace.model.{CloudPlatform => WSMCloudPlatform}
+import bio.terra.datarepo.model.{CloudPlatform => DRCloudPlatform}
+
+object CloudPlatform extends Enumeration {
+  type CloudPlatform = Value
+  val GCP: CloudPlatform = Value("GCP")
+  val AZURE: CloudPlatform = Value("AZURE")
+  val UNKNOWN: CloudPlatform = Value("UNKNOWN")
```
aherbst-broad, line 11: I remain dubious about exposing an `UNKNOWN` cloud platform. This feels like it should be a 500 since the system is in an inconsistent state and very likely a bug (i.e, how would a consumer of this api reasonably accommodate this state?)

cahrens: @aherbst-broad billing projects can be in invalid/errored states. In that case, the message field is used to indicate the cause of the issue.

In terra-ui, this is accommodated by not displaying an icon at all (theoretically we could have an "unknown" icon if we wished), and there is an error popup that displays the error message. throwing a 500 would not be as friendly.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: I verified that the CloudPlatform is required in BPM when projects are created

core/src/main/scala/org/broadinstitute/dsde/rawls/model/CloudPlatform.scala
blakery: ~I'm more inclined to agree, now that I've verified that CloudPlatform is required in BPM.
@cahrens Was there another reason that I'm forgetting here?~
nvm: stale page

aherbst-broad (approved):
🚢

PR #2053, opened 2022-09-29
Aj 582 clone snapshot references along with workspace

calypsomatic:
[AJ-582 Cloning a workspace also clones snapshots-by-reference](https://broadworkbench.atlassian.net/browse/AJ-582)
Cloning workspaces previously ignored snapshot-by-references.  This PR includes them.  
Note: If the user of the newly cloned workspace does not have permission to view the snapshot, then it will still appear in the workspace, but the user will be unable to view any data, as below:
Screen Shot 2022-09-29 at 1 34 36 PM
Also, this specifically deals exclusively with GCP resources, so expanding beyond that will need further work.   I'm guessing that can be a follow-on ticket when the time comes.
The clone workspace API in Firecloud Orchestration calls the rawlsDao, so i believe no further work should be required there.
This PR is half-draft as it does not address the case of very large numbers of snapshots in one workspace, but I am putting it out for comment while I think about that.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -207,12 +207,14 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
                                destinationWorkspaceId: UUID,
                                name: String,
                                ctx: RawlsRequestContext
-  ): Unit = {
+  ): DataRepoSnapshotResource = {
     val requestBody = new CloneReferencedResourceRequestBody()
       .name(name)
-      .cloningInstructions(CloningInstructionsEnum.NOTHING)
+      .cloningInstructions(CloningInstructionsEnum.REFERENCE)
       .destinationWorkspaceId(destinationWorkspaceId)
-    getReferencedGcpResourceApi(ctx).cloneGcpDataRepoSnapshotReference(requestBody, sourceWorkspaceId, snapshotId)
+    getReferencedGcpResourceApi(ctx)
+      .cloneGcpDataRepoSnapshotReference(requestBody, sourceWorkspaceId, snapshotId)
+      .getResource()
   }
 
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -201,4 +201,18 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
         ),
       workspaceId
     )
+
+  def cloneSnapshotByReference(sourceWorkspaceId: UUID,
+                               snapshotId: UUID,
+                               destinationWorkspaceId: UUID,
+                               name: String,
+                               ctx: RawlsRequestContext
+  ): Unit = {
+    val requestBody = new CloneReferencedResourceRequestBody()
+      .name(name)
+      .cloningInstructions(CloningInstructionsEnum.NOTHING)
+      .destinationWorkspaceId(destinationWorkspaceId)
+    getReferencedGcpResourceApi(ctx).cloneGcpDataRepoSnapshotReference(requestBody, sourceWorkspaceId, snapshotId)
+  }
+
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -66,6 +66,21 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       Future.successful(snapshotRef)
     }
 
+  def cloneSnapshotByReference(sourceWorkspaceId: UUID,
+                               targetWorkspaceId: UUID,
+                               snapshot: DataRepoSnapshotResource
+  ): DataRepoSnapshotResource = {
+    if (!workspaceStubExists(targetWorkspaceId, ctx)) {
+      workspaceManagerDAO.createWorkspace(targetWorkspaceId, ctx)
+    }
+    workspaceManagerDAO.cloneSnapshotByReference(sourceWorkspaceId,
+                                                 snapshot.getMetadata().getResourceId(),
+                                                 targetWorkspaceId,
+                                                 snapshot.getMetadata().getName(),
+                                                 ctx
+    )
+  }
+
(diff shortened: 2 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -201,4 +201,20 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
       workspaceId
     )
+
+  def cloneSnapshotByReference(sourceWorkspaceId: UUID,
+                               snapshotId: UUID,
+                               destinationWorkspaceId: UUID,
+                               name: String,
+                               ctx: RawlsRequestContext
+  ): DataRepoSnapshotResource = {
+    val requestBody = new CloneReferencedResourceRequestBody()
```
calypsomatic, line 211: There's no real need for this to return `DataRepoSnapshotResource` as I don't think it's ever used, although in theory it could be under some future use cases.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -66,6 +66,21 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       Future.successful(snapshotRef)
     }
 
+  def cloneSnapshotByReference(sourceWorkspaceId: UUID,
+                               targetWorkspaceId: UUID,
+                               snapshot: DataRepoSnapshotResource
+  ): DataRepoSnapshotResource = {
+    if (!workspaceStubExists(targetWorkspaceId, ctx)) {
```
calypsomatic, line 74: Since this method is currently only used internally and doesn't care about the return value, I haven't wrapped it in a `Future`. I am happy to adjust upon request.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1070,6 +1075,22 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                                                         dataAccess,
                                                         s1
                                 ) { destWorkspaceContext =>
+                                  // TODO: Without limit to make sure all snapshots are fetched
+                                  val snapshotResponse: Future[SnapshotListResponse] = snapshotServiceConstructor(s1)
```
calypsomatic: Looking for input on what kind of numbers we might expect to see here and ideas for how best to make sure we get all snapshots.  Looping with a large limit until done?  Is there a way to get all without the limit?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -93,7 +95,8 @@ object WorkspaceService {
                   googleIamDao: GoogleIamDAO,
                   terraBillingProjectOwnerRole: String,
                   terraWorkspaceCanComputeRole: String,
-                  terraWorkspaceNextflowRole: String
+                  terraWorkspaceNextflowRole: String,
+                  snapshotServiceConstructor: RawlsRequestContext => SnapshotService
   )(
```
calypsomatic, line 100: Needed to add this `SnapshotService` into the `WorkspaceService` which led to a lot of changes throughout the code.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
davidangb: have you inspected the `DataRepoSnapshotResource` that is actually returned from this call? From a quick read of the code, I think it might always be null - though I could have misread

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -66,6 +66,21 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+      workspaceManagerDAO.createWorkspace(targetWorkspaceId, ctx)
+    }
```
davidangb, line 76: As a standalone method, this is fine. But in the current PR, `WorkspaceService` may retrieve tens or hundreds of references, then call `cloneSnapshotByReference` for each one - which means we'll execute `workspaceStubExists` over and over. Each call to `workspaceStubExists` is a new REST request to WSM, so this is pretty inefficient.

I also commented elsewhere in the PR - I suggest `SnapshotService` should have a "clone all references" method. In that method, you could call `workspaceStubExists` and `createWorkspace` (if necessary) once at the top of the method, then retrieve and clone a whole bunch of references.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb: Some of the WFL-enabled workspaces could have multiple hundreds of references, so we definitely need to handle large numbers. I think the best way to handle it is by paging through N results at a time, until the next page returns 0 results. 100 could be a good page size.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1070,6 +1075,22 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+                                      100
+                                    )
+                                  for {
+                                    snapshots <- snapshotResponse
+                                    snapshot <- snapshots.gcpDataRepoSnapshots
+                                  } snapshotServiceConstructor(s1).cloneSnapshotByReference(
+                                    sourceWorkspaceContext.workspaceIdAsUUID,
+                                    destWorkspaceContext.workspaceIdAsUUID,
+                                    snapshot
+                                  )
```
davidangb, line 1082: I suggest moving all of this code into `SnapshotService`, under a new "clone all snapshot references" method which just accepts the source workspace and destination workspace as arguments. `SnapshotService` should know how to manipulate snapshot references; `WorkspaceService` should be as agnostic as possible about those.

davidangb (commented):
comments inline!

I'm struck looking at this code at how much more efficient and encapsulated it would be if WSM would clone the stub workspace - and all of its references - instead of doing up here in Rawls with potentially hundreds of api calls to WSM. What's your thinking on that? Is that worth invetigating?

calypsomatic:
> comments inline!
> 
> I'm struck looking at this code at how much more efficient and encapsulated it would be if WSM would clone the stub workspace - and all of its references - instead of doing up here in Rawls with potentially hundreds of api calls to WSM. What's your thinking on that? Is that worth investigating?

Yeah, I got caught up in just seeing how to get it to work at all.  Sure seems like it WSM should at the very least have a batch clone call.  I'll poke around and see what I turn up!

calypsomatic:
Superseded by https://github.com/broadinstitute/rawls/pull/2082

PR #2080, opened 2022-10-17
WOR-568 Finish port to sam client lib

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-568

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -517,15 +517,6 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     }
   }
 
-  // TODO
-  @deprecated(message = "user listUserResources instead")
-  override def getPoliciesForType(resourceTypeName: SamResourceTypeName,
-                                  userInfo: UserInfo
-  ): Future[Set[SamResourceIdWithPolicyName]] = {
-    val url = samServiceURL + s"/api/resources/v1/${resourceTypeName.value}"
-    retry(when401or5xx)(() => pipeline[Set[SamResourceIdWithPolicyName]](userInfo) apply RequestBuilding.Get(url))
-  }
-
   override def listUserResources(resourceTypeName: SamResourceTypeName,
                                  ctx: RawlsRequestContext
   ): Future[Seq[SamUserResource]] =
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -230,7 +230,7 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
     retry(when401or5xx) { () =>
       val callback = new SamApiCallback[sam.model.UserStatus]("createUserV2")
 
-      usersApi(ctx).createUserV2Async("", callback)
+      usersApi(ctx).createUserV2Async(null, callback)
 
       callback.future
         .map { userStatus =>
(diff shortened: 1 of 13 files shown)


dvoet:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -893,56 +893,40 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     }
 
     for {
-      workspacePolicies <- traceWithParent("getPolicies", ctx)(_ =>
-        samDAO.getPoliciesForType(SamResourceTypeNames.workspace, ctx.userInfo)
-      )
-      // filter out the policies that are not related to access levels, if a user has only those ignore the workspace
+      workspaceResources <- samDAO.listUserResources(SamResourceTypeNames.workspace, ctx)
```
dvoet, line 896: This is the spot where transition to `listUserResources` matters most. The return structure is substantially different. The old way relied on policy naming to figure out access levels (an artifact of implementation using ldap). The new way uses role names. Fortunately for workspaces roles have the same name as policies. There is also a difference in cardinality: if the user is in multiple policies for a resource the old way would return each but in the new way all the users roles/actions are in a singe resource entry (which also contains info about which are inherited or public).

dvoet:
jenkins retest

dvoet:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -170,7 +106,7 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
                                    ctx: RawlsRequestContext
   ): Future[SamPolicySyncStatus] =
     retry(when401or5xx) { () =>
-      val callback = new SamApiCallback[SyncStatus]("syncStatus")
+      val callback = new SamApiCallback[sam.model.SyncStatus]("syncStatus")
```
dvoet, line 109: I consistently added the package prefix to all model classes coming from the sam client lib to try to keep things straight

MatthewBemis (approved):
:shipit:

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -374,18 +374,15 @@ class UserService(
...
+      projectUsers <- samDAO.listAllResourceMemberIds(SamResourceTypeNames.billingProject, projectName.value, ctx)
       _ <- projectUsers.toList.traverse(destroyPet(_, projectName))
     } yield ()
 
   private def destroyPet(userIdInfo: UserIdInfo, projectName: GoogleProjectId): Future[Unit] =
     for {
       petSAJson <- samDAO.getPetServiceAccountKeyForUser(projectName, RawlsUserEmail(userIdInfo.userEmail))
       petUserInfo <- gcsDAO.getUserInfoUsingJson(petSAJson)
-      _ <- samDAO.deleteUserPetServiceAccount(projectName, petUserInfo)
+      _ <- samDAO.deleteUserPetServiceAccount(projectName, ctx)
```
cahrens, line 385: @dvoet with this change, lines 386-387 appear to be unnecessary… `petUserInfo` is no longer used. I'm moving this code around in my PR, so I am wondering if I can just delete those lines (in which case I will fold line 385 into 377, as `destroyPet` is only used in that one place.

Just wanted to check in case those calls are doing some other side-effect like checking permissions.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -791,7 +791,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     for {
       petSAJson <- samDAO.getPetServiceAccountKeyForUser(projectName, RawlsUserEmail(userIdInfo.userEmail))
       petUserInfo <- gcsDAO.getUserInfoUsingJson(petSAJson)
-      _ <- samDAO.deleteUserPetServiceAccount(projectName, petUserInfo)
+      _ <- samDAO.deleteUserPetServiceAccount(projectName, ctx)
```
cahrens, line 794: @dvoet same question here-- should lines 792 and 793 be removed?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: Discussed this with Doug via Slack, needs to change to `ctx.copy(userInfo = petUserInfo)`. Fix in https://github.com/broadinstitute/rawls/pull/2095/commits/1c1121c139f9c50f2dc7afcc974b38249eda0371

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Fix in https://github.com/broadinstitute/rawls/pull/2095/commits/1c1121c139f9c50f2dc7afcc974b38249eda0371

PR #2095, opened 2022-10-25
[WOR-558] Delete landing zones on billing project deletion

cahrens:
Ticket: [WOR-558](https://broadworkbench.atlassian.net/browse/WOR-558)

Note that this PR is against the feature branch for landing zone creation.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -10,12 +10,15 @@ import org.broadinstitute.dsde.rawls.model.{
   ErrorReport,
   ErrorReportSource,
   RawlsBillingProject,
+  RawlsBillingProjectName,
   RawlsRequestContext,
+  SamBillingProjectActions,
   SamBillingProjectPolicyNames,
   SamBillingProjectRoles,
   SamPolicy,
   SamResourcePolicyName,
-  SamResourceTypeNames
+  SamResourceTypeNames,
+  UserInfo
 }
 import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport, StringValidationUtils}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
@@ -24,7 +27,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 /**
  * Knows how to provision billing projects with external cloud providers (that is, implementors of the
- * BillingProjectCreator trait)
+ * BillingProjectLifecycle trait)
  *
  * All billing projects are created following this algorithm:
  * 1. Pre-flight validation with a billing project creator. Right now, the creator is determined according
@@ -36,8 +39,8 @@ import scala.concurrent.{ExecutionContext, Future}
 class BillingProjectOrchestrator(ctx: RawlsRequestContext,
                                  samDAO: SamDAO,
                                  billingRepository: BillingRepository,
-                                 googleBillingProjectCreator: BillingProjectCreator,
-                                 bpmBillingProjectCreator: BillingProjectCreator,
+                                 googleBillingProjectLifecycle: BillingProjectLifecycle,
+                                 bpmBillingProjectLifecycle: BillingProjectLifecycle,
                                  config: MultiCloudWorkspaceConfig
 )(implicit val executionContext: ExecutionContext)
     extends StringValidationUtils
@@ -48,9 +51,9 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
    * Creates a "v2" billing project, using either Azure managed app coordinates or a Google Billing Account
    */
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
-    val billingProjectCreator = createProjectRequest.billingInfo match {
-      case Left(_)  => googleBillingProjectCreator
-      case Right(_) => bpmBillingProjectCreator
+    val billingProjectLifecycle = createProjectRequest.billingInfo match {
+      case Left(_)  => googleBillingProjectLifecycle
+      case Right(_) => bpmBillingProjectLifecycle
     }
     val billingProjectName = createProjectRequest.projectName
 
@@ -58,13 +61,13 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
       _ <- validateBillingProjectName(createProjectRequest.projectName.value)
 
       _ = logger.info(s"Validating billing project creation request [name=${billingProjectName.value}]")
-      _ <- billingProjectCreator.validateBillingProjectCreationRequest(createProjectRequest, ctx)
+      _ <- billingProjectLifecycle.validateBillingProjectCreationRequest(createProjectRequest, ctx)
 
       _ = logger.info(s"Creating billing project record [name=${billingProjectName}]")
       _ <- createV2BillingProjectInternal(createProjectRequest, ctx)
 
       _ = logger.info(s"Created billing project record, running post-creation steps [name=${billingProjectName.value}]")
-      creationStatus <- billingProjectCreator.postCreationSteps(createProjectRequest, config, ctx).recoverWith {
+      creationStatus <- billingProjectLifecycle.postCreationSteps(createProjectRequest, config, ctx).recoverWith {
         case t: Throwable =>
           logger.error(s"Error in post-creation steps for billing project [name=${billingProjectName.value}]")
           rollbackCreateV2BillingProjectInternal(createProjectRequest).map(throw t)
@@ -130,20 +133,67 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
         )
       )
     } yield {}
+
+  def deleteBillingProjectV2(projectName: RawlsBillingProjectName): Future[Unit] =
+    for {
+      _ <- samDAO
+        .userHasAction(SamResourceTypeNames.billingProject,
+                       projectName.value,
+                       SamBillingProjectActions.deleteBillingProject,
+                       ctx
+        )
+        .flatMap {
+          case true => Future.successful()
+          case false =>
+            Future.failed(
+              new RawlsExceptionWithErrorReport(
+                errorReport = ErrorReport(StatusCodes.Forbidden, "You must be a project owner.")
+              )
+            )
+        }
+      billingProfileId <- billingRepository.getBillingProfileId(projectName)
+      projectLifecycle = billingProfileId match {
+        case None    => googleBillingProjectLifecycle
+        case Some(_) => bpmBillingProjectLifecycle
+      }
+      _ <- billingRepository.failUnlessHasNoWorkspaces(projectName)
+      _ <- projectLifecycle.preDeletionSteps(projectName, ctx)
+      _ <- unregisterBillingProjectWithUserInfo(projectName, ctx.userInfo)
+    } yield {}
+
+  // This code also lives in UserService.
+  def unregisterBillingProjectWithUserInfo(projectName: RawlsBillingProjectName,
+                                           ownerUserInfo: UserInfo
+  ): Future[Unit] =
+    for {
+      _ <- billingRepository.deleteBillingProject(projectName)
+      _ <- samDAO
+        .deleteResource(SamResourceTypeNames.billingProject,
+                        projectName.value,
+                        ctx.copy(userInfo = ownerUserInfo)
+        ) recoverWith { // Moving this to the end so that the rawls record is cleared even if there are issues clearing the Sam resource (theoretical workaround for https://broadworkbench.atlassian.net/browse/CA-1206)
+        case t: Throwable =>
+          logger.warn(
+            s"Unexpected failure deleting billing project (while deleting billing project in Sam) for billing project `${projectName.value}`",
+            t
+          )
+          throw t
+      }
+    } yield {}
 }
 
 object BillingProjectOrchestrator {
   def constructor(samDAO: SamDAO,
                   billingRepository: BillingRepository,
-                  googleBillingProjectCreator: GoogleBillingProjectCreator,
-                  bpmBillingProjectCreator: BpmBillingProjectCreator,
+                  googleBillingProjectLifecycle: GoogleBillingProjectLifecycle,
+                  bpmBillingProjectLifecycle: BpmBillingProjectLifecycle,
                   config: MultiCloudWorkspaceConfig
   )(ctx: RawlsRequestContext)(implicit executionContext: ExecutionContext): BillingProjectOrchestrator =
     new BillingProjectOrchestrator(ctx,
                                    samDAO,
                                    billingRepository,
-                                   googleBillingProjectCreator,
-                                   bpmBillingProjectCreator,
+                                   googleBillingProjectLifecycle,
+                                   bpmBillingProjectLifecycle,
                                    config
     )
 
@@ -167,3 +217,5 @@ class ServicePerimeterAccessException(errorReport: ErrorReport) extends RawlsExc
 class GoogleBillingAccountAccessException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
 
 class LandingZoneCreationException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
+
+class BillingProjectDeletionException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -116,18 +116,16 @@ class BpmBillingProjectLifecycle(
 
   override def preDeletionSteps(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext): Future[Unit] =
     for {
-      _ <- billingRepository.getCreationStatus(projectName).flatMap {
+      _ <- billingRepository.getCreationStatus(projectName).map {
         case CreationStatuses.CreatingLandingZone =>
-          Future.failed(
-            new BillingProjectDeletionException(
-              RawlsErrorReport(
-                s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created"
-              )
+          throw new BillingProjectDeletionException(
+            RawlsErrorReport(
+              s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created"
             )
           )
-        case _ => Future.successful()
+        case _ => ()
       }
-      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
+      _ <- billingRepository.getLandingZoneId(projectName).map {
         case Some(landingZoneId) =>
           val landingZoneResponse = blocking {
             // Note that this actually just starts a landing zone deletion job (and thus returns quickly).
@@ -143,11 +141,9 @@ class BpmBillingProjectLifecycle(
               logger.info(
                 s"Deleting BPM-backed billing project ${projectName}, initiated deletion of landing zone ${landingZoneId} with jobID ${landingZoneResponse.getJobReport.getId}"
               )
-              Future.successful()
           }
         case None =>
           logger.warn(s"Deleting BPM-backed billing project ${projectName}, but no associated landing zone to delete")
-          Future.successful()
       }
     } yield {}
 }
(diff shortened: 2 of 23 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -218,12 +218,6 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         workspaceRec <- dataAccess.workspaceQuery.findByIdQuery(submissionRec.workspaceId).result.map(_.head)
       } yield (RawlsUserEmail(submissionRec.submitterEmail), workspaceRec)
 
-    def getPetSAUserInfo(googleProjectId: GoogleProjectId, submitterEmail: RawlsUserEmail): Future[UserInfo] =
```
cahrens, line 222: Changed to use [existing code](https://github.com/broadinstitute/rawls/blob/885608b490e9b1418f7edc5622fb7fc3bb716962/core/src/main/scala/org/broadinstitute/dsde/rawls/util/AuthUtil.scala#L12) in `AutUtil`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -791,7 +791,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     for {
       petSAJson <- samDAO.getPetServiceAccountKeyForUser(projectName, RawlsUserEmail(userIdInfo.userEmail))
```
cahrens, line 792: This could also be changed to use `AuthUtil`, but the code comments indicate that this code will all be deleted once workspace migration is complete (https://broadworkbench.atlassian.net/browse/CA-1118).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -160,7 +160,7 @@ object UserService {
     for {
       petSAJson <- samDAO.getPetServiceAccountKeyForUser(projectName, RawlsUserEmail(userIdInfo.userEmail))
```
cahrens, line 167: There is quite a bit of duplicated code here between `UserService` and `WorkspaceService`, but supposedly it can all be deleted once workspace migration is complete (https://broadworkbench.atlassian.net/browse/CA-1118).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -130,20 +133,67 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+        case None    => googleBillingProjectLifecycle
+        case Some(_) => bpmBillingProjectLifecycle
+      }
+      _ <- billingRepository.failUnlessHasNoWorkspaces(projectName)
+      _ <- projectLifecycle.preDeletionSteps(projectName, ctx)
+      _ <- unregisterBillingProjectWithUserInfo(projectName, ctx.userInfo)
+    } yield {}
+
+  // This code also lives in UserService.
+  def unregisterBillingProjectWithUserInfo(projectName: RawlsBillingProjectName,
```
cahrens, line 165: Is it worth trying to share this code? Currently this file does not have a dependency on UserService, and I would like to avoid introducing that.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
cahrens: This is just opportunistic cleanup that I noticed while trying to trace through some recent changes to use the SamClient (which caused me some rebasing pain).

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -94,6 +94,74 @@ object UserService {
       "roles/bigquery.jobUser" -> Set(s"group:${ownerGroupEmail.value}", s"group:${computeUserGroupEmail.value}")
     )
 
+  // TODO - once workspace migration is complete and there are no more v1 workspaces or v1 billing projects, we can remove this https://broadworkbench.atlassian.net/browse/CA-1118
+  def deleteGoogleProjectIfChild(projectName: RawlsBillingProjectName,
```
cahrens, line 104: Moved up to companion object to allow sharing from `GoogleBillingProjectLifecycle`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: This is cleanup from Sam client change PR.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -291,4 +292,202 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
+    when(
+      samDAO.deleteResource(SamResourceTypeNames.billingProject, billingProjectName.value, testContext)
+    ).thenReturn(Future.successful())
+
+    val billingRepository = mock[BillingRepository]
+    when(billingRepository.failUnlessHasNoWorkspaces(billingProjectName)(executionContext))
+      .thenReturn(Future.successful())
+    when(billingRepository.deleteBillingProject(billingProjectName)).thenReturn(Future.successful(true))
+
+    // Mock Azure project
```
cahrens, line 408: I should remove this comment now that this method can create either type of project.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -477,7 +477,7 @@ class UserServiceSpec
       ).thenReturn(Future.successful(Set(ownerIdInfo)))
       when(mockSamDAO.getPetServiceAccountKeyForUser(project.googleProjectId, ownerUserInfo.userEmail))
         .thenReturn(Future.successful(petSAJson))
-      when(mockSamDAO.deleteUserPetServiceAccount(project.googleProjectId, testContext))
+      when(mockSamDAO.deleteUserPetServiceAccount(project.googleProjectId, testContext.copy(userInfo = ownerUserInfo)))
```
cahrens, line 488: Related to Sam client fixes.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -99,4 +102,36 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
+                s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created"
+              )
+            )
+          )
+        case _ => Future.successful()
+      }
+      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
+        case Some(landingZoneId) =>
+          val landingZoneResponse = blocking {
+            workspaceManagerDAO.deleteLandingZone(UUID.fromString(landingZoneId), ctx)
```
aherbst-broad, line 133: Hopefully this is zippy, I suspect we may need to spawn another async job for this one as well (not in this PR).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -99,4 +102,36 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
+          Future.failed(
+            new BillingProjectDeletionException(
+              RawlsErrorReport(
+                s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created"
+              )
+            )
+          )
+        case _ => Future.successful()
+      }
+      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
```
blakery: Nit: If you change this from `.flatMap` to `.map`, you won't need to return `Future.successful()`

You're already throwing the exception directly when you get the errorReport, and you're blocking on the call to WSM to delete the landing zone, so you aren't really getting any use out of `.flatMap`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -99,4 +102,36 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
+        case Some(landingZoneId) =>
+          val landingZoneResponse = blocking {
+            workspaceManagerDAO.deleteLandingZone(UUID.fromString(landingZoneId), ctx)
+          }
+          val errorReport = Option(landingZoneResponse.getErrorReport)
+          errorReport match {
```
blakery: Nit: it would probably read a little cleaner to match directly on the `Option(landingZoneResponse.getErrorReport)`, instead of assigning it to a separate value, especially since you're immediately deconstructing it.
```scala
Option(landingZoneResponse.getErrorReport) match {
    case Some(errorReport) => ....
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
blakery: I like opportunistic cleanup 😀

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
blakery: Seems reasonable as is to me. The logic is basically just doing two things with separate resources - not like it's super complex.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -130,20 +133,67 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+                                           ownerUserInfo: UserInfo
+  ): Future[Unit] =
+    for {
+      _ <- billingRepository.deleteBillingProject(projectName)
+      _ <- samDAO
+        .deleteResource(SamResourceTypeNames.billingProject,
+                        projectName.value,
+                        ctx.copy(userInfo = ownerUserInfo)
+        ) recoverWith { // Moving this to the end so that the rawls record is cleared even if there are issues clearing the Sam resource (theoretical workaround for https://broadworkbench.atlassian.net/browse/CA-1206)
```
blakery, line 174: I don't understand the part about clearing the record - this doesn't appear to do anything aside from log the message and re-throw the exception

blakery (commented):
Looks good, just a question on `BpmBillingProjectLifecycle`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: @blakery so do I need `flatMap` at line 108, or can that one also be simplified? I'm again confused about map vs. flatMap and mostly was just excited to figure out a syntax that works.

cahrens: If I replace `flatMap` with `map` I won't be able to log the earning about the landing zone not existing. Perhaps not a big deal, but I thought it was good to note because it isn't "expected".

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -99,4 +102,36 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
+      _ <- billingRepository.getCreationStatus(projectName).flatMap {
+        case CreationStatuses.CreatingLandingZone =>
+          Future.failed(
+            new BillingProjectDeletionException(
+              RawlsErrorReport(
+                s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created"
+              )
+            )
+          )
+        case _ => Future.successful()
```
cahrens: Am I correct that here I don't want to replace `flatMap` with `map` because I do actually want to handle all the possible creation statuses (`map` maps None -> None, but that doesn't help me here)?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -99,4 +101,35 @@ class BpmBillingProjectCreator(billingRepository: BillingRepository,
...
+          }
+          Option(landingZoneResponse.getErrorReport) match {
+            case Some(errorReport) =>
+              throw new BillingProjectDeletionException(
+                RawlsErrorReport(StatusCode.int2StatusCode(errorReport.getStatusCode), errorReport.getMessage)
+              )
+            case None => Future.successful()
+          }
+        case None =>
+          logger.warn(s"Deleting azure-backed billing project ${projectName}, but no associated landing zone to delete")
```
aherbst-broad: Nit: These won't always be azure-backed billing projects (i.e,. when we start using BPM for GCP projects deleting those projects will always hit this line).

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -239,14 +246,108 @@ class BpmBillingProjectCreatorSpec extends AnyFlatSpec {
...
       testContext
     )
     ScalaFutures.whenReady(result.failed) { exception =>
-      exception.getMessage.contains(landingZoneErrorMessage)
+      exception shouldBe a[LandingZoneCreationException]
+      assert(exception.getMessage.contains(landingZoneErrorMessage))
+    }
+  }
+
+  behavior of "postCreationSteps"
```
aherbst-broad: Is this section testing postCreationSteps or preDeletionSteps?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: @aherbst-broad delete is async…. it will just start the job, so should be fast. We decided in the ticket that we are going "fire and forget", and not worry if the deletion ends up failing for some reason.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: True, but we will have to change the code at that point anyway because GCP projects likely will not have landing zones.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
cahrens: Definitely `preDeletionSteps`-- thanks, will fix.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: I will add a code comment that the operation is asynchronous, and we aren't tracking whether it ultimately succeeds.

aherbst-broad: Ah got it, thanks.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
cahrens: This comment was already in the code I copied over… I think they are saying that they delete the Sam record after deleting the project so that the project is deleted even if the Sam call fails.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
cahrens: Fixed

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: Fixed

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: Fixed

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: `flatMap` will not handle `None` or failures either. What `flatMap` will do, is take a success (as in a `Future` or `Try`) or value (in the case of an `Option`), and then your function returns a value of the same type.
eg:
```scala
val optPresent: Option[Int] = Some(1)
optPresent.flatMap(i => Some(i + 1)  // will be Some(2)

val optMissing: Option[Int] = None
optMissing.flatMap(i => Some(i + 1)  // will still be None
```

if you want to handle the `None`  for an option, you would either use `getOrElse`, or do a conditional check on `isEmpty`, or handle it in a match.
For a `Future` or `Try`, you'd use `recover`.

The reason this doesn't give you a compiler error, is because all your branches either throw an exception or explicitly return a future of the correct type
The reason it works like you expect, is because the call to `getCreationStatus` already throws an exception on None.

One thing you could do, to make it a bit more obvious, would be to have  `getCreationStatus` return an `Option`, and then handle it something like this:
```scala
for {
    status  <- billingRepository.getCreationStatus(projectName)
    _ = status match {
        case None =>  throw new RawlsException(s"Billing Project ${projectName.value} does not exist in Rawls database")
        case Some(CreationStatuses.CreatingLandingZone) =>
            throw new RawlsExceptionWithErrorReport( s"Billing project ${projectName.value} cannot be deleted because its landing zone is still being created")
        case Some(_) => () // do nothing
    }
...
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: The explanation above applies here too. 
You will never get `None` here, because `getLandingZoneId` throws an exception there - so you'd want `recover` or `recoverWith` to log it.

cahrens: @blakery I think this one is different-- None is possible in this case, meaning that the billing project existed, but the landingZoneId was None. The exception that is thrown by `billingRepository.getLandingZoneId` is when the billing project doesn't exist.

The type of `landingZoneId` is Option[String], whereas the type of `status` is CreationStatuses.CreationStatus.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: OK, I made this change, but I don't understand how it's any better. In this case, I wasn't handling None here because it's not actually an option-- I was just using Future.successful for other statuses.

cahrens: We talked, and I misunderstood what Blake was suggested. New code has been pushed up.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: We talked, and I misunderstood what Blake was suggested. New code has been pushed up.

blakery (approved):

PR #2097, opened 2022-10-27
[AJ-659] add default value for new ignoreEmptyColumns submission column

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-659
Liquibase change to put a default value for ignoreEmptyOutputs to fix handling of existing columns

See test failure ticket: https://broadworkbench.atlassian.net/browse/QA-2033

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet id="add_empty_columns" author="jsafer" logicalFilePath="dummy">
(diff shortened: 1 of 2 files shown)


calypsomatic (approved):

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet id="add_empty_columns" author="jsafer" logicalFilePath="dummy">
+        <addDefaultValue tableName="SUBMISSION" columnName="IGNORE_EMPTY_OUTPUTS" columnDataType="BOOLEAN"
+                         defaultValueBoolean="false" />
```
MatthewBemis: I think there is one more step needed to backfill the existing rows. There are a couple of options:

A) Make the column non-nullable before setting the default value. I believe this will instruct this particular changeset to also set the value for pre-existing rows to `defaultValueBoolean`. With the column set to nullable (as in it's current state), I think it will only set the default for new rows, not existing rows
B) Add another changeset to explicitly backfill all existing rows

I imagine the intent here is to have a value set for each existing and new row, so option A is probably the preferred approach since it enforces that these columns aren't null, which the code isn't able to handle.

jordansafer: I see, makes sense!

MatthewBemis (approved):
lgtm!

davidangb (approved):
Looks good to me, assuming this passes tests

jordansafer:
jenkins retest

PR #2098, opened 2022-10-28
fix liquibase syntax

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-659
Syntax fix for the latest liquibase changeset!

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
     <changeSet id="make_empty_columns_nonnull" author="jsafer" logicalFilePath="dummy">
-        <addNotNullConstraint tableName="SUBMISSION" columnName="IGNORE_EMPTY_OUTPUTS" columnDataType="BOOLEAN"
-                              defaultNullValue="0"/>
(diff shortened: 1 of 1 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
-    <changeSet id="add_empty_columns" author="jsafer" logicalFilePath="dummy">
-        <addNotNullConstraint tableName="SUBMISSION" columnName="IGNORE_EMPTY_OUTPUTS" columnDataType="BOOLEAN"
-                              defaultNullValue="false"/>
```
jordansafer, line 5: The issue was that `defaultNullValue="false"` was trying to set the boolean value to a string value, false. Liquibase added some type autodetection support in release 4.5, but we are on the 3.xxx major version update schedule, so we don't have access to these changes. As it stands, the addNotNullConstraint command will always attempt to use a string. This means that `defaultNullValue="0"` will also fail, since liquibase will attempt to use the string "0" as opposed to the 1 bit value 0.  https://github.com/liquibase/liquibase/issues/1054

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20221027_add_default_ignore_empty_outputs.xml
```diff
@@ -1,7 +1,9 @@
...
+    <changeSet id="make_empty_columns_nonnull" author="jsafer" logicalFilePath="dummy">
+        <sql>
+            UPDATE SUBMISSION SET IGNORE_EMPTY_OUTPUTS = 0 WHERE IGNORE_EMPTY_OUTPUTS IS NULL;
+            ALTER TABLE SUBMISSION MODIFY IGNORE_EMPTY_OUTPUTS BIT(1) NOT NULL;
```
davidangb: add `DEFAULT 0` to this?

jordansafer: added-

calypsomatic (approved):

davidangb (approved):

PR #2105, opened 2022-11-01
WOR-525 part 2

blakery:
[WOR-525](https://broadworkbench.atlassian.net/browse/WOR-525)<Put notes here to help reviewer understand this PR>

Check status of landing zone jobs when a user queries a billing project with status `CreationStatuses.CreatingLandingZone`, and update the project when the job is complete.

This was changed to be triggered on user query instead of a background job because of issues in accessing data outside of a user request.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -246,14 +254,66 @@ class UserService(
...
+            updateLandingZoneFailure("Landing Zone result marked as successful, but no landing zone returned")
+          case JobReport.StatusEnum.SUCCEEDED => updateLandingZoneSuccess(result.getLandingZone.getId)
+        }
+      case Some(result) if result.getLandingZone != null && result.getLandingZone.getId != null =>
+        updateLandingZoneSuccess(result.getLandingZone.getId)
+      case Some(result) if result.getErrorReport != null =>
+        updateLandingZoneFailure(result.getErrorReport.getMessage)
+      // no job report, no landing zone, and no error report
+      // return project with status as error, but don't update anything in the database, so it retries in the future, and no data is lost for now
+      case Some(_) =>
```
blakery: I was getting kind of paranoid at this point, but.... 🤷

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
```diff
@@ -62,4 +62,11 @@ class BillingRepository(dataSource: SlickDataSource) {
       )
     }
 
+  /**
+    * Whenever this is called, the status is implied to be CreationStatus.Ready
+    */
+  def updateLandingZoneId(projectName: RawlsBillingProjectName, landingZoneId: UUID): Future[Int] =
```
cahrens, line 64: I think we want this to take the status as an argument. For instance, you will want to set the billing project to ERROR if the landing zone creation fails.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -285,6 +285,15 @@ trait RawlsBillingProjectComponent {
         .withProjectName(projectName)
         .setCreationStatus(status, message)
 
+    /**
+      * Whenever this is called, the status is implied to be CreationStatus.Ready
+      */
+    def updateLandingZone(projectName: RawlsBillingProjectName, landingZoneId: UUID): WriteAction[Int] =
```
cahrens, line 291: Again, I think we want the status as an argument.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
```diff
@@ -98,7 +98,7 @@ class WorkspaceResourceMonitor(
         })
     )
   } yield
-    if (jobResults.reduce((a, b) => a && b)) { // true iff all runners for the given job completed successfully
+    if (jobResults.forall(a => a)) { // true iff all runners for the given job completed successfully
```
cahrens, line 101: Why did you make this change? And does it mean that the comment should be updated?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -246,14 +255,77 @@ class UserService(
...
+    * For billing projects in the status of CreationStatuses.CreatingLandingZone,
+    * retrieve the landing zone status and update the project appropriately.
+    * Deletes the record tracking the landing zone job, if the job is completed.
+    */
+  def updateLandingZoneStatus(billingProject: RawlsBillingProject): RawlsBillingProject = {
+    // if there's no landing zone creation in progress, this is a no-op
+    if (billingProject.status != CreationStatuses.CreatingLandingZone) return billingProject
+    val record = Await.result(
+      workspaceResourceRecordDao.selectByBillingProject(billingProject.projectName).map(_.headOption),
+      DurationInt(30).seconds
```
cahrens: Why the 30 second interval?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
cahrens: OK, I see that you are setting the status to ERROR in another way.

Perhaps just rename the method to something clearer? Like `landingZoneCreationCompleted` or …. I don't know, leave it as `updateLandingZoneId`, and use `updateCreationStatus` to set the status? One of the things I'm thinking about is that the Landing Zone team has a ticket to return the landing zone ID when creation is first kicked off, at which point we can store it then and wouldn't want this side effect of the creation status being set.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -27,6 +27,7 @@ import org.mockito.{ArgumentMatchers, Mockito}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatestplus.mockito.MockitoSugar.mock
 
+import java.util.UUID
```
cahrens: I imagine this is rebase cruft.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -55,7 +60,7 @@ class UserServiceSpec
   val defaultMockGcsDAO: GoogleServicesDAO = new MockGoogleServicesDAO("test")
   val defaultMockServicePerimeterService: ServicePerimeterService = mock[ServicePerimeterService](RETURNS_SMART_NULLS)
   val defaultBillingProfileManagerDAO: BillingProfileManagerDAO = mock[BillingProfileManagerDAO](RETURNS_SMART_NULLS)
-
+//  val defaultWorkspaceManagerDAO: WorkspaceManagerDAO = mock[WorkspaceManagerDAO](RETURNS_SMART_NULLS)
```
cahrens: Remove line.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -1238,6 +1247,249 @@ class UserServiceSpec
...
+      CreationStatuses.Ready,
+      None,
+      None,
+      azureManagedAppCoordinates = Some(AzureManagedAppCoordinates(null, null, null)),
+      billingProfileId = Some(billingProfile.getId.toString)
+    )
+    val userService = spy(getUserService())
+    val result = userService.mapCloudPlatform(project, Some(billingProfile), Set())
+    result shouldBe RawlsBillingProjectResponse(Set(), project, CloudPlatform.AZURE)
+    verify(userService).updateLandingZoneStatus(project) // update landing status was called, but it was a no-op
```
cahrens: Is there a way to verify that it was a no-op?

cahrens:
This looks good-- I have a few questions/comments, and it would be good to add test coverage for the exception handling.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
blakery: The behavior is exactly the same, except that it won't cause an exception on on empty collection

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: The actual number arbitrary. 
You have to specify a duration though, and Duration.Inf is probably bad form in real code. 
This is just a db call, so it's generally basically instantaneous, and 30 seconds seemed like enough time to be able to assume there's an issue with the db.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
blakery: I'm using the default mocks, so it doing anything that's not a no-op will make it blow up

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
cahrens: Yep, Jesus has a PR up now to include the landingZoneId in the creation response. I can fold that in with my open PR, after yours has merged. So definitely it would be good to make setting the landing zone ID only set the ID.

https://github.com/DataBiosphere/terra-workspace-manager/pull/886/files

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -285,6 +285,15 @@ trait RawlsBillingProjectComponent {
         .withProjectName(projectName)
         .setCreationStatus(status, message)
 
+    /**
+      * Whenever this is called, the status is implied to be CreationStatus.Ready
```
aherbst-broad: 🤔 I don't follow this comment given the code below -- how does this method affect the billing project status?

cahrens:
@blakery I'm confused on the commits why it includes the commits that are already on the feature branch. It would be good to squash all your commits into a single one before merging to make things easier (it would probably would have eased your rebasing as well).

blakery:
Squashing commits is part of merging. 
The commits are on my branch because I didn't create a new one after the previous merge. So nothing was changed from the perspective of my branch when it was merged in. I think those commits just get re-squashed.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
aherbst-broad: Is there a reason this needs to be a synchronous await? If not, it'd be good to retain consistency with the rest of the codebase, where we don't generally await futures outside of tests but rather wrap them in a for comprehension or flatmap.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -246,14 +255,78 @@ class UserService(
...
+          updateLandingZoneFailure(result.getErrorReport.getMessage)
+        // no job report, no landing zone, and no error report
+        // return project with status as error, but don't update anything in the database, so it retries in the future, and no data is lost for now
+        case Some(_) =>
+          billingProject.copy(status = CreationStatuses.Error,
+                              message = Some("Unable to retrieve landing zone results")
+          )
+      }
+    catch {
+      case e: ApiException =>
```
aherbst-broad: Should we allow 5xx errors to be retry-able and thus not push the billing project status to error? That is, if WSM is temporarily unavailable it would be good to be resilient to that situation and allow the billing project to retry.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectCreatorSpec.scala
```diff
@@ -68,6 +68,7 @@ class BpmBillingProjectCreatorSpec extends AnyFlatSpec {
     )
     when(bpm.listManagedApps(ArgumentMatchers.eq(coords.subscriptionId), ArgumentMatchers.eq(testContext)))
       .thenReturn(Seq())
+
```
aherbst-broad: Nitpick: typo, may as well remove this to keep the blame clean.

aherbst-broad (commented):
Looking good, had some questions about the await.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
blakery: I just forgot to remove that comment after making some changes (the changes were both farther up and lower down on the call stack). I'll remove it.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: It doesn't persist the error state, so it will re-attempt in future requests.
I figured since this was a user request that UIs could be waiting on, it might be make sense to avoid spending time in retries. It probably does make sense for 5XX though.

blakery: I suggest that even for 5XXs, this is not worth it here:

- The existing trait that provides retry functionality elsewhere requires an `ActorSystem`
- Even if it would work, shoe-horning it in here makes for really convoluted syntax (I just spent way too much time trying to make it work - it was kind of hideous)
- Re-implementing it with decent testing, correct config of timeouts and n-retries, etc would take more time than we probably want to spend

After writting the above, I found the Retry that doesn't require the ActorSystem (it's in a companion object at the bottom of the file). I have a reasonable solution that uses it, but it has two drawbacks:
1. it extends the unit tests by however long you set your timeout
2. there's no way of only retrying some exceptions - it's all or nothing

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -285,6 +285,15 @@ trait RawlsBillingProjectComponent {
         .withProjectName(projectName)
         .setCreationStatus(status, message)
 
+    /**
```
cahrens: @blakery this comment still needs to be removed.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -254,6 +265,76 @@ class UserService(
       RawlsBillingProjectResponse(roles, project.copy(message = message, status = CreationStatuses.Error))
   }
 
+  def updateLandingZoneStatus(billingProject: Option[RawlsBillingProject]): Future[Option[RawlsBillingProject]] =
+    billingProject match {
+      case None    => Future.successful(billingProject)
```
cahrens, line 270: I don't quite understand… how would we get into this with billingProject being None?

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -1266,6 +1311,233 @@ class UserServiceSpec
       .get shouldEqual CloudPlatform.UNKNOWN.toString
   }
 
+  /* it should "not attempt to update landing zones for non-azure projects" in {
```
cahrens: @blakery why is this one commented out?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: `getBillingProject` uses a method that returns an optional
`updateLandingZoneStatus` returns a future now
Every way I tried, the `<-` in `getBillingProject` kept getting interpreted as the .map of the optional, not the future
I could make it work with some really convoluted syntax, but overloading the function like this was the least confusing way I could find to implement it

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
blakery: Because I didn't clean up after myself and remove it
And because it was happening as a side effect, before I updated `updateLandingZoneStatus` to use a future, and it was more documentation then a condition that needed to be tested specifically

cahrens (approved):
Note the scalafmt issue though.

PR #2116, opened 2022-11-08
Ignore flaky test (WOR-611).

cahrens:
Ticket: [WOR-611](https://broadworkbench.atlassian.net/browse/WOR-611)
This unit test was just introduced in https://github.com/broadinstitute/rawls/pull/2113 and is flaky. It always passes for me locally, but has failed twice in ci, including on the [merge commit](https://github.com/broadinstitute/rawls/actions/runs/3421947562/jobs/5698747428#step:8:9960).

Disabling the test while I figure out what the issue is.
---

**PR checklist**

- [ ] Include the JIRA issue number in the PR description and title
- [ ] Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green, including CI tests
- [ ] **Squash commits and merge** to develop (branches are automatically deleted after merging)
- [ ] Inform other teams of any substantial changes via Slack and/or email

PR #2113, opened 2022-11-07
Delete unreferenced BPM billing profile on project deletion (WOR-577).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-577
This change allows Terra managed app deployments to be reused once the Rawls billing project is deleted (through Swagger/curl).

Log output from manual testing:

````
rawls [INFO] [16:24:22.710] [scala-execution-context-global-56] o.b.d.r.b.BpmBillingProjectLifecycle - Deleting BPM-backed billing project RawlsBillingProjectName(CARNov7), initiated deletion of landing zone a30683c0-0cb7-42f3-94fc-8ce4a701c087 with jobID b6084e84-b353-42b4-9f88-ed08620d9fe2
rawls [INFO] [16:24:23.676] [scala-execution-context-global-56] o.b.d.r.b.BpmBillingProjectLifecycle - Deleting BPM-backed billing project RawlsBillingProjectName(CARNov7), deleting billing profile record aa3e1d57-94c1-45e8-a83a-30bb6e4bfbb4
````

---

**PR checklist**

- [x] Include the JIRA issue number in the PR description and title
- [ ] Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green, including CI tests
- [ ] **Squash commits and merge** to develop (branches are automatically deleted after merging)
- [ ] Inform other teams of any substantial changes via Slack and/or email

PR #2117, opened 2022-11-09
Fix flaky test (WOR-611).

cahrens:
Ticket: <[WOR-611](https://broadworkbench.atlassian.net/browse/WOR-611)>

Related to https://github.com/broadinstitute/rawls/pull/2113 (and test temporarily skipped in https://github.com/broadinstitute/rawls/pull/2116).

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -145,7 +145,7 @@ class BpmBillingProjectLifecycle(
         case None =>
           logger.warn(s"Deleting BPM-backed billing project ${projectName}, but no associated landing zone to delete")
       }
-      _ <- billingRepository.getBillingProfileId(projectName).map {
+      _ <- billingRepository.getBillingProfileId(projectName).flatMap {
         case Some(billingProfileId) =>
           val numOtherProjectsWithProfile = for {
             allProjectsWithProfile <- billingRepository
@@ -167,6 +167,7 @@ class BpmBillingProjectLifecycle(
           logger.warn(
             s"Deleting BPM-backed billing project ${projectName}, but no associated billing profile record to delete"
           )
+          Future.successful()
       }
     } yield {}
 }
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
@@ -384,7 +384,7 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
     }
   }
 
-  ignore should "delete the billing profile if no other project references it" in {
+  it should "delete the billing profile if no other project references it" in {
     val billingProfileId = UUID.randomUUID()
     val billingProjectName = RawlsBillingProjectName(s"project_for_${billingProfileId}")
 


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -116,16 +116,16 @@ class BpmBillingProjectLifecycle(
 
   override def preDeletionSteps(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext): Future[Unit] =
     for {
-      _ <- billingRepository.getCreationStatus(projectName).map {
+      _ <- billingRepository.getCreationStatus(projectName).flatMap {
```
ehigham: These two are equivalent and a matter of personal taste. I think Blake suggested not using `flatMap` here as
- you didn't need to call another function that returned a `Future` and suggested instead could use `map` and return `()`.
- `flatMap`-ing in a `for`-comp is a bit odd as each subsequent statement (`res <- expr`) is actually syntactic sugar for `flatMap`.

ehigham: As an aside, I prefer the use of `flatMap` because it captures the idea that the following expression has side effects - namely, you're throwing an exception. In Scala, throwing in the in the context of `Future` with `map` or `flatMap`, however, is perfectly legal.
I could go on but I'm not sure how productive it'll be for fear it'll just lead to more confusion!

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -141,11 +141,13 @@ class BpmBillingProjectLifecycle(
...
                 s"Deleting BPM-backed billing project ${projectName}, initiated deletion of landing zone ${landingZoneId} with jobID ${landingZoneResponse.getJobReport.getId}"
               )
+              Future.successful()
           }
         case None =>
           logger.warn(s"Deleting BPM-backed billing project ${projectName}, but no associated landing zone to delete")
+          Future.successful()
       }
-      _ <- billingRepository.getBillingProfileId(projectName).map {
+      result <- billingRepository.getBillingProfileId(projectName).flatMap {
```
ehigham: You don't need to store this in a variable and yield it. The bug was that this was a `Future[Future[Unit]]` rather than a `Future[Unit]`.

ehigham (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: Hmmm, I unfortunately don't see the difference between lines 119 and 150. In both cases, the result of the function call is a Future. The return type of  `getCreationStatus` is `Future[CreationStatus]`, whereas the return type of `getBillingProfileId` is ` Future[Option[String]]`. I was hoping for a general rule like "use flatMap whenever you are dealing with Futures". Is the issue here that the call to `getBillingProfileId` is the last call in the for comprehension??

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -125,7 +125,7 @@ class BpmBillingProjectLifecycle(
           )
         case _ => ()
       }
-      _ <- billingRepository.getLandingZoneId(projectName).map {
+      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
```
cahrens: This code is pretty much identical to the code at line 150 (return type of `Future[Option[String]]`, so I'm guessing this should be a `flatMap`? Or perhaps it doesn't because it's not the last one in the for comprehension?

Still confused….

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -141,11 +141,13 @@ class BpmBillingProjectLifecycle(
...
+              Future.successful()
           }
         case None =>
           logger.warn(s"Deleting BPM-backed billing project ${projectName}, but no associated landing zone to delete")
+          Future.successful()
       }
-      _ <- billingRepository.getBillingProfileId(projectName).map {
+      _ <- billingRepository.getBillingProfileId(projectName).flatMap {
         case Some(billingProfileId) =>
           val numOtherProjectsWithProfile = for {
```
ehigham, line 152: `numOtherProjectsWithProfile: Future[Int]`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: I find it helpful to substitute types into the signatures of `map` and `flatMap`.

Here's the thing we want to use, wrapped up in a `Future`:
```scala
billingRepository.getBillingProfileId(projectName) : Future[Option[String]]`
```

In the case where there's `Some(profileId)`, we want to perform some async computation to fetch all the the other projects that share that profile and delete the profile if none exist. this function has the following signature `Option[String] -> Future[Unit]`. Why? Because `getBillingProjectsWithProfile` is async and returns `Future`.

Let's substitute this into the signatures for `map` and `flatMap`, starting with `map`:
```scala
// in the definition of `trait Future[A]`
def map[B](f: A -> B): Future[B]

substitute `A`:
def map[B](f: Option[String] -> B): Future[B]

substitute `f` signature to get `B`:
def map[Future[Unit]](Option[String] -> Future[Unit]): Future[Future[Unit]]
```
And there's our problem - our Futures are interleaved. `Future` is async and runs eagerly in its own thread. If we wait for the outer future to be done, there's no guarantee that the inner future will be done. We need to synchronize/compose these effects so that they're done sequentially. For that, we use `flatMap`: 
```scala
// in the definition of `trait Future[A]`
def flatMap[B](f: A -> Future[B]): Future[B]

substitute `A`:
def flatMap[B](f: Option[String] -> Future[B]): Future[B]

substitute `f` signature to get `B`:
def flatMap[Unit](Option[String] -> Future[Unit]): Future[Unit]
```
Thus, when we wait for the result to be done, we can guarantee that the effects from our continuation, `f`, have been performed. The key observation is that if we want to do anything asynchronous with our value boxed up in a `Future`, use `flatMap`, otherwise use `map`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: I meant to post my long reply to this. See below - hope it helps :)

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: Thanks @ehigham! I think I mostly get it at the moment (though I'm a bit confused why "A" in the above isn't Future[Option[String]]). But I understand the general principle at the end-- "The key observation is that if we want to do anything asynchronous with our value boxed up in a Future, use flatMap, otherwise use map".

cahrens:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: I simplified the code back to the minimum necessary changes. Still interested at some point in discussing the alternative of moving the inner asynchronous call out, just for my own edification.

ehigham (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: Sure - we can pair on it when you're free if you like.

blakery (approved):

PR #2122, opened 2022-11-17
PROD-750: Add heartbeat and liveness checks to submission monitor

cjllanwarne:
Per PROD-650, we're still not sure **why** the submission monitor actor is getting jammed up, but we do think we know which function it's happening in (technically, which function gets run but then doesn't log that it completed, and doesn't get triggered again - at least not for a very long time). So - we can at least catch the situation when it happens and log it. ~And since we're there, we might as well restart the actor too~.

This should:
- Let us count how often this is happening
- Give us sentry alerts every time this happens
- ~Give the monitor a chance to reset its state by being killed and restarted~

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -135,11 +135,11 @@ class SubmissionMonitorActor(val workspaceName: WorkspaceName,
       checkCurrentWorkflowStatusCounts(true) pipeTo parent
     case SubmissionMonitorHeartbeat =>
       val secondsSinceLastMonitorPass = OffsetDateTime.now().toEpochSecond - lastMonitorPass.toEpochSecond
-      logger.debug(s"submission monitor heartbeat. Last monitor pass (or actor startup): ${lastMonitorPass} ($secondsSinceLastMonitorPass ago)")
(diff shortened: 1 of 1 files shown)


dvoet:
If the query does not finish, why do we think doing it again will? I think there is a danger of this gobbling up all the database threads: 

1. query db
2. query doesn't finish in time so start new query
3. goto 2

davidangb (approved):
this looks pretty good to me, and I'm giving it a :+1:.

Up to you - you might also want to issue a `setReceiveTimeout` and add a case for `akka.actor.ReceiveTimeout`. See:
* https://github.com/broadinstitute/rawls/blob/01300fa09496a08306fd1f9500c8f78427631baa/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala#L59
* https://github.com/broadinstitute/rawls/blob/01300fa09496a08306fd1f9500c8f78427631baa/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/EntityStatisticsCacheMonitor.scala#L87

note that the technique to restart the actor in EntityStatisticsCacheMonitor is really untrustful of the supervisor - your reliance on the supervisor here is more graceful.

cjllanwarne:
@dvoet would you recommend not throwing the exception and just logging that this has happened? Or moving to aborting? Do you have any sense which query it is that's taking too long?

dvoet:
I don't know what is going on. I would try innotop when this is happening to identify if there is a query hanging. See the [rawls play book](https://docs.google.com/document/d/1bR5A91dcXP0QTCnJolszF3oIArA3jy-DZPANv6pyEek/edit?usp=sharing), search for innotop

kpierre13 (approved):

aednichols (approved):
Nice!

PR #2130, opened 2022-11-21
WOR-613

blakery:
[WOR-613](https://broadworkbench.atlassian.net/browse/WOR-613)

Adds a new Sam endpoint to get creds for the user's default pet
Moves the landing zone status update back to the monitor on the back-end (creating LandingZoneCreationStatusRunner)
enables the monitor

---

**PR checklist**


.github/workflows/rawls-build-tag-publish-and-run-tests.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
           # manually recalculate b/c env context is broken https://github.com/actions/runner/issues/480
-          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ github.run_id }}-${{ matrix.terra-env }}", "version-template": "${{ matrix.terra-env }}", "custom-version-json": "${{ needs.rawls-build-tag-publish-job.outputs.custom-version-json }}" }'
+          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt}}-${{ matrix.terra-env }}", "version-template": "${{ matrix.terra-env }}", "custom-version-json": "${{ needs.rawls-build-tag-publish-job.outputs.custom-version-json }}" }'
```
cahrens: @blakery this commit has merged now, so you should rebase your PR on the head of develop to pick up this fix.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -528,6 +528,15 @@ class HttpSamDAO(baseSamServiceURL: String, serviceAccountCreds: Credential)(imp
       callback.future
     }
 
+  override def getUserArbitraryPetServiceAccountKey(userEmail: String): Future[String] =
+    retry(when401or5xx) { () =>
+      val callback = new SamApiCallback[String]("getUserArbitraryPetServiceAccountKey")
+
+      googleApi(rawlsSAContext).getUserArbitraryPetServiceAccountKeyAsync(userEmail, callback)
+
+      callback.future
```
cahrens, line 537: Do you know what `callback.future` does here?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
```diff
@@ -13,19 +14,26 @@ object WorkspaceManagerResourceMonitorRecord {
     type JobType = Value
     val AzureLandingZoneResult: Value = Value("AzureLandingZoneResult")
   }
+
+  implicit sealed class JobStatus(val complete: Boolean)
+
+  case object Complete extends JobStatus(true)
+
+  case object Incomplete extends JobStatus(false)
```
ehigham, line 22: This looks a bit fishy to me. Your sum type already encodes whether or not the job's complete.
Would it make sense to encode "finished" as an `Option[Instant]` and an failure message as`Option[String]` on the job record itself?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
```diff
@@ -92,7 +92,7 @@ class WorkspaceResourceMonitor(
     jobResults <- Future.sequence(
       registeredRunners
         .getOrElse(job.jobType, List())
-        .map(_.run(job).recover { case t =>
+        .map(_.run(job).map(_.complete).recover { case t =>
```
cahrens, line 95: What does this do?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+  samDAO: SamDAO,
+  workspaceManagerDAO: WorkspaceManagerDAO,
+  billingRepository: BillingRepository,
+  gcsDAO: GoogleServicesDAO
+) extends WorkspaceManagerResourceJobRunner
+    with LazyLogging {
+
+  override val jobType: JobType = JobType.AzureLandingZoneResult
+
+  def failureMessage(result: AzureLandingZoneResult): Some[String] = {
```
ehigham: This file has a lot of noise from null checking. In Scala (or any fp language) we shouldn't be using null; instead use `Option` and chain off that. If this is scrutinizing a java type with nulls, then [those should be parsed out into a a scala case class.](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

ehigham: Why `Some[String]`? Why not just `String`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+          s"Job to monitor AzureLandingZoneResult for billing project $billingProjectName created with id ${job.jobControlId} but no user email set"
+        )
+        val errorMsg =
+          s"Unable to update ${billingProjectName.value} with landing zone status because no user email is stored on monitoring job"
+        billingRepository.updateCreationStatus(billingProjectName, CreationStatuses.Error, Some(errorMsg))
+        return Future.successful(Complete)
+    }
+    getUserCtx(userEmail)
+      .map(userCtx =>
+        Try(workspaceManagerDAO.getCreateAzureLandingZoneResult(job.jobControlId.toString, userCtx)) match {
```
cahrens, line 69: How does this differ from doing try/catch?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+            billingRepository.updateCreationStatus(billingProjectName, CreationStatuses.Error, message)
+            Incomplete
+          case Success(result) if result.getJobReport != null =>
+            result.getJobReport.getStatus match {
+              // the job just isn't done yet - return indicates this runner isn't finished with the job so it's rescheduled
+              case JobReport.StatusEnum.RUNNING => Incomplete
+              // a normal success
+              case JobReport.StatusEnum.SUCCEEDED if result.getLandingZone != null =>
+                billingRepository.updateCreationStatus(billingProjectName, CreationStatuses.Ready, None)
+                billingRepository.updateLandingZoneId(billingProjectName, result.getLandingZone.getId)
```
cahrens: I don't think you need to worry about updating the landing zone ID-- we set it at creation time and there is no reason it should change.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+                Complete
+              // set the error, and indicate this runner is finished with the job
+              case _ =>
+                billingRepository
+                  .updateCreationStatus(billingProjectName, CreationStatuses.Error, failureMessage(result))
+                Complete
+            }
+          // this case is only hit if the job report is null but the landing zone is present in the response
+          case Success(result) if result.getLandingZone != null && result.getLandingZone.getId != null =>
```
cahrens: I'm confused about this case. How would you know it's complete? And I don't think we need to worry about updating the ID.

(FWIW, the ID may always be present now)

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+    val msg = () match {
+      case _ if result.getErrorReport != null && result.getErrorReport.getMessage != null =>
+        result.getErrorReport.getMessage
+      case _ if result.getJobReport != null =>
+        result.getJobReport.getStatus match {
+          case JobReport.StatusEnum.FAILED => "Failure Reported, but no errors returned"
+          case JobReport.StatusEnum.SUCCEEDED if result.getLandingZone == null =>
+            "Result marked as success, but no landing zone returned"
+          case JobReport.StatusEnum.RUNNING | JobReport.StatusEnum.SUCCEEDED =>
+            "Invalid Failure" // FIXME: improve message (we should never get into this state)
```
ehigham: Then throw or assert that this shouldn't be hit?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+          case null => "No status reported"
+        }
+      // no job report, no error report (or at least no error report message), also no lz (since that would still be a success)
+      case _ => "No landing zone, job report, or errors returned from landing zone creation"
```
ehigham: So is this an illegal state? How would we get into this situation?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+  workspaceManagerDAO: WorkspaceManagerDAO,
+  billingRepository: BillingRepository,
+  gcsDAO: GoogleServicesDAO
+) extends WorkspaceManagerResourceJobRunner
+    with LazyLogging {
+
+  override val jobType: JobType = JobType.AzureLandingZoneResult
+
+  def failureMessage(result: AzureLandingZoneResult): Some[String] = {
+    val msg = () match {
```
ehigham: Why pattern match on `Unit`? Why not use a normal if statement or `Option` chaining?

ehigham (commented):
unsolicited feedback...

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
cahrens: I see now… the error handling is just for the single `getCreateAzureLandingZoneResult` call.  Or actually, is it for the `map`?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+            billingRepository.updateCreationStatus(billingProjectName, CreationStatuses.Error, failureMessage(result))
+            Complete
+          // no landing zone, no error report, and no job report: set job status to failure but retry in the future
+          case Success(result) =>
+            billingRepository.updateCreationStatus(billingProjectName, CreationStatuses.Error, failureMessage(result))
+            Incomplete
+        }
+      )
+      .recover { case t: Throwable =>
+        logger.error(
```
cahrens: Is this the correct error message? I thought the recover was associated with the Try.

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -1794,7 +1794,7 @@ class SubmissionMonitorSpec(_system: ActorSystem)
         })).flatten
         submissionList.forall(_.status == SubmissionStatuses.Done.toString) && submissionList.length == numSubmissions
       },
-      max = 30 seconds,
+      max = 60 seconds,
```
cahrens: This is a known sporadic test failure-- we may not want to increase the timeout interval, but hopefully the JIRA ticket will provide more information (https://broadinstitute.slack.com/archives/C03BCARP494/p1669048189832469).

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunnerSpec.scala
```diff
@@ -0,0 +1,455 @@
...
+      .getUserCtx(ArgumentMatchers.eq(userEmail))(ArgumentMatchers.any())
+
+    whenReady(runner.run(monitorRecord))(_ shouldBe WorkspaceManagerResourceMonitorRecord.Incomplete)
+    verify(billingRepository).updateCreationStatus(
+      ArgumentMatchers.eq(billingProjectName),
+      ArgumentMatchers.eq(CreationStatuses.Error),
+      ArgumentMatchers.any[Some[String]]()
+    )
+  }
+  // TODO: test throwing a random exception from somewhere weird where it wouldn't be handled, like the failure message builder,
```
cahrens: Is this something you still want to do?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunnerSpec.scala
```diff
@@ -0,0 +1,455 @@
...
+    val ctx = mock[RawlsRequestContext]
+    val billingProjectName = RawlsBillingProjectName(UUID.randomUUID().toString)
+    val monitorRecord = WorkspaceManagerResourceMonitorRecord(
+      UUID.randomUUID(),
+      WorkspaceManagerResourceMonitorRecord.JobType.AzureLandingZoneResult,
+      None,
+      Some(billingProjectName.value),
+      Some(userEmail),
+      new Timestamp(Instant.now().toEpochMilli)
+    )
```
cahrens: It seems like some of this setup could be moved to a beforeEach or helper method to make the individual test cases easier to read.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunnerSpec.scala
blakery: No: the structure changed so that it's irrelevant - I'll remove the comment

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunnerSpec.scala
blakery: You're probably correct on that. 
I'm generally averse to setting up mocks like that, since it often degrades into a series of optional args to customize them; but here there's enough repetition to make it worth it

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
cahrens: I'm not sure we should consider this a success case-- how do we know the creation is complete?

blakery:
I restructured the error handling to be clearer and simplified it since we don't need to update the landing zone id.
I think the changes should address most of the earlier comments.

blakery:
there is still the question on the test timeout
I can put back the original value, but then it fails consistently for me, and has so far passed consistently with the longer timeout. 
So I'd need to ignore the test, or something like that instead, since the unit tests are required to pass to merge changes (and because it appears to be unrelated to changes in this PR).

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
cahrens: Workflows team has OK'd this timeout increase. https://broadinstitute.slack.com/archives/C1EH66VCM/p1669216629798999

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
cahrens: @blakery still wondering about this one

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -0,0 +1,115 @@
...
+          s"Job to monitor AzureLandingZoneResult created with id ${job.jobControlId} but no billing project set"
+        )
+        return Future.successful(Complete) // nothing more this runner can do with it
+    }
+    val userEmail = job.userEmail match {
+      case Some(email) => email
+      case None =>
+        logger.error(
+          s"Job to monitor AzureLandingZoneResult for billing project $billingProjectName created with id ${job.jobControlId} but no user email set"
```
cahrens: Should this be `${billingProjectName.value}` (does it matter)?

blakery: The string value of the project includes its object type, so it reads: `RawlsBillingProject(projectName)`.
So I removed the part before that so it will read a bit cleaner.
I figure having the actual object name in the message might make it _slightly_ more helpful

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
blakery: The difference isn't huge: it's largely syntactical
this:
```scala
  val result: Try[AzureLandingZoneResult]  = try {
    Success(workspaceManagerDAO.getCreateAzureLandingZoneResult(job.jobControlId.toString, userCtx))
  } catch {
    case t: Throwable => Failure(t)
  }
```
is exactly the same as this:
```scala
   val tryVal: Try[AzureLandingZoneResult] = Try(workspaceManagerDAO.getCreateAzureLandingZoneResult(job.jobControlId.toString, userCtx))
```

The main reason to write it this way is that it makes it convenient for pattern matching.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
blakery: I changed the return type of the job runner, to be a _wrapped_ value of a boolean:
```scala
  implicit sealed class JobStatus(val complete: Boolean)

  case object Complete extends JobStatus(true)

  case object Incomplete extends JobStatus(false)
```
This means that the return type will either be `Complete` or `Incomplete`, which has 2 advantages:
1. It reads a bit more clearly in the job runner itself - instead of returning raw booleans everywhere, you're returning a type that makes it clear what it indicates
2. Because it's a wrapped type, the handling has to be a little bit more explicit, which reduces the chance of doing something unintended

The change here is a result of 2

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
blakery: It's the future that the `SamApiCallback` wraps - so it's just returning that

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
blakery: The job runner doesn't update anything on the job itself: it just indicates if this particular runner is finished with a specific job.
The idea is that if we needed to have more than one thing happen for a particular event, we could add more than one runner (which was probably overthinking it in the first place).

I just added this because it made things easier to read and more explicit than unwrapped booleans.

ehigham (approved):

PR #2131, opened 2022-11-22
Add more error handling to cleanup resources (WOR-635).

cahrens:
Ticket: <[WOR-635](https://broadworkbench.atlassian.net/browse/WOR-635)>

Adds more robustness in the Azure billing project creation/deletion so that we will clean up landing zones and billing profiles if at all possible. 

Note that WSM landing zone deletion code does not use the billing profile at all, so it is safe to delete the billing profile while landing zone deletion is underway.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -1,6 +1,10 @@
 package org.broadinstitute.dsde.rawls.billing
 
+import akka.http.scaladsl.model.StatusCodes.InternalServerError
 import akka.http.scaladsl.model.{StatusCode, StatusCodes}
+import bio.terra.profile.model.ProfileModel
+import bio.terra.workspace.model.CreateLandingZoneResult
+import cats.implicits.{catsSyntaxFlatMapOps, toTraverseOps}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.WorkspaceManagerResourceMonitorRecordDao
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -97,26 +104,106 @@ class BpmBillingProjectLifecycle(
...
+        _ <- billingRepository.updateLandingZoneId(createProjectRequest.projectName, landingZoneId.get)
         _ <- resourceMonitorRecordDao.create(
           UUID.fromString(landingZoneResponse.getJobReport.getId),
           JobType.AzureLandingZoneResult,
-          projectName
+          projectName.value
         )
         _ <- billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
-      } yield CreationStatuses.CreatingLandingZone
+      } yield CreationStatuses.CreatingLandingZone).recoverWith { case t: Throwable =>
```
cahrens: I was surprised that the catch down at line 126 doesn't cover an error thrown within the for comprehension. Is there a cleaner way for me to write this code to avoid the duplicated cleanup logic?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -97,26 +104,106 @@ class BpmBillingProjectLifecycle(
...
+          _ <- cleanupBillingProfile(profileModelId, projectName, ctx)
+        } yield throw t
+      }
     } catch {
-      case exception: Exception => Future.failed(exception)
+      case t: Throwable =>
+        for {
+          _ <- cleanupLandingZone(landingZoneId, projectName, ctx)
+          _ <- cleanupBillingProfile(profileModelId, projectName, ctx)
+        } yield throw t
```
cahrens: Does this throw produce the same result as the previous Future.failed call at line 116?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -97,26 +104,106 @@ class BpmBillingProjectLifecycle(
...
     }
   }
 
+  private def cleanupLandingZone(landingZoneId: Option[UUID],
+                                 projectName: RawlsBillingProjectName,
+                                 ctx: RawlsRequestContext
+  ): Future[Unit] =
```
cahrens, line 148: This method could have just returned `Unit`, but for symmetry with `cleanupBillingProfile` I made it return `Future[Unit]`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -97,26 +104,106 @@ class BpmBillingProjectLifecycle(
...
+                )
+            }
+          case Failure(e) =>
+            logger.warn(
+              s"Unable to delete landing zone with ID ${id} for BPM-backed billing project ${projectName.value}.",
+              e
+            )
+        }
+        Future.successful()
+      case None =>
```
cahrens: This case will only be hit during cleanup that happens due to a failed creation call (depending on where the error is hit, it could be expected that no landing zone exists). In the delete case we log it as a warning and don't call this method at all because we expect the landing zone to exist in that case.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -97,26 +104,106 @@ class BpmBillingProjectLifecycle(
...
+            logger.info(
+              s"Deleting BPM-backed billing project ${projectName.value}, deleting billing profile record ${id}"
+            )
+            billingProfileManagerDAO.deleteBillingProfile(id, ctx)
+          case num =>
+            logger.info(
+              s"Deleting BPM-backed billing project ${projectName.value}, but not deleting billing profile record ${id} because ${num} other project(s) reference it"
+            )
+        }
+      case None =>
```
cahrens: Same note here about this case only being hit during a creation error.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -128,44 +215,16 @@ class BpmBillingProjectLifecycle(
...
       }
-      _ <- billingRepository.getLandingZoneId(projectName).map {
+      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
         case Some(landingZoneId) =>
-          val landingZoneResponse = blocking {
-            // Note that this actually just starts a landing zone deletion job (and thus returns quickly).
-            // We are not attempting to ensure that the landing zone deletion completes successfully.
-            workspaceManagerDAO.deleteLandingZone(UUID.fromString(landingZoneId), ctx)
-          }
-          Option(landingZoneResponse.getErrorReport) match {
```
cahrens, line 138: The refactoring introduced a change in behavior here-- we now don't throw a `BillingProjectDeletionException` if we get back an error report from the landing zone deletion, and instead just log it and proceed with deleting the billing profile.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -373,21 +432,23 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
     val bp =
       new BpmBillingProjectLifecycle(repo, bpm, workspaceManagerDAO, mock[WorkspaceManagerResourceMonitorRecordDao])
 
-    val result = bp.preDeletionSteps(
-      billingProjectName,
-      testContext
+    Await.result(bp.preDeletionSteps(
```
cahrens, line 482: This is the test case related to the change in behavior-- we will still delete the billing profile in this case and just log the issue with deleting the landing zone.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: Yeah and it's not a huge rewrite! When it comes to error handling and cleaning up resources on error, lexical scoping is your friend (though it does lead to some indentation). This allows you to register separate error handlers to clean up errors as the stack gets popped. Doing so can really clean up your error handling logic as you don't have to worry about Option types - you know the resource exists. The general pattern is as follows: 
```scala
createResourceA.flatMap { resourceA => 
   createResourceB(A).flatMap { resourceB
      useResourceB(resourceB).recoverWith { t: Throwable =>
         destroyB(resourceB) >> failed(t)
      }
   }.recoverWith { t: Throwable => 
     destroyA(resourceA) >> failed(t)
   }
}
```

Applying this pattern to the function above becomes: 
```scala
    override def postCreationSteps(createProjectRequest: CreateRawlsV2BillingProjectFullRequest,
                                 config: MultiCloudWorkspaceConfig,
                                 ctx: RawlsRequestContext
  ): Future[CreationStatus] = {
    val projectName = createProjectRequest.projectName

    def createBillingProfile: Future[ProfileModel] =
      Future(blocking {
        val profileModel = billingProfileManagerDAO.createBillingProfile(
          projectName.value,
          createProjectRequest.billingInfo,
          ctx
        )
        logger.info(s"Created Billing Profile with ID ${profileModel.getId} for Billing Project $projectName.")
        profileModel
      })

    // This starts a landing zone creation job. There is a separate monitor that polls to see when it
    // completes and then updates the billing project status accordingly.
    def createLandingZone(profileModel: ProfileModel): Future[CreateLandingZoneResult] =
      Future(blocking {
        workspaceManagerDAO.createLandingZone(
          config.azureConfig.get.landingZoneDefinition,
          config.azureConfig.get.landingZoneVersion,
          profileModel.getId,
          ctx
        )
      })

    createBillingProfile.flatMap { profileModel =>
      createLandingZone(profileModel)
        .flatMap { landingZone =>
          (for {
            _ <- Option(landingZone.getErrorReport).traverse_ { errorReport =>
              Future.failed(
                new LandingZoneCreationException(
                  RawlsErrorReport(StatusCode.int2StatusCode(errorReport.getStatusCode), errorReport.getMessage)
                )
              )
            }
           // potential NullPointerException in `landingZone.getJobReport.getId`?
            _ = logger.info(s"initiated creation of landing zone ${landingZone.getLandingZoneId} with jobId ${landingZone.getJobReport.getId}")
            _ <- billingRepository.updateLandingZoneId(createProjectRequest.projectName, landingZone.getLandingZoneId)
            _ <- resourceMonitorRecordDao.create(
              UUID.fromString(landingZone.getJobReport.getId),
              JobType.AzureLandingZoneResult,
              projectName.value
            )
          } yield CreationStatuses.CreatingLandingZone).recoverWith { case t: Throwable =>
            // no Option here
            cleanupLandingZone(landingZone.getLandingZoneId, projectName, ctx) >> Future.failed(t)
          }
        }
        .recoverWith { case t: Throwable =>
          // no Option here
          cleanupBillingProfile(profileModel.getId, projectName, ctx) >> Future.failed(t)
        }
    }
  }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -75,11 +76,17 @@ class BpmBillingProjectLifecycle(
                                  config: MultiCloudWorkspaceConfig,
                                  ctx: RawlsRequestContext
   ): Future[CreationStatus] = {
-    val projectName = createProjectRequest.projectName.value
+    val projectName = createProjectRequest.projectName
+    var profileModelId = None: Option[UUID]
+    var landingZoneId = None: Option[UUID]
     try {
```
ehigham: Try to avoid java-style `try catch finally` blocks because it encourages imperative-style variable setting and complicates error handling logic (like below) - using error handling with lexical scoping gives better granularity and clarity over the state of the world. See my comment below.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: I am of the mind that anything making an external call should really return `Future` (or better yet,`IO`).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: You can greatly simplify this function and `cleanupBillingProfile` by not having to take `Optional` (see above). Here's an example of how you could implement this. I'd be happy to work through the other one with you if you're interested. 
```scala
  private def cleanupLandingZone(landingZoneId: UUID,
                                 ctx: RawlsRequestContext
  ): Future[Unit] =
    (for {
      // Note that this actually just starts a landing zone deletion job (and thus returns quickly).
      // We are not attempting to ensure that the landing zone deletion completes successfully.
      landingZoneResponse <- Future(blocking {
        val response = workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)
        logger.info(
          s"Initiated deletion of landing zone $landingZoneId for BPM-backed billing project ${projectName.value}."
        )
        response
      })

     // Option.map (or traverse) is your friend when it comes to null-checking and throwing
      _ = Option(landingZoneResponse.getErrorReport).map { errorReport =>
        logger.warn(
          s"Unable to delete landing zone with ID $landingZoneId for BPM-backed " +
            s"billing project ${projectName.value}: ${errorReport.getMessage}."
        )
      }
    } yield ()).recover { case t: Throwable =>
      logger.warn(
        s"Unable to delete landing zone with ID $landingZoneId for BPM-backed billing project ${projectName.value}.",
        t
      )
    }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: Yes - for `Future`s it does because `Future` can encode the idea of failure with an exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -75,45 +80,130 @@ class BpmBillingProjectLifecycle(
...
-      for {
-        _ <- billingRepository.updateLandingZoneId(createProjectRequest.projectName,
-                                                   landingZoneResponse.getLandingZoneId
+      })
+
+    createBillingProfile.flatMap { profileModel =>
+      createLandingZone(profileModel)
+        .flatMap { landingZone =>
+          (for {
+            _ <- Option(landingZone.getErrorReport).traverse { errorReport =>
```
cahrens, line 113: @ehigham the `traverse` documentation seems to indicate that it expects something iterable… but obviously it also works here, which is a bit confusing.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -75,45 +80,130 @@ class BpmBillingProjectLifecycle(
...
+              Future.failed(
+                new LandingZoneCreationException(
+                  RawlsErrorReport(StatusCode.int2StatusCode(errorReport.getStatusCode), errorReport.getMessage)
+                )
+              )
+            }
+            _ <- Option(landingZone.getJobReport) match {
```
cahrens: @ehigham I agree we should avoid a potential NullPointerException. Is there a way to do this that is like `traverse` but only gives you the None case?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: It's using https://typelevel.org/cats/typeclasses/traverse.html

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ehigham: `getOrElse` would make you handle the `None` case explicitly I guess?

cahrens:
I think I've responded to all of @ehigham's comments (which were very helpful!).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: Right, but the documentation doesn't mean much of anything to me… in particular, it seems to imply that the thing you are using traverse with needs to be iterable, and that is not the case here.

ehigham: That type class, `Traverse`, doesn't require the thing to be interable... Scala type classes are a whole other thing and reading them is painful until you've done it enough haha. We can got though that stuff some other time!

ehigham (approved):
Looks great.

cahrens:
> Looks great.

Good thing, since you wrote most of it. :)

blakery (approved):

PR #2140, opened 2022-12-01
WX-854 Use supported Terra region in test

jgainerdewar:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/WX-854)

These tests started failing because when trying to run the workflow, we couldn't write to the bucket. This changes the region to one supported by Terra.

---

**PR checklist**

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
@@ -247,7 +247,7 @@ class RawlsApiSpec
       }(owner.makeAuthToken(billingScopes))
     }
 
-    "should be able to create workspace and run sub-workflow tasks in non-US regions" taggedAs(MethodsTest) ignore in {
+    "should be able to create workspace and run sub-workflow tasks in non-US regions" taggedAs(MethodsTest) in {
       implicit val token: AuthToken = studentBToken
 
       // this will create a method with a workflow containing 3 sub-workflows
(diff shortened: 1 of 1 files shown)


jgainerdewar:
jenkins retest

jgainerdewar:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -247,7 +247,7 @@ class RawlsApiSpec
       }(owner.makeAuthToken(billingScopes))
     }
 
-    "should be able to create workspace and run sub-workflow tasks in non-US regions" taggedAs(MethodsTest) in {
+    "should be able to create workspace and run sub-workflow tasks in non-US regions" taggedAs(MethodsTest) ignore in {
```
davidangb: change `ignore in` to just `ignore`, I think:
```suggestion
    "should be able to create workspace and run sub-workflow tasks in non-US regions" taggedAs(MethodsTest) ignore {
```

davidangb (approved):

aednichols (approved):
Looks OK to me on the basis that we do not publicly claim support for European regions [0] 

The purpose of this test is to verify that we launch compute in the same region as the workspace bucket, to avoid egress. Its purpose is not specifically to check whether compute works in Finland.

[0] https://support.terra.bio/hc/en-us/articles/360058964552-Customizing-where-your-data-are-stored-and-analyzed

PR #2142, opened 2022-12-02
[WOR-529] Orchestrate Cloning Azure Workspaces

ehigham:
RR: https://broadworkbench.atlassian.net/browse/WOR-529 Add `cloneMultiCloudWorkspace` to the `MultiCloudWorkspaceService`. This verifies that the user has the necessary cloud-agnostic permissions to clone the source workspace then delegates to the either `cloneAzureWorkspace` (WOR-625) or the GCP `WorkspaceService` to clone the google workspace.
I've included a bit of refactoring/flattening the source code to avoid duplicating work and more code reuse.

Ticket: <Link to Jira ticket>

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -202,22 +174,59 @@ trait WorkspaceSupport {
...
+            StatusCodes.BadRequest,
+            s"Workspace bucket location must be a single " +
               s"region of format: $singleRegionPattern or the default bucket location ('US')."
           )
-          throw new RawlsExceptionWithErrorReport(errorReport = err)
+        )
+      }
+    }
+
+  /**
```
ehigham, line 194: These have been moved from the workspace service so we can call them directly in the MultiCloudWorkspaceService

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -51,22 +57,35 @@ object MultiCloudWorkspaceService {
...
     )
+
+  def getCloudPlatform(p: RawlsBillingProject): WorkspaceCloudPlatform =
+    AzureBillingProject.unapply(p).as(Azure).getOrElse(Gcp)
+
+  object AzureBillingProject {
+    def unapply(p: RawlsBillingProject): Option[(RawlsBillingProjectName, UUID)] =
+      for {
+        profileId <- p.billingProfileId
+        uuid <- Try(UUID.fromString(profileId)).toOption
```
ehigham: I'm not sure if it's better to throw `InternalServerError` if the profile id is malformed - I'm guessing this is something we create ourselves and the user has no control over the billing project id?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -60,60 +62,61 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Test
...
     val samDAO = new MockSamDAO(slickDataSource)
-    val mcWorkspaceService = MultiCloudWorkspaceService.constructor(
-      slickDataSource,
-      workspaceManagerDAO,
-      mock[BillingProfileManagerDAO],
-      samDAO,
-      config,
-      workbenchMetricBaseName
-    )(testContext)
+    val mcWorkspaceService = spy(
```
ehigham, line 79: spy the `MultiCloudWorkspaceService` so we can partially mock some of its methods. I hate inheritance.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -387,6 +402,162 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Test
     testAsyncCreationFailure(StatusEnum.SUCCEEDED, StatusEnum.FAILED)
   }
 
+  behavior of "cloneMultiCloudWorkspace"
```
ehigham, line 389: note behaviour for cloning GCS workspaces is covered in the workspace service

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -2256,7 +2254,10 @@ class WorkspaceServiceSpec
     val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
 
     val workspace =
-      Await.result(services.workspaceService.cloneWorkspace(baseWorkspace.toWorkspaceName, workspaceRequest),
+      Await.result(services.mcWorkspaceService.cloneMultiCloudWorkspace(services.workspaceService,
```
ehigham, line 2257: sigh scalaFmt

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -5,8 +5,8 @@ import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.stream.Materializer
```
ehigham, line 5: Sorry for the noise in this file. Recommend viewing side-by side and hide whitespace changes.
Main changes:
- Pull out common methods into `WorkspaceSupport`
- Refactor copy + clone to remove work duplication
- Flatten deeply-nested methods that were affected by ^

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3057,64 +3001,57 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-    * @param op
-    * @tparam T
-    * @return
+    * takes a RawlsBillingProject and checks that Rawls has the appropriate permissions on the underlying Billing
+    * Account on Google.  Does NOT check if Terra _User_ has necessary permissions on the Billing Account.  Updates
+    * BillingProject to persist latest 'invalidBillingAccount' info.  Returns TRUE if user has right IAM access, else
+    * FALSE
     */
-  def withBillingProjectContext[T](billingProjectName: String, parentContext: RawlsRequestContext)(
-    op: (RawlsBillingProject) => Future[T]
```
ehigham, line 3125: This pattern was causing very nested code and it's less flexible than  returning `Future[RawlsBillingProject]`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: As discussed, if the profileId exists, you should look up the billing profile record to get the cloud platform. Otherwise assume GCP (example: https://github.com/broadinstitute/rawls/blob/20575d547dbab706c5b712c5548203acbc8d34a9/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala#L315)

And yes, I think an InternalServerError would be appropriate if profileId is malformed. Seems like it might be worth a wrapper function, as we do it other places (https://github.com/broadinstitute/rawls/pull/2142/files#diff-518dd904cf9703f8db8998f3dd2eab4c5ef7b5488d59532eb194a391370e63e7R107).

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -2539,7 +2539,7 @@ class WorkspaceApiServiceSpec extends ApiServiceSpec {
           ArgumentMatchers.eq(SamResourceTypeNames.billingProject),
           ArgumentMatchers.eq(billingProjectName),
           ArgumentMatchers.eq(SamBillingProjectActions.createWorkspace),
-          ArgumentMatchers.argThat(userInfoEq(testContext))
+          any[RawlsRequestContext]
```
cahrens, line 2590: Why did you make this change?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -387,6 +402,162 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Test
...
+      .when(mcWorkspaceService)
+      .getWorkspaceContext(equalTo(testData.azureWorkspace.toWorkspaceName), any())
+
+    val result = intercept[RawlsExceptionWithErrorReport] {
+      Await.result(
+        mcWorkspaceService.cloneMultiCloudWorkspace(
+          mock[WorkspaceService](RETURNS_SMART_NULLS),
+          testData.azureWorkspace.toWorkspaceName,
+          WorkspaceRequest(
+            testData.billingProject.projectName.value,
```
cahrens, line 471: Is this billingProject a GCP billing project? I'm having trouble when reading the test knowing that we are trying to clone to a GCP workspace.

cahrens: Actually, the same billing project is used below for an Azure workspace (https://github.com/broadinstitute/rawls/pull/2142/files#diff-b8196c4af84188faa6eb1f65595077eb0b25f6c06f5a8e64b5c60e4832b93118R545), so I'm even more confused.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -439,6 +439,12 @@ object WorkspaceType {
       case _              => throw new RawlsException(s"Invalid WorkspaceType [$wt]")
     }
 
+  def toCloudPlatform(t: WorkspaceType): WorkspaceCloudPlatform.WorkspaceCloudPlatform =
```
cahrens: I'm not sure if we want to bake this assumption in, as we will use McWorkspaces for GCP before too long. I think the McWorkspace has a cloud context/platform variable that can be checked (and RawlsWorkspace does map to GCP).

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala
ehigham: because the context passed to this function is not the same as the test context. I'm less keen on prescribing request contexts as it makes refactoring less easy.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
ehigham: I'm not too sure what the whole "multi-cloud" thing really means - maybe here's a good place to ask? No workspace is "multi-cloud" - it's either GCP- or Azure-hosted so what does multi-cloud really mean?

cahrens: Actually, I think I'm wrong. The mcWorkspace's "azureContext" vs. "gcpContext" determines the cloud. This is the code in WorkspaceServices' listWorkspaces where we get the cloudPlatform.

But it is a good question… I thought McWorkspace were "multi-cloud" because they could be either Azure or GCP. But I also recall discussions that someday a McWorkspace might have multiple billing profiles (for different clouds). @aherbst-broad can you explain what "multi-cloud" is meant to refer to?

```
                  val cloudPlatform = workspace.workspaceType match {
                    case WorkspaceType.McWorkspace =>
                      Option(workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, ctx)) match {
                        case Some(mcWorkspace) if mcWorkspace.getAzureContext != null =>
                          Option(WorkspaceCloudPlatform.Azure)
                        case Some(mcWorkspace) if mcWorkspace.getGcpContext != null =>
                          Option(WorkspaceCloudPlatform.Gcp)
                        case _ =>
                          throw new RawlsException(
                            s"unexpected state, no cloud context found for workspace ${workspace.workspaceId}"
                          )
                      }

Comment (aherbst-broad) [reply model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala]: So, "McWorkspace" was a term I introduced in the early going that I somewhat regret. It was in reference to the "MC Terra Architecture", a term which has been floating around to mean "not rawls". 

What it has really come to mean, though, is that this is a WSM-owned workspace rather than a legacy rawls workspace. It's defacto azure-only today; someday, as we embark on that migration, MC workspaces will also support GCP. 

To add to the potential complexity, a workspace manager workspace _does_ support multiple cloud contexts (i.e, you can have one GCP cloud context, and one Azure cloud context), but that is not something we will be leveraging any time soon.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala (hunk 2 above)
Comment (aherbst-broad) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala]: Agreed, if the profile ID is malformed that's a programming error and a 500.

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala [hunk 11]
```diff
@@ -81,43 +100,82 @@ class MultiCloudWorkspaceService(ctx: RawlsRequestContext,
...
+      billingProject <- getBillingProjectContext(RawlsBillingProjectName(destWorkspaceRequest.namespace))
+      _ <- requireCreateWorkspaceAction(billingProject.projectName)
+      _ <- dataSource.inTransaction(_ => failIfWorkspaceExists(destWorkspaceRequest.toWorkspaceName))
+      clone <- (WorkspaceType.toCloudPlatform(sourceWs.workspaceType), getCloudPlatform(billingProject)) match {
+        case (Gcp, Gcp)     => wsService.cloneWorkspace(sourceWs, billingProject, destWorkspaceRequest)
+        case (Azure, Azure) => cloneAzureWorkspace(sourceWs, billingProject, destWorkspaceRequest)
+        case (wscp, bpcp) =>
+          Future.failed(
+            RawlsExceptionWithErrorReport(
+              ErrorReport(
```
aherbst-broad, line 207: I like the pattern matching and catch of the mismatch here, good belts and suspenders 👍

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
ehigham: yeah - unless it has `azure` in the name or `mc`, assume gcp

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -72,90 +76,61 @@ trait WorkspaceSupport {
...
-              samDAO.listUserRolesForResource(SamResourceTypeNames.billingProject, workspaceRequest.namespace, ctx)
-            )
+  def requireBillingProjectOwnerAccess(projectName: RawlsBillingProjectName,
+                                       parentContext: RawlsRequestContext
+  ): Future[Unit] =
+    for {
+      billingProjectRoles <- traceWithParent("listUserRolesForResource", parentContext)(context =>
+        samDAO.listUserRolesForResource(SamResourceTypeNames.billingProject, projectName.value, context)
+      )
+      _ <- ApplicativeThrow[Future].raiseUnless(billingProjectRoles.contains(SamBillingProjectRoles.owner)) {
```
aherbst-broad, line 125: TIL about `ApplicativeThrow`

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -101,46 +91,99 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+                  s
+                )
+              }
+            }
+        }
+
+      // Default to the legacy implementation if no workspace was been created
+      // This can happen if there's
+      // - no azure config
+      // - no billing profile or the billing profile's cloud platform is Azure
```
cahrens: Should this "cloud platform is GCP"?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -151,17 +194,23 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
-        case (Gcp, Gcp)     => wsService.cloneWorkspace(sourceWs, billingProject, destWorkspaceRequest)
-        case (Azure, Azure) => cloneAzureWorkspace(sourceWs, billingProject, destWorkspaceRequest)
-        case (wscp, bpcp) =>
+      billingProfileOpt <- getBillingProfile(billingProject)
+      clone <- (WorkspaceType.toCloudPlatform(sourceWs.workspaceType), billingProfileOpt) match {
+
+        case (Azure, Some(profile)) if profile.getCloudPlatform == CloudPlatform.AZURE =>
+          cloneAzureWorkspace(sourceWs, billingProject, profile, destWorkspaceRequest)
+
+        case (Gcp, profileOpt) if profileOpt.map(_.getCloudPlatform).contains(CloudPlatform.GCP) =>
```
cahrens: How does this also handle the case of the billing profile being None?

cahrens: Ah, I believe you fixed this in a later commit.

cahrens (approved):
I'm a bit unclear about some of the context changes in WorkspaceService, but in general I think the code is easier to read now.

Did have a question about one code comment that I think might be incorrect-- https://github.com/broadinstitute/rawls/pull/2142#discussion_r1044762727

Oh, and test failures!

cahrens (changes requested):
Ugh, just noticed test failures.

ehigham:
> Ugh, just noticed test failures.

I won't merge until they pass haha

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: thank you!

cahrens (approved):

PR #2162, opened 2022-12-16
[WOR-625] Wire up WSM Cloning Operations in Rawls

ehigham:
RR: https://broadworkbench.atlassian.net/browse/WOR-625 Call WSM to initiate the workspace clone
Hand off monitoring WSM job to the WorkspaceResourceMonitor (todo [WOR-553] extend support for CloneWorkspace job types)
Add "smart constructors" to `WorkspaceManagerResourceMonitorRecord` so we supply all the fields required for each job type.

I haven't added integration tests - i feel there are other things to implement before that (ie tracking the clone job in wsm), not least that that sort of end-to-end test would be too long running for PR tests and to stay on-top of any drooled state.
This may be a bit controversial, but either wsm's implementation is correct or it isnt and I don't believe here is that right place to test that.

Ticket: <Link to Jira ticket>

---

**PR checklist**

[WOR-553]: https://broadworkbench.atlassian.net/browse/WOR-553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WOR-625]: https://broadworkbench.atlassian.net/browse/WOR-625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -76,7 +76,7 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
       new CloneWorkspaceRequest()
         .destinationWorkspaceId(workspaceId)
         .displayName(displayName)
-        .spendProfile(spendProfile.toString)
+        .spendProfile(spendProfile.getId.toString)
         .azureContext(
           new AzureContext()
             .tenantId(spendProfile.getTenantId.toString)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.rawls.dataaccess.workspacemanager
 
 import akka.actor.ActorSystem
 import akka.stream.Materializer
+import bio.terra.profile.model.ProfileModel
 import bio.terra.workspace.api.{ReferencedGcpResourceApi, ResourceApi, WorkspaceApi}
 import bio.terra.workspace.client.ApiClient
 import bio.terra.workspace.model._
@@ -67,10 +68,24 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
   override def cloneWorkspace(sourceWorkspaceId: UUID,
                               workspaceId: UUID,
                               displayName: String,
-                              spendProfileId: UUID,
-                              location: String,
-                              ctx: RawlsRequestContext
-  ): CloneWorkspaceResult = ???
+                              spendProfile: ProfileModel,
+                              ctx: RawlsRequestContext,
+                              location: Option[String]
+  ): CloneWorkspaceResult =
+    getWorkspaceApi(ctx).cloneWorkspace(
+      new CloneWorkspaceRequest()
+        .destinationWorkspaceId(workspaceId)
+        .displayName(displayName)
+        .spendProfile(spendProfile.toString)
+        .azureContext(
+          new AzureContext()
+            .tenantId(spendProfile.getTenantId.toString)
+            .subscriptionId(spendProfile.getSubscriptionId.toString)
+            .resourceGroupId(spendProfile.getManagedResourceGroupId)
+        )
+        .location(location.orNull),
+      sourceWorkspaceId
+    )
 
   override def createAzureWorkspaceCloudContext(workspaceId: UUID,
                                                 azureTenantId: String,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -64,6 +64,14 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
         .stage(WorkspaceStageModel.MC_WORKSPACE)
     )
 
+  override def cloneWorkspace(sourceWorkspaceId: UUID,
+                              workspaceId: UUID,
+                              displayName: String,
+                              spendProfileId: UUID,
+                              location: String,
+                              ctx: RawlsRequestContext
+  ): CloneWorkspaceResult = ???
+
   override def createAzureWorkspaceCloudContext(workspaceId: UUID,
                                                 azureTenantId: String,
                                                 azureResourceGroupId: String,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -26,7 +26,7 @@ object WorkspaceManagerDAO {
     val byteBuffer = ByteBuffer.allocate(16)
     byteBuffer.putLong(uuid.getMostSignificantBits)
     byteBuffer.putLong(uuid.getLeastSignificantBits)
-    base64Url().encode(byteBuffer.array())
+    base64Url().omitPadding().encode(byteBuffer.array())
   }
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -13,6 +13,8 @@ import scala.util.Try
 
 object WorkspaceManagerDAO {
 
+  // Remove/Move into ShortUUID after [PF-1268]
+
   /** @see bio.terra.stairway.ShortUUID#get() */
   def decodeShortUuid(shortUuid: String): Option[UUID] =
     Try {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.rawls.dataaccess.workspacemanager
 
+import bio.terra.profile.model.ProfileModel
 import bio.terra.workspace.client.ApiException
 import bio.terra.workspace.model._
 import org.broadinstitute.dsde.rawls.model.{DataReferenceDescriptionField, DataReferenceName, RawlsRequestContext}
@@ -21,9 +22,9 @@ trait WorkspaceManagerDAO {
   def cloneWorkspace(sourceWorkspaceId: UUID,
                      workspaceId: UUID,
                      displayName: String,
-                     spendProfileId: UUID,
-                     location: String,
-                     ctx: RawlsRequestContext
+                     spendProfile: ProfileModel,
+                     ctx: RawlsRequestContext,
+                     location: Option[String] = None
   ): CloneWorkspaceResult
 
   def createAzureWorkspaceCloudContext(workspaceId: UUID,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -3,10 +3,32 @@ package org.broadinstitute.dsde.rawls.dataaccess.workspacemanager
 import bio.terra.profile.model.ProfileModel
 import bio.terra.workspace.client.ApiException
 import bio.terra.workspace.model._
+import com.google.common.io.BaseEncoding.base64Url
 import org.broadinstitute.dsde.rawls.model.{DataReferenceDescriptionField, DataReferenceName, RawlsRequestContext}
 import org.broadinstitute.dsde.workbench.model.{ErrorReportSource, WorkbenchEmail}
 
+import java.nio.ByteBuffer
 import java.util.UUID
+import scala.util.Try
+
+object WorkspaceManagerDAO {
+
+  /** @see bio.terra.stairway.ShortUUID#get() */
+  def decodeShortUuid(shortUuid: String): Option[UUID] =
+    Try {
+      val bytes = ByteBuffer.wrap(base64Url.decode(shortUuid))
+      val hi = bytes.getLong(0)
+      val lo = bytes.getLong(8)
+      new UUID(hi, lo)
+    }.toOption
+
+  def encodeShortUUID(uuid: UUID): String = {
+    val byteBuffer = ByteBuffer.allocate(16)
+    byteBuffer.putLong(uuid.getMostSignificantBits)
+    byteBuffer.putLong(uuid.getLeastSignificantBits)
+    base64Url().encode(byteBuffer.array())
+  }
+}
 
 trait WorkspaceManagerDAO {
   val errorReportSource = ErrorReportSource("WorkspaceManager")
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -17,6 +17,15 @@ trait WorkspaceManagerDAO {
                                       spendProfileId: String,
                                       ctx: RawlsRequestContext
   ): CreatedWorkspace
+
+  def cloneWorkspace(sourceWorkspaceId: UUID,
+                     workspaceId: UUID,
+                     displayName: String,
+                     spendProfileId: UUID,
+                     location: String,
+                     ctx: RawlsRequestContext
+  ): CloneWorkspaceResult
+
   def createAzureWorkspaceCloudContext(workspaceId: UUID,
                                        azureTenantId: String,
                                        azureResourceGroupId: String,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
@@ -1,10 +1,12 @@
 package org.broadinstitute.dsde.rawls.monitor.workspace
 
 import akka.actor.{Actor, Props}
+import cats.Applicative
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.config.WorkspaceManagerResourceMonitorConfig
 import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.JobType.JobType
+import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.{Incomplete, JobStatus}
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{
   WorkspaceManagerResourceJobRunner,
   WorkspaceManagerResourceMonitorRecord
@@ -19,10 +21,10 @@ import scala.util.Failure
 
 object WorkspaceResourceMonitor extends {
 
-  def props(config: Config, dataSource: SlickDataSource, jobRunners: List[WorkspaceManagerResourceJobRunner])(implicit
-    executionContext: ExecutionContext
+  def props(config: Config, dataSource: SlickDataSource, jobRunners: Map[JobType, WorkspaceManagerResourceJobRunner])(
+    implicit executionContext: ExecutionContext
   ): Props = {
-    val jobDao = new WorkspaceManagerResourceMonitorRecordDao(dataSource)
+    val jobDao = WorkspaceManagerResourceMonitorRecordDao(dataSource)
     val monitor = new WorkspaceResourceMonitor(jobDao, jobRunners)
     Props(new WorkspaceMonitorRouter(WorkspaceManagerResourceMonitorConfig(config), monitor))
   }
@@ -51,7 +53,7 @@ class WorkspaceMonitorRouter(val config: WorkspaceManagerResourceMonitorConfig, 
 
     // This monitor is always on and polling, and we want that default poll rate to be low, maybe once per minute.
     // if more jobs are active, we want to poll more frequently, say ~once per 5 seconds
-    case CheckDone(count) if count == 0 =>
+    case CheckDone(0) =>
       context.system.scheduler.scheduleOnce(config.defaultRetrySeconds seconds, self, CheckNow)
 
     case CheckDone(_) =>
@@ -66,43 +68,51 @@ class WorkspaceMonitorRouter(val config: WorkspaceManagerResourceMonitorConfig, 
 
 }
 
-class WorkspaceResourceMonitor(
+case class WorkspaceResourceMonitor(
   jobDao: WorkspaceManagerResourceMonitorRecordDao,
-  jobRunners: List[WorkspaceManagerResourceJobRunner]
+  jobRunners: Map[JobType, WorkspaceManagerResourceJobRunner]
 )(implicit val executionContext: ExecutionContext)
     extends LazyLogging {
 
-  val registeredRunners: Map[JobType, List[WorkspaceManagerResourceJobRunner]] = jobRunners.groupBy(_.jobType)
-
   /**
     * Run all jobs, and wait for them to complete
+    *
     * @return the message containing the number of uncompleted jobs
     */
   def checkJobs(): Future[CheckDone] = for {
     jobs: Seq[WorkspaceManagerResourceMonitorRecord] <- jobDao.selectAll()
     jobResults <- Future.sequence(jobs.map(runJob))
-  } yield CheckDone(jobResults.count(_ == false))
+  } yield CheckDone(jobResults.count(!_.isDone))
 
   /**
     * Runs the job in all job runners configured for the type of the job
     * Deletes the job if all runners have completed successfully
+    *
     * @return true if all runners for the job completed successfully, false if any failed
     */
-  def runJob(job: WorkspaceManagerResourceMonitorRecord): Future[Boolean] = for {
-    jobResults <- Future.sequence(
-      registeredRunners
-        .getOrElse(job.jobType, List())
-        .map(_.run(job).map(_.complete).recover { case t =>
-          logger.error(s"Exception monitoring WSM Job", t)
-          false
-        })
-    )
-  } yield
-    if (jobResults.forall(a => a)) { // true iff all runners for the given job completed successfully
-      jobDao.delete(job) // remove completed jobs
-      true
-    } else {
-      false
-    }
+  def runJob(job: WorkspaceManagerResourceMonitorRecord): Future[JobStatus] =
+    for {
+      status <- jobRunners.getOrElse(job.jobType, NoopJobRunner)(job).recover { case t: Throwable =>
+        logger.warn(
+          "Failure monitoring WSM job " +
+            s"[ jobId = ${job.jobControlId}" +
+            s", jobType = ${job.jobType}" +
+            s"]",
+          t
+        )
+        Incomplete
+      }
 
+      _ <- Applicative[Future].whenA(status.isDone) {
+        jobDao.delete(job)
+      }
+
+    } yield status
+}
+
+case object NoopJobRunner extends WorkspaceManagerResourceJobRunner {
+  override def apply(job: WorkspaceManagerResourceMonitorRecord)(implicit
+    executionContext: ExecutionContext
+  ): Future[JobStatus] =
+    Future.successful(Incomplete)
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
@@ -92,7 +92,7 @@ case class WorkspaceResourceMonitor(
     */
   def runJob(job: WorkspaceManagerResourceMonitorRecord): Future[JobStatus] =
     for {
-      status <- jobRunners.getOrElse(job.jobType, NoopJobRunner)(job).recover { case t: Throwable =>
+      status <- jobRunners.getOrElse(job.jobType, AlwaysIncompleteJobRunner)(job).recover { case t: Throwable =>
         logger.warn(
           "Failure monitoring WSM job " +
             s"[ jobId = ${job.jobControlId}" +
@@ -110,7 +110,7 @@ case class WorkspaceResourceMonitor(
     } yield status
 }
 
-case object NoopJobRunner extends WorkspaceManagerResourceJobRunner {
+case object AlwaysIncompleteJobRunner extends WorkspaceManagerResourceJobRunner {
   override def apply(job: WorkspaceManagerResourceMonitorRecord)(implicit
     executionContext: ExecutionContext
   ): Future[JobStatus] =
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -4,16 +4,26 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.profile.model.{CloudPlatform, ProfileModel}
 import bio.terra.workspace.model.JobReport.StatusEnum
-import bio.terra.workspace.model.{CreateCloudContextResult, CreateControlledAzureRelayNamespaceResult}
+import bio.terra.workspace.model.{
+  CloneWorkspaceResult,
+  CreateCloudContextResult,
+  CreateControlledAzureRelayNamespaceResult
+}
 import cats.Apply
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.billing.BillingProfileManagerDAO
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
-import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadWriteAction}
+import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.JobType
+import org.broadinstitute.dsde.rawls.dataaccess.slick.{
+  DataAccess,
+  ReadWriteAction,
+  WorkspaceManagerResourceMonitorRecord
+}
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
-import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
+import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource, WorkspaceManagerResourceMonitorRecordDao}
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
+import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceType.{McWorkspace, RawlsWorkspace}
 import org.broadinstitute.dsde.rawls.model.{
   AzureManagedAppCoordinates,
@@ -92,6 +102,10 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
     for {
       billingProject <- traceWithParent("getBillingProjectContext", parentContext) { s =>
         getBillingProjectContext(RawlsBillingProjectName(workspaceRequest.namespace), s)
+      }
+
+      _ <- traceWithParent("requireCreateWorkspaceAccess", parentContext) { childContext =>
+        requireCreateWorkspaceAction(billingProject.projectName, childContext)
       }
 
       billingProfileOpt <- traceWithParent("getBillingProfile", parentContext) { s =>
@@ -189,17 +203,20 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       sourceWs <- getWorkspaceContextAndPermissions(sourceWorkspaceName, SamWorkspaceActions.read)
       billingProject <- getBillingProjectContext(RawlsBillingProjectName(destWorkspaceRequest.namespace))
       _ <- requireCreateWorkspaceAction(billingProject.projectName)
-      _ <- dataSource.inTransaction(_ => failIfWorkspaceExists(destWorkspaceRequest.toWorkspaceName))
       billingProfileOpt <- getBillingProfile(billingProject)
       clone <- (sourceWs.workspaceType, billingProfileOpt) match {
 
         case (McWorkspace, Some(profile)) if profile.getCloudPlatform == CloudPlatform.AZURE =>
-          cloneAzureWorkspace(sourceWs, billingProject, profile, destWorkspaceRequest)
+          traceWithParent("cloneAzureWorkspace", ctx) { s =>
+            cloneAzureWorkspace(sourceWs, billingProject, profile, destWorkspaceRequest, s)
+          }
 
         case (RawlsWorkspace, profileOpt)
             if profileOpt.isEmpty ||
               profileOpt.map(_.getCloudPlatform).contains(CloudPlatform.GCP) =>
-          wsService.cloneWorkspace(sourceWs, billingProject, destWorkspaceRequest)
+          traceWithParent("cloneRawlsWorkspace", ctx) { s =>
+            wsService.cloneWorkspace(sourceWs, billingProject, destWorkspaceRequest, s)
+          }
 
         case (wsType, profileOpt) =>
           Future.failed(
@@ -218,12 +235,73 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
   def cloneAzureWorkspace(sourceWorkspace: Workspace,
                           project: RawlsBillingProject,
                           profile: ProfileModel,
-                          request: WorkspaceRequest
-  ): Future[Workspace] = Future.failed(
-    RawlsExceptionWithErrorReport(
-      ErrorReport(StatusCodes.NotImplemented, "WOR-625")
-    )
-  )
+                          request: WorkspaceRequest,
+                          parentContext: RawlsRequestContext
+  ): Future[Workspace] = {
+
+    def createNewWorkspaceContext(): Future[Workspace] =
+      dataSource.inTransaction { access =>
+        for {
+          _ <- failIfWorkspaceExists(request.toWorkspaceName)
+          workspaceId = UUID.randomUUID()
+          newWorkspace <- createMultiCloudWorkspaceInDatabase(
+            workspaceId.toString,
+            request.toWorkspaceName,
+            request.attributes,
+            access,
+            parentContext
+          )
+        } yield newWorkspace
+      }
+
+    for {
+      // The call to WSM is asynchronous. Before we fire it off, allocate a new workspace record
+      // to avoid naming conflicts - we'll erase it should the clone request to WSM fail.
+      (newWorkspace, cloneResult) <- createNewWorkspaceContext().flatMap { newWorkspace =>
+        traceWithParent("workspaceManagerDAO.cloneWorkspace", parentContext) { context =>
+          Future
+            .successful((newWorkspace, _: CloneWorkspaceResult))
+            .ap(Future(blocking {
+              workspaceManagerDAO.cloneWorkspace(
+                sourceWorkspace.workspaceIdAsUUID,
+                newWorkspace.workspaceIdAsUUID,
+                request.name,
+                profile.getId,
+                request.bucketLocation.getOrElse(""),
+                context
+              )
+            }))
+        }.recoverWith { case t: Throwable =>
+          logger.warn(
+            "Clone workspace request to workspace manager failed for " +
+              s"[ sourceWorkspaceName=${sourceWorkspace.toWorkspaceName}" +
+              s", newWorkspaceName=${newWorkspace.toWorkspaceName}" +
+              s"]",
+            t
+          )
+          dataSource.inTransaction(_.workspaceQuery.delete(newWorkspace.toWorkspaceName)) *> Future.failed(t)
+        }
+      }
+
+      _ = logger.info(
+        "Created workspace record and clone job for azure workspace " +
+          s"[ workspaceId=${newWorkspace.workspaceId}" +
+          s", workspaceName=${newWorkspace.toWorkspaceName}" +
+          s", cloneJobReportId=${cloneResult.getJobReport.getId}" +
+          s"]"
+      )
+
+      // hand off monitoring the clone job to the resource monitor
+      _ <- WorkspaceManagerResourceMonitorRecordDao(dataSource).create(
+        WorkspaceManagerResourceMonitorRecord.forCloneWorkspace(
+          UUID.fromString(cloneResult.getJobReport.getId),
+          newWorkspace.workspaceIdAsUUID,
+          parentContext.userInfo.userEmail
+        )
+      )
+
+    } yield newWorkspace
+  }
 
   /**
    * Creates a "multi-cloud" workspace, one that is managed by Workspace Manager.
@@ -292,7 +370,13 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       savedWorkspace: Workspace <- traceWithParent("saveMultiCloudWorkspaceToDB", parentContext)(_ =>
         dataSource.inTransaction(
           dataAccess =>
-            createMultiCloudWorkspaceInDatabase(workspaceId.toString, workspaceRequest, dataAccess, parentContext),
+            createMultiCloudWorkspaceInDatabase(
+              workspaceId.toString,
+              workspaceRequest.toWorkspaceName,
+              workspaceRequest.attributes,
+              dataAccess,
+              parentContext
+            ),
           TransactionIsolation.ReadCommitted
         )
       )
@@ -391,19 +475,20 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
     }
 
   private def createMultiCloudWorkspaceInDatabase(workspaceId: String,
-                                                  workspaceRequest: MultiCloudWorkspaceRequest,
+                                                  workspaceName: WorkspaceName,
+                                                  attributes: AttributeMap,
                                                   dataAccess: DataAccess,
                                                   parentContext: RawlsRequestContext
   ): ReadWriteAction[Workspace] = {
     val currentDate = DateTime.now
     val workspace = Workspace.buildMcWorkspace(
-      namespace = workspaceRequest.namespace,
-      name = workspaceRequest.name,
+      namespace = workspaceName.namespace,
+      name = workspaceName.name,
       workspaceId = workspaceId,
       createdDate = currentDate,
       lastModified = currentDate,
       createdBy = ctx.userInfo.userEmail.value,
-      attributes = workspaceRequest.attributes
+      attributes = attributes
     )
     traceDBIOWithParent("saveMultiCloudWorkspace", parentContext)(_ =>
       dataAccess.workspaceQuery.createOrUpdate(workspace)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -4,17 +4,12 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.profile.model.{CloudPlatform, ProfileModel}
 import bio.terra.workspace.model.JobReport.StatusEnum
-import bio.terra.workspace.model.{
-  CloneWorkspaceResult,
-  CreateCloudContextResult,
-  CreateControlledAzureRelayNamespaceResult
-}
+import bio.terra.workspace.model.{CreateCloudContextResult, CreateControlledAzureRelayNamespaceResult}
 import cats.Apply
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.billing.BillingProfileManagerDAO
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
-import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.JobType
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{
   DataAccess,
   ReadWriteAction,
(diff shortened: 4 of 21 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3226,15 +3226,12 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     }
 
     for {
-      _ <- traceDBIOWithParent("requireCreateWorkspaceAccess", parentContext) { childContext =>
-        DBIO.from(requireCreateWorkspaceAction(billingProject.projectName, childContext))
-      }
+      _ <- failIfWorkspaceExists(workspaceRequest.toWorkspaceName)
```
ehigham, line 3229: moved this here as it needs to be in the same transaction that the workspace record is being created in.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3226,15 +3226,12 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     }
 
     for {
-      _ <- traceDBIOWithParent("requireCreateWorkspaceAccess", parentContext) { childContext =>
-        DBIO.from(requireCreateWorkspaceAction(billingProject.projectName, childContext))
-      }
```
ehigham, line 3231: This can be lifted out from the transaction

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
```diff
@@ -66,43 +68,51 @@ class WorkspaceMonitorRouter(val config: WorkspaceManagerResourceMonitorConfig,
...
-      registeredRunners
-        .getOrElse(job.jobType, List())
-        .map(_.run(job).map(_.complete).recover { case t =>
-          logger.error(s"Exception monitoring WSM Job", t)
-          false
-        })
-    )
-  } yield
-    if (jobResults.forall(a => a)) { // true iff all runners for the given job completed successfully
-      jobDao.delete(job) // remove completed jobs
```
ehigham, line 102: we were firing and forgetting this, hence moving into the for-comp body

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,14 +233,73 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+          "Clone workspace request to workspace manager failed for " +
+            s"[ sourceWorkspaceName=${sourceWorkspace.toWorkspaceName}" +
+            s", newWorkspaceName=${newWorkspace.toWorkspaceName}" +
+            s"]",
+          t
+        )
+        dataSource.inTransaction(_.workspaceQuery.delete(newWorkspace.toWorkspaceName)) >> Future.failed(t)
+      }
+
+      _ = clonedWorkspaceCounter.inc()
```
ehigham: should any other metrics get incremented? many the multicloud counter?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,14 +233,73 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+      newWorkspace <- createNewWorkspaceRecord()
+
+      cloneResult <- traceWithParent("workspaceManagerDAO.cloneWorkspace", parentContext) { context =>
+        Future(blocking {
+          workspaceManagerDAO.cloneWorkspace(
+            sourceWorkspace.workspaceIdAsUUID,
+            newWorkspace.workspaceIdAsUUID,
+            request.name,
+            profile.getId,
+            request.bucketLocation.getOrElse(""),
```
marctalbott: I believe we'll want to force this to be `eastus` for now which we can pull from `azureConfig` like we do when creating Azure workspaces

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: This seems fine to me. Alternatively, we could add a new `clonedMultiCloudWorkspaceCounter` or `clonedAzureWorkspaceCounter` to mirror the created workspace counters. We can always combine those to see total new workspaces or cloned workspaces if we want

ehigham: Good idea - I'll stick with the  `MultiCloud` convention as a synonym for 'Workspace-Manager'-managed

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: thanks

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,21 +229,86 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+        } yield newWorkspace
+      }
+
+    for {
+      // Guard for forwards compatibility:
+      // We only support cloning azure workspace storage containers into the Azure Config's
+      // `defaultRegion` for now.
+      _ <- request.bucketLocation.traverse_(throwUnlessIsDefaultRegion)
+
+      // The call to WSM is asynchronous. Before we fire it off, allocate a new workspace record
```
cahrens, line 252: Can't we wait to write the record until we at least get back a success from starting the clone? Similar to what we do on workspace creation?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: Christina tells me that this location is determined by the landing zone's storage account. I'm now in a bit of a knot as to how to support this in a generic way as I understand this route is meant to work for both GCP and azure workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: I think we should change that behaviour too to avoid contention issues - reserve the workspace id/name before using it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: Maybe we won't have to worry about this as we transition from rawls (and given this method is called `createAzureWorkspace`), but if we were, I might consider refactoring the `ProfileModel` to be a tagged-union instead of an all-in-one jobby. Likewise workspaces and billing projects.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,21 +229,80 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+      }
+
+    for {
+      // The call to WSM is asynchronous. Before we fire it off, allocate a new workspace record
+      // to avoid naming conflicts - we'll erase it should the clone request to WSM fail.
+      newWorkspace <- createNewWorkspaceRecord()
+
+      cloneResult <- traceWithParent("workspaceManagerDAO.cloneWorkspace", parentContext) { context =>
+        Future(blocking {
+          workspaceManagerDAO.cloneWorkspace(
```
ehigham, line 258: List of required arguments inspired by
https://github.com/DataBiosphere/terra-workspace-manager/blob/809d8ea4871cb757dd467cd594856359295e2345/service/src/test/java/bio/terra/workspace/service/workspace/AzureWorkspaceTest.java#L195-L200

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -64,6 +65,28 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
+                              spendProfile: ProfileModel,
+                              ctx: RawlsRequestContext,
+                              location: Option[String]
+  ): CloneWorkspaceResult =
+    getWorkspaceApi(ctx).cloneWorkspace(
+      new CloneWorkspaceRequest()
+        .destinationWorkspaceId(workspaceId)
+        .displayName(displayName)
+        .spendProfile(spendProfile.toString)
+        .azureContext(
```
marctalbott, line 80: no-op for this PR, but just noting that `azureContext` can be omitted once the `bpm-azure-enabled` feature flag is turned on in all WSM environments ([link to feature flag in WSM](https://github.com/DataBiosphere/terra-workspace-manager/blob/beaf0e1bd34426318b94e75bdb357be1f99a9f25/service/src/main/resources/application.yml#L207)). It is currently turned on in non-production environments. Once it is turned on in prod as well, we can omit these coordinates and WSM will fetch them from BPM directly based on the spend profile ID we supply.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: Ah right, Christina's correct -- WSM won't do anything with the location if it's included for an Azure workspace anyways. 

Azure workspace location isn't really supported fully at the moment. An Azure workspace's location is fully determined by its Landing Zone. Last I heard, users were expected to create a new LZ if they want to use a different Azure region. At the moment, only the `eastus` region is supported as far as I know. 

So I guess for this use case, the thing to do is just drop the requested location if it's included when cloning an Azure workspace. We could also reject the request entirely if a location is provided for an Azure workspace, although that might be overkill. The Terra UI won't be including a location when cloning an Azure workspace

cahrens: Terra-ui currently _does_ send a location when creating or cloning an Azure workspace because it sends a default value (on create we just ignore it for an Azure workspace). But I don't think it should for Azure workspaces because we don't allow the user to choose a location for those workspaces… I can update my PR to stop sending `bucketLocation` if the workspace's billing project is Azure.

This is the current paylod  for an Azure workspace:

````
attributes: {description: ""}
authorizationDomain: []
bucketLocation: "US-CENTRAL1"
copyFilesWithPrefix: "notebooks/"
name: "CARAzureDec13 copy"
namespace: "CARDec13Work"

Comment (ehigham) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala]: That would be why you were hitting that error because I had previously enforced that if you specified a location, it was whatever our default value was.
I think we should ignore the bucketLocation property. If we ever allow users to create storage containers in locations other than that determined by their LZ, we should add a distinct field for this.

**Comment by ehigham** [comment]
jenkins retest plz

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala (hunk 5 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala]: Ignoring it is fine, but I still think that it's odd that we send it all for Azure workspaces, especially since the value we are sending is GCP-specific.

I'm not going to include it for Azure billing projects-- https://github.com/DataBiosphere/terra-ui/pull/3607/commits/d1b478ee840ab850bbe314e81480e273c2751ef7

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala [hunk 9]
```diff
@@ -216,21 +229,80 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+          )
+        } yield newWorkspace
+      }
+
+    for {
+      // The call to WSM is asynchronous. Before we fire it off, allocate a new workspace record
+      // to avoid naming conflicts - we'll erase it should the clone request to WSM fail.
+      newWorkspace <- createNewWorkspaceRecord()
+
+      cloneResult <- traceWithParent("workspaceManagerDAO.cloneWorkspace", parentContext) { context =>
```
cahrens, line 256: You need to check the `cloneResult` to see if it included an ErrorReport. If so, you need to delete the Rawls workspace record (if you've already created it… I still think you should wait to create until you know that the clone was kicked off successfully).

I had a previous clone attempt that failed due, and now I'm seeing this, indicating that the Rawls record wasn't deleted.

image

cahrens: This was the error I hit when trying to clone the first time. I thought it was because my branch was out-of-date, but I pulled and am still seeing it:

image


{"causes":[],"exceptionClass":"java.lang.IllegalArgumentException","message":"Invalid UUID string: 1MJcwag2S4WZiGfDqKnjuA","source":"rawls","stackTrace":[{"className":"java.util.UUID","fileName":"UUID.java","lineNumber":280,"methodName":"fromString1"},{"className":"java.util.UUID","fileName":"UUID.java","lineNumber":258,"methodName":"fromString"},{"className":"org.broadinstitute.dsde.rawls.workspace.MultiCloudWorkspaceService","fileName":"MultiCloudWorkspaceService.scala","lineNumber":290,"methodName":"$anonfun$cloneAzureWorkspace$7"},{"className":"scala.concurrent.impl.Promise$Transformation","fileName":"Promise.scala","lineNumber":470,"methodName":"run"},{"className":"java.util.concurrent.ForkJoinTask$RunnableExecuteAction","fileName":"ForkJoinTask.java","lineNumber":1395,"methodName":"exec"},{"className":"java.util.concurrent.ForkJoinTask","fileName":"ForkJoinTask.java","lineNumber":373,"methodName":"doExec"},{"className":"java.util.concurrent.ForkJoinPool$WorkQueue","fileName":"ForkJoinPool.java","lineNumber":1182,"methodName":"topLevelExec"},{"className":"java.util.concurrent.ForkJoinPool","fileName":"ForkJoinPool.java","lineNumber":1655,"methodName":"scan"},{"className":"java.util.concurrent.ForkJoinPool","fileName":"ForkJoinPool.java","lineNumber":1622,"methodName":"runWorker"},{"className":"java.util.concurrent.ForkJoinWorkerThread","fileName":"ForkJoinWorkerThread.java","lineNumber":165,"methodName":"run"}]}

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,21 +229,80 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+          s"[ workspaceId=${newWorkspace.workspaceId}" +
+          s", workspaceName=${newWorkspace.toWorkspaceName}" +
+          s", cloneJobReportId=${cloneResult.getJobReport.getId}" +
+          s"]"
+      )
+
+      // hand off monitoring the clone job to the resource monitor
+      _ <- WorkspaceManagerResourceMonitorRecordDao(dataSource).create(
+        WorkspaceManagerResourceMonitorRecord.forCloneWorkspace(
+          UUID.fromString(cloneResult.getJobReport.getId),
```
cahrens: Apparently this isn't a UUID! (see thread above)

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
cahrens: Good because it doesn't make sense to be sending an Azure context if the workspace could be GCP.

Is there a ticket for this cleanup work?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
```diff
@@ -66,43 +68,51 @@ class WorkspaceMonitorRouter(val config: WorkspaceManagerResourceMonitorConfig,
...
+      }
+
+    } yield status
+}
 
+case object NoopJobRunner extends WorkspaceManagerResourceJobRunner {
+  override def apply(job: WorkspaceManagerResourceMonitorRecord)(implicit
+    executionContext: ExecutionContext
+  ): Future[JobStatus] =
+    Future.successful(Incomplete)
```
cahrens, line 117: What is the status Incomplete?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -63,10 +68,10 @@ object MultiCloudWorkspaceService {
   * This service knows how to provision a new "multi-cloud" workspace, a workspace managed by terra-workspace-manager.
   */
 class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
-                                 workspaceManagerDAO: WorkspaceManagerDAO,
+                                 val workspaceManagerDAO: WorkspaceManagerDAO,
```
cahrens, line 71: What does it mean if there is no val? Should `billingProfileManagerDAO` also be fixed?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
```diff
@@ -66,43 +68,51 @@ class WorkspaceMonitorRouter(val config: WorkspaceManagerResourceMonitorConfig,
...
-  } yield
-    if (jobResults.forall(a => a)) { // true iff all runners for the given job completed successfully
-      jobDao.delete(job) // remove completed jobs
-      true
-    } else {
-      false
-    }
+  def runJob(job: WorkspaceManagerResourceMonitorRecord): Future[JobStatus] =
+    for {
+      status <- jobRunners.getOrElse(job.jobType, NoopJobRunner)(job).recover { case t: Throwable =>
```
cahrens: Where does the "running" actually happen?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: I still don't understand why we don't wait to verify that kicking off the clone succeeded (by checking `cloneResult`, which I would probably rename as `cloneJobResult`) before creating the Rawls record. That operation is synchronous.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -154,14 +154,11 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
...
-    when(
-      wsmResouceRecordDao.create(
-        landingZoneJobId,
-        JobType.AzureLandingZoneResult,
-        createRequest.projectName.value,
-        testContext.userInfo.userEmail
-      )
-    ).thenReturn(Future.successful())
+
+    doReturn(Future.successful())
```
cahrens, line 159: This means we lose checking that the correct information was stored. I would prefer checking the expected arguments.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -467,12 +468,20 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
     val testProject3Name = RawlsBillingProjectName("project3")
     val testProject3 = RawlsBillingProject(testProject3Name, CreationStatuses.Ready, Option(billingAccountName), None)
 
-    val testAzureProjectName = RawlsBillingProjectName("azure")
-    val testAzureProject = RawlsBillingProject(testAzureProjectName,
-                                               CreationStatuses.Ready,
-                                               Option(billingAccountName),
-                                               None,
-                                               billingProfileId = Some(UUID.randomUUID().toString)
+    val azureBillingProfile = new ProfileModel()
```
cahrens, line 471: Why create an actual ProfileModel instance?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
cahrens: Oh, I think this is if the job type doesn't match a runner (after seeing the test "mark any jobs that doesnt have a registered handler as incomplete"). I would suggest a comment or rename the JobRunner as something like `NoRegisteredJobRunner`.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
cahrens: Nevermind.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -404,135 +393,186 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
...
+              _ <- mcWorkspaceService.cloneMultiCloudWorkspace(
+                mock[WorkspaceService](RETURNS_SMART_NULLS),
+                testData.azureWorkspace.toWorkspaceName,
+                WorkspaceRequest(
+                  testData.azureWorkspace.namespace,
+                  testData.azureWorkspace.name,
+                  Map.empty
+                )
+              )
+            } yield fail(),
```
cahrens: Why `yield fail()`?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -404,135 +393,186 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
...
-      )
+        }
+
+        result.errorReport.statusCode.value shouldBe StatusCodes.Conflict
+      }
     }
 
-    result.errorReport.statusCode.value shouldBe StatusCodes.BadRequest
-  }
+  it should "not create a workspace record if the request to Workspace Manager fails" in
```
cahrens, line 458: Will also want to test for an error being in the return result object.

ehigham: I'm going to let the runner handle that

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
ehigham: Because if it gets to this point the test has failed

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: If a user sends two requests in quick succession (script error, bug in UI, whatever, if it could happen it will happen), we could send two clone requests for the same workspace name. One will then subsequently fail with a resource conflict. The idea is I want to allocate/reserve the workspace name and then send the clone request so that this sort of thing doesn't happen.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: I'm going to let the runner take care of that. It'll already have to to handle this error so it makes sense to do it in one place rather than duplicating this logic.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: Thanks! Good catch

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
ehigham: This is a continutation from Blakes last PR. In his runner implementation, there was always a 1-to-1 relationship between runners and job types. When a runner runs a job, it returns a Status that's either `Incomplete` or `Complete` that's used to schedule the next run. This 1-to-1 relationship is not the case and to make the simplest change possible without redesigning everying, I just returned incomplete for the job types we don't have runners for. I didn't want to make this PR explode more than it needed to. If we did https://broadworkbench.atlassian.net/browse/WOR-533 before this, I wouldn't need to do this.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
ehigham: a runner is now a function. See the parens around job

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitor.scala
ehigham: I prefer to name it in terms of what it does. That said, noop might mean that it ends in a completed state. Perhaps `AlwaysIncompleteJobRunner`?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: I don't really know tbh. I think a `val` on a property means that it's public. I'm using this in testing to mock some return values from the workspace manager dao

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
ehigham: I'll see what I can do. Having "smart constructors" has cost us the ease of doing this though I think it's a good compromise to make sure we're constructing objects with the right state.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: Do you know if this is a UUID for landing zones?

ehigham: ugh this is a pain because it's a binary number in the databsae.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
cahrens: But you can't let the runner handle it if there is no returned job ID. And even if there is a returned Job id with an error status, I'm not sure it makes sense to have the runner poll again and do the cleanup (though I guess that helps isolate cleanup code?).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: It's probably worth spending some time figuring out why a UUID isn't being returned-- the simplest thing is probably to actually make it a UUID.

With a lot of WSM APIs, you actually pass in the ID that you want to use. I don't know how/this Job ID is getting generated.

ehigham: @marctalbott the swagger page says that this ID is caller provided. Is there any way to make this be a UUID? Otherwise I'm gonna have to do some kind of horror in liquibase.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
ehigham: > if there is no returned job ID

So you're telling me that it doesn't always return an ID?

> though I guess that helps isolate cleanup code?

That's exactly what I'm going after

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: @cahrens, can you give me an example? This is the first time I'm experiencing this stuff.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
ehigham: I've added back some argument verification at the end

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: Ah - it's via a job control. Sadly clone doesn't expose this. Perhaps it should?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
```diff
@@ -1,11 +1,34 @@
...
+      new UUID(hi, lo)
+    }.toOption
+
+  def encodeShortUUID(uuid: UUID): String = {
+    val byteBuffer = ByteBuffer.allocate(16)
+    byteBuffer.putLong(uuid.getMostSignificantBits)
+    byteBuffer.putLong(uuid.getLeastSignificantBits)
+    base64Url().encode(byteBuffer.array())
+  }
+}
```
ehigham: I wish this wasn't necessary

cahrens: I suggest adding a comment with the ticket number (PF-1268) so we know when we can update cloneWorkspace and remove this.

ehigham:
At this point, clone workspace returns sucessfully. When the UI tries to get this workspace, however, I see the following in the logs: 
```
rawls [INFO] [19:49:48.879] [scala-execution-context-global-334] o.b.d.r.w.MultiCloudWorkspaceService - Created workspace record and clone job for azure workspace [ workspaceId='d6281294-28a0-4ba3-8f8e-1dad6a78cfbf', workspaceName='CARDec13Work/clone-workspace-test-clone', cloneJobReportId='47SOE88pQnmwyJ3EWSBKvQ']
rawls [ERROR] [19:49:50.394] [rawls-akka.actor.default-dispatcher-14] akka.actor.ActorSystemImpl - HttpMethod(GET) http://rawls-sbt:8080/api/workspaces/CARDec13Work/clone-workspace-test-clone?fields=accessLevel%2CazureContext%2CcanCompute%2CcanShare%2Cowners%2Cworkspace%2Cworkspace.attributes%2Cworkspace.authorizationDomain%2Cworkspace.cloudPlatform%2Cworkspace.isLocked%2Cworkspace.workspaceId%2CworkspaceSubmissionStats: 501 Not Implemented entity: {"causes":[],"message":"Non-azure MC workspaces not supported [id=d6281294-28a0-4ba3-8f8e-1dad6a78cfbf]","source":"rawls","stackTrace":[],"statusCode":501}
```
The workspace record exits but the clone must have failed as subsequent calls to get the workspace respond in 404.
Seems that workspace manager also agrees  - getting the job responds in an error 
```
curl -X GET "https://workspace.dsde-dev.broadinstitute.org/api/job/v1/jobs/47SOE88pQnmwyJ3EWSBKvQ"
{
  "message": "Workspace d6281294-28a0-4ba3-8f8e-1dad6a78cfbf not found.",
  "statusCode": 404,
  "causes": []
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
ehigham: sure

ehigham: Maybe move these into the ShortUUID class itself

cahrens: Would be nice, but it's in Stairway. I don't think Rawls depends on Stairway.

ehigham:
Logging the CloneWorkspaceResult from WSM: 
```
[ jobId=pmZTD22BQ2i7OeM8taNftA, jobStatus=RUNNING, resultURL=https://workspace.dsde-dev.broadinstitute.org]
```

So seems the job gets created without any error. Looking at the logs, for WSM, seems cloning fails while creating the storage container, though they offer little more insight than that 
https://cloudlogging.app.goo.gl/zAhqVvRU9eFdv4u58

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
ehigham: It does - at least calling ShortUUID.get works in tests, so maybe just test code does?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
ehigham: https://broadworkbench.atlassian.net/browse/WOR-682

ehigham:
jenkins retest

cahrens:
@ehigham after much log digging, found this-- CreateAzureContextFlight is what failed:

````
java.lang.IllegalArgumentException: UUID string too large
	at java.base/java.util.UUID.fromString1(Unknown Source)
	at java.base/java.util.UUID.fromString(Unknown Source)
	at bio.terra.workspace.service.spendprofile.SpendProfileService.getSpendProfileFromBpm(SpendProfileService.java:139)
	at bio.terra.workspace.service.spendprofile.SpendProfileService.authorizeLinking(SpendProfileService.java:86)
	at bio.terra.workspace.service.spendprofile.SpendProfileService$$FastClassBySpringCGLIB$$ec6e75dd.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:793)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed(MethodInvocationProceedingJoinPoint.java:89)
	at io.opencensus.contrib.spring.aop.Handler.proceed(Handler.java:41)
	at io.opencensus.contrib.spring.aop.CensusSpringAspect.trace(CensusSpringAspect.java:70)
	at jdk.internal.reflect.GeneratedMethodAccessor294.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs(AbstractAspectJAdvice.java:634)
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod(AbstractAspectJAdvice.java:624)
	at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke(AspectJAroundAdvice.java:72)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:175)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:708)
	at bio.terra.workspace.service.spendprofile.SpendProfileService$$EnhancerBySpringCGLIB$$ef91efac.authorizeLinking(<generated>)
	at bio.terra.workspace.service.spendprofile.SpendProfileService$$FastClassBySpringCGLIB$$ec6e75dd.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
	at org.springframework.aop.framework.CglibAopProxy.invokeMethod(CglibAopProxy.java:386)
	at org.springframework.aop.framework.CglibAopProxy.access$000(CglibAopProxy.java:85)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:704)
	at bio.terra.workspace.service.spendprofile.SpendProfileService$$EnhancerBySpringCGLIB$$1312968.authorizeLinking(<generated>)
	at bio.terra.workspace.service.workspace.flight.CheckSpendProfileStep.doStep(CheckSpendProfileStep.java:57)
````	



````
ERROR 2022-12-21T19:49:49.994Z [resource.labels.containerName: workspacemanager] Step Result exception java.lang.IllegalArgumentException: UUID string too large at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source) at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source) at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Unknown Source) at java.base/java.lang.reflect.Constructor.newInstance(Unknown Source) at bio.terra.workspace.service.job.StairwayExceptionSerializer.deserialize(StairwayExceptionSerializer.java:130) at bio.terra.stairway.impl.FlightDao.makeFlightStateList(FlightDao.java:1022) at bio.terra.stairway.impl.FlightDao.getFlightStateInner(FlightDao.java:852) at bio.terra.stairway.impl.FlightDao.lambda$getFlightState$11(FlightDao.java:831) at bio.terra.stairway.impl.DbRetry.perform(DbRetry.java:88) at bio.terra.stairway.impl.DbRetry.retry(DbRetry.java:66) at bio.terra.stairway.impl.FlightDao.getFlightState(FlightDao.java:831) at bio.terra.stairway.impl.StairwayImpl.getFlightState(StairwayImpl.java:530) at bio.terra.stairway.impl.StairwayImpl.waitForFlight(StairwayImpl.java:504) at bio.terra.workspace.service.resource.controlled.flight.clone.workspace.AwaitCreateCloudContextFlightStep.doStep(AwaitCreateCloudContextFlightStep.java:31) at bio.terra.stairway.impl.FlightRunner.stepWithRetry(FlightRunner.java:248) at bio.terra.stairway.impl.FlightRunner.runSteps(FlightRunner.java:179) at bio.terra.stairway.impl.FlightRunner.fly(FlightRunner.java:110) at bio.terra.stairway.impl.FlightRunner.run(FlightRunner.java:68) at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) at java.base/java.util.concurrent.FutureTask.run(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) at java.base/java.lang.Thread.run(Unknown Source)
  {
    "insertId": "yq7or6fncc23f88j",
    "jsonPayload": {
      "message": "Step Result exception\njava.lang.IllegalArgumentException: UUID string too large\n\tat java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)\n\tat java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source)\n\tat java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)\n\tat java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Unknown Source)\n\tat java.base/java.lang.reflect.Constructor.newInstance(Unknown Source)\n\tat bio.terra.workspace.service.job.StairwayExceptionSerializer.deserialize(StairwayExceptionSerializer.java:130)\n\tat bio.terra.stairway.impl.FlightDao.makeFlightStateList(FlightDao.java:1022)\n\tat bio.terra.stairway.impl.FlightDao.getFlightStateInner(FlightDao.java:852)\n\tat bio.terra.stairway.impl.FlightDao.lambda$getFlightState$11(FlightDao.java:831)\n\tat bio.terra.stairway.impl.DbRetry.perform(DbRetry.java:88)\n\tat bio.terra.stairway.impl.DbRetry.retry(DbRetry.java:66)\n\tat bio.terra.stairway.impl.FlightDao.getFlightState(FlightDao.java:831)\n\tat bio.terra.stairway.impl.StairwayImpl.getFlightState(StairwayImpl.java:530)\n\tat bio.terra.stairway.impl.StairwayImpl.waitForFlight(StairwayImpl.java:504)\n\tat bio.terra.workspace.service.resource.controlled.flight.clone.workspace.AwaitCreateCloudContextFlightStep.doStep(AwaitCreateCloudContextFlightStep.java:31)\n\tat bio.terra.stairway.impl.FlightRunner.stepWithRetry(FlightRunner.java:248)\n\tat bio.terra.stairway.impl.FlightRunner.runSteps(FlightRunner.java:179)\n\tat bio.terra.stairway.impl.FlightRunner.fly(FlightRunner.java:110)\n\tat bio.terra.stairway.impl.FlightRunner.run(FlightRunner.java:68)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)\n\tat java.base/java.util.concurrent.FutureTask.run(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)\n\tat java.base/java.lang.Thread.run(Unknown Source)\n",
      "thread": "pool-19-thread-1",
      "context": "default",
      "requestId": "eZQGnM5J",
      "serviceContext": {
        "service": "workspace",
        "version": "0.254.526-SNAPSHOT"
      },
      "logger": "bio.terra.stairway.StepResult"
    },
    "resource": {
      "type": "k8s_container",
      "labels": {
        "pod_name": "workspacemanager-deployment-5494b6fc5f-8wqmw",
        "location": "us-central1-a",
        "project_id": "broad-dsde-dev",
        "container_name": "workspacemanager",
        "namespace_name": "terra-dev",
        "cluster_name": "terra-dev"
      }
    },
    "timestamp": "2022-12-21T19:49:49.994Z",
    "severity": "ERROR",
    "labels": {
      "k8s-pod/app_kubernetes_io/part-of": "terra",
      "k8s-pod/app_kubernetes_io/instance": "workspacemanager",
      "k8s-pod/pod-template-hash": "5494b6fc5f",
      "k8s-pod/app_kubernetes_io/component": "workspacemanager",
      "k8s-pod/app_kubernetes_io/managed-by": "Helm",
      "k8s-pod/app_kubernetes_io/name": "workspacemanager",
      "k8s-pod/deployment": "workspacemanager-deployment",
      "k8s-pod/helm_sh/chart": "workspacemanager-0.101.0",
      "compute.googleapis.com/resource_name": "gke-terra-dev-default-v3-97df8767-mt93"
    },
    "logName": "projects/broad-dsde-dev/logs/stdout",
    "trace": "projects/broad-dsde-dev/traces/ca5886cdab2848281a472f73d40823a1",
    "sourceLocation": {
      "file": "bio/terra/stairway/StepResult.java",
      "line": "33",
      "function": "bio.terra.stairway.StepResult.<init>"
    },
    "receiveTimestamp": "2022-12-21T19:49:54.596888854Z",
    "spanId": "1cbe7944151a0b76",
    "traceSampled": true
  }

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala [hunk 19]
```diff
@@ -64,6 +65,28 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
+                              displayName: String,
+                              spendProfile: ProfileModel,
+                              ctx: RawlsRequestContext,
+                              location: Option[String]
+  ): CloneWorkspaceResult =
+    getWorkspaceApi(ctx).cloneWorkspace(
+      new CloneWorkspaceRequest()
+        .destinationWorkspaceId(workspaceId)
+        .displayName(displayName)
+        .spendProfile(spendProfile.toString)
```
cahrens: @ehigham this could be the source of the issue-- try passing `spendProfile.getId.toString`.

cahrens: Yes, that allowed me to get a workspace (which displayed an error!). You should add unit test coverage for this method in `HttpWorkspaceManagerDAO` to make sure you are sending the correct args to WSM.

image

cahrens: Error shown is because we get back no resources for Request URL: https://workspace.dsde-dev.broadinstitute.org/api/workspaces/v1/bc1289e5-cb15-47f6-b117-96b494cb2df7/resources?stewardship=CONTROLLED&limit=1000
. Waiting to see if it resolves after some time….

ehigham: Ah good catch. That was a silly mistake!

ehigham: I'm surprised I didn't get something back from wsm complaining about this

cahrens: You did… it just was very buried!

ehigham: earlier I meant!

cahrens: Yeah, all the casting of strings to UUIDs is very suspect in WSM. Either declare them as UUIDs from the get-go, or at least add better error handling.

ehigham: Added a unit test. Hopefully I got it right again and didn't just code the test wrong too haha

cahrens:
My last attempt to clone an Azure workspace (before holiday) did not blow up, but it also did not clone the workspace files. https://broadinstitute.slack.com/archives/C02L9L96YB1/p1671729706281989?thread_ts=1671656066.181489&cid=C02L9L96YB1

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitorSpec.scala
```diff
@@ -28,86 +36,66 @@ class WorkspaceResourceMonitorSpec extends AnyFlatSpec with Matchers with Mockit
...
-    val job0 = new WorkspaceManagerResourceMonitorRecord(
+    val job0 = WorkspaceManagerResourceMonitorRecord.forAzureLandingZone(
       UUID.randomUUID(),
-      JobType.AzureLandingZoneResult,
-      None,
-      Some("bpId1"),
-      None,
-      Timestamp.from(Instant.now())
+      RawlsBillingProjectName("bpId1"),
+      RawlsUserEmail("simply-sausages@gmail.com")
```
marctalbott, line 42: ? Do they make good sausages?

marctalbott (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/WorkspaceResourceMonitorSpec.scala
ehigham: The best.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: I do think you want to at least verify if you get a JobResult with a JobID in this method. I'm pretty sure we have seen cases in other WSM methods where we get an ErrorReport and no JobID because the operation immediately failed.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -216,21 +229,98 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+      )
+
+      // We can't specify the jobId to WSM at the time of writing and
+      // a 22-character base64url-encoded short-UUID is generated instead.
+      jobId = WorkspaceManagerDAO.decodeShortUuid(cloneResult.getJobReport.getId).getOrElse {
+        // If this happens we're in trouble. Our database needs a UUID but whatever WSM gave us
+        // was not something we know how to convert into a UUID. I don't think we should delete the
+        // workspace record but I'm not going to write an elaborate error handler to clean this up,
+        // instead favouring PF-1269 as a better solution.
+        val message =
```
cahrens, line 293: Don't you need to delete the Rawls workspace record in this case?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
cahrens: I am still thinking we need coverage for the case of no job ID.

cahrens:
@ehigham I'm done re-reviewing. This LGTM with the exception of the comments I added around the actual call to WSM manager's clone method.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: It's such an unlikely thing to happen that I don't think putting more time into writing and testing this is worth it. If it does happen, it'll be logged and the workspace record will exist so we can try to find out what happened

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
ehigham: Are there cases when WSM returns a non-error code but does not return a job id? If there are [strong opinion], that's weird and should be fixed.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: Ha, I totally didn't read your comment in the code! Sure, that's fine.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: If we get an error report why is the request responding with a non error code in the http response? That's really weird - akin to mixing error codes and exceptions when people objected to exceptions "for performance reasons"

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
cahrens: I don't think it's weird. A job report will only be created if there is a job to be executed. If an error is encountered that prevents getting to the point of creating the job (for example, permissions), I would expect an ErrorReport and no JobReport. But we can discuss at standup to get the opinions of others.

cahrens: Also might be good to ask in the terra-workspace-manager Slack channel about what the expectations are. Should one always assume a JobReport will be returned (with a jobId), or should one handle the case of no JobReport and inspect the ErrorReport).

cahrens: Ah, I missed non-error code in your message. Yes, I would expect an error code in that case. Are you then logging the message in the ErrorReport if the server call has a non-success status?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -215,21 +228,98 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+            context
+          )
+        })
+      }.recoverWith { case t: Throwable =>
+        logger.warn(
+          "Clone workspace request to workspace manager failed for " +
+            s"[ sourceWorkspaceName='${sourceWorkspace.toWorkspaceName}'" +
+            s", newWorkspaceName='${newWorkspace.toWorkspaceName}'" +
+            s"]",
+          t
```
cahrens, line 272: So `recoverWith` will capture if the method returns an error status? If so, please include the ErrorReport message if one is present in `cloneResult`.

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: As we discussed, the cause of the error should be in the exception.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
cahrens: We discussed in standup and feel like anything that would prevent a JobReport from being created should return an error status code.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
ehigham: As discussed in standup, we don't expect to get back a 200 without a job id

PR #2153, opened 2022-12-08
WOR-595

blakery:
Ticket: [WOR-595](https://broadworkbench.atlassian.net/browse/WOR-595)

Adds a check prior to workspace deletion that throw an error if there are child resources blocking deletion

---

**PR checklist**

- [ ] Include the JIRA issue number in the PR description and title
- [ ] Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green, including CI tests
- [ ] **Squash commits and merge** to develop (branches are automatically deleted after merging)
- [ ] Inform other teams of any substantial changes via Slack and/or email

PR #2177, opened 2023-01-03
[WOR-683] Don't check google children on deletion for Azure workspaces.

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-683

This fixes an issue that was recently introduced in https://github.com/broadinstitute/rawls/pull/2153

---

**PR checklist**

[WOR-683]: https://broadworkbench.atlassian.net/browse/WOR-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.rawls.workspace
 
-import akka.http.scaladsl.model.{StatusCode, StatusCodes}
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.stream.Materializer
 import bio.terra.workspace.client.ApiException
@@ -624,7 +624,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       } yield ()
     }
 
-  def assertNoChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+  def assertNoGoogleChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+    _ <-
+      if (workspace.googleProjectId.value.isEmpty) {
+        Future.failed(
+          RawlsExceptionWithErrorReport(
+            ErrorReport(StatusCodes.InternalServerError,
+                        "Cannot call this method on a workspace with no googleProjectId"
+            )
+          )
+        )
+      } else {
+        Future.successful()
+      }
     workspaceChildren <- samDAO
       .listResourceChildren(SamResourceTypeNames.workspace, workspace.workspaceId, ctx)
       .map(
@@ -650,8 +662,12 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                                       parentContext: RawlsRequestContext
   ): Future[Option[String]] =
     for {
-      _ <- assertNoChildrenBlockingWorkspaceDeletion(workspaceContext)
-
+      _ <-
+        if (!isAzureMcWorkspace(maybeMcWorkspace)) {
+          assertNoGoogleChildrenBlockingWorkspaceDeletion(workspaceContext)
+        } else {
+          Future.successful()
+        }
       _ <- traceWithParent("requesterPaysSetupService.revokeAllUsersFromWorkspace", parentContext)(_ =>
         requesterPaysSetupService.revokeAllUsersFromWorkspace(workspaceContext) recoverWith { case t: Throwable =>
           logger.warn(
@@ -734,7 +750,6 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         }
       )
 
-      // Delete workspace manager record (which will only exist if there had ever been a TDR snapshot in the WS)
       _ <- traceWithParent("deleteWorkspaceInWSM", parentContext)(_ =>
         Future(workspaceManagerDAO.deleteWorkspace(workspaceContext.workspaceIdAsUUID, ctx)).recoverWith {
           // this will only ever succeed if a TDR snapshot had been created in the WS, so we gracefully handle all exceptions here
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -625,18 +625,14 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     }
 
   def assertNoGoogleChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
-    _ <-
-      if (workspace.googleProjectId.value.isEmpty) {
-        Future.failed(
-          RawlsExceptionWithErrorReport(
-            ErrorReport(StatusCodes.InternalServerError,
-                        "Cannot call this method on a workspace with no googleProjectId"
-            )
-          )
+    _ <- ApplicativeThrow[Future].raiseWhen(workspace.googleProjectId.value.isEmpty) {
+      RawlsExceptionWithErrorReport(
+        ErrorReport(
+          StatusCodes.InternalServerError,
+          s"Cannot call this method on workspace ${workspace.workspaceId} with no googleProjectId"
         )
-      } else {
-        Future.successful()
-      }
+      )
+    }
     workspaceChildren <- samDAO
       .listResourceChildren(SamResourceTypeNames.workspace, workspace.workspaceId, ctx)
       .map(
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -650,8 +662,12 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                                       parentContext: RawlsRequestContext
   ): Future[Option[String]] =
     for {
-      _ <- assertNoChildrenBlockingWorkspaceDeletion(workspaceContext)
-
+      _ <-
+        if (!isAzureMcWorkspace(maybeMcWorkspace)) {
```
cahrens: It is questionable what we should do for McWorkspaces that are GCP, but we are already using this check further down in the method before calling `maybeDeleteGoogleProject`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -734,7 +750,6 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         }
       )
 
-      // Delete workspace manager record (which will only exist if there had ever been a TDR snapshot in the WS)
```
cahrens, line 737: Comment is obsolete.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -624,7 +624,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       } yield ()
     }
 
-  def assertNoChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+  def assertNoGoogleChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+    _ <-
+      if (workspace.googleProjectId.value.isEmpty) {
```
cahrens: If there's a cleaner way to write this if/else, definitely let me know. I spent some time Googling and didn't hit upon something better… though I suspect there is a way to avoid having to explicitly have successful "else" condition.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -624,7 +624,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       } yield ()
     }
 
-  def assertNoChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+  def assertNoGoogleChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+    _ <-
```
cahrens: With this assertion of the googleProjectId, the existing "delete an Azure workspace" test point in `WorkspaceApiServiceSpec` fails without the code change below at line 666.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -624,7 +624,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       } yield ()
     }
 
-  def assertNoChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
+  def assertNoGoogleChildrenBlockingWorkspaceDeletion(workspace: Workspace): Future[Unit] = for {
```
ehigham, line 627: I think there's still a problem when the workspace is a v1 workspace in which case the workspace won't have a google project resource under it in sam but that google project might have resouces under it (eg leo stuff)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ehigham: It's more stylistic sugar, but `ApplicativeThrow` has many common exception-throwing patterns (though scala's type system makes this almost a ugly):
```scala 
ApplicativeThrow[Future].raiseWhen(condition: Bool)(t: Throwable)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -624,7 +624,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+    _ <-
+      if (workspace.googleProjectId.value.isEmpty) {
+        Future.failed(
+          RawlsExceptionWithErrorReport(
+            ErrorReport(StatusCodes.InternalServerError,
+                        "Cannot call this method on a workspace with no googleProjectId"
```
ehigham: I like the assertion. Do you know if we log internal server errors? I'd like to put some info in the logs because this condition should never really be hit (we should have determined that it's an azure or mc workspace before this function is called).
Could you add a log statement at some appropriate level (error maybe?) to this and include the workspace name (ie `namespace/name`) + any other information you think useful?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ehigham: Good thing to flag though I'm not too worried right now. I presume much of this functionality will be moved into wsm for gcp workspaces and wsm knows/will know which cloud platform a workspace is hosted on.

ehigham (changes requested):
Request changes/clarification on the subject of v1 workspaces. Changes otherwise look good, though logging the server error would be nice to have

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ehigham: More stylistic sugar (read: not important, but if you like) You can also use `Applicative` to handle these types of conditional effects. 
```scala
Applicative[Future].whenA(condition: Bool)(action: => Future[Unit])
Applicative[Future].unlessA(condition: Bool)(action: => Future[Unit])
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Presumably we rapidly no longer care about v1 workspaces though.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Thanks, much nicer.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Thanks, I like it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: It will get logged. I added the workspaceId to the message, and that should be sufficient.

ehigham: Workspace id is a little painful as it's hard to query for; it's stored as binary and rawls' mysql version lacks the helpful utils to convert uuids.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: I'm not sure that we can stop caring about them until they're officially unsupported at the end of the month unfortunately. One way to address this would be to change the line where we list out the `googleProjectChildren` to only do so if there is a `google-project` that is a child of the workspace. This would be more correct anyways since any children of the `google-project` resource would only block deletion if the `google-project` was a child of the workspace. So it'd be something like:

```
    workspaceChildren <- samDAO
      .listResourceChildren(SamResourceTypeNames.workspace, workspace.workspaceId, ctx)
    maybeWorkspaceGoogleProject = workspaceChildren.find(_.resourceTypeName == SamResourceTypeNames.googleProject.value)
    googleProjectChildren <- maybeWorkspaceGoogleProject.traverse { workspaceGoogleProject => samDAO.listResourceChildren(SamResourceTypeNames.googleProject, workspaceGoogleProject.resourceId, ctx)}
    blockingChildren = workspaceChildren.filter(c =>
      c.resourceTypeName != SamResourceTypeNames.googleProject.value || workspace.googleProjectId.value != c.resourceId
    ).toList ::: googleProjectChildren.getOrElse(Seq.empty).toList
```

ehigham: exactly

ehigham (approved):
Created https://broadworkbench.atlassian.net/browse/WOR-691 for v1 workspaces

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: We very commonly log IDs, and I search for them quite often… albeit, not through a direct SQL query.

cahrens: In any case, I would not expect that we will ever actually hit this line of code. It's here mostly to ensure that our existing test coverage ensures this method isn't called for Azure workspaces (as the Sam stuff is mocked out, and thus doesn't blow up in the way that the call to real Sam does).

I would imagine the namespace/name will end up being in the log by virtue of the API call itself.

marctalbott (approved):

PR #2201, opened 2023-01-19
WOR-533

blakery:
Ticket: [WOR-533](https://broadworkbench.atlassian.net/browse/WOR-533)

Adding runner for CloneWorkspaceContainer jobs
Patterns are similar to those for LandingZoneCreationStatus
A bit of complexity in the tests was added because for the landing zone status, we have the `BillingRepository` to mock, making tests easy; but we don't have anything similar for workspaces. 
I considered adding something; but then I started looking at the workspace service, and the `WorkspaceSupport` trait, and thought better of it

---

**PR checklist**

[WOR-533]: https://broadworkbench.atlassian.net/browse/WOR-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
@@ -1,17 +1,10 @@
 package org.broadinstitute.dsde.rawls.monitor.workspace.runners
 
-import bio.terra.workspace.model.{CloneWorkspaceResult, JobReport}
+import bio.terra.workspace.model.{CloneControlledAzureStorageContainerResult, CloneWorkspaceResult, JobReport}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SamDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.dataaccess.slick.{
-  WorkspaceManagerResourceJobRunner,
-  WorkspaceManagerResourceMonitorRecord
-}
-import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.{
-  Complete,
-  Incomplete,
-  JobStatus
-}
+import org.broadinstitute.dsde.rawls.dataaccess.slick.{WorkspaceManagerResourceJobRunner, WorkspaceManagerResourceMonitorRecord}
+import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.{Complete, Incomplete, JobStatus}
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
 import org.broadinstitute.dsde.rawls.model.{RawlsRequestContext, Workspace}
 import org.joda.time.DateTime
@@ -63,7 +56,7 @@ class CloneWorkspaceContainerRunner(
         logFailure(msg, Some(t))
         cloneFail(workspaceId, msg).map(_ => Incomplete)
       case Success(ctx) =>
-        Try(workspaceManagerDAO.getCloneWorkspaceResult(workspaceId, job.jobControlId.toString, ctx)) match {
+        Try(workspaceManagerDAO.getCloneAzureStorageContainerResult(workspaceId, job.jobControlId.toString, ctx)) match {
           case Success(result) => handleCloneResult(workspaceId, result)
           case Failure(t) =>
             val msg = s"Api call to get clone result from workspace manager failed with: ${t.getMessage}"
@@ -74,7 +67,7 @@ class CloneWorkspaceContainerRunner(
 
   }
 
-  def handleCloneResult(workspaceId: UUID, result: CloneWorkspaceResult)(implicit
+  def handleCloneResult(workspaceId: UUID, result: CloneControlledAzureStorageContainerResult)(implicit
     executionContext: ExecutionContext
   ): Future[JobStatus] = Option(result.getJobReport).map(_.getStatus) match {
     case Some(JobReport.StatusEnum.RUNNING) => Future.successful(Incomplete)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
@@ -47,18 +47,13 @@ class CloneWorkspaceContainerRunner(
         return Future.successful(Complete) // nothing more this runner can do with it
     }
 
-    val workspace = getWorkspace(workspaceId).map(_.getOrElse {
-      logFailure(s"no workspace found for workspace id: $workspaceId")
-      return Future.successful(Complete)
-    })
-
     val userEmail = job.userEmail match {
       case Some(email) => email
       case None =>
         val msg =
           s"Unable to update clone status for workspace $workspaceId because no user email set on monitoring job"
         logFailure(msg)
-        return workspace.flatMap(ws => cloneFail(ws, msg).map(_ => Complete))
+        return cloneFail(workspaceId, msg).map(_ => Complete)
     }
 
     getUserCtx(userEmail).transformWith {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+      cloneFail(workspaceId, msg).map(_ => Complete)
+  }
+
+  def cloneSuccess(wsId: UUID, finishTime: DateTime)(implicit
+    executionContext: ExecutionContext
+  ): Future[Option[Workspace]] =
+    getWorkspace(wsId).flatMap {
+      case Some(workspace) =>
+        val updatedWS = workspace.copy(completedCloneWorkspaceFileTransfer = Some(finishTime))
+        dataSource.inTransaction(_.workspaceQuery.createOrUpdate(updatedWS)).map(Option(_))
```
cahrens, line 113: What does the `.map(Option(_))` do?

blakery: It just wraps the result of `dataSource.inTransaction(_.workspaceQuery.createOrUpdate(updatedWS))` in an `Option`, so the return type is correct.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
```diff
@@ -0,0 +1,120 @@
...
+  gcsDAO: GoogleServicesDAO
+) extends WorkspaceManagerResourceJobRunner
+    with LazyLogging {
+
+  override def apply(
+    job: WorkspaceManagerResourceMonitorRecord
+  )(implicit executionContext: ExecutionContext): Future[JobStatus] = {
+
+    def logFailure(msg: String, t: Option[Throwable] = None): Unit = {
+      val logMessage = s"CloneWorkspaceResult monitoring job with id ${job.jobControlId} failed: $msg"
```
cahrens: Should be `CloneWorkspaceContainerResult`.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
```diff
@@ -0,0 +1,120 @@
+package org.broadinstitute.dsde.rawls.monitor.workspace.runners
+
+import bio.terra.workspace.model.{CloneControlledAzureStorageContainerResult, CloneWorkspaceResult, JobReport}
```
cahrens: Shouldn't need to import CloneWorkspaceResult.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
```diff
@@ -0,0 +1,120 @@
...
+  workspaceManagerDAO: WorkspaceManagerDAO,
+  dataSource: SlickDataSource,
+  gcsDAO: GoogleServicesDAO
+) extends WorkspaceManagerResourceJobRunner
+    with LazyLogging {
+
+  override def apply(
+    job: WorkspaceManagerResourceMonitorRecord
+  )(implicit executionContext: ExecutionContext): Future[JobStatus] = {
+
```
marctalbott, line 36: Should we be checking that the record has the correct JobType?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+      mock[WorkspaceManagerDAO],
+      mock[SlickDataSource],
+      mock[GoogleServicesDAO]
+    )
+    whenReady(runner(monitorRecord.copy(workspaceId = None)))(
+      _ shouldBe WorkspaceManagerResourceMonitorRecord.Complete
+    )
+  }
+
+  it should "return a completed status if no workspace is found for the id" in {
```
marctalbott, line 79: Not sure I really understand this test case. 

Is this the case where we clone a workspace and the destination workspace isn't found in Rawls after it's finished cloning in WSM? That seems like it should raise an error or otherwise fail in some way

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+        mock[WorkspaceManagerDAO],
+        mock[SlickDataSource],
+        mock[GoogleServicesDAO]
+      )
+    )
+
+    doAnswer(answer =>
+      Future.successful(Some(workspace.copy(errorMessage = Some(answer.getArgument(1).asInstanceOf[String]))))
+    ).when(runner)
+      .cloneFail(ArgumentMatchers.any(), ArgumentMatchers.any())(ArgumentMatchers.any[ExecutionContext]())
```
marctalbott, line 124: Will this test fail if `cloneFail` isn't called?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+        mock[SamDAO],
+        mock[WorkspaceManagerDAO],
+        mock[SlickDataSource],
+        mock[GoogleServicesDAO]
+      )
+    )
+    val completedTime = "2023-01-16T10:08:48.541-05:00"
+    val expectedTime = DateTime.parse(completedTime)
+    val report = new JobReport().status(JobReport.StatusEnum.SUCCEEDED).completed(completedTime)
+    doAnswer { answer =>
```
marctalbott, line 236: It feels strange to me to mock out functions in the runner when that's the class we're testing. Do we need separate tests for `cloneFail` and `cloneSuccess`?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
blakery: This is just that it completes the _job_ if the workspace is not there: `Complete` vs `Incomplete` is just whether or not the runner is finished with the job.
In this case, an error is logged, but we can't set the message on the workspace, because... well, no workspace.

In practice, this is an unlikely scenario, but it could occur with some manual intervention

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
blakery: The main reason cloneFail and cloneSuccess are there in the first place, is to allow method spying, since we're using slick directly (which is both a pain to mock, and tends to end up with very broad argument matchers when you do), instead of a proper dao/repository that can be mocked for workspaces (like we do for billing). 
I considered the idea of creating something along the lines of the 'BillingRepository', until I saw the `WorkspaceSupport` trait used by `WorkspaceService` and thought better of it for now.

Personally, I don't see much value in writing tests for those, since they are just a strict pass-through for updating the workspace with the given parameters, using straightforward already-in-place operations in slick.
LMK if you want to jump on a call or something to go over it further.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
blakery: I added a verify check, so it will now

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
blakery: I wasn't considering it to be important, since the job types are mapped up-stream, to give us more flexibility in terms of enabling/disabling, multiple runners on a job, etc. 

But this probably _should_ explicitly fail on any _other_ type of job..... 

I'm convinced: I'll put in a check.

I'm thinking of having it throw an exception explicitly, since 
1. I think that counts as exceptional behavior (not an expected failure case)
2. there's logic in the monitor that logs things in a way that seems correct for this case

cahrens:
jenkins retest

cahrens:
jenkins retest

cahrens (approved):
I ran the branch locally. It cleaned up all the old entries in the monitor table, and when I cloned a workspace with files it correctly added an entry and then removed it when all the files had copied over.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunnerSpec.scala
marctalbott: That makes sense. Mocking slick is definitely a pain

marctalbott (approved):

PR #2211, opened 2023-01-23
[SU-284] Endpoint for batch updating billing project members + email invites

MatthewBemis:
This PR adds a new endpoint to batch update the members of a billing project. Currently users have to be added one by one which can be a painful UX, and using a managed group isn't always the best option.

This PR also implements the new billing project email invite / pending access for unregistered users.

Ticket: https://broadworkbench.atlassian.net/browse/SU-284

This also relates to https://broadworkbench.atlassian.net/browse/WOR-601

---

**PR checklist**

[SUP-758]: https://broadworkbench.atlassian.net/browse/SUP-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -401,6 +401,44 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
+    patch:
+      tags:
+        - billing_v2
+      summary: batch update members of billing project
+      description: batch update members of billing project. only owners may perform this action.
+      operationId: batchUpdateBillingProjectMembersV2
+      parameters:
+        - $ref: '#/components/parameters/billingProjectIdPathParam'
+        - name: inviteUsersNotFound
+          in: query
+          description: true to invite unregistered users, false to ignore
+          required: true
+          schema:
+            type: boolean
+            default: false
+      requestBody:
+        description: billing project access updates
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/BatchProjectAccessUpdate'
+      responses:
+        204:
+          description: No Content
+        403:
+          description: You must be a project owner to alter the members of a project
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        409:
+          description: You have specified a member that does not exist in the system. Use the inviteUsersNotFound query param to invite them
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
   /api/billing/v2/{projectId}/spendReport:
     get:
       tags:
@@ -5652,6 +5690,22 @@ components:
           role:
             description: the role to modify for the specified member
             $ref: '#/components/schemas/ProjectRole'
+    BatchProjectAccessUpdate:
+        required:
+          - membersToAdd
+          - membersToRemove
+        type: object
+        properties:
+          membersToAdd:
+            description: The list of members to add to the billing project
+            type: array
+            items:
+              $ref: '#/components/schemas/ProjectAccessUpdate'
+          membersToRemove:
+            description: The list of members to remove from the billing project
+            type: array
+            items:
+              $ref: '#/components/schemas/ProjectAccessUpdate'
     SpendReport:
       required:
         - spendDetails
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
@@ -789,10 +789,10 @@ class UserService(
       val membersToAdd = batchProjectAccessUpdate.membersToAdd
       val membersToRemove = batchProjectAccessUpdate.membersToRemove
 
-      collectMissingUsers(membersToAdd.map(_.email), ctx) flatMap { membersToInvite =>
-        if (membersToInvite.isEmpty || inviteUsersNotFound) {
+      collectMissingUsers(membersToAdd.map(_.email), ctx) flatMap { missingUsers =>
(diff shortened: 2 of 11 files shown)


MatthewBemis:
jenkins retest - wsm never started

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5652,6 +5690,22 @@ components:
...
+          membersToAdd:
+            description: The list of members to add to the billing project
+            type: array
+            items:
+              $ref: '#/components/schemas/ProjectAccessUpdate'
+          membersToRemove:
+            description: The list of members to remove from the billing project
+            type: array
+            items:
+              $ref: '#/components/schemas/ProjectAccessUpdate'
```
aherbst-broad, line 5708: Does this mean consumers of this API need to provide a role when removing users from a project? Would it make sense to have this be an array of emails?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -776,6 +781,42 @@ class UserService(
...
 
+  def batchUpdateBillingProjectMembers(projectName: RawlsBillingProjectName,
+                                       batchProjectAccessUpdate: BatchProjectAccessUpdate,
+                                       inviteUsersNotFound: Boolean
+  ): Future[Unit] =
+    requireProjectAction(projectName, SamBillingProjectActions.alterPolicies) {
+      val membersToAdd = batchProjectAccessUpdate.membersToAdd
+      val membersToRemove = batchProjectAccessUpdate.membersToRemove
+
+      collectMissingUsers(membersToAdd.map(_.email), ctx) flatMap { membersToInvite =>
```
aherbst-broad: Nit: I was getting turned around on the naming of `membersToInvite`, perhaps `missingUsers`?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -776,6 +781,42 @@ class UserService(
...
+              addUserToBillingProjectV2(projectName, projectAccessUpdate)
+            }
+            removals <- Future.traverse(membersToRemove) { projectAccessUpdate =>
+              removeUserFromBillingProjectV2(projectName, projectAccessUpdate)
+            }
+          } yield notificationDAO.fireAndForgetNotifications(invites)
+        } else
+          Future.failed(
+            new RawlsExceptionWithErrorReport(
+              ErrorReport(StatusCodes.Conflict, s"Users ${membersToInvite.mkString(",")} have not signed up for Terra")
```
aherbst-broad: Not for this PR, but: I would like to get away from rendering presentation layer language in errors for consumption by the UI and instead render the error into a semantically meaningful structure that the UI (or really any api consumer like a cli tool) can then consume and render however it chooses. Something like this:
```
409
{
"missingUsers": ["foo@bar.com", "bar@baz.com"]
}
```

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -401,6 +401,44 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
+    patch:
```
aherbst-broad, line 404: Is this a thing that needs an equivalent passthrough in orch?

aherbst-broad (commented):
Looking good had a couple questions.

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: The reason for implementing it this way is that a user can have multiple roles on the billing project, so this would allow the caller to revoke a single role. That said, I can't think of a good use-case for a user to have multiple roles so I could buy the argument that this is just an implementation detail of Sam that's leaking out and there's no strong reason to present it to the end-user this way.

I'm not tied to this current implementation, happy to make your suggested change if you'd like

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
MatthewBemis: Renamed in c3609ba61d3ab4ddc6e065a2208eb655446b6e60

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: I'm not sure- I don't think anything will _need_ this to be added to Orch unless AoU wants to use this, but I'm also not sure what our default procedure is these days. Do we add all new Rawls endpoints to Orch by default?

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: I took a look at the current [delete billing project member endpoint](https://github.com/broadinstitute/rawls/blob/5fb91e5e39cbc35c3a813dcfdda9c27dc45a119e/core/src/main/resources/swagger/api-docs.yaml#L592) and it accepts only an email. So, I'd say this new endpoint should consume only emails instead of email+role tuples, to keep it consistent.

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: Fine to leave rawls-only for now if there are no non-terra UI consumers expected.

core/src/main/resources/swagger/api-docs.yaml
MatthewBemis: That endpoint also requires the role name to be specified as part of the [path](https://github.com/broadinstitute/rawls/blob/5fb91e5e39cbc35c3a813dcfdda9c27dc45a119e/core/src/main/resources/swagger/api-docs.yaml#L591). So leaving it as-is would be the more consistent approach

nawatts (approved):

aherbst-broad (approved):

PR #2213, opened 2023-01-24
Make methods private if they can be (WOR-721).

cahrens:
In preparation for some filtering of workspaces (filtering out V1 workspaces) in WOR-721, reduce the public API of `WorkspaceService`.

---

**PR checklist**


cahrens:
jenkins restest

cahrens:
jenkins retest

marctalbott (approved):

blakery (approved):
If you want, it looks like you could just nest these:
```scala
def outerFunction( ):  =  {
  def helper1(): = {}
  def helper2(): = {}
  helper1()
  helper2()
}

```

That's usually what I do when I have a one-off helper function that I don't want to test separately, since it makes it clear what it belongs with, and doesn't clutter the namespace. 

Then if I need to use it elsewhere, I'll extract it.

It's strictly style/preference though - just thought I'd point it out in case you liked that better.

PR #2214, opened 2023-01-25
Limit WorkspaceService, EntityService, and SnapshotService to V2 workspaces (WOR-721).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-721

This PR changes `WorkspaceService` so that it generally only returns/accepts v2 workspaces. I say _generally_ because there are the following exceptions:

1. The delete endpoint continues to accept v1 workspaces for the purpose of cleaning up old workspaces (do we want to add logging or a counter if it is called with a v1 workspace?).
2. Admin endpoints have not been limited to v2 workspaces.
3. [getMethodInputsOutputs](https://github.com/broadinstitute/rawls/blob/00060a63a7a5fd65ce8b2e09c5d0bbf219bf7a39/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L2020), which does not seem to be based on a workspace.
4. [workflowQueueStatus](https://github.com/broadinstitute/rawls/blob/00060a63a7a5fd65ce8b2e09c5d0bbf219bf7a39/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L2588), which does not seem to be based on a workspace.
5. [getGenomicsOperationV2](https://github.com/broadinstitute/rawls/blob/00060a63a7a5fd65ce8b2e09c5d0bbf219bf7a39/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L2766). This is called with a workspaceId, but it immediately calls into cromiam. **I do not know how this endpoint is being used.**
6. Methods that are based on billingProjects instead of workspaces (`createGoogleProject`, `setProjectBillingAccount`, `setupGoogleProjectIam`, `renameAndLabelProject`, `getSpendReportTableName`). Note that these are mostly used by the migration actor.

With these change, the v1 workspace I'm on ("rtitle test") does not show in the workspace list, cannot be directly accessed, and does not contribute tags to the tags filter select. I do get an error about a persistent disk needing to be migrated because "rtitle test" has a disk… that's because the Leo disk API is returning the disk, and yet no workspace that is returned from the `getList` method has a googleProject account matching the disk's. I will next be looking into Leo.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
@@ -97,10 +97,9 @@ class BillingRepository(dataSource: SlickDataSource) {
 
   def failUnlessHasNoWorkspaces(projectName: RawlsBillingProjectName)(implicit ec: ExecutionContext): Future[Unit] =
     dataSource.inTransaction { dataAccess =>
-      for {
-        workspaces <- dataAccess.workspaceQuery.listWorkspacesByNamespace(projectName)
-        (v1Workspaces, v2Workspaces) = workspaces.partition(ws => ws.workspaceVersion == WorkspaceVersions.V1)
-        _ = v2Workspaces map { _ =>
+      dataAccess.workspaceQuery.listWithBillingProject(projectName) map { workspaces =>
+        val (v1Workspaces, v2Workspaces) = workspaces.partition(ws => ws.workspaceVersion == WorkspaceVersions.V1)
+        v2Workspaces map { _ =>
           throw new RawlsExceptionWithErrorReport(
             ErrorReport(
               StatusCodes.BadRequest,
@@ -108,7 +107,7 @@ class BillingRepository(dataSource: SlickDataSource) {
             )
           )
         }
-        _ = v1Workspaces map { _ =>
+        v1Workspaces map { _ =>
           throw new RawlsExceptionWithErrorReport(
             ErrorReport(
               StatusCodes.BadRequest,
@@ -116,6 +115,7 @@ class BillingRepository(dataSource: SlickDataSource) {
             )
           )
         }
-      } yield {}
+        ()
+      }
     }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
@@ -3,7 +3,12 @@ package org.broadinstitute.dsde.rawls.billing
 import akka.http.scaladsl.model.StatusCodes
 import org.broadinstitute.dsde.rawls.dataaccess.SlickDataSource
 import org.broadinstitute.dsde.rawls.model.CreationStatuses.CreationStatus
-import org.broadinstitute.dsde.rawls.model.{ErrorReport, RawlsBillingProject, RawlsBillingProjectName, WorkspaceVersions}
+import org.broadinstitute.dsde.rawls.model.{
+  ErrorReport,
+  RawlsBillingProject,
+  RawlsBillingProjectName,
+  WorkspaceVersions
+}
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 
 import java.util.UUID
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
@@ -1,11 +1,10 @@
 package org.broadinstitute.dsde.rawls.billing
 
 import akka.http.scaladsl.model.StatusCodes
-import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.dataaccess.SlickDataSource
-
 import org.broadinstitute.dsde.rawls.model.CreationStatuses.CreationStatus
-import org.broadinstitute.dsde.rawls.model.{ErrorReport, RawlsBillingProject, RawlsBillingProjectName}
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, RawlsBillingProject, RawlsBillingProjectName, WorkspaceVersions}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 
 import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
@@ -93,15 +92,25 @@ class BillingRepository(dataSource: SlickDataSource) {
 
   def failUnlessHasNoWorkspaces(projectName: RawlsBillingProjectName)(implicit ec: ExecutionContext): Future[Unit] =
     dataSource.inTransaction { dataAccess =>
-      dataAccess.workspaceQuery.countByNamespace(projectName) map { count =>
-        if (count == 0) ()
-        else
+      for {
+        workspaces <- dataAccess.workspaceQuery.listWorkspacesByNamespace(projectName)
+        (v1Workspaces, v2Workspaces) = workspaces.partition(ws => ws.workspaceVersion == WorkspaceVersions.V1)
+        _ = v2Workspaces map { _ =>
           throw new RawlsExceptionWithErrorReport(
             ErrorReport(
               StatusCodes.BadRequest,
               "Project cannot be deleted because it contains workspaces."
             )
           )
-      }
+        }
+        _ = v1Workspaces map { _ =>
+          throw new RawlsExceptionWithErrorReport(
+            ErrorReport(
+              StatusCodes.BadRequest,
+              "Project cannot be deleted because it contains v1 workspaces. Contact support for help."
+            )
+          )
+        }
+      } yield {}
     }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -348,6 +348,9 @@ trait WorkspaceComponent {
     ): ReadAction[Seq[Workspace]] =
       loadWorkspaces(findV2WorkspacesByIdsQuery(workspaceIds), attributeSpecs)
 
+    def listWorkspacesByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] =
+      loadWorkspaces(findByNamespaceQuery(namespaceName))
+
     def countByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Int] =
       findByNamespaceQuery(namespaceName).size.result
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -348,9 +348,6 @@ trait WorkspaceComponent {
     ): ReadAction[Seq[Workspace]] =
       loadWorkspaces(findV2WorkspacesByIdsQuery(workspaceIds), attributeSpecs)
 
-    def listWorkspacesByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] =
-      loadWorkspaces(findByNamespaceQuery(namespaceName))
-
     def countByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Int] =
       findByNamespaceQuery(namespaceName).size.result
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -326,6 +326,11 @@ trait WorkspaceComponent {
     ): ReadAction[Option[Workspace]] =
       loadWorkspace(findByNameQuery(workspaceName), attributeSpecs)
 
+    def findV2WorkspaceByName(workspaceName: WorkspaceName,
+                              attributeSpecs: Option[WorkspaceAttributeSpecs] = None
+    ): ReadAction[Option[Workspace]] =
+      loadWorkspace(findV2WorkspaceByNameQuery(workspaceName), attributeSpecs)
+
     def findById(workspaceId: String): ReadAction[Option[Workspace]] =
       loadWorkspace(findByIdQuery(UUID.fromString(workspaceId)))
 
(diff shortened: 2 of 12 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -338,8 +343,10 @@ trait WorkspaceComponent {
     ): ReadAction[Seq[Workspace]] =
       loadWorkspaces(findByIdsQuery(workspaceIds), attributeSpecs)
 
-    def listByNamespaces(namespaceNames: Seq[RawlsBillingProjectName]): ReadAction[Seq[Workspace]] =
```
cahrens, line 341: This method, and its helper query, was unused.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -174,6 +173,33 @@ trait WorkspaceSupport {
       case Some(workspace) => op(workspace)
     }
 
+  def getV2WorkspaceContextAndPermissions(workspaceName: WorkspaceName,
```
cahrens, line 176: These are copies of the original methods, with the ultimate query (defined in `withV2WorkspaceContext`) modified to the v2 version.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -566,6 +566,8 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       .getPolicy(SamResourceTypeNames.workspace, workspaceId, SamWorkspacePolicyNames.owner, ctx)
       .map(_.memberEmails)
 
+  // Do not limit workspace deletion to V2 workspaces so that we can clean up old V1 workspaces as needed.
+  // Possibly create a counter when this is called on a v1 workspace?
```
cahrens: Will remove the comment when we decide whether or not to do this.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -903,9 +905,13 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         .collect { case Success(workspaceId) =>
           workspaceId
         }
-        .toSeq
+      v2WorkspaceIdsForUser <- dataSource.inTransaction { dataAccess =>
```
cahrens, line 913: Modifying `workspaceQuery.getTags` to limit to v2 workspaces was unattractive because the tag database does not store the workspace version. I didn't really want to do a join there, so I'm first filtering the user's workspaces to v2 workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1551,7 +1557,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
 
   def sendChangeNotifications(workspaceName: WorkspaceName): Future[String] =
     for {
-      workspaceContext <- getWorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.own)
+      workspaceContext <- getV2WorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.own)
```
cahrens, line 1565: Apparently there is no test coverage for this method. Should I add it?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -847,10 +849,10 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
     withLibraryAttributeNamespaceCheck(operations.map(_.name)) {
       for {
         isCurator <- tryIsCurator(ctx.userInfo.userEmail)
-        workspace <- getWorkspaceContext(workspaceName) flatMap { workspace =>
+        workspace <- getV2WorkspaceContext(workspaceName) flatMap { workspace =>
```
cahrens, line 857: What the heck is a "library"? Is it dead code?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2754,7 +2768,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
 
   def getAccessInstructions(workspaceName: WorkspaceName): Future[Seq[ManagedGroupAccessInstructions]] =
     for {
-      workspaceId <- loadWorkspaceId(workspaceName)
+      workspaceId <- loadV2WorkspaceId(workspaceName)
```
cahrens, line 2842: Hmmm, another method lacking test coverage.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: I think a counter is worthwhile for both v1 and v2 workspace deletion calls, either here or in a separate PR+ticket. We don't have great observability on the frequency of calls to this API for v1 workspaces so it's either a counter or a logging statement and grepping.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: I think this is a firecloud-ui feature, perhaps [this](https://portal.firecloud.org/#workspaces/anvil-datastorage/1000G-high-coverage-2019). I see no evidence of this in terra-ui.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: I think so.

aherbst-broad (approved):
Looks straightforward

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -338,8 +343,10 @@ trait WorkspaceComponent {
...
-      loadWorkspaces(findByNamespacesQuery(namespaceNames))
+    def listV2WorkspacesByIds(workspaceIds: Seq[UUID],
+                              attributeSpecs: Option[WorkspaceAttributeSpecs] = None
+    ): ReadAction[Seq[Workspace]] =
+      loadWorkspaces(findV2WorkspacesByIdsQuery(workspaceIds), attributeSpecs)
 
     def countByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Int] =
```
marctalbott, line 351: This is used to determine whether a user can delete a billing project. It could be very confusing for a user trying to delete a billing project that has only v1 workspaces in it. The error message will indicate that there are workspaces to delete, but they will not be able to see said workspaces. We do still need to prevent billing project deletion if there are _any_ workspaces in the billing project, so maybe we just need to amend the error message the user will see so that they know they may need to reach out to support for help

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -326,6 +326,11 @@ trait WorkspaceComponent {
     ): ReadAction[Option[Workspace]] =
```
marctalbott, line 326: This is still getting used by `WorkspaceSupport.getWorkspaceContext` which is used by a bunch of entity and snapshot service things. I think we'll want to switch those to stop returning v2 workspaces as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -326,6 +326,11 @@ trait WorkspaceComponent {
...
       loadWorkspace(findByNameQuery(workspaceName), attributeSpecs)
 
+    def findV2WorkspaceByName(workspaceName: WorkspaceName,
+                              attributeSpecs: Option[WorkspaceAttributeSpecs] = None
+    ): ReadAction[Option[Workspace]] =
+      loadWorkspace(findV2WorkspaceByNameQuery(workspaceName), attributeSpecs)
+
     def findById(workspaceId: String): ReadAction[Option[Workspace]] =
```
marctalbott, line 334: This is still being called by the `CloneWorkspaceContainerRunner` which should probably switch to `findV2WorkspaceById` although sort of a nonissue since that runner will theoretically only ever touch v2 workspaces

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: There's a bunch of [orch library endpoints](https://firecloud-orchestration.dsde-dev.broadinstitute.org/#/Library) as well. It looks like DUOS might use the feature, but I don't know much about what it does or how it works

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Created a [ticket](https://broadworkbench.atlassian.net/browse/WOR-753) to add a counter. I would think we want one for both creating workspaces and deleting them. If I have time, I can do it this sprint.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Added in https://github.com/broadinstitute/rawls/pull/2214/commits/215f0fd9af3b924d4f774c464e23262e55d0f283

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: Added in https://github.com/broadinstitute/rawls/pull/2214/commits/215f0fd9af3b924d4f774c464e23262e55d0f283

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
cahrens: Right, I'm not trying to prevent all internal usages to the original APIs… I tried that first and it was a mess. Instead I'm just preventing v1 workspace access/usage at the service layer. In addition to the fact that the runner only works on Azure workspaces, clone operations can't be started on v1 workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
cahrens: I think it would be good to do an explicit query for the v2 workspaces, and if none exist, do a query for v1 workspaces.

1. If v2 workspaces exist, show existing message, "Project cannot be deleted because it contains workspaces".
2. If _only_ v1 workspaces exist, show "Project cannot be deleted because it contains v1 workspaces."

marctalbott: I like the sound of that or possibly doing a group by to get the counts for v2 and v1 in a single transaction. Your solution would be easier to rip out later. The only other thing would be to modify the error message for v1 workspaces to indicate that users should contact support for help deleting the billing project

cahrens: Implemented in https://github.com/broadinstitute/rawls/pull/2214/commits/7e7b84e6047d8f98d304462af21994dd92571c10

cahrens: I looked into adding test coverage for the v1 case, but it was getting ugly. I tested it manually.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/UserApiServiceSpec.scala
```diff
@@ -137,6 +139,8 @@ class UserApiServiceSpec extends ApiServiceSpec {
       sealRoute(services.userRoutes) ~>
       check {
         assertResult(StatusCodes.BadRequest)(status)
+        val responseString = Unmarshal(response.entity).to[String].futureValue
+        assert(responseString.contains("Project cannot be deleted because it contains workspaces."))
```
cahrens, line 143: This billing project does contain a v1 workspace, but it contains around 22 workspaces in total (mostly v2).

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -348,6 +348,9 @@ trait WorkspaceComponent {
     ): ReadAction[Seq[Workspace]] =
       loadWorkspaces(findV2WorkspacesByIdsQuery(workspaceIds), attributeSpecs)
 
+    def listWorkspacesByNamespace(namespaceName: RawlsBillingProjectName): ReadAction[Seq[Workspace]] =
```
marctalbott: There is already a method that does the same thing `listWithBillingProject`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
```diff
@@ -93,15 +97,25 @@ class BillingRepository(dataSource: SlickDataSource) {
 
   def failUnlessHasNoWorkspaces(projectName: RawlsBillingProjectName)(implicit ec: ExecutionContext): Future[Unit] =
     dataSource.inTransaction { dataAccess =>
-      dataAccess.workspaceQuery.countByNamespace(projectName) map { count =>
-        if (count == 0) ()
-        else
+      for {
+        workspaces <- dataAccess.workspaceQuery.listWorkspacesByNamespace(projectName)
```
marctalbott: Since there's only one query here, the for-comp isn't strictly necessary. You could `.map` `listWorkspacesByNamespace` and then do all the same things within that block. It's a style preference thing though so up to you!

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
cahrens: Ah, thanks! I got confused because I thought that map would iterate over all the workspaces, which wasn't what I wanted. But I guess it does not because what is actually returned is a `ReadAction`.

I made your suggested change.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
cahrens: Thank you, I missed that! TBH, I don't really see the point of just a few methods being in `WorkspaceExtensions`.

aherbst-broad (approved):

PR #2224, opened 2023-02-02
[WOR-667] Delete Azure workspace if creation fails partway through

marctalbott:
Ticket: [WOR-667](https://broadworkbench.atlassian.net/browse/WOR-667)
- Catch exceptions from WSM when orchestrating calls required to create an Azure workspace so that we can rollback workspace creation if one of the calls fails.
- Move creation of workspace record in Rawls earlier in the createAzureWorkspace process
- Remove `MultiCloudWorkspaceRequest` -- this lets us share code more easily between `cloneWorkspace` and `createWorkspace` and this case class wasn't being used for very much at this point.

---

**PR checklist**

[WOR-667]: https://broadworkbench.atlassian.net/browse/WOR-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -430,10 +430,17 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       _ = logger.info(
         s"Created Azure storage container in WSM [workspaceId = ${workspaceId}, containerId = ${containerResult.getResourceId}]"
       )
-    } yield savedWorkspace).recoverWith { case e: ApiException =>
+    } yield savedWorkspace).recoverWith { case e @ (_: ApiException | _: WorkspaceManagerCreationFailureException) =>
       logger.info(s"Error creating workspace ${workspaceRequest.toWorkspaceName} [workspaceId = ${workspaceId}]", e)
       for {
-        _ <- Future(workspaceManagerDAO.deleteWorkspace(workspaceId, ctx))
+        _ <- Future(blocking {
+          workspaceManagerDAO.deleteWorkspace(workspaceId, ctx)
+        }).recover { case e =>
+          logger.info(
+            s"Error cleaning up workspace ${workspaceRequest.toWorkspaceName} in WSM [workspaceId = ${workspaceId}",
+            e
+          )
+        }
         _ <- dataSource.inTransaction(_.workspaceQuery.delete(workspaceRequest.toWorkspaceName))
       } yield throw new RawlsExceptionWithErrorReport(
         ErrorReport(StatusCodes.InternalServerError,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.workspace
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.profile.model.{CloudPlatform, ProfileModel}
+import bio.terra.workspace.client.ApiException
 import bio.terra.workspace.model.JobReport.StatusEnum
 import bio.terra.workspace.model._
 import cats.Apply
@@ -121,13 +122,8 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
             .traverse { _ =>
               traceWithParent("createMultiCloudWorkspace", parentContext) { s =>
                 createMultiCloudWorkspace(
-                  MultiCloudWorkspaceRequest(
-                    workspaceRequest.namespace,
-                    workspaceRequest.name,
-                    workspaceRequest.attributes,
-                    WorkspaceCloudPlatform.Azure,
-                    profileModel.getId.toString
-                  ),
+                  workspaceRequest,
+                  profileModel,
                   s
                 )
               }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -527,6 +511,23 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       case Right(_) => ()
     }
 
+  private def createNewWorkspaceRecord(workspaceId: UUID,
+                                       request: WorkspaceRequest,
+                                       parentContext: RawlsRequestContext
+  ): Future[Workspace] =
+    dataSource.inTransaction { access =>
```
marctalbott, line 518: Doing this all in a single transaction eliminates the potential for duplicate requests to conflict with each other

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -233,33 +226,19 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
 
     val wsmConfig = multiCloudWorkspaceConfig.workspaceManager
       .getOrElse(throw new RawlsException("WSM app config not present"))
-
-    def createNewWorkspaceRecord(): Future[Workspace] =
```
marctalbott, line 237: Moved into private method down below

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -4,6 +4,7 @@ import akka.actor.PoisonPill
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import akka.http.scaladsl.model.{StatusCode, StatusCodes}
 import akka.http.scaladsl.testkit.ScalatestRouteTest
+import bio.terra.profile.model.ProfileModel
```
marctalbott, line 7: All changes in this file are fallout from removing `MultiCloudWorkspaceRequest`

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -233,33 +226,19 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
-          )
-        } yield newWorkspace
-      }
+    val workspaceId = UUID.randomUUID()
 
     for {
       // The call to WSM is asynchronous. Before we fire it off, allocate a new workspace record
       // to avoid naming conflicts - we'll erase it should the clone request to WSM fail.
-      newWorkspace <- createNewWorkspaceRecord()
+      newWorkspace <- createNewWorkspaceRecord(workspaceId, request, parentContext)
```
blakery, line 234: It would be nice to connect the provisional creation with the deletion on failure, in a more direct way, with some consistency across instances (it seems the pattern happens a couple of times in this file). Might be more trouble than it's worth, though.

marctalbott: Not entirely sure I understand. Do you mean you'd like to see the rollback logic directly after each WSM call when creating or cloning a workspace instead of just the one catchall I've added for workspace creation?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -145,18 +145,6 @@ object WorkspaceVersions {
     )
 }
 
-case class MultiCloudWorkspaceRequest(
```
aherbst-broad, line 148: 👍 nice cleanup.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -463,7 +430,24 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       _ = logger.info(
         s"Created Azure storage container in WSM [workspaceId = ${workspaceId}, containerId = ${containerResult.getResourceId}]"
       )
-    } yield savedWorkspace
+    } yield savedWorkspace).recoverWith { case e @ (_: ApiException | _: WorkspaceManagerCreationFailureException) =>
```
aherbst-broad, line 433: Is there any chance WSM gives back a 409 in case there's a duplicate? Thinking through the retry case here.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: Good catch. 

If the workspace id is already in use when WSM first receives the request, it will [return a 409](https://github.com/DataBiosphere/terra-workspace-manager/blob/f607490bea34363768e5d9cb4c4ba5d379ddd4fa/service/src/main/java/bio/terra/workspace/app/controller/WorkspaceApiController.java#L143). If one of the later steps in the workspace creation flight fails due to a duplicate workspace, WSM will return a 400 instead 🙃 (see [`WorkspaceService`](https://github.com/DataBiosphere/terra-workspace-manager/blob/f607490bea34363768e5d9cb4c4ba5d379ddd4fa/service/src/main/java/bio/terra/workspace/service/workspace/WorkspaceService.java#L113) and [`CreateWorkspaceStep`](https://github.com/DataBiosphere/terra-workspace-manager/blob/f607490bea34363768e5d9cb4c4ba5d379ddd4fa/service/src/main/java/bio/terra/workspace/service/workspace/flight/CreateWorkspaceStep.java#L39) which both throw [`DuplicateWorkspaceException`](https://github.com/DataBiosphere/terra-workspace-manager/blob/f607490bea34363768e5d9cb4c4ba5d379ddd4fa/service/src/main/java/bio/terra/workspace/service/workspace/exceptions/DuplicateWorkspaceException.java#L6) which extends `BadRequest`, not `Conflict`).

WSM only checks the workspace id for uniqueness. Given the low likelihood of a UUID collision, I question whether it's worth accounting for it but 🤷. EDIT: I've decided it's not worth it as I don't expect us to have any UUID collisions

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: Looked into how `cloneWorkspace` does its rollback, it does clean up the Rawls record if any WSM calls fail but it doesn't delete the workspace in WSM. I think it's fine as is so I'm going to leave it rather than expand the scope of this ticket

PR #2230, opened 2023-02-10
WOR-778 fix access check with default pet

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-778

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2735,6 +2738,10 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         petKey,
         expectedGoogleBucketPermissions
       )
+      _ <- ApplicativeThrow[Future].raiseWhen(useDefaultPet && expectedGoogleProjectPermissions.nonEmpty) {
+        new RawlsException("user has workspace read-only access yet has expected google project permissions")
+      }
+
```
dvoet, line 2743: I am torn about throwing this exception. On the one hand, it is an unexpected case. On the other hand, this will probably lead to unanticipated failure if we ever choose to add google project permissions to read-only users.

The alternative is that we just don't do the google project access check for read-only users. But that also means that if we ever do add google project permissions we will silently not be checking them.

Opinions welcome.

aherbst-broad: I prefer to throw the exception instead of silently not checking them; it seems like that's a sure programmer error we'd  want to catch in testing should the day arrive when we do add the google project perms for readonly users.

aherbst-broad (approved):

cahrens (approved):
Is there any unit test coverage for this API? I was hoping to see some test cases as well.

dvoet:
There is not currently any unit test coverage as this was mainly developed for integration tests. I can add some if you like.

aherbst-broad:
@dvoet a small unit test would be good, yeah.

aherbst-broad (approved):
Thumbing this to unblock but I'd still love to see a unit test.

dvoet:
jenkins retest

dvoet:
jenkins retest

PR #2239, opened 2023-02-16
AJ-862: filter by name

davidangb:
adds filter-by-name capability to the `entityQuery` API. This will allow users (once UI adopts this feature) to search for e.g. a sample_id in a sample table and return the one row they are looking for.

Performance on a 30k row anonymized table: https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/da_theiagen_anonymized_clone/data

```
30k rows:
Full table query: 3200-3400ms
Entity name exact match query: 1200ms

300 rows:
Full table query: 1200ms
Entity name exact match query: 1200ms
```

remaining TODOs:
- [x] quantify, even manually/anecdotally, the latency impact of using `entityNameFilter` instead of `filterTerms` on a large table
- [x] ensure Jenkins automation tests pass
- [ ] update Orchestration's swagger definition to include the new query parameter, test it works via Orch

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -1,23 +1,18 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
-import io.opencensus.trace.{AttributeValue => OpenCensusAttributeValue, Span}
+import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
+import org.broadinstitute.dsde.rawls.model.AttributeName.toDelimitedName
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
 import org.broadinstitute.dsde.rawls.util.TracingUtils.traceDBIOWithParent
-import org.broadinstitute.dsde.rawls.{
-  model,
-  RawlsException,
-  RawlsExceptionWithErrorReport,
-  RawlsFatalExceptionWithErrorReport
-}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport, RawlsFatalExceptionWithErrorReport, model}
 import slick.jdbc.{GetResult, JdbcProfile, SQLActionBuilder}
 
 import java.sql.Timestamp
 import java.util.{Date, UUID}
 import scala.annotation.tailrec
-import scala.collection.immutable.TreeSeqMap
 import scala.language.postfixOps
 
 //noinspection TypeAnnotation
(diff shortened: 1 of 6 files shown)


davidangb:
Jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
```diff
@@ -3418,6 +3418,77 @@ class EntityApiServiceSpec extends ApiServiceSpec {
...
+                                        calculateNumPages(expectedEntities.size, pageSize)
+              ),
+              expectedEntities
+            )
+          ) {
+            responseAs[EntityQueryResponse]
+          }
+        }
+    }
+  }
```
davidangb, line 3490: could you add a unit test for field selection while filtering by name?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -550,12 +552,20 @@ trait EntityComponent {
...
+        val formattedAttributePairs =
+          desiredFields.map(attributeName => s"('${attributeName.namespace}', '${attributeName.name}')").mkString(", ")
+        val attributesFilter =
+          if (desiredFields.isEmpty) "" else s"and (a.namespace, a.name) in ($formattedAttributePairs)"
+
         sql"""#${baseEntityAndAttributeSql(
             workspaceContext
-          )} where e.name = ${entityName} and e.entity_type = ${entityType} and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}"""
+          )} where e.name = ${entityName} and e.entity_type = ${entityType} and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}
+          #${attributesFilter}"""
```
davidangb: This is vulnerable to SQL injection. Try specifying a field of `')) OR 1=1; --` via the swagger UI (it'll be an error). The vulnerability here is because we're building a string literal and then using that literal via `#${attributesFilter}` instead of using bind parameters.

Take a look at WorkspaceComponent: https://github.com/broadinstitute/rawls/blob/5a23db18d962b2955a9ffca2240b054394aa6bcf/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala#L214-L218 which does a very similar thing but uses bind params. I think you could implement it like this:

```scala
        // user requested specific attributes. include them in the where clause.
        val attrNamespaceNameTuples = reduceSqlActionsWithDelim(desiredFields.toSeq.map { attrName =>
          sql"(${attrName.namespace}, ${attrName.name})"
        })
        val attributesFilter =
          if (desiredFields.isEmpty) sql""
          else concatSqlActions(sql" and (a.namespace, a.name) in (", attrNamespaceNameTuples, sql")")

        concatSqlActions(
          sql"""#${baseEntityAndAttributeSql(
              workspaceContext
            )} where e.name = ${entityName} and e.entity_type = ${entityType} and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}""",
          attributesFilter
        )
```

jordansafer:
Updated timing data for sql statement execution times:

| num cols |  full scan | exact name match (all cols) | 
| ------ | ----- | -------- | 
| all columns 30k row table | 1973, 1962, 1959         | 149, 149, 149 | 
| 2 columns 30k row table | 1933, 1938, 1949 | 159, 156, 151 |
| all columns 300 row table |  99, 94, 98  | 96, 92, 95  | 

The new filter term has clear impact, showing over 10x speedup in query time over the full table scan. Notably, the full table scan query time increases from an average of 97 ms to an average of 1965 ms, a 20x increase when switching from a 300 row to 30k row table (100x increase in size). For the same increase in table size, the exact match query time increases from an average of 94 ms to an average of 155 ms, only a 60% increase in runtime.

The filtered attribute query logic has marginal impact, if any. Queries where logic was filtered by attribute were about 1% faster on the full table scan and 1% slower on the exact match query. Given the small sample sizes, it's hard to conclude there's any impact from filtering attributes as part of the sql query.

davidangb (commented):
important comment inline!

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiServiceSpec.scala
jordansafer: on it!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
jordansafer: I'll test it out!

davidangb (commented):
lgtm! Here's a thumb: :+1:. Since I started this PR I can't give it a real approval, though

calypsomatic (approved):

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -1025,7 +1026,7 @@ class WorkspaceJsonSupport extends JsonSupport {
 
   implicit val EntityTypeMetadataFormat: RootJsonFormat[EntityTypeMetadata] = jsonFormat3(EntityTypeMetadata)
 
-  implicit val EntityQueryFormat: RootJsonFormat[EntityQuery] = jsonFormat7(EntityQuery)
+  implicit val EntityQueryFormat: RootJsonFormat[EntityQuery] = jsonFormat8(EntityQuery)
```
aaronkanzer, line 1029: Mini q -- why this change? Does it have to do something with the # of keys in the json?

davidangb: yes, exactly. since we added `entityNameFilter` as a brand new field to `EntityQuery`, we have to change this JsonFormat to be an 8-tuple instead of a 7-tuple. It's a quirk of the `spray-json` library.

aaronkanzer: ah got it -- thanks!

aaronkanzer (approved):

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
jordansafer: Yeah! It's changed due to the additional argument/param to entity query

PR #2240, opened 2023-02-17
AJ-864: filter-by-column

davidangb:
Adds filter-by-column capability. I will make a separate PR to update Orch's swagger definitions once this merges.

---

David's performance tests:
* against workspace `davidan-bp-3/da-theiagen_anonymized_clone-6` in dev
* entityType `qxeuvwzusdwv` (32865 rows)
* filterTerms: `csrgy`
* columnFilter: `bbxqzvoxhhcu=csrgy`
* measured by running Rawls locally and capturing the execution time of the relevant SQL queries as reported by Slick

time, ms | filterTerms |   | columnFilter |  
-- | -- | -- | -- | --
  | count query | results query | count query | results query
run 1 | 74 | 669 | 74 | 292
run 2 | 73 | 691 | 75 | 290
run 3 | 72 | 649 | 74 | 278
run 4 | 72 | 665 | 75 | 284
run 5 | 72 | 662 | 69 | 277
avg | 72.6 | 667.2 | 73.4 | 284.2

---

Bria's performance tests:
30k rows:
Full table query: 300-1000ms
Filter term query: ~2000ms
Filter by column query: ~800ms

3k rows:
Full table query: ~300ms
Filter term query: ~400ms
Filter by column query: ~200-400ms

---

TODOs:
- [x] lots lots lots more tests, specifically around interactions with other features like field selection, sorting by columns other than "name", page size and offset; also need tests for filtering on values inside arrays
- [x] refactoring input validation out to a separate method
- [x] quantify, even manually/anecdotally, the latency impact of using columnFilter instead of filterTerms on a large table. We've added an additional subquery and I don't know how this impacts performance.
- [ ] update Orchestration's swagger definition to include the new query parameter, test it works via Orch

Note: the diff in this PR has a bunch of `scalafmt` changes; you might want to ignore whitespace in the diff

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -1,23 +1,18 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import akka.http.scaladsl.model.StatusCodes
-import io.opencensus.trace.{AttributeValue => OpenCensusAttributeValue, Span}
+import io.opencensus.trace.{Span, AttributeValue => OpenCensusAttributeValue}
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
+import org.broadinstitute.dsde.rawls.model.AttributeName.toDelimitedName
 import org.broadinstitute.dsde.rawls.model.{Workspace, _}
 import org.broadinstitute.dsde.rawls.util.CollectionUtils
 import org.broadinstitute.dsde.rawls.util.TracingUtils.traceDBIOWithParent
-import org.broadinstitute.dsde.rawls.{
-  model,
-  RawlsException,
-  RawlsExceptionWithErrorReport,
-  RawlsFatalExceptionWithErrorReport
-}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport, RawlsFatalExceptionWithErrorReport, model}
 import slick.jdbc.{GetResult, JdbcProfile, SQLActionBuilder}
 
 import java.sql.Timestamp
 import java.util.{Date, UUID}
 import scala.annotation.tailrec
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -480,24 +480,55 @@ trait EntityComponent {
...
+          entityQuery.columnFilter match {
+            case None => sql""
+            case Some(columnFilter) =>
+              val shardId = determineShard(workspaceContext.workspaceIdAsUUID)
+
+              sql""" and e.id in (
+                select filter_e.id
+                from ENTITY filter_e, ENTITY_ATTRIBUTE_#${shardId} filter_a
+                where filter_a.owner_id = filter_e.id
+                and filter_a.namespace = ${columnFilter.attributeName.namespace}
```
jordansafer, line 493: a silly optimization question, if we moved the `and filter_a.name` to be the before the namespace check of the where clause, would this speed things up? Does sql first check everything for owner and namespace, then check name? I'm thinking that namespace match will almost always be true, but the name check will almost always be false, so we'll save some checking if we check the name first. I don't know if it's really true based on how sql executes things and whether there's any significant improvement from the change though-
EDIT: based on quick search, looks like the query optimizer _should_ take care of this in most cases

aaronkanzer (approved):
Thank you for the very robust tests!

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: agreed - this really should be handled by the query optimizer. I also grabbed the SQL query and `EXPLAIN`-ed it before and after moving the `filter_a.name` clause earlier - the execution plan was the same in both cases:

filter-by-column-explain2

PR #2247, opened 2023-02-27
AJ-242: list-all-entities streams

davidangb:
Adds JSON streaming for list-all-entities, and removes the `pageSizeLimit` cap on how many entities can be returned from this API. Obsoletes #1536 and #1670.

Requires broadinstitute/firecloud-develop#3253.

`EntityService` and `EntityServiceSpec` contain the important changes in this PR. Most of the rest of the files are syntax/compatibility changes that should have no real effect.

Non-automated testing shows that:
* time to first byte is significantly faster for large data tables and equivalent for small data tables
* time to complete the request is slower for large data tables and equivalent for small data tables
* memory usage does not spike for large data tables

TODOs:
- [x] refactor `EntityService.listEntities` to better support unit tests
- [ ] (maybe) remove test-only SQL query methods - _do this in a follow-on PR to keep PRs managable_
- [x] review code for any refactoring
- [x] perf test
- [x] memory test

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -73,6 +73,15 @@ data-source {
   }
 }
 
+slick {
+  # batchSize is used for writes, to group inserts/updates
+  # this must be explicitly utilized via custom business logic
+  batchSize = 2000
+  # fetchSize is used during Slick streaming to set the size of pages
+  # this must be explicitly set via withStatementParameters
+  fetchSize = 5000
+}
+
 wdl-parsing {
   # number of parsed WDLs to cache
   cache-max-size = 7500
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -64,6 +64,12 @@ case class EntityRecordWithInlineAttributes(id: Long,
                                             deletedDate: Option[Timestamp]
 ) extends EntityRecordBase
 
+// result structure from entity and attribute list raw sql
+case class EntityAndAttributesResult(entityRecord: EntityRecord,
+                                     attributeRecord: Option[EntityAttributeRecord],
+                                     refEntityRecord: Option[EntityRecord]
+)
+
 object EntityComponent {
   // the length of the all_attribute_values column, which is TEXT, -1 becaue i'm nervous
   val allAttributeValuesColumnSize = 65534
@@ -224,12 +230,6 @@ trait EntityComponent {
 
     type EntityQuery = Query[EntityTable, EntityRecord, Seq]
     type EntityAttributeQuery = Query[EntityAttributeTable, EntityAttributeRecord, Seq]
-
-    // result structure from entity and attribute list raw sql
-    case class EntityAndAttributesResult(entityRecord: EntityRecord,
-                                         attributeRecord: Option[EntityAttributeRecord],
-                                         refEntityRecord: Option[EntityRecord]
-    )
 
     // Raw queries - used when querying for multiple AttributeEntityReferences
 
@@ -1432,12 +1432,12 @@ trait EntityComponent {
       Entity(entityRecord.name, entityRecord.entityType, attributes)
 
     def unmarshalEntities(
-      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesResult]
+      entityAttributeRecords: Seq[EntityAndAttributesResult]
     ): Seq[Entity] =
       unmarshalEntitiesWithIds(entityAttributeRecords).map { case (_, entity) => entity }
 
     def unmarshalEntitiesWithIds(
-      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesResult]
+      entityAttributeRecords: Seq[EntityAndAttributesResult]
     ): Seq[(Long, Entity)] = {
       val allEntityRecords = entityAttributeRecords.map(_.entityRecord).distinct
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -12,7 +12,9 @@ import org.broadinstitute.dsde.rawls.{
   RawlsExceptionWithErrorReport,
   RawlsFatalExceptionWithErrorReport
 }
+import slick.dbio.Effect.Read
 import slick.jdbc.{GetResult, JdbcProfile, SQLActionBuilder}
+import slick.sql.SqlStreamingAction
 
 import java.sql.Timestamp
 import java.util.{Date, UUID}
@@ -223,6 +225,12 @@ trait EntityComponent {
     type EntityQuery = Query[EntityTable, EntityRecord, Seq]
     type EntityAttributeQuery = Query[EntityAttributeTable, EntityAttributeRecord, Seq]
 
+    // result structure from entity and attribute list raw sql
+    case class EntityAndAttributesResult(entityRecord: EntityRecord,
+                                         attributeRecord: Option[EntityAttributeRecord],
+                                         refEntityRecord: Option[EntityRecord]
+    )
+
     // Raw queries - used when querying for multiple AttributeEntityReferences
 
     // noinspection SqlDialectInspection,DuplicatedCode
@@ -287,12 +295,6 @@ trait EntityComponent {
     private object EntityAndAttributesRawSqlQuery extends RawSqlQuery {
       val driver: JdbcProfile = EntityComponent.this.driver
 
-      // result structure from entity and attribute list raw sql
-      case class EntityAndAttributesResult(entityRecord: EntityRecord,
-                                           attributeRecord: Option[EntityAttributeRecord],
-                                           refEntityRecord: Option[EntityRecord]
-      )
-
       // tells slick how to convert a result row from a raw sql query to an instance of EntityAndAttributesResult
       implicit val getEntityAndAttributesResult = GetResult { r =>
         // note that the number and order of all the r.<< match precisely with the select clause of baseEntityAndAttributeSql
@@ -327,6 +329,17 @@ trait EntityComponent {
           left outer join ENTITY e_ref on a.value_entity_ref = e_ref.id"""
 
       // Active actions: only return entities and attributes with their deleted flag set to false
+
+      // TODO: DRY up activeStreamForType and activeActionForType
+      // almost the same as "activeActionForType" except 1) adds a sort by e.id; 2) returns a stream
+      def activeStreamForType(workspaceContext: Workspace,
+                              entityType: String
+      ): SqlStreamingAction[Seq[EntityAndAttributesResult], EntityAndAttributesResult, Read] =
+        sql"""#${baseEntityAndAttributeSql(workspaceContext)}
+        where e.deleted = false
+        and e.entity_type = ${entityType}
+        and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}
+        order by e.id""".as[EntityAndAttributesResult]
 
       def activeActionForType(workspaceContext: Workspace,
                               entityType: String
@@ -794,6 +807,14 @@ trait EntityComponent {
     def listEntities(workspaceContext: Workspace): ReadAction[TraversableOnce[Entity]] =
       EntityAndAttributesRawSqlQuery.actionForWorkspace(workspaceContext) map (query => unmarshalEntities(query))
 
+    // almost the same as "listActiveEntitiesOfType" except 1) does not unmarshal entities; 2) returns a stream
+    def streamActiveEntityAttributesOfType(workspaceContext: Workspace,
+                                           entityType: String
+    ): SqlStreamingAction[Seq[EntityAndAttributesResult], EntityAndAttributesResult, Read] =
+      EntityAndAttributesRawSqlQuery.activeStreamForType(workspaceContext, entityType)
+
+    // TODO: if only used by tests, can this be removed?
+    // currently only used by tests
     def listActiveEntitiesOfType(workspaceContext: Workspace, entityType: String): ReadAction[TraversableOnce[Entity]] =
       EntityAndAttributesRawSqlQuery.activeActionForType(workspaceContext, entityType) map (query =>
         unmarshalEntities(query)
@@ -1410,22 +1431,19 @@ trait EntityComponent {
     private def unmarshalEntity(entityRecord: EntityRecord, attributes: AttributeMap): Entity =
       Entity(entityRecord.name, entityRecord.entityType, attributes)
 
-    private def unmarshalEntities(
-      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesRawSqlQuery.EntityAndAttributesResult]
+    def unmarshalEntities(
+      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesResult]
     ): Seq[Entity] =
       unmarshalEntitiesWithIds(entityAttributeRecords).map { case (_, entity) => entity }
 
-    private def unmarshalEntitiesWithIds(
-      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesRawSqlQuery.EntityAndAttributesResult]
+    def unmarshalEntitiesWithIds(
+      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesResult]
     ): Seq[(Long, Entity)] = {
       val allEntityRecords = entityAttributeRecords.map(_.entityRecord).distinct
 
       // note that not all entities have attributes, thus the collect below
       val entitiesWithAttributes = entityAttributeRecords.collect {
-        case EntityAndAttributesRawSqlQuery.EntityAndAttributesResult(entityRec,
-                                                                      Some(attributeRec),
-                                                                      refEntityRecOption
-            ) =>
+        case EntityAndAttributesResult(entityRec, Some(attributeRec), refEntityRecOption) =>
           ((entityRec.id, attributeRec), refEntityRecOption)
       }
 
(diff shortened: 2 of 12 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -328,13 +330,24 @@ trait EntityComponent {
...
+        and e.entity_type = ${entityType}
+        and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}"""
+
+      // almost the same as "activeActionForType" except 1) adds a sort by e.id; 2) returns a stream
+      def activeStreamForType(workspaceContext: Workspace,
+                              entityType: String
+      ): SqlStreamingAction[Seq[EntityAndAttributesResult], EntityAndAttributesResult, Read] =
+        concatSqlActions(listTypeSql(workspaceContext, entityType), sql" order by e.id").as[EntityAndAttributesResult]
+
+      // TODO: if only used by tests, can this be removed?
```
davidangb, line 345: I suggest doing this in a follow-on PR to keep PRs manageable

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -794,6 +807,14 @@ trait EntityComponent {
     def listEntities(workspaceContext: Workspace): ReadAction[TraversableOnce[Entity]] =
       EntityAndAttributesRawSqlQuery.actionForWorkspace(workspaceContext) map (query => unmarshalEntities(query))
 
+    // almost the same as "listActiveEntitiesOfType" except 1) does not unmarshal entities; 2) returns a stream
+    def streamActiveEntityAttributesOfType(workspaceContext: Workspace,
+                                           entityType: String
+    ): SqlStreamingAction[Seq[EntityAndAttributesResult], EntityAndAttributesResult, Read] =
+      EntityAndAttributesRawSqlQuery.activeStreamForType(workspaceContext, entityType)
+
+    // TODO: if only used by tests, can this be removed?
```
davidangb, line 816: I suggest doing this in a follow-on PR to keep PRs manageable

core/src/main/resources/reference.conf
```diff
@@ -73,6 +73,15 @@ data-source {
   }
 }
 
+slick {
```
aaronkanzer, line 76: Just curious where the values here came from? Is there something specific about `2000` and `5000`?

davidangb: `batchSize = 2000` is a pre-existing value set in rawls.conf: https://github.com/broadinstitute/firecloud-develop/blob/59558bd76cc869004b850e999dc9a50d5243e4c7/base-configs/rawls/rawls.conf.ctmpl#L601 … I'm just copying it down from rawls.conf to reference.conf for the sake of thoroughness.

`fetchSize = 5000` is the value that seemed to work best for me in testing - and I realize I explained that in https://github.com/broadinstitute/firecloud-develop/pull/3253 but not here, my bad!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
aaronkanzer: Could you just move them into the test files as utility functions for now? Or would the follow-up PR have more in mind with these functions?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -339,35 +344,156 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
-                }.map { r =>
-                  r.toSeq
-                }
-              }
+  /*
+   * Queries the db for a stream of entity attributes.
+   */
+  private def listEntitiesDbSource(workspaceContext: Workspace,
+                                   entityType: String
+  ): Source[EntityAndAttributesResult, NotUsed] = {
```
aaronkanzer, line 352: Might be my lack of Scala -- but what is the need for `NotUsed` here?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -339,35 +344,156 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
+
+    // translate the Slick stream to a Source
+    Source.fromPublisher(dataSource.database.stream(allAttrsStream))
+  }
+
+  /**
+    * Given a Source containing entity attributes, scroll through that source and combine
+    * attributes into entities. Emit a Source of entities.
+    * <p>
+    * IMPORTANT: this !!requires!! that the incoming Source of attributes is ordered by
```
aaronkanzer, line 371: Is there any way to enforce/check this ordering before any processing of entities? It seems `incomplete/duplicate` could be problematic

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1410,22 +1431,19 @@ trait EntityComponent {
...
+    def unmarshalEntities(
+      entityAttributeRecords: Seq[EntityAndAttributesResult]
     ): Seq[Entity] =
       unmarshalEntitiesWithIds(entityAttributeRecords).map { case (_, entity) => entity }
 
-    private def unmarshalEntitiesWithIds(
-      entityAttributeRecords: Seq[entityQuery.EntityAndAttributesRawSqlQuery.EntityAndAttributesResult]
+    def unmarshalEntitiesWithIds(
+      entityAttributeRecords: Seq[EntityAndAttributesResult]
     ): Seq[(Long, Entity)] = {
```
aaronkanzer, line 1441: I know this line wasn't changed -- but why is `Long` used here instead of something like `AnyVal`?

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -339,35 +344,156 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
+          AttrAccum(Seq(), None)
+
+        // if previous is empty but current is not, it's the first element
+        case (EmptyElement, curr: AttrAccum) =>
+          curr
+
+        // midstream, we notice that the current entity is the same as the previous entity.
+        // keep gathering attributes for this entity, and don't emit an entity yet.
+        case (prev: AttrAccum, curr: AttrAccum) if prev.accum.head.entityRecord.id == curr.accum.head.entityRecord.id =>
+          val newAccum = prev.accum ++ curr.accum
```
aaronkanzer, line 413: Not sure if `:::` would be preferred in terms of concatenation?

(Anyways -- an interesting read: https://stackoverflow.com/questions/6559996/scala-list-concatenation-vs -- was just looking up syntax of Scala, and noticed the second voted response there)

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -339,35 +344,156 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
+       until we notice that the current element is from a different entity than the previous attribute;
+       when that happens, we marshal and emit an entity.
+     */
+    class EntityCollector extends GraphStage[FlowShape[AttrAccum, AttrAccum]] {
+      val in = Inlet[AttrAccum]("EntityCollector.in")
+      val out = Outlet[AttrAccum]("EntityCollector.out")
+      override val shape = FlowShape(in, out)
+
+      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+        private var prev: AttributeStreamElement = EmptyElement // note: var!
```
aaronkanzer, line 447: why the `note: var!`?

I'm curious what I should be on the lookout for here ha

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
```diff
@@ -253,13 +259,16 @@ class CaseSensitivitySpec extends AnyFreeSpec with Matchers with TestDriverCompo
           }
         }
 
+        implicit val actorSystem = ActorSystem() // needed for stream materialization below
```
aaronkanzer: Do you need to declare this twice? Seems you already have it at the top of the class (~Line 51-52ish)

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityServiceSpec.scala
```diff
@@ -463,4 +450,160 @@ class EntityServiceSpec
       ex.errorReport.message shouldBe "Can't find attribute name non-existent-attribute"
       ex.errorReport.statusCode shouldBe Some(StatusCodes.NotFound)
   }
+
+  // all following tests can use the same exemplar data, no need to re-create it for each unit test
```
aaronkanzer, line 454: An overall question for this test suite -- do you want to include a unit test when the input has some  `EntityAndAttributesResult` with `refEntityRecord` as some value, and others with `None`? Would it be useful, or is this state not possible?

It seems that these unit tests have either some value there or `None` for all values in a given Sequence

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -339,35 +344,156 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
+              emit(out, gatherOrOutput(prev, EmptyElement))
+              completeStage()
+            }
+          }
+        )
+      }
+    }
+
+    val pipeline = dbSource
+      .map(x => AttrAccum(Seq(x), None)) // transform EntityAndAttributesResult to AttrAccum
```
aaronkanzer: Could we use a more descriptive variable name here instead of `x`?

davidangb:
Now that broadinstitute/firecloud-develop#3253 has merged, jenkins retest.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I could refactor to move these into test-only code - that would still involve a good number of lines of code changing, and I wanted to limit the number of lines changing in this PR. But if you think it's worth it I'll do it in this PR instead of in a follow-on!

aaronkanzer: Next PR sounds good!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: The `Source` class requires something here, and `NotUsed` "is used in generic type signatures wherever the actual value is of no importance" ([link](https://doc.akka.io/api/akka/current/akka/NotUsed.html)). More specifically, `Source.fromPublisher` (which I don't have control over) returns a `Source[T, NotUsed]` so that's the return type for this method.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: This is returning a `Seq` of 2-tuples, where the tuple consists of the entity id and the `Entity` object. The entity id is the db's primary key, which is a Long.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: Ah, but since these are `Seq`s and not `List`s, `:::` isn't applicable; `Seq` only has `++`. I'd have to convert them to `List`s first: `prev.accum.toList ::: curr.accum.toList` and then convert (even implicitly) the resultant `List` back to a `Seq`.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: mutable variables are generally frowned upon in Scala - so this is a hint to anyone reading the code who might be assuming everything is an immutable `val`. I tried to be very careful with this mutable code - and since I can't change the signatures of the methods I'm overriding in `GraphStage` I needed to use a `var`.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/local/CaseSensitivitySpec.scala
davidangb: good catch; removed.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: I struggled with this a whole bunch. The options I see are:
1. Inspect the entire result set before processing it to verify it is in the correct order. This defeats the entire purpose of streaming.
2. While iterating over the result set, detect if the current entity id is less than the previous entity id. Since we require that the result set is ordered by entity id, this would be an error case. However, since this is midstream we have already sent the `200 OK` status to the caller, and we could easily have already written some entities to the response payload. So, all we could do at this point is stop streaming after sending partial results, or potentially even encode an error response inside of a fake entity, a la`{"name": "ERROR", "entityType": "ERROR", "attributes": { "errorDescription" : "… error message here … "} }`.
3. Assume that this method is always called with an appropriate result set and enforce that via PR review and code inspection

I landed on the last, since option 2 has all kinds of weirdness about it. Do you have good suggestions on how to handle this? I will keep looking at other error-trapping options.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: updated.

core/src/test/scala/org/broadinstitute/dsde/rawls/entities/EntityServiceSpec.scala
davidangb: I've updated one of the unit tests to use refs!

calypsomatic (approved):
So far as I can understand it, looks good!

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
aaronkanzer: Hmmm this is tough ha

To the overall goal, I think option 3 is the best approach to your point -- just to clarify, what would an example incoming payload look like in terms of _what exactly_ would be ordered? Just want to make sure I understand this entirely

aaronkanzer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: The relevant code is the `order by e.id` here: https://github.com/broadinstitute/rawls/blob/6fd5c51f9fa44ed6133227d345d33042877f95b3/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L340-L343.

I suppose that technically we don't actually care about _ordering_ - we just care that all attributes for any given entity are contiguous within the input stream. Remember that the input stream has one element per attribute, and the output stream has one element per entity. The implementation of our stream processing is that as soon as we see that the previous input element and the current input element belong to different entities, we output the previous entity and start collecting a new one. If attributes belonging to the same entity were not contiguous, we'd output that entity twice, and each would have only a portion of its attributes.

So, in pseudocode, this is a good input:

| entityId | entityName | attributeId | attributeName | attributeValue |
| --- | --- | --- | --- | --- |
| 1 | entity-one | 100 | color | red |
| 1 | entity-one | 101 | size | M |
| 2 | entity-two | 102 | color | green |
| 3 | entity-three | 103 | color | blue |
| 3 | entity-three | 104 | size | L |

which will produce the desired output of three entities:

```json
[
  {"name": "entity-one", "attributes": { "color": "red", "size": "M" }},
  {"name": "entity-two", "attributes": { "color": "green"}},
  {"name": "entity-three", "attributes": { "color": "blue", "size": "L" }}
]
```

but this is a bad input:

| entityId | entityName | attributeId | attributeName | attributeValue |
| --- | --- | --- | --- | --- |
| 1 | entity-one | 100 | color | red |
| 2 | entity-two | 102 | color | green |
| 3 | entity-three | 103 | color | blue |
| 1 | entity-one | 101 | size | M |
| 3 | entity-three | 104 | size | L |

which would produce undesirable output:
```json
[
  {"name": "entity-one", "attributes": { "color": "red" }},
  {"name": "entity-two", "attributes": { "color": "green"}},
  {"name": "entity-three", "attributes": { "color": "blue" }},
  {"name": "entity-one", "attributes": { "size": "M" }},
  {"name": "entity-three", "attributes": { "size": "L"}}
]
```

aaronkanzer: Thank you for all of this 🙌 

Where my mind went to was wondering if we created a hash set alongside the streaming, where we would do a quick check of the hash set to see if an entity ID was encountered yet...but I assume that that would defeat the purpose of the streaming 🤔 

I think the `order_by` clause is probably enough then here? Let me know!

davidangb: I think the hashset approach would be essentially the same as option 2 above - we'd still be detecting an error midstream and have to figure out how to handle that. The more I think about it, though, we may want more error-handling for cases where something ELSE goes wrong, like losing the connection to the db. I'll try to handle that in the follow-on PR.

aaronkanzer: Sounds good -- thanks for all the explanations here!

PR #2248, opened 2023-03-02
AJ-242: list-all-entities streaming tweaks: logging, exception handling, test-only method organization

davidangb:
Follow-on work to #2247 

This PR adds:
* consolidation/organization of `EntityComponent` methods that are only used in unit tests
* additional logging for the list-all-entities stream
    * by default, logs when the upstream/downstream finishes and in case of an exception
    * has a clear place (`onElement`) to enable logging for each element in the stream if a developer wants to do so
* additional error-handling
    * adds a `recover` for the stream to output exception messages to the caller before shutting down the stream
    * adds a safety check for out-of-order input to the stream

This is what exception info at the end of the response looks like:
Screenshot 3-2-2023 at 02 09 PM

And here's the extra logging in the typical success case:
```
rawls [INFO] [19:15:06.514] [rawls-akka.actor.default-dispatcher-14] akka.stream.Materializer - [gatherEntities] Downstream finished, cause: SubscriptionWithCancelException$StageWasCompleted$: null
rawls [INFO] [19:15:07.076] [rawls-akka.actor.default-dispatcher-5] akka.stream.Materializer - [gatherEntities] Upstream finished.
```

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -330,24 +330,15 @@ trait EntityComponent {
 
       // Active actions: only return entities and attributes with their deleted flag set to false
 
-      private def listTypeSql(workspaceContext: Workspace, entityType: String) =
-        sql"""#${baseEntityAndAttributeSql(workspaceContext)}
-        where e.deleted = false
-        and e.entity_type = ${entityType}
-        and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}"""
-
       // almost the same as "activeActionForType" except 1) adds a sort by e.id; 2) returns a stream
       def activeStreamForType(workspaceContext: Workspace,
                               entityType: String
       ): SqlStreamingAction[Seq[EntityAndAttributesResult], EntityAndAttributesResult, Read] =
-        concatSqlActions(listTypeSql(workspaceContext, entityType), sql" order by e.id").as[EntityAndAttributesResult]
-
-      // TODO: if only used by tests, can this be removed?
-      // currently only used by listActiveEntitiesOfType, which is only used by tests
-      def activeActionForType(workspaceContext: Workspace,
-                              entityType: String
-      ): ReadAction[Seq[EntityAndAttributesResult]] =
-        listTypeSql(workspaceContext, entityType).as[EntityAndAttributesResult]
+        sql"""#${baseEntityAndAttributeSql(workspaceContext)}
+                where e.deleted = false
+                and e.entity_type = ${entityType}
+                and e.workspace_id = ${workspaceContext.workspaceIdAsUUID} order by e.id"""
+          .as[EntityAndAttributesResult]
 
       def activeActionForRefs(workspaceContext: Workspace,
                               entityRefs: Set[AttributeEntityReference]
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -330,24 +330,15 @@ trait EntityComponent {
 
       // Active actions: only return entities and attributes with their deleted flag set to false
 
-      private def listTypeSql(workspaceContext: Workspace, entityType: String) =
-        sql"""#${baseEntityAndAttributeSql(workspaceContext)}
-        where e.deleted = false
-        and e.entity_type = ${entityType}
-        and e.workspace_id = ${workspaceContext.workspaceIdAsUUID}"""
-
```
davidangb, line 340: I deleted `listTypeSql` in favor of inlining its sql where necessary.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -413,13 +413,21 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
         // marshal them into an Entity object, emit that Entity, and start a new accumulator
         // for the new/current entity
-        case (prev: AttrAccum, curr: AttrAccum) if prev.accum.head.entityRecord.id != curr.accum.head.entityRecord.id =>
+        case (prev: AttrAccum, curr: AttrAccum) if prev.accum.head.entityRecord.id < curr.accum.head.entityRecord.id =>
           entityFinished(prev.accum, curr.accum)
 
+        // midstream, we notice that the current entity's id is LESS than the previous entity's id.
+        // this breaks the assumption that entities are ordered by their ids ascending, and indicates a coding
+        // error. Throw an exception.
+        case (prev: AttrAccum, curr: AttrAccum) if prev.accum.head.entityRecord.id > curr.accum.head.entityRecord.id =>
```
jordansafer, line 426: I'm a little confused by this. What case would lead to the entity record id's being out of order? I guess this is a hypothetical error, and I'm not sure what a good example of this would be

jordansafer (commented):
Looks good! +1 on these changes, I had one question for understanding on the change where we're throwing an error if entityIds aren't increasing-

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
davidangb: ah, this was discussed at length in the first PR, see https://github.com/broadinstitute/rawls/pull/2247#discussion_r1120767592. It will only happen if a future Rawls developer changes the database query, which is definitely a hypothetical. Do you think I should leave this clause out?

jordansafer: I think it's good. I think my vote would be to catch it through unit tests if this did happen, but this is an improvement, and I don't think it's worth the effort of changing it to unit testing when we have this logic written in this PR :)

jordansafer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -257,11 +257,26 @@ trait EntityApiService extends UserInfoDirectives {
...
-                complete {
-                  entityServiceConstructor(ctx).listEntities(WorkspaceName(workspaceNamespace, workspaceName),
-                                                             entityType
+                import spray.json._
+
+                // listEntities returns a source of Entity. We will stream-output each successful entity to the user,
+                // and if an exception occurs midstream we want to output the exception (wrapped in an ErrorReport)
+                // as the final element. Akka and Spray have a hard time with the mixed element types in the stream,
+                // so we manually convert them all to JsValue here and then output the stream of JsValue.
+                val jsValueSource = entityServiceConstructor(ctx)
```
aaronkanzer, line 266: What is `jsValue` here, is it an abbreviation of some sort?

davidangb: the type of `jsValueSource` is `Future[Source[JsValue, NotUsed]]` (where `JsValue` is `spray.json.JsValue`), so this variable name is hinting at its type.

aaronkanzer (approved):

PR #2249, opened 2023-03-03
AJ-864: enable column filter on primary key; remove name filter

davidangb:
Enhancement to #2239 and #2240:

* if the user specifies the id column `${entityType}_id` in `columnFilter`, treat it exactly the same as if the user specified an entity name in the now-deleted `entityNameFilter`
* remove `entityNameFilter`

This consolidates two query parameters into one, simplifying the API slightly, and allows the `${entityType}_id` pseudo-column to function exactly the same as other columns. This should simplify UI development.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
@@ -50,70 +50,59 @@ trait EntityApiService extends UserInfoDirectives {
                          'sortDirection.?,
                          'filterTerms.?,
                          'filterOperator.?,
-                         'entityNameFilter.?,
                          'columnFilter.?
-              ) {
-                (page,
-                 pageSize,
-                 sortField,
-                 sortDirection,
-                 filterTerms,
-                 filterOperator,
-                 entityNameFilter,
-                 columnFilterStringOpt
-                ) =>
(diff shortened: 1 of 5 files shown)


aaronkanzer (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -50,70 +50,59 @@ trait EntityApiService extends UserInfoDirectives {
...
-                      k -> Try(s.map(_.toInt))
-                    }
-                    val sortDirectionTry =
-                      sortDirection.map(dir => Try(SortDirections.fromString(dir))).getOrElse(Success(Ascending))
-                    val operatorTry =
-                      filterOperator.map(op => Try(FilterOperators.fromString(op))).getOrElse(Success(And))
+              ) { (page, pageSize, sortField, sortDirection, filterTerms, filterOperator, columnFilterStringOpt) =>
+                parameterSeq { allParams =>
+                  val toIntTries = Map("page" -> page, "pageSize" -> pageSize).map { case (k, s) =>
+                    k -> Try(s.map(_.toInt))
```
aaronkanzer, line 57: What does `k, s` stand for here? Could we use more descriptive variable names?

jordansafer (approved):

PR #2254, opened 2023-03-09
[WOR-644] Tag Sentry events for billing project creation failures

marctalbott:
Ticket: [WOR-644](https://broadworkbench.atlassian.net/browse/WOR-644)

- Add custom tag to server exceptions thrown during billing project creation (includes any landing zone or billing profile failures).
- Alerts will be configured in Sentry to notify the #dsp-workspaces-alerts channel when issues with the Workspaces team tag are seen. See this test exception I set up in dev as an example - https://broad-institute.sentry.io/alerts/rules/firecloud-dev/13984611/details/

---

**PR checklist**

[WOR-644]: https://broadworkbench.atlassian.net/browse/WOR-644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -63,7 +63,7 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
     def tagAndCaptureSentryEvent(e: Throwable): Unit = {
       val sentryEvent = new SentryEvent()
-      sentryEvent.setTag("team", "workspaces")
+      sentryEvent.setTag("component", "billing")
       sentryEvent.setThrowable(e)
       Sentry.captureEvent(sentryEvent)
       throw e
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -1,7 +1,11 @@
 package org.broadinstitute.dsde.rawls.billing
 
 import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.model.StatusCodes.ServerError
+import bio.terra.profile.client.{ApiException => BpmApiException}
+import bio.terra.workspace.client.{ApiException => WsmApiException}
 import com.typesafe.scalalogging.LazyLogging
+import io.sentry.{Sentry, SentryEvent}
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
 import org.broadinstitute.dsde.rawls.model.{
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -57,15 +61,22 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
    * Creates a "v2" billing project, using either Azure managed app coordinates or a Google Billing Account
    */
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
+    def tagAndCaptureSentryEvent(e: Throwable): Unit = {
```
marctalbott, line 64: This would be worth extracting into a separate class and renaming it `captureWorkspacesSentryEvent` or something along those lines if we want to alert more often on other failing operations we care about (workspace CRUD, spend reporting, etc.). Leaving it here right now as it's all that's required for the current scope of the ticket. Absolutely open to expanding the scope of the ticket and adding these events to other Workspaces-owned operations if others think it'd be worth doing all at once -- it'd be a small-ish lift

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -84,7 +95,14 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
         None
       )
-    } yield {}
+    } yield {}).recover {
+      case e: RawlsExceptionWithErrorReport =>
+        e.errorReport.statusCode.collect { case _: ServerError =>
+          tagAndCaptureSentryEvent(e)
+        }
+      case wsmException: WsmApiException if wsmException.getCode >= 500 => tagAndCaptureSentryEvent(wsmException)
+      case bpmException: BpmApiException if bpmException.getCode >= 500 => tagAndCaptureSentryEvent(bpmException)
```
marctalbott, line 106: There might be a less repetitive way to do this, but it wasn't coming to mind and at least this is pretty clear about which exceptions we want to alert on.

marctalbott (commented):
One reservation I have here is putting our current org layout into the code (ie tagging the event as a workspaces team event) because it adds to the cost of any potential reorgs in the future. I started with this as it felt like the clearest tagging system, but I'm now leaning towards using a tag related to the actual Rawls functionality as it may be more future-proof (I hope the Workspaces team outlasts Rawls, but I've hoped that about all of my past teams 🙃). A tag like `operation -> billingProjectCreation` would work just as well for Sentry alerting and would make our lives easier in any future re-orgs

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
blakery: I really like this style of alerting
Not needing to dig for the stack trace is great.
Alert => click => see problem

I'd be in favor or using this more generally - although we should probably try it out for realsies before going too ham.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
blakery: There are some ways, but since it's just a simple function call, I think it's good as is

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
aherbst-broad: Agreed, let's get confidence in this approach before using it elsewhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -57,15 +61,22 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
    * Creates a "v2" billing project, using either Azure managed app coordinates or a Google Billing Account
    */
   def createBillingProjectV2(createProjectRequest: CreateRawlsV2BillingProjectFullRequest): Future[Unit] = {
+    def tagAndCaptureSentryEvent(e: Throwable): Unit = {
+      val sentryEvent = new SentryEvent()
+      sentryEvent.setTag("team", "workspaces")
```
aherbst-broad: Small bit of pushback on the nomenclature here; I don't think we should bake the notion of the specific team into this code but rather the functional area of concern. That is, there may come a day when the "Broad Workspaces" team is no longer responsible for billing project failures and it'd be a bit of a pain to have to ship a PR to reflect that reality. 


Perhaps ("component","billing")?  Swat tests are setup [similarly](https://github.com/broadinstitute/rawls/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsTestGroupAnnotations.java).

aherbst-broad (approved):
Looks good, had a comment about the naming.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
marctalbott: Agreed. I like `component`, I had been thinking `operation` but broader would be better

blakery (approved):

PR #2255, opened 2023-03-10
WOR-772 Add handling for landing zone delete jobs to resource monitor

blakery:
Ticket: [WOR-772](https://broadworkbench.atlassian.net/browse/WOR-772)

1. Split BP creation into two steps: an initial phase, and finalize phase
* Initial phase (previously pre deletion steps) is invoked by the orchestrator
* The orchestrator schedules a job to call the finalize step
2. Add runners for the deletion jobs

~~When we were talking about how to handle the deletion for gcp billing projects, we were debating how far to go right now, and how much work it would be. It turned out to be super simple to just make the gcp delete work the same way, so that's the way it is currently structured in this PR. But it would also be easy to change the orchestrator to call the finalize step immediately for gcp projects. I thought it was more logical and easier to follow as currently structured, so that's how I left it for now.~~
I put back the synchronous behavior for google project deletion, because the async was breaking swatomation

I was able to delete both GCP and Azure billing projects with these changes.

---

**PR checklist**

[WOR-772]: https://broadworkbench.atlassian.net/browse/WOR-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -208,46 +214,39 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+          projectName,
+          ctx.userInfo.userEmail,
+          monitorJobType
+        )
+      )
+      _ <- billingRepository.updateCreationStatus(projectName, CreationStatuses.Deleting, None)
+    } yield ()
 
-  // This code also lives in UserService.
-  def unregisterBillingProjectWithUserInfo(projectName: RawlsBillingProjectName,
```
blakery, line 216: Moved to the `BillingProjectLifecycle`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -208,46 +214,39 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
         case Some(_) => bpmBillingProjectLifecycle
       }
       _ <- billingRepository.failUnlessHasNoWorkspaces(projectName)
-      _ <- projectLifecycle.preDeletionSteps(projectName, ctx)
-      _ <- unregisterBillingProjectWithUserInfo(projectName, ctx.userInfo)
-    } yield {}
+      (jobControlId, monitorJobType) <- projectLifecycle.initiateDelete(projectName, ctx)
+      _ <- resourceMonitorRecordDao.create(
```
blakery: If we wanted to keep the synchronous behavior for GCP projects for now, we would either match on the monitorJobType, or return an additional value when we choose the lifecycle, and then for GCP projects, call the finalizeDeletion step, instead of scheduling a job.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -267,13 +266,3 @@ object BillingProjectOrchestrator {
     )
   }
 }
-
-class DuplicateBillingProjectException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
```
blakery, line 271: Moved to `BillingProjectLifecycle`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -22,19 +28,20 @@ import org.broadinstitute.dsde.rawls.model.{
...
-  billingRepository: BillingRepository,
+  val samDAO: SamDAO,
+  val billingRepository: BillingRepository,
   billingProfileManagerDAO: BillingProfileManagerDAO,
-  workspaceManagerDAO: HttpWorkspaceManagerDAO,
+  workspaceManagerDAO: WorkspaceManagerDAO,
   resourceMonitorRecordDao: WorkspaceManagerResourceMonitorRecordDao
 )(implicit val executionContext: ExecutionContext)
-    extends BillingProjectLifecycle
-    with LazyLogging {
```
blakery, line 37: This still uses `LazyLogging`: it just gets it by way of `BillingProjectLifecycle`, where it was needed for `unregisterBillingProject`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -176,33 +184,39 @@ class BpmBillingProjectLifecycle(
...
+    landingZoneId: UUID,
+    ctx: RawlsRequestContext
+  ): DeleteAzureLandingZoneResult = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+    case Failure(e: ApiException) =>
+      val msg = s"Unable to delete landing zone: ${e.getMessage}"
+      throw RawlsExceptionWithErrorReport(StatusCode.int2StatusCode(e.getCode), msg)
+    case Failure(t) =>
       logger.warn(
-        s"Unable to delete landing zone with ID $landingZoneId for BPM-backed billing project ${projectName.value}.",
+        s"Unable to delete landing zone with ID $landingZoneId for BPM-backed billing project.",
```
blakery: The resulting error message gets stored on the project, now - so it seemed superfluous to pass in the project name just to tack it on to an error message.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -212,7 +226,7 @@ class BpmBillingProjectLifecycle(
   private def cleanupBillingProfile(profileModelId: UUID,
                                     projectName: RawlsBillingProjectName,
                                     ctx: RawlsRequestContext
-  ): Future[Unit] = {
+  ): Future[Boolean] = {
```
blakery: I think I changed this to return a boolean just to make some debugging quicker - it seems harmless to leave it in

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -54,10 +66,19 @@ class GoogleBillingProjectLifecycle(samDAO: SamDAO, gcsDAO: GoogleServicesDAO)(i
...
     // lives in a companion object (which makes straight mocking impossible), and the method will be removed
     // once workspace migration is complete. Note also that the more "integration" level test BillingApiServiceV2Spec
     // does verify that code in this method is executed when a Google-based project is deleted.
     deleteGoogleProjectIfChild(projectName, ctx.userInfo, gcsDAO, samDAO, ctx)
+      .map(_ => (UUID.randomUUID(), GoogleBillingProjectDelete))
+
+  override def finalizeDelete(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
+    executionContext: ExecutionContext
+  ): Future[Unit] =
+    unregisterBillingProject(projectName, ctx)
```
blakery, line 81: This is just what the orchestrator was doing after the `preDeletionSteps` was finished

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
```diff
@@ -15,6 +15,13 @@ object WorkspaceManagerResourceMonitorRecord {
     type JobType = Value
     val AzureLandingZoneResult: Value = Value("AzureLandingZoneResult")
     val CloneWorkspaceContainerResult: Value = Value("CloneWorkspaceContainerResult")
+
+    val GoogleBillingProjectDelete: Value = Value("GoogleBillingProjectDelete")
+    val AzureBillingProjectDelete: Value = Value("AzureBillingProjectDelete")
+    // bpm based project with no azure landing zone to delete
+    // AzureBillingProjectDelete and OtherBpmBillingProjectDelete map to the same runner
+    // This is a separate type to reflect previous synchronous behavior
+    val OtherBpmBillingProjectDelete: Value = Value("OtherBpmBillingProjectDelete")
```
blakery: it looks kludgy, but I wanted to preserve the existing behavior

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/CloneWorkspaceContainerRunner.scala
```diff
@@ -128,9 +129,4 @@ class CloneWorkspaceContainerRunner(
     dataAccess.workspaceQuery.findById(wsId.toString)
   }
 
-  def getUserCtx(userEmail: String)(implicit executionContext: ExecutionContext): Future[RawlsRequestContext] = for {
```
blakery, line 131: This was just moved to a common interface. It was the third time needing it.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/GoogleBillingProjectDeleteRunner.scala
```diff
@@ -0,0 +1,75 @@
...
+  Complete,
+  Incomplete,
+  JobStatus
+}
+import org.broadinstitute.dsde.rawls.model.{CreationStatuses, RawlsBillingProjectName}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+class GoogleBillingProjectDeleteRunner(
```
blakery: This is super simple - most of it is just sanity checking some values. Then it just calls the same code the orchestrator was, before.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -917,11 +920,6 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
       verify(services.samDAO).deleteUserPetServiceAccount(ArgumentMatchers.eq(project.googleProjectId),
                                                           any[RawlsRequestContext]
       )
-      verify(services.samDAO).deleteResource(
```
blakery, line 920: Hot Take: These checks add no value, and we should remove them. The only thing they check is if we've set up enough mocks to not throw unexpected NPEs.

cahrens:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -176,33 +184,39 @@ class BpmBillingProjectLifecycle(
...
+    *  starts a landing zone deletion job
+    *  does not ensure that the landing zone deletion completes successfully.
+    */
+  private def cleanupLandingZone(
+    landingZoneId: UUID,
+    ctx: RawlsRequestContext
+  ): DeleteAzureLandingZoneResult = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+    case Failure(e: ApiException) =>
+      val msg = s"Unable to delete landing zone: ${e.getMessage}"
+      throw RawlsExceptionWithErrorReport(StatusCode.int2StatusCode(e.getCode), msg)
```
cahrens: Should you throw a more specific type of exception here? Perhaps `BillingProjectDeletionException` or `LandingZoneDeletionException`?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -176,33 +184,39 @@ class BpmBillingProjectLifecycle(
...
+    case Failure(t) =>
       logger.warn(
-        s"Unable to delete landing zone with ID $landingZoneId for BPM-backed billing project ${projectName.value}.",
+        s"Unable to delete landing zone with ID $landingZoneId for BPM-backed billing project.",
         t
       )
-    }
+      throw t
```
cahrens: Same note here about type of exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -176,33 +184,39 @@ class BpmBillingProjectLifecycle(
...
+      logger.info(
+        s"Initiated deletion of landing zone $landingZoneId for BPM-backed billing project."
+      )
+      Option(landingZoneResponse.getErrorReport) match {
+        case Some(errorReport) =>
+          val msg = s"Unable to delete landing zone with ID $landingZoneId for BPM-backed " +
+            s"billing project: ${errorReport.getMessage}."
+          logger.warn(msg)
+          val status = Option(errorReport.getStatusCode).map(code => StatusCode.int2StatusCode(code))
+          throw RawlsExceptionWithErrorReport(
```
cahrens: Same note here about type of exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: But I find it confusing to understand why it is returning true vs. false. If the boolean stays, you need to update the documentation to explain what it means.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -242,21 +260,28 @@ class BpmBillingProjectLifecycle(
           )
         case _ => ()
       }
-      _ <- billingRepository.getLandingZoneId(projectName).flatMap {
+      (jobControlId, eventType) <- billingRepository.getLandingZoneId(projectName).map {
         case Some(landingZoneId) =>
```
cahrens, line 264: Is `landingZoneId` the same as `jobControlId`? If so, please make the names consistent.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -242,21 +260,28 @@ class BpmBillingProjectLifecycle(
...
-      _ <- billingRepository.getBillingProfileId(projectName).flatMap {
-        case Some(billingProfileId) =>
-          cleanupBillingProfile(UUID.fromString(billingProfileId), projectName, ctx)
-        case None =>
-          logger.warn(
-            s"Deleting BPM-backed billing project ${projectName}, but no associated billing profile record to delete"
-          )
-          Future.successful()
+          logger.warn(s"Deleting BPM-backed billing project $projectName, but no associated landing zone to delete")
+          (UUID.randomUUID(), OtherBpmBillingProjectDelete)
```
cahrens: Why return a random UUID here?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -242,21 +260,28 @@ class BpmBillingProjectLifecycle(
...
+
+  override def finalizeDelete(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
+    executionContext: ExecutionContext
+  ): Future[Unit] = for {
+    billingProfileId <- billingRepository.getBillingProfileId(projectName)
+    _ <- billingProfileId match {
+      case Some(id) => cleanupBillingProfile(UUID.fromString(id), projectName, ctx)
+      case None =>
+        logger.warn(
+          s"Deleting BPM-backed billing project $projectName, but no associated billing profile record was deleted"
```
cahrens: I think it would be good to make it clear that the billing profile record wasn't deleted because none existed-- this message makes it seem like it could have been there, but we didn't delete it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -54,10 +66,19 @@ class GoogleBillingProjectLifecycle(samDAO: SamDAO, gcsDAO: GoogleServicesDAO)(i
...
-  override def preDeletionSteps(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext): Future[Unit] =
+  override def initiateDelete(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
+    executionContext: ExecutionContext
+  ): Future[(UUID, JobType)] =
     // Note: GoogleBillingProjectLifecycleSpec does not test that this method is called because the method
     // lives in a companion object (which makes straight mocking impossible), and the method will be removed
     // once workspace migration is complete. Note also that the more "integration" level test BillingApiServiceV2Spec
     // does verify that code in this method is executed when a Google-based project is deleted.
     deleteGoogleProjectIfChild(projectName, ctx.userInfo, gcsDAO, samDAO, ctx)
+      .map(_ => (UUID.randomUUID(), GoogleBillingProjectDelete))
```
cahrens: Again, the usage of a random UUID is confusing. If the UUID is not necessary for these types, can you make the argument an Option?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -10,17 +15,24 @@ import org.broadinstitute.dsde.rawls.model.{
...
   deleteGoogleProjectIfChild,
   syncBillingProjectOwnerPolicyToGoogleAndGetEmail
 }
 
+import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
 
-class GoogleBillingProjectLifecycle(samDAO: SamDAO, gcsDAO: GoogleServicesDAO)(implicit
+class GoogleBillingProjectLifecycle(
+  val billingRepository: BillingRepository,
```
cahrens, line 32: We don't yet use the `BillingRepository` for GCP billing projects, so passing it in feels a bit odd. But on the other hand, I know that we hope to use it relatively soon.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -400,20 +410,40 @@ object BootMonitors extends LazyLogging {
...
         dataSource,
         Map(
           JobType.AzureLandingZoneResult ->
-            new LandingZoneCreationStatusRunner(samDAO, workspaceManagerDAO, new BillingRepository(dataSource), gcsDAO),
+            new LandingZoneCreationStatusRunner(samDAO, workspaceManagerDAO, billingRepo, gcsDAO),
           JobType.CloneWorkspaceContainerResult ->
-            new CloneWorkspaceContainerRunner(samDAO, workspaceManagerDAO, dataSource, gcsDAO)
+            new CloneWorkspaceContainerRunner(samDAO, workspaceManagerDAO, dataSource, gcsDAO),
+          JobType.AzureBillingProjectDelete -> bpmBillingProjectDeleteRunner,
+          JobType.OtherBpmBillingProjectDelete -> bpmBillingProjectDeleteRunner
```
cahrens: Why is the mapping for `GoogleBillingProjectDeleteRunner` not here?

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycleSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.concurrent.{Await, ExecutionContext, Future}
+// so we don't have to define a new type in order to call test methods in our instantiations of the interface
+import scala.language.reflectiveCalls
+
+class BillingProjectLifecycleSpec extends AnyFlatSpec {
+  implicit val executionContext: ExecutionContext = TestExecutionContext.testExecutionContext
+
+  behavior of "unregisterBillingProject"
+
+  it should "delete the project even if Sam deleteResource fails" in {
```
cahrens, line 28: Was this pre-existing behavior to delete the project in this case?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: good point

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -374,176 +413,175 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
     )
 
     Await.result(bpo.deleteBillingProjectV2(billingProjectName), Duration.Inf)
 
-    verify(billingRepository, Mockito.times(1)).failUnlessHasNoWorkspaces(billingProjectName)(executionContext)
-    verify(billingRepository, Mockito.times(1)).deleteBillingProject(billingProjectName)
-    verify(billingProjectLifecycle, Mockito.times(1)).preDeletionSteps(billingProjectName, testContext)
-    verify(samDAO, Mockito.times(1))
-      .deleteResource(SamResourceTypeNames.billingProject, billingProjectName.value, testContext)
+    verify(monitorRecordDao).create(ArgumentMatchers.argThat(matchedExpectedEvent))
```
cahrens, line 523: I would definitely be tempted to fold this into the previous test case.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -568,63 +603,63 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
...
+    verify(workspaceManagerDAO, Mockito.never()).deleteLandingZone(ArgumentMatchers.any(), ArgumentMatchers.any())
+    verify(bpm, Mockito.never()).deleteBillingProfile(ArgumentMatchers.any[UUID], ArgumentMatchers.eq(testContext))
   }
 
   it should "delete the landing zone if the id exists" in {
     val repo = mock[BillingRepository]
     when(repo.getCreationStatus(billingProjectName)).thenReturn(Future.successful(CreationStatuses.Ready))
     when(repo.getLandingZoneId(billingProjectName)).thenReturn(Future.successful(Some(landingZoneId.toString)))
-    // Mock no associated billing profile to delete
-    when(repo.getBillingProfileId(billingProjectName)).thenReturn(Future.successful(None))
```
cahrens, line 611: Why is this not needed?

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -568,63 +603,63 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
...
 
     val bpm = mock[BillingProfileManagerDAO]
     val workspaceManagerDAO = mock[HttpWorkspaceManagerDAO]
     when(workspaceManagerDAO.deleteLandingZone(landingZoneId, testContext))
-      .thenReturn(new DeleteAzureLandingZoneResult().jobReport(new JobReport().id("fake-id")))
+      .thenReturn(
+        new DeleteAzureLandingZoneResult().jobReport(new JobReport().id("C1024C05-40A6-4A12-B12E-028E445AEC3B"))
```
cahrens: Is the actual UUID meaningful?

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -651,63 +686,62 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
     val bpm = mock[BillingProfileManagerDAO]
     val workspaceManagerDAO = mock[HttpWorkspaceManagerDAO]
     when(workspaceManagerDAO.deleteLandingZone(landingZoneId, testContext)).thenReturn(
-      new DeleteAzureLandingZoneResult().errorReport(new ErrorReport().statusCode(500).message(landingZoneErrorMessage))
+      new DeleteAzureLandingZoneResult()
+        .landingZoneId(UUID.fromString("c1024c05-40a6-4a12-b12e-028e445aec3b"))
```
cahrens: Is the actual UUID meaningful?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,464 @@
...
+      ArgumentMatchers.eq(DeletionFailed),
+      ArgumentMatchers.any[Some[String]]()
+    )
+  }
+
+  it should "set an error status and message on the project and return incomplete when landing zone call returns 500" in {
+    val ctx = mock[RawlsRequestContext]
+    val wsmDao = mock[WorkspaceManagerDAO]
+    val wsmExceptionMessage = "looking for this to be reported in the billing project message"
+    val landingZoneId = "7db89d7c-ef8b-4eaa-aef8-a62bd66cb095"
```
cahrens, line 140: If the landing zone ID is not meaningful, I think it's better to create a random one.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: The orchestrator needs a UUID so it can schedule the follow-up job.

If there wasn't a landing zone to delete, it still needs to schedule the job.

I think your point below or making it an `Option`, and then letting the orchestrator generate the UUID if it needs it is the way to go, though.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,464 @@
...
+
+class BPMBillingProjectDeleteRunnerSpec extends AnyFlatSpec with MockitoSugar with Matchers with ScalaFutures {
+  implicit val executionContext: ExecutionContext = TestExecutionContext.testExecutionContext
+
+  behavior of "initial setup and basic requirements updating billing project status"
+
+  val userEmail = "user@email.com"
+  val billingProjectName: RawlsBillingProjectName = RawlsBillingProjectName("fake_name")
+
+  it should "return a completed status if the billing project name is None" in {
```
cahrens: ID is none?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,464 @@
...
+    val runner = new BPMBillingProjectDeleteRunner(
+      mock[SamDAO],
+      mock[GoogleServicesDAO],
+      mock[WorkspaceManagerDAO],
+      mock[BillingRepository],
+      mock[BillingProjectLifecycle]
+    )
+    whenReady(runner(monitorRecord.copy(billingProjectId = None)))(
+      _ shouldBe WorkspaceManagerResourceMonitorRecord.Complete
+    )
```
cahrens, line 49: Verify the args to `updateCreationStatus`?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,464 @@
...
+      ArgumentMatchers.eq(DeletionFailed),
+      ArgumentMatchers.any[Some[String]]()
+    )
+  }
+
+  it should "set an error status and message from the error report for failed jobs" in {
+    val ctx = mock[RawlsRequestContext]
+    val wsmDao = mock[WorkspaceManagerDAO]
+    val wsmExceptionMessage = "looking for this to be reported in the billing project message"
+    val landingZoneId = "7db89d7c-ef8b-4eaa-aef8-a62bd66cb095"
```
cahrens: If the landing zone ID is not meaningful, I think it's better to create a random one (same comment for subsequent test cases below).

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,464 @@
...
+    whenReady(runner(monitorRecord))(_ shouldBe WorkspaceManagerResourceMonitorRecord.Complete)
+
+    verify(billingRepository).updateCreationStatus(
+      ArgumentMatchers.eq(billingProjectName),
+      ArgumentMatchers.eq(DeletionFailed),
+      ArgumentMatchers.any[Some[String]]()
+    )
+  }
+
+  it should "continue running when the landing zone delete job is still running" in {}
```
cahrens: Looks like you forgot to come back to this test case.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycleSpec.scala
blakery: yes
there's even a comment in the function that this is the desired behavior

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -221,7 +202,7 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
           resourceMonitorRecordDao
             .create(
               WorkspaceManagerResourceMonitorRecord.forBillingProjectDelete(
-                jobControlId,
+                jobControlId.getOrElse(UUID.randomUUID()),
```
cahrens, line 212: Why is the random UUID needed here?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
blakery: I removed the `GoogleBillingProjectDeleteRunner`, because I put back the sync behavior for GCP projects, in order to avoid breaking swatomation

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
cahrens: Never mind, just saw your comment.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
cahrens: But the class still exists. Should it then be deleted?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
blakery: It is used: it's required by a method in the shared trait, called in the finalizeDelete step.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
blakery: This was originally a test in the orchestrator. By moving it to the lifecycle, it narrows the scope of what we're testing, and eliminates the need to mock that call.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
blakery: It's a `RawlsBillingProjectName`, and frequently referenced as 'name', so that's what I defaulted to. The error message in the runner class just says 'no billing project is set', so I'll just use that wording.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
blakery: There is no call up `updateCreationStatus` here, because there's no billing project set on the job, and therefore no call to update it.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
cahrens: I was reacting to `monitorRecord.copy(billingProjectId = None)` and didn't realize that is a `RawlsBillingProjectName`. In retrospect, perhaps that variable should have been called something else.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
blakery: I've been reading up on pact, and how random values that change across test runs should be avoided. 
It isn't applicable here, and this project is a long way from being able to effectively use contracts, but it stuck in my head. 
I don't have a problem generating a random uuid here, if it makes it clearer.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
cahrens: Ah, right!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
cahrens: To me it makes it clearer that the value is not meaningful (ie., it isn't set to that value in order to match up with other test data). But there could be other ways to achieve that if you want a consistent value. For instance, just naming the variable "dummyLandingZoneId" or putting a code comment that the specific value is not important would help me.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
blakery: fixed

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
blakery: I removed all usages, but not the class itself. Deleted.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: I was using the existing message - you're right though: updated message for clarity.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
cahrens: @blakery I still thinks this should be backed out, or the meaning of the boolean should be documented.

blakery: I didn't see your comment earlier - removed the return value

cahrens: I recall there was also a mock in a test that returned a value because of this.

blakery: There shouldn't be; it's a private method.

cahrens: Sorry, I saw [this](https://github.com/broadinstitute/rawls/pull/2255/files#diff-95269d8fdd62fc54d856044c882baeafa08f98d9a9a3919025f51fd3594d495eR104) and assumed it was related. What is the return value here for?

cahrens: Ugh, link isn't working… I think because the file is collapsed by default. I'll find it again and put a comment directly there. https://github.com/broadinstitute/rawls/pull/2255/files#r1134477026

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
```diff
@@ -0,0 +1,500 @@
...
+      billingRepository.updateCreationStatus(
+        ArgumentMatchers.eq(billingProjectName),
+        ArgumentMatchers.eq(DeletionFailed),
+        ArgumentMatchers.any[Some[String]]()
+      )
+    ).thenAnswer { invocation =>
+      val message: Option[String] = invocation.getArgument(2)
+      assert(message.get.toLowerCase.contains("request context"))
+      assert(message.get.contains(userEmail))
+      Future.successful(1)
```
cahrens, line 104: @blakery why does this return 1?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycle.scala
```diff
@@ -1,29 +1,81 @@
...
-                                            ctx: RawlsRequestContext
+trait BillingProjectLifecycle extends LazyLogging {
+
+  // It's probably reasonable to expect that the billing project lifecyle always includes these two resources
+  val samDAO: SamDAO
+  val billingRepository: BillingRepository
+
+  // This code also lives in UserService as unregisterBillingProjectWithUserInfo
+  // if this was scala 3.x, we could just use a parameterized trait and this would work basically everywhere
+  protected def unregisterBillingProject(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
```
marctalbott: `rollbackCreateV2BillingProjectInternal` in the orchestrator does basically the same thing too -- might be worth refactoring that to just call this

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -208,46 +215,47 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
         case Some(_) => bpmBillingProjectLifecycle
       }
       _ <- billingRepository.failUnlessHasNoWorkspaces(projectName)
-      _ <- projectLifecycle.preDeletionSteps(projectName, ctx)
-      _ <- unregisterBillingProjectWithUserInfo(projectName, ctx.userInfo)
-    } yield {}
+      (jobControlId, monitorJobType) <- projectLifecycle.initiateDelete(projectName, ctx)
+      _ <- monitorJobType match {
+        case GoogleBillingProjectDelete => projectLifecycle.finalizeDelete(projectName, ctx)
+        case _ =>
```
marctalbott: I'd prefer to see specific JobTypes here -- if a `CloneWorkspaceContainerResult` job were to show up somehow, we'd need to throw

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -221,17 +231,19 @@ class BpmBillingProjectLifecycle(
...
     }
   }
 
-  override def preDeletionSteps(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext): Future[Unit] =
+  override def initiateDelete(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
+    executionContext: ExecutionContext
+  ): Future[(Option[UUID], JobType)] =
     for {
       _ <- billingRepository.getCreationStatus(projectName).map {
         case CreationStatuses.CreatingLandingZone =>
```
marctalbott: We may want to move this status check up into the Orchestrator and prevent deletion unless the billing project is in the `Ready` or `Error` state (and maybe `DeletionFailed`, I haven't gotten that far in the PR)

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -242,21 +254,28 @@ class BpmBillingProjectLifecycle(
...
+    executionContext: ExecutionContext
+  ): Future[Unit] = for {
+    billingProfileId <- billingRepository.getBillingProfileId(projectName)
+    _ <- billingProfileId match {
+      case Some(id) => cleanupBillingProfile(UUID.fromString(id), projectName, ctx)
+      case None =>
+        logger.warn(
+          s"Deleting BPM-backed billing project $projectName, but no associated billing profile record to delete"
+        )
+        Future.successful(false)
```
marctalbott: Why does this return false?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -0,0 +1,131 @@
...
+          .map(_ => Complete)
+    }
+    getUserCtx(userEmail).transformWith {
+      case Success(userCtx) =>
+        job.jobType match {
+          case AzureBillingProjectDelete =>
+            billingRepository.getLandingZoneId(projectName).flatMap {
+              case Some(landingZoneId) =>
+                handleLandingZoneDeletion(job.jobControlId, projectName, UUID.fromString(landingZoneId), userCtx)
+              case None =>
```
marctalbott: I'm a bit confused about why this is error-worthy here, but when we initiate the delete we're okay with a BPM billing project not having a landing zone. If we want to enforce that all BPM billing projects have a landing zone, we should do that. If we don't, then I think we can get rid of the `OtherBpmBillingProjectDelete` job type entirely and make this runner finalize deletion immediately if there's no landing zone.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -0,0 +1,131 @@
...
+                val msg = s"No landing zone id set on Azure billing project"
+                logger.error(msg)
+                billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Complete)
+            }
+          case _ => billingLifecycle.finalizeDelete(projectName, userCtx).map(_ => Complete)
+        }
+      case Failure(t) =>
+        val msg = s"Unable to complete billing project deletion: unable to retrieve request context for $userEmail"
+        logger.error(s"${job.jobType} job ${job.jobControlId} for billing project: $projectName failed: $msg", t)
+        billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Incomplete)
```
marctalbott: I don't fully understand why we're setting the `CreationStatus` to `DeletionFailed` but the job status is still `Incomplete`. It seems to me like we only want to set `CreationStatus` when we're giving up on deletion, but the `Incomplete` job status means we'll retry deletion right?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunnerSpec.scala
blakery: That's because the method in the `BillingRepository` returns the number of objects updated.
It wasn't anything I touched at all in this PR - just mocked it

cahrens (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycleSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+  SamResourceTypeNames
+}
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatestplus.mockito.MockitoSugar.mock
+import org.mockito.Mockito.{verify, when, RETURNS_SMART_NULLS}
+
+import java.util.UUID
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext, Future}
+// so we don't have to define a new type in order to call test methods in our instantiations of the interface
```
marctalbott: I'm not really grokking this comment -- is it required for `exposeUnregisterBillingProject`?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
marctalbott: Not sure I agree with that. These are verifying that the Sam call to delete the `billing-project` resource is actually being made by whatever service makes that call. Without it, we won't know if there was a bug in `unregisterBillingProject` and it wasn't actually trying to delete the `billing-project` resource which might lead to us leaking Sam resources.

Whether _this_ is the place for us to verify that is a different question. It doesn't really feel like the best place for it, but I'd prefer to see the check get moved to the proper place than get deleted entirely

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycle.scala
blakery: Well spotted - always happy to remove unnecessary code

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: I'm not sure the orchestrator is the correct place for this. The behavior for different statuses could differ significantly based on the type of project, and I think it's appropriate for that differentiation of behavior to be in the `BillingProjectLifecycle`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: It was matching the return type of `cleanupBillingProfile` which was returning a boolean earlier on, and I missed this in the clean-up (and it worked with types, since the return value was ignored)
cleaned up now

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectLifecycleSpec.scala
blakery: yes - otherwise, since we're just creating an anonymous instantiation of the interface, the compiler see `exposeUnregisterBillingProject`. I think this was fixed in scala 3 
But `exposeUnregisterBillingProject` is no longer needed, since I changed the visibility on `unregisterBillingProject`, so I'll remove the comment as well

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
blakery: This way we set the status and message so the user sees what's going on, so something doesn't fail silently indefinitely if there's a real problem; and we're retrying because this could easily be caused by something transient, like a networking error.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
blakery: I think the intent was to preserve some existing functionality, but it might not have mapped cleanly when splitting into phases.
`OtherBpmBillingProjectDelete` is keyed on the presence of the landing zone, so as written, we _shouldn't_ get into this particular state, and it's really just a loophole in how things have been defined.

I think that the correct way to do this is log the message here (and not on the init side) and proceed to the deletion. Then instead of both job types just have `BpmBillingProjectDelete`.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
marctalbott: What do you think of keeping the status as `Deleting` but adding the error message?

Alternatively, we could split `DeletionFailed` to two separate statuses, `RetryableDeletionFailure` and `FatalDeletionFailure`.

I'm mostly concerned about our ability to debug billing project deletion failures. It would be nice to be able to tell if Rawls is actively retrying for a given billing project or whether it's given up.

On a separate note, it also might be worth adding a retry limit to keep billing projects from getting endlessly retried

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
marctalbott: Fair enough, I think I'd still like to see us checking other statuses to make sure we're only deleting projects that we're confident we can delete. 

It seems like we should throw if the billing project status is `Creating`, `Deleting` or `AddingToPerimeter`

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
blakery: This got cleaned up with the job types and handling of billing projects with no landing zone, discussed below

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
blakery: Those checks are already done in unit tests for the appropriate classes (and with considerably more specificity).
These were written when the overall unit test coverage for the project was much sparser.

For the record, I wasn't necessarily saying they needed to be removed _now_ - more just provoking discussion (and a bit of venting frustration, tbh).

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
blakery: I like the idea of keeping the `Deleting` status, but still setting the error message: it gives a pretty clean representation of the state and intent of the service.

The problem with the retry limit part is determining what that limit should be (it's nearly always going to be arbitrary), and tracking the state (add another field to the job?).
Also the retry period is on the order of a minute or two, and this is an edge case - so I think it would just be introducing unnecessary complexity.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: There's a collection of terminal states for the statuses.
It ended up being cleanest to just check in the orchestrator that the project is in one of those states.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -63,24 +62,18 @@ class BPMBillingProjectDeleteRunner(
...
+        billingRepository.getLandingZoneId(projectName).flatMap {
+          case Some(landingZoneId) =>
+            handleLandingZoneDeletion(job.jobControlId, projectName, UUID.fromString(landingZoneId), userCtx)
+          case None => billingLifecycle.finalizeDelete(projectName, userCtx).map(_ => Complete)
         }
       case Failure(t) =>
         val msg = s"Unable to complete billing project deletion: unable to retrieve request context for $userEmail"
         logger.error(s"${job.jobType} job ${job.jobControlId} for billing project: $projectName failed: $msg", t)
-        billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Incomplete)
+        billingRepository.updateCreationStatus(projectName, Deleting, Some(msg)).map(_ => Incomplete)
```
marctalbott, line 76: There's a few other cases where we're setting the status as `DeletionFailed` with an `Incomplete` status. They all look like similar cases where we expect the failure to be transient and we want to retry. Can we keep those statuses as `Deleting` as well? I think `DeletionFailed` will be a more helpful status for us if we can treat it as a fatal error that requires some manual intervention

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -210,28 +189,33 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+            ErrorReport(
+              s"Billing project ${projectName.value} cannot be deleted when in status of $status"
+            )
+          )
+      }
+      jobControlId <- projectLifecycle.initiateDelete(projectName, ctx)
+      _ <- deleteType match {
         case GoogleBillingProjectDelete => projectLifecycle.finalizeDelete(projectName, ctx)
-        case _ =>
+        case BpmBillingProjectDelete =>
```
marctalbott, line 208: Is Scala okay with this being a non-exhaustive match? I would've expected you to need a `_` case that throws here

marctalbott (approved):
Just a couple of comments -- looking pretty good

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
blakery: it knows that it's either one or the other, because it's assigned up above, and those are the only two possibilities assigned

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
cahrens: @blakery @marctalbott did this get followed up on?

PR #2260, opened 2023-03-15
WOR-722 - follow up for billing project status clean-up

blakery:
Ticket: [WOR-772](https://broadworkbench.atlassian.net/browse/WOR-772)

clean up project statuses for consistency
[Follow up for changes requested here](https://github.com/broadinstitute/rawls/pull/2255#discussion_r1136045139)

**PR checklist**

[WOR-772]: https://broadworkbench.atlassian.net/browse/WOR-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
@@ -92,16 +92,17 @@ class BPMBillingProjectDeleteRunner(
   ): Future[JobStatus] =
     Try(workspaceManagerDAO.getDeleteLandingZoneResult(jobId.toString, lzId, ctx)) match {
       case Failure(e: ApiException) =>
-        val (msg, jobStatus) = e.getCode match {
-          case 500 => (s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}", Incomplete)
-          case 404 => (s"Unable to find jobId $jobId in WSM for Landing Zone deletion", Complete)
+        e.getCode match {
+          case 500 =>
+            val msg = s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}"
+            billingRepository.updateCreationStatus(projectName, Deleting, Some(msg)).map(_ => Incomplete)
+          case 404 =>
+            val msg = s"Unable to find jobId $jobId in WSM for Landing Zone deletion"
+            billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Complete)
           case code =>
-            (
-              s"API call to get Landing Zone deletion operation failed with status code $code: ${e.getMessage}",
-              Incomplete
-            )
+            val msg = s"API call to get Landing Zone deletion operation failed with status code $code: ${e.getMessage}"
+            billingRepository.updateCreationStatus(projectName, Deleting, Some(msg)).map(_ => Incomplete)
         }
-        billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => jobStatus)
       case Failure(e) =>
         val msg = s"Api call to get landing zone delete job $jobId from workspace manager failed: ${e.getMessage}"
         billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Incomplete)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -92,16 +92,17 @@ class BPMBillingProjectDeleteRunner(
   ): Future[JobStatus] =
     Try(workspaceManagerDAO.getDeleteLandingZoneResult(jobId.toString, lzId, ctx)) match {
       case Failure(e: ApiException) =>
-        val (msg, jobStatus) = e.getCode match {
-          case 500 => (s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}", Incomplete)
-          case 404 => (s"Unable to find jobId $jobId in WSM for Landing Zone deletion", Complete)
+        e.getCode match {
+          case 500 =>
+            val msg = s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}"
```
cahrens: The 500 case is now basically the same as the one below at line 103. I'm not sure if there is value in handling the 500 separately. I think that including the jobId and the status code in the message is probably sufficient.

blakery: Makes sense to me
In the first pass, I was using the generic jobs endpoint, where all of this gets hit a lot more frequently, because the generic job endpoint returns its error reports as non-2xx
Since this is using the job-specific endpoint, 'normal' errors are returned with 200 and an error report, so there's a lot less value in handling specific error cases now

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -92,16 +92,14 @@ class BPMBillingProjectDeleteRunner(
...
     Try(workspaceManagerDAO.getDeleteLandingZoneResult(jobId.toString, lzId, ctx)) match {
       case Failure(e: ApiException) =>
-        val (msg, jobStatus) = e.getCode match {
-          case 500 => (s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}", Incomplete)
-          case 404 => (s"Unable to find jobId $jobId in WSM for Landing Zone deletion", Complete)
+        e.getCode match {
+          case 404 =>
+            val msg = s"Unable to find jobId $jobId in WSM for Landing Zone deletion"
+            billingRepository.updateCreationStatus(projectName, DeletionFailed, Some(msg)).map(_ => Complete)
           case code =>
```
marctalbott: I'm not sure retrying a 403 will change anything. 400s and 401s also may not be worth retrying

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/BPMBillingProjectDeleteRunner.scala
```diff
@@ -92,16 +92,13 @@ class BPMBillingProjectDeleteRunner(
...
-          case 500 => (s"Landing Zone deletion operation with jobId $jobId failed: ${e.getMessage}", Incomplete)
-          case 404 => (s"Unable to find jobId $jobId in WSM for Landing Zone deletion", Complete)
-          case code =>
-            (
-              s"API call to get Landing Zone deletion operation failed with status code $code: ${e.getMessage}",
-              Incomplete
-            )
+        val msg = s"API call to get Landing Zone deletion operation failed with status code ${e.getCode}: ${e.getMessage}"
+        e.getCode match {
+          case code if code > 400 && code < 500 =>
```
cahrens: Did you intentionally leave 400 out of this?

blakery: Kind of. I was thinking of 401, 403, and 404 as the important ones to catch (those are the only non-500 exceptions I've seen from any of the async job endpoints, and I didn't see any others in the service code). 

So it wasn't exactly intentional, but it also wasn't exactly a logic error.

I can make it inclusive of 400; it's only the 500 and 404 that I feel strongly about personally, since those are what I've actually encountered with these endpoints (also why I have explicit tests for them).
It _does_ look like it's an off-by-one error like that.

cahrens: I think it makes sense to include 400 as not retryable-- 

```
(https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)
The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

cahrens (approved):

marctalbott (approved):

PR #2266, opened 2023-03-17
WOR-779 update rawls spend report api to call bpm for azure billing projects

sergiygetlin:
Spend report endpoint now can return Azure spend data.

Ticket: https://broadworkbench.atlassian.net/browse/WOR-779

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -65,7 +65,6 @@ trait BillingProfileManagerDAO {
 
   def getStatus(): SystemStatus
 
-  @throws(classOf[BpmAzureSpendReportBadRequest])
   @throws(classOf[BpmAzureSpendReportApiException])
   def getAzureSpendReport(billingProfileId: UUID,
                           spendReportStartDate: Date,
@@ -76,8 +75,8 @@ trait BillingProfileManagerDAO {
 
 class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
 
-class BpmAzureSpendReportBadRequest(message: String) extends Exception(message)
-class BpmAzureSpendReportApiException(message: String, cause: Throwable = null) extends Exception(message, cause)
+class BpmAzureSpendReportApiException(val statusCode: Int, message: String, cause: Throwable = null)
+    extends Exception(message, cause)
 
 object BillingProfileManagerDAO {
   val BillingProfileRequestBatchSize = 1000
@@ -197,7 +196,6 @@ class BillingProfileManagerDAOImpl(
 
   override def getStatus(): SystemStatus = apiClientProvider.getUnauthenticatedApi().serviceStatus()
 
-  @throws(classOf[BpmAzureSpendReportBadRequest])
   @throws(classOf[BpmAzureSpendReportApiException])
   def getAzureSpendReport(billingProfileId: UUID,
                           spendReportStartDate: Date,
@@ -214,13 +212,9 @@ class BillingProfileManagerDAOImpl(
         )
     catch {
       case ex: ApiException =>
-        ex.getCode match {
-          case StatusCodes.BadRequest.intValue =>
-            val bpmErrorMessageJson = ex.getMessage.parseJson
-            val bpmErrorMessage = bpmErrorMessageJson.convertTo[BpmAzureReportErrorMessage]
-            throw new BpmAzureSpendReportBadRequest(bpmErrorMessage.message)
-          case _ => throw new BpmAzureSpendReportApiException(ex.getMessage, ex)
-        }
-      case ex: Exception => throw new BpmAzureSpendReportApiException(ex.getMessage, ex)
+        logger.info(s"Failed to get Azure spend report for billing profile [id=${billingProfileId}]")
+        val bpmErrorMessageJson = ex.getMessage.parseJson
+        val bpmErrorMessage = bpmErrorMessageJson.convertTo[BpmAzureReportErrorMessage]
+        throw new BpmAzureSpendReportApiException(ex.getCode, bpmErrorMessage.message, ex)
     }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -214,12 +214,12 @@ class BillingProfileManagerDAOImpl(
         )
     catch {
       case ex: ApiException =>
-        if (ex.getCode == StatusCodes.BadRequest.intValue) {
-          val bpmErrorMessageJson = ex.getMessage.parseJson
-          val bpmErrorMessage = bpmErrorMessageJson.convertTo[BpmAzureReportErrorMessage]
-          throw new BpmAzureSpendReportBadRequest(bpmErrorMessage.message)
-        } else {
-          throw new BpmAzureSpendReportApiException(ex.getMessage, ex)
+        ex.getCode match {
+          case StatusCodes.BadRequest.intValue =>
+            val bpmErrorMessageJson = ex.getMessage.parseJson
+            val bpmErrorMessage = bpmErrorMessageJson.convertTo[BpmAzureReportErrorMessage]
+            throw new BpmAzureSpendReportBadRequest(bpmErrorMessage.message)
+          case _ => throw new BpmAzureSpendReportApiException(ex.getMessage, ex)
         }
       case ex: Exception => throw new BpmAzureSpendReportApiException(ex.getMessage, ex)
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -14,7 +14,7 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsRequestContext
 }
 
-import java.util.UUID
+import java.util.{Date, UUID}
 import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
@@ -52,6 +52,12 @@ trait BillingProfileManagerDAO {
   ): Unit
 
   def getStatus(): SystemStatus
+
+  def getAzureSpendReport(billingProfileId: UUID,
+                          spendReportStartDate: Date,
+                          spendReportEndDate: Date,
+                          ctx: RawlsRequestContext
+  )(implicit ec: ExecutionContext): Future[SpendReport]
 }
 
 class ManagedAppNotFoundException(errorReport: ErrorReport) extends RawlsExceptionWithErrorReport(errorReport)
@@ -173,4 +179,19 @@ class BillingProfileManagerDAOImpl(
       )
 
   override def getStatus(): SystemStatus = apiClientProvider.getUnauthenticatedApi().serviceStatus()
+
+  def getAzureSpendReport(billingProfileId: UUID,
+                          spendReportStartDate: Date,
+                          spendReportEndDate: Date,
+                          ctx: RawlsRequestContext
+  )(implicit ec: ExecutionContext): Future[SpendReport] =
+    Future.apply(
+      apiClientProvider
+        .getSpendReportingApi(ctx)
+        .getSpendReport(
+          billingProfileId,
+          spendReportStartDate,
+          spendReportEndDate
+        )
+    )
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -15,16 +15,7 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsBillingAccountName,
   RawlsRequestContext
 }
-import spray.json.{
-  DefaultJsonProtocol,
-  DeserializationException,
-  JsArray,
-  JsNumber,
-  JsObject,
-  JsString,
-  JsValue,
-  RootJsonFormat
-}
+import spray.json.DefaultJsonProtocol
 
 import java.util.{Date, UUID}
 import scala.annotation.tailrec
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -22,6 +25,53 @@ case class SpendReportingAggregationKeyWithSub(key: SpendReportingAggregationKey
...
+    def mapSpendReportingForDateRange(
+      spendReportingForDateRange: bio.terra.profile.model.SpendReportingForDateRange
+    ): SpendReportingForDateRange =
+      SpendReportingForDateRange(
+        spendReportingForDateRange.getCost,
+        spendReportingForDateRange.getCredits,
+        spendReportingForDateRange.getCurrency,
+        Option.empty,
+        Option.empty,
+        category = Option.apply(TerraSpendCategories.withName(spendReportingForDateRange.getCategory.toString))
```
marctalbott: You can actually omit the `.apply` entirely (one of the syntactic sugar/magic elements of the `.apply` method). `Option.apply(...) == Option(...)`

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -22,6 +25,53 @@ case class SpendReportingAggregationKeyWithSub(key: SpendReportingAggregationKey
...
+      spendReport.getSpendSummary.getCredits,
+      spendReport.getSpendSummary.getCurrency,
+      Option.apply(DateTime.parse(spendReport.getSpendSummary.getStartTime)),
+      Option.apply(DateTime.parse(spendReport.getSpendSummary.getEndTime))
+    )
+
+    SpendReportingResults.apply(spendDetails, spendSummary)
+  }
+}
+
```
marctalbott: Curious what @blakery thinks, but it might be worth making these convertors `apply` methods in the respective case classes. Something like this could be a little simpler

```suggestion
case class SpendReportingResults(spendDetails: Seq[SpendReportingAggregation], spendSummary: SpendReportingForDateRange) {
  def apply(spendReport: bio.terra.profile.model.SpendReport): SpendReportingResults = {
    val spendDetails: Seq[SpendReportingAggregation] =
      spendReport.getSpendDetails.asScala
        .map(SpendReportingAggregation)
        .toList

    val spendSummary = SpendReportingForDateRange(
      spendReport.getSpendSummary.getCost,
      spendReport.getSpendSummary.getCredits,
      spendReport.getSpendSummary.getCurrency,
      Option(DateTime.parse(spendReport.getSpendSummary.getStartTime)),
      Option(DateTime.parse(spendReport.getSpendSummary.getEndTime))
    )

    SpendReportingResults(spendDetails, spendSummary)
  }
  
  case class SpendReportingForDateRange(
    cost: String,
    credits: String,
    currency: String,
    startTime: Option[DateTime] = None,
    endTime: Option[DateTime] = None,
    workspace: Option[WorkspaceName] = None,
    googleProjectId: Option[GoogleProject] = None,
    category: Option[TerraSpendCategory] = None,
    subAggregation: Option[SpendReportingAggregation] = None
  ) {
      def apply(
      spendReportingForDateRange: bio.terra.profile.model.SpendReportingForDateRange
    ): SpendReportingForDateRange =
      SpendReportingForDateRange(
        spendReportingForDateRange.getCost,
        spendReportingForDateRange.getCredits,
        spendReportingForDateRange.getCurrency,
        Option.empty,
        Option.empty,
        category = Option(TerraSpendCategories.withName(spendReportingForDateRange.getCategory.toString))
      )
    }

case class SpendReportingAggregation(aggregationKey: SpendReportingAggregationKey,
                                     spendData: Seq[SpendReportingForDateRange]
) {
    def apply(
      spendReportingAggregation: bio.terra.profile.model.SpendReportingAggregation
    ): SpendReportingAggregation = {

      val spendData = spendReportingAggregation.getSpendData.asScala
        .map(SpendReportingForDateRange)
        .toList

      SpendReportingAggregation(
        SpendReportingAggregationKeys.withName(spendReportingAggregation.getAggregationKey.name()),
        spendData
      )
    }
  }

```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -288,4 +304,44 @@ class SpendReportingService(
...
+    end: DateTime,
+    aggregations: Set[SpendReportingAggregationKeyWithSub]
+  ): Future[SpendReportingResults] =
+    for {
+      billingProject <- billingRepository.getBillingProject(project)
+
+      report <- getReportData(billingProject.get, project, start, end, aggregations)
+
+      result = report
+    } yield result
```
marctalbott: We can return `report` directly

```suggestion
    } yield report
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -173,4 +194,30 @@ class BillingProfileManagerDAOImpl(
...
+      apiClientProvider
+        .getSpendReportingApi(ctx)
+        .getSpendReport(
+          billingProfileId,
+          spendReportStartDate,
+          spendReportEndDate
+        )
+    catch {
+      case ex: ApiException =>
+        if (ex.getCode == StatusCodes.BadRequest.intValue) {
```
marctalbott: It could be good to make sure all BPM exceptions are getting propagated up to the user correctly, not just 400s.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
sergiygetlin: Agree. This will be more idiomatic solution in this case. I initially start using this approach per @blakery recommendation, but by some reason it didn't work. Most likely I did something wrong. So, I unblocked myself and used another approach.

sergiygetlin: Apply should belong to the corresponding companion object if I am not mistaken

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
```diff
@@ -217,4 +223,121 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar {
...
+  behavior of "BpmAzureReportErrorMessageJsonProtocol"
+
+  it should "serialize BpmAzureReportErrorMessage into json" in {
+    val bpmError = BpmAzureReportErrorMessage("customError", 400)
+    val jsObject: JsValue = bpmError.toJson
+
+    jsObject shouldNot equal(null)
+
+    // jsObject.fields should contain key "message"
+    // jsObject.fields should contain key "statusCode"
```
cahrens: Do you want to actually assert these?

cahrens: Or perhaps not because of the next test case…

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: Why is BadRequest handled in a special way? A comment about the use case could be helpful.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -288,4 +304,43 @@ class SpendReportingService(
...
+    end: DateTime
+  ): Future[SpendReportingResults] =
+    try {
+      val spendReport =
+        bpmDao.getAzureSpendReport(UUID.fromString(billingProfileId), start.toDate, end.toDate, ctx)
+
+      Future.successful(SpendReportingResultsConvertor(spendReport))
+    } catch {
+      case ex: BpmAzureSpendReportBadRequest =>
+        throw RawlsExceptionWithErrorReport(StatusCodes.BadRequest, ex.getMessage)
```
cahrens: @blakery Relating to our conversation at the end of standup… would it be better for this to return Future.failed(RawlsExceptionWithErrorReport..)? I'm unclear now about when it is important to return a failed future vs. just throwing the exception.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -288,4 +304,43 @@ class SpendReportingService(
...
+      getSpendForGCPBillingProject(project, start, end, aggregations)
+    } else {
+      getSpendForAzureBillingProject(billingProject.billingProfileId.get, start, end)
+    }
+
+  private def getSpendForAzureBillingProject(
+    billingProfileId: String,
+    start: DateTime,
+    end: DateTime
+  ): Future[SpendReportingResults] =
```
cahrens, line 335: @blakery  and I chatted. This could just return `SpendReportingResults`, and the call could be wrapped with a Future where it's actually called (line 328).

As the method is currently written (with return type of Future[SpendReportingResults], it would be better to return Future.failed(exception) than to throw the exception outside of a Future.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
sergiygetlin: To parse response from BPM and return validation message only

cahrens: And that is the only case where the response is JSON?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
```diff
@@ -25,56 +24,71 @@ case class SpendReportingAggregationKeyWithSub(key: SpendReportingAggregationKey
...
+      .map(sd =>
+        SpendReportingAggregation(
+          aggregationKey = SpendReportingAggregationKeys.withName(sd.getAggregationKey.name()),
+          spendData = sd.getSpendData.asScala
+            .map(srRange =>
+              SpendReportingForDateRange(
+                srRange.getCost,
+                srRange.getCredits,
+                srRange.getCurrency,
+                Option.when(srRange.getStartTime != null)(DateTime.parse(srRange.getStartTime)),
```
marctalbott: `Option(srRange.getStartTime).map(DateTime.parse)` would accomplish the same thing and is a bit simpler IMO

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
sergiygetlin: This is the only case where I need to extract message, otherwise I return ErrorReport with the whole exception

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
sergiygetlin: I think I don't need this test at all.

cahrens (approved):
This LGTM, but I obviously am no Scala expert. I'm trusting that @marctalbott and/or @blakery will review for that.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -173,4 +196,31 @@ class BillingProfileManagerDAOImpl(
...
+        .getSpendReportingApi(ctx)
+        .getSpendReport(
+          billingProfileId,
+          spendReportStartDate,
+          spendReportEndDate
+        )
+    catch {
+      case ex: ApiException =>
+        ex.getCode match {
+          case StatusCodes.BadRequest.intValue =>
```
marctalbott: If we're going to parse out the error message from BPM like this for 400s, I think it would be worth doing so for all errors from BPM so that we can return the message without the extra JSON.

marctalbott: This also wraps everything into these exceptions here, only to unwrap and rewrap them into `RawlsExceptionWithErrorReport` in the `SpendReportingService`. I'd prefer to just wrap it up in the exception we want to throw here and simplify the exception handling in `SpendReportingService`.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala
sergiygetlin: I somehow missed this option

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
sergiygetlin: I simplified exception handling at DAO layer a little bit, but still wrapping original exception into DAO specific one. Left service layer as is.

core/src/test/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModelSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+  object TestData {
+    val defaultCurrency = "USD"
+    val defaultAzureCredits = "0" // always 0 for Azure
+
+    def buildBPMEmptyReport: SpendReport = {
+      // some period
+      val from = DateTime.now().minusMonths(2)
+      val to = from.plusMonths(1)
+
+      // BVM return following spendReport in case specific data doesn't exist
```
marctalbott, line 25: For GCP, Rawls returns a 404 when there's no data. I don't really have a preference between a 404 or returning an empty report like BPM does today, but they should at least be consistent. What do you think?

Making them consistent falls outside the scope of this PR in my opinion, but wanted to note the inconsistency

marctalbott (approved):
Looks good to me! Wanted to get your thoughts on one inconsistency between BPM and Rawls, but I think it's out of scope for this PR and can be addressed in a follow-up PR

core/src/test/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModelSpec.scala
sergiygetlin: Yes, I agree. I think it would be definitely better to keep responses consistent. It probably requires changes not only in Rawls, but in BPM too.

PR #2277, opened 2023-03-21
WOR-837 - Make Terra resilient if a landing zone fails to create

blakery:
Ticket: [WOR-837](https://broadworkbench.atlassian.net/browse/WOR-837)

The main change was to proceed with deletion immediately if there is no landing zone to delete. 

I was able to delete a billing project where retrieving the landing zone delete results had failed, after manually deleting the landing zone id on the billing project.

---

**PR checklist**

[WOR-837]: https://broadworkbench.atlassian.net/browse/WOR-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -9,10 +9,7 @@ import io.sentry.{Sentry, SentryEvent}
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, WorkspaceManagerResourceMonitorRecordDao}
 import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord
-import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord.JobType.{
-  BpmBillingProjectDelete,
-  GoogleBillingProjectDelete
-}
+
 import org.broadinstitute.dsde.rawls.model.{
   CreateRawlsV2BillingProjectFullRequest,
   CreationStatuses,
@@ -189,9 +186,9 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
             )
         }
       billingProfileId <- billingRepository.getBillingProfileId(projectName)
-      (deleteType, projectLifecycle) = billingProfileId match {
-        case None    => (GoogleBillingProjectDelete, googleBillingProjectLifecycle)
-        case Some(_) => (BpmBillingProjectDelete, bpmBillingProjectLifecycle)
+      projectLifecycle = billingProfileId match {
+        case None    => googleBillingProjectLifecycle
+        case Some(_) => bpmBillingProjectLifecycle
       }
       _ <- billingRepository.failUnlessHasNoWorkspaces(projectName)
       _ <- billingRepository.getCreationStatus(projectName).map { status =>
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -189,9 +186,9 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
             )
         }
       billingProfileId <- billingRepository.getBillingProfileId(projectName)
-      (deleteType, projectLifecycle) = billingProfileId match {
-        case None    => (GoogleBillingProjectDelete, googleBillingProjectLifecycle)
-        case Some(_) => (BpmBillingProjectDelete, bpmBillingProjectLifecycle)
+      projectLifecycle = billingProfileId match {
```
blakery, line 188: I realized I could make the behavior more consistent by using the return value of `BillingProjectLifecycle.initiateDelete` to indicate whether or not the deletion has to wait on a job in order to finish.
So if we get `Some(jobId)`: schedule the job. If we get `None`: finalize the delete immediately.

This eliminates the need for a dummy job id, and makes the logic simpler.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -37,6 +34,8 @@ class GoogleBillingProjectLifecycle(
 ) extends BillingProjectLifecycle {
   implicit val errorReportSource: ErrorReportSource = ErrorReportSource("rawls")
 
+  override val deleteJobType: JobType = GoogleBillingProjectDelete
```
blakery, line 37: making this part of the `BillingProjectLifecycle` makes the orchestrator more open-ended and generic

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -596,7 +596,7 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
 
   behavior of "initiateDelete"
 
-  it should "succeed if the landing zone and billing profiles id do not exist" in {
+  it should "return None for the jobId if the landing zone does not exist" in {
```
blakery, line 599: The old test title was a lie.
A separate test for the missing bp id was added below - although that circumstance shouldn't really come up with how things are handled currently.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -611,7 +611,9 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
                                      mock[WorkspaceManagerResourceMonitorRecordDao]
       )
 
-    Await.result(bp.initiateDelete(billingProjectName, testContext), Duration.Inf)
+    val jobId = Await.result(bp.initiateDelete(billingProjectName, testContext), Duration.Inf)
+
+    assert(jobId.isEmpty)
 
     verify(workspaceManagerDAO, Mockito.never()).deleteLandingZone(ArgumentMatchers.any(), ArgumentMatchers.any())
     verify(bpm, Mockito.never()).deleteBillingProfile(ArgumentMatchers.any[UUID], ArgumentMatchers.eq(testContext))
```
cahrens, line 619: Don't you also want to verify that the billing project was actually deleted (`verify(repo).deleteBillingProject(ArgumentMatchers.eq(billingProjectName`)? That's the change in behavior isn't it-- before the billing project wouldn't be deleted, but now it is.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
cahrens: Is the variable actually needed though? I don't see that anything besides a test references it.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
blakery: No, because there's no change in behavior in `BpmBillingProjectLifecycle`. It's all in the orchestrator.
I only updated the tests here because I saw a deficiency.

I did add a test to explicitly hit the code path of the behavior in the orchestrator when `BpmBillingProjectLifecycle.initiateDelete` returns `None` though. It was being tested before, but kind of piecemeal.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
blakery: The orchestrator uses it to determine the type of job it needs to schedule, instead of hard-coding the logic there.
This makes it so all of the specific logic, outside of determining which lifecycle to use, is all contained here.

cahrens: Right, I see it now.

cahrens (approved):

marctalbott (approved):

PR #2278, opened 2023-03-22
[WOR-816] Avoid throwing 500 errors from checkWorkspaceCloudPermissions for requester pays and workspaces with bad billing

cahrens:
Both of the specific errors (requester pays and bad billing) mentioned in https://broadworkbench.atlassian.net/browse/WOR-816 are cases of `StorageExceptions` being thrown from `testSAGoogleBucketIam`. 

There was some discussion of possibly avoiding the error on a RequestPays workspaces by using a different account to make the query, but that will not work because the user does not have write access (which Nick thought of later in the Slack thread discussion). And given that we want to handle the bad billing case as well, I think it makes more sense just to ensure that we retain the original 4xx status code, which ensures that the errors won't propagate to Sentry-- `RawlsExceptionWithErrorReport` instances are only [sent to Sentry](https://github.com/broadinstitute/rawls/blob/93afe2d5401da2b026a8359f612814379f5cd5aa/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala#L41) if they have 5xx status codes. 

Before on a requester pays workspace:
image
image

After on same workspace:
image
image

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -7,6 +7,7 @@ import bio.terra.workspace.model.WorkspaceDescription
 import cats.implicits._
 import cats.{Applicative, ApplicativeThrow, MonadThrow}
 import com.google.api.services.cloudbilling.model.ProjectBillingInfo
+import com.google.cloud.storage.StorageException
 import com.typesafe.scalalogging.LazyLogging
 import io.opencensus.scala.Tracing.startSpanWithParent
 import io.opencensus.trace.{AttributeValue => OpenCensusAttributeValue, Span, Status}
@@ -50,6 +51,7 @@ import org.joda.time.DateTime
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
+import java.io.IOException
 import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
@@ -2736,7 +2738,12 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         GcsBucketName(workspace.bucketName),
         petKey,
         expectedGoogleBucketPermissions
-      )
+      ).recoverWith { case t: StorageException =>
+         // Throw with the status code of the exception (for example 403 for invalid billing, 400 for requester pays)
+         // instead of a 500 to avoid Sentry notifications.
+         Future.failed(new RawlsExceptionWithErrorReport(ErrorReport(t.getCode, t)))
+      }
+
       _ <- ApplicativeThrow[Future].raiseWhen(useDefaultPet && expectedGoogleProjectPermissions.nonEmpty) {
         new RawlsException("user has workspace read-only access yet has expected google project permissions")
       }
@@ -2744,7 +2751,10 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       projectIamResults <- gcsDAO.testSAGoogleProjectIam(GoogleProject(workspace.googleProjectId.value),
                                                          petKey,
                                                          expectedGoogleProjectPermissions
-      )
+      ).recoverWith { case t: IOException =>
+        // Throw a 400 to avoid Sentry notifications.
+        Future.failed(new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, t)))
+      }
 
       missingBucketPermissions = expectedGoogleBucketPermissions -- bucketIamResults
       missingProjectPermissions = expectedGoogleProjectPermissions -- projectIamResults
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -2734,27 +2734,31 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         _.filterNot(_.value.startsWith("resourcemanager."))
       )
 
-      bucketIamResults <- gcsDAO.testSAGoogleBucketIam(
-        GcsBucketName(workspace.bucketName),
-        petKey,
-        expectedGoogleBucketPermissions
-      ).recoverWith { case t: StorageException =>
-         // Throw with the status code of the exception (for example 403 for invalid billing, 400 for requester pays)
-         // instead of a 500 to avoid Sentry notifications.
-         Future.failed(new RawlsExceptionWithErrorReport(ErrorReport(t.getCode, t)))
-      }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2732,19 +2734,31 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-      projectIamResults <- gcsDAO.testSAGoogleProjectIam(GoogleProject(workspace.googleProjectId.value),
-                                                         petKey,
-                                                         expectedGoogleProjectPermissions
-      )
+      projectIamResults <- gcsDAO
+        .testSAGoogleProjectIam(GoogleProject(workspace.googleProjectId.value),
+                                petKey,
+                                expectedGoogleProjectPermissions
+        )
+        .recoverWith { case t: IOException =>
```
cahrens, line 2758: We don't know of a case where this throws an exception and causes a Sentry alert, but I figured it was worth doing the same sort of handling. Unfortunately though we can't get at the original status code. This is where I found that the method is declared to throw IOException: https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/java/latest/com/google/api/services/cloudresourcemanager/CloudResourceManager.Organizations.html

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2732,19 +2734,31 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-        petKey,
-        expectedGoogleBucketPermissions
-      )
+      bucketIamResults <- gcsDAO
+        .testSAGoogleBucketIam(
+          GcsBucketName(workspace.bucketName),
+          petKey,
+          expectedGoogleBucketPermissions
+        )
+        .recoverWith { case t: StorageException =>
```
cahrens, line 2743: The underlying Google method `testIAMPermissions` only declares that it throws StorageException.

https://javadoc.io/static/com.google.cloud/google-cloud-storage/1.113.13/com/google/cloud/storage/spi/v1/StorageRpc.html

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2732,19 +2734,31 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+          // Throw with the status code of the exception (for example 403 for invalid billing, 400 for requester pays)
+          // instead of a 500 to avoid Sentry notifications.
+          Future.failed(new RawlsExceptionWithErrorReport(ErrorReport(t.getCode, t)))
```
aherbst-broad, line 2746: Should this be t.getMessage rather than the exception itself?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: I'm confused about how this is a CloudResourceManager. When I drill into the code I get this rawls [method](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L690), then this wb-libs [method](https://github.com/broadinstitute/workbench-libs/blob/1a66428276165d7e9a2a1e8819474ef6029b72c2/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala#L554), and then finally this google sdk [method](https://github.com/googleapis/java-storage/blob/46e6a51693d1eb5ecf7c88859cdfa859dc4d5e51/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L4015), which looks like it throws `StorageException` once again.

aherbst-broad (commented):
Had a question about the IOException vs StorageException

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: You get more information if you pass the throwable itself. Do you have concerns about that information being included?

```
  def apply(statusCode: StatusCode, throwable: Throwable)(implicit source: ErrorReportSource): ErrorReport =
    ErrorReport(source.source,
                message(throwable),
                Some(statusCode),
                causes(throwable),
                throwable.getStackTrace,
                Option(throwable.getClass)
    )

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala (hunk 1 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala:2758]: Let me take a look again.

Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala:2758]: @aherbst-broad I believe you are in the wrong method. This one calls this rawls [method](https://github.com/broadinstitute/rawls/blob/45f8e07012299a1670ac84e8f144a4302a56c71b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L696), which calls `cloudResourceManager.projects().testIamPermissions(project.value, testRequest)`.

```
    public TestIamPermissions testIamPermissions(java.lang.String resource, com.google.api.services.cloudresourcemanager.model.TestIamPermissionsRequest content) throws java.io.IOException {
      TestIamPermissions result = new TestIamPermissions(resource, content);
      initialize(result);
      return result;
    }

aherbst-broad: Ah that makes sense, thanks for the explanation.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: Meh, ok. I have misgivings about our continual surfacing of stack traces in the UI, but that's a different discussion.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: This won't end up showing a stack trace to the user. If this cloud permissions check fails we just continue to show the banner about the syncing permissions, unless the failure specifically has the requester pays error message in it (in which case we just treat the workspace as if its permissions are done syncing).

PR #2279, opened 2023-03-22
ID-492 Fastpass Feature Branch

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-492
Design doc: https://docs.google.com/document/d/1PimgphrgXwaB2NRzy4uhyJKT13DZ4Yr2CfEZmcts1qY/edit

Operation FastPass

---

**PR checklist**


tlangs:
jenkins retest

tlangs:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/test/api/RawlsApiSpec.scala
```diff
@@ -1134,7 +1134,7 @@ class RawlsApiSpec
   }
 
   // Retrieves roles with policy emails for bucket acls and checks that service account is set up correctly
-  private def getBucketRolesWithEmails(
+  private def getNonConditionalBucketRolesWithEmails(
```
cahrens, line 1140: It might help to update the documentation to explain what "non-conditional" refers to.

core/src/main/resources/reference.conf
```diff
@@ -128,4 +128,8 @@ dataRepoEntityProvider {
   bigQueryMaximumBytesBilled = 100GB
 }
 
-sam.timeout = 3 minutes
\ No newline at end of file
+sam.timeout = 3 minutes
+
+prometheus {
```
cahrens, line 133: Was this change intentional? Ah, I see some openTelemetry stuff below. Can you explain how it's being used in the PR?

core/src/main/scala/org/broadinstitute/dsde/rawls/config/FastPassConfig.scala
```diff
@@ -0,0 +1,21 @@
...
+import java.time.Duration
+
+case class FastPassConfig(
+  enabled: Boolean,
+  grantPeriod: Duration
+)
+
+object FastPassConfig {
+  def apply(conf: Config): FastPassConfig = {
+    val fastPassConfig = conf.getConfig("fastPass")
```
cahrens, line 15: Can you link the firecould-develop PR that adds these values to the rawls conf file?

core/src/main/resources/reference.conf
cahrens: Should the fastPass config also be added to this file?

tlangs: FastPass is configured per-env, but prometheus/opentelemetry is the same for all envs.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/FastPassGrantComponent.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  import driver.api._
+
+  class FastPassGrantTable(tag: Tag) extends Table[FastPassGrantRecord](tag, "FASTPASS_GRANTS") {
+
+    def id = column[Long]("id", O.PrimaryKey, O.AutoInc)
+
+    def workspaceId = column[UUID]("workspace_id")
+
+    def userSubjectId = column[String]("user_subject_id", O.PrimaryKey)
```
cahrens: I'm not entirely sure what "userSubject" is, but will it be unique if the same user has FastPass grants in 2 different workspaces (wondering about the primary key designation)?

tlangs: `userSubjectId` is the unique ID used by Sam for a user in Terra. Its used here to tie all the Pet Service Accounts of a user together so that they're removed when a user is removed.

core/src/main/scala/org/broadinstitute/dsde/rawls/config/FastPassConfig.scala
tlangs: https://github.com/broadinstitute/firecloud-develop/pull/3267

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/FastPassGrantComponent.scala
cahrens: I probably just am not understanding the syntax… but why is it marked as a primary key?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
```diff
@@ -917,11 +918,7 @@ object WorkspaceMigrationActor {
 
           _ <- rolesToRemove.traverse_ { case (identity, roles) =>
             val Array(memberType, email) = identity.toString.split(":")
-            googleIamDao.removeIamRoles(googleProject,
-                                        WorkbenchEmail(email),
-                                        GoogleIamDAO.MemberType.stringToMemberType(memberType),
-                                        roles
-            )
+            googleIamDao.removeRoles(googleProject, WorkbenchEmail(email), IamMemberTypes.withName(memberType), roles)
```
cahrens, line 921: Did the method change names?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1164,7 +1174,10 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                 _ <- methodConfig.traverse_(dataAccess.methodConfigurationQuery.create(destWorkspaceContext, _))
               } yield ()
             })
-
+            _ <- traceDBIOWithParent("FastPassService.setupFastPassClonedWorkspace", parentContext)(childContext =>
```
cahrens, line 1178: Do we need to worry about the comment at line 1100-- "Orchestration has its own implementation of cloneWorkspace. When changing something here, you may also need to update orchestration's implementation"?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,417 @@
...
+      val ((accountEmail, accountType, resourceType, resourceName), grants) = grouped
+      val organizationRoles = grants.map(_.organizationRole).toSet
+      val workbenchEmail = WorkbenchEmail(accountEmail.value)
+      val googleProject = GoogleProject(googleProjectId.value)
+      () =>
+        removeFastPassGrants(resourceType, resourceName, workbenchEmail, accountType, organizationRoles, googleProject)
+    }
+    for {
+      // this `flatMap` is necessary to run the IAM Updates in sequence, as to not sent conflicting policy updates to GCP
+      _ <- DBIO.from(iamUpdates.foldLeft(Future.successful())((a, b) => a.flatMap(_ => b())))
```
dvoet, line 228: perhaps a simplification for a later date: seems like this could be done with 2 total calls to google, 1 to update project iam and another to update bucket iam. Then you don't need to worry about serializing correctly

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,417 @@
...
+        .map(fromBucketPolicy)
+    } yield {
+      // Role binding quotas do not de-duplicate member emails, hence the conversion of Sets to Lists
+      val numProjectRoleBindings = projectPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedProjectBindings = numProjectRoleBindings + maxProjectRoleBindingsPerUser
+
+      val numBucketRoleBindings = bucketPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedBucketBindings = numBucketRoleBindings + maxBucketRoleBindingsPerUser
+
+      expectedProjectBindings < policyBindingsQuotaLimit && expectedBucketBindings < policyBindingsQuotaLimit
```
dvoet, line 430: I would log the numbers in event quotas are exceeded

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+    for {
+      _ <- DBIO.from(addUserAndPetToProjectIamRoles(workspace.googleProjectId, projectIamRoles, userAndPet, condition))
+      _ <- writeGrantsToDb(
+        workspace.workspaceId,
+        userAndPet,
+        gcpResourceType = IamResourceTypes.Project,
+        workspace.googleProjectId.value,
+        projectIamRoles,
+        expirationDate
+      )
```
dvoet: example of where mixing `DBIO` and `Future` is a little touchy. The roles have been added to google yet future errors may trigger a rollback and the records will not be written to the db.

dvoet (approved):
I don't see any red flags so I approve given the urgency.

I know other places in rawls mix `Future` and `DBIO` but it is generally a bad practice because it may hold transactions open longer than necessary and the order of operations may not be what you think it is (e.g. `Future`s execute before the transaction commits and perhaps regardless of eventual transaction disposition). I flagged one case where this is especially touchy.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      terraBucketWriterRole
+    )
+
+  val policyBindingsQuotaLimit = 1500
+  val maxProjectRoleBindingsPerUser = 6 // 3 possible roles for each user and pet service account
+  val maxBucketRoleBindingsPerUser = 2 // 1 possible role for each user and pet service account
+
+}
+
+class FastPassService(protected val ctx: RawlsRequestContext,
```
dvoet, line 75: add some doc about what FastPass is. It is a nice name but is not immediately apparent what it means to someone not in the know.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/WorkspaceMigrationActor.scala
tlangs: I had to deprecate the old methods, and because of a quirk in scala compiling, overloaded methods are not allowed to have different sets of default arguments. https://stackoverflow.com/questions/66562353/scala-multiple-overloaded-alternatives-of-method-define-default-arguments.

So, in order to not break any existing usages of `addIamRoles` and `removeIamRoles`, I had to create a new method.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/FastPassGrantComponent.scala
tlangs: Later additions to the table have made it so that `user_subject_id` shouldn't be a primary key. I would recommend editing the liquibase changeset in this PR instead of making another changeset. It just makes the merged DB changes cleaner.

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
```diff
@@ -171,6 +171,10 @@ class MockSamDAO(dataSource: SlickDataSource)(implicit executionContext: Executi
 
   override def getUserArbitraryPetServiceAccountKey(userEmail: String): Future[String] = ???
 
+  override def getUserPetServiceAccount(ctx: RawlsRequestContext,
+                                        googleProjectId: GoogleProjectId
+  ): Future[WorkbenchEmail] =
+    Future.successful(WorkbenchEmail("pet-110347448408766049948@broad-dsde-dev.iam.gserviceaccount.com"))
```
cahrens, line 177: Is the specific email address meaningful in any way?

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -278,13 +303,14 @@ class WorkspaceServiceSpec
...
+      terraWorkspaceNextflowRole,
+      terraBucketReaderRole,
+      terraBucketWriterRole,
       rawlsWorkspaceAclManager,
-      multiCloudWorkspaceAclManager
+      multiCloudWorkspaceAclManager,
+      fastPassServiceConstructor
     ) _
 
     def cleanupSupervisor =
```
cahrens, line 319: It would be good to test that the proper calls to FastPass are made for create, clone, and delete (and policy removals).

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      projectPolicy <- googleIamDao
+        .getProjectPolicy(GoogleProject(workspace.googleProjectId.value))
+        .map(fromProjectPolicy)
+      bucketPolicy <- googleStorageDAO
+        .getBucketPolicy(GcsBucketName(workspace.bucketName), Some(GoogleProject(workspace.googleProjectId.value)))
+        .map(fromBucketPolicy)
+    } yield {
+      // Role binding quotas do not de-duplicate member emails, hence the conversion of Sets to Lists
+      val numProjectRoleBindings = projectPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedProjectBindings = numProjectRoleBindings + maxProjectRoleBindingsPerUser
```
cahrens: Why is the max number being added on?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+                      protected val terraWorkspaceNextflowRole: String,
+                      protected val terraBucketReaderRole: String,
+                      protected val terraBucketWriterRole: String
+)(implicit protected val executionContext: ExecutionContext, val openTelemetry: OpenTelemetryMetrics[IO])
+    extends LazyLogging {
+
+  private val openTelemetryTags: Map[String, String] = Map("service" -> "FastPassService")
+  private def samWorkspaceRoleToGoogleProjectIamRoles(samResourceRole: SamResourceRole) =
+    samResourceRole match {
+      case SamWorkspaceRoles.projectOwner =>
```
cahrens, line 92: It would be good to have these role groups defined in a single place. For instance, these also appear in `WorkspaceService.setupGoogleProjectIam`. We want to make sure that the set of role definitions stays in sync.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+        Set(terraBillingProjectOwnerRole, terraWorkspaceCanComputeRole, terraWorkspaceNextflowRole)
+      case SamWorkspaceRoles.owner      => Set(terraWorkspaceCanComputeRole, terraWorkspaceNextflowRole)
+      case SamWorkspaceRoles.canCompute => Set(terraWorkspaceCanComputeRole, terraWorkspaceNextflowRole)
+      case _                            => Set.empty[String]
+    }
+
+  private def samWorkspaceRolesToGoogleBucketIamRoles(samResourceRole: SamResourceRole) =
```
cahrens, line 99: Can you use `WorkspaceService.getGoogleBucketPermissionsFromRoles`?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      quotaAvailable =>
+        if (quotaAvailable) {
+          logger.info(
+            s"Adding FastPass access for ${ctx.userInfo.userEmail} in workspace being cloned ${parentWorkspace.toWorkspaceName}"
+          )
+          val expirationDate = DateTime.now(DateTimeZone.UTC).plus(config.grantPeriod.toMillis)
+          for {
+            petEmail <- DBIO.from(samDAO.getUserPetServiceAccount(ctx, childWorkspace.googleProjectId))
+            userAndPet = UserAndPetEmails(WorkbenchEmail(ctx.userInfo.userEmail.value), petEmail)
+            _ <- setupBucketRoles(parentWorkspace, Set(SamWorkspaceRoles.reader), userAndPet, expirationDate)
```
cahrens: Why is there no need to set up project roles on a cloned workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+                                    policyName: SamResourcePolicyName,
+                                    email: String
+  ): ReadWriteAction[Unit] =
+    for {
+      maybeUserSubjectId <- DBIO.from(samDAO.getUserIdInfo(email, ctx).map {
+        case SamDAO.User(x) => x.googleSubjectId
+        case _              => None
+      })
+      if maybeUserSubjectId.isDefined
+      userSubjectId = maybeUserSubjectId.orNull
```
cahrens: It looks really odd to me to not have the block starting at line 179 indented, but perhaps there's something different about "if" in Scala?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
```diff
@@ -0,0 +1,62 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="garwood" id="create_FASTPASS_GRANTS">
+        <createTable tableName="FASTPASS_GRANTS"
+                     remarks="Records information about fastpass grants.">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="USER_SUBJECT_ID"
+                    type="VARCHAR(254)">
+                <constraints primaryKey="true"/>
```
gpolumbo-broad: Do we need to have a composite key here?  I think it would fine/better/easier to make the `id` column the PK and just index the `USER_SUBJECT_ID` column

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      samPolicy <- DBIO.from(samDAO.getPolicy(SamResourceTypeNames.workspace, workspace.workspaceId, policyName, ctx))
+      organizationRoles = samPolicy.roles.flatMap(samWorkspaceRoleToGoogleProjectIamRoles) ++ samPolicy.roles.flatMap(
+        samWorkspaceRolesToGoogleBucketIamRoles
+      )
+      fastPassGrants <- dataAccess.fastPassGrantQuery.findFastPassGrantsForUserInWorkspace(
+        workspace.workspaceIdAsUUID,
+        RawlsUserSubjectId(userSubjectId)
+      )
+      userAndRoleSpecificFastPassGrants = fastPassGrants.filter(grant =>
+        organizationRoles.contains(grant.organizationRole)
```
cahrens: What rules are we filtering out (not removing) here?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="garwood" id="create_FASTPASS_GRANTS">
+        <createTable tableName="FASTPASS_GRANTS"
+                     remarks="Records information about fastpass grants.">
```
gpolumbo-broad, line 8: Suggest adding a link can be added here to the ticket and/or design doc

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: Would it be safer to write the records to the db first (is there any harm if the to-be-written cleanup code calls google to remove roles that did not actually get granted?)?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
gpolumbo-broad, line 1: Not essential, but seeing this column names on their own, I'm not completely sure what they all represent like `ACCOUNT_TYPE` and `RESOURCE_TYPE`.  Are these referencing Terra accounts or resources?  Or Googley stuff?  This is a nit, but maybe renaming the columns so they are more descriptive or slightly less desirable, adding some `remarks` to the column definitions that explain what they are.  Don't need to do this to get my thumb, but I think it would be better for the long term.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -158,6 +167,8 @@ object Boot extends IOApp with LazyLogging {
     )
 
     initAppDependencies[IO](conf, appName, metricsPrefix).use { appDependencies =>
+      implicit val openTelemetryMetrics: OpenTelemetryMetrics[IO] = appDependencies.openTelemetry
```
gpolumbo-broad, line 170: Not a fan of implicits… if we can avoid using them in the future, I think we should try to do that.  For now, this would probably be a pain to go through and fix, so I guess it can stay in this PR, but if we have a chance to be _explicit_ instead, even if it is more verbose, that is my preference.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+object FastPassService {
+  def constructor(config: FastPassConfig,
+                  googleIamDao: GoogleIamDAO,
+                  googleStorageDAO: GoogleStorageDAO,
+                  samDAO: SamDAO,
+                  terraBillingProjectOwnerRole: String,
+                  terraWorkspaceCanComputeRole: String,
+                  terraWorkspaceNextflowRole: String,
+                  terraBucketReaderRole: String,
+                  terraBucketWriterRole: String
```
gpolumbo-broad, line 50: I know this a "quick and dirty" implementation, but I would consider setting this up to take an arbitrary collection of roles.  Who knows how long this stuff will stick around and how the roles we need to Fast Pass will change.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
cahrens: Ah nevermind, finally got to FastPassServiceSpec.scala.

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
```diff
@@ -0,0 +1,560 @@
...
+      ).thenReturn(Future.successful(bucketPolicy))
+
+      val newWorkspaceName = "space_for_workin"
+      val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
+      val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+      val workspaceFastPassGrants =
+        runAndWait(fastPassGrantQuery.findFastPassGrantsForWorkspace(workspace.workspaceIdAsUUID))
+
+      workspaceFastPassGrants should have size 0
+  }
```
cahrens, line 565: Also need to add test coverage for `removePolicyFastPassesForUser`.

cahrens (changes requested):
I'm done with my initial review. I had a few suggestions/questions, and the primary key designation needs to be removed from `userSubjectId`.

I will also try out the bee.

cahrens:
Question-- do auth domains add any wrinkles to this implementation (I don't have a good understanding of auth domains)?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/FastPassGrantComponent.scala
```diff
@@ -0,0 +1,153 @@
...
+trait FastPassGrantComponent {
+  this: DriverComponent with WorkspaceComponent =>
+
+  import driver.api._
+
+  class FastPassGrantTable(tag: Tag) extends Table[FastPassGrantRecord](tag, "FASTPASS_GRANTS") {
+
+    def id = column[Long]("id", O.PrimaryKey, O.AutoInc)
+
+    def workspaceId = column[UUID]("workspace_id")
```
gpolumbo-broad, line 67: I'm confused how this is working right.  It must be if you've gotten this far in the development.  I see conversions of `UUID` -> `String` and `String` -> `UUID`, but the datatype on the column is `BINARY(16)`, so I'd expect to see something in this class that converts `BINARY(16)` -> `String` -> `UUID` and vice versa, but I'm not seeing it.  Am I missing it?  Have you confirmed for sure that this field is serializing and deserializing properly?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      googleStorageDAO,
+      samDAO,
+      terraBillingProjectOwnerRole,
+      terraWorkspaceCanComputeRole,
+      terraWorkspaceNextflowRole,
+      terraBucketReaderRole,
+      terraBucketWriterRole
+    )
+
+  val policyBindingsQuotaLimit = 1500
```
gpolumbo-broad, line 69: Link to where this is coming from so that we know why this is 1500 and can adjust it accordingly if the rules change.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+  val maxProjectRoleBindingsPerUser = 6 // 3 possible roles for each user and pet service account
+  val maxBucketRoleBindingsPerUser = 2 // 1 possible role for each user and pet service account
```
gpolumbo-broad: Where did these numbers come from?  Why are they set to these values and what would I need to know if I need to change these values in the future?

Also - do we want these values hardcoded or should they come from `FastPassConfig`?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
gpolumbo-broad: +1

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+      // Role binding quotas do not de-duplicate member emails, hence the conversion of Sets to Lists
+      val numBucketRoleBindings = bucketPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedBucketBindings = numBucketRoleBindings + maxBucketRoleBindingsPerUser
+
+      expectedBucketBindings < policyBindingsQuotaLimit
+    }
+
+  private def conditionFromExpirationDate(expirationDate: DateTime): Expr =
+    Expr(
+      s"FastPass access for ${ctx.userInfo.userEmail.value} for while IAM propagates through Google Groups",
```
gpolumbo-broad:
```suggestion
      s"Temporary access directly granted to ${ctx.userInfo.userEmail.value} while we wait for Sam managed permissions to propagate through Google Groups",
```

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+                              userAndPet: UserAndPetEmails,
+                              gcpResourceType: IamResourceType,
+                              resourceName: String,
+                              organizationRoles: Set[String],
+                              expiration: DateTime
+  ): ReadWriteAction[Unit] = {
+    val rolesToWrite =
+      Seq((userAndPet.userEmail, IamMemberTypes.User), (userAndPet.petEmail, IamMemberTypes.ServiceAccount)).flatMap(
+        tuple => organizationRoles.map(r => (tuple._1, tuple._2, r))
+      )
```
gpolumbo-broad, line 331: nit - This took me a little while to grok.  This statement is doing quite a few things.  Suggest adding a type declaration onto the `val`.  Also, in the `organizationRoles.map` the fact that it was just called `r` meant that my brain kind of missed it.  Maybe it's just me, but I thought this thing was making a `Tuple2` because of the `r` and the lack of a type declaration.  I don't know, maybe this would read better if you `flatMap` over the `organizationRoles` instead of the `Seq[Tuple2]`?  I don't know.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+        workspace.bucketName,
+        bucketIamRoles,
+        expirationDate
+      )
+    } yield ()
+  }
+
+  private def writeGrantsToDb(workspaceId: String,
+                              userAndPet: UserAndPetEmails,
+                              gcpResourceType: IamResourceType,
```
gpolumbo-broad, line 323: Naming inconsistency.  I personally think `GcpResourceType` is better than `IamResourceType`, since that's really what we're referring too and that enum might be more useful if named as such.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -0,0 +1,444 @@
...
+        .map(fromBucketPolicy)
+    } yield {
+      // Role binding quotas do not de-duplicate member emails, hence the conversion of Sets to Lists
+      val numProjectRoleBindings = projectPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedProjectBindings = numProjectRoleBindings + maxProjectRoleBindingsPerUser
+
+      val numBucketRoleBindings = bucketPolicy.bindings.toList.flatMap(_.members.toList).size
+      val expectedBucketBindings = numBucketRoleBindings + maxBucketRoleBindingsPerUser
+
+      expectedProjectBindings < policyBindingsQuotaLimit && expectedBucketBindings < policyBindingsQuotaLimit
```
gpolumbo-broad, line 430: Do we also need to check that: 
```expectedProjectBindings + expectedBucketBinding < policyBindingsQuotaLimit``` ?  

Put another way, is the Policy Binding Quota at the Project level or at the resource (project/bucket) level?

File: core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala [hunk 31]
```diff
@@ -0,0 +1,444 @@
...
+      } else {
+        logger.info(
+          s"Not enough IAM Policy Role Binding quota available to add FastPass access for ${ctx.userInfo.userEmail.value} in workspace ${workspace.toWorkspaceName}"
+        )
+        DBIO.successful()
+      }
+    }
+  }
+
+  def setupFastPassForUserInClonedWorkspace(parentWorkspace: Workspace,
```
gpolumbo-broad, line 142:
```suggestion
  def setupFastPassForUserInClonedParentWorkspace(parentWorkspace: Workspace,
```
This may be kind of a nit, but upon my first read through of this, I was expecting this method to set up ALL of the FastPass permissions for the Parent AND the Child.  So the naming change is an attempt to clarify what this method does vs. `setupFastPassForUserInNewWorkspace`.  This might also benefit from some comments explaining that this method is meant to work in conjunction with `setupFastPassForUserInNewWorkspace` to setup the right permissions on the Parent from the Child. 

EDIT: A-ha!  See Christina's comment, she was also confused by this.  Add some comments to help differentiate.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
```diff
@@ -0,0 +1,62 @@
...
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="USER_SUBJECT_ID"
+                    type="VARCHAR(254)">
+                <constraints nullable="false"/>
```
gpolumbo-broad, line 14: Suggest adding an index on this column

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
```diff
@@ -0,0 +1,62 @@
...
+            </column>
+            <column name="ACCOUNT_EMAIL"
+                    type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="ACCOUNT_TYPE"
+                    type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="WORKSPACE_ID"
```
gpolumbo-broad, line 24: Suggest adding an index on this column

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230316_fastpass_grants.xml
```diff
@@ -0,0 +1,62 @@
...
+            </column>
+            <column name="RESOURCE_TYPE"
+                    type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="RESOURCE_NAME"
+                    type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="ROLE_NAME"
```
gpolumbo-broad, line 36: Suggest adding an index on this column

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
gpolumbo-broad: Agreed - suggest adding a comment that describes what this thing needs to be doing in simple english

gpolumbo-broad (changes requested):
I am part way through the code review but have made it through `FastPassService`, which I think is the meat.  In any case, I had enough comments that I wanted to post these and then keep reviewing.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
dvoet: what I am getting at here is that the order this code is written is not necessarily the order it is executed. I think no matter the order these commands are written here, google will be updated before the transaction commits and even if the transaction rolls back.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
Ghost-in-a-Jar: It looks like orch just calls out to the clone workspace endpoint on rawls, since we arent changing any endpoint behavior we should be fine.

cahrens: Thanks for checking. Perhaps we should consider changing the comment in the future….

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
Ghost-in-a-Jar: it was copied from other tests in rawls..... 🤷

cahrens: OK, I just wondered because whenever I see something that isn't obviously fake I wonder if it means that there was some setup someplace with this specific account (so the test is dependent on that). So if the test passes with an arbitrary email address I would suggest changing it, but it's not a big deal.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: added a clarifying comment, this function is to check whether we can complete the operation and stay below the quota

cahrens: But isn't `numProjectRoleBindings` the actual number of bindings that will be added?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: we are planning on ripping the FastPassService class out of rawls so we want to keep it as independent as possible

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: because the project roles only concern compute jobs and bigquery stuff -- all of that should be happening in the child project

cahrens: I'm sorry, I don't follow. Don't we want to grant the same permissions for a cloned workspace that we do for a newly created workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: added a clarifying comment, this is actually to short circuit the for-comp (i didnt know you could do that either)

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: Added a clarifying comment

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: Oh, line 178 is related to the loop above it?? Is it possible to actually put it on line 177 to make it clearer?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: Marking resolved based on the fact that the goal is to remove `FastPassService` in the not-too-near future.

tlangs:
@cahrens We found a small bug while testing to make sure that removing the creator of a workspace also removes their FastPass access. We will confirm that we've fixed this bug before merging.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: i think the confusion is coming from the variable names `maxProjectRoleBindingsPerUser` this is the number of role bindings we will be adding in the operation not the total quota. so numProjectRoleBindings + maxProjectRoleBindingsPerUser should be less than the quota to go forward with the operation

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: No because a cloned workspace gets access to the parent workspace's bucket to prevent data duplication, but it doesnt get access to the compute so the parent doesnt pay for that.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: I agree that this name change is helpful… I also wasn't getting that this was for the parent workspace and not the newly created one.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: Ah, I see. I was still confused about which workspace this was operating on… I do think the name change proposed here would help-- https://github.com/broadinstitute/rawls/pull/2279/files/4c376bd8e97a1429395eaf012fcfe2be6f296687#r1153708146.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
cahrens: I think I understand this now (and the code had changed!).

Ghost-in-a-Jar: no it is just a line by itself in the for comp that will short circuit the for comp if it is false. We updated with a comment and new lines to make it more clear

tlangs:
@cahrens We need to implement some new functionality in Rawls to support dynamically removing FastPass grants on role change, so instead of doing that, we're just going to remove _all_ FastPass grants when a role changes. This means that if the workspace creator is removed from the `OWNER` role after workspace creation, all FastPass grants will be wiped and there won't be any concern for inappropriate resource access.

Ghost-in-a-Jar (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
Ghost-in-a-Jar: no, the policy binding limits are per resource type and projects vs buckets are different resource types

cahrens (approved):
Looks like all my questions are answered-- a few nits to clean up later, but in general this looks good.

PR #2283, opened 2023-03-24
AJ-873: Launch WDS from rawls during workspace creation

aaronkanzer:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-873

In this PR:
* include the Leonardo client library in Rawls
* include Leo and WDS values in config (see also https://github.com/broadinstitute/firecloud-develop/pull/3274, in progress)
* during workspace creation, auto-deploy a WDS; catch errors so WDS failures don't prevent workspace creation
* during workspace clone, auto-deploy a WDS; catch errors so WDS failures don't prevent workspace clone
* unit tests

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
```diff
@@ -0,0 +1,25 @@
+package org.broadinstitute.dsde.rawls.workspace
+
+import org.broadinstitute.dsde.workbench.client.leonardo.api.AppsV2Api;
+import org.broadinstitute.dsde.workbench.client.leonardo.model.CreateAppRequest;
+import org.broadinstitute.dsde.workbench.client.leonardo.model.AppType;
+
+import scala.concurrent.Future
```
davidangb: nit: unused import

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
davidangb, line 1: I suggest renaming this to be a "DAO" instead of a "Client", moving this class to the `org.broadinstitute.dsde.rawls.dataaccess` package, and following the DAO conventions there for classes like `SamDAO` or `MethodRepoDAO`.

Specifically - make a base `LeonardoDAO` trait that defines the method signatures, then rename this implementation class as `HttpLeonardoDAO`. Having the base trait (think: a Java interface or abstract class) makes unit testing/mocking much easier down the line.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
```diff
@@ -0,0 +1,25 @@
...
+  def getAppsV2leonardoApi(accessToken: String): AppsV2Api = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(leonardoBasePath)
+    new AppsV2Api(apiClient)
+  }
+
+  def createWDSInstance(token: String, workspaceId: String, appName: String): Unit = {
+    val createAppRequest = new CreateAppRequest();
+    createAppRequest.setAppType(AppType.CROMWELL);
```
davidangb: thinking out loud: should we make "CROMWELL" a value in rawls.conf instead of hardcoding it here?

When we want to switch "CROMWELL" to "WDS", having the app type hardcoded here requires a full release; changing a conf value could be done a bit easier. But there's not THAT much difference.

automation/project/Dependencies.scala
```diff
@@ -13,7 +13,7 @@ object Dependencies {
   val workbenchGoogle2V = s"0.25-${workbenchLibsHash}"
   val workbenchModelV  = s"0.15-${workbenchLibsHash}"
   val workbenchMetricsV  = s"0.5-${workbenchLibsHash}"
-
+  val leonardoClient: ModuleID = "org.broadinstitute.dsde.workbench" %% "leonardo-client" % "1.3.5-178c33e")
```
davidangb: is this change necessary? I don't see the `leonardoClient` being used at all in the automation subproject.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -995,7 +996,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
             } else {
               DBIO.from(Future(Map()))
             }
-
+
```
davidangb: nit: avoid whitespace-only changes like this

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  def getAppsV2leonardoApi(accessToken: String): AppsV2Api = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(leonardoBasePath)
+    new AppsV2Api(apiClient)
+  }
+
+  def createWDSInstance(token: String, workspaceId: String, appName: String): Unit = {
+    val createAppRequest = new CreateAppRequest();
```
davidangb: Scala does not require trailing `;`, and it is idiomatic to omit those. Check the whole PR for this!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
davidangb, line 1: digging in from the API definition down … start at `WorkspaceApiService` line 34, and see that it calls `MultiCloudWorkspaceService.createMultiCloudOrRawlsWorkspace()`, which calls `createMultiCloudWorkspace()`. Creating an Azure workspace does not call this method `WorkspaceService.createWorkspace()` ! I think this code needs to move, probably into `MultiCloudWorkspaceService.createWorkspace()`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3435,6 +3436,14 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
       )
 
+      // If an Azure workspace -- After the workspace has been created, create a WDS instance via Leonardo corresponding with the workspace
+      // TODO: determine how to determine if Azure workspace if (isAzureMcWorkspace(maybeLoadMcWorkspace(savedWorkspace))) (
+      _ = logger.info(s"creating WDS instance - workspace:'${workspaceName}' - UUID:${workspaceId}")
+      _ <- traceWithParent("createWDSInstance", parentContext) { context =>
+        val leonardoClient = new LeonardoClient("a base path");
+        leonardoClient.createWDSInstance("a token", workspaceId, "workspace Name that should be casted to String");
+      }
+
```
davidangb: this may be moot because the code needs to move, but there's a few suggestions here:
* suggest moving the logging statement inside the closure instead of outside
* the `for … yield` block containing this new code is a series of `DBIOAction`s, so this code also needs to be a `DBIOAction`:
  * use `traceDBIOWithParent` instead of `traceWithParent`
  * `traceDBIOWithParent` requires its inner code to return a DBIO, which you can create with `DBIO.from()`
  * `DBIO.from` requires its inner code to be a Future, which you can create with `Future`
* the "context" variable returned from `traceWithParent`/`traceDBIOWithParent` is totally unused; you can tell the compiler to ignore this by naming it `_`.
* you can get the current user's access token from `parentContext.userInfo` as the first argument to leonardoClient.createWDSInstance
* you can use `s"…"` string interpolation to set the app name as the third argument to `leonardoClient.createWDSInstance`


```suggestion
      _ <- traceDBIOWithParent("createWDSInstance", parentContext) { _ =>
        logger.info(s"creating WDS instance - workspace:'${workspaceName}' - UUID:${workspaceId}")
        val leonardoClient = new LeonardoClient("a base path")
        DBIO.from(Future(leonardoClient.createWDSInstance(parentContext.userInfo.accessToken.token,
          workspaceId, s"wds-${workspaceId}")))
      }
```

davidangb (commented):
structural and Scala-idiomatic comments inline!

How are you envisioning you can unit test this?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -448,6 +448,17 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
         )
       )
     }
+    // Necessary for DBIO.from calls
+    import dataSource.dataAccess.driver.api.DBIO
+
+    for {
+      _ <- traceDBIOWithParent("createWDSInstance", parentContext) { _ =>
```
aaronkanzer: @davidangb Perhaps if you have a second to look....

Slightly stumped here how to conform/transform the return value here to work with `DBIO` -- see the following error. Curious to get your thoughts on what might be the best practice here -- thanks again

```
[error] /app/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala:455:9: type mismatch;
[error]  found   : slick.dbio.DBIOAction[Unit,slick.dbio.NoStream,slick.dbio.Effect with slick.dbio.Effect]
[error]  required: scala.concurrent.Future[org.broadinstitute.dsde.rawls.model.Workspace]
[error]       _ <- traceDBIOWithParent("createWDSInstance", parentContext) { _ =>
```

davidangb: Couple comments … 
1. I'd think you want to move the call to create the WDS app up a bit, perhaps just after creation of the storage container. This would place the call inside the `for … yield` that starts at line 383. By moving the app-deployment call inside the for-yield, the return value of this entire `createWorkspace()` method reverts to being the the `yield savedWorkspace` at line 433, and you can just ignore the result of the app-deployment call.
2. Within a for-yield, each step (each statement that uses `<-`) needs to be of the same type. In this for-yield, each of these statements returns a `Future`, so I don't think you want to use `DBIO` here, you want to use `Future`. The creation of the cloud context from lines 397-402 is a good template for this.

I could chat about this live tomorrow if this isn't clear!

aaronkanzer: This helped quite a bit!

The only change I needed to do was use the `traceWithParent` wrapper instead of the `traceDBIOWithParent` in order for the code to compile -- let me know if this is of concern

aaronkanzer:
Sample Curl command to test when running Rawls locally against `https://leonardo.dsde-dev.broadinstitute.org`

```
curl -X 'POST' \
  'https://local.broadinstitute.org:20443/api/workspaces' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer $TOKEN' \
  -H 'Content-Type: application/json' \
  -d '{"namespace":"azure-dev-2023-01-23","name":"<something-unique>","authorizationDomain":[],"attributes":{"description":"this is a test"},"copyFilesWithPrefix":"notebooks/"}
'
```

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -438,13 +438,16 @@ object Boot extends IOApp with LazyLogging {
       val resourceBufferService = new ResourceBufferService(resourceBufferDAO, resourceBufferConfig)
       val resourceBufferSaEmail = resourceBufferConfig.saEmail
 
+      val leonardoConfig = LeonardoConfig.apply(conf)
```
davidangb: By convention, this class extracts the subtrees of `conf` and passes only what each `*Config` class needs. For example,
```scala
val resourceBufferConfig = ResourceBufferConfig(conf.getConfig("resourceBuffer"))
```
as opposed to
```scala
val resourceBufferConfig = ResourceBufferConfig(conf)
```
So I think you'd want to do:
```scala
val leonardoConfig = LeonardoConfig.apply(conf.getConfig("leonardo"))
```
which will of course require some changes to the `LeonardoConfig` class itself.

Note that this is just the convention of this particular file and I'm suggesting this for consistency; the code works just fine as it is.

Additionally, I don't think you need the `.apply()` here:
```scala
val leonardoConfig = LeonardoConfig(conf.getConfig("leonardo"))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/config/LeonardoConfig.scala
```diff
@@ -0,0 +1,15 @@
...
+
+case object LeonardoConfig {
+  def apply[T <: LeonardoConfig](conf: Config): LeonardoConfig = {
+    val leonardoConfig = conf.getConfig("leonardo")
+
+    new LeonardoConfig(
+      leonardoConfig.getString("server"),
+    )
+  }
+}
```
davidangb: I think the `[T <: LeonardoConfig]` is extraneous here. Take a look at `ResourceBufferConfig` or `DataRepoEntityProviderConfig` as examples.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.broadinstitute.dsde.rawls.util.HttpClientUtilsStandard
+import org.broadinstitute.dsde.workbench.client.leonardo.ApiClient
+
+class HttpLeonardoDAO(
+  config: LeonardoConfig
+)(implicit val system: ActorSystem, val materializer: Materializer, val executionContext: ExecutionContext)
+    extends DsdeHttpDAO
+    with LeonardoDAO {
+  val http = Http(system)
+  val httpClientUtils = HttpClientUtilsStandard()
```
davidangb: is there any reason this needs to extend `DsdeHttpDAO`? If you leave that out, you can then simplify a lot more by omitting `(implicit val system: ActorSystem, val materializer: Materializer, val executionContext: ExecutionContext)` and
```scala
  val http = Http(system)
  val httpClientUtils = HttpClientUtilsStandard()
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  def getAppsV2leonardoApi(accessToken: String): AppsV2Api = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(config.baseUrl)
+    new AppsV2Api(apiClient)
+  }
+
+  def createWDSInstance(token: String, workspaceId: UUID, appName: String, appType: String): Unit = {
```
davidangb: nit: since this accepts `appType` as an argument, I'd name this method more generically, like `def createApp` - it's not specific to WDS.

You could even have both this generic `def createApp` method and a WDS-specific method that encapsulates the expected app name and app type, like
```scala
  def createApp(token: String, workspaceId: UUID, appName: String, appType: String): Unit = {
    val createAppRequest = new CreateAppRequest()
    val appTypeEnum = AppType.fromValue(appType)
    createAppRequest.setAppType(appTypeEnum)
    getAppsV2leonardoApi(token).createAppV2(workspaceId.toString, appName, createAppRequest);
  }

  def createWdsApp(token: String, workspaceId: UUID): Unit = {
    createApp(token, workspaceId, s"wds-${workspaceId}", config.appType)
  }
```
See my comments further down on `MultiCloudWorkspaceService` for more about this.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -40,6 +40,7 @@ import org.broadinstitute.dsde.rawls.serviceperimeter.ServicePerimeterService
 import org.broadinstitute.dsde.rawls.user.UserService
 import org.broadinstitute.dsde.rawls.util.TracingUtils._
 import org.broadinstitute.dsde.rawls.util._
+import org.broadinstitute.dsde.rawls.dataaccess.HttpLeonardoDAO
```
davidangb: nit: unused import

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
```diff
@@ -79,7 +83,7 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar {
 
   it should "fail when provided with Google billing account information" in {
     val provider = mock[BillingProfileManagerClientProvider](RETURNS_SMART_NULLS)
-    val config = new MultiCloudWorkspaceConfig(true, None, None)
+    val config = new MultiCloudWorkspaceConfig(true, None, None, None)
```
davidangb: I'm curious: why do some of the `MultiCloudWorkspaceConfig` constructors accept `Some(leonardoConfig)`, but others accept `None`? Can you just use `None` everywhere?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -0,0 +1,38 @@
...
+    with LeonardoDAO {
+  val http = Http(system)
+  val httpClientUtils = HttpClientUtilsStandard()
+
+  def getAppsV2leonardoApi(accessToken: String): AppsV2Api = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(config.baseUrl)
+    new AppsV2Api(apiClient)
+  }
```
davidangb, line 18: can `getAppsV2leonardoApi` be private?

project/Dependencies.scala
```diff
@@ -129,6 +129,7 @@ object Dependencies {
   val billingProfileManager = excludeJakarta("bio.terra" % "billing-profile-manager-client" % "0.1.109-SNAPSHOT")
   val terraCommonLib = tclExclusions(excludeJakarta("bio.terra" % "terra-common-lib" % "0.0.63-SNAPSHOT" classifier "plain"))
   val sam: ModuleID = excludeJakarta("org.broadinstitute.dsde.workbench" %% "sam-client" % "0.1-f554115")
+  val leonardo: ModuleID = excludeJakarta("org.broadinstitute.dsde.workbench" % "leonardo-client_2.11" % "1.3.5-178c33e")
```
davidangb: I haven't looked at the new Leo client - does it need `excludeJakarta`?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -49,6 +54,7 @@ object MultiCloudWorkspaceService {
                   billingProfileManagerDAO: BillingProfileManagerDAO,
                   samDAO: SamDAO,
                   multiCloudWorkspaceConfig: MultiCloudWorkspaceConfig,
+                  leonardoConfig: LeonardoConfig,
                   workbenchMetricBaseName: String
   )(ctx: RawlsRequestContext)(implicit ec: ExecutionContext, system: ActorSystem): MultiCloudWorkspaceService =
     new MultiCloudWorkspaceService(
```
davidangb, line 60: Here's some structural/architectural suggestions:
* instead of accepting `leonardoConfig: LeonardoConfig` as an argument, accept a `leonardoDAO: LeonardoDAO`. Construct the DAO in `Boot` to supply to this argument. This would allow you, in unit tests, to supply a mock dao and still test this service.
* move the `appType = "CROMWELL"` value out of the `multiCloudWorkspaces.azureConfig` config block and into the `leonardo` config block, and make it part of the `LeonardoConfig` class. This eliminates a bunch of changes in `MultiCloudWorkspaceConfig`; it also feels like it encapsulates Leo app names inside Leo-related classes instead of in Azure-wide classes. The only classes that would be looking at that config value are now `LeonardoConfig` and `HttpLeonardoDAO`. Might also want to rename `appType = "CROMWELL"` to `wdsType = "CROMWELL"` to make it clear which "app" you're concerned with.
* here in `MultiCloudWorkspaceService` - assuming you've broken out a generic `createApp()` and a WDS-specific `createWdsApp()` in `LeonardoDAO` - you can call the more specific `createWdsApp()`. This further encapsulates logic into the Leo dao and out of this class.

core/src/test/resources/reference.conf
```diff
@@ -130,3 +131,7 @@ multiCloudWorkspaces {
         leonardoWsmApplicationId = "leo"
     }
 }
+
+leonardo {
+    server = "https://leonardo.dsde-dev.broadinstitute.org"
+}
```
davidangb: this will need corresponding changes in https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/rawls/rawls.conf.ctmpl before merging this PR.

davidangb (commented):
comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
aaronkanzer: Removed

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
aaronkanzer: Updated

automation/project/Dependencies.scala
aaronkanzer: Removed

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/LeonardoClient.scala
aaronkanzer: Removed

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aaronkanzer: Updated to `traceWithParent` where workspace creation occurs in Rawls

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
aaronkanzer: Updated

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
aaronkanzer: Removed

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
aaronkanzer: Refactored here to use `createApp` and `createWdsInstance` as separate functions in `HttpLeonardoDAO`

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
aaronkanzer: Refactored so this got resolved

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
aaronkanzer: I wasn't able to get it to be private nor protected within the `trait` definition for `LeonardoDAO` -- was there something you had in mind there?

project/Dependencies.scala
aaronkanzer: Removed `excludeJakarta`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aaronkanzer: Took suggestions into account -- `leonardoDAO` is now used throughout here

core/src/main/scala/org/broadinstitute/dsde/rawls/config/LeonardoConfig.scala
aaronkanzer: Updated

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAOSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+//      dao.createWDSInstance(
+//        "token",
+//        workspaceId,
+//        "something-fun")
+//    }
+  }
+
+  it should "call Leonardo createAppV2 endpoint during app creation" in {
+    val dao = new HttpLeonardoDAO(mockServer.mockServerBaseUrl, "CROMWELL")
+    // This isn't a real unit test, just an example to try to mock out a Leo call...
```
aaronkanzer: Cc @davidangb

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAOSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  val testConf: Config = ConfigFactory.load()
+  val leonardoConfig: LeonardoConfig = LeonardoConfig.apply(testConf)
+
+  "HttpLeonardoDAO" should "get an AppsV2Api object when calling getAppsV2leonardoApi during app creation" in {
+    val dao = new MockLeonardoDAO(leonardoConfig.baseUrl, "CROMWELL")
+    val result = dao.getAppsV2leonardoApi("token")
+    assert(result.isInstanceOf[AppsV2Api])
+  }
+
```
aaronkanzer, line 48: I decided to omit unit tests for other methods since they seem to specifically invoke the Leonardo Client (e.g. regressions should be captured in both contract tests and Leonardo Client tests)

Let me know if you disagree here -- I timeboxed doing simple tests for number of times invoked with no success (I know the code works manually, but the unit testing for it has been a much bigger engineering lift than reward here)

core/src/main/resources/reference.conf
```diff
@@ -128,6 +128,11 @@ dataRepoEntityProvider {
   bigQueryMaximumBytesBilled = 100GB
 }
 
+leonardo {
+    wdsType = "CROMWELL"
+    server = "https://leonardo.dsde-dev.broadinstitute.org"
```
yuliadub: just wanted to check that hardcoding to "dev" is intentional?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
davidangb: I don't think `getAppsV2leonardoApi` needs to be defined in the trait. The trait could only list `createWDSInstance` and  `createApp` - those are the methods that outside callers should be using. `getAppsV2leonardoApi` is specific to this class and could be private here and not exposed anywhere else.

core/src/main/scala/org/broadinstitute/dsde/rawls/config/LeonardoConfig.scala
```diff
@@ -0,0 +1,15 @@
+package org.broadinstitute.dsde.rawls.config
+
+import com.typesafe.config.Config
+
+case class LeonardoConfig(baseUrl: String, wdsType: String)
```
davidangb, line 5: this is now unused outside of tests, I believe. You could do away with it entirely, or update `Boot` to use it, like:
```scala
      val leonardoConfig: LeonardoConfig = LeonardoConfig(conf.getConfig("leonardo"))
      val leonardoDAO: LeonardoDAO =
        new HttpLeonardoDAO(leonardoConfig.baseUrl, leonardoConfig.wdsType)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -449,13 +449,18 @@ object Boot extends IOApp with LazyLogging {
       val resourceBufferService = new ResourceBufferService(resourceBufferDAO, resourceBufferConfig)
       val resourceBufferSaEmail = resourceBufferConfig.saEmail
 
+      val leonardoConfig = conf.getConfig("leonardo")
+      val leonardoDAO: LeonardoDAO =
+        new HttpLeonardoDAO(leonardoConfig.getString("server"), leonardoConfig.getString("wdsType"));
```
davidangb: nit: trailing semicolon not needed (so hard to drop that habit!)

also see my comment on `LeonardoConfig`, below

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -0,0 +1,33 @@
...
+  def getAppsV2leonardoApi(accessToken: String): AppsV2Api = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(accessToken)
+    apiClient.setBasePath(baseUrl)
+    new AppsV2Api(apiClient)
+  }
+
+  def createWDSInstance(token: String, workspaceId: UUID, appName: String): Unit = {
+    createApp(token, workspaceId, appName, wdsType);
+  }
```
davidangb: I suggest moving the knowledge of how to name the WDS app into here, instead of keeping it in `MultiCloudWorkspaceService`:
```scala
  def createWDSInstance(token: String, workspaceId: UUID): Unit = {
    createApp(token, workspaceId, s"wds-${workspaceId}", wdsType);
  }
```
Trying to keep all the WDS-specific logic in one place (such as "CROMWELL" vs "WDS" for the app type).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -426,6 +416,19 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
           )
         )
       )
+
+      _ <- traceWithParent("createWDSInstance", parentContext) { _ =>
+        logger.info(
+          s"Creating WDS instance - workspace:'${workspaceRequest.toWorkspaceName.name}' - UUID:${workspaceId}"
```
davidangb: nit: this will only log the workspace name, not the namespace+name. `${workspaceRequest.toWorkspaceName}` should get you both.

core/src/main/resources/reference.conf
davidangb: since this will get overridden by the values in https://github.com/broadinstitute/firecloud-develop/pull/3274, it would only cause a problem if the value in rawls.conf is missing and it falls back to using this value.

might be safest to use something like `server = "add-leo-hostname-here"` to catch that corner case?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAOSpec.scala
davidangb: The most valuable test in my opinion is the one you wrote in MultiCloudWorkspaceServiceSpec: when creating a workspace, assert that we call `LeonardoDAO. createWDSInstance()`.

I think you could do away with the test in this file, it's really only testing code inside the mock dao, and it's testing behavior of the client library, not really testing any of our own code.

Potentially you could write a unit test that validates that the WDS-specific `createWDSInstance()` method supplies `createApp()` with the right values, but that's also not a super high priority.

davidangb (commented):
more comments inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -269,6 +278,10 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
             workspaceManagerDAO.enableApplication(workspaceId, wsmConfig.leonardoWsmApplicationId, context)
           })
         }
+
+        // create a WDS application in Leo
+        _ <- createWdsAppInWorkspace(workspaceId, parentContext)
```
calypsomatic: Just checking - this is expected to fail until cloning is fully developed?  I tried cloning a workspace on my bee and this call went through - it attempted to create a CROMWELL app - but failed due to lack of a storage container.

davidangb: ahh, good catch. It is not expected to fail. I moved WDS(+CBAS) app creation after storage container creation for clones - but I haven't retested yet to validate that, I will do that as soon as I can

davidangb:
transient error downloading from artifactory; jenkins retest

davidangb:
[org.broadinstitute.dsde.test.api.WorkspaceApiSpec.Rawls should not allow readers to import method configs from another workspace](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/60936/testReport/org.broadinstitute.dsde.test.api/WorkspaceApiSpec/Rawls_should_not_allow_readers_to_import_method_configs_from_another_workspace/)
Error Details
503 Service Temporarily Unavailable

Jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
calypsomatic: Looks good now!

calypsomatic (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
davidangb: @calypsomatic update: this is now working for clones, in my testing.

yuliadub (approved):

PR #2295, opened 2023-03-28
ID-507 Add fastpass monitor to cleanup expired grants.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-507

Create an actor that regularly cleans up the expired fast pass grants. This involves making a api call to google and then deleting the grant from the db. I reduced the amt of io where possible with group-by's.

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -133,6 +133,3 @@ sam.timeout = 3 minutes
 prometheus {
   endpointPort = 9098
 }
-
-fastPassMonitor.period = 10m
-fastPassMonitor.period = ${?FASTPASS_MONITOR_PERIOD}
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -36,6 +36,15 @@ health-monitor-dispatcher {
   throughput = 1
 }
 
+fast-pass-monitor-dispatcher {
+  type = Dispatcher
+  executor = "thread-pool-executor"
+  thread-pool-executor {
+    fixed-pool-size = 8
+  }
+  throughput = 1
+}
+
 gcs {
   defaultLocation = "us-central1"
   bucketLogsMaxAge = "180"
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -133,3 +133,6 @@ sam.timeout = 3 minutes
 prometheus {
   endpointPort = 9098
 }
+
+fastPassMonitor.period = 10m
+fastPassMonitor.period = ${?FASTPASS_MONITOR_PERIOD}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -76,6 +76,7 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
+import scala.jdk.DurationConverters.JavaDurationOps
 import scala.language.{higherKinds, postfixOps}
 
 object Boot extends IOApp with LazyLogging {
@@ -483,9 +484,10 @@ object Boot extends IOApp with LazyLogging {
           .withDispatcher("fast-pass-monitor-dispatcher"),
         "fast-pass-monitor"
       )
+      val fastPassMonitorCleanupPeriod = conf.getDuration("fastPassMonitor.cleanupPeriod").toScala
       system.scheduler.scheduleAtFixedRate(
-        1 minute,
-        10 minute,
+        10 seconds,
+        fastPassMonitorCleanupPeriod,
         fastPassMonitor,
         FastPassMonitor.DeleteExpiredGrants
       )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -31,7 +31,7 @@ import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.drs.{DrsHubResolver, MarthaResolver}
 import org.broadinstitute.dsde.rawls.dataaccess.slick.DataAccess
 import org.broadinstitute.dsde.rawls.entities.{EntityManager, EntityService}
-import org.broadinstitute.dsde.rawls.fastpass.FastPassService
+import org.broadinstitute.dsde.rawls.fastpass.{FastPassMonitor, FastPassService}
 import org.broadinstitute.dsde.rawls.genomics.GenomicsService
 import org.broadinstitute.dsde.rawls.google.{HttpGoogleAccessContextManagerDAO, HttpGooglePubSubDAO}
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -592,6 +592,7 @@ object Boot extends IOApp with LazyLogging {
           slickDataSource,
           gcsDAO,
           appDependencies.httpGoogleIamDAO,
+          appDependencies.httpGoogleStorageDAO,
           samDAO,
           notificationDAO,
           pubSubDAO,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -485,12 +485,14 @@ object Boot extends IOApp with LazyLogging {
         "fast-pass-monitor"
       )
 
-      system.scheduler.scheduleAtFixedRate(
-        10 seconds,
-        fastPassConfig.monitorCleanupPeriod.toScala,
-        fastPassMonitor,
-        FastPassMonitor.DeleteExpiredGrants
-      )
+      if (fastPassConfig.enabled) {
+        system.scheduler.scheduleAtFixedRate(
+          10 seconds,
+          fastPassConfig.monitorCleanupPeriod.toScala,
+          fastPassMonitor,
+          FastPassMonitor.DeleteExpiredGrants
+        )
+      }
 
       val workspaceServiceConstructor: RawlsRequestContext => WorkspaceService = WorkspaceService.constructor(
         slickDataSource,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -464,6 +464,7 @@ object Boot extends IOApp with LazyLogging {
         FastPassService.constructor(
           fastPassConfig,
           appDependencies.httpGoogleIamDAO,
+          appDependencies.httpGoogleIamDAO,
           appDependencies.httpGoogleStorageDAO,
           samDAO,
           terraBillingProjectOwnerRole = gcsConfig.getString("terraBillingProjectOwnerRole"),
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -473,6 +473,23 @@ object Boot extends IOApp with LazyLogging {
           terraBucketWriterRole = gcsConfig.getString("terraBucketWriterRole")
         )
 
+      val fastPassMonitor = system.actorOf(
+        FastPassMonitor
+          .props(
+            slickDataSource.dataAccess,
+            appDependencies.httpGoogleIamDAO,
+            appDependencies.httpGoogleStorageDAO
+          )
+          .withDispatcher("fast-pass-monitor-dispatcher"),
+        "fast-pass-monitor"
+      )
+      system.scheduler.scheduleAtFixedRate(
+        1 minute,
+        10 minute,
+        fastPassMonitor,
+        FastPassMonitor.DeleteExpiredGrants
+      )
+
       val workspaceServiceConstructor: RawlsRequestContext => WorkspaceService = WorkspaceService.constructor(
         slickDataSource,
         methodRepoDAO,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -4,7 +4,6 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
 import cats.effect._
-import cats.effect.syntax.resource
 import cats.implicits._
 import com.codahale.metrics.SharedMetricRegistries
 import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets
@@ -31,7 +30,7 @@ import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.drs.{DrsHubResolver, MarthaResolver}
 import org.broadinstitute.dsde.rawls.dataaccess.slick.DataAccess
 import org.broadinstitute.dsde.rawls.entities.{EntityManager, EntityService}
-import org.broadinstitute.dsde.rawls.fastpass.{FastPassMonitor, FastPassService}
+import org.broadinstitute.dsde.rawls.fastpass.FastPassService
 import org.broadinstitute.dsde.rawls.genomics.GenomicsService
 import org.broadinstitute.dsde.rawls.google.{HttpGoogleAccessContextManagerDAO, HttpGooglePubSubDAO}
 import org.broadinstitute.dsde.rawls.jobexec.MethodConfigResolver
@@ -64,7 +63,7 @@ import org.broadinstitute.dsde.workbench.google.{
 import org.broadinstitute.dsde.workbench.google2._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.oauth2.{ClientId, ClientSecret, OpenIDConnectConfiguration}
-import org.broadinstitute.dsde.workbench.openTelemetry.{OpenTelemetryMetrics, OpenTelemetryMetricsInterpreter}
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.http4s.Uri
 import org.http4s.blaze.client.BlazeClientBuilder
 
@@ -76,7 +75,6 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
-import scala.jdk.DurationConverters.JavaDurationOps
 import scala.language.{higherKinds, postfixOps}
 
 object Boot extends IOApp with LazyLogging {
@@ -473,26 +471,6 @@ object Boot extends IOApp with LazyLogging {
           terraBucketReaderRole = gcsConfig.getString("terraBucketReaderRole"),
           terraBucketWriterRole = gcsConfig.getString("terraBucketWriterRole")
         )
-
-      val fastPassMonitor = system.actorOf(
-        FastPassMonitor
-          .props(
-            slickDataSource,
-            appDependencies.httpGoogleIamDAO,
-            appDependencies.httpGoogleStorageDAO
-          )
-          .withDispatcher("fast-pass-monitor-dispatcher"),
-        "fast-pass-monitor"
-      )
-
-      if (fastPassConfig.enabled) {
-        system.scheduler.scheduleAtFixedRate(
-          10 seconds,
-          fastPassConfig.monitorCleanupPeriod.toScala,
-          fastPassMonitor,
-          FastPassMonitor.DeleteExpiredGrants
-        )
-      }
 
       val workspaceServiceConstructor: RawlsRequestContext => WorkspaceService = WorkspaceService.constructor(
         slickDataSource,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -484,10 +484,10 @@ object Boot extends IOApp with LazyLogging {
           .withDispatcher("fast-pass-monitor-dispatcher"),
         "fast-pass-monitor"
       )
-      val fastPassMonitorCleanupPeriod = conf.getDuration("fastPassMonitor.cleanupPeriod").toScala
+
       system.scheduler.scheduleAtFixedRate(
         10 seconds,
-        fastPassMonitorCleanupPeriod,
+        fastPassConfig.monitorCleanupPeriod.toScala,
         fastPassMonitor,
         FastPassMonitor.DeleteExpiredGrants
       )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
@@ -476,7 +476,7 @@ object Boot extends IOApp with LazyLogging {
       val fastPassMonitor = system.actorOf(
         FastPassMonitor
           .props(
-            slickDataSource.dataAccess,
+            slickDataSource,
             appDependencies.httpGoogleIamDAO,
             appDependencies.httpGoogleStorageDAO
           )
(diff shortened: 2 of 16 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/FastPassMonitor.scala
```diff
@@ -0,0 +1,71 @@
...
+  override def receive: Receive = { case DeleteExpiredGrants =>
+    deleteExpiredGrants()
+  }
+
+  private def deleteExpiredGrants(): Unit =
+    dataAccess.fastPassGrantQuery.findExpiredFastPassGrants().map { grants =>
+      grants.to(LazyList).groupBy(_.workspaceId).foreach { case (workspaceId, grants) =>
+        dataAccess.workspaceQuery.findById(workspaceId).map { maybeWorkspace =>
+          val workspace = maybeWorkspace.getOrElse(throw new RuntimeException(s"Could not find workspace $workspaceId"))
+          grants.groupBy(_.accountEmail).foreach { case (_, grants) =>
```
tlangs: We want to group by `userSubjectId`. The same `userSubjectId` will group the users email and pet account email.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/FastPassMonitor.scala
```diff
@@ -0,0 +1,71 @@
...
+
+class FastPassMonitor private (dataAccess: DataAccess, googleIamDao: GoogleIamDAO, googleStorageDao: GoogleStorageDAO)
+    extends Actor {
+  import context.dispatcher
+  override def receive: Receive = { case DeleteExpiredGrants =>
+    deleteExpiredGrants()
+  }
+
+  private def deleteExpiredGrants(): Unit =
+    dataAccess.fastPassGrantQuery.findExpiredFastPassGrants().map { grants =>
```
tlangs: We probably want to transactionalize this somewhere. Probably at the user-level, but somewhere.

tlangs (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/FastPassMonitor.scala
```diff
@@ -0,0 +1,80 @@
...
+  ): Unit = {
+    val organizationRoles = grants.map(_.organizationRole).toSet
+    grants foreach { grant =>
+      val memberType: MemberType = matchGrantMemberType(grant)
+      grant.resourceType match {
+        case GcpResourceTypes.Project =>
+          googleIamDao.removeIamRoles(GoogleProject(workspace.googleProjectId.value),
+                                      WorkbenchEmail(grant.accountEmail.value),
+                                      memberType,
+                                      organizationRoles
```
tlangs: The `organizationRole`s for projects and buckets are different, so you'll have to group the `organizationRoles` by resource type

tlangs (changes requested):

core/src/main/resources/reference.conf
```diff
@@ -127,4 +127,7 @@ dataRepoEntityProvider {
   bigQueryMaximumBytesBilled = 100GB
 }
 
-sam.timeout = 3 minutes
\ No newline at end of file
+sam.timeout = 3 minutes
+
+fastPassMonitor.period = 10m
+fastPassMonitor.period = ${?FASTPASS_MONITOR_PERIOD}
```
tlangs: These configs should go in firecloud-develop. Rawls doesn't support the same config stuff as Sam 😭

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/FastPassMonitor.scala
```diff
@@ -0,0 +1,93 @@
...
+  private def deleteExpiredGrants(): Unit =
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.fastPassGrantQuery.findExpiredFastPassGrants().map { expiredGrants =>
+        logger.info(s"Found ${expiredGrants.size} total expired grants")
+        expiredGrants.to(LazyList).groupBy(_.workspaceId).foreach { case (workspaceId, workspaceGrants) =>
+          logger.info(s"Found ${workspaceGrants.size} expired grants for workspace $workspaceId")
+          dataAccess.workspaceQuery.findV2WorkspaceByIdQuery(UUID.fromString(workspaceId)).map { workspace =>
+            workspaceGrants.groupBy(_.accountEmail).foreach { case (accountEmail, accountEmailGrants) =>
+              logger.info(
+                s"Removing ${accountEmailGrants.size} grants for ${accountEmail} from workspace ${workspace.namespace}/${workspace.name}"
```
tlangs: you can use `workspace.toWorkspaceName` instead of building the workspaceNamespace/workspaceName yourself.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,101 @@
...
+            dataAccess.workspaceQuery.findByIdOrFail(workspaceId).flatMap { workspace =>
+              DBIO.sequence(workspaceGrants.groupBy(_.accountEmail).map { case (_, accountEmailGrants) =>
+                // Remove grants for a given workspace and accountEmail
+                removeGrantsForAccountEmailInWorkspace(dataAccess, workspace, accountEmailGrants)
+              })
+            }
+          })
+        }
+        .map(_ => DBIO.successful(()))
+    }
```
Ghost-in-a-Jar, line 74: We tried rewriting this with a for comp but I think it is easier to follow the grouping logic using a stream and groupby's.

core/src/main/resources/reference.conf
```diff
@@ -133,3 +133,6 @@ sam.timeout = 3 minutes
 prometheus {
   endpointPort = 9098
 }
+
+fastPassMonitor.period = 10m
+fastPassMonitor.period = ${?FASTPASS_MONITOR_PERIOD}
```
tlangs: These configs should go in fc-develop.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -473,6 +474,24 @@ object Boot extends IOApp with LazyLogging {
...
+        FastPassMonitor
+          .props(
+            slickDataSource,
+            appDependencies.httpGoogleIamDAO,
+            appDependencies.httpGoogleStorageDAO
+          )
+          .withDispatcher("fast-pass-monitor-dispatcher"),
+        "fast-pass-monitor"
+      )
+      val fastPassMonitorCleanupPeriod = conf.getDuration("fastPassMonitor.cleanupPeriod").toScala
```
tlangs: The config should be added to `FastPassConfig.scala`

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,101 @@
...
+        case _ => throw new RuntimeException(s"Unsupported resource type ${resourceTypeGrant.resourceType}")
+      }
+      openTelemetry.incrementCounter("fastpass-monitor-google-call", tags = openTelemetryTags).unsafeToFuture()
+    }
+    // Once all the api calls have been made remove all the corresponding grants from the db
+    DBIO.from(
+      openTelemetry.incrementCounter("fastpass-monitor-grant-delete", tags = openTelemetryTags).unsafeToFuture()
+    )
+    DBIO.sequence(grants.map(grant => dataAccess.fastPassGrantQuery.delete(grant.id)))
+  }.map(_ => DBIO.successful(()))
```
tlangs: The extra `()` in `DBIO.successful(())` is unnecessary

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,101 @@
...
+                // Remove grants for a given workspace and accountEmail
+                removeGrantsForAccountEmailInWorkspace(dataAccess, workspace, accountEmailGrants)
+              })
+            }
+          })
+        }
+        .map(_ => DBIO.successful(()))
+    }
+
+  private def removeGrantsForAccountEmailInWorkspace(dataAccess: DataAccess,
```
tlangs: I'm a little confused as to how all the `DBIO` calls are all getting executed in this function. It looks like their being called, but not actually returned by the function.

tlangs (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
Ghost-in-a-Jar: They are the last line in the block so they are getting returned

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
gpolumbo-broad: This stuff I think is worth putting into comments on this method for whomever comes in here next and needs to work with it and says, "why did they do it like this?  why didn't they…"  This is the "meat" of the FastPass cleanup, so some explanation of how it was approached this way and why is I think warranted here.  Same thing with the notes about reducing limiting the I/O - which, btw, where were you trying to limit?  DB read/writes?  Google read/writes?  Both?  Are there particularly hairy situations we can get in where instead of making a handful of requests we make thousands?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,101 @@
...
+          googleStorageDao.removeIamRoles(
+            GcsBucketName(workspace.bucketName),
+            WorkbenchEmail(resourceTypeGrant.accountEmail.value),
+            resourceTypeGrant.accountType,
+            organizationRoles,
+            userProject = Some(GoogleProject(workspace.googleProjectId.value))
+          )
+        case _ => throw new RuntimeException(s"Unsupported resource type ${resourceTypeGrant.resourceType}")
+      }
+      openTelemetry.incrementCounter("fastpass-monitor-google-call", tags = openTelemetryTags).unsafeToFuture()
```
gpolumbo-broad: We may want to have separate counters for Project IAM and Bucket IAM.  I am thinking about quotas and trying to track down if one or the other of these starts causing us grief.

gpolumbo-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -473,6 +474,26 @@ object Boot extends IOApp with LazyLogging {
           terraBucketWriterRole = gcsConfig.getString("terraBucketWriterRole")
         )
 
+      val fastPassMonitor = system.actorOf(
```
marctalbott: We boot up all other akka monitors in `BootMonitors.scala` and only do so for back Rawls. This should follow that pattern unless there's a reason that front Rawls needs to be running this as well

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,125 @@
...
+                               googleIamDao: GoogleIamDAO,
+                               googleStorageDao: GoogleStorageDAO
+)(implicit openTelemetry: OpenTelemetryMetrics[IO])
+    extends Actor
+    with LazyLogging {
+  import context.dispatcher
+
+  private val openTelemetryTags: Map[String, String] = Map("service" -> "FastPassMonitor")
+  override def receive: Receive = { case DeleteExpiredGrants =>
+    openTelemetry.incrementCounter("fastpass-monitor-receive-msg", tags = openTelemetryTags).unsafeToFuture()
```
marctalbott: Curious about this OpenTelemetry stuff, what's the difference between this and the other metrics logic that Rawls uses?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,125 @@
...
+        .flatMap { expiredGrants =>
+          logger.info(s"Found ${expiredGrants.size} total expired grants")
+          // Convert to stream and group by workspaceId
+          DBIO.sequence(expiredGrants.to(LazyList).groupBy(_.workspaceId).map { case (workspaceId, workspaceGrants) =>
+            logger.info(s"Found ${workspaceGrants.size} expired grants for workspace $workspaceId")
+            // Query workspace info then group by accountEmail
+            dataAccess.workspaceQuery.findByIdOrFail(workspaceId).flatMap { workspace =>
+              DBIO.sequence(workspaceGrants.groupBy(_.accountEmail).map { case (_, accountEmailGrants) =>
+                // Remove grants for a given workspace and accountEmail
+                removeGrantsForAccountEmailInWorkspace(dataAccess, workspace, accountEmailGrants)
```
marctalbott: What will happen if one of these fails and retries don't work? Will the monitor continue on to remove the other grants or will it get blocked on the single failing grant?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,125 @@
...
+    // Projects and buckets need different api calls, so group by resourceType
+    accountEmailGrants.groupBy(_.resourceType).map { case (resourceType, resourceTypeGrants) =>
+      val organizationRoles = resourceTypeGrants.map(_.organizationRole).toSet
+      // The grouped resourceTypeGrants are the same except for roles, so we can just take the first one
+      val resourceTypeGrant = resourceTypeGrants.head
+      // For a given resource type, remove the roles for the accountEmail
+      resourceType match {
+        case IamResourceTypes.Project =>
+          googleIamDao.removeRoles(GoogleProject(workspace.googleProjectId.value),
+                                   WorkbenchEmail(resourceTypeGrant.accountEmail.value),
```
marctalbott: nit: `resourceTypeGrant.accountEmail` is already a `WorkbenchEmail`

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,125 @@
...
+        case _ => throw new RuntimeException(s"Unsupported resource type ${resourceTypeGrant.resourceType}")
+      }
+    }
+    // Once all the api calls have been made remove all the corresponding grants from the db
+    DBIO.from(
+      openTelemetry
+        .incrementCounter("fastpass-monitor-grant-delete", count = accountEmailGrants.size, tags = openTelemetryTags)
+        .unsafeToFuture()
+    )
+    DBIO.sequence(accountEmailGrants.map(grant => dataAccess.fastPassGrantQuery.delete(grant.id)))
```
marctalbott: Might be worth adding a `delete` that takes a list of ids and deletes them all rather than deleting one at a time

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
marctalbott: The other metrics stuff I'm referring to is in `RawlsInstrumented` by the way

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
```diff
@@ -0,0 +1,381 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.{BeforeAndAfterAll, OptionValues}
+
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.{Duration, _}
+import scala.concurrent.{Await, ExecutionContext}
+import scala.language.postfixOps
+
+//noinspection NameBooleanParameters,TypeAnnotation,EmptyParenMethodAccessedAsParameterless,ScalaUnnecessaryParentheses,RedundantNewCaseClass,ScalaUnusedSymbol
```
marctalbott, line 56: What's this comment about?

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
```diff
@@ -0,0 +1,381 @@
...
+    testCode: (TestApiService) => T
+  ) = {
+    val apiService = new TestApiService(dataSource, user, fastPassEnabled, fastPassGrantPeriod)
+    try
+      testCode(apiService)
+    finally
+      apiService.cleanupSupervisor
+  }
+
+  "FastPassMonitor" should "remove expired fastpass grants" in withTestDataServicesFastPassGrantPeriodZero { services =>
```
marctalbott, line 322: I'd like to see a test that checks that a single troublesome grant doesn't block up the entire monitor

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
```diff
@@ -0,0 +1,381 @@
...
+    mockServer.startServer()
+  }
+
+  override def afterAll(): Unit = {
+    mockServer.stopServer
+    super.afterAll()
+  }
+
+  // noinspection TypeAnnotation,NameBooleanParameters,ConvertibleToMethodValue,UnitMethodIsParameterless
+  class TestApiService(dataSource: SlickDataSource,
```
marctalbott, line 99: I'm surprised you had to add all of this, I'm curious about what made it necessary?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
tlangs: It doesn't require the same setup in terra-helmfile like other rawls metrics do. Makes it much simpler to add new metrics.

marctalbott: 👍 awesome, that is good to know, thanks!

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
tlangs: This is copied from `WorkspaceServiceSpec`, and it seems to work 🤷‍♂️

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
tlangs: I've added a test in the latest commit.

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitorSpec.scala
tlangs: Keeps the linter from complaining. It was copied from `WorkspaceServiceSpec`

tlangs:
@marctalbott I've refactored the monitor so that an error when calling a google api won't derail the removals of other FastPass grants. I've also created a test to make sure this works. 

I've also split things into separate transactions. That way, a failure in removing one grant won't make the database out-of-date for the rest of the ones that successfully get removed.

tlangs:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
tlangs: re-written as a for-comp. Easier to do when getting the grants and removing them are in two different transactions.

tlangs (approved):

tlangs:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -38,88 +35,66 @@ class FastPassMonitor private (dataSource: SlickDataSource,
...
   /*
    * Remove google roles for expired grants and delete the grants from the database.
    * This is done in a transaction so that if any of the google calls fail, the grants are not deleted.
    * In order to reduce the number of api and db calls, we group the grants by workspaceId, accountEmail, and resourceType.
    */
-  private def deleteExpiredGrants(): Future[ReadWriteAction[Unit]] =
+  private def deleteExpiredGrants(): Future[Unit] =
+    for {
+      grantsGroupedByEmail <- findFastPassGrantsToRemove()
+    } yield Future.sequence(removeFastPassGrants(grantsGroupedByEmail))
```
marctalbott: The type of this for-comp is `Future[Future[Iterable[Unit]]]`. If you move the yield into the for-comp you can get rid of the nested futures which may not run as expected

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -38,88 +35,66 @@ class FastPassMonitor private (dataSource: SlickDataSource,
...
-              DBIO.sequence(workspaceGrants.groupBy(_.accountEmail).map { case (_, accountEmailGrants) =>
-                // Remove grants for a given workspace and accountEmail
-                removeGrantsForAccountEmailInWorkspace(dataAccess, workspace, accountEmailGrants)
-              })
-            }
-          })
+      for {
+        expiredGrants <- dataAccess.fastPassGrantQuery.findExpiredFastPassGrants()
+        _ = logger.info(s"Found ${expiredGrants.size} FastPass grants to clean up")
+        groupedByWorkspaceId = expiredGrants.groupBy(_.workspaceId)
```
marctalbott, line 60: I'm not fully grokking why we need all of these `groupBys`. From what I can tell, a `FastPassGrant` has all the information Rawls needs to revoke the grant except for the workspace's Google project. Could we instead iterate over a list of `(FastPassGrant, GoogleProject)`s and revoke the IAM grant accordingly? 

Let me know if you'd like to hop on a call and discuss!

tlangs: Is it true that workspaces are 1:1 with a google project?

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
tlangs: Fixed

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
marctalbott: Yeah, any workspace that doesn't have its own Google project have been inaccessible to users for a month or two -- Rawls won't return them except in Admin endpoints.

tlangs: ah, ok. Thanks PPW! Updated and pushed

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -0,0 +1,100 @@
...
+        groupedByWorkspace <- DBIO.sequence(groupedByWorkspaceId.map { case (workspaceId, workspaceGrants) =>
+          dataAccess.workspaceQuery.findByIdOrFail(workspaceId).map(workspace => workspace -> workspaceGrants)
+        })
+        _ = groupedByWorkspace.foreach(t =>
+          logger.info(s"Found ${t._2.size} FastPass grants in ${t._1.toWorkspaceName} to clean up")
+        )
+        groupedByEmail = groupedByWorkspace.flatMap { case (workspace, workspaceGrants) =>
```
marctalbott, line 67: Sort of the same question, but why do we need to group by email here?

tlangs: We need to group by email because the calls to the google iam api in workbench-libs works on a per-email basis

marctalbott: We group by email in `FastPassService.removeFastPassGrantsInWorkspaceProject` as well, will that suffice?

marctalbott (approved):
Generally looks good, just have the one remaining question about whether we need the `groupBy` in the monitor

PR #2296, opened 2023-03-29
ID-525 FastPass OpenTelemetry

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-525

Instead of using the hacked-in StatsD metrics in Rawls, FastPass will use the new standard: OpenTelemetry.

This PR adds metrics for per-user FastPass grants and removals, as well as per-principal IAM grants and removals.

I've also re-written how FastPass grants are removed when a workspace is deleted. Before, we were only removing the grants for the requester. Now, we're removing all grants for all users in the workspace.

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -128,3 +128,7 @@ dataRepoEntityProvider {
 }
 
 sam.timeout = 3 minutes
+
+prometheus {
+  endpointPort = 9098
+}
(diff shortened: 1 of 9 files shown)


core/src/main/resources/reference.conf
```diff
@@ -127,4 +127,8 @@ dataRepoEntityProvider {
   bigQueryMaximumBytesBilled = 100GB
 }
 
-sam.timeout = 3 minutes
\ No newline at end of file
+sam.timeout = 3 minutes
+
+prometheus {
+  endpointPort = 9098
+}
```
Ghost-in-a-Jar, line 134: should we add the fastpass configs here too? i noticed they are only in test reference rn

tlangs: The FastPass configs are done in firecloud-develop, unfortunately.

Ghost-in-a-Jar (approved):

PR #2300, opened 2023-03-30
WOR-837 - remove landing zone id from billing project when landing zone creation fails

blakery:
Ticket: [WOR-837](https://broadworkbench.atlassian.net/browse/WOR-837)

The change made earlier was to continue the deletion process if no landing zone id is present.

The landing zone id is set on billing project from the initial response to the lz creation request to WSM. 

When attempting to delete a nonexistent landing zone, WSM returns a 403, not a 404 - so we can’t use the api call status to determine correct behavior.

The additional change here should be to update the LandingZoneCreationStatusRunner to unset the landing zone id if creation fails, since we should be trusting the landing zone service to clean up after itself, and the landing zone id should be useless at that point anyway.

---

**PR checklist**

[WOR-837]: https://broadworkbench.atlassian.net/browse/WOR-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
@@ -90,7 +90,7 @@ class BillingRepository(dataSource: SlickDataSource) {
         .landingZoneId
     }
 
-  def updateLandingZoneId(projectName: RawlsBillingProjectName, landingZoneId: UUID): Future[Int] =
+  def updateLandingZoneId(projectName: RawlsBillingProjectName, landingZoneId: Option[UUID]): Future[Int] =
     dataSource.inTransaction { dataAccess =>
       dataAccess.rawlsBillingProjectQuery.updateLandingZone(projectName, landingZoneId)
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
@@ -86,6 +86,7 @@ class LandingZoneCreationStatusRunner(
                     CreationStatuses.Error,
                     Some(s"Landing Zone creation failed: $msg")
                   )
+                  .flatMap(_ => billingRepository.updateLandingZoneId(billingProjectName, None))
                   .map(_ => Complete)
               case None =>
                 val msg = Option(result.getErrorReport)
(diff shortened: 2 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
```diff
@@ -90,7 +90,7 @@ class BillingRepository(dataSource: SlickDataSource) {
         .landingZoneId
     }
 
-  def updateLandingZoneId(projectName: RawlsBillingProjectName, landingZoneId: UUID): Future[Int] =
+  def updateLandingZoneId(projectName: RawlsBillingProjectName, landingZoneId: Option[UUID]): Future[Int] =
```
blakery, line 93: This had to be changed so that `None` could be passed as a value to un-set the landing zone id.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
```diff
@@ -86,6 +86,7 @@ class LandingZoneCreationStatusRunner(
                     CreationStatuses.Error,
                     Some(s"Landing Zone creation failed: $msg")
                   )
+                  .flatMap(_ => billingRepository.updateLandingZoneId(billingProjectName, None))
```
marctalbott, line 89: Is there any chance that the landing zone still exists in a partial state and we'll be orphaning cloud resources by removing its ID from the billing project record?

cahrens: I do somewhat worry about this. It might be good to run this by @sergiygetlin to get his opinion.

blakery: I was talking this over with @aherbst-broad yesterday.
Our conclusion at the time was basically that we have to trust the landing zone service to clean up after itself.
If it does, the id is useless anyway, and all that happens with it is that you get a 403 when you try to use it, like when we go to delete it (which is indistinguishable from a 403 for other reasons).

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/LandingZoneCreationStatusRunner.scala
cahrens: I have a problem with trust. :) But I will work on it.

marctalbott (approved):

PR #2307, opened 2023-04-04
[WOR-816] Don't directly log HttpGoogleServicesDAO errors to Sentry

cahrens:
This change is to resolve the following scenario:

1) We want to check if Google permissions have synced. If they have not, we expect a 400 or 403 from the Google APIs.
2) At the place where we make the call for this permission check, we catch the Google exception and wrap it in a RawlsExceptionWithErrorReport, retaining its status code. (Implemented in https://github.com/broadinstitute/rawls/pull/2278-- before 2278 the uncaught exception would get wrapped as a 500 error, which got logged to Sentry)
3) We don't log 4xx RawlsExceptionWithErrorReports to Sentry, which is the behavior we want.
4) BUT this logback handling was causing all `HttpGoogleServicesDAO` errors to log to Sentry anyway.

Since uncaught errors from `HttpGoogleServicesDAO` will get wrapped with a 500 error and logged to Sentry, there is no need to also directly log them to Sentry via logback… it leads to duplicate logging (which is behavior we saw before #2278 was implemented).

~~The change in behavior though is that 4xx errors from `HttpGoogleServicesDAO` will not get logged at all if the errors are caught and wrapped with a RawlsExceptionWithErrorReport with the appropriate status code. That's exactly the behavior we want for this permission checking method, but is it generally OK?~~

Solution modified to only suppress `HttpGoogleServicesDAO` logging for 2 specific error messages.

### DIFF
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -67,7 +67,6 @@
         <appender-ref ref="file"/>
         <appender-ref ref="console"/>
         <appender-ref ref="SYSLOG"/>
-        <appender-ref ref="Sentry" />
     </logger>
 
     <logger name="org.broadinstitute.dsde.rawls.google.HttpGooglePubSubDAO" level="debug" additivity="false">
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -67,6 +67,7 @@
         <appender-ref ref="file"/>
         <appender-ref ref="console"/>
         <appender-ref ref="SYSLOG"/>
+        <appender-ref ref="Sentry" />
     </logger>
 
     <logger name="org.broadinstitute.dsde.rawls.google.HttpGooglePubSubDAO" level="debug" additivity="false">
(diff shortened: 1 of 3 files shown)


cahrens:
jenkins retest

blakery (approved):
It makes sense to me that wrapped 4xx errors not to be logged to sentry, generally.

davidangb (commented):
My only concern here is any exceptions thrown by HttpGoogleServicesDAO outside of the context of an API route - such as any of the back-rawls monitors or scheduled actors. Those wouldn't get caught by the `RawlsApiService.exceptionHandler` and as of this PR also wouldn't get caught by `sentry-logback`, and therefore wouldn't make it to Sentry at all. I agree though that double-logging exceptions is bad.

aherbst-broad (commented):
I agree with @davidangb , we need to ensure exceptions from actors aren't going to get dropped on the floor.

cahrens:
OK, I'm going to go back to seeing if there is a way to filter out these specific exceptions (perhaps via `SentryEventFilter`).

cahrens:
@davidangb @marctalbott @aherbst-broad I changed the approach so that we only filter Sentry events directly from `HttpGoogleServicesDAO` if their throwable message contains the 2 specific strings that cause us a lot of spam-- requester pays workspaces and workspaces associated with a disabled billing account. Hopefully these are cases that we wouldn't care about either from actors/monitors.

Since I only did the filtering on the `HttpGoogleServicesDAO` logger, any unhandled exceptions from `RawlsApiService` will still get logged to Sentry through the normal `RawlsExceptionWithErrorReport` mechanism.

davidangb (approved):
lgtm, thank you!

aherbst-broad (approved):

PR #2316, opened 2023-04-11
AJ-939 pass source workspace Id to createApp call

calypsomatic:
As part of [AJ-939](https://broadworkbench.atlassian.net/browse/AJ-939), WDS needs to receive a sourceWorkspaceId when it is being cloned.  Here Rawls passed this id when calling the `createApp` API, which has the sourceWorkspaceId added as a parameter in https://github.com/DataBiosphere/leonardo/pull/3281.

[AJ-939]: https://broadworkbench.atlassian.net/browse/AJ-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


project/Dependencies.scala
```diff
@@ -133,7 +133,8 @@ object Dependencies {
   val billingProfileManager = excludeJakarta("bio.terra" % "billing-profile-manager-client" % "0.1.112-SNAPSHOT")
   val terraCommonLib = tclExclusions(excludeJakarta("bio.terra" % "terra-common-lib" % "0.0.63-SNAPSHOT" classifier "plain"))
   val sam: ModuleID = excludeJakarta("org.broadinstitute.dsde.workbench" %% "sam-client" % "0.1-f554115")
-  val leonardo: ModuleID = "org.broadinstitute.dsde.workbench" % "leonardo-client_2.11" % "1.3.5-178c33e"
+  val leonardo: ModuleID = "org.broadinstitute.dsde.workbench" % "leonardo-client_2.11" % "1.3.5-f0dfd05-SNAP"
```
calypsomatic: Note: the leonardo client will need to be a version that includes https://github.com/DataBiosphere/leonardo/pull/3281

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -17,11 +17,19 @@ class HttpLeonardoDAO(leonardoConfig: LeonardoConfig) extends LeonardoDAO {
...
+  override def createApp(token: String,
+                         workspaceId: UUID,
+                         appName: String,
+                         appType: String,
+                         sourceWorkspaceId: Option[UUID]
+  ): Unit = {
     val createAppRequest = new CreateAppRequest()
+    if (sourceWorkspaceId.nonEmpty) {
+      createAppRequest.setSourceWorkspaceId(sourceWorkspaceId.get.toString)
+    }
```
davidangb, line 37: a more scala-idiomatic implementation:
```scala
    sourceWorkspaceId.foreach { sourceid =>
      createAppRequest.setSourceWorkspaceId(sourceid.toString)
    }
```
check out this thread of opinions: https://stackoverflow.com/questions/6819373/why-is-foreach-better-than-get-for-scala-options and the scaladoc for `foreach`: https://www.scala-lang.org/api/2.13.8/scala/Option.html#foreach[U](f:A=%3EU):Unit

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -17,11 +17,19 @@ class HttpLeonardoDAO(leonardoConfig: LeonardoConfig) extends LeonardoDAO {
...
     val appTypeEnum = AppType.fromValue(appType)
     createAppRequest.setAppType(appTypeEnum)
     getAppsV2leonardoApi(token).createAppV2(workspaceId.toString, appName, createAppRequest)
```
davidangb: for unit testing, I recommend breaking out the logic to construct the `CreateAppRequest` into a standalone unit-testable method, something like this:
```scala
  override def createApp(token: String,
                         workspaceId: UUID,
                         appName: String,
                         appType: String,
                         sourceWorkspaceId: Option[UUID]
                        ): Unit = {
    val createAppRequest = buildAppRequest(appType, sourceWorkspaceId)
    getAppsV2leonardoApi(token).createAppV2(workspaceId.toString, appName, createAppRequest)
  }

  protected[dataaccess] def buildAppRequest(appType: String, sourceWorkspaceId: Option[UUID]): CreateAppRequest = {
    val createAppRequest = new CreateAppRequest()
    sourceWorkspaceId.foreach { sourceId =>
      createAppRequest.setSourceWorkspaceId(sourceId.toString)
    }
    val appTypeEnum = AppType.fromValue(appType)
    createAppRequest.setAppType(appTypeEnum)

    createAppRequest
  }
```

Then, you can unit test directly against `buildAppRequest` with different values for `sourceWorkspaceId`.

Note: I'm not sure if you need the `protected[dataaccess]` on `buildAppRequest` - just guessing without actually trying it what visibility is best for being unit-testable from `HttpLeonardoDAOSpec`

davidangb (approved):
comments inline!

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAOSpec.scala
```diff
@@ -42,8 +42,26 @@ class HttpLeonardoDAOSpec extends TestKit(ActorSystem("HttpLeonardoDAOSpec")) wi
...
       )
   }
 
+  it should "call the createApp API with source workspace id" in {
+    val workspaceId: UUID = UUID.randomUUID()
+    val sourceWorkspaceId: UUID = UUID.randomUUID()
+
+    val leonardoDAO = Mockito.spy(
+      new HttpLeonardoDAO(leonardoConfig)
+    )
```
davidangb: nit: don't need a `spy` here, can just do:
```scala
val leonardoDAO = new HttpLeonardoDAO(leonardoConfig)
```
I wouldn't hold the PR for this, but if you're changing other stuff anyway might as well change this too.

davidangb (approved):
one nit inline, Igtm!

Is there any coordination necessary with the Leo PR? Does it matter if Rawls starts sending this extra argument to Leo before Leo is updated to receive it?

calypsomatic:
> one nit inline, Igtm!
> 
> Is there any coordination necessary with the Leo PR? Does it matter if Rawls starts sending this extra argument to Leo before Leo is updated to receive it?

I'll say yes - we'll need to update Rawls with an appropriate version of the Leo client that accepts the extra parameter, so this PR will have to wait for Leo to be updated.

yuliadub:
not question related to the code added, but I am a bit confused - why are there so many commits that are showing up as part of this PR (from Aaron and David too)?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
```diff
@@ -17,14 +17,28 @@ class HttpLeonardoDAO(leonardoConfig: LeonardoConfig) extends LeonardoDAO {
...
+  protected[dataaccess] def buildAppRequest(appType: String, sourceWorkspaceId: Option[UUID]): CreateAppRequest = {
     val createAppRequest = new CreateAppRequest()
+    sourceWorkspaceId.foreach { sourceId =>
+      createAppRequest.setSourceWorkspaceId(sourceId.toString)
+    }
     val appTypeEnum = AppType.fromValue(appType)
     createAppRequest.setAppType(appTypeEnum)
-    getAppsV2leonardoApi(token).createAppV2(workspaceId.toString, appName, createAppRequest)
+
+    createAppRequest
```
yuliadub, line 41: is this line meant to be here? (not sure if is scala syntax I may not be familiar with)

calypsomatic:
> not question related to the code added, but I am a bit confused - why are there so many commits that are showing up as part of this PR (from Aaron and David too)?

I started working off Aaron's branch that David took over before it got merged.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpLeonardoDAO.scala
calypsomatic: Yep, in scala you don't have to write `return createAppRequest`, it just returns whatever the last line is

yuliadub: ahhh cool!

yuliadub (approved):

yuliadub:
> > not question related to the code added, but I am a bit confused - why are there so many commits that are showing up as part of this PR (from Aaron and David too)?
> 
> I started working off Aaron's branch that David took over before it got merged.

ah that totally makes sense! I wish github had a better way of indicating that. thanks for explaining!

PR #2317, opened 2023-04-12
ID-551 FastPass ACL Updates

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-551

Implement FastPass for when users are added to an existing workspace. 

When  a user is added to a workspace, Rawls will check with Sam to make sure the user is enabled. If so, Rawls will set up the appropriate FastPass grants for the user. These grants are cleaned up just like other grants.

This required a bit of a refactoring of the FastPass code. 

I've taken into account some of the changes in https://github.com/broadinstitute/rawls/pull/2318 and made calls to FastPass even more resilient to errors. This required letting `FastPassService` start its own DB transactions, because a failure in the transaction, even if caught, would bubble up and cause the overall operation calling FastPass to fail. Now, `FastPassService` methods start their own DB transaction and return a `Future` to make sure other operations in Rawls aren't affected by failures.

Testing was a bear. I ended up re-writing most of the tests to be independent of other services. I've also implemented a `MockFastPassService` that sets up all of its mocked calls to Sam and Google internally, rather than sharing those DAOs with whatever else the test is doing. 

Bit of a monster PR, but it significantly expands FastPass functionality!

---

**PR checklist**


tlangs:
Oh geez, I need to squash some commits.

tlangs:
jenkins retest

tlangs:
Jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1175,17 +1176,20 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-                .setupFastPassForUserInClonedWorkspace(sourceWorkspaceContext, destWorkspaceContext)
-            )
             _ = clonedWorkspaceCounter.inc()
 
           } yield (sourceWorkspaceContext, destWorkspaceContext),
         // read committed to avoid deadlocks on workspace attr scratch table
         TransactionIsolation.ReadCommitted
       )
-
+      _ <- traceWithParent("FastPassService.setupFastPassClonedWorkspace", parentContext)(childContext =>
```
aherbst-broad, line 1185: I want to make sure I'm understanding this change. 

Previously, we were setting up the fastpass grants for the user on the source workspace with READER inside the clone workspace DB transaction. 

Now, we'll be setting up the fastpass grants on the source workspace _outside_ of the db transaction but in a new fastpass transaction _and_ initiating the sync of the grants. Is that right?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -3550,9 +3554,6 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
           samDAO.getPetServiceAccountKeyForUser(savedWorkspace.googleProjectId, ctx.userInfo.userEmail)
         )
       )
-      _ <- traceDBIOWithParent("FastPassService.setupFastPassNewWorkspace", parentContext)(childContext =>
```
aherbst-broad, line 3553: This was moved to L257 right?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1456,10 +1460,10 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
             } else Future.successful()
 
           _ <- workspaceAclManager.maybeShareWorkspaceNamespaceCompute(policyAdditions, workspaceName, ctx)
-          _ <- Future.traverse(policyRemovals.map(_._2)) { case email =>
-            dataSource.inTransaction { dataAccess =>
-              fastPassServiceConstructor(ctx, dataAccess).removeFastPassesForUserInWorkspace(workspace, email)
-            }
+
+          // Sync FastPass grants once ACLs are updated
+          _ <- Future.traverse(policyRemovals.map(_._2) ++ policyAdditions.map(_._2)) { case email =>
```
aherbst-broad, line 1465: So the change here is to sync both shares and unshares right?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
tlangs: That's right. Because of how Slick works, any uncaught error within a `DBIO` causes the entire transaction to fail if its not caught. Since we do for-comps and wrap `Future`s in `DBIO`, every single API requests or DB call would needs its own `cleanup`. Its far easier to just start a new transaction that can fail after the workspace is set up so that workspace operations aren't interfered with by a wayward FastPass error.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
tlangs: yup!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
tlangs: Yes. `FastPassService` actually reaches out to Sam to get the user's permissions on the workspace. As a result, we don't need to pass how the users' access has changed, just that the users' access did change. `FastPassService` also checks Sam to see if the user is enabled, so that a disabled user doesn't get access to a workspace through being added.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -205,175 +204,163 @@ class FastPassService(protected val ctx: RawlsRequestContext,
...
+      .inTransaction { implicit dataAccess =>
+        for {
+          maybeUserStatus <- DBIO.from(samDAO.getUserStatus(ctx))
+          if maybeUserStatus.isDefined
+          samUserInfo = maybeUserStatus.map(SamUserInfo.fromSamUserStatus).orNull
+
+          petEmail <- DBIO.from(samDAO.getUserPetServiceAccount(ctx, childWorkspace.googleProjectId))
+          userType = getUserType(samUserInfo.userEmail)
+          userAndPet = UserAndPetEmails(samUserInfo.userEmail, userType, petEmail)
+          _ <- removeParentBucketReaderGrant(parentWorkspace, samUserInfo)
```
Ghost-in-a-Jar, line 229: Why do we need to do this for a cloned workspace?

tlangs: see: https://github.com/broadinstitute/rawls/pull/2326

tlangs: Basically, if a user clones a single workspace many times, they can run up against a quota that says there can only be 20 policy bindings for a user to the same role with different conditions. So, we just clear out the previous bucket reader grant and re-make it again with a new condition.

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -205,175 +204,163 @@ class FastPassService(protected val ctx: RawlsRequestContext,
...
-              )
-              _ <- DBIO
-                .from(
-                  openTelemetry.incrementCounter("fastpass-granted-user", tags = openTelemetryTags).unsafeToFuture()
-                )
-            } yield ()
-          } else {
-            logger.info(
-              s"Not enough IAM Policy Role Binding quota available to add FastPass access for ${ctx.userInfo.userEmail.value} in parent workspace ${parentWorkspace.toWorkspaceName}"
+          _ <- removeFastPassesForUserInWorkspace(workspace, samUserInfo)
```
Ghost-in-a-Jar, line 268: Just to make sure I understand this -- during the sync we remove any grants they have and then add the ones they should have?

tlangs: Yup!

tlangs:
jenkins retest

tlangs:
jenkins retest

Ghost-in-a-Jar (approved):

PR #2320, opened 2023-04-15
ID-563 Make FastPassMonitor not run IAM updates in parallel

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-563

`FastPassService.removeFastPassGrantsInWorkspaceProject` already groups the grants its removing by email and makes sure that they don’t conflict by executing policy updates in series. `FastPassMonitor` no longer pre-groups the expired grants and instead let `FastPassService` do it.

I've modified `FastPassService.removeFastPassGrantsInWorkspaceProject` to return a list of errors and the failed FastPass grants to the caller, so that `FastPassMonitor` can output useful log messages. I've also added a test in `FastPassServiceSpec` to make sure `FastPassService.removeFastPassGrantsInWorkspaceProject` is applying all the policy removals it can, and collecting the errors once finished.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
@@ -12,7 +12,7 @@ import org.broadinstitute.dsde.rawls.fastpass.FastPassService.{
   policyBindingsQuotaLimit,
   possibleBucketRoleBindingsPerUser,
   possibleProjectRoleBindingsPerUser,
-  removeFastPassGrantsInWorkspaceProject,
+  RemovalFailure,
   SAdomain,
   UserAndPetEmails
 }
@@ -83,7 +83,6 @@ object FastPassService extends LazyLogging {
   protected val openTelemetryTags: Map[String, String] = Map("service" -> "FastPassService")
 
   type RemovalFailure = (Throwable, Seq[FastPassGrant])
-  type RemovalResult = Either[RemovalFailure, Seq[FastPassGrant]]
 
   /**
     * Remove the FastPass grants in a Google Project.
@@ -118,6 +117,9 @@ object FastPassService extends LazyLogging {
     executionContext: ExecutionContext,
     openTelemetry: OpenTelemetryMetrics[IO]
   ): ReadWriteAction[Seq[RemovalFailure]] = {
+
+    type RemovalResult = Either[RemovalFailure, Seq[FastPassGrant]]
+
     logger.info(
       s"Removing ${fastPassGrants.size} FastPass grants in Google Project: ${googleProjectId.value}"
     )
@@ -402,7 +404,7 @@ class FastPassService(protected val ctx: RawlsRequestContext,
 
   private def removeFastPassGrantsInWorkspaceProject(fastPassGrants: Seq[FastPassGrant],
                                                      googleProjectId: GoogleProjectId
-  ): ReadWriteAction[Unit] =
+  ): ReadWriteAction[Seq[RemovalFailure]] =
     FastPassService.removeFastPassGrantsInWorkspaceProject(fastPassGrants,
                                                            googleProjectId,
                                                            dataAccess,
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -121,13 +136,30 @@ object FastPassService extends LazyLogging {
...
+        DBIO.from(
+          iamUpdates
+            .foldLeft(Future.successful[Seq[RemovalResult]](Seq.empty[RemovalResult]))((a, b) =>
+              a.flatMap(results => b().map(_ +: results))
+            )
+        )
+      successfulIamRemovals = iamRemovals.collect { case Right(seq) => seq }.flatten
+      failedIamRemovals = iamRemovals.collect { case Left(e) => e }
+      _ <- removeGrantsFromDb(successfulIamRemovals.map(_.id), dataAccess, optCtx)
+    } yield failedIamRemovals
```
gpolumbo-broad, line 165: I spent a while looking at this and I don't know whether I am asking you to change anything or not :)  From a code smell perspective, this is making my spidey sense tingle (pardon the mixed metaphors).  There's a lot of massaging of types and shapes of data in this entire method that is hard to track.   I'm glad you figured it out and I have no doubt it was hard, but if I had a reason to come in here and modify this code, I'd be pretty scared.  I don't know "the" answer, but some things I would consider would be extracting methods.  

There's one big block of code in the `.map` and then there seems to be another logical block in the `for` comp.  Within `for` comps, when I see more than one or two statements in the `yield`, that is also something that I might consider extracting.  

I guess what I am asking is for you to take another look at this to see if you can do some simple refactorings that might simplify this a bit.  If you can't do anything meaningful, then that's fine, just let me know that you tried and I'll throw you a thumb.

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
```diff
@@ -644,4 +689,159 @@ class FastPassServiceSpec
...
+                                                             testData.workspaceNoAttrs.googleProjectId,
+                                                             slickDataSource.dataAccess,
+                                                             services.googleIamDAO,
+                                                             services.googleStorageDAO,
+                                                             None
+      )(executionContext, services.openTelemetry)
+
+    runAndWait(DBIO.seq(project1Removal, project2Removal))
+
+    val endTime = DateTime.now()
```
gpolumbo-broad: https://www.baeldung.com/java-8-date-time-intro

Recommend replacing Joda time with with `java.time` libraries.  I think you can replace this stuff with `java.time.Instant` and `java.time.Duration`.

gpolumbo-broad (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
tlangs: Yeah, this is really messy, but in a way scala makes hard to untangle. Dealing with concurrency and collecting errors is always a sore spot, so I'm using the standard Scala `Either` library to collect results and errors. 

The way we have to `flatmap` over the updates is because we need to force the Futures to execute sequentially. This, combined with the error handling, makes it really hard to refactor for readability. 

I did my best to pull the type out so that it wasn't muddying the waters, but I think this is just the result of 2 "hard" problems in software and Scala's hard-to-read-but-idiomatic way of solving it.

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
tlangs: No more Joda time in FastPass. Phew, that was quite the brain-scratcher of a change. Time zones are hard!

tlangs:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
```diff
@@ -679,4 +723,159 @@ class FastPassServiceSpec
...
+
+    // Each IAM update takes 1 second.
+    // Per workspace, there are 2 project calls and 2 bucket calls (user and pet).
+    // Updates within a workspace happen serially, but two workspaces can be updated in parallel.
+    // So, the updates should take 4 seconds. We allow for a range of 3-5 seconds to take hardware speed into account.
+    // A value less than 3 means all updates ran concurrently when they shouldn't have.
+    // A value more than 5 means the separate workspaces were updated sequentially, which is also incorrect behavior.
+    seconds should be > 2
+    seconds should be < 6
+  }
```
Ghost-in-a-Jar, line 1238: good job replicating the behavior in a test

Ghost-in-a-Jar (approved):

gpolumbo-broad (approved):

tlangs:
jenkins retest

PR #2326, opened 2023-04-18
ID-564 One FastPass grant per cloning user

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-564

Simply remove the users current FastPass grant on the cloned workspace before adding a new one.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
@@ -313,8 +313,8 @@ class FastPassService(protected val ctx: RawlsRequestContext,
   ): ReadWriteAction[Unit] = {
     val predicate = (g: FastPassGrant) =>
       g.resourceType.equals(IamResourceTypes.Bucket) &&
-      g.resourceName.equals(parentWorkspace.bucketName) &&
-      g.organizationRole.equals(terraBucketReaderRole)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -281,6 +281,7 @@ class FastPassService(protected val ctx: RawlsRequestContext,
 
               userType = getUserType(samUserInfo.userEmail)
               userAndPet = UserAndPetEmails(samUserInfo.userEmail, userType, petEmail)
+              _ <- removeParentBucketReaderGrant(parentWorkspace, samUserInfo)
```
aherbst-broad, line 284: Why remove the grants if they're going to be re-added immediately after?

tlangs: The grants are conditional, and each grant resets the expiration timer. We could pull the policy, modify the condition, and re-push it, but that's a much larger change with the same number of API calls as just removing and re-adding the role binding with a new expiration.

aherbst-broad: Makes sense, thanks for the explanation.

aherbst-broad (approved):

PR #2336, opened 2023-04-25
AJ-1047: mysql 5.6->5.7 for tests

davidangb:
work in progress, seeing how tests do with a 5.7 mysql image.

going to 5.7 would also require a change to the local-env script (AJ-1042), and changes to BEEs (AJ-1039). Those are out of scope for this PR.

Output from github actions:
```
[info] DataAccessSpec:
[info] DataAccess
[info] - should use utf8 for MySQL's character_set_server (142 milliseconds)
[info]   + actual character set: utf8 
[info] - should be testing against MySQL 5.7 (89 milliseconds)
[info]   + actual version string: 5.7.41 
```


davidangb:
error registering users during swatomation startup. Jenkins retest.

yuliadub (approved):

PR #2349, opened 2023-05-04
AJ-429: attribute temp table should have same character set and collation as real tables

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-429

Found while investigating Rawls compatibility with MySQL 8.

The `ENTITY_ATTRIBUTE_TEMP` temp table, created by the `createEntityAttributeTempTable()` stored procedure, did not have the same `utf8_bin` collation for its namespace and name columns as the real tables.

This hasn't seemed to cause problems on MySQL 5.6, but it does cause one unit test to fail on MySQL 8: CaseSensitivitySpec's "should update all attributes correctly when batch updating entities".

Let's address the temp table now.

---

**PR checklist**


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230504_update_entity_attr_temp_table_procedure.xml
davidangb, line 1: compare this to the same procedure as created in [`20210407_add_entity_attr_temp_table_procedure.xml`](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20210407_add_entity_attr_temp_table_procedure.xml), and note the changes to the `namespace` and `name` columns.

Those columns now use the same character set/collation as the real tables; you can see that in [`20220517_alter_attribute_name_collation.xml`](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20220517_alter_attribute_name_collation.xml).

yuliadub (approved):

aaronkanzer (approved):

PR #2397, opened 2023-06-15
[WOR-1064] Wireup creation of protected data workspaces in Azure

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1064)
## Why
We need to be able to create "protected" Azure workspaces. The policy wireup and protected data landing zone segments are wired up in workspace manager and TPS, Rawls now needs to know whether to attach such a policy during the creation of a workspace.

## This PR
* Adds a `protectedData` boolean to the create workspace request. If this is set and the workspace being created is Azure, Rawls will attach the `terra:protected-data` poilcy input to the creation request to WSM. WSM, in turn, will verify that the target landing zone is likewise protected and fail out if this is not the case. 
* Cloning just works, as the cloned workspace inherits the policy and the same check takes place (i.e., an attempt to clone a protected workspace into a non-protected landing zone will fail out)
---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
@@ -19,6 +19,13 @@ trait WorkspaceManagerDAO {
                                       ctx: RawlsRequestContext
   ): CreatedWorkspace
 
+  def createProtectedWorkspaceWithSpendProfile(workspaceId: UUID,
+                                      displayName: String,
+                                      spendProfileId: String,
+                                      ctx: RawlsRequestContext
+                                     ): CreatedWorkspace
+
+
   def cloneWorkspace(sourceWorkspaceId: UUID,
                      workspaceId: UUID,
                      displayName: String,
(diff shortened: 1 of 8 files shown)


cahrens:
You need to update the Swagger API (and ultimately orchestration, and rawls-model I think).

Also looks like some test coverage is missing.

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerDAO.scala
```diff
@@ -19,6 +19,12 @@ trait WorkspaceManagerDAO {
                                       ctx: RawlsRequestContext
   ): CreatedWorkspace
 
+  def createProtectedWorkspaceWithSpendProfile(workspaceId: UUID,
```
cahrens, line 22: Nit: instead of a separate method, I would have just added an argument to `createWorkpaceWithSpendProfile`.

cahrens (approved):

cahrens:
@aherbst-broad you are likely hitting the Leo issue in fiab-start, so there's no point in re-running automation tests until that is resolved.

blakery (approved):

aherbst-broad:
jenkins retest

PR #2415, opened 2023-06-27
AJ-1112 Numbers shouldn't stringify into scientific notation

calypsomatic:
Ticket: [AJ-1112](https://broadworkbench.atlassian.net/browse/AJ-1112)
If an entity attribute was a number that was larger than about 9 digits, it would be converted to scientific notation when downloaded to a TSV.  (e.g. 2305150090 -> 2.30515009E+9).  This was due to the nature of the `BigDecimal` object that such numbers were unmarshalled as when pulled from the database, which by default limits the string version of the number to 9 digits.  This PR changes stringification of `AttributeNumber`s to plainString and adds a test to check for it.
Since this changes the rawls Model, there will be two follow-on PRs to complete this ticket:
1. After this PR merges, a new model will need to be released and rawls will be updated to reflect that.
2. Since the issue appears upon TSV download which happens through Firecloud-Orchestration, once the model is released orch will need to be updated to use the new model.

---

**PR checklist**

[AJ-1112]: https://broadworkbench.atlassian.net/browse/AJ-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
+++ b/model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
@@ -4,13 +4,11 @@ import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
 import org.scalatest.Assertions
-import org.scalatest.flatspec.AnyFlatSpecLike
 import org.scalatest.freespec.AnyFreeSpec
-import org.scalatest.matchers.should.Matchers
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
-class AttributeSpec extends AnyFreeSpec with Assertions with Matchers {
+class AttributeSpec extends AnyFreeSpec with Assertions {
 
   "AttributeName model" - {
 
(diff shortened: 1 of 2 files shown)


model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
```diff
@@ -4,11 +4,13 @@ import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
 import org.scalatest.Assertions
+import org.scalatest.flatspec.AnyFlatSpecLike
```
davidangb: nit: unused import

model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
```diff
@@ -686,4 +688,26 @@ class AttributeSpec extends AnyFreeSpec with Assertions {
...
+        (AttributeNumber(223456789100L), "223456789100")
+      )
+
+      numbers.foreach { case (attribute, string) =>
+        assertResult(string)(AttributeStringifier(attribute))
+
+      }
+
+    }
+  }
```
davidangb: I wouldn't hold the PR for this, but I always prefer to make individual tests for the various inputs instead of one test that contains an internal loop. When they're bundled into one test with a loop, it'll short-circuit at the first failure, which means that as a developer I might not get all the feedback I care about … like knowing that three of these inputs will fail.  You could do something like:
```scala
  "AttributeStringifier" - {

    val numbers = List(
      (AttributeNumber(BigDecimal(1234567890)), "1234567890"),
      (AttributeNumber(BigDecimal(2234567890000d)), "2234567890000"),
      (AttributeNumber(BigDecimal(4123456789000011d)), "4123456789000011"),
      (AttributeNumber(BigDecimal(2234567891098765432L)), "2234567891098765432"),
      (AttributeNumber(BigDecimal(22345678910987000L)), "22345678910987000"),
      (AttributeNumber(22345678900d), "22345678900"),
      (AttributeNumber(41234567890000d), "41234567890000"),
      (AttributeNumber(223456789100L), "223456789100")
    )

    numbers.foreach { case (attribute, string) =>
      s"should not stringify large number [$string] in scientific notation" in {
        assertResult(string)(AttributeStringifier(attribute))
      }
    }
  }
```

model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
```diff
@@ -4,11 +4,13 @@ import org.broadinstitute.dsde.rawls.RawlsException
...
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
 import org.scalatest.Assertions
+import org.scalatest.flatspec.AnyFlatSpecLike
 import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
-class AttributeSpec extends AnyFreeSpec with Assertions {
+class AttributeSpec extends AnyFreeSpec with Assertions with Matchers {
```
davidangb: the `Matchers` trait (and its import) aren't actually used

calypsomatic:
jenkins retest

davidangb (commented):
I know this is still in draft but I took a look anyway, please forgive!

model/src/test/scala/org/broadinstitute/dsde/rawls/model/AttributeSpec.scala
calypsomatic: Yeah I messed around with how to write the test a bit as you can see from the unused imports!  Happy to rewrite.

davidangb (approved):

PR #2417, opened 2023-06-27
[WOR-1090] Remove v1 billing project creation and Deployment Manager from Rawls

marctalbott:
Ticket: [WOR-1090](https://broadworkbench.atlassian.net/browse/WOR-1090)
* Google removed the `genomics.googleapis.com` API which we were attempting to enable in our Google projects. The necessary RBS fixes have gone out to fix workspace creation, but Deployment Manager is still attempting to enable this API when Rawls issues a Google project creation request as part of the v1 billing project creation flow. As a result, v1 billing project creation is broken. This functionality has been deprecated for ~9 months and has not been used in production in the last 90 days. Rather than fix Deployment Manager, we are taking this opportunity to remove v1 billing project creation and Deployment Manager from Rawls entirely. 
* The remainder of the v1 billing APIs will be removed as a part of [WOR-659](https://broadworkbench.atlassian.net/browse/WOR-659), and Deployment Manager will be retired as a part of [WOR-725](https://broadworkbench.atlassian.net/browse/WOR-725)

---

**PR checklist**

[WOR-1090]: https://broadworkbench.atlassian.net/browse/WOR-1090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WOR-659]: https://broadworkbench.atlassian.net/browse/WOR-659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WOR-725]: https://broadworkbench.atlassian.net/browse/WOR-725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -606,50 +569,16 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
...
-        }
-      }
-
-      // Now we've got an access token, test IAM permissions to see if the SA has permission to create projects.
-      probeSACredential = buildCredentialFromAccessToken(tokenResponse.getAccessToken, billingProbeEmail)
-      hasAccess <- testBillingAccountAccess(billingAccountName, probeSACredential)
-    } yield hasAccess
-  }
+  override def testTerraBillingAccountAccess(billingAccountName: RawlsBillingAccountName): Future[Boolean] =
+    testBillingAccountAccess(billingAccountName, getBillingServiceAccountCredential)
```
marctalbott, line 573: This is the only place where I changed functionality rather than just removed deprecated code. We used to check that this service account that is in `terra-billing@terra.bio` had permission on the user's billing account as a proxy for whether the user had granted access to `terra-billing@terra.bio`. I've replaced it with a check that our billing SA has access to the billing account. This is the SA that we use to update the billing account on workspace Google projects, so it's the account that actually needs access in order for workspace creation work.

cahrens: @marctalbott but it looks like `testBillingAccountAccess` itself calls `testTerraBillingAccountAccess`. How does this not cause infinite recursion?

marctalbott: There's some slightly confusing overloading going on here.

The `testBillingAccountAccess` method that calls this takes a `UserInfo` along with the billing account name which then calls this method. This method calls a different `testBillingAccountAccess` that takes a `Credential` and issues the actual access check to GCP

core/src/main/resources/reference.conf
```diff
@@ -49,12 +49,6 @@ gcs {
   defaultLocation = "us-central1"
   bucketLogsMaxAge = "180"
   google-api-uri = "https://www.googleapis.com"
-
-  deploymentManager {
```
cahrens, line 53: Does this mean there are helm file deletions or firecloud develop deletions that need to happen?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -13,7 +13,7 @@ tags:
   - name: admin
   - name: billing
   - name: billing_v2
-  - name: entities
+  - name: entitiesf
```
cahrens: I believe this is a typo.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
```diff
@@ -55,18 +55,7 @@ trait BillingApiService extends UserInfoDirectives {
                 }
               }
           }
-      } ~
-        path("billing") {
-          post {
```
cahrens, line 60: Is there an orchestration passthrough that needs to be removed?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
```diff
@@ -415,8 +410,8 @@ class BillingApiServiceV2Spec extends ApiServiceSpec with MockitoSugar {
   it should "return 400 when creating a project with a name that is too short" in withEmptyDatabaseAndApiServices {
     services =>
       Post("/billing/v2",
-           CreateRawlsBillingProjectFullRequest(RawlsBillingProjectName("short"),
-                                                services.gcsDAO.accessibleBillingAccountName,
+        CreateRawlsV2BillingProjectFullRequest(RawlsBillingProjectName("short"),
```
cahrens, line 413: Wait, this was a V2 test that was calling the wrong API??

cahrens:
@marctalbott skimming through all this deleted code was an absolute delight! Just a few questions and a typo, but otherwise LGTM.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiService.scala
marctalbott: Yes, PR coming soon

core/src/main/resources/reference.conf
marctalbott: yeah, no helmfile, but there's some firecloud-develop deletion that will need to happen

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2Spec.scala
marctalbott: It was calling the right API, but with the wrong payload. The payloads are nearly identical so the test still worked

core/src/main/resources/swagger/api-docs.yaml
marctalbott: fixed!

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
aherbst-broad: It'd be worth renaming one of them to aid readability and avoid any confusion. I can this being hard to debug 6 months from now.

aherbst-broad (approved):
Love the deletions, thanks for doing this cleanup.

cahrens:
jenkins retest

PR #2433, opened 2023-07-07
[WOR-1035] Add ability to call WSM async delete workspace API

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1035)

## Why
We want to shift to calling the new [WSM async workspace deletion API.](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Workspace/deleteWorkspaceV2). Since the new endpoint is async, it's a single relatively fast call to initiate the deletion followed by quick polling calls. This coupled with it's use of POST mitigates potential timeouts/retries by upstream proxies.

## This PR
* ... is the first step of introducing this endpoint to Rawls and a related POST endpoint for Rawls workspace deletion. 
   * **Note:** The rawls API will not be async this time around. There is design work planned to get full async workspace deletion but this is not that effort.
* Adds the needed methods to the WSM DAO
* Adds the call to the new DAO methods and associated polling wireup in the `MultiCloudWorkspaceService`. The only consumer of this is the creation rollback code [here](https://github.com/broadinstitute/rawls/blob/87ba8f60e27c16e623919f68cdf0e649c031bf5d/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala#L537C11-L537C24).
* Adds a new polling timeout for deletion. This is wired up in [this related FC develop PR](https://github.com/broadinstitute/firecloud-develop/pull/3416). I've given it a relatively generous 10 minutes.

## TODO
* Swat tests are failing as the aforementioned FC develop needs to be merged.
* As noted, the actual wireup for deletion and the related rawls API is forthcoming in future PRs.

---

**PR checklist**

- [ ] Include the JIRA issue number in the PR description and title
- [ ] Make sure Swagger is updated if API changes
  - [ ] **...and Orchestration's Swagger too!**
- [ ] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green, including CI tests
- [ ] **Squash commits and merge** to develop (branches are automatically deleted after merging)
- [ ] Inform other teams of any substantial changes via Slack and/or email

PR #2441, opened 2023-07-10
[WOR-1035] Wire up workspace deletion for MC workspaces

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1035)

## Why
This is the second in the series of PRs to introduce a POST workspace deletion endpoint and to shift over to the async WSM deletion API. See this [initial PR](https://github.com/broadinstitute/rawls/pull/2433) and the linked ticket for the broader context. The intention is to rebase and merge this to `develop` once #2433 merges. 

## This PR
* Shifts the entrypoint for workspace deletion to a new method in the `MultiCloudWorkspaceService`, `deleteMultiCloudOrRawlsWorkspace`. The service delegates to the `WorkspaceService` in the event that the workspace being deleted is a `rawls` workspace, in a manner very similar to workspace creation.
    * The MC workspace deletion code in `WorkspaceService` will be removed in a follow-on PR

---

**PR checklist**


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -547,8 +547,8 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
       workbenchMetricBaseName
     )(testContext)
     val request = WorkspaceRequest(
-      "fake_ns" + UUID.randomUUID().toString,
-      "fake_name",
+      s"fake_ns_${UUID.randomUUID()}",
+      s"fake_name_${UUID.randomUUID()}",
```
aherbst-broad, line 553: I was bumping into naming collisions when running these tests multiple times against the same db

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -92,6 +92,57 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+      result: Option[String] <- workspace.workspaceType match {
+        case RawlsWorkspace => workspaceService.deleteWorkspace(workspaceName)
+        case McWorkspace    => deleteMultiCloudWorkspace(workspace)
+      }
+    } yield result
+
+  private def deleteMultiCloudWorkspace(workspace: Workspace): Future[Option[String]] =
+    for {
+      _ <- deleteWorkspaceInWSM(workspace.workspaceIdAsUUID).recover { case e: ApiException =>
+        if (e.getCode == StatusCodes.NotFound.intValue) {
```
blakery, line 120: There isn't any plausibility of a user being able to delete a workspace in one service but not the other, right? 
Just mentioning it because a lot of the time, services will return 404 if the user can't access the resource. 
I don't think there's actually a problem here: just bringing it up in case I'm wrong about something.

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aherbst-broad: 404 in WSM generally means the resource is not there. WSM will return a 403 otherwise.

marctalbott: Also, the `workspace` resource we're checking the `delete` action against is the same resource WSM will check against and delete. So if Sam tells Rawls that the user can delete the workspace, Sam will tell WSM the same thing

marctalbott (approved):

PR #2444, opened 2023-07-16
QA-2077: Proposed changes to Rawls e2e Azure tests using new infrastructure stack

ichengchang:
Ticket: [QA-2077](https://broadworkbench.atlassian.net/browse/QA-2077), [QA-2078](https://broadworkbench.atlassian.net/browse/QA-2078), [QA-2082](https://broadworkbench.atlassian.net/browse/QA-2082), [QA-2086](https://broadworkbench.atlassian.net/browse/QA-2086), [QA-2089](https://broadworkbench.atlassian.net/browse/QA-2089)

This PR proposes architectural changes to the Rawls e2e Azure tests focusing on separating the infrastructure related stack from the rest of the test stack which is mainly the test code itself.

In the new approach (please refer to this [doc](https://broadworkbench.atlassian.net/wiki/spaces/QA/pages/2807431169/Identify+Requirements+for+BEEs+to+support+Azure+E2E+Testing) for further details about the new stack),

- the billing project is created once and used by all test cases instead of creating a fresh billing project to run each test, saving about 15 seconds of wait time per test.

- access tokens required for testing are injected into the tests. Instead of obtaining access tokens through the use of sensitive service account keys, the new approach use `Workload Identity Federation` which does not leak sensitive data of multiple service account keys.

This is the related [PR](https://github.com/broadinstitute/terra-github-workflows/pull/86) on the infrastructure side.

---

**PR checklist**

[QA-2077]: https://broadworkbench.atlassian.net/browse/QA-2077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[QA-2078]: https://broadworkbench.atlassian.net/browse/QA-2078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-2082]: https://broadworkbench.atlassian.net/browse/QA-2082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-2086]: https://broadworkbench.atlassian.net/browse/QA-2086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QA-2089]: https://broadworkbench.atlassian.net/browse/QA-2089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -1,9 +1,6 @@
 name: rawls-run-azure-e2e-tests
 
 on:
-  schedule:
```
ungwudik, line 4: Why are we removing this?

ichengchang: To ensure this doesn't get accidentally triggered on schedule during POC and testing. We'll reinstate it once this approach has been thoroughly vetted.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,6 +13,46 @@ on:
...
         type: boolean
+      owner-subject:
+        description: 'Owner (used for creating tmp billing project)'
+        required: true
+        default: 'hermione.owner@quality.firecloud.org'
+        type: string
+      non-owner-subject:
+        description: 'Student (non-owner used for e2e testing)'
+        required: true
+        default: 'harry.potter@quality.firecloud.org'
```
jyang-broad: does this email have to exist? or can we try and set
non-owner-${{ github.run_attempt }}@quality.firecloud.org

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -105,14 +204,28 @@ jobs:
         with:
           workflow: .github/workflows/rawls-swat-tests.yaml
           repo: broadinstitute/terra-github-workflows
-          ref: refs/heads/main
+          ref: refs/heads/iv-az-e2e-1
```
jyang-broad: notes to swap this back before merge

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,6 +13,46 @@ on:
...
+        type: string
+      access-token-scopes:
+        description: 'List of OAuth 2.0 access scopes to be included in the generated token'
+        required: true
+        default: 'profile, email, openid'
+        type: string
+      tenant-id:
+        description: 'Azure tenant ID'
+        required: true
+        default: 'fad90753-2022-4456-9b0a-c7e5b934e408'
```
jyang-broad: add a comment on which mrg this is (assuming it's staticTestingMrg)

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,6 +13,46 @@ on:
...
+        type: string
+      subscription-id:
+        description: 'Azure subscription ID'
+        required: true
+        default: 'f557c728-871d-408c-a28b-eb6b2141a087'
```
jyang-broad: can you add a comment on which subscription this is?

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,9 +128,71 @@ jobs:
...
+        uses: google-github-actions/auth@v1
+        with:
+          token_format: 'access_token'
+          workload_identity_provider: 'projects/1038484894585/locations/global/workloadIdentityPools/github-wi-pool/providers/github-wi-provider'
+          service_account: ${{ inputs.service-account }}
+          access_token_scopes: ${{ inputs.access-token-scopes }}
+          access_token_subject: ${{ inputs.non-owner-subject }}
+          export_environment_variables: false
+
+      - name: Generate a random billing project name
```
jyang-broad, line 149: nice

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -105,14 +204,28 @@ jobs:
         with:
           workflow: .github/workflows/rawls-swat-tests.yaml
           repo: broadinstitute/terra-github-workflows
-          ref: refs/heads/main
+          ref: refs/heads/iv-az-e2e-1
+          token: ${{ env.TOKEN }}
+          inputs: '{ "bee-name": "${{ env.BEE_NAME }}", "ENV": "qa", "ref": "refs/heads/${{ needs.init-github-context.outputs.branch }}", "test-group-name": "workspaces_azure", "test-command": "${{ env.rawls_test_command }}", "java-version": "17", "billing-project": "${{ needs.az-e2e-params-gen.outputs.project-name }}", "e2e": "true", "service-account": "${{ inputs.service-account }}" }'
```
jyang-broad: not an issue for PoC but java verions might not want to be hardcoded.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -105,14 +204,28 @@ jobs:
...
+
+  delete-billing-project-v2-from-bee-workflow:
```
jyang-broad, line 207: If we're going to blow up the bee, do we need to do this step? (could be useful for resetting tests and not creating a bee though)

jyang-broad (approved):
Looks good! moves responsibility of billing projects out of the test stack and into GHA

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: This email would need to be a registered Terra user and a valid Google user in `quality.firecloud.org` domain along with appropriate proxy groups. Otherwise, we would need to add a new user `non-owner-${{ github.run_attempt }}@quality.firecloud.org` to `quality.firecloud.org` with all groups, then register it with Sam API and accept the TOS before it can be used in testing.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Yup!

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Updated comment.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: If developers want to do some debugging without recreating BEE, they could use this to remove the billing project and dissociate from the LZ. Otherwise if the BEE will be destroyed, this step is redundant, including it won't have much impact on the test performance.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Added comment.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Yup!

cahrens (changes requested):
Please let me know when this PR is cleaned up and ready for review.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,6 +13,46 @@ on:
...
+        type: string
+      mrg-id:
+        description: 'Azure Managed Resource Group name. The default is staticTestingMrg within subscription 8201558-dsp-azure-testing.'
+        required: true
+        default: 'staticTestingMrg'
+        type: string
+      landing-zone-id:
+        description: 'Landing Zone ID. An existing LZID tag within a given MRG.'
+        required: true
+        default: 'f41c1a97-179b-4a18-9615-5214d79ba600'
```
cahrens: Note that the landing zone will need to be recreated from time to time, which means that all the places that reference the ID will need to be updated. So it would be good to think of how to centralize this so we don't have to go around and change a lot of workflow files.

ichengchang: Agreed. Maybe we should have a wilder discussion about a centralized process to enable these IDs to be ingested from the maintainers of these landing zone resources?

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Probably can leave it up to `Rawls` team to decide how they would like to approach this.

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
```diff
@@ -30,48 +24,59 @@ import spray.json._
...
-
-  private val azureManagedAppCoordinates = AzureManagedAppCoordinates(
-    UUID.fromString("fad90753-2022-4456-9b0a-c7e5b934e408"),
-    UUID.fromString("f557c728-871d-408c-a28b-eb6b2141a087"),
-    "staticTestingMrg",
-    Some(UUID.fromString("f41c1a97-179b-4a18-9615-5214d79ba600"))
-  )
+class AzureWorkspacesSpec extends AnyFlatSpec with Matchers with BeforeAndAfterAll with CleanUp {
+  // The values of the following vars are injected from the pipeline.
+  var billingProject: String = _
```
jdcanas, line 37: How exactly does this work if you want to run the tests locally?

ichengchang: You can set up local environment variables for the tests to use. Let me create some templates for that.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -131,7 +246,7 @@ jobs:
   notify-slack-on-failure:
     runs-on: ubuntu-latest
     needs: [init-github-context, rawls-build-tag-publish-job, create-bee-workflow, rawls-swat-e2e-test-job, destroy-bee-workflow] # Want to notify regardless of which step fails
-    if: ${{ failure() }}
+    if: false
```
jyang-broad: I think we need to revert this, assuming this was forced to false for testing.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +134,105 @@ jobs:
...
     runs-on: ubuntu-latest
-    needs: [rawls-swat-e2e-test-job, init-github-context]
+    needs: [init-github-context, rawls-swat-e2e-test-job, delete-billing-project-v2-from-bee-workflow]
     if: ${{ needs.init-github-context.outputs.delete-bee && always() }} # always run to confirm bee is destroyed unless explicitly requested not to
-    permissions:
-      contents: 'read'
-      id-token: 'write'
     steps:
       - name: dispatch to terra-github-workflows
         uses: broadinstitute/workflow-dispatch@v3
```
jyang-broad, line 234: while we're here, can we add `wait-for-completion: false` to destroy bee? ultimately, we don't need to pay attention to this as this would get cleaned up by devops tooling in case of failure

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +134,105 @@ jobs:
...
+            "java-version": "17",
+            "billing-project": "${{ needs.params-gen.outputs.project-name }}",
+            "pipeline-inject": "RAWLS_AZURE_E2E",
+            "user-subjects": ${{ toJson(env.USER_SUBJECTS) }}
+          }'
+
+  delete-billing-project-v2-from-bee-workflow:
+    runs-on: ubuntu-latest
+    needs: [init-github-context, params-gen, rawls-swat-e2e-test-job]
+    if: false
```
jyang-broad: assuming we should remove this or change it to if-always.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +134,105 @@ jobs:
...
+    steps:
+      - name: dispatch to terra-github-workflows
+        uses: broadinstitute/workflow-dispatch@v3
+        with:
+          workflow: .github/workflows/delete-billing-project-v2-from-bee.yaml
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/iv-az-e2e-1
           token: ${{ env.TOKEN }}
-          inputs: '{ "bee-name": "${{ env.BEE_NAME }}", "ENV": "qa", "ref": "refs/heads/${{ needs.init-github-context.outputs.branch }}", "test-group-name": "workspaces_azure", "test-command": "${{ env.rawls_test_command }}", "java-version": "17" }'
+          inputs: '{
```
jyang-broad, line 219: can we set `wait-for-completion: false` here? This might cause it collide with destroy-bee, but ultimately this will get cleaned up one way or another so we don't need to watch for failures.

jyang-broad (approved):
Looked mostly at the GHA setup, looks good! There's a bunch of switches that need to be reverted to final form (if: statements, ref: branches)

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Reverted

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Added `wait-for-completion: false`

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Do you mean `wait-for-completion: true` as this needs to be completed BEFORE `bee-destroy` correct?

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -1,5 +1,17 @@
...
 
+# This workflow can be run on schedule OR dispatched (manually or by another dispatcher workflow)
+# When thw workflow is triggered on schedule, it will use the parameters specified on
+# the RHS of || in init-github-context.extract-inputs step. Otherwise, if the workflow is
+# triggered by a dispatch event, it will use the parameters specified on LHS of || in
+# the init-github-context.extract-inputs step.
+#
+# Ex.
+#
+# echo "student-subjects=${{ toJson(inputs.student-subjects || '["harry.potter@quality.firecloud.org","lavender.brown@quality.firecloud.org"]') }}" >> "$GITHUB_OUTPUT"
```
cahrens, line 11: I'm not clear yet why we need to be able to specify the values different for a schedule vs. dispatch. Can you explain what the use case is?

I'm wondering if we can simplify this and remove the LHS/RHS parsing, as it is difficult to read and remember.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,23 +28,47 @@ on:
...
+      service-account: ${{ steps.extract-inputs.outputs.service-account }}
     steps:
       - name: Get inputs or use defaults
         id: extract-inputs
         run: |
-          echo "branch=${{ inputs.branch || 'develop' }}" >> "$GITHUB_OUTPUT"
-          echo "delete-bee=${{ inputs.delete-bee || false }}" >> "$GITHUB_OUTPUT"
+          echo "branch=${{ inputs.branch                            || 'develop' }}" >> "$GITHUB_OUTPUT"
+          echo "delete-bee=${{ inputs.delete-bee                    || false }}" >> "$GITHUB_OUTPUT"
+          echo "owner-subject=${{ inputs.owner-subject              || 'hermione.owner@quality.firecloud.org' }}" >> "$GITHUB_OUTPUT"
```
cahrens, line 69: Why isn't the default value specification up at line 34 sufficient?

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +131,107 @@ jobs:
...
+
       - name: dispatch to terra-github-workflows
         env:
           rawls_test_command: "testOnly -- -l ProdTest -l NotebooksCanaryTest -n org.broadinstitute.dsde.test.api.WorkspacesAzureTest"
         uses: broadinstitute/workflow-dispatch@v3
         with:
           workflow: .github/workflows/rawls-swat-tests.yaml
           repo: broadinstitute/terra-github-workflows
-          ref: refs/heads/main
+          ref: refs/heads/iv-az-e2e-1
```
cahrens: Will need to remember to swap this back to main.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: The `if: false` should be removed.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -127,6 +241,7 @@ jobs:
           ref: refs/heads/main
           token: ${{ env.TOKEN }}
           inputs: '{ "bee-name": "${{ env.BEE_NAME }}" }'
+          wait-for-completion: false
```
cahrens, line 241: Why set this to false?

automation/application.bee.conf.template
```diff
@@ -0,0 +1,71 @@
+# To run against a local UI, set baseUrl = "http://local.broadinstitute.org/"
+
+
```
cahrens: Why the blank lines?

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: You see the same pattern with `branch` and `delete-bee`. 

```
          echo "branch=${{ inputs.branch                            || 'develop' }}" >> "$GITHUB_OUTPUT"
          echo "delete-bee=${{ inputs.delete-bee                    || false }}" >> "$GITHUB_OUTPUT"
```

Scheduled workflows are configured to run at specific times or intervals `without user interaction`, and they do not provide a mechanism for users to input values during scheduling.

Inputs are commonly associated with workflows triggered by events, pull requests, or dispatch actions, where users can specify inputs when triggering the workflow.

To enable the use of the same code for both scheduled and manual (dispatched) runs, we have adopted an LHS/RHS parsing approach. Otherwise we will have to do something below which doesn't look as straightforward as doing the same thing in a single line of code.

```
if [[ -n "${{ inputs.branch }}" ]]; then
  echo "branch=${{ inputs.branch }}" >> "$GITHUB_OUTPUT"
else
  echo "branch=develop" >> "$GITHUB_OUTPUT"
fi
```

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Yup! Once the pipeline code in the other PR is merged. I'll swap this back to `main`.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Once the pipeline code in the other PR is merged. I'll swap this back to `always()` as @jyang-broad suggested.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: @jyang-broad suggested this and it makes sense. The workflow does not have to wait for the `bee-destroy` job to finish since its success or failure has nothing to do with the test. There is also a job scheduled to clean up left-over BEEs on a regular basis.

automation/application.bee.conf.template
ichengchang: The blank lines are removed.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: That's essentially the same reasons I gave for the previous comment.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: Ok, I see.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: Do you mind if we leave the TODOs as unresolved until they are done? It just makes it easier for me to keep track of things that still remain.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: Can we leave this as unresolved until then?

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +131,106 @@ jobs:
...
+  attach-landing-zone-to-bee-workflow:
+    runs-on: ubuntu-latest
+    needs: [init-github-context, create-bee-workflow, params-gen]
+    steps:
+      - name: dispatch to terra-github-workflows
+        uses: broadinstitute/workflow-dispatch@v3
+        with:
+          workflow: attach-landing-zone-to-bee.yaml
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/iv-az-e2e-1
```
cahrens: Reminder to update this ref when the other PR merges (no need to respond, just logging the TODO).

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +131,106 @@ jobs:
...
         uses: broadinstitute/workflow-dispatch@v3
         with:
           workflow: .github/workflows/rawls-swat-tests.yaml
           repo: broadinstitute/terra-github-workflows
-          ref: refs/heads/main
+          ref: refs/heads/iv-az-e2e-1
+          token: ${{ env.TOKEN }}
+          inputs: '{
+            "bee-name": "${{ env.BEE_NAME }}",
+            "ENV": "qa",
```
cahrens, line 197: Is QA the correct env? I'm not sure what that refers to.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: And actually, it looks like there are 3 places where this appears… I own't comment on them anymore because I'm sure you will change all at the same time.

automation/setup-local-env-for-e2e.sh
```diff
@@ -0,0 +1,425 @@
...
+      USERS_METADATA_JSON_B64=$(printf '%s' "${USERS_METADATA_JSON}" | base64 -b 0)
+    else
+      echo "Error: No suitable base64 encoding option found."
+      exit 1
+    fi
+
+    echo "export USERS_METADATA_B64=\"${USERS_METADATA_JSON_B64}\"" >> ${TEST_RESOURCES}/$e2eEnv
+}
+
+attachLandingZoneToBillingProject() {
```
cahrens: I worry about duplicating all of this… isn't it basically `attach-landing-zone-to-bee.yaml` github workflow? One can't just run the workflow against the bee that they wish to use for running the tests locally?

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
```diff
@@ -31,48 +29,59 @@ import spray.json._
...
+      response.accessLevel should be(Some(ProjectOwner))
+    } finally {
+      Rawls.workspaces.delete(projectName, workspaceName)
+      assertNoAccessToWorkspace(projectName, workspaceName)
     }
   }
 
   it should "allow access to WorkspaceManager API" in {
-    implicit val token = owner.makeAuthToken()
+    // implicit val token = owner.makeAuthToken()
```
cahrens: Remove commented-out line.

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
```diff
@@ -81,171 +90,169 @@ class AzureWorkspacesSpec extends AnyFlatSpec with Matchers with CleanUp {
...
+      Set.empty,
+      Map("disableAutomaticAppCreation" -> "true")
+    )
+    try {
+      val response = workspaceResponse(Rawls.workspaces.getWorkspaceDetails(projectName, workspaceName))
+      response.workspace.name should be(workspaceName)
+      response.workspace.cloudPlatform should be(Some(WorkspaceCloudPlatform.Azure))
+
+      // nonOwner is not a member of the workspace, should not be able to write
+      // val userToken = nonOwner.makeAuthToken()
```
cahrens: Remove commented-out line.

automation/setup-local-env-for-e2e.sh
blakery: I don't think that would work: isn't this setting environmental variables so a locally running instance of rawls can connect to a remote bee?

cahrens: No, I'm talking about the code starting at line 338 that is hitting orch (running in a bee) to create a billing project, and then polling on the billing project creation status.

cahrens:
@ichengchang there is a test case that we need to add to this test, at least until we have a new automated test in place that tests billing project creation and deletion (actually creating and attaching a landing zone)-- we need a test case that verifies that we can delete the billing project with no errors. Billing project deletion used to be covered by these tests because each one created a billing project and deleted it, and if the billing project failed to delete the test would fail. However, now the billing project is deleted in a Github workflow, and the workflow does not fail if the billing project deletion fails… I think that's fine for other teams because that's not what they are trying to test. But on our team, we DO want to test that we can delete a billing project correctly.

For instance, we just had a e2e test failure where billing project deletion failed due to a permissions error. We need to investigate this and continue to monitor it, so I don't want to lose the test coverage.

https://github.com/broadinstitute/terra-github-workflows/actions/runs/6036120646/job/16377802377

I can imagine that you either introduce a test that just uses `withTemporaryBillingProject` to create and destroy a billing project (it would be best if it is the LAST test point that runs because if it fails, the other tests will fail due to the MRG being in use), or you introduce a parameter to the delete billing project workflow that specifies whether the workflow should fail if the billing project cannot be deleted.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -16,23 +28,47 @@ on:
...
+        type: string
+      student-subjects:
+        description: 'A JSON array of Student subjects used for E2E testing'
+        required: true
+        default: '["harry.potter@quality.firecloud.org","ron.weasley@quality.firecloud.org"]'
+        type: string
+      service-account:
+        description: 'Email address or unique identifier of the Google Cloud service account for which to generate credentials'
+        required: true
+        default: 'firecloud-qa@broad-dsde-qa.iam.gserviceaccount.com'
```
jyang-broad, line 44: has this SA been tested? I copied this setup elsewhere to test a different flow and ran into a WIF-not-enabled issue and I don't see an entry in terraform to enable it.

ichengchang: The `WIF` isn't used in `rawls-run-azure-e2e-tests.yaml`, at least for now there's not a need for it because everything is taken care of in `terra-github-workflows`. That's why I decided not to include a terraform entry for it.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: The `if: false` has been flipped to `if: always()`

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: The input `ENV` is used to set the vault path like this `secret/dsde/firecloud/${{ inputs.ENV }}/common/firecloud-account.json`. Since the `BEEs` use data from the `qa` path, the `ENV` has been set to `qa`.

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
ichengchang: Commented line removed

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
ichengchang: Removed

automation/setup-local-env-for-e2e.sh
ichengchang: You can provide the billing project option `--billingProject <value>` if you use the `attach-billing-project-to-landing-zone.yaml` workflow in `terra-github-workflows` to attach it to the LZ. The method `attachBillingProjectToLandingZone` will be bypassed if the billing project is provided.

ichengchang:
> @ichengchang there is a test case that we need to add to this test, at least until we have a new automated test in place that tests billing project creation and deletion (actually creating and attaching a landing zone)-- we need a test case that verifies that we can delete the billing project with no errors. Billing project deletion used to be covered by these tests because each one created a billing project and deleted it, and if the billing project failed to delete the test would fail. However, now the billing project is deleted in a Github workflow, and the workflow does not fail if the billing project deletion fails… I think that's fine for other teams because that's not what they are trying to test. But on our team, we DO want to test that we can delete a billing project correctly.
> 
> For instance, we just had a e2e test failure where billing project deletion failed due to a permissions error. We need to investigate this and continue to monitor it, so I don't want to lose the test coverage.
> 
> https://github.com/broadinstitute/terra-github-workflows/actions/runs/6036120646/job/16377802377
> 
> I can imagine that you either introduce a test that just uses `withTemporaryBillingProject` to create and destroy a billing project (it would be best if it is the LAST test point that runs because if it fails, the other tests will fail due to the MRG being in use), or you introduce a parameter to the delete billing project workflow that specifies whether the workflow should fail if the billing project cannot be deleted.

@cahrens We now have the following new github actions in `terra-github-workflows` [PR](https://github.com/broadinstitute/terra-github-workflows/pull/86)

```
.github/actions/run-delete-billing-project/action.yaml
.github/actions/run-billing-project-poller/action.yaml
```
along with the workflows
```
.github/workflows/delete-billing-project-v2-from-bee.yaml
```

I have a sense these can be used to support the use case (failed billing project deletion) you need, perhaps by adding a flag to report deletion failure.

automation/setup-local-env-for-e2e.sh
cahrens: So do we think there is sufficient value in duplicating all this code related to creating a billing project and polling for the billing project to complete? Again, I worry that this pattern is going to result in the copies of this code in every repository that wishes to use the `attach-landing-zone-to-billing-project` action.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -91,33 +131,107 @@ jobs:
...
+            "ref": "refs/heads/${{ needs.init-github-context.outputs.branch }}",
+            "test-group-name": "workspaces_azure",
+            "test-command": "${{ env.rawls_test_command }}",
+            "java-version": "17",
+            "billing-project": "${{ needs.params-gen.outputs.project-name }}",
+            "e2e-env": "${{ env.E2E_ENV }}",
+            "user-subjects": ${{ toJson(env.USER_SUBJECTS) }}
+          }'
+
+  delete-billing-project-v2-from-bee-workflow:
```
cahrens, line 207: @ichengchang if the billing project fails to delete, will the workflow report the failure? I want to make sure that is in place before we replace the existing version of the e2e test (related to https://github.com/broadinstitute/rawls/pull/2444#issuecomment-1703172297), otherwise we lose coverage.

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
```diff
@@ -31,48 +29,58 @@ import spray.json._
...
-  val owner: Credentials = UserPool.userConfig.Owners.getUserCredential("hermione")
-  val nonOwner: Credentials = UserPool.chooseStudent
-
-  private val azureManagedAppCoordinates = AzureManagedAppCoordinates(
-    UUID.fromString("fad90753-2022-4456-9b0a-c7e5b934e408"),
-    UUID.fromString("f557c728-871d-408c-a28b-eb6b2141a087"),
-    "staticTestingMrg",
-    Some(UUID.fromString("f41c1a97-179b-4a18-9615-5214d79ba600"))
-  )
+class AzureWorkspacesSpec extends AnyFlatSpec with Matchers with BeforeAndAfterAll with CleanUp {
```
cahrens: I am having trouble running this test locally. I don't know why I named the test class differently from the file… can we rename the class as WorkspacesAzureApiSpec so it matches the filename? Otherwise IntelliJ does not seem to want to run the test.

automation/setup-local-env-for-e2e.sh
```diff
@@ -0,0 +1,425 @@
...
+#    --mrgId staticTestingMrg                                  \
+#    --landingZoneId f41c1a97-179b-4a18-9615-5214d79ba600
+#
+# Replace rawls-593017852-dev-1 with the name of an existing BEE environment you want your tests to run against
+# Replace (tenantId, subscriptionId, mrgId, landingZoneId) with new coordinates
+# Please refer to https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/attach-landing-zone-to-bee.yaml
+#
+# This script generates a e2e.env file. You can run tests locally as follows.
+#
+# source ${TEST_RESOURCES}/$e2eEnv
```
cahrens: Typo here-- prints out as `source /Users/roberts/Documents/source/rawls/automation/src/test/resources/$e2eEnv`.

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: The `delete-billing-project-v2-from-bee` workflow now accepts a `silent-on-failure` flag. Set it to "false" will cause `exit 1` on failure.

automation/setup-local-env-for-e2e.sh
```diff
@@ -0,0 +1,425 @@
...
+    echo "#    --mrgId staticTestingMrg                                  \\"
+    echo "#    --landingZoneId f41c1a97-179b-4a18-9615-5214d79ba600"
+    echo "#"
+    echo "# Replace rawls-593017852-1-dev with the name of an existing BEE environment you want your tests to run against"
+    echo "# Replace (tenantId, subscriptionId, mrgId, landingZoneId) with new coordinates"
+    echo "# Please refer to https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/attach-landing-zone-to-bee.yaml"
+    echo "#"
+    echo "# This script generates a \$e2eEnv file. You can run tests locally as follows."
+    echo "#"
+    echo "# source ${TEST_RESOURCES}/\$e2eEnv"
```
cahrens: Or actually, I guess the typo is here? Duplicating all this documentation in a source comment and echo statements seems like it is asking for trouble.

Typo here-- prints out as source /Users/roberts/Documents/source/rawls/automation/src/test/resources/$e2eEnv.

We could also add a comment about how to configure the environment file in IntelliJ. Perhaps:

"If you are running the test via IntelliJ, in the Run/Debug Configuration use the EnvFile tab to provide the path to the generated .env file."

automation/setup-local-env-for-e2e.sh
ichengchang: I agree with you on this point. We should make `--billingProject` a required argument and it will up to the developer to provide a valid billing project already attached to a landing zone.

automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
ichengchang: Updated class name to `WorkspacesAzureApiSpec`

automation/setup-local-env-for-e2e.sh
ichengchang: Updated the help section including IntelliJ run/debug config.

automation/setup-local-env-for-e2e.sh
ichengchang: Removed redundant comments here because it's duplicated in help.

automation/setup-local-env-for-e2e.sh
ichengchang: Simplified this setup script. Make `billingProject` required. We no longer need Azure Coordinates setup in this script.

cahrens (approved):

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: updated to `main`

.github/workflows/rawls-run-azure-e2e-tests.yaml
ichengchang: Updated to `main`

PR #2460, opened 2023-07-25
[WOR-1065] Multiregional bucket migration pipeline

marctalbott:
Ticket: [WOR-1065](https://broadworkbench.atlassian.net/browse/WOR-1065)
* Add new pipeline to migrate multiregional buckets to a single region. 
* Based on existing PPW migration pipeline. I've done my best to call out notable changes between the two pipelines
* Currently runs from start of migration up until the first STS job is issued. Future PRs will port over the remaining functionality

---

**PR checklist**

[WOR-1065]: https://broadworkbench.atlassian.net/browse/WOR-1065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -456,6 +468,41 @@ object BootMonitors extends LazyLogging {
...
+                                                    config: Config,
+                                                    gcsDAO: HttpGoogleServicesDAO,
+                                                    googleIamDAO: GoogleIamDAO,
+                                                    dataSource: SlickDataSource,
+                                                    workspaceService: RawlsRequestContext => WorkspaceService,
+                                                    storageService: GoogleStorageService[IO],
+                                                    storageTransferService: GoogleStorageTransferService[IO],
+                                                    samDAO: SamDAO
+  ) = {
+    val actorConfig = MultiregionalBucketMigrationActor.Config(
```
marctalbott: Changed how this works from the PPW pipeline so that I could fetch the `defaultBucketLocation` from the `gcs` config (value is `us-central1`). I had trouble getting config values from a different config block using the implicit reader like PPW does.

An alternative would be to add in a multiregional bucket migration-specific block of config

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
```diff
@@ -131,4 +131,5 @@
     <include file="changesets/20230316_fastpass_grants.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20230504_update_entity_attr_temp_table_procedure.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20230628_multiregional_bucket_migration.xml" relativeToChangelogFile="true"/>
+    <include file="changesets/20230719_autoupdate_updated_multiregional_bucket.xml" relativeToChangelogFile="true"/>
```
marctalbott, line 134: missed this trigger for the STS and History tables in the initial database table setup

local-dev/templates/docker-rsync-local-rawls.sh
```diff
@@ -88,6 +88,7 @@ start_server () {
     -v rawls-shared-source:/app -w /app \
     -v sbt-cache:/root/.sbt \
     -v jar-cache:/root/.ivy2 \
+    -v ~/.ivy2/local:/root/.ivy2/local \
```
marctalbott, line 91: Required to use locally published workbench-libs versions when running local Rawls, not sure if others have run into this problem before

local-dev/templates/rawls.conf.ctmpl
```diff
@@ -119,7 +119,7 @@ dataRepo {
 
 resourceBuffer {
   projectPool {
-    regular = "cwb_ws_dev_v6"
+    regular = "cwb_ws_dev_v7"
```
marctalbott, line 122: RBS project pool was bumped in this PR - https://github.com/DataBiosphere/terra-resource-buffer/pull/271

project/Dependencies.scala
```diff
@@ -82,7 +82,7 @@ object Dependencies {
   val mysqlConnector: ModuleID =  "com.mysql"                         % "mysql-connector-j"  % "8.0.33"
   val liquibaseCore: ModuleID =   "org.liquibase"                 % "liquibase-core"        % "4.17.2"
 
-  val workbenchLibsHash = "7dbb965"
+  val workbenchLibsHash = "f099feb1-SNAP"
```
marctalbott: Will bump once https://github.com/broadinstitute/workbench-libs/pull/1433 merges

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -150,8 +150,8 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
       updatedCol,
       finishedCol,
       outcomeCol,
-      workspaceBucketIamRemovedCol,
       messageCol,
+      workspaceBucketIamRemovedCol,
```
marctalbott, line 154: Moved so ResultSet could be correctly parsed

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -100,23 +100,23 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
 
   import driver.api._
 
-  implicit def setValueObject[A <: ValueObject]: SetParameter[A] =
+  implicit def setMRBValueObject[A <: ValueObject]: SetParameter[A] =
```
marctalbott, line 103: This and other `MRB` renamings in this file were done to avoid collisions with PPW-related readers.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+        _ <- raiseWhen(workspaceBillingAccount != billingProjectBillingAccount) {
+          makeError(
+            "billing account on workspace differs from billing account on billing project",
+            Map(
+              "workspaceBillingAccount" -> workspaceBillingAccount,
+              "billingProject" -> billingProject.projectName,
+              "billingProjectBillingAccount" -> billingProjectBillingAccount
+            )
+          )
+        }
```
marctalbott, line 294: These checks were all moved from `configureGoogleProject`. We still need to check that billing information is up to date before attempting to migrate a workspace's bucket, but we're not configuring the Google project, so I moved the check here

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+
+            actorSaIdentity = serviceAccount(userInfo.userEmail.value)
+            _ <- storageService
+              .overrideIamPolicy(
+                GcsBucketName(workspace.bucketName),
+                Map(StorageRole.StorageAdmin -> NonEmptyList.one(actorSaIdentity)),
+                bucketSourceOptions =
+                  if (requesterPaysEnabled) List(BucketSourceOption.userProject(googleProjectToBill.value))
+                  else List.empty,
+                version = 3
```
marctalbott, line 315: Using new parameter from https://github.com/broadinstitute/workbench-libs/pull/1433

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+          "migrationId" -> migrationId,
+          "workspace" -> workspaceName,
+          "retry" -> retryCount,
+          "retried" -> now
+        )
+      )
+    } yield ()
+
+  // Create bucket in us-central1-a
+  final def createBucketInTargetRegion(migration: MultiregionalBucketMigration,
```
marctalbott, line 483: This has been changed to not take a source and destination Google project at this migration works within a single Google project. Bucket location is always `us-central1` at the moment as defined by `defaultBucketLocation`, but we could amend that in the future if we wanted to support other regions

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
```diff
@@ -78,6 +80,9 @@ trait DataAccess
       TableQuery[PpwStorageTransferJobs].delete andThen // FK to V1_WORKSPACE_MIGRATION_QUERY
       migrationRetryQuery.delete andThen // FK to V1_WORKSPACE_MIGRATION_QUERY
       workspaceMigrationQuery.delete andThen // FK to workspace
+      TableQuery[MultiregionalStorageTransferJobs].delete andThen // FK to MULTIREGIONAL_BUCKET_MIGRATION_HISTORY
```
aherbst-broad: Curious why this is necessary if there is an on delete cascade FK in the multiregion xfer jobs table.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
aherbst-broad: I think the bucket migration stuff should be it's own config so it's 100% obvious which configs need to be flipped where as this pipeline evolves.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -224,7 +224,7 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
 
         _ <- MonadThrow[ReadWriteAction].raiseWhen(workspace.isLocked) {
           new RawlsExceptionWithErrorReport(
-            ErrorReport(StatusCodes.BadRequest, s"'$workspaceName' cannot be migrated it is locked.")
+            ErrorReport(StatusCodes.BadRequest, s"'$workspaceName' cannot be migrated as it is locked.")
```
aherbst-broad: Nit: I'd adjust this to make it clear that this is a _bucket_ migration, not a workspace migration.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -281,16 +281,13 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
...
-            select null from WORKSPACE where id <> w.id and namespace = w.namespace
-        )
+        where m.#$startedCol is not null and m.#$finishedCol is null
         """.as[Int].head
 
 // The following query uses raw parameters. In this particular case it's safe to do as the
 // values of the `activeStatuses` are known and controlled by us. In general one should use
 // bind parameters for user input to avoid sql injection attacks.
-    final def nextMigration(onlyChild: Boolean) = OptionT[ReadWriteAction, (Long, UUID, String)] {
+    final def nextMigration(rateLimit: Boolean) = OptionT[ReadWriteAction, (Long, UUID, String)] {
```
aherbst-broad: Can you explain what this param is for?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+import java.util.UUID
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+/**
+  * The `MultiregionalBucketMigrationActor` is an Akka Typed Actor [1] that migrates a v2 workspace's
+  * multiregional GCS bucket to a single GCP region.
+  *
+  * The actor migrates workspaces by executing a series of operations (or pipeline) successively.
```
aherbst-broad: Nit: `The actor migrations buckets`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+
+  val storageTransferJobs = MultiregionalStorageTransferJobs.storageTransferJobs
+
+  final def restartMigration: MigrateAction[Unit] =
+    restartFailuresLike(FailureModes.noBucketPermissionsFailure, FailureModes.gcsUnavailableFailure) |
+      reissueFailedStsJobs
+
+  final def startMigration: MigrateAction[Unit] =
+    for {
+      (maxAttempts, maxReties) <- asks(d => (d.maxConcurrentAttempts, d.maxRetries))
```
aherbst-broad: Nit: typo `maxReties` -> `maxRetries`

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
aherbst-broad: I think this makes sense. Is the strictly needed to migrate a bucket or is it more of a Good Clean Living™ thing?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+            bucketInfo <- IO.fromOption(bucketOpt)(noWorkspaceBucketError(migration, workspace))
+            requesterPaysEnabled = Option(bucketInfo.requesterPays()).exists(_.booleanValue())
+
+            actorSaIdentity = serviceAccount(userInfo.userEmail.value)
+            _ <- storageService
+              .overrideIamPolicy(
+                GcsBucketName(workspace.bucketName),
+                Map(StorageRole.StorageAdmin -> NonEmptyList.one(actorSaIdentity)),
+                bucketSourceOptions =
+                  if (requesterPaysEnabled) List(BucketSourceOption.userProject(googleProjectToBill.value))
```
aherbst-broad, line 313: Why does the user project need to be included here if it's requester pays? Does that get clobbered if omitted?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+
+        now <- nowTimestamp
+        _ <- inTransaction { dataAccess =>
+          import dataAccess.setOptionValueObject
+          dataAccess.multiregionalBucketMigrationQuery.update2(
+            migration.id,
+            dataAccess.multiregionalBucketMigrationQuery.tmpBucketCol,
+            Some(tmpBucketName),
+            dataAccess.multiregionalBucketMigrationQuery.tmpBucketCreatedCol,
+            Some(now)
```
aherbst-broad, line 363: Curious why this is needed given the triggers being added.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+                                       sourceBucketName: GcsBucketName,
+                                       destBucketName: GcsBucketName
+  ): MigrateAction[Unit] =
+    MigrateAction { env =>
+      def pollForBucketToBeCreated(interval: FiniteDuration, deadline: Deadline): IO[Unit] =
+        IO.sleep(interval).whileM_ {
+          for {
+            _ <- IO.raiseWhen(deadline.isOverdue()) {
+              WorkspaceMigrationException(
+                message = "Workspace migration failed: timed out waiting for bucket creation",
```
aherbst-broad: nit: bucket migration failed.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+              logBucket = GcsBucketName(
+                GoogleServicesDAO.getStorageLogsBucketName(workspace.googleProjectId)
+              ).some,
+              location = env.defaultBucketLocation.some,
+              autoclassEnabled = true
+            )
+            .compile
+            .drain
+
+          _ <- pollForBucketToBeCreated(interval = 100.milliseconds, deadline = 10.seconds.fromNow)
```
aherbst-broad: Was the 10.seconds value take straight from PPW? I'm wondering if that needs to be more generous.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+            "iamConfigured" -> now
+          )
+        )
+      } yield ()
+    }
+
+  final def issueBucketTransferJob(readMigrations: DataAccess => ReadAction[Vector[MultiregionalBucketMigration]],
+                                   getSrcBucket: (GcsBucketName, GcsBucketName) => GcsBucketName,
+                                   getDstBucket: (GcsBucketName, GcsBucketName) => GcsBucketName,
+                                   getColumn: DataAccess => ColumnName[Option[Timestamp]]
```
aherbst-broad: Nitpick (as I know this was copied from the existing PPW stuff): Took me a bit to grok what this was doing with the call on L615, might be better named `getTimestampColumnToUpdate` or something equally explicit.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -0,0 +1,857 @@
...
+                data = Map(
+                  "migrationId" -> migration.id,
+                  "workspace" -> workspace.toWorkspaceName,
+                  "googleProject" -> workspace.googleProjectId,
+                  "bucketName" -> destBucketName
+                )
+              )
+            }
+
+            // Don't need a requester pays project for the bucket in the new google project
```
aherbst-broad: The bit referenced here is forthcoming in a future PR, correct?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/MultiregionalBucketMigrationActorSpec.scala
```diff
@@ -0,0 +1,915 @@
...
+  it should "not prevent a workspace from being deleted if the migration was retried" in
+    runMigrationTest {
+      for {
+        _ <- inTransaction { dataAccess =>
+          createAndScheduleWorkspace(testData.workspace) >>=
+            dataAccess.multiregionalBucketMigrationRetryQuery.getOrCreate
+        }
+        wsService <- MigrateAction.asks(_.workspaceService)
+        _ <- MigrateAction.fromFuture {
+          wsService.deleteWorkspace(testData.workspace.toWorkspaceName)
```
aherbst-broad, line 881: What is the failure mode here? The workspace is locked and therefore undeletable?

aherbst-broad (commented):
Looking good, had some questions.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
marctalbott: :+1: I'll break it out

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
marctalbott: We do need a valid billing account on the workspace so that we can create buckets and perform other GCP operations. The billing project billing account isn't such a big deal and we have some checks in place now that make it safer to assume that the billing project's and workspace's billing accounts are the same. I still felt there wasn't much downside to keeping these extra checks in place for the sake of Good Clean Living™

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/DataAccess.scala
marctalbott: It's not strictly necessary and I did a lot of the unit testing without it in place. I added it as an afterthought for the sake of symmetry. I'm not attached to it though, so happy to remove it since it's unnecessary

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
marctalbott: For PPW, `onlyChild` migrations could run even when the pipeline was rate limited by GCP quotas since `onlyChild` migrations didn't consume any GCP quota. This boolean served two purposes for PPW: (1) ensure that `onlyChild` migrations would still be run when the pipeline was otherwise blocked and (2) to block workspaces that **weren't** `onlyChild` migrations from being selected by `nextMigration`. 

I've repurposed this parameter so that only (2) applies. `rateLimit` will be true if the pipeline is blocked (detected by the presence of certain GCP errors) or if we have reached the limit of concurrent migrations. When `rateLimit` is true, `nextMigration` won't return any workspaces and the pipeline will stop until we are no longer rate limited. 

An alternative I hadn't considered previously would be to check the conditional before calling `nextMigration` and return `None` instead of calling `nextMigration` at all. This alternative feels simpler and clearer to me, so I'm going to make that change

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
marctalbott: The triggers will only update the `UPDATED` column in their tables. Other columns still need to be explicitly updated depending on what step in the pipeline has finished

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
marctalbott: Yeah, this was taken directly. I'm not sure if PPW ran into any issues with this timing out, but I see no reason not to bump it up to 30 seconds to give us some extra breathing room

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
marctalbott: If a bucket is requester pays, you always need to include the Google project to bill your operation to. The Google project to bill isn't needed (and may even be rejected, I'm not sure) if the bucket isn't requester pays.

In this case, we're billing to `broad-dsde-ENV` (which is `googleProjectToBill`) similar to how we're consuming `broad-dsde-ENV`'s quotas. There may be a yet-to-be-written ticket to move this and the quota over to use the user's project

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
marctalbott: Yes, requester pays will be enabled on the bucket as part of [WOR-1098](https://broadworkbench.atlassian.net/browse/WOR-1098).

This comment also needs to be updated as the bucket is in a new _region_, not a new google project

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/MultiregionalBucketMigrationActorSpec.scala
marctalbott: [Relevant ticket](https://broadworkbench.atlassian.net/browse/PPWM-105)

There was a problem in PPW where cascade delete wasn't set up on the FK from the retry table to the migration history table, so workspaces that had retried migrations couldn't be deleted as the FK constraint prevented it

local-dev/templates/docker-rsync-local-rawls.sh
cahrens: I have not.

cahrens (approved):
I can't pretend to understand the Scala in MultiregionalBucketMigrationActor with anything approaching competency, but I'm not concerned given that this code was largely a port from PPW (which worked successfully) and it appears to have strong unit test coverage (not that I can understand much of the unit tests either).

aherbst-broad (approved):

PR #2472, opened 2023-07-31
[WOR-1154] Check for 403 when polling on WSM deletion

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1154)

## Context
The initial cutover of this caused flakiness with our Azure end to end tests, where the deletion result polling would occasionally get back a 403 from WSM. There is a race in WSM during workspace deletion where the Sam resource is deleted before the workspace record is deleted. This opens the possibility of getting a 403 while the deletion job is still in flight. We should guard against this possibility in Rawls.

## This PR
* Adds a check against 403s from the deletion job result endpoint during workspace deletion. Since a 403 here means that the Sam resource is gone and the workspace is effectively unreadable, I think this is a safe assumption to make in Rawls. 
* [This](https://github.com/broadinstitute/rawls/actions/runs/5728207068) is an E2E test run that bumped into the 403 and completed successfully.

---

**PR checklist**


cahrens (approved):
Thanks for getting to the root of this!

marctalbott (approved):

PR #2474, opened 2023-08-01
[WOR-1097] Monitor running STS jobs in multiregional bucket migration pipeline

marctalbott:
Ticket: [WOR-1097](https://broadworkbench.atlassian.net/browse/WOR-1097)
* port over PPW migration code to check the status of STS jobs and update the migration table once they're complete

---

**PR checklist**

[WOR-1097]: https://broadworkbench.atlassian.net/browse/WOR-1097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
@@ -19,14 +19,18 @@ import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SamDAO, SlickDataSource}
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Implicits._
-import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{Failure, Success, toTuple}
+import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils.Outcome.{toTuple, Failure, Success}
 import org.broadinstitute.dsde.rawls.monitor.migration.MigrationUtils._
 import org.broadinstitute.dsde.rawls.monitor.migration.MultiregionalBucketMigrationActor.MigrateAction._
 import org.broadinstitute.dsde.rawls.workspace.WorkspaceService
 import org.broadinstitute.dsde.workbench.google.GoogleIamDAO
 import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.ObjectDeletionOption.DeleteSourceObjectsAfterTransfer
 import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.ObjectOverwriteOption.OverwriteObjectsAlreadyExistingInSink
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/MultiregionalBucketMigrationActorSpec.scala
```diff
@@ -911,4 +911,256 @@ class MultiregionalBucketMigrationActorSpec extends AnyFlatSpecLike with Matcher
...
+          getTransferJobs = inTransaction { _ =>
+            storageTransferJobs
+              .sortBy(_.id.asc)
+              .result
+              .map(_.toList)
+          }
+
+          transferJobsBefore <- getTransferJobs
+
+          _ = Thread.sleep(500)
```
marctalbott: Without this, the test was flaking because `transferJobsMid(0).updated` and `transferJobsBefore(0).updated` were often equal. Adding this sleep has made the test pass consistently. There is another test that I had to do this for that was flaking in the last PR. 

I am sadly not sure why this is happening with these tests, but doesn't appear to be an issue with the PPW tests

aherbst-broad: Interesting. Thinking through this a bit: since the updated field on the transfer jobs table is at best second granularity, if the refresh code updates the jobs in the same second as the previous update it seems likely that the updated timestamps between the before and after lists will be equal. So, a Thread.sleep() here seems reasonable if a bit awkward. Consider bumping the sleep to a full 1000 to avoid any chance they'll be read in the same second.

I am surprised that the PPW tests do not flake in a similar manner, perhaps the refresh in that code always takes > 1 second?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -702,6 +706,145 @@ object MultiregionalBucketMigrationActor {
...
+            IO.pure {
+              Failure(
+                s"""Transfer job "${operation.getTransferJobName}" failed: """ +
+                  s"""operation "${operation.getName}" failed with errors: """ +
+                  operation.getErrorBreakdownsList
+              )
+            }
+          case other =>
+            // shouldn't really happen, but something could have gone wrong in the wire.
+            // report the error and try again later. todo: don't try indefinitely
```
blakery, line 773: In this case, I think manual intervention is probably a reasonable way to stop an indefinite retry.

Looking through the status enum, the most likely to occur is probably `QUEUED`, in which case this would be the correct action.

blakery (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/MultiregionalBucketMigrationActorSpec.scala
marctalbott: Just double checked the PPW tables, and the `updated` column goes up to 6 digits of precision while the multiregional bucket table's `updated` column doesn't. It only really matters for these tests, but I'll look into adding some additional precision to deflake the tests over these sleeps

PR #2482, opened 2023-08-09
[WOR-1061] Add bucket migration APIs for scheduling and fetching past attempts

marctalbott:
Ticket: [WOR-1061](https://broadworkbench.atlassian.net/browse/WOR-1061)
* Add new admin APIs to schedule workspaces for bucket migrations one at a time, in a list, or by billing project. Also add admin APIs to list migration attempts for a single workspace or workspaces in a billing proejct.
* Add new `BucketMigrationService` to handle actual scheduling
* Prevent workspaces from being unlocked in the middle of a bucket migration

---

**PR checklist**

[WOR-1061]: https://broadworkbench.atlassian.net/browse/WOR-1061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -939,7 +939,7 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
-  /api/admin/migrations:
+  /api/admin/workspaces/migrations:
     post:
       tags:
         - admin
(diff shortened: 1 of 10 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -939,7 +939,7 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
-  /api/admin/migrations:
+  /api/admin/workspaces/migrations:
```
marctalbott, line 942: this is getting ripped out soon, but just fixing this to be the actual API

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -967,6 +967,137 @@ paths:
                  $ref: '#/components/schemas/ErrorReport'
         500:
            $ref: '#/components/responses/RawlsInternalError'
+  /api/admin/bucketMigration/workspaces/{workspaceNamespace}/{workspaceName}:
```
marctalbott, line 970: I opted to group all of these APIs under `/bucketMigration`, but would be open to alternative API structures

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1693,7 +1693,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
 
         dataSource.inTransaction { dataAccess =>
           import dataAccess.WorkspaceExtensions
-          dataAccess.workspaceMigrationQuery.isMigrating(workspaceContext).flatMap {
```
marctalbott, line 1696: This has been a dead code path since we began excluding v1 workspaces via `getV2WorkspaceContext` above. We do still want to prevent users from unlocking workspaces while they're in the middle of a bucket migration

core/src/test/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationServiceSpec.scala
```diff
@@ -0,0 +1,260 @@
...
+          }
+          postMigrationAttempts.foreach { case (_, attempts) =>
+            attempts should not be empty
+          }
+        },
+        Duration.Inf
+      )
+  }
+
+  it should "schedule any eligible workspaces in a billing project for migration and return any failures" in withMinimalTestDatabase {
```
marctalbott, line 106: This is a departure from PPW. For PPW, specifying a list of workspaces with an ineligible workspace would fail the entire list. I felt this was likely to give us a headache later (e.g. trying to migrate a billing project with a single locked workspace would be impossible), so we will now migrate all possible workspaces and return an error to the user to let them know that some workspaces failed to migrate.

Alternatively, we could return a successful status code always (or perhaps if at least 1 succeeds) and simply list out the outcomes for each workspace.

```
200
[
  ws1 -> 'Success' (OR normal Metadata response),
  ws2 -> 'Failure', 'error message'
]
```

core/src/main/resources/swagger/api-docs.yaml
cahrens: I think I would prefer to keep it under "workspaces". So  `/api/admin/workspaces/{workspaceNamespace}/{workspaceName}/bucketMigration`.

marctalbott: Okay! And then an equivalent `api/admin/billing/{projectId}/bucketMigration` for the billing project based APIs?

blakery (approved):

core/src/main/resources/swagger/api-docs.yaml
cahrens: Oh, I hadn't gotten any further in the PR, ha! Now that you call out the billing use case, perhaps since the APIs can work either via workspace or billing project it is cleaner to leave it the way you have it with `/api/admin/bucketMigration/`.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -967,6 +967,137 @@ paths:
...
+             'application/json':
+               schema:
+                 $ref: '#/components/schemas/ErrorReport'
+        500:
+           $ref: '#/components/responses/RawlsInternalError'
+  /api/admin/bucketMigration/billing/{projectId}:
+    get:
+      tags:
+        - admin
+      summary: list bucket migration attempts for workspaces in billing project
```
cahrens, line 1054: Nit: GCP workspaces only

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -967,6 +967,137 @@ paths:
...
+        - admin
+      summary: list bucket migration attempts for workspace
+      description: list bucket migration attempts for workspace
+      operationId: listBucketMigrationsForWorkspace
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespacePathParam'
+        - $ref: '#/components/parameters/workspaceNamePathParam'
+      responses:
+        200:
+          description: Successful Request
```
cahrens, line 982: What do you get back if you specify an Azure workspace?

marctalbott: Good question, I hadn't thought to test that out. It would presumably get scheduled and then the migration actor would start trying to migrate it and fail out relatively early, but possibly only after the workspace had been locked. I'll add a check to limit this to Rawls workspaces only and prevent Azure/WSM workspaces from being scheduled at all

core/src/test/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationServiceSpec.scala
cahrens: In this case, what would the result of calling GET on the billing project be? I assume you would get the migration attempts for the ones that started migrating, but what do you get for the ineligible ones?

cahrens: Nevermind, just had to read down further!

cahrens (approved):

PR #2491, opened 2023-08-15
[WOR-1188] Extract workspace aggregation to a separate class

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1188)

## Context
We want to include policies in the workspace listing response. 

For the existing Rawls `getWorkspace`  codepath, we are making two identical calls to WSM's `getWorkspace` API: one so we can determine the cloud platform and one to get policies. This is a redundant 100+ ms roundtrip to WSM. For getting a single workspace, not a huge deal. For listing workspaces, this is expensive.
 
Rather than try to shoehorn a call to the existing `getAzureCloudContextFromWorkspaceManager` method in WorkspaceService, I elected to perform some refactoring with the goal of making this code more focused and testable, and making the error handling clearer.

To that end, I've introduced the idea of an `AggregatedWorkspace`. This is the combination of a "rawls" workspace and any other data from upstream services. For now, the only upstream service is Workspace Manager. I've similarly introduced an `AggregatedWorkspaceService` that isolates the aggregation to a more focused class rather than dumping it into the already huge `WorkspaceService` class.

The `AggregatedWorkspaceService` class exposes a single method for now -- `getAggregatedWorkspace`.  This has the WSM connection and stitching logic. 

An additional goal was to introduce clearer exception handling. I think the upper layers of Rawls should not have to care about handling WSM client API exceptions -- this is a burden for developers to always have to remember to handle 404, 500s, etc. Instead this new service returns more semantically meaningful exceptions that upper layers can handle as they see fit.

## This PR
* Implements the refactoring and associated tests as described above
* Adds a listWorkspace integration test to the E2E tests. 
---

**PR checklist**


automation/src/test/scala/org/broadinstitute/dsde/test/api/WorkspacesAzureApiSpec.scala
```diff
@@ -158,6 +159,32 @@ class AzureWorkspacesSpec extends AnyFlatSpec with Matchers with CleanUp {
     }
   }
 
+  it should "allow listing workspaces" in {
```
aherbst-broad, line 162: New E2E test -- this feels like a pretty critical user journey.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
```diff
@@ -0,0 +1,97 @@
...
+    * @throws AggregateWorkspaceNotFoundException when the source workspace references a workspace manager
+    *                                             record that is not present
+    * @throws WorkspaceAggregationException when an error is encountered pulling data from aggregating upstream systems
+    *                                       
+    * @throws InvalidCloudContextException when an aggregated workspace does not have an Azure cloud context
+    */
+  def getAggregatedWorkspace(workspace: Workspace, ctx: RawlsRequestContext): AggregatedWorkspace = {
+    val span = startSpanWithParent("getWorkspaceFromWorkspaceManager", ctx.tracingSpan.orNull)
+
+    workspace.workspaceType match {
```
aherbst-broad, line 46: This is a pretty close copypasta from [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L455).

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3205,7 +3205,9 @@ class WorkspaceServiceSpec
     }
   }
 
-  "listWorkspaces" should "list the correct cloud platform and state for Azure and Google workspaces" in withTestDataServices {
+  behavior of "listWorkspaces"
```
aherbst-broad, line 3208: Tweak to make running the `listWorkspaces` tests easier in the IDE -- I like right clicking on this top-level parent and running all related tests.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3288,77 +3297,76 @@ class WorkspaceServiceSpec
       ) should contain theSameElementsAs expected
   }
 
-  "listWorkspaces" should "not return a MC workspace that does not have a cloud context" in withTestDataServices {
```
aherbst-broad, line 3291: The rest of the diff for this class is scalafmt.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspace.scala
```diff
@@ -0,0 +1,41 @@
...
+  Workspace,
+  WorkspaceCloudPlatform,
+  WorkspacePolicy,
+  WorkspaceType
+}
+
+/**
+  * Represents the aggregation of a "rawls" workspace with any data from
+  * external sources (i.e,. workspace manager cloud context, policies, etc.)
+  * @param rawlsWorkspace Source rawls worksapce
```
marctalbott: It'd be good to somehow indicate the difference here between `rawlsWorkspace` and `WorkspaceType.RawlsWorkspace` since `rawlsWorkspace` will not always be a `WorkspaceType.RawlsWorkspace`. Maybe `baseWorkspace`?

marctalbott (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
blakery: Hiding whitespace fixes it.

aherbst-broad:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
```diff
@@ -0,0 +1,97 @@
...
+}
+
+import java.util.UUID
+import scala.jdk.CollectionConverters._
+
+/**
+  * Responsible for aggregating workspace data together from various sources (i.e, WSM).
+  * @param workspaceManagerDAO DAO for talking to WSM
+  */
+class AggregatedWorkspaceService(workspaceManagerDAO: WorkspaceManagerDAO) extends LazyLogging {
```
blakery, line 25: I love that this class has a single dependency, which can be easily mocked, and creates a clean separation of concerns.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspace.scala
```diff
@@ -0,0 +1,41 @@
...
+  * @param azureCloudContext Azure cloud context (if present)
+  * @param policies Terra policies
+  */
+case class AggregatedWorkspace(
```
blakery, line 21: I want to say that there could be some sort of re-work of the different workspace objects floating around (`WorkspaceDetails`, `WorkspaceXYZResponse`, etc) instead of creating a new one, but I like this as is because it's 1. internal to the service 2. localized to this part of the code, and 3. has a single clear purpose, instead of trying to over-generalize.

blakery (approved):
Looks like a great step in making this more testable and maintainable!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspace.scala
aherbst-broad: Yeah, I've been ruminating on this as well. My desire is to clean this up a bit as we move into a workspaces v2 api world. I think we do have a good "base" Workspace model, and a need for an aggregated model like this, but the complexity goes up as we have the different models between the list / get responses.

aherbst-broad:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1078,19 +1015,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
                   // be reachable when running against the dev environment.
-                  case ex: ApiException
-                      if (WorkspaceType.McWorkspace == workspace.workspaceType) && (ex.getCode == StatusCodes.NotFound.intValue) =>
+                  case ex: AggregateWorkspaceNotFoundException =>
                     logger.warn(
                       s"MC Workspace ${workspace.name} (${workspace.workspaceIdAsUUID}) does not exist in the current WSM instance. "
                     )
                     None
-                  case ex: InvalidCloudContextException =>
+                  case ex: WorkspaceAggregationException =>
```
cahrens, line 1030: @aherbst-broad @blakery I think we lost some functionality here introduced in https://github.com/broadinstitute/rawls/pull/2464-- this was handling a missing CloudContext during the list to avoid throwing an exception. As far as I can tell, with the refactoring in this PR we will still get an InvalidCloudContextException from AggregatedWorkspaceService, but it will no longer be handled here.

cahrens: Though actually the unit test still existed-- https://github.com/broadinstitute/rawls/pull/2491/files#diff-05872226f80cea9f996d66c8df13e83d05cf72a93a2fec9fa8d5a3e54550f32eR3300. I will try running the test tomorrow and see how this is still being handled.

cahrens: Ah, I see. The `InvalidCloudContextException` ends up getting wrapped in a `WorkspaceAggregationException`.

PR #2492, opened 2023-08-15
WOR-659 Remove v1 billing and workspaces code - Part 1

blakery:
Ticket: [WOR-659](https://broadworkbench.atlassian.net/browse/WOR-659)

In this PR:
* Remove v1 billing endpoints
* remove v1 billing project specific deletion steps
* Rename BillingApiService to ServicePerimeterApiService, to reflect remaining functionality 

---

**PR checklist**

[WOR-659]: https://broadworkbench.atlassian.net/browse/WOR-659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -97,74 +92,6 @@ object UserService {
...
-          _ <- samDAO.deleteResource(SamResourceTypeNames.googleProject,
-                                     projectName.value,
-                                     ctx.copy(userInfo = userInfoForSam)
-          )
-        } yield ()
-      )
-    }
-  }
-
-  private def deletePetsInProject(projectName: GoogleProjectId,
```
blakery, line 147: It took awhile to convince myself that this wasn't needed, particularly since it was checked in some tests for the v2 billing functionality.
Eventually I figured out that it only happens when `samDAO.listResourceChildren` returns projects, which only happens for v1.
So the tests were basically ensuring backwards compatibility with v1 projects.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -137,15 +137,24 @@ class HttpGoogleServicesDAOSpec extends AnyFlatSpec with Matchers with MockitoTe
...
+  /*
+    FIXME: This test has likely been broken for its entire existence, so marking it as ignored for now
+       the test was not waiting for the future returned by the call to googleServicesDAO.setupWorkspace to complete before moving on.
+       So it was failing (when verifying the call to googleStorageService.insertBucket),
+       but the test framework would wait for child processes to complete before reporting.
+       If another test was executing when it was reported, the failure would show up in that test,
+       but with the actual failure point listed as this test, far down in the stack trace (or however deep in the call stack it was when it failed)
+       If no test was active, whether because all tests had completed, or the framework was in between tests, the failure would not be reported at all.
+   */
+  it should "create the workspace's GCS bucket" ignore {
```
blakery, line 149: I could also replicate the behavior on develop, by running all tests in the `dataaccess` package.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3595,94 +3499,6 @@ paths:
...
-          description: You must be a project owner to delete billing project
-          content:
-            'application/json':
-              schema:
-                $ref: '#/components/schemas/ErrorReport'
-        500:
-          $ref: '#/components/responses/RawlsInternalError'
   /api/user/billingAccounts:
     get:
       tags:
```
marctalbott, line 3562: maybe remove the `billing` tag from this API so we can fully remove the `billing` section and just have the `billingV2`?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -723,32 +557,25 @@ class UserService(
...
+        _ <- Future.traverse(policies) { policy =>
+          samDAO
+            .addUserToPolicy(
+              SamResourceTypeNames.billingProject,
+              projectName.value,
+              policy,
+              projectAccessUpdate.email,
+              ctx
+            )
+            .recoverWith { case regrets: Throwable if policy == SamBillingProjectPolicyNames.canComputeUser =>
```
marctalbott: Wondering if we can remove this `recoverWith` block at this point. Older billing projects will still have this policy, but the policy _should_ no longer be in use for any real auth decisions so it may be time to stop supporting changes to the policy and stop adding users to the policy

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -291,207 +284,19 @@ class UserServiceSpec
...
-      val userService = getUserService(dataSource, mockSamDAO)
-
-      val actual = intercept[RawlsExceptionWithErrorReport] {
-        Await.result(userService.deleteBillingProject(defaultBillingProjectName), Duration.Inf)
-      }
-      actual.errorReport.statusCode shouldEqual Option(StatusCodes.Forbidden)
-    }
-  }
 
   it should "not delete the Google project when unregistering a Billing project" in {
```
marctalbott: nit: test should be renamed since there is no longer a Google project to delete when unregistering. Maybe "delete the billing project records from Sam and Rawls when unregistering a billing project"

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -789,16 +786,6 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
   private def isAzureMcWorkspace(maybeMcWorkspace: Option[WorkspaceDescription]): Boolean =
     maybeMcWorkspace.flatMap(mcWorkspace => Option(mcWorkspace.getAzureContext)).isDefined
 
-  // TODO - once workspace migration is complete and there are no more v1 workspaces or v1 billing projects, we can remove this https://broadworkbench.atlassian.net/browse/CA-1118
```
marctalbott: We may need to come up with a plan for the remaining v1 workspaces (~13000) in prod, as this will break workspace deletion for those workspaces and leave billing projects with these workspaces undeleteable. Probably out of scope for this ticket, but just calling it out

marctalbott (approved):
Just a few small comments, but generally looks good. Nice to see all of this deleted!

core/src/main/resources/swagger/api-docs.yaml
blakery: `/api/user/billingAccounts` is still has the `billing` tag.
I guess we could give it the `billingV2` tag, but it looks a bit odd, since it doesn't have v2 in its path, and all the other v2 endpoints do (and it hasn't changed).

marctalbott: This one is technically a user API and doesn't really fit in the `billing` section on its own. So I'd be fine with either moving it to `billingV2` or dropping it from the `billing` sections entirely and just letting it live in the user section

aherbst-broad: I'd prefer to remove the entirety of billingv1 to avoid confusion and leave things tidy. What about leaving it with users (i.e., remove just the billing tag)? It's odd that it was in both.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -409,16 +297,7 @@ class UserService(
                                    3600,
                                    RawlsUserSubjectId("0")
       )
-      for {
-        _ <- deleteGoogleProjectIfChild(projectName,
```
aherbst-broad, line 413: Why is this call not needed anymore?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: Now that I look at it, this recover block doesn't look like a great idea in the first place. If it fails, it still returns success. It logs the error, but you'd have to know to look for it.
I guess the main thing it would do, is make sure the user is added to any remaining policies in the list.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: Let's huddle about this a bit, I want to get a plan in place before we break something here.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: The idea was to not throw on failure because v2 billing projects don't have this policy. We still don't want to throw, but I think we can skip adding users to the `canCompute` billing project policy entirely now. We have [WOR-658](https://broadworkbench.atlassian.net/browse/WOR-658) to clean up the policies on older billing projects

core/src/main/resources/swagger/api-docs.yaml
blakery: I didn't even think about the fact that it already had both tags.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
blakery: `deleteGoogleProjectIfChild` does clean-up that is specific to v1 projects, so it's marked for removal as part of this ticket.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
blakery: Pulling it out of this PR, for now.
The other one is more focused around workspaces, anyway.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAOSpec.scala
aherbst-broad: 👍 This should get a tech debt ticket.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -137,15 +137,24 @@ class HttpGoogleServicesDAOSpec extends AnyFlatSpec with Matchers with MockitoTe
...
     val userInfo = UserInfo(RawlsUserEmail("fake@email.com"),
                             OAuth2BearerToken("some-token"),
                             300,
                             RawlsUserSubjectId("193481341723041"),
                             None
     )
 
-    val googleStorageService = mock[GoogleStorageService[IO]](RETURNS_SMART_NULLS)
+    val googleStorageService = mock[GoogleStorageService[IO]](RETURNS_DEEP_STUBS)
```
aherbst-broad, line 157: Why does this need RETURNS_DEEP_STUBS?

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -125,10 +119,9 @@ class UserServiceSpec
 
       val userService = getUserService(dataSource, servicePerimeterService = mockServicePerimeterService)
 
-      val actual =
+      val actual: Unit =
         userService.addProjectToServicePerimeter(defaultServicePerimeterName, project.projectName).futureValue
-      val expected = ()
-      actual shouldEqual expected
+      actual shouldEqual ()
```
aherbst-broad, line 124: This doesn't seem like a meaningful assertion; if the future did anything else the test would fail anyway wouldn't it?

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ServicePerimeterApiServiceSpec.scala
```diff
@@ -0,0 +1,151 @@
+package org.broadinstitute.dsde.rawls.webservice
```
aherbst-broad, line 1: Glad there is a dedicated spec for this api service now.

aherbst-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAOSpec.scala
blakery: It was just an initial attempt to get the test to pass, without investing multiple (additional) hours on it.
It did make the test get farther, but then it got to some internal methods that were difficult to spy.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
blakery: That's correct. I think I auto-resolved that line from an IntelliJ inspection..

PR #2494, opened 2023-08-16
[WOR-1062] Add bucket migration progress API

marctalbott:
Ticket: [WOR-1062](https://broadworkbench.atlassian.net/browse/WOR-1062)
* when checking STS job progress, also fetch transfer progress stats (objects and bytes transferred so far along with the total amount to be transferred)
    * write these stats to the db regardless of if the operation has finished
* add enum `MultiregionalBucketMigrationStep` of user-facing migration steps to be returned in progress API

---

**PR checklist**

[WOR-1062]: https://broadworkbench.atlassian.net/browse/WOR-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
@@ -63,7 +63,7 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
     for {
       // most recent migration attempt, need the two STS jobs if they exist and need to turn them into STSJobProgress
       attemptOpt <- dataAccess.multiregionalBucketMigrationQuery
-        .getAttempt(UUID.fromString(workspace.workspaceId))
+        .getAttempt(workspace.workspaceIdAsUUID)
         .value
       attempt = attemptOpt.getOrElse(
         throw new RawlsExceptionWithErrorReport(
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
@@ -61,7 +61,6 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
   )(dataAccess: DataAccess): ReadWriteAction[Option[MultiregionalBucketMigrationProgress]] = {
     import dataAccess.driver.api._
     for {
-      // most recent migration attempt, need the two STS jobs if they exist and need to turn them into STSJobProgress
       attemptOpt <- dataAccess.multiregionalBucketMigrationQuery
         .getAttempt(workspace.workspaceIdAsUUID)
         .value
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
@@ -86,13 +86,13 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
           .result
           .headOption
 
-    } yield
-      MultiregionalBucketMigrationProgress(
-          workspace.toWorkspaceName,
-          MultiregionalBucketMigrationStep.fromMultiregionalBucketMigration(attempt),
-          STSJobProgress.fromMultiregionalStorageTransferJob(tempTransferJob),
-          STSJobProgress.fromMultiregionalStorageTransferJob(finalTransferJob)
-        ).some
+    } yield MultiregionalBucketMigrationProgress(
+      workspace.toWorkspaceName,
+      MultiregionalBucketMigrationStep.fromMultiregionalBucketMigration(attempt),
+      attempt.outcome,
+      STSJobProgress.fromMultiregionalStorageTransferJob(tempTransferJob),
+      STSJobProgress.fromMultiregionalStorageTransferJob(finalTransferJob)
+    ).some
   }
 
   def getBucketMigrationAttemptsForWorkspace(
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -21,6 +29,73 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+  ): Future[Map[String, Option[MultiregionalBucketMigrationProgress]]] =
+    asFCAdmin {
+      for {
+        workspaces <- dataSource.inTransaction { dataAccess =>
+          dataAccess.workspaceQuery.listWithBillingProject(billingProjectName)
+        }
+        progress <- workspaces.traverse { workspace =>
+          dataSource
+            .inTransaction(getBucketMigrationProgress(workspace))
+            .recover(_ => None)
```
marctalbott, line 53: Don't want to throw a 404 if a single workspace hasn't been migrated yet, so catch the 404 from `getBucketMigrationProgress` and return None instead

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -47,7 +122,7 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
               dataAccess.multiregionalBucketMigrationQuery.getMigrationAttempts(workspace)
             }
             .map { attempts =>
-              workspace.name -> attempts.mapWithIndex(
+              workspace.toWorkspaceName.toString -> attempts.mapWithIndex(
```
marctalbott, line 125: Sort of unnecessary to include the namespace here since all workspaces are in the same namespace, but I think it's clearer as the full `WorkspaceName` is globally unique while `name` isn't

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -939,73 +939,93 @@ object MultiregionalBucketMigrationActor {
       // Transfer operations are listed after they've been started. For bucket-to-bucket transfers
       // we expect at most one operation. Polling the latest operation will allow for jobs to be
       // restarted in the cloud console.
-      outcome <- liftF {
+      operation <- liftF {
```
marctalbott, line 942: Before, we were only updating the db when the transfer job had completed. This doesn't work if we want transfer stats while the job is in progress, so I've reworked this to support that

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiService.scala
```diff
@@ -13,6 +13,7 @@ import org.broadinstitute.dsde.rawls.bucketMigration.BucketMigrationService
 import org.broadinstitute.dsde.rawls.model.ExecutionJsonSupport._
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport._
 import org.broadinstitute.dsde.rawls.model._
+import org.broadinstitute.dsde.rawls.monitor.migration.MultiregionalBucketMigrationJsonSupport._
```
marctalbott, line 16: This file is best viewed with whitespace hidden. I restructured the akka routes a bit to make them more hierarchical so adding any future routes will be easier

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -96,6 +96,84 @@ object MultiregionalBucketMigrationFailureModes {
...
+object MultiregionalBucketMigrationStep extends Enumeration {
+  type MultiregionalBucketMigrationStep = Value
+  val ScheduledForMigration, PreparingTransferToTempBucket, TransferringToTempBucket, PreparingTransferToFinalBucket,
+    TransferringToFinalBucket, FinishingUp, Finished = Value
+
+  def fromMultiregionalBucketMigration(migration: MultiregionalBucketMigration): MultiregionalBucketMigrationStep =
+    migration match {
+      case m if m.finished.isDefined                                             => Finished
+      case m if m.tmpBucketTransferred.isDefined || m.tmpBucketDeleted.isDefined => FinishingUp
+      case m if m.tmpBucketTransferJobIssued.isDefined                           => TransferringToFinalBucket
```
aherbst-broad, line 108: Is this transferring into or out of the tmp bucket?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -96,6 +96,84 @@ object MultiregionalBucketMigrationFailureModes {
...
+  }
+
+  def fromTransferOperation(operation: TransferOperation): Option[STSJobProgress] =
+    for {
+      op <- Option(operation)
+      counters <- Option(op.getCounters)
+    } yield STSJobProgress(counters.getBytesFoundFromSource,
+                           counters.getBytesCopiedToSink,
+                           counters.getObjectsFoundFromSource,
+                           counters.getObjectsCopiedToSink
```
aherbst-broad: Is it possible for the sdk to give back nulls for these values?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
marctalbott: It looks like it defaults to 0, but better safe than sorry, I'll wrap them in Options just in case

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
marctalbott: Transferring out of the tmp bucket. I find the wording a bit confusing, but you can see an example [here](https://github.com/broadinstitute/rawls/blob/4a3b2e2c86b93ed00d7e439713c763de87508942/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala#L347) in the conditions we use to determine the next step in the migration. We'll only delete the temporary bucket once the `tmpBucketTransferred` col is defined

blakery (approved):

PR #2501, opened 2023-08-21
[WOR-1186] Async workspace deletion API wireup

aherbst-broad:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1186

## Context
This is the first step in the effort to build out asynchronous azure workspace deletion. This builds out the needed API endpoint and service modifications up to the point of depositing a monitor record in the workspace manager monitoring table. [WOR-1193](https://broadworkbench.atlassian.net/browse/WOR-1193) will pick up where this leaves off an implement the backend job to delete a workspace.

## This PR
* Adds a new versioned workspace endpoint:
   ```DELETE /api/workspaces/v2/<namespace>/<name>```
* Adds a new method to the `MultiCloudWorkspaceService`
   * When presented with an "MC" workspace, it creates the needed record on the monitor table and gives back the job ID
   * When presented with a "rawls" workspace, deletes the workspace synchronously and gives back the bucketName.
 * Introduces a `WorkspaceDeletionResult` type. For now this will contain the jobId for the async path. 
    * I've added a `gcpDeletionContext` field that will contain the `bucketName` for some semblance of compatibility with clients. This is not fully backwards compatible with the existing API as the v1 API actually [renders](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala#L291) a user facing message and returns that to the UI. My hunch is to give back just the needed metadata and let the rendering of a user facing string be a UI concern.

---

**PR checklist**

[WOR-1193]: https://broadworkbench.atlassian.net/browse/WOR-1193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

blakery (approved):

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -148,6 +148,19 @@ object WorkspaceVersions {
     )
 }
 
+case class GcpWorkspaceDeletionContext(bucketName: Option[String])
```
marctalbott: `bucketName` can be unwrapped from the `Option`. All GCP workspaces will have buckets, and Azure workspace deletion results won't have a `GcpWorkspaceDeletionContext`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -109,13 +112,62 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+        s"V2 Starting deletion of workspace [workspaceId=${workspace.workspaceId}, name=${workspaceName.name}, billingProject=${workspace.namespace}, user=${ctx.userInfo.userSubjectId.value}]"
+      )
+
+      _ = workspace.state match {
+        case WorkspaceState.Ready => true
+        case WorkspaceState.Deleting =>
+          throw new RawlsExceptionWithErrorReport(
+            ErrorReport(StatusCodes.Conflict, "Workspace is already being deleted.")
+          )
+        case _ =>
```
marctalbott, line 137: Do we need to allow workspaces in `CreateFailed`, `DeleteFailed`, and `UpdateFailed` to be deleted so users can clean up their failed workspaces?

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
blakery: That's a good point.

I'm trying to think through the implications of that. We might need to make some steps more resilient to things like incomplete records or resources, for that to work.
On the other hand, if deleting doesn't work, it just goes from one failed state to another? 

Maybe we should allow this after the 'Nuke it from orbit' plan is in place?

cahrens: I think we should allow deletion for any state besides Deleting.

aherbst-broad: I will update. We don't want to admit transitions from CREATING and UPDATING (per the [design spec](https://broadworkbench.atlassian.net/wiki/spaces/WOR/pages/2819653639/Workspace+Async+Deletion)) so I will fail those out.

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

PR #2505, opened 2023-08-25
[WOR-1207] Only start migrations if workspace bucket is US multiregion

marctalbott:
Ticket: [WOR-1207](https://broadworkbench.atlassian.net/browse/WOR-1207)
* Grab bucket location for all workspaces and only start migration if workspace bucket is in US multiregion. Fail otherwise

---

**PR checklist**

[WOR-1207]: https://broadworkbench.atlassian.net/browse/WOR-1207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
@@ -289,14 +289,11 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
         .as[Int]
         .map(_.head > 0)
 
-    final def scheduleAndGetMetadata: WorkspaceName => ReadWriteAction[MultiregionalBucketMigrationMetadata] =
+    final def scheduleAndGetMetadata: Workspace => ReadWriteAction[MultiregionalBucketMigrationMetadata] =
       (schedule _) >=> getMetadata
 
-    final def schedule(workspaceName: WorkspaceName): ReadWriteAction[Long] =
+    final def schedule(workspace: Workspace): ReadWriteAction[Long] =
       for {
-        workspaceOpt <- workspaceQuery.findByName(workspaceName)
-        workspace <- MonadThrow[ReadWriteAction].fromOption(workspaceOpt, NoSuchWorkspaceException(workspaceName))
-
         maybePastBucketMigration <- getAttempt(workspace.workspaceIdAsUUID).value
         id <- maybePastBucketMigration match {
           case None          => scheduleFirstMigrationAttempt(workspace)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
@@ -3,8 +3,8 @@ package org.broadinstitute.dsde.rawls.monitor.migration
 import akka.http.scaladsl.model.StatusCodes
 import cats.MonadThrow
 import cats.data.OptionT
-import cats.implicits.catsSyntaxFunction1FlatMap
 import com.google.storagetransfer.v1.proto.TransferTypes.TransferOperation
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -289,22 +288,20 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
...
-    final def scheduleAndGetMetadata: WorkspaceName => ReadWriteAction[MultiregionalBucketMigrationMetadata] =
-      (schedule _) >=> getMetadata
+    final def scheduleAndGetMetadata
+      : (Workspace, Option[String]) => ReadWriteAction[MultiregionalBucketMigrationMetadata] =
+      schedule(_, _).flatMap(getMetadata)
 
-    final def schedule(workspaceName: WorkspaceName): ReadWriteAction[Long] =
+    final def schedule(workspace: Workspace, location: Option[String]): ReadWriteAction[Long] =
       for {
-        workspaceOpt <- workspaceQuery.findByName(workspaceName)
```
marctalbott, line 297: We always need to fetch the workspace before trying to schedule it for migration because we need the workspace's bucket name, so `schedule` now takes a `Workspace` and it's up to the `BucketMigrationService` to fetch the workspace first. Has the added benefit of not loading each workspace twice when scheduling all workspaces in a billing project.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigration.scala
```diff
@@ -322,6 +319,15 @@ trait MultiregionalBucketMigrationHistory extends DriverComponent with RawSqlQue
           )
         }
 
+        _ <- MonadThrow[ReadWriteAction].raiseWhen(!location.getOrElse("").equals("US")) {
```
marctalbott, line 322: This check needs to take place only when scheduling a first migration or we would prevent restarts for workspace's that had already had their final workspace bucket created in a single region.

cahrens:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationServiceSpec.scala
```diff
@@ -248,6 +256,42 @@ class BucketMigrationServiceSpec extends AnyFlatSpec with TestDriverComponent {
     exception.errorReport.statusCode shouldBe Some(StatusCodes.BadRequest)
   }
 
+  it should "fail for non-US multiregion buckets" in withMinimalTestDatabase { _ =>
+    val adminService = mockBucketMigrationServiceForAdminUser()
+
+    val nonUsWorkspace = minimalTestData.workspace.copy(name = "nonUsWorkspace",
```
cahrens, line 262: Nit: I would be more comfortable if the bucketName was one of the actual single region names that we expect… just to be very clear that we don't match on it.

cahrens (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationServiceSpec.scala
marctalbott: Agreed! This is just the bucket name though, not the region. The region we're checking is returned by the mocked `Bucket` down on [l. 277](https://github.com/broadinstitute/rawls/pull/2505/files/13393d1253edefffd14aa4311fc0cea74cd75063#diff-cfab8b4fd2e70ce3dea44657883d0898a691834bca3aff618c351909d8576b9cR277) which is one of the regions we expose in the UI (`northamerica-northeast1`)

cahrens: Ah, I was confusing bucketName and region. Thanks!

aherbst-broad (approved):

PR #2512, opened 2023-08-31
AJ-1308 don't import protected snapshots into unprotected workspaces

calypsomatic:
[AJ-1308](https://broadworkbench.atlassian.net/browse/AJ-1308)
Protected snapshots-by-reference should not be allowed in unprotected workspaces.  Future work should rely on terra policy service to identify protected snapshots and workspaces, but to fill in the compliance gap in the short term, this PR checks the protected status of a snapshot by `isSecureMonitoringEnabled` on its root dataset and of the workspace by the `fc-secure` prefix of its bucket.  
In order to check the root dataset of a snapshot, I needed to fetch the snapshot from TDR; this means adding a `DataRepoDAO` to the `SnapshotService`, with associated changes wherever `SnapshotService` is used.

**PR checklist**

- [ ] ~~Make sure Swagger is updated if API changes~~
  - [ ] ~~**...and Orchestration's Swagger too!**~~
- [ ] ~~If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).~~

[AJ-1308]: https://broadworkbench.atlassian.net/browse/AJ-1308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -22,7 +22,6 @@ import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
 
 import java.util.UUID
 import scala.annotation.tailrec
-import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
@@ -80,7 +79,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     if (!workspaceContext.bucketName.startsWith("fc-secure")) {
       // if not, check if snapshot is protected
       val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
-      if (sources.exists(_.getDataset.isSecureMonitoringEnabled)) {
+      if (sources.asScala.exists(_.getDataset.isSecureMonitoringEnabled)) {
         throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
       }
     }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -3,7 +3,8 @@ package org.broadinstitute.dsde.rawls.snapshot
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.workspace.model._
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
 import org.broadinstitute.dsde.rawls.model.{
@@ -13,6 +14,7 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsRequestContext,
   SamWorkspaceActions,
   SnapshotListResponse,
+  Workspace,
   WorkspaceAttributeSpecs,
   WorkspaceName
 }
@@ -20,6 +22,7 @@ import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
 
 import java.util.UUID
 import scala.annotation.tailrec
+import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
@@ -28,16 +31,18 @@ object SnapshotService {
   def constructor(dataSource: SlickDataSource,
                   samDAO: SamDAO,
                   workspaceManagerDAO: WorkspaceManagerDAO,
-                  terraDataRepoUrl: String
+                  terraDataRepoUrl: String,
+                  dataRepoDAO: DataRepoDAO
   )(ctx: RawlsRequestContext)(implicit executionContext: ExecutionContext): SnapshotService =
-    new SnapshotService(ctx, dataSource, samDAO, workspaceManagerDAO, terraDataRepoUrl)
+    new SnapshotService(ctx, dataSource, samDAO, workspaceManagerDAO, terraDataRepoUrl, dataRepoDAO)
 }
 
 class SnapshotService(protected val ctx: RawlsRequestContext,
                       val dataSource: SlickDataSource,
                       val samDAO: SamDAO,
                       workspaceManagerDAO: WorkspaceManagerDAO,
-                      terraDataRepoInstanceName: String
+                      terraDataRepoInstanceName: String,
+                      dataRepoDAO: DataRepoDAO
 )(implicit protected val executionContext: ExecutionContext)
     extends FutureSupport
     with WorkspaceSupport
@@ -53,6 +58,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       if (!workspaceStubExists(wsid, ctx)) {
         workspaceManagerDAO.createWorkspace(wsid, ctx)
       }
+      validateProtectedStatus(workspaceContext, snapshot)
       // create the requested snapshot reference
       val snapshotRef = workspaceManagerDAO.createDataRepoSnapshotReference(wsid,
                                                                             snapshot.snapshotId,
@@ -63,6 +69,19 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
                                                                             ctx
       )
       Future.successful(snapshotRef)
+    }
+
+  def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
+    // logically it might make more sense to check if the snapshot is protected before the workspace
+    // but that is a more expensive check
+    // check if workspace is protected
+    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
+      // if not, check if snapshot is protected
+      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
+      sources.filter(source => source.getDataset.isSecureMonitoringEnabled) match {
+        case Nil => // this is fine
+        case _   => throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
+      }
     }
 
   def getSnapshot(workspaceName: WorkspaceName, referenceId: String): Future[DataRepoSnapshotResource] = {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -78,9 +78,9 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     if (!workspaceContext.bucketName.startsWith("fc-secure")) {
       // if not, check if snapshot is protected
       val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
-      sources.filter(source => source.getDataset.isSecureMonitoringEnabled) match {
-        case Nil => // this is fine
-        case _   => throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
+      if(sources.exists(_.getDataset.isSecureMonitoringEnabled)) {
+        throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
+      }
       }
     }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -71,6 +71,8 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       Future.successful(snapshotRef)
     }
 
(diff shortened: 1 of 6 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -2086,14 +2087,25 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
       WorkspaceType.RawlsWorkspace,
       WorkspaceState.Ready
     )
-
+    val adWorkspace = Workspace(
```
calypsomatic: I was hesitant to modify any existing test workspaces in case it would break any tests, so I've created a new one here.  I did see at least one existing test workspace somewhere that seemed to be representing a workspace with an authorization domain, so if it makes more sense to simply change and use that one, I could do that instead.

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockDataRepoDAO.scala
```diff
@@ -14,6 +14,7 @@ class MockDataRepoDAO(instanceName: String) extends DataRepoDAO {
     snap.id(snapshotId)
     snap.name("snapshotName")
     snap.description("snapshotDescription")
+    snap.source(java.util.List.of(new SnapshotSourceModel().dataset(new DatasetSummaryModel())))
```
calypsomatic, line 17: Necessary since now all tests that use `getSnapshot` will be checking the dataset.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -65,6 +71,19 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       Future.successful(snapshotRef)
     }
 
+  def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
+    // logically it might make more sense to check if the snapshot is protected before the workspace
+    // but that is a more expensive check
+    // check if workspace is protected
+    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
```
dvoet, line 79: Relying on the bucket name makes me queasy, but I think that ship has sailed.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -65,6 +71,19 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+      // if not, check if snapshot is protected
+      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
+      sources.filter(source => source.getDataset.isSecureMonitoringEnabled) match {
+        case Nil => // this is fine
+        case _   => throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
```
calypsomatic: What is the appropriate exception to throw here?

dvoet:
```suggestion
      if(sources.exists(_.getDataset.isSecureMonitoringEnabled)) {
        throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
      }
```

dvoet (approved):
functionality looks good, 1 style suggestion

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
calypsomatic: Same

jladieu: Possibly include a comment explaining that this is a bit of a hack and that this responsibility should be moved to Terra policy service when it's ready?  Bonus points for including a ticket if that work is tracked somewhere.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -65,6 +71,19 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+    // logically it might make more sense to check if the snapshot is protected before the workspace
+    // but that is a more expensive check
+    // check if workspace is protected
+    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
+      // if not, check if snapshot is protected
+      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
+      if(sources.exists(_.getDataset.isSecureMonitoringEnabled)) {
+        throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
+      }
+      }
```
jladieu: maybe a nit: but the close curlies look a bit goofy to me...is this a formatting fix, or an intentional Scala thing?

calypsomatic: Not a nit!  This actually is the cause of everything breaking all of a sudden!

nawatts: Seconded.

nawatts (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
jladieu: I think the important thing about setting up a new sample in the test fixture is to have its name make clear why it's new/different from the other ones, and have the minimal setup required to distinguish how it's different from the default.

1) Naming: In this case I'm not sure what `ad` prefix means, but I guess I can infer from your comment that it means authorization domain...possibly make the name a bit more verbose like: `workspaceWithAuthorizationDomain` will help for a reader that isn't familiar with the ad shorthand.

2) Setup: I can't really tell from the setup code which traits of the new sample workspace make it have an authorization domain?  Ideally, the pattern would be (not literally, because I have no idea how to do this in Scala 😆 :
```
val workspaceWithAuthorizationDomain = <setupDefaultWorkspace>.onlySpecifyOverridesThatAreVerySpecificToAuthorizationDomain()
```

jladieu (approved):
Looks like Scala to me!

I like the test coverage.

I had a couple thoughts/questions that aren't language specific, but feel free to disregard if they don't make sense.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
cahrens: Can we also do the TDR check here for Azure protected workspaces? Or does that happen elsewhere (@aherbst-broad)?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
cahrens: I agree that "adWorkspace" meant nothing to me.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
aherbst-broad: @cahrens not sure I follow -- WSM is the service that enforces policy for Azure protected workspaces. I am wondering if there is a way to perform this check in WSM today since these are fundamentally "rawls stage" WSM workspaces that are holding the snapshots. Fine for now though.

aherbst-broad (approved):
lgtm

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
dvoet: @aherbst-broad that is actually a very interesting point. This code path is already creating a snapshot ref in WSM which looks like it _should_ be doing the policy check which should hit [this error](https://github.com/DataBiosphere/terra-workspace-manager/blob/bb07fcb5d4f080604683aadd3670601d9a88a4e4/service/src/main/java/bio/terra/workspace/service/policy/PolicyValidator.java#L100C70-L100C70). But it is my understanding that importing by reference is rare with rawls workspaces and it is really the import by copy operation that needs to be protected. @calypsomatic, do we know that importing by reference is actually a problem?

dvoet (changes requested):
not a change per se, but I want to make sure we know what we are doing

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
dvoet: Perhaps the workspace created does not have a cloud context and therefore no cloud platform to check and so the check passes?

calypsomatic: Snapshot-by-reference into Azure workspaces doesn't work at all right now - it doesn't fail gracefully but it does fail, which is at least compliant and something we intend to implement in the future.  Snapshot by reference is pretty rare and not a big problem, but we deemed it still worth covering the compliance gap.  More information in [this doc](https://docs.google.com/document/d/1lxmWErJczgdCZabGRNFQDD2We6afe4YhdV4HF2bHLGQ/edit#heading=h.v9bx55hrki7h).

calypsomatic: Comparing trying to import into an azure workspace before vs after my change: previously, it failed with the error "Workspace landing zone type [CromwellBaseResourcesFactory] does not support protected data" - but now seems like it hits my check first and gives "Unable to add protected snapshot to unprotected workspace".  If anyone here has knowledge or intuition about how desirable/undesirable this effect is, I'm willing to hear it - I can also check with product next week when everyone's back!

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/TestDriverComponent.scala
```diff
@@ -2086,14 +2087,25 @@ trait TestDriverComponent extends DriverComponent with DataAccess with DefaultIn
...
+      Map.empty
+    )
     override def save() =
       DBIO.seq(
         workspaceQuery.createOrUpdate(workspace),
         workspaceQuery.createOrUpdate(workspace2),
         // TODO (CA-1235): Remove these after PPW Migration is complete
         workspaceQuery.createOrUpdate(v1Workspace),
         workspaceQuery.createOrUpdate(v1Workspace2),
+        workspaceQuery.createOrUpdate(adWorkspace),
```
davidangb: Adding the new `adWorkspace` here, inside `MinimalTestData`, means that any unit test that relies on `withMinimalTestDatabase` will save one more workspace to the db and have one more workspace to clean up when done. This just a few more rows, but it's extra overhead for a BUNCH of tests that don't need it, and Rawls unit tests are already slow. Can you add this extra workspace just for the test(s) that actually need it?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -65,6 +71,20 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+  // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
+  // We'll have to use this workaround for identifying protected status
+  def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
+    // logically it might make more sense to check if the snapshot is protected before the workspace
+    // but that is a more expensive check
+    // check if workspace is protected
+    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
+      // if not, check if snapshot is protected
+      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
+      if (sources.exists(_.getDataset.isSecureMonitoringEnabled)) {
```
davidangb: Because this class already imports `import scala.jdk.CollectionConverters._`, if you change `if (sources.exists` to `if (sources.asScala.exists`, you can eliminate the deprecated `import scala.collection.convert.ImplicitConversions ...` line.

davidangb (commented):
a couple comments inline - I realize this may be moot if you reimplement this PR in WSM.

dvoet (approved):

davidangb (approved):
nice, thank you!

calypsomatic:
jenkins retest

jladieu (approved):

calypsomatic:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -65,6 +70,20 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+  // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
+  // We'll have to use this workaround for identifying protected status
+  def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
+    // logically it might make more sense to check if the snapshot is protected before the workspace
+    // but that is a more expensive check
+    // check if workspace is protected
+    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
+      // if not, check if snapshot is protected
+      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
+      if (sources.asScala.exists(_.getDataset.isSecureMonitoringEnabled)) {
```
pshapiro4broad, line 82: It's been a while since I've looked at scala but does this mean that a snapshot without any sources will not check this flag? Under what circumstances would a snapshot's sources be empty?

The only way I can see that sources would be missing is if the `include` param to the web API call didn't include `SOURCES`. Right now rawls is depending on the default API behavior to provide this field. It seems like the code should fail if that field is missing, since that means something has gone wrong.

https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/api/data-repository-openapi.yaml#L642

PR #2519, opened 2023-09-05
[WM-2064] Return 400 when method request is invalid for /inputsOutputs 

salonishah11:
Ticket: https://broadworkbench.atlassian.net/browse/WM-2064

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
```diff
@@ -70,14 +70,15 @@ object WorkspaceManagerResourceMonitorRecord {
...
-                          ): WorkspaceManagerResourceMonitorRecord = {
+  ): WorkspaceManagerResourceMonitorRecord =
     WorkspaceManagerResourceMonitorRecord(
       jobRecordId,
       JobType.WorkspaceDelete,
       workspaceId = Some(workspaceId),
-      billingProjectId = None, userEmail = Some(userEmail.value), createdTime = Timestamp.from(Instant.now())
+      billingProjectId = None,
+      userEmail = Some(userEmail.value),
+      createdTime = Timestamp.from(Instant.now())
```
salonishah11, line 80: Edited by `sbt scalafmtAll`. Without this Github formatter check fails.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -129,7 +129,9 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       )
 
       _ = workspace.state match {
-        case WorkspaceState.Ready | WorkspaceState.CreateFailed | WorkspaceState.DeleteFailed | WorkspaceState.UpdateFailed => true
+        case WorkspaceState.Ready | WorkspaceState.CreateFailed | WorkspaceState.DeleteFailed |
+            WorkspaceState.UpdateFailed =>
+          true
```
salonishah11, line 134: Edited by `sbt scalafmtAll`. Without this Github formatter check fails.

dillydally414 (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
```diff
@@ -62,6 +62,18 @@ trait MethodWiths {
         )
     }
 
+    method.validate match {
+      case Some(_) => fetchMethod
+      case None =>
+        DBIO.failed(
+          new RawlsExceptionWithErrorReport(
+            errorReport =
+              ErrorReport(StatusCodes.BadRequest, s"Invalid method request. One or more values in request are invalid.")
```
cjllanwarne: Feels like we're duplicating the dictionary definition of a 400 response in the message.

Since we haven't touched user data yet, just validating the request itself, is it valid to tell the user which field is bad? Or is that a bigger change? Is this a standard message or one you made up? I might lean towards leaving it empty or simplifying to "invalid request" or "invalid parameters"

cjllanwarne (approved):
LGTM. One minor quibble about the wording but not a blocker

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
salonishah11: No this isn't a standard message. It is hard to specify which fields are invalid here because `MethodRepoMethod` can be one of `DockstoreMethod`, `DockstoreToolsMethod` or `AgoraMethod` and AgoraMethod has different fields from the Dockstore methods. I was a little hesitant to change `.validate` method to return which fields are invalid as it would have a wider change effect. I can leave the error message as `Invalid request` as that would be easier.

salonishah11:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSupport.scala
cjllanwarne: Sounds good to me 👍

PR #2520, opened 2023-09-05
WOR-1206 Remove v1 workspaces code

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1206

also includes https://broadworkbench.atlassian.net/browse/WOR-662

Note that while the PPW change sets were removed from slick updates, the tables themselves are not being removed at this point. They simply won't be re-checked on db updates, and won't be re-run for a new database.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -258,7 +258,7 @@ trait WorkspaceComponent {
       loadWorkspaces(workspaceQuery)
 
     def listWithBillingProject(billingProject: RawlsBillingProjectName): ReadAction[Seq[Workspace]] =
-      workspaceQuery.withBillingProject(billingProject).read
+      workspaceQuery.withBillingProject(billingProject).withVersion(WorkspaceVersions.V2).read
```
blakery: The filtering was being done inline where this was being called, so it seemed easier to just put this here instead.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -298,10 +298,7 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
   }
 
   def determineCromwellBackendFromWorkspaceVersion(workspaceVersion: WorkspaceVersion): CromwellBackend =
-    workspaceVersion match {
-      case WorkspaceVersions.V1 => defaultNetworkCromwellBackend
-      case WorkspaceVersions.V2 => highSecurityNetworkCromwellBackend
-    }
+    highSecurityNetworkCromwellBackend
```
blakery: I think we could keep going here, and remove this method and remove defaultNetworkCromwellBackend from this classes parameters. 
But I was wary of how far the changes for that might stretch.

If anyone feels strongly about it, I can clean it up with a follow-up PR.

aherbst-broad (commented):
Is liquibase ok with these migrations being moved? What happens when liquibase runs where there are changesets that no longer exist?

blakery:
> Is liquibase ok with these migrations being moved? What happens when liquibase runs where there are changesets that no longer exist?

Liquibase doesn't care if there is a changeset for a given changelog row.
[Documentation link](https://docs.liquibase.com/workflows/liquibase-community/trimming-changelogs.html#:~:text=Removing%20unneeded%20changesets%20is,s%20and%20you%20are%20done)

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
```diff
@@ -98,15 +98,7 @@
     <include file="changesets/20211006_add_sharded_column.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20210924_populate_sharded_entity_tables.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20211117_drop_unused.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20211115_v1_workspace_migration_history.xml" relativeToChangelogFile="true"/>
```
marctalbott, line 101: This won't remove the tables from existing DBs, is that the intention?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
```diff
@@ -98,23 +98,14 @@ class BillingRepository(dataSource: SlickDataSource) {
...
               StatusCodes.BadRequest,
               "Project cannot be deleted because it contains workspaces."
             )
           )
         }
-        v1Workspaces map { _ =>
-          throw new RawlsExceptionWithErrorReport(
-            ErrorReport(
-              StatusCodes.BadRequest,
-              "Project cannot be deleted because it contains v1 workspaces. Contact support for help."
```
marctalbott, line 114: Will the DB allow the billing project record to be deleted with v1 workspace records still referencing it?

marctalbott: It's likely that Sam will prevent this as well as the `billing-project` `owner` policy group is a member of the `project-owner` `workspace` policy group.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
marctalbott: How about removing this method and leaving the rest?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -188,28 +188,15 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
                                                    billingProjectSyncOutcome: Outcome
```
marctalbott: `billingProjectSyncOutcome` is no longer used

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/FailureModes.scala
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.dsde.rawls.monitor.migration
+
+object FailureModes {
```
marctalbott, line 3: I'll remove this entirely in the next bucket migration PR -- there is a duplicated set of `MultiregionalBucketMigrationFailureModes` that I neglected to actually switch over to

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -638,7 +637,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       )
 
       _ <- traceWithParent("requesterPaysSetupService.revokeAllUsersFromWorkspace", parentContext)(_ =>
```
aherbst-broad: Update this trace to match the new method name

aherbst-broad (approved):
Happy to see the deletion of the old codepaths.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
blakery: That's correct: I thought we decided we weren't going to worry about deleting the actual database tables right now.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingRepository.scala
blakery: Bah: you're right...
When I pulled out this part, I was thinking it would still get the workspaces, and still error.
But I also changed the query to only return v2 workspaces.
This is the one place where we still need to get v1 workspaces.

blakery: I removed the version from the query, so this should still throw on v1 workspaces, now (I never changed the places that were filtering them out manually anyway).

marctalbott (approved):

PR #2531, opened 2023-09-12
ID-784 Rawls Exercises User Bucket Permissions for Access Checks

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-784

`checkWorkspaceCloudPermissions`, which is called by Terra UI when loading a workspace, is supposed to tell the UI if the user has all the appropriate permissions. However, due to permissions propagation issues, the permissions google _says_ the user has and the permissions the user _actually_ has can be different! So, I've inserted a block that gets the workspace bucket location into the method to actually exercise the `storage.buckets.get` permission, and if that call fails, vend a FastPass Grant to the user. This is what the UI does currently, and just does so repeatedly while waiting for permissions to propagate. If we vend a FastPass Grant in Rawls on the first instance of this check, we can severely cut down how long user are waiting for permissions.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -123,6 +123,8 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
 
   val SingleRegionLocationType: String = "region"
 
+  val REQUESTER_PAYS_ERROR_SUBSTRINGS = Seq("requester pays", "UserProjectMissing")
+
   override def updateBucketIam(bucketName: GcsBucketName,
                                policyGroupsByAccessLevel: Map[WorkspaceAccessLevel, WorkbenchEmail],
                                userProject: Option[GoogleProjectId]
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -621,6 +621,26 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
...
 
+  def testSAGoogleBucketGetLocation(googleProject: GoogleProject, bucketName: GcsBucketName, saKey: String)(implicit
+    executionContext: ExecutionContext
+  ): Future[Option[String]] = {
+    implicit val async = IO.asyncForIO
+    val credentials = ServiceAccountCredentials.fromStream(new ByteArrayInputStream(saKey.getBytes))
+    val storageServiceResource = GoogleStorageService.fromCredentials(credentials)
+    storageServiceResource
+      .use { storageService =>
+        storageService.getBucket(googleProject, bucketName)
```
marctalbott, line 637: You need to provide a `BucketGetOption` with the `userProject` set if it's requester pays or this will fail

tlangs: How would we know which user-project to use? Instead, I'm thinking that we do what the UI does [here](https://github.com/DataBiosphere/terra-ui/blob/31ed64d4f3f3261cd69aeb1d4a987f1ea7e16381/src/libs/ajax/ajax-common.ts#L100) and just exclude requester-pays errors from triggering the FastPass functionality.

marctalbott (approved):

Ghost-in-a-Jar (approved):

PR #2534, opened 2023-09-15
AJ-908: validate internal ordering of db rows in entityQuery

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-908

Context:
* We are hoping to reduce instances of heavy/expensive garbage collection pauses in Rawls.
* We see that these garbage collections predominantly happen a few minutes after the hour - 3:04, 6:03, 7:03, 11:05, etc.
* We see that there exists more than one user who is regularly making multiple calls to the [`entityQuery`](https://rawls.dsde-dev.broadinstitute.org/#/entities/entityQuery) API on the hour (this must be a cron job).
* We hope that reducing memory consumption of the `entityQuery` API will alleviate these garbage collection pauses.
* We hope that moving the `entityQuery` API to a steaming model instead of fully materializing entire result sets will reduce its memory consumption.

We previously moved the [`list_entities`](https://rawls.dsde-dev.broadinstitute.org/#/entities/list_entities) API to a streaming model in #2247. The key to moving the API to a stream is ensuring that rows are returned from the DB to Rawls in the correct order; else, the stream results will be invalid.

The `entityQuery` API is much more involved than the `list_entities` API. Its SQL queries are complex and difficult to reason about as a human. While I _think_ that the rows are currently returned in an order that is valid for streaming, I am not 100% confident.

This PR adds validation of those results. It inspects the ordering of db rows for the `entityQuery` API, then logs either "entityAttributeRecords is properly ordered? TRUE" or "entityAttributeRecords is properly ordered? FALSE".

My proposal is to merge this PR and let it run in prod for 1-2 weeks and use log analysis to check for any database result sets that are not properly ordered. If we find that 100% of the result sets are in the correct order, I'll have a lot more confidence in moving the API to streaming.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -1615,11 +1615,6 @@ trait EntityComponent {
             s"entityAttributeRecords is properly ordered? FALSE; length: ${entityAttributeRecords.length}; " +
               s"workspaceId: ${rec.workspaceId}; entityType: ${rec.entityType}; first offender: ${rec.name}"
           )
-          entityAttributeRecords foreach { eaar =>
-            logger.info(s"$eaar")
-          }
-          // TODO: don't actually throw an error; for this commit, we're throwing to see if any unit tests fail
-          throw new RawlsException("EntityAndAttributesResult ordering problem")
         }
       }
       // TEMP: end validation
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -391,7 +391,7 @@ trait EntityComponent {
         * @param sortFieldName
         * @return
         */
-      private def pageOfEntityIdsSelectQuery(workspaceId: UUID, entityType: String, sortFieldName: String) = {
+      private def paginationSubquery(workspaceId: UUID, entityType: String, sortFieldName: String) = {
         val shardId = determineShard(workspaceId)
 
         val (sortColumns, sortJoin) = sortFieldName match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1563,13 +1564,61 @@ trait EntityComponent {
...
-      entityAttributeRecords: Seq[EntityAndAttributesResult]
+      entityAttributeRecords: Seq[EntityAndAttributesResult],
+      validateOrdering: Boolean = false
     ): Seq[Entity] =
-      unmarshalEntitiesWithIds(entityAttributeRecords).map { case (_, entity) => entity }
+      unmarshalEntitiesWithIds(entityAttributeRecords, validateOrdering).map { case (_, entity) => entity }
+
+    // tail-recursive method that validates if all entity ids are contiguous within the Seq[EntityAndAttributesResult]
+    // an order of 3,3,3,2,2,2,1,1 is ok
+    // an order of 1,2,3,1 is bad, because the 1 is repeated non-contiguously
```
jladieu, line 1574: ~can you clarify the return type in the comment too?  I can't quickly tell if a non-empty return value means stuff is good or bad here :)~

Oops, got to the tests later on, which answered this for me.  I still think a comment right here might be helpful though.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -1563,13 +1564,61 @@ trait EntityComponent {
...
+      // short-circuit if empty.
+      if (entityAttributeRecords.isEmpty) {
+        return Seq.empty[(Long, Entity)]
+      }
+
+      if (validateOrdering) {
+        val isProperlyOrdered: Option[EntityRecord] =
+          validateEntityIdOrdering(-1, entityAttributeRecords.head, Set.empty[Long], entityAttributeRecords.tail)
+
+        if (isProperlyOrdered.isEmpty) {
```
jladieu: This led me to ask the question about return type earlier... I am confused that we're looking for `isProperlyOrdered` to be empty, but then printing a log message that says properly ordered `TRUE` -- seems like it should be the opposite case?

jladieu: Possibly rename this to `isUnordered`?

calypsomatic (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: Good feedback, added comments!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
davidangb: I've updated the variable name here, good call

davidangb:
[Test Result](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/66961/testReport/) (1 failure / +1)
[org.broadinstitute.dsde.test.api.WorkspaceApiSpec.Rawls should add workspace Google project to billing project's service perimeter](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/66961/testReport/junit/org.broadinstitute.dsde.test.api/WorkspaceApiSpec/Rawls_should_add_workspace_Google_project_to_billing_project_s_service_perimeter/)

Jenkins retest

davidangb:
fiab-start failure. Jenkins retest.

jladieu (approved):

PR #2550, opened 2023-09-25
[WOR-1259] Bucket migration list workspaces API

marctalbott:
Ticket: [WOR-1259](https://broadworkbench.atlassian.net/browse/WOR-1259)
* Add API to fetch workspaces for display in UI bucket migrations page
* Include unmigrated workspaces with US multiregion bucket, workspaces currently migrating, and workspaces that have successfully migrated in the past 7 days or that have failed migration.

---

**PR checklist**

[WOR-1259]: https://broadworkbench.atlassian.net/browse/WOR-1259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3936,7 +3936,7 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/NotificationType'
-  /api/workspaces/v2:
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}:
     delete:
       tags:
         - workspaces_v2
@@ -3974,6 +3974,33 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         404:
           description: Workspace not found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/bucketMigration:
+    get:
+      tags:
+        - workspaces_v2
+      summary: get workspaces eligible for bucket migration and their migration progress if it exists
+      description: get workspaces eligible for bucket migration and their migration progress if it exists
+      operationId: getEligibleWorkspacesForBucketMigrationWithProgress
+      responses:
+        200:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  type: object
+                  description: workspaceNamespace/workspaceName -> bucket migration progress
+                  additionalProperties:
+                    $ref: '#/components/schemas/MultiregionalBucketMigrationProgress'
+        403:
+          description: You must be an admin to list migration attempts
           content:
             'application/json':
               schema:
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3999,12 +3999,6 @@ paths:
                   description: workspaceNamespace/workspaceName -> bucket migration progress
                   additionalProperties:
                     $ref: '#/components/schemas/MultiregionalBucketMigrationProgress'
-        403:
-          description: You must be an admin to list migration attempts
-          content:
-            'application/json':
-              schema:
-                $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
 components:
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
@@ -10,6 +10,8 @@ import org.broadinstitute.dsde.rawls.model.{
   ErrorReport,
   RawlsBillingProjectName,
   RawlsRequestContext,
+  SamResourceTypeNames,
+  SamWorkspaceRoles,
   Workspace,
   WorkspaceName
 }
@@ -18,7 +20,11 @@ import org.broadinstitute.dsde.rawls.monitor.migration._
 import org.broadinstitute.dsde.rawls.util.{RoleSupport, WorkspaceSupport}
 import org.broadinstitute.dsde.rawls.{NoSuchWorkspaceException, RawlsExceptionWithErrorReport}
 
+import java.sql.Timestamp
+import java.time.Instant
+import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO, val gcsDAO: GoogleServicesDAO)(
   val ctx: RawlsRequestContext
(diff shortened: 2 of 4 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3936,7 +3936,7 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/NotificationType'
-  /api/workspaces/v2:
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}:
```
aherbst-broad, line 3939: Good catch

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +3980,33 @@ paths:
...
+            'application/json':
+              schema:
+                type: array
+                items:
+                  type: object
+                  description: workspaceNamespace/workspaceName -> bucket migration progress
+                  additionalProperties:
+                    $ref: '#/components/schemas/MultiregionalBucketMigrationProgress'
+        403:
+          description: You must be an admin to list migration attempts
```
aherbst-broad: Is this true? I thought this is a non-admin endpoint.

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +33,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
     with WorkspaceSupport
     with LazyLogging {
 
+  def getEligibleOrMigratingWorkspaces: Future[Map[String, Option[MultiregionalBucketMigrationProgress]]] = {
+    def getOwnedWorkspaces: Future[Seq[Workspace]] = for {
```
aherbst-broad, line 38: 🌱 This would be useful in the workspacerepository. Might be out of scope for now because I think it's best to keep things consistent with the idioms of this class.

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +33,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+        .flatMap(_.toOption)
+      workspaces <- dataSource.inTransaction(_.workspaceQuery.listV2WorkspacesByIds(workspaceIds))
+    } yield workspaces
+
+    def getEligibleUnmigratedWorkspaces(
+      workspaces: Seq[Workspace]
+    ): Future[Seq[(Workspace, Option[MultiregionalBucketMigrationProgress])]] =
+      workspaces
+        .traverse { workspace =>
+          getBucketLocation(workspace).map(workspace -> _)
```
aherbst-broad, line 53: Any danger of us 429ing google if there are many workspaces in this list?

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +33,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+        .traverse {
+          case (workspace, progress @ Some(MultiregionalBucketMigrationProgress(_, Some(outcome), _, _))) =>
+            dataSource
+              .inTransaction(_.multiregionalBucketMigrationQuery.getMigrationAttempts(workspace))
+              .map { migrations =>
+                if (
+                  migrations.exists(
+                    _.finished
+                      .getOrElse(Timestamp.from(Instant.MIN))
+                      .after(Timestamp.from(Instant.now().minusSeconds(86400 * 7))) || outcome.isFailure
```
aherbst-broad: Consider expressing this in units of days, second arithmetic is always wonky with things like DST etc

or

Extract the magic 86400 * 7 number to a constant

core/src/test/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationServiceSpec.scala
```diff
@@ -491,6 +500,135 @@ class BucketMigrationServiceSpec extends AnyFlatSpec with TestDriverComponent {
...
+              "reason the migration failed".some
+            )
+          } yield ()
+        },
+        Duration.Inf
+      )
+
+      val res = Await.result(bucketMigrationService.getEligibleOrMigratingWorkspaces, Duration.Inf)
+      res.map {
+        case (name, None) if name.equals(usWorkspace.toWorkspaceName.toString) =>
```
aherbst-broad: 🤔 It took me a bit to grok this assertion block. It might be clearer to express this as explicit gets against the map rather than a iteration over the map which is what this appears to be.

aherbst-broad (approved):

core/src/main/resources/swagger/api-docs.yaml
marctalbott: You're right, forgot to change this when copying this over from another endpoint

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: GCS docs say we're safe - https://cloud.google.com/storage/quotas

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: I did think about incorporating the WorkspaceRepository into this service, but decided to hold off for now. In general though, would we want something like `getOwnedWorkspaces` in the WorkspaceRepository given it makes external requests to Sam?

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +34,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
     with WorkspaceSupport
     with LazyLogging {
 
+  def getEligibleOrMigratingWorkspaces: Future[Map[String, Option[MultiregionalBucketMigrationProgress]]] = {
+    def getOwnedWorkspaces: Future[Seq[Workspace]] = for {
+      userWorkspaces <- samDAO.listUserResources(SamResourceTypeNames.workspace, ctx)
+      workspaceIds = userWorkspaces
+        .filter(_.hasRole(SamWorkspaceRoles.owner))
```
cahrens, line 41: Following up on the discussion in https://github.com/broadinstitute/sam/pull/1183#issuecomment-1721434522… should this be checking for an "own" action instead of an "owner" role? I'm still trying to work through what one should infer outside of Sam with respect to roles.

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +34,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+
+      // get migration progress if it exists
+      workspacesWithProgressOpt <- workspaces.traverse { workspace =>
+        dataSource
+          .inTransaction(getBucketMigrationProgress(workspace))
+          .recover(_ => None)
+          .map(workspace -> _)
+      }
+
+      (scheduledWorkspaces, unscheduledWorkspaces) = workspacesWithProgressOpt.partition(_._2.isDefined)
```
cahrens, line 102: Can you add a comment about what field this is partitioning on?

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,6 +34,77 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+
+      eligibleWorkspaces <- getEligibleUnmigratedWorkspaces(unscheduledWorkspaces.map(_._1))
```
cahrens: Also this field.

cahrens: Actually, why doesn't `getEligibleUnmigratedWorkspaces` return only workspace? You don't use the empty  `Option.empty[MultiregionalBucketMigrationProgress]`, do you?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +3980,27 @@ paths:
...
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  type: object
+                  description: workspaceNamespace/workspaceName -> bucket migration progress
+                  additionalProperties:
+                    $ref: '#/components/schemas/MultiregionalBucketMigrationProgress'
```
cahrens, line 4001: So this progress reflects only the last migration attempt? (I'm reviewing the other PR and realize that one get can get an array of `WorkspaceMigrationMetadata` per workspace)

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: It should be yes, this directly contradicts what I said in that thread by relying on the role name rather than an explicit action.

This API lists the `workspace` resources that the user has access to along with the roles and non-default actions that they have. The `own` action is granted by default in the `owner` role, which means it is not present in this API response. The only way to check if the user has the `own` action is to query each `workspace` resource individually (either by [listing the actions they have](https://sam.dsde-dev.broadinstitute.org/#/Resources/resourceActionsV2) or by [explicitly checking for the `own` action](https://sam.dsde-dev.broadinstitute.org/#/Resources/resourcePermissionV2)). 

This is the lazy and technically incorrect way of doing this permissions check (although relatively safe unless someone removes the `own` action from the `owner` workspace role). I could add the additional authz checks, but I'd say that in this case it's not a huge deal since this is just reading the workspaces and we'll check for the `own` action before actually allowing migration.

cahrens: This code looks at actions.own instead: https://github.com/broadinstitute/rawls/pull/2555/files#diff-cc8ea874ea1b95b1e65e8d09795f295476565b0a44d36b924ad355e659ba61eeR46

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: I don't, it's just to make combining `eligibleWorkspaces` and `migratingOrRecentlyFinishedWorkspaces` in the response easier

core/src/main/resources/swagger/api-docs.yaml
marctalbott: That's correct, it's just the most recent attempt. The array of `WorkspaceMigrationMetadata` is a bit of carryover from PPW. The way we're doing restarts for this migration makes it far less likely for there to actually be multiple distinct migration attempts. In PPW, calling the `startMigration` API a second time would trigger a new distinct migration attempt. For this migration, calling the `startMigration` API a second time will attempt to restart the previous migration attempt

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
cahrens: Oh, I misread the code. I thought somehow only the first part of the response from `getEligibleUnmigratedWorkspaces` was getting assigned to the variable `eligibleWorkspaces`. Would you mind naming it something more like `eligibleWorkspacesAndEmptyProgress` or …. I don't know, something that makes it clear that this is more than just a list of workspaces?

cahrens: Or perhaps the return types are enough of the helper methods. As usual, I simply find Scala difficult to read.

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: It does, that's the correct way of doing it. It uses a different Sam API that is per-`workspace` resource. I wanted to do it the right way when possible (which includes all cases where workspaces are actually scheduled for migration) -- it's just expensive in this situation due to the available Sam APIs.

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: :+1: I'll rename it -- it took me a bit to remind myself of what was happening there too

cahrens: Great, then I think we would need the other to be `migratingOrRecentlyFinishedWorkspacesAndProgress`.

cahrens (approved):

cahrens:
jenkins retest

PR #2551, opened 2023-09-27
AJ-1348: disable alpha-only integration tests; replace with unit tests

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1348

We have a few integration tests (aka Jenkins swatomation) that run exclusively in the alpha environment.

This PR ignores/disables all of those tests. Most of them already had equivalent unit tests, and this PR adds a couple unit tests to make up the remainder of the coverage.

Note that I am NOT yet fully deleting the alpha tests, simply ignoring them. I want to make sure that reducing to 0 tests does not cause problems for the Jenkins job and this makes it super easy to turn a test back on if necessary. I will delete the alpha tests fully once I see Jenkins succeed. I will leave the AJ-1348 ticket open until the tests are deleted.

---

**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.rawls.snapshot
 
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import bio.terra.datarepo.model.{DatasetSummaryModel, SnapshotModel, SnapshotSourceModel}
+import bio.terra.workspace.client.ApiException
 import bio.terra.workspace.model._
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -32,48 +33,64 @@ import scala.jdk.CollectionConverters._
 
 class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSugar with TestDriverComponent {
```
davidangb, line 35: the diff below is ugly. to summarize: I extracted `defaultMockSamDao()` and `defaultMockWorkspaceManagerDao()` methods out of the "create a new snapshot reference to a TDR snapshot" test. Then, I added two new tests which call `defaultMockSamDao()` and `defaultMockWorkspaceManagerDao()`. I think it's a little easier to see as a split diff instead of  as a unified diff.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -108,6 +127,92 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+                                                               Option(DataReferenceDescriptionField("foo")),
+                                                               UUID.randomUUID()
+                                         )
+          ),
+          Duration.Inf
+        )
+
+        // assert that the service checked to see if the workspace exists
+        verify(mockWorkspaceManagerDAO, times(1)).getWorkspace(
+          any[UUID],
```
jladieu: I like using `any` matchers by default when setting up test _stubs_, but I think using `any` in a _verify_ can let a certain category of surprise bugs creep through...this is exactly what bit me when I was trying to make changes to the UI and using overly loose verifies (see [comment](https://github.com/DataBiosphere/terra-ui/pull/4235/commits/5ea60c1a957813e441cc919cef7b099226cff67b#r1340175189)) -- the example I provided is even looser than this, I don't even check the arg list...  but hopefully the point makes sense.

Is there a specific workspace ID from the test fixture you can look for here vs. `any`?

I'm okay with leaving the `RawlsRequestContext` matcher loose assuming it's not particularly important for the code under test.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -108,6 +127,92 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+          any[RawlsRequestContext]
+        )
+
+        // assert that the service DID NOT call WSM's createWorkspace
+        verify(mockWorkspaceManagerDAO, times(0)).createWorkspace(
```
jladieu, line 211: In the opposite direction from my other comment...this is appropriate use of the `any` because we want to make sure it didn't get called _at all_ or _with anything_

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -108,6 +127,92 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+                                                             Option(DataReferenceDescriptionField("foo")),
+                                                             UUID.randomUUID()
+                                       )
+        ),
+        Duration.Inf
+      )
+
+      // assert that the service checked to see if the workspace exists
+      verify(mockWorkspaceManagerDAO, times(1)).getWorkspace(
+        any[UUID],
```
jladieu: Same comment as below about verifying for a more specific UUID here.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -32,48 +33,64 @@ import scala.jdk.CollectionConverters._
...
-          any[String],
-          any[CloningInstructionsEnum],
-          any[RawlsRequestContext]
-        )
+    mockSamDAO
+  }
+
+  // create a mockito-powered WorkspaceManagerDAO that always returns success for creating
+  // a snapshot reference. Tests can add more responses/expectations to the mock returned by this method.
+  private def defaultMockWorkspaceManagerDao() = {
```
jladieu, line 61: maybe to support my comment later on, you can pass in a `workspaceId` from the caller, and use that instead of something randomly generated -- then you can look for a hardcoded string in later verification

jladieu (approved):
Awesome work!  I left some suggestions in the theme of improving the binding between the test fixture setup & the verification step, but all LGTM.

davidangb:
@jladieu good call on using specific arguments instead of `any` - I've updated to reflect that!

nawatts (approved):

PR #2552, opened 2023-09-27
[WOR-732] Remove leo app enabling workaround during clone

aherbst-broad:
[Ticket WOR-732](https://broadworkbench.atlassian.net/browse/WOR-732)

## Context
WSM was not enabling applications properly during workspace clone due to a bug. Now that the bug has been [remediated](https://github.com/DataBiosphere/terra-workspace-manager/commit/0c7b05dc00fab0aa07ae47326ce3f7f717a5cda1) we no longer need the workaround of disabling/enabling to trigger the correct behavior. 

## This PR
* Removes the hack and updates associated tests.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -58,40 +58,20 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
                                                spendProfileId: String,
                                                billingProjectNamespace: String,
                                                applicationIds: Seq[String],
+                                               policyInputs: Option[WsmPolicyInputs],
                                                ctx: RawlsRequestContext
-  ): CreatedWorkspace =
-    getWorkspaceApi(ctx).createWorkspace(
-      new CreateWorkspaceRequestBody()
-        .id(workspaceId)
-        .displayName(displayName)
-        .spendProfile(spendProfileId)
-        .stage(WorkspaceStageModel.MC_WORKSPACE)
-        .projectOwnerGroupId(billingProjectNamespace)
-        .applicationIds(applicationIds.asJava)
-    )
-
-  override def createProtectedWorkspaceWithSpendProfile(workspaceId: UUID,
-                                                        displayName: String,
-                                                        spendProfileId: String,
-                                                        billingProjectNamespace: String,
-                                                        ctx: RawlsRequestContext
   ): CreatedWorkspace = {
-    val policyInputs = new WsmPolicyInputs()
-    val protectedPolicyInput = new WsmPolicyInput()
-    protectedPolicyInput.name("protected-data")
-    protectedPolicyInput.namespace("terra")
-    protectedPolicyInput.additionalData(List().asJava)
+    val request = new CreateWorkspaceRequestBody()
+      .id(workspaceId)
+      .displayName(displayName)
+      .spendProfile(spendProfileId)
+      .stage(WorkspaceStageModel.MC_WORKSPACE)
+      .projectOwnerGroupId(billingProjectNamespace)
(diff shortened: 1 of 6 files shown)


marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -635,6 +625,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
                                                                   workspaceRequest.name,
                                                                   spendProfileId,
```
cahrens: Don't we also want to start Leo when creating a `protectedWorkspaceWithSpendProfile` (line 616)?

This is the sort of reason why I don't like having the 2 different methods, vs. just making whether or not it is protected a boolean arg on a single method.

And do we need to do anything on clone?

cahrens (changes requested):

aherbst-broad:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aherbst-broad: Yeah good call, I refactored a bit in this [commit](https://github.com/broadinstitute/rawls/pull/2552/commits/ca812746a7305aca5f73cb573f007ce8d6739c2e) to coalesce the 2 methods into a single one consuming an optional policy inputs object. 

The enable/disable was in the cloneworkspace method of this class. WSM now does the right thing on clone and pulls the app IDs from the source workspace and enables the apps properly. I tested in a BEE and got a WDS booted up on a clone with this change.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -57,39 +57,21 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
-    )
+    val request = new CreateWorkspaceRequestBody()
+      .id(workspaceId)
+      .displayName(displayName)
+      .spendProfile(spendProfileId)
+      .stage(WorkspaceStageModel.MC_WORKSPACE)
+      .projectOwnerGroupId(billingProjectNamespace)
+      .applicationIds(applicationIds.asJava)
+
+    policyInputs.map(request.policies)
```
cahrens, line 72: This looks confusing to me-- it populates the request with the policies? Is it somehow handling the Option?

cahrens (approved):
I have a question about how the policyInputs are populated, but I do assume it works. 

I didn't look at the swatomation test failures.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
marctalbott: This is equivalent to:
```
policyInputsOpt.map { actualPolicyInputs =>
    request.policies(actualPolicyInputs)
}
```
(I renamed `policyInputs` here for clarity)


Calling `.map` on an `Option` will only execute the code inside the block if the `Option` is defined. In this case, we'll only set the request policies if `policyInputs` is defined. 

You can also omit the `()` and parameter if the function called inside the `.map` accepts the same type that is inside the `Option` which is why `.map(request.policies)` is okay.

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

PR #2555, opened 2023-09-28
[WOR-1247] Add multiregional bucket migration APIs for owners

marctalbott:
Ticket: [WOR-1247](https://broadworkbench.atlassian.net/browse/WOR-1247)
* I wanted to avoid too much code duplication between the owner and admin APIs. This approach worked out _okay_ -- there are parts of it that are easy and/or clean, but it wasn't the most fun to test and I worry it's unnecessarily difficult to reason about. Still, I wanted to get eyes on it to get some input as I've been sitting on it longer than I'd like and because it's hard to know if it's too difficult to understand when I've been staring at it this long

---

**PR checklist**

[WOR-1247]: https://broadworkbench.atlassian.net/browse/WOR-1247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,33 +30,173 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
...
+    asFCAdminWithWorkspace(workspaceName) {
+      migrateWorkspaceBucketInternal
+    }
+
+  def migrateWorkspaceBucket(workspaceName: WorkspaceName): Future[MultiregionalBucketMigrationMetadata] =
+    asOwnerWithWorkspace(workspaceName) {
+      migrateWorkspaceBucketInternal
+    }
+
+  def adminMigrateAllWorkspaceBuckets(
```
marctalbott, line 230: I didn't bother abstracting out `asFCAdminWithWorkspaces` or `asOwnerWithWorkspaces` because these are currently the only methods that accept an arbitrary list of workspaces. If we end up going with this approach, I'll end up doing that when working on [WOR-1274](https://broadworkbench.atlassian.net/browse/WOR-1274)

core/src/main/scala/org/broadinstitute/dsde/rawls/model/SamModel.scala
```diff
@@ -83,6 +83,7 @@ object SamBillingProjectActions {
   val readSpendReport = SamResourceAction("read_spend_report")
   val updateBillingAccount = SamResourceAction("update_billing_account")
   val deleteBillingProject = SamResourceAction("delete")
+  val own = SamResourceAction("own")
```
marctalbott, line 86: see https://github.com/broadinstitute/sam/pull/1183

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
```diff
@@ -27,33 +30,173 @@ class BucketMigrationService(val dataSource: SlickDataSource, val samDAO: SamDAO
     with WorkspaceSupport
     with LazyLogging {
 
-  def getBucketMigrationProgressForWorkspace(
-    workspaceName: WorkspaceName
-  ): Future[Option[MultiregionalBucketMigrationProgress]] =
+  /**
+    * Helper functions to enforce appropriate authz and load workspace(s) if authz passes
+    */
+  def asFCAdminWithWorkspace[T](workspaceName: WorkspaceName)(op: Workspace => Future[T]): Future[T] =
```
marctalbott: These and the `Internal` functions below had to be made public to make unit testing easier and less repetitive. This is one of the things I'm questioning

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -607,6 +607,89 @@ paths:
...
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+    post:
+      tags:
+        - billing_v2
+      summary: start bucket migration attempt for Google workspaces in billing project
```
cahrens: It would be good to include what happens to workspaces in the following states:

1. Not eligible for migration
2. Migration previously attempted but failed

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -607,6 +607,89 @@ paths:
...
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+  /api/billing/v2/{projectId}/bucketMigration/progress:
+    get:
+      tags:
+        - billing_v2
+      summary: get bucket migration progress for workspaces in billing project
```
cahrens, line 686: Does the return object include workspaces that have never been scheduled for migration (either because they were not eligible, or because they just had never been started)?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +4063,114 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/bucketMigration:
+    get:
+      tags:
+        - workspaces_v2
+      summary: list bucket migration attempts for workspace
```
cahrens: What does this return for a workspace that has never been scheduled for migration? An empty array?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +4063,114 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/WorkspaceMigrationMetadata'
+        403:
+          description: You must be an owner to start migration attempts
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
```
cahrens, line 4212: What do you get if you attempt to start a migration for an ineligible workspace?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +4063,114 @@ paths:
...
+        200:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/MultiregionalBucketMigrationProgress'
+        403:
+          description: You must be an owner to list migration attempts
```
cahrens: Copy/paste error

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3980,6 +4063,114 @@ paths:
...
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/bucketMigration/progress:
+    get:
+      tags:
+        - workspaces_v2
+      summary: get bucket migration progress for workspace
```
cahrens, line 4217: What do you get if you ask for progress for an ineligible workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
cahrens: How is this "asFCAdmin"?

cahrens:
This general approach looks OK to me, but why do you have to call the internal method signatures in the unittest?

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
aherbst-broad: 🤔 Getting my head around this. I see that the tests are call these methods to ensure that we are properly disallowing access in the correct situations which makes sense. I further observe that the consuming methods within `BucketMigrationService` like `migrateWorkspaceBucketsInBillingProject(billingProjectName)` do not have corresponding tests. Would it make sense to get a test around those instead and hide these methods?

That being said, I think this is an "ok" implementation as-is. The API is sound and nothing is preventing us from refactoring this in the future.

aherbst-broad (commented):
Initial comment on the approach

core/src/main/scala/org/broadinstitute/dsde/rawls/bucketMigration/BucketMigrationService.scala
marctalbott: Yeah, I made them public because I was trying to avoid a lot of repetitive tests, but I'm thinking now that a bit of repetition in the tests would be preferred if we can actually directly test the real entrypoints to this service

core/src/main/resources/swagger/api-docs.yaml
marctalbott: The same as if you ask for progress for a workspace that hasn't been migrated which is a 404

core/src/main/resources/swagger/api-docs.yaml
marctalbott: A 400. I'll update the swagger

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Yes, that's correct

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Yes, it includes the workspace name in the response and maps to `null`

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Agreed, thanks for the close read on this Swagger, I'm realizing now that I didn't do a great job of keeping it accurate. Fix to come :)

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -653,12 +659,24 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/WorkspaceMigrationMetadata'
+        400:
+          description: Some or all workspaces in billing project were ineligible for migration.
```
cahrens: So you cannot call this method on a billing project where some of the workspaces are not in US multi-region? I would just expect it to ignore those workspaces instead of erroring, as otherwise the method doesn't seem very useful.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1014,12 +1044,24 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/WorkspaceMigrationMetadata'
+        400:
+          description: Some or all workspaces in billing project were ineligible for migration.
```
cahrens: This endpoint is just for a single workspace, so it shouldn't be checking other workspaces in the billing project.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1124,12 +1184,24 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/WorkspaceMigrationMetadata'
+        400:
+          description: Some or all workspaces in billing project were ineligible for migration.
```
cahrens: Same comment as above-- you cannot call this method on a billing project where some of the workspaces are not in US multi-region? I would just expect it to ignore those workspaces instead of erroring, as otherwise the method doesn't seem very useful.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1096,7 +1150,7 @@ paths:
                 type: array
                 items:
                   type: object
-                  description: workspaceNamespace/workspaceName -> bucket migration metadata
+                  description: workspaceNamespace/workspaceName -> bucket migration metadata. bucket migration metadata is an empty list for workspaces that have not been scheduled for migration before
```
cahrens: Nit on al of these-- I would transform into completely sentences or at least remove the period. For instance:

workspaceNamespace/workspaceName -> bucket migration metadata (which will be an empty list for workspaces that have never been scheduled for migration)

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4067,8 +4145,8 @@ paths:
     get:
       tags:
         - workspaces_v2
-      summary: list bucket migration attempts for workspace
-      description: list bucket migration attempts for workspace
+      summary: list bucket migration attempts for workspace. empty list for workspaces that have not been migrated before
+      description: list bucket migration attempts for workspace. empty list for workspaces that have not been migrated before
```
cahrens, line 4149: Nit: complete sentences or no period

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4106,12 +4190,24 @@ paths:
             'application/json':
               schema:
                 $ref: '#/components/schemas/WorkspaceMigrationMetadata'
+        400:
+          description: Some or all workspaces in billing project were ineligible for migration.
```
cahrens: This is just a single workspace, no?

core/src/main/resources/swagger/api-docs.yaml
cahrens: Reading the test, I think the missing detail here is that even though a 400 is thrown, any eligible workspaces in the billing project *are* scheduled for migration. This seems like it could be rather confusing, but at the very least it should be clarified in the Swagger doc.

marctalbott: It will migrate any workspaces that are eligible and the exception will include the list of errors, each of which will include the workspace in question

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Yup, copy/paste error.

core/src/main/resources/swagger/api-docs.yaml
marctalbott: It is, this was a copy-paste error as well

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Including the response from above for visibility
> It will migrate any workspaces that are eligible and the exception will include the list of errors, each of which will include the workspace in question

cahrens (approved):

aherbst-broad (approved):

PR #2556, opened 2023-09-29
AJ-1309: Disable cross-platform snapshots-by-reference

jladieu:

Update `SnapshotService` to enforce stricter validation around the `cloudPlatform` of the snapshot & workspace.  Basically: disallow Azure-based snapshots by reference entirely, and futureproof the code to ensure that if we _do_ start supporting Azure snapshots by reference that we'll have to bake in support at the appropriate touchpoints in this area of the code.

---

**PR checklist**

- [x] Make sure Swagger is updated if API changes (N/A)
  - [x] **...and Orchestration's Swagger too!**(N/A)
- [x] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala). (N/A)

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -158,7 +158,11 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
   ): DataRepoSnapshotResource = {
     val snapshot = new DataRepoSnapshotAttributes().instanceName(instanceName).snapshot(snapshotId.toString)
     val commonFields =
-      new ReferenceResourceCommonFields().name(name.value).cloningInstructions(CloningInstructionsEnum.NOTHING)
+      new ReferenceResourceCommonFields()
+        .name(name.value)
+        // Note: that we're ignoring the passed in `cloningInstructions` is a known issue, addressed as part of
+        // https://github.com/broadinstitute/rawls/pull/2081
+        .cloningInstructions(CloningInstructionsEnum.NOTHING)
     description.map(d => commonFields.description(d.value))
     val request = new CreateDataRepoSnapshotReferenceRequestBody().snapshot(snapshot).metadata(commonFields)
     getReferencedGcpResourceApi(ctx).createDataRepoSnapshotReference(request, workspaceId)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
@@ -1,0 +1,54 @@
+package org.broadinstitute.dsde.rawls.snapshot
+
+import bio.terra.datarepo.model.{CloudPlatform => SnapshotCloudPlatform}
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.model.WorkspaceCloudPlatform.WorkspaceCloudPlatform
+import org.broadinstitute.dsde.rawls.model.{Workspace, WorkspaceCloudPlatform}
+
+object SnapshotReferenceCreationValidator {
+  def constructor(workspaceContext: Workspace, snapshot: WrappedSnapshot): SnapshotReferenceCreationValidator =
+    new SnapshotReferenceCreationValidator(workspaceContext, snapshot)
+}
+
+// thrown to disallow creation of a snapshot reference on an unsupported platform
+class UnsupportedPlatformException(message: String) extends RawlsException(message)
+// thrown to disallow creation of a snapshot reference across cloud boundaries
+class PlatformBoundaryException(message: String) extends RawlsException(message)
+// thrown to disallow creation of a snapshot reference across a protected data boundary
+class ProtectedDataException(message: String) extends RawlsException(message)
+
+class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val snapshot: WrappedSnapshot) {
+
+  // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
+  // We'll have to use this workaround for identifying protected status.
+  @throws(classOf[ProtectedDataException])
+  def validateProtectedStatus(): Unit =
+    if (!isWorkspaceProtected && snapshot.isProtected) {
+      throw new ProtectedDataException("Unable to add protected snapshot to unprotected workspace.")
+    }
+
+  // Throws an exception when the given snapshot cannot be referenced by the given workspace due to crossing an
+  // unsupported platform boundary.
+  @throws(classOf[PlatformBoundaryException])
+  def validateWorkspacePlatformCompatibility(workspacePlatform: WorkspaceCloudPlatform): Unit =
+    // Defining all acceptable combinations is overkill when all we really need to do is prevent anything Azure, but
+    // this is here as a safeguard against us introducing support for Azure without deliberately addressing the issue of
+    // snapshots by ref across cloud platforms.
+    (snapshot.platform, workspacePlatform) match {
+      case (SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp) => // ok
+      case _ =>
+        throw new PlatformBoundaryException(
+          "Snapshots by reference are not supported across the given cloud boundaries (" +
+            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform})."
+        )
+    }
+
+  @throws(classOf[UnsupportedPlatformException])
+  def validateSnapshotPlatform(): Unit =
+    if (snapshot.platform == SnapshotCloudPlatform.AZURE) {
+      throw new UnsupportedPlatformException("Snapshots by reference are not supported for Azure datasets.")
+    }
+
+  // TODO: get this information from a more authoritative source rather than relying on the hardcoded bucket prefix
+  private def isWorkspaceProtected: Boolean = workspaceContext.bucketName.startsWith("fc-secure")
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.rawls.model.{
   SamWorkspaceActions,
   SnapshotListResponse,
   WorkspaceAttributeSpecs,
+  WorkspaceCloudPlatform,
   WorkspaceName
 }
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
@@ -73,7 +74,9 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
         snapshotValidator.validateWorkspacePlatformCompatibility(wsmWorkspace.getCloudPlatform)
       } catch {
         case _: AggregateWorkspaceNotFoundException =>
-          // create the stub workspace in WSM if it does not already exist
+          // if a WSM workspace does not already exist, assume the platform is GCP, confirm platform compatibility,
+          // and then create a stub workspace
+          snapshotValidator.validateWorkspacePlatformCompatibility(WorkspaceCloudPlatform.Gcp)
           workspaceManagerDAO.createWorkspace(wsid, rawlsWorkspace.workspaceType, ctx)
       }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.rawls.snapshot
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.workspace.model._
 import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
@@ -13,12 +14,11 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsRequestContext,
   SamWorkspaceActions,
   SnapshotListResponse,
-  Workspace,
   WorkspaceAttributeSpecs,
   WorkspaceName
 }
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
-import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
+import org.broadinstitute.dsde.rawls.workspace.{AggregateWorkspaceNotFoundException, AggregatedWorkspaceService}
 
 import java.util.UUID
 import scala.annotation.tailrec
@@ -47,41 +47,47 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     with WorkspaceSupport
     with LazyLogging {
 
-  def createSnapshot(workspaceName: WorkspaceName, snapshot: NamedDataRepoSnapshot): Future[DataRepoSnapshotResource] =
+  def createSnapshot(workspaceName: WorkspaceName,
+                     snapshotIdentifiers: NamedDataRepoSnapshot
+  ): Future[DataRepoSnapshotResource] =
     getV2WorkspaceContextAndPermissions(workspaceName,
                                         SamWorkspaceActions.write,
                                         Some(WorkspaceAttributeSpecs(all = false))
-    ).flatMap { workspaceContext =>
-      val wsid = workspaceContext.workspaceIdAsUUID // to avoid UUID parsing multiple times
-      // create the stub workspace in WSM if it does not already exist
-      if (!workspaceStubExists(wsid, ctx)) {
-        workspaceManagerDAO.createWorkspace(wsid, workspaceContext.workspaceType, ctx)
+    ).flatMap { rawlsWorkspace =>
+      val wsid = rawlsWorkspace.workspaceIdAsUUID // to avoid UUID parsing multiple times
+      val snapshot =
+        new WrappedSnapshot(dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken))
+      val snapshotValidator = new SnapshotReferenceCreationValidator(rawlsWorkspace, snapshot)
+      val aggregatedWorkspaceService = new AggregatedWorkspaceService(workspaceManagerDAO)
+
+      // prevent snapshots from disallowed platforms
+      snapshotValidator.validateSnapshotPlatform()
+
+      // prevent disallowed access across workspace or dataset protection boundaries
+      snapshotValidator.validateProtectedStatus()
+
+      try {
+        // if there's an existing WSM workspace, make sure its platform is compatible
+        // with that of the snapshot's dataset.
+        val wsmWorkspace = aggregatedWorkspaceService.fetchAggregatedWorkspace(rawlsWorkspace, ctx)
+        snapshotValidator.validateWorkspacePlatformCompatibility(wsmWorkspace.getCloudPlatform)
+      } catch {
+        case _: AggregateWorkspaceNotFoundException =>
+          // create the stub workspace in WSM if it does not already exist
+          workspaceManagerDAO.createWorkspace(wsid, rawlsWorkspace.workspaceType, ctx)
       }
-      validateProtectedStatus(workspaceContext, snapshot)
+
       // create the requested snapshot reference
-      val snapshotRef = workspaceManagerDAO.createDataRepoSnapshotReference(wsid,
-                                                                            snapshot.snapshotId,
-                                                                            snapshot.name,
-                                                                            snapshot.description,
-                                                                            terraDataRepoInstanceName,
-                                                                            CloningInstructionsEnum.NOTHING,
-                                                                            ctx
+      val snapshotRef = workspaceManagerDAO.createDataRepoSnapshotReference(
+        wsid,
+        snapshotIdentifiers.snapshotId,
+        snapshotIdentifiers.name,
+        snapshotIdentifiers.description,
+        terraDataRepoInstanceName,
+        CloningInstructionsEnum.NOTHING,
+        ctx
       )
       Future.successful(snapshotRef)
-    }
-
-  // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
-  // We'll have to use this workaround for identifying protected status
-  private def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
-    // logically it might make more sense to check if the snapshot is protected before the workspace
-    // but that is a more expensive check
-    // check if workspace is protected
-    if (!workspaceContext.bucketName.startsWith("fc-secure")) {
-      // if not, check if snapshot is protected
-      val sources = dataRepoDAO.getSnapshot(snapshot.snapshotId, ctx.userInfo.accessToken).getSource
-      if (sources.asScala.exists(_.getDataset.isSecureMonitoringEnabled)) {
-        throw new RawlsException("Unable to add protected snapshot to unprotected workspace.")
-      }
     }
 
   def getSnapshot(workspaceName: WorkspaceName, referenceId: String): Future[DataRepoSnapshotResource] = {
@@ -249,9 +255,6 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       workspaceManagerDAO.deleteDataRepoSnapshotReference(workspaceContext.workspaceIdAsUUID, snapshotUuid, ctx)
     }
   }
-
-  private def workspaceStubExists(workspaceId: UUID, ctx: RawlsRequestContext): Boolean =
-    Try(workspaceManagerDAO.getWorkspace(workspaceId, ctx)).isSuccess
 
   private def validateSnapshotId(snapshotId: String): UUID =
     Try(UUID.fromString(snapshotId)) match {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -34,7 +34,14 @@ object SnapshotService {
                   terraDataRepoUrl: String,
                   dataRepoDAO: DataRepoDAO
   )(ctx: RawlsRequestContext)(implicit executionContext: ExecutionContext): SnapshotService =
-    new SnapshotService(ctx, dataSource, samDAO, workspaceManagerDAO, terraDataRepoUrl, dataRepoDAO)
+    new SnapshotService(ctx,
+                        dataSource,
+                        samDAO,
+                        workspaceManagerDAO,
+                        terraDataRepoUrl,
+                        dataRepoDAO,
+                        new AggregatedWorkspaceService(workspaceManagerDAO)
+    )
 }
 
 class SnapshotService(protected val ctx: RawlsRequestContext,
@@ -42,7 +49,8 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
                       val samDAO: SamDAO,
                       workspaceManagerDAO: WorkspaceManagerDAO,
                       terraDataRepoInstanceName: String,
-                      dataRepoDAO: DataRepoDAO
+                      dataRepoDAO: DataRepoDAO,
+                      aggregatedWorkspaceService: AggregatedWorkspaceService
 )(implicit protected val executionContext: ExecutionContext)
     extends FutureSupport
     with WorkspaceSupport
@@ -59,7 +67,6 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       val snapshot =
         new WrappedSnapshot(dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken))
       val snapshotValidator = new SnapshotReferenceCreationValidator(rawlsWorkspace, snapshot)
-      val aggregatedWorkspaceService = new AggregatedWorkspaceService(workspaceManagerDAO)
 
       // prevent snapshots from disallowed platforms
       snapshotValidator.validateSnapshotPlatform()
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
@@ -1,0 +1,23 @@
+package org.broadinstitute.dsde.rawls.snapshot
+
+import bio.terra.datarepo.model.{CloudPlatform, DatasetSummaryModel, SnapshotModel}
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+
+// A basic wrapper class to simplify interaction with the SnapshotModel from TDR
+class WrappedSnapshot(snapshot: SnapshotModel) {
+
+  // While getSource returns a list of SnapshotSourceModels, this is an artifact and we can make an assumption that
+  // a given snapshot will have exactly one source that will always have a dataset.  Violations of this assumption
+  // indicate a bug, in which case, crash and burn.
+  private def snapshotDataset(): DatasetSummaryModel = getOnlyElement(snapshot.getSource.asScala.toList).getDataset
+
+  private def getOnlyElement[A](list: List[A]): A = list match {
+    case head :: Nil => head
+    case _           => throw new NoSuchElementException(s"Expected exactly one element, but found ${list.size}")
+  }
+
+  def isProtected: Boolean = snapshotDataset().isSecureMonitoringEnabled()
+
+  def platform: CloudPlatform = snapshotDataset().getCloudPlatform
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.rawls.snapshot
 
 import bio.terra.datarepo.model.{CloudPlatform, DatasetSummaryModel, SnapshotModel}
+import org.broadinstitute.dsde.rawls.RawlsException
 
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 
@@ -13,11 +14,21 @@ class WrappedSnapshot(snapshot: SnapshotModel) {
   private def snapshotDataset(): DatasetSummaryModel = getOnlyElement(snapshot.getSource.asScala.toList).getDataset
 
   private def getOnlyElement[A](list: List[A]): A = list match {
-    case head :: Nil => head
-    case _           => throw new NoSuchElementException(s"Expected exactly one element, but found ${list.size}")
+    case onlyElement :: Nil => onlyElement
+    case _                  => throw new NoSuchElementException(s"Expected exactly one element, but found ${list.size}")
   }
 
-  def isProtected: Boolean = snapshotDataset().isSecureMonitoringEnabled()
+  def isProtected: Boolean = Option(snapshotDataset().isSecureMonitoringEnabled)
+    .getOrElse(
+      throw new RawlsException(
+        s"Snapshot ${snapshot.getId}'s DatasetSummaryModel had null value for secure monitoring; cannot continue"
+      )
+    )
 
-  def platform: CloudPlatform = snapshotDataset().getCloudPlatform
+  def platform: CloudPlatform = Option(snapshotDataset().getCloudPlatform)
+    .getOrElse(
+      throw new RawlsException(
+        s"Snapshot ${snapshot.getId}'s DatasetSummaryModel had null value for cloud platform; cannot continue"
+      )
+    )
 }
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
@@ -9,7 +9,12 @@ import bio.terra.workspace.model._
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
 import org.broadinstitute.dsde.rawls.model.WorkspaceType.WorkspaceType
-import org.broadinstitute.dsde.rawls.model.{DataReferenceDescriptionField, DataReferenceName, ErrorReport, RawlsRequestContext}
+import org.broadinstitute.dsde.rawls.model.{
+  DataReferenceDescriptionField,
+  DataReferenceName,
+  ErrorReport,
+  RawlsRequestContext
+}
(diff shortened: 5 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -162,7 +162,11 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
   ): DataRepoSnapshotResource = {
     val snapshot = new DataRepoSnapshotAttributes().instanceName(instanceName).snapshot(snapshotId.toString)
     val commonFields =
-      new ReferenceResourceCommonFields().name(name.value).cloningInstructions(CloningInstructionsEnum.NOTHING)
+      new ReferenceResourceCommonFields()
+        .name(name.value)
+        // Note: that we're ignoring the passed in `cloningInstructions` is a known issue, addressed as part of
+        // https://github.com/broadinstitute/rawls/pull/2081
+        .cloningInstructions(CloningInstructionsEnum.NOTHING)
```
calypsomatic, line 165: Oh man but that PR is never gonna happen

jladieu: 😆

jladieu: Fine by me, just wanted to save myself a headscratch the next time I wander into this area and squint at it.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -10,15 +10,7 @@ import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
...
-import org.broadinstitute.dsde.rawls.model.{
-  DataReferenceDescriptionField,
-  DataReferenceName,
-  NamedDataRepoSnapshot,
-  RawlsRequestContext,
-  SamResourceAction,
-  SamResourceTypeNames,
-  SamUserStatusResponse
-}
+import org.broadinstitute.dsde.rawls.model.{DataReferenceDescriptionField, DataReferenceName, NamedDataRepoSnapshot, RawlsRequestContext, SamResourceAction, SamResourceTypeNames, SamUserStatusResponse}
```
jladieu: I'm not sure why my formatter smooshed all these into one line -- I liked it better before, but lost the battle with the IDE...any suggestions of what I did wrong here are welcome.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -47,43 +47,48 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
-    ).flatMap { workspaceContext =>
-      val wsid = workspaceContext.workspaceIdAsUUID // to avoid UUID parsing multiple times
-      // create the stub workspace in WSM if it does not already exist
-      if (!workspaceStubExists(wsid, ctx)) {
-        workspaceManagerDAO.createWorkspace(wsid, ctx)
+    ).flatMap { rawlsWorkspace =>
+      val wsid = rawlsWorkspace.workspaceIdAsUUID // to avoid UUID parsing multiple times
+      val snapshot = dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken)
+      val validator = new SnapshotReferenceCreationValidator(rawlsWorkspace, snapshot)
+      val aggregatedWorkspaceService = new AggregatedWorkspaceService(workspaceManagerDAO)
```
davidangb: hm - TODO here is to check with the workspaces team. The `Workspace` object has a `workspaceType: WorkspaceType` which I think is a valid proxy for Gcp vs. Azure. You may also be able to determine the cloud platform from the `googleProjectNumber` or `bucketName` fields. If Workspaces team says these are safe to rely on, you could eliminate the need for `AggregatedWorkspaceService` entirely, including the extra outbound REST call.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -47,43 +47,48 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+          // if there's an existing WSM workspace, make sure its platform is compatible
+          // with that of the snapshot's dataset.
+          validator.validatePlatforms(wsmWorkspace.getCloudPlatform)
+        case Failure(_) =>
+          // Pre-emptively fail (and prevent creation of a workspace stub here if snapshot is Azure). We can remove this
+          // check once we support Azure-based snapshots by reference.
+          validator.validateSnapshotPlatform()
+
+          // create the stub workspace in WSM if it does not already exist
+          workspaceManagerDAO.createWorkspace(wsid, ctx)
```
davidangb: I think this call is misplaced. We need to call `workspaceManagerDAO.createWorkspace(wsid, ctx)` when:
1. this is a GCP workspace
2. there does not already exist a workspace record in WSM

if condition 1 above is true, then `aggregatedWorkspaceService.getAggregatedWorkspace(rawlsWorkspace, ctx)` will always return `Success` and we'll never reach this code.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val snapshot: SnapshotModel) {
+
+  // TODO: get this information from a more authoritative source rather than relying on the hardcoded bucket prefix
+  private def isWorkspaceProtected: Boolean = workspaceContext.bucketName.startsWith("fc-secure")
+
+  private def isSnapshotProtected: Boolean = snapshot.getSource.asScala.exists(_.getDataset.isSecureMonitoringEnabled)
+
+  private def snapshotPlatform: SnapshotCloudPlatform = {
+    val snapshotSourceDatasets = snapshot.getSource.asScala
+    val datasetCloudPlatforms = snapshotSourceDatasets.map(_.getDataset.getCloudPlatform).filter(_ != null).toSet
```
davidangb: what you have here is fine; to be Scala-idiomatic, you'd work with `Option` instead of `null`, like:
```
val datasetCloudPlatforms = snapshotSourceDatasets.flatMap(x => Option(x.getDataset.getCloudPlatform)).toSet
```
(boy, I hope that's functionally equivalent; I didn't actually try it)

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+
+    // assume GCP if datasets don't come from a source explicitly marked with another platform
+    if (datasetCloudPlatforms.isEmpty) {
+      return SnapshotCloudPlatform.GCP
+    }
```
davidangb: I don't know if this is safe - we should check with the TDR team

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+    // fail if datasets sources somehow span multiple clouds
+    if (datasetCloudPlatforms.size > 1) {
+      throw new IllegalStateException(
+        s"snapshot ${snapshot.getId} unexpectedly has more than one cloudPlatform: $datasetCloudPlatforms"
+      )
+    }
+
+    // fail if dataset sourced from something other than GCP aren't all explicitly marked as such
+    // as this might indicate co-mingled GCP (blank) & non-GCP datasets
+    val onlyCloudPlatform = datasetCloudPlatforms.take(1).last
```
davidangb: this could be just:
```scala
val onlyCloudPlatform = datasetCloudPlatforms.head
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+    onlyCloudPlatform match {
+      case SnapshotCloudPlatform.GCP => onlyCloudPlatform
+      case _ =>
+        if (snapshotSourceDatasets.forall(_.getDataset.getCloudPlatform == onlyCloudPlatform)) onlyCloudPlatform
+        else
+          throw new IllegalStateException(
+            s"snapshot ${snapshot.getId} unexpectedly has sources from more than one cloudPlatform"
+          )
```
davidangb: how would we ever reach this second case? We've already run `toSet` which eliminates duplicates, and we already have a check for `datasetCloudPlatforms.size > 1` … this feels redundant.

If you combine a bunch of these comments, I think you can simplify this method to:
```scala
  private def snapshotPlatform: SnapshotCloudPlatform = {
    val snapshotSourceDatasets = snapshot.getSource.asScala
    val datasetCloudPlatforms = snapshotSourceDatasets.flatMap(x => Option(x.getDataset.getCloudPlatform)).toSet

    // fail if datasets sources somehow span multiple clouds
    if (datasetCloudPlatforms.size > 1) {
      throw new IllegalStateException(
        s"snapshot ${snapshot.getId} unexpectedly has more than one cloudPlatform: $datasetCloudPlatforms"
      )
    }

    // we're left with either zero or one cloud platform from the snapshot sources
    datasetCloudPlatforms.headOption match {
      // assume GCP if datasets don't come from a source explicitly marked with another platform
      case None => SnapshotCloudPlatform.GCP
      case Some(platform) => platform
    }
  }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+  // We'll have to use this workaround for identifying protected status.
+  @throws(classOf[ProtectedDataException])
+  def validateProtectedStatus(): Unit =
+    if (!isWorkspaceProtected && isSnapshotProtected) {
+      throw new ProtectedDataException("Unable to add protected snapshot to unprotected workspace.")
+    }
+
+  @throws(classOf[UnsupportedPlatformException])
+  @throws(classOf[PlatformBoundaryException])
+  def validatePlatforms(workspacePlatform: WorkspaceCloudPlatform): Unit = {
```
davidangb: see my other comment - it feels slightly off that the `SnapshotReferenceCreationValidator` constructor accepts a `Workspace`, but this method requires a `WorkspaceCloudPlatform` argument … potentially, you can read the platform right from the Workspace object and not need the argument.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,89 @@
...
+    // here as a safeguard against us introducing support for Azure without deliberately addressing the issue of
+    // snapshots by ref across cloud platforms.
+    val acceptableCombinations = Map((SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp))
+
+    if (!acceptableCombinations.get(snapshotPlatform).contains(workspacePlatform)) {
+      throw new PlatformBoundaryException(
+        "Snapshots by reference are not supported across the given cloud boundaries " +
+          s"(snapshot: ${snapshotPlatform.toString.toLowerCase}, workspace: ${workspacePlatform.toString.toLowerCase})."
+      )
+    }
```
davidangb, line 44: one of my favorite features of Scala is its rich pattern-matching. Again, the code you have is fine but to be Scalalike:
```scala
    (snapshotPlatform, workspacePlatform) match {
      case (SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp) => // ok
      case _ =>
        throw new PlatformBoundaryException(
          "Snapshots by reference are not supported across the given cloud boundaries " +
            s"(snapshot: ${snapshotPlatform.toString.toLowerCase}, workspace: ${workspacePlatform.toString.toLowerCase})."
        )
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -47,43 +47,48 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+
+      Try(aggregatedWorkspaceService.getAggregatedWorkspace(rawlsWorkspace, ctx)) match {
+        case Success(wsmWorkspace) =>
+          // if there's an existing WSM workspace, make sure its platform is compatible
+          // with that of the snapshot's dataset.
+          validator.validatePlatforms(wsmWorkspace.getCloudPlatform)
+        case Failure(_) =>
+          // Pre-emptively fail (and prevent creation of a workspace stub here if snapshot is Azure). We can remove this
+          // check once we support Azure-based snapshots by reference.
+          validator.validateSnapshotPlatform()
```
davidangb: if `aggregatedWorkspaceService.getAggregatedWorkspace(rawlsWorkspace, ctx)` is `Failure`, doesn't it indicate a pretty serious failure, like a totally missing workspace? Is there a point to validating the snapshot in such a case, since import would fail no matter what?

davidangb (commented):
a bunch of Scala syntax and idiom feedback, also a couple comments on business logic. Thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Just to clarify, but is your list an *and* or an *or* :)

I want to make sure I'm clear on whether or not this should be happening as a side effect for Azure workspaces (my hunch was no, and that the current behavior, which creates a workspace no matter what, isn't desirable).

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: @nmalfroy I'm trying to determine the cloud platform of a given TDR snapshot here, to decide whether or not to allow it to be referenced by a workspace.

I'm getting this particular `snapshot` from a call to `dataRepoDAO.getSnapshot` from `SnapshotService` which calls this method.

I tried to dig into the TDR code to understand what assumptions I could make here but couldn't get to a definitive understanding of whether I can assume `cloudPlatform` is always set.  Can you advise on if I should treat `cloudPlatform` as non-nullable, and get rid of this assumption that _if it's not present then it must be GCP_?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: You're right this is really gross and where I thought I might be over-engineering... thanks for pulling me back 😆 

By way of explanation, I think this got all bloated because I wasn't sure if I could assume `getCloudPlatform` was always set for every dataset, and wanted to handle the case where there were blanks (defaulted to GCP) co-mingled with non-blanks (which were Azure), and fail in that situation.  Hopefully @nmalfroy can confirm that it's always present and then we can eliminate this!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: Love it, this is exactly the kind of syntactic sugar I was hoping to learn :)

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Maybe this gets clearer/sorted out once I understand the decision logic for when it's appropriate to call `createWorkspace`, but I'll try to explain what I'm doing here a bit better.

I am trying to tease apart the check for the existing workspace from the decision to backfill/create one because I want to make sure we _only_ backfill/create a workspace for valid snapshot requests.

This `Failure` block is what we expect _when we haven't created a workspace yet_

If the snapshot request isn't valid (eg: because it's for an Azure snapshot), it didn't seem right to have side effects and go create a workspace anyway, so I wanted to fail fast here in that case without chugging along and backfilling.

Does that make sense?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: Love it!  Looks like it did the trick.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: I thought this felt clumsy too...I think it's a side effect of me wanting to do some validation on the snapshot platform before having a guaranteed WSM workspace (with its cloudPlatform) in hand.  If we can determine the platform earlier, based on what's in Rawls only, then we don't need the WSM workspace cloudPlatform to decide this.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: `WorkspaceType` looks like it's either `McWorkspace` (which, regretfully, I can no longer read as anything but MickWorkspace, sorry if I've thus infected you), or `RawlsWorkspace`.  I interpreted the `McWorkspace` type as potentially being GCP OR Azure (or in the futuretimes, both).

@aherbst-broad do you have any advice here?  I thought using `AggregatedWorkspaceService` might be the way to get from WSM a determination of whether a workspace is for Azure vs. GCP, but maybe I can shortcut it by just looking at `WorkspaceType` and obviate the need for `AggregatedWorkspaceService` entirely?

(For extra context, I'm trying to determine workspace cloud platform to compare to the cloud platform of a TDR snapshot, and to make decisions around whether to allow a snapshot by ref)

aherbst-broad: It'd preferable to leverage the AggregatedWorkspaceService here and rely on the truth from WSM if present. That is, the AggregatedWorkspaceService should Do The Right Thing™ (or be made to do so) and give you back a workspace with maybe the needed cloud platform information from WSM. If there is WSM info present, use that for decision making, if not, assume GCP. This would keep us from having to migrate one more thing as we move toward a WSM-based future.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,92 @@
...
+class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val snapshot: SnapshotModel) {
+
+  // TODO: get this information from a more authoritative source rather than relying on the hardcoded bucket prefix
+  private def isWorkspaceProtected: Boolean = workspaceContext.bucketName.startsWith("fc-secure")
+
+  private def isSnapshotProtected: Boolean = snapshot.getSource.asScala.exists(_.getDataset.isSecureMonitoringEnabled)
+
+  private def snapshotPlatform: SnapshotCloudPlatform = {
+    val snapshotSourceDatasets = snapshot.getSource.asScala
+    val datasetCloudPlatforms = snapshotSourceDatasets.flatMap(src => Option(src.getDataset.getCloudPlatform)).toSet
```
davidangb: if the reality is that `getCloudPlatform` can return null, and if it's null we should assume GCP, I'd suggest handling the defaulting up here:
```scala
    val datasetCloudPlatforms = snapshotSourceDatasets
      .map(src => Option(src.getDataset.getCloudPlatform).getOrElse(SnapshotCloudPlatform.GCP))
      .toSet
```
this could eliminate complexity below.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
davidangb: ah, I didn't think about blanks and non-blanks. Hopefully if any blanks exist they are Java `null` and not `""`!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: good call on using `AggregatedWorkspaceService` here to future-proof the code - that makes sense. 

@jladieu what I'm really trying to avoid here is that if the current workspace is an Azure workspace, we're going to add a call to WSM and then immediately fail the import because we don't support Azure. What do you think about refactoring the code so that if the _snapshot_ being imported is an Azure snapshot, we fail fast before reaching out to WSM to determine the workspace's cloud?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: Responding to this comment and the subesequent comment at once … what I believe the flow should be is:
1. Does the workspace exist _in Rawls_? If not, reject the request. This already happens inside `getV2WorkspaceContextAndPermissions()`, which is the first line of `SnapshotService.createSnapshot()`.
2. Is the snapshot-being-imported an Azure snapshot? If so, reject the request.
3. Is the workspace an Azure workspace? If so, reject the request.
4. Does the workspace exist _in WSM_? If not, create it.
5. Create the snapshot reference.

Specific to the `Success` or `Failure` of `aggregatedWorkspaceService.getAggregatedWorkspace()` - if the workspace is of type `WorkspaceType.RawlsWorkspace`, you'll get a `Success` as long as the workspace exists _in Rawls_, even if that workspace doesn't exist _in WSM_. So I don't think it is a reliable test for step 4 above.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: My list is an "and" - hopefully the comment I added just before this one helps to clarify. And you're correct, if this is an Azure workspace we should not make any attempt to create a stub in WSM.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Ok I totally understand your concern about adding an unnecessary call that we could avoid by failing early; I'll need to re-organize the flow a bit here but your comment below provides a good roadmap!  Thanks for digging into the detail on this.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: Discussed this in Slack and can confirm `cloudPlatform` should always be set and its absence would be considered a bug.  This should let me unroll a bunch of this extra complexity.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceServiceSpec.scala
```diff
@@ -31,33 +33,33 @@ import scala.language.postfixOps
...
-    "test-namespace",
-    "test-name",
-    "aWorkspaceId",
-    "aBucket",
-    Some("workflow-collection"),
-    new DateTime(),
-    new DateTime(),
-    "test",
-    Map.empty
+    namespace = "test-namespace",
```
jladieu: These extra param names are just to aid readability.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -87,14 +80,33 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+      )
+    )
+      .thenReturn(
+        new SnapshotModel()
+          .id(java.util.UUID.randomUUID())
+          .name("snapshot")
+          .description("snapshot description")
+          .source(
+            java.util.List
+              .of(new SnapshotSourceModel().dataset(new DatasetSummaryModel().cloudPlatform(SnapshotCloudPlatform.GCP)))
```
jladieu, line 113: Now that we actively check the snapshot -> dataset -> cloudPlatform, this object needs to be adequately hydrated for all the existing tests to continue passing.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -135,7 +147,7 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
       val mockWorkspaceManagerDAO = defaultMockWorkspaceManagerDao()
       when(
         mockWorkspaceManagerDAO.getWorkspace(any[UUID], any[RawlsRequestContext])
-      ).thenAnswer(_ => new ApiException(404, "Workspace does not exist"))
+      ).thenAnswer(_ => throw new ApiException(404, "Workspace does not exist"))
```
jladieu, line 164: I bumped into some type problems here that precipitated this change, I don't know exactly why they only showed up after I started moving things around though.  But specifically, my interpretation of the compiler error was that it was treating the `ApiException` as the return value rather than as a thrown value, and the compiler couldn't cast it to the expected return type.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -161,13 +173,13 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
       )
 
       // assert that the service checked to see if the workspace exists
-      verify(mockWorkspaceManagerDAO, times(1)).getWorkspace(
```
jladieu, line 164: just dropping this `times(1)` throughout, as I noticed elsewhere we omit this and it's redundant with the default behavior of `verify` without the 2nd arg

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
jladieu: This was actually a problematic change in that it was causing `scalafmt` to fail after pushing up the commit.  I'm not sure why, but updating my "Class count to use import with '_'" made the IDE stop smooshing the imports.
image

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Reorganized this quite a bit after incorporating input from TDR & Workspaces.  We should now fail fast on unsupported platforms without incurring a hit to WSM, and `AggregatedWorkspaceService` is able to support GCP Workspaces (even Rawlsy ones!) as a result of https://github.com/broadinstitute/rawls/pull/2563

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: This should be all set now, PTAL again :)

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: See resulting `WrappedSnapshot` for the simplified goodness.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: See newly simplified `WrappedSnapshot` which de-uglies things a lot.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: I hope this is in better shape now, see https://github.com/broadinstitute/rawls/pull/2556#discussion_r1346170905

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockWorkspaceManagerDAO.scala
```diff
@@ -23,7 +28,8 @@ class MockWorkspaceManagerDAO(
 
   val references: TrieMap[(UUID, UUID), DataRepoSnapshotResource] = TrieMap()
 
-  def mockGetWorkspaceResponse(workspaceId: UUID) = new WorkspaceDescription().id(workspaceId)
+  def mockGetWorkspaceResponse(workspaceId: UUID) =
+    new WorkspaceDescription().id(workspaceId).gcpContext(new GcpContext().projectId("some-project-id"))
```
jladieu, line 32: Now that we actively check the snapshot -> dataset -> cloudPlatform, this object needs to be adequately hydrated for all the existing tests to continue passing.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+
+// A basic wrapper class to simplify interaction with the SnapshotModel from TDR
+class WrappedSnapshot(snapshot: SnapshotModel) {
+
+  // While getSource returns a list of SnapshotSourceModels, this is an artifact and we can make an assumption that
+  // a given snapshot will have exactly one source that will always have a dataset.  Violations of this assumption
+  // indicate a bug, in which case, crash and burn.
+  private def snapshotDataset(): DatasetSummaryModel = getOnlyElement(snapshot.getSource.asScala.toList).getDataset
```
davidangb, line 14: is this true, that snapshots expect to have only one source dataset? If that's the case, it might simplify code in Import Service and Terra UI too.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  private def getOnlyElement[A](list: List[A]): A = list match {
+    case head :: Nil => head
```
davidangb: nit: since `head` is a method of `Iterable`, I suggest using a different variable name here to avoid confusion, like:
```suggestion
    case x :: Nil => x
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
```diff
@@ -0,0 +1,23 @@
...
+    case _           => throw new NoSuchElementException(s"Expected exactly one element, but found ${list.size}")
+  }
+
+  def isProtected: Boolean = snapshotDataset().isSecureMonitoringEnabled()
+
+  def platform: CloudPlatform = snapshotDataset().getCloudPlatform
```
davidangb: When calling Java methods from Scala, I'm always cautious about nulls. I don't know how much we can trust these return values, you may want to do something like:
```suggestion
  def isProtected: Boolean = Option(snapshotDataset().isSecureMonitoringEnabled)
    .getOrElse(
      throw new RawlsException(
        s"Snapshot ${snapshot.getId}'s DatasetSummaryModel had null value for secure monitoring; cannot continue"
      )
    )

  def platform: CloudPlatform = Option(snapshotDataset().getCloudPlatform)
    .getOrElse(
      throw new RawlsException(
        s"Snapshot ${snapshot.getId}'s DatasetSummaryModel had null value for cloud platform; cannot continue"
      )
    )
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -47,43 +47,49 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+      try {
+        // if there's an existing WSM workspace, make sure its platform is compatible
+        // with that of the snapshot's dataset.
+        val wsmWorkspace = aggregatedWorkspaceService.fetchAggregatedWorkspace(rawlsWorkspace, ctx)
+        snapshotValidator.validateWorkspacePlatformCompatibility(wsmWorkspace.getCloudPlatform)
+      } catch {
+        case _: AggregateWorkspaceNotFoundException =>
+          // create the stub workspace in WSM if it does not already exist
+          workspaceManagerDAO.createWorkspace(wsid, rawlsWorkspace.workspaceType, ctx)
       }
```
davidangb: thinking out loud … 

If this is a GCP workspace, and if the stub workspace does not exist, `fetchAggregatedWorkspace` will throw `AggregateWorkspaceNotFoundException` and this code will never call `snapshotValidator.validateWorkspacePlatformCompatibility(…)`.

In our current business logic, this is fine, because we have already short-circuited if we have an Azure snapshot. I wonder if it this behavior could bite us in the future. I am not coming up with the use case, but it feels wiggly.

You could do something like:
```scala
      val cloudPlatform = try {
        // if there's an existing WSM workspace, make sure its platform is compatible
        // with that of the snapshot's dataset.
        val wsmWorkspace = aggregatedWorkspaceService.fetchAggregatedWorkspace(rawlsWorkspace, ctx)
        wsmWorkspace.getCloudPlatform
      } catch {
        case _: AggregateWorkspaceNotFoundException =>
          // create the stub workspace in WSM if it does not already exist
          workspaceManagerDAO.createWorkspace(wsid, rawlsWorkspace.workspaceType, ctx)
          WorkspaceCloudPlatform.Gcp
      }

      snapshotValidator.validateWorkspacePlatformCompatibility(cloudPlatform)
```

davidangb (commented):
A handful of more comments, this is very close!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
jladieu: Yes, from discussing this with @nmalfroy, there will only ever be one `source` dataset, so while it's a List, it'll always be a singleton.  And the `cloudPlatform` should always be set on it.

If TDR ever _changes_ this assumption, as in, make it something other than exactly one source, they'll do so loudly!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
jladieu: `cloudPlatform` should always be present, so I got to get rid of this extra complexity

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
jladieu: Nice, fixed this in aec9590c3cf809713fb67c12388f57b4e4663214

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
jladieu: Fixed this in aec9590c3cf809713fb67c12388f57b4e4663214, although I try to avoid single-character variable names unless I'm really cramped for space...so I chose `onlyElement`

That said, I don't _actually_, _really_ understand what the variable is referring to here because we're deep in quirky Scala stuff, so if that name doesn't actually make sense please let me know!!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Thanks so much for squinting at this a bit more...and you're right, I do think some stuff could slip through later on if we make changes to earlier validation.

I fixed this in fa0de1036a6daf61013e266179f03de49b97ffc6 by adding a check on an assumed platform of `Gcp` right before creating the stub.  This comes with the benefit of making it so we don't create a stub for an invalid request.

That said, I think this is almost impossible to actually test without also removing/changing the logic that will prevent us from getting here, so I am going with the low-cov/faith approach that this works.  You ok with that?

The alternative is I could create a really crazy mock that just bypasses/skips earlier validations and makes sure this still triggers, but I thought that might be tricky to actually do.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
davidangb: Looks like you've got it right. Scala implements `List` as a collection of elements terminated by a `Nil`. If you were working with a `List[String]`, you could have:
| List length | `List` constructor | pattern-match |
|--------|--------|--------|
| 0 | List() | Nil |
| 1  | List("foo") | x :: Nil |
| 2  | List("foo", "bar") | x :: y :: Nil |

you can also pattern-match against `List`s by looking at the head (one element) and tail (N elements). `List("foo", "bar", "baz", "qux")` will pattern match against `x :: y`, where `x` is a `String` of "foo" and `y` is a `List("bar", "baz", "qux")`.

davidangb (approved):
nice, thanks for sticking with this!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
jladieu: 🤯

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -47,43 +48,51 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
-      val wsid = workspaceContext.workspaceIdAsUUID // to avoid UUID parsing multiple times
-      // create the stub workspace in WSM if it does not already exist
-      if (!workspaceStubExists(wsid, ctx)) {
-        workspaceManagerDAO.createWorkspace(wsid, workspaceContext.workspaceType, ctx)
+    ).flatMap { rawlsWorkspace =>
+      val wsid = rawlsWorkspace.workspaceIdAsUUID // to avoid UUID parsing multiple times
+      val snapshot =
+        new WrappedSnapshot(dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken))
+      val snapshotValidator = new SnapshotReferenceCreationValidator(rawlsWorkspace, snapshot)
+      val aggregatedWorkspaceService = new AggregatedWorkspaceService(workspaceManagerDAO)
```
calypsomatic: Is this creating a new `AggregatedWorkspaceService` for every snapshot creation?  It doesn't seem to rely on any specific input - can we pull it up to the class level?

calypsomatic (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Done in 1202723762cbaeae8115211c94fe8fe1f730b05a

_Also, good call!_

jladieu:
image

PR #2563, opened 2023-10-03
AJ-1309: Allow GCP aggregated workspace

jladieu:

This PR includes three refactoring commits which should lower the barrier to using `AggregatedWorkspaceService` in `SnapshotService`.

---

**PR checklist**

- [x] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala). _not doing this as it's an internal only change_
- [x] Inform other teams of any substantial changes via Slack and/or email _(n/a, this is a minor refactoring change that unlocks more functional changes)_

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -6,7 +6,13 @@ import bio.terra.profile.model.ProfileModel
 import bio.terra.workspace.api.{ReferencedGcpResourceApi, ResourceApi, WorkspaceApi}
 import bio.terra.workspace.client.ApiClient
 import bio.terra.workspace.model._
-import org.broadinstitute.dsde.rawls.model.{DataReferenceDescriptionField, DataReferenceName, RawlsRequestContext}
+import org.broadinstitute.dsde.rawls.model.WorkspaceType.WorkspaceType
+import org.broadinstitute.dsde.rawls.model.{
+  DataReferenceDescriptionField,
+  DataReferenceName,
+  RawlsRequestContext,
+  WorkspaceType
+}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 
 import java.util.UUID
@@ -50,8 +56,16 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
   override def getWorkspace(workspaceId: UUID, ctx: RawlsRequestContext): WorkspaceDescription =
     getWorkspaceApi(ctx).getWorkspace(workspaceId, null) // use default value for role
 
-  override def createWorkspace(workspaceId: UUID, ctx: RawlsRequestContext): CreatedWorkspace =
-    getWorkspaceApi(ctx).createWorkspace(new CreateWorkspaceRequestBody().id(workspaceId))
+  override def createWorkspace(workspaceId: UUID,
+                               workspaceType: WorkspaceType,
+                               ctx: RawlsRequestContext
+  ): CreatedWorkspace = {
+    val stage = workspaceType match {
+      case WorkspaceType.RawlsWorkspace => WorkspaceStageModel.RAWLS_WORKSPACE
+      case WorkspaceType.McWorkspace    => WorkspaceStageModel.MC_WORKSPACE
+    }
+    getWorkspaceApi(ctx).createWorkspace(new CreateWorkspaceRequestBody().id(workspaceId).stage(stage))
+  }
 
   override def createWorkspaceWithSpendProfile(workspaceId: UUID,
                                                displayName: String,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.rawls.snapshot
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.workspace.model._
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
 import org.broadinstitute.dsde.rawls.dataaccess.{SamDAO, SlickDataSource}
@@ -19,6 +18,7 @@ import org.broadinstitute.dsde.rawls.model.{
   WorkspaceName
 }
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, WorkspaceSupport}
+import org.broadinstitute.dsde.rawls.{RawlsException, RawlsExceptionWithErrorReport}
 
 import java.util.UUID
 import scala.annotation.tailrec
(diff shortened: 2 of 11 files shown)


model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -759,6 +757,8 @@ case class AzureManagedAppCoordinates(tenantId: UUID,
                                       landingZoneId: Option[UUID] = None
 )
 
+case class GcpProjectInfo(projectId: GoogleProjectId)
```
jladieu: Does this merit publishing a new `rawls-model`?

jladieu:
I'd welcome advice on whether and what kind of verification would be worthwhile above and beyond the unit tests.  The changed code is used by `ListWorkspaces` and `GetWorkspace` so I'm assuming poking around the UI in Workspaces is a good place to start.

jladieu:
After discussing with @davidangb and realizing that, while by design `WorkspaceType` and `Azure/GCP` are orthogonal, in practice `WorkspaceType.RAWLS` === `GCP` and `WorkspaceType.MCWorkspace` === `Azure` -- so these changes amount to futureproofing only.  That said, it still seemed like a net positive, so I'm rolling ahead with these.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -72,7 +72,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
 
   // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
   // We'll have to use this workaround for identifying protected status
-  def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
+  private def validateProtectedStatus(workspaceContext: Workspace, snapshot: NamedDataRepoSnapshot): Unit =
```
jladieu, line 75: Just opportunistic warning tidying.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -108,7 +108,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     }
 
   // AS-787 - rework the data so that it's in the same place in the JSON with a list and get snapshot responses
-  def massageSnapshots(references: ResourceList): SnapshotListResponse = {
+  private def massageSnapshots(references: ResourceList): SnapshotListResponse = {
```
jladieu, line 111: Just opportunistic warning tidying.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -127,7 +127,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
       // if we fail with a 404, it means we have no stub in WSM yet. This is benign and functionally equivalent
       // to having no references, so return the empty list.
       case Failure(ex: bio.terra.workspace.client.ApiException) if ex.getCode == 404 =>
-        new SnapshotListResponse(Seq.empty[DataRepoSnapshotResource])
+        SnapshotListResponse(Seq.empty[DataRepoSnapshotResource])
```
jladieu, line 130: Just opportunistic warning tidying.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
aherbst-broad: I think this is only being consumed internally by the AggregatedWorkspaceService, so I'd say it's not necessary.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
```diff
@@ -38,46 +40,42 @@ class AggregatedWorkspaceService(workspaceManagerDAO: WorkspaceManagerDAO) exten
...
-                        )
-                          .getOrElse(List.empty)
-                      )
-                    )
-                  )
-                  .getOrElse(List.empty)
+                convertPolicies(wsmInfo)
               )
-            case None =>
+            case (Some(gcpContext), None) =>
```
aherbst-broad: 👍 this should put us in a better place for real MC GCP workspaces.

aherbst-broad (approved):
This generally makes sense, thanks for doing this.

jladieu:
@aherbst-broad @davidangb I added a third commit to this PR and reset the reviews; this third commit provides the access path for `SnapshotService` to fetch a Rawls based workspace using the 'unoptimized' version of what used to be named `getAggregatedWorkspace` and is now `optimizedFetchAggregatedWorkspace`.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -50,8 +56,16 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
-    getWorkspaceApi(ctx).createWorkspace(new CreateWorkspaceRequestBody().id(workspaceId))
+  override def createWorkspace(workspaceId: UUID,
+                               workspaceType: WorkspaceType,
+                               ctx: RawlsRequestContext
+  ): CreatedWorkspace = {
+    val stage = workspaceType match {
+      case WorkspaceType.RawlsWorkspace => WorkspaceStageModel.RAWLS_WORKSPACE
+      case WorkspaceType.McWorkspace    => WorkspaceStageModel.MC_WORKSPACE
+    }
+    getWorkspaceApi(ctx).createWorkspace(new CreateWorkspaceRequestBody().id(workspaceId).stage(stage))
```
davidangb, line 67: My first reaction: how did this ever work for MC_WORKSPACEs previously????

Looking deeper at it, `MultiCloudWorkspaceService` never calls this `createWorkspace`; it calls `createWorkspaceWithSpendProfile` instead. I wonder if these two methods are worth combining. That's out of scope for this PR (we could yak-shave forever).

aherbst-broad (approved):

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
davidangb: `GoogleProjectId` is already a wrapper around `String` for type safety - do we actually need this additional layer of wrapper?

If we do need it, I note that we don't have a `JsonFormat` for `GcpProjectInfo`, so we would likely fail to de/serialize it in JSON requests/responses. I don't think `GcpProjectInfo` is used anywhere that needs serdes, so this is ok, just raising it to keep an eye on

davidangb (approved):
:+1: and thank you for this. Two comments inline, neither of which should block merging.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
jladieu: I'd love to avoid touching this file, although I thought maybe we'd eventually (YAGNI?) cram more stuff into `GcpProjectInfo`...is this potentially a landmine for the next person if I leave it in here without `JsonFormat`?

jladieu: Ok dropped it...very happy with the outcome!  Thanks for the suggestion.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
jladieu: Agreed to punt this....

PR #2568, opened 2023-10-05
[WOR-1245] Don't issue FastPass grants when a workspace's bucket is migrating

marctalbott:
Ticket: [WOR-1245](https://broadworkbench.atlassian.net/browse/WOR-1245)
* Before issuing a FastPass grant, check if the workspace has an in-progress multiregion bucket migration and don't issue the FastPass grant if it does.
* The ticket also mentions adding resilience to this when restoring bucket IAM, but this would require more wide-reaching changes in Rawls. Since I was able to manually verify that this will prevent FastPass grants on a migrating workspace, this feels like overkill and I've decided not to include it in this PR.

---

**PR checklist**

[WOR-1245]: https://broadworkbench.atlassian.net/browse/WOR-1245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
@@ -8,7 +8,14 @@ import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.config.FastPassConfig
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadWriteAction}
 import org.broadinstitute.dsde.rawls.dataaccess.{GoogleServicesDAO, SamDAO, SlickDataSource}
-import org.broadinstitute.dsde.rawls.fastpass.FastPassService._
+import org.broadinstitute.dsde.rawls.fastpass.FastPassService.{
+  openTelemetryTags,
+  policyBindingsQuotaLimit,
+  removeFastPassGrants,
+  RemovalFailure,
+  SAdomain,
+  UserAndPetEmails
+}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
```diff
@@ -296,6 +299,14 @@ class FastPassService(protected val ctx: RawlsRequestContext,
     dataSource
       .inTransaction { implicit dataAccess =>
         for {
+          isMigrating <- dataAccess.multiregionalBucketMigrationQuery.isMigrating(workspace)
+          _ = if (isMigrating) {
+            throw new RawlsExceptionWithErrorReport(
```
cahrens, line 304: This won't cause Sentry noise, will it?

cahrens (approved):

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala
marctalbott: From what I can tell it won't. This gets caught further down on [line 336](https://github.com/broadinstitute/rawls/blob/07a738b79fc4fa0c1d516d59ea1e5c38707016bc/core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassService.scala#L336) along with other FastPass failures. FastPass failures in general aren't getting sent to Sentry, so these won't either. If I'm wrong and these do end up getting sent to Sentry, it should be infrequent at the very least and we can always mute it if it gets too noisy.

PR #2570, opened 2023-10-06
AJ-1309: Disable crossplatform snapshots by ref

jladieu:

Update `SnapshotService` to enforce stricter validation around the `cloudPlatform` of the snapshot & workspace.  Basically: disallow Azure-based snapshots by reference entirely, and futureproof the code to ensure that if we _do_ start supporting Azure snapshots by reference that we'll have to bake in support at the appropriate touchpoints in this area of the code.

This is a re-attempt of https://github.com/broadinstitute/rawls/pull/2556, which was reverted in: https://github.com/broadinstitute/rawls/pull/2569

The new attempt has been updated to include 438c6f61aaea7cce421cc9d6c4d78c19376ce947 to better handle legacy Rawls workspaces.

d07d524951388c3350786f7790297dbab8422e83 is a pure re-apply of the first PR, so it should represent the same set of changes as reviewed in https://github.com/broadinstitute/rawls/pull/2556

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
@@ -1,0 +1,54 @@
+package org.broadinstitute.dsde.rawls.snapshot
+
+import bio.terra.datarepo.model.{CloudPlatform => SnapshotCloudPlatform}
+import org.broadinstitute.dsde.rawls.RawlsException
+import org.broadinstitute.dsde.rawls.model.WorkspaceCloudPlatform.WorkspaceCloudPlatform
+import org.broadinstitute.dsde.rawls.model.{Workspace, WorkspaceCloudPlatform}
+
+object SnapshotReferenceCreationValidator {
+  def constructor(workspaceContext: Workspace, snapshot: WrappedSnapshot): SnapshotReferenceCreationValidator =
+    new SnapshotReferenceCreationValidator(workspaceContext, snapshot)
+}
+
+// thrown to disallow creation of a snapshot reference on an unsupported platform
+class UnsupportedPlatformException(message: String) extends RawlsException(message)
+// thrown to disallow creation of a snapshot reference across cloud boundaries
+class PlatformBoundaryException(message: String) extends RawlsException(message)
+// thrown to disallow creation of a snapshot reference across a protected data boundary
+class ProtectedDataException(message: String) extends RawlsException(message)
+
+class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val snapshot: WrappedSnapshot) {
+
+  // Ideally this would rely on Terra Policy Service, but until TPS is enabled for GCP
+  // We'll have to use this workaround for identifying protected status.
+  @throws(classOf[ProtectedDataException])
+  def validateProtectedStatus(): Unit =
+    if (!isWorkspaceProtected && snapshot.isProtected) {
+      throw new ProtectedDataException("Unable to add protected snapshot to unprotected workspace.")
+    }
+
+  // Throws an exception when the given snapshot cannot be referenced by the given workspace due to crossing an
+  // unsupported platform boundary.
+  @throws(classOf[PlatformBoundaryException])
+  def validateWorkspacePlatformCompatibility(workspacePlatform: WorkspaceCloudPlatform): Unit =
+    // Defining all acceptable combinations is overkill when all we really need to do is prevent anything Azure, but
+    // this is here as a safeguard against us introducing support for Azure without deliberately addressing the issue of
+    // snapshots by ref across cloud platforms.
+    (snapshot.platform, workspacePlatform) match {
+      case (SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp) => // ok
+      case _ =>
+        throw new PlatformBoundaryException(
+          "Snapshots by reference are not supported across the given cloud boundaries (" +
+            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform})."
+        )
+    }
+
+  @throws(classOf[UnsupportedPlatformException])
+  def validateSnapshotPlatform(): Unit =
+    if (snapshot.platform == SnapshotCloudPlatform.AZURE) {
+      throw new UnsupportedPlatformException("Snapshots by reference are not supported for Azure datasets.")
+    }
+
+  // TODO: get this information from a more authoritative source rather than relying on the hardcoded bucket prefix
+  private def isWorkspaceProtected: Boolean = workspaceContext.bucketName.startsWith("fc-secure")
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
@@ -1,0 +1,34 @@
+package org.broadinstitute.dsde.rawls.snapshot
+
+import bio.terra.datarepo.model.{CloudPlatform, DatasetSummaryModel, SnapshotModel}
+import org.broadinstitute.dsde.rawls.RawlsException
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+
+// A basic wrapper class to simplify interaction with the SnapshotModel from TDR
+class WrappedSnapshot(snapshot: SnapshotModel) {
+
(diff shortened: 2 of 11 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceServiceSpec.scala
```diff
@@ -31,7 +31,7 @@ import scala.language.postfixOps
 
 class AggregatedWorkspaceServiceSpec extends AnyFlatSpec with MockitoTestUtils {
 
-  private val rawlsWorkspace = Workspace(
+  private val legacyRawlsWorkspace = Workspace(
```
jladieu, line 34: Just renamed this and `mcRawlsWorkspace` to clarify that these are the Rawls side of things (as opposed to being WSM workspaces, which are individually mocked in the specs).

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceServiceSpec.scala
```diff
@@ -102,14 +104,19 @@ class AggregatedWorkspaceServiceSpec extends AnyFlatSpec with MockitoTestUtils {
...
     val svc = new AggregatedWorkspaceService(wsmDao)
 
-    val aggregatedWorkspace = svc.fetchAggregatedWorkspace(mcWorkspace, defaultRequestContext)
+    val aggregatedWorkspace = svc.fetchAggregatedWorkspace(mcRawlsWorkspace, defaultRequestContext)
 
-    aggregatedWorkspace.baseWorkspace shouldBe mcWorkspace
-    aggregatedWorkspace.googleProjectId shouldBe Some(GoogleProjectId("project-id"))
+    aggregatedWorkspace.baseWorkspace shouldBe mcRawlsWorkspace
     aggregatedWorkspace.getCloudPlatform shouldBe WorkspaceCloudPlatform.Gcp
+    aggregatedWorkspace.azureCloudContext shouldBe None
```
jladieu, line 118: Reordered asserts throughout to confirm which cloudPlatform first (which IMO is the most important detail), then azure details, then GCP details.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceServiceSpec.scala
```diff
@@ -119,7 +126,7 @@ class AggregatedWorkspaceServiceSpec extends AnyFlatSpec with MockitoTestUtils {
     val svc = new AggregatedWorkspaceService(wsmDao)
 
     intercept[AggregateWorkspaceNotFoundException] {
-      svc.optimizedFetchAggregatedWorkspace(mcWorkspace, defaultRequestContext)
+      svc.fetchAggregatedWorkspace(mcRawlsWorkspace, defaultRequestContext)
```
jladieu, line 129: oops, in the initial PR I meant to be only testing the `optimized` method in the below stanza but missed switching these over

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1388,12 +1395,14 @@ class WorkspaceServiceSpec
...
     )
     when(services.workspaceManagerDAO.getWorkspace(any[UUID], any[RawlsRequestContext])).thenReturn(
-      new WorkspaceDescription().azureContext(
-        new AzureContext()
-          .tenantId("fake_tenant_id")
-          .subscriptionId("fake_sub_id")
-          .resourceGroupId("fake_mrg_id")
-      )
+      new WorkspaceDescription()
+        .stage(WorkspaceStageModel.MC_WORKSPACE)
```
jladieu, line 1399: @aherbst-broad I had to introduce `stage` in a bunch of different spots throughout this file (and various mocks), which is making me a little nervous about the potential scope of impact of this change.  Looking for validation that this is _ok_ and makes things reflect the expected reality, rather than me shoving an assumption where it doesn't belong.

davidangb (approved):
:+1: assuming this looks good in your BEE

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
aherbst-broad: I think this makes sense given the changes in the associated case statement in the AggregatedWorkspaceService.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,54 @@
...
+    (snapshot.platform, workspacePlatform) match {
+      case (SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp) => // ok
+      case _ =>
+        throw new PlatformBoundaryException(
+          "Snapshots by reference are not supported across the given cloud boundaries (" +
+            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform})."
+        )
+    }
+
+  @throws(classOf[UnsupportedPlatformException])
```
aherbst-broad, line 46: Interesting, I forgot about this annotation 👍

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.broadinstitute.dsde.rawls.model.WorkspaceCloudPlatform.WorkspaceCloudPlatform
+import org.broadinstitute.dsde.rawls.model.{Workspace, WorkspaceCloudPlatform}
+
+object SnapshotReferenceCreationValidator {
+  def constructor(workspaceContext: Workspace, snapshot: WrappedSnapshot): SnapshotReferenceCreationValidator =
+    new SnapshotReferenceCreationValidator(workspaceContext, snapshot)
+}
+
+// thrown to disallow creation of a snapshot reference on an unsupported platform
+class UnsupportedPlatformException(message: String) extends RawlsException(message)
```
aherbst-broad, line 14: I like these granular exception types

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
```diff
@@ -0,0 +1,34 @@
...
+  private def snapshotDataset(): DatasetSummaryModel = getOnlyElement(snapshot.getSource.asScala.toList).getDataset
+
+  private def getOnlyElement[A](list: List[A]): A = list match {
+    case onlyElement :: Nil => onlyElement
+    case _                  => throw new NoSuchElementException(s"Expected exactly one element, but found ${list.size}")
+  }
+
+  def isProtected: Boolean = Option(snapshotDataset().isSecureMonitoringEnabled)
+    .getOrElse(
+      throw new RawlsException(
```
aherbst-broad, line 23: Any way to get coverage on this exception handling?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/WrappedSnapshot.scala
jladieu: Working on it now...actually this was a good callout because as I started to add a test for `isSecureMonitoringEnabled`, it uncovered the fact that `isSecureMonitoringEnabled` just defaults to `false` if it's not explicitly set, due to it being a bool.  This is clearly not ideal...but I'm not exactly sure the best way to do anything about it, Scala wiz advice welcome here! cc @davidangb

jladieu: @aherbst-broad @davidangb added one more commit with the additional test coverage requested

You'll note it includes an `ignore`d spec with the _actual, desirable_ behavior, and a copout 'working as implemented' spec that clarifies current behavior, with great disappointment.  I kind of want to roll forward with this anyway because I don't think I'm _introducing_ the described problem, just surfacing it.

Final thoughts, in the form of thumbs up or down, would be welcome.

jladieu: Discussed this with @davidangb and he helped me realize I was just missing a key part of the fixture, which was to set the boolean explicitly to `null` (otherwise it'd use the default value of the instance initializers).

aherbst-broad (approved):

PR #2576, opened 2023-10-11
Use json logformat

jladieu:
Experimenting with logback in rawls

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1,4 +1,5 @@
 package org.broadinstitute.dsde.rawls.workspace
+//import net.logstash.logback.argument.StructuredArguments
 
 import akka.http.scaladsl.model.StatusCodes
 import akka.stream.Materializer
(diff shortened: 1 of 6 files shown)


jladieu:
Just enabling logback seemed to have the desired effect in GCP console:

Before (in prod), many messages per stacktrace:
image

After Logback (with faked error report, in bee), only one message per error report:
image

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -892,7 +894,22 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       }
     }
 
-  def getTags(query: Option[String], limit: Option[Int] = None): Future[Seq[WorkspaceTag]] =
+  private def dumpLogs(): Unit = {
```
jladieu, line 897: This is a fake crash I created to simulate an error when querying the tags endpoint; I did this because I wasn't able to get logging to work in the test environment (it's stuck on the NOP logger despite my appeals).

Also, I've got history with slf4j warnings, which is that I've never in my life been able to get it to work from a not working state....so I gave up my attempts after I saw it was involved.

davidangb:
A couple links to check out:
* How Sam logs in json format: https://github.com/broadinstitute/sam/blob/2015f5b67e6614285767f19a52b9a21466a4e0cf/src/main/resources/logback.xml#L11
* How Leo logs in json format: https://github.com/DataBiosphere/leonardo/blob/17342f265026378a67da613b0e2fe8b13041ffc9/http/src/main/resources/logback.xml#L4
* How Logit's logstash is configured to accept different formats: https://github.com/broadinstitute/workbench-elk-metrics/blob/master/logstash/logstash.conf

jladieu:
After looking at the downstream impact of enabling JSON formatting on Kibana messaging, I opted not to pursue introducing StructuredFormatting here, see [doc](https://docs.google.com/document/d/1DPjzCClHdrSeNGwviVxySO06t4CXrCI6_3XQIWLpySg/edit#heading=h.h5blczaf11ps) for details.

PR #2580, opened 2023-10-13
AJ-1350: Add `LogSpec` to confirm existing logging.

jladieu:
Ticket: [AJ-1350](https://broadworkbench.atlassian.net/browse/AJ-1350)

This adds a spec to be used as a baseline for future iteration/experimentation with logging appenders & encoders.

[AJ-1350]: https://broadworkbench.atlassian.net/browse/AJ-1350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/LogSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/LogSpec.scala
@@ -1,0 +1,64 @@
+package org.broadinstitute.dsde.rawls
+
+import akka.http.scaladsl.model.StatusCodes
+import ch.qos.logback.core.OutputStreamAppender
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.rawls.model.ErrorReport
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito.{doAnswer, spy}
+import org.scalatest.Assertions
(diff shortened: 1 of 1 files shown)


calypsomatic (approved):

nawatts (approved):

davidangb (approved):
I see how this is helpful if we are making changes to logging formats. On the surface, it's just asserting on the behavior of third-party libraries, though - maybe add a comment to explain?

I'm fine with the implementation as-is, and don't want to open more rabbit holes, but I'm also curious if an approach like this would work too: https://stackoverflow.com/questions/2169330/java-junit-capture-the-standard-input-output-for-use-in-a-unit-test

jladieu:
2ee20dde8c44b88827d987784dca4f9ca18863ac adds a clarifying comment for _why_ this spec even exists

jladieu:
@davidangb Thanks for the link and suggestion!  I think if I had known about this approach before digging into the log4j rabbit hole I'd have adopted it.  As it stands though, I think I like the current result which lets you specify mocking at the appender & class level and not rely on an assumption that everything just goes to Sysout, so I think I'll keep it as is...but pocket that link so I know how to do that in the future =)

PR #2589, opened 2023-10-18
[WOR-1283] store an error message when workspace deletion fails

marctalbott:
Ticket: [WOR-1283](https://broadworkbench.atlassian.net/browse/WOR-1283)
* Store an error message on the workspace record when workspace deletion fails

---

**PR checklist**

[WOR-1283]: https://broadworkbench.atlassian.net/browse/WOR-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunner.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunner.scala
@@ -40,8 +40,8 @@ import scala.util.{Failure, Success}
   * 2. Fire a request to Leonardo to delete any applications
   * 3. Fire a request to Leonardo to delete any runtimes
   * 4. Fire a request to WSM to delete the WSM workspace (including any cloud resources like containers, batch pools, etc.)
-  * 5. Delete the rawls record 
-  * 
+  * 5. Delete the rawls record
+  *
   * If a downstream operation fails or times out during deletion, we mark the workspace as `DeleteFailed` and finish the job
   */
 class WorkspaceDeletionRunner(val samDAO: SamDAO,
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -30,6 +30,11 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
       access.workspaceQuery.updateState(workspaceId, state)
     }
 
+  def setFailedState(workspaceId: UUID, state: WorkspaceState, message: String): Future[Int] =
```
marctalbott, line 33: I went this route over adding an optional `message` param to `updateState` with the hope that it would make it clear to us in the future that you need to include an error message when you're setting a workspace to a `*Failed` state.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunner.scala
```diff
@@ -101,7 +103,9 @@ class WorkspaceDeletionRunner(val samDAO: SamDAO,
           s"Workspace deletion failed [workspaceId=${workspaceId}, jobControlId=${job.jobControlId}, jobType=${job.jobType}]",
           t
         )
-        workspaceRepository.updateState(job.workspaceId.get, WorkspaceState.DeleteFailed).map(_ => Complete)
+        workspaceRepository
+          .setFailedState(job.workspaceId.get, WorkspaceState.DeleteFailed, t.getMessage)
```
cahrens, line 107: Shouldn't we append the jobControlID?

cahrens (approved):

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunner.scala
marctalbott: We discussed this during standup, but responding here too in case we ever need to reference it in the future:
There's no need to append the `jobControlId` here since all of the exceptions that `runStep` will throw include the `jobControlId` in their message -- see lines 129, 141 and 156.

PR #2591, opened 2023-10-19
[WOR-1309] Don't error for non-Ready MC Workspaces that have no cloud context.

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1309

The particular use case is that we need to be able to query workspace details for a workspace that is deleting-- there is a period of time when WSM has already deleted the cloud context, but has not yet finished deleting the workspace itself.

We decided to generalize the change though to any workspace that is not in the Ready state.

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4134,7 +4134,7 @@ paths:
       tags:
         - workspaces_v2
       summary: Delete workspace
-      description: Starts a job to delete a workspace and all of its contents. Note that this is an in-progress API and currently unsupported.
+      description: Starts a job to delete a workspace and all of its contents.
       operationId: deleteWorkspaceV2
       parameters:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5703,7 +5703,7 @@ components:
           format: date-time
         cloudPlatform:
           type: string
-          description: The cloud platform of the workspace
+          description: The cloud platform of the workspace. If the workspace is not in the Ready state, the cloud platform may be missing.
           enum:
             - GCP
             - AZURE
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
@@ -30,12 +30,17 @@ class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val sn
   // Throws an exception when the given snapshot cannot be referenced by the given workspace due to crossing an
   // unsupported platform boundary.
   @throws(classOf[PlatformBoundaryException])
-  def validateWorkspacePlatformCompatibility(workspacePlatform: WorkspaceCloudPlatform): Unit =
+  def validateWorkspacePlatformCompatibility(workspacePlatform: Option[WorkspaceCloudPlatform]): Unit =
     // Defining all acceptable combinations is overkill when all we really need to do is prevent anything Azure, but
     // this is here as a safeguard against us introducing support for Azure without deliberately addressing the issue of
     // snapshots by ref across cloud platforms.
     (snapshot.platform, workspacePlatform) match {
-      case (SnapshotCloudPlatform.GCP, WorkspaceCloudPlatform.Gcp) => // ok
+      case (SnapshotCloudPlatform.GCP, Some(WorkspaceCloudPlatform.Gcp)) => // ok
+      case (_, None) =>
+        throw new PlatformBoundaryException(
+          "Snapshots by reference are not supported into a workspace with no cloud context (" +
+            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform})."
+        )
       case _ =>
         throw new PlatformBoundaryException(
           "Snapshots by reference are not supported across the given cloud boundaries (" +
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
@@ -44,7 +44,7 @@ class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val sn
       case _ =>
         throw new PlatformBoundaryException(
           "Snapshots by reference are not supported across the given cloud boundaries (" +
-            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform})."
+            s"snapshot: ${snapshot.platform}, workspace: ${workspacePlatform.get})."
         )
     }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -83,7 +83,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
         case _: AggregateWorkspaceNotFoundException =>
           // if a WSM workspace does not already exist, assume the platform is GCP, confirm platform compatibility,
           // and then create a stub workspace
-          snapshotValidator.validateWorkspacePlatformCompatibility(WorkspaceCloudPlatform.Gcp)
+          snapshotValidator.validateWorkspacePlatformCompatibility(Some(WorkspaceCloudPlatform.Gcp))
           workspaceManagerDAO.createWorkspace(wsid, rawlsWorkspace.workspaceType, ctx)
       }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
@@ -70,19 +70,27 @@ class AggregatedWorkspaceService(workspaceManagerDAO: WorkspaceManagerDAO) exten
             azureCloudContext = None,
             convertPolicies(wsmInfo)
           )
-        case (WorkspaceStageModel.MC_WORKSPACE, None, None, WorkspaceState.Deleting) =>
-          AggregatedWorkspace(
-            workspace,
-            googleProjectId = None,
-            azureCloudContext = None,
-            convertPolicies(wsmInfo)
-          )
-        case (_, _, _, _) =>
+        case (WorkspaceStageModel.MC_WORKSPACE, Some(_), Some(_), _) =>
           throw new InvalidCloudContextException(
             ErrorReport(
               StatusCodes.NotImplemented,
               s"Unexpected state, expected exactly one set of cloud metadata for workspace ${workspace.workspaceId}"
             )
+          )
+        case (WorkspaceStageModel.MC_WORKSPACE, None, None, WorkspaceState.Ready) =>
+          throw new InvalidCloudContextException(
+            ErrorReport(
+              StatusCodes.NotImplemented,
+              s"Unexpected state, no cloud metadata for ready workspace ${workspace.workspaceId}"
+            )
+          )
+        case (WorkspaceStageModel.MC_WORKSPACE, None, None, _) =>
+          // Tolerate no cloud context for a workspace that is not ready.
+          AggregatedWorkspace(
+            workspace,
+            googleProjectId = None,
+            azureCloudContext = None,
+            convertPolicies(wsmInfo)
           )
       }
     } catch {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
@@ -7,15 +7,7 @@ import com.typesafe.scalalogging.LazyLogging
 import io.opencensus.scala.Tracing.startSpanWithParent
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.workspacemanager.WorkspaceManagerDAO
-import org.broadinstitute.dsde.rawls.model.{
-  AzureManagedAppCoordinates,
-  ErrorReport,
-  GoogleProjectId,
-  RawlsRequestContext,
-  Workspace,
-  WorkspacePolicy,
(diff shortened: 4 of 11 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
```diff
@@ -30,12 +30,17 @@ class SnapshotReferenceCreationValidator(val workspaceContext: Workspace, val sn
   // Throws an exception when the given snapshot cannot be referenced by the given workspace due to crossing an
   // unsupported platform boundary.
   @throws(classOf[PlatformBoundaryException])
-  def validateWorkspacePlatformCompatibility(workspacePlatform: WorkspaceCloudPlatform): Unit =
+  def validateWorkspacePlatformCompatibility(workspacePlatform: Option[WorkspaceCloudPlatform]): Unit =
```
cahrens, line 33: This is to handle the single usage by SnapshotService, at line https://github.com/broadinstitute/rawls/pull/2591/files#diff-3fb111d317a44913c11c6b540e6a2d05fdf5715c28ae0a0bb87921bb8e1057eeL80

Before the change `getCloudPlatform` would have thrown an exception for a workspace in this state.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -337,14 +337,15 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
               // determine which functions to use for the various part of the response
               def bucketOptionsFuture(): Future[Option[WorkspaceBucketOptions]] =
                 if (options.contains("bucketOptions")) {
-                  wsmContext.azureCloudContext match {
+                  wsmContext.googleProjectId match {
```
cahrens, line 340: Before this keyed off of `azureCloudContext`, returning bucket details if None. I have swapped the check so that we return bucket details if `googleProjectId` is not None, to handle the case of them both being None (plus it just makes more sense!).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -353,7 +354,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                 traceWithParent("getUserComputePermissions", s1)(_ =>
                   getUserComputePermissions(workspaceContext.workspaceIdAsUUID.toString,
                                             accessLevel,
-                                            Some(wsmContext.getCloudPlatform)
+                                            wsmContext.getCloudPlatform
```
cahrens, line 357: If this ends up being None, `getUserComputePermissions` will default to GCP permission model (Azure is slightly more lenient):

```
  private def getUserComputePermissions(workspaceId: String,
                                        userAccessLevel: WorkspaceAccessLevel,
                                        cloudPlatform: Option[WorkspaceCloudPlatform]
  ): Future[Boolean] =
    cloudPlatform match {
      case Some(WorkspaceCloudPlatform.Azure) =>
        Future.successful(userAccessLevel >= WorkspaceAccessLevels.Write)
      case _ if userAccessLevel >= WorkspaceAccessLevels.Owner =>
        Future.successful(true)
      case default =>
        samDAO.userHasAction(SamResourceTypeNames.workspace, workspaceId, SamWorkspaceActions.compute, ctx)
    }

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala [hunk 4]
```diff
@@ -427,7 +428,7 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                   WorkspaceDetails.fromWorkspaceAndOptions(workspaceContext,
                                                            authDomain,
                                                            useAttributes,
-                                                           Some(wsmContext.getCloudPlatform)
+                                                           wsmContext.getCloudPlatform
```
cahrens, line 431: This could already handle None.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1006,23 +1007,29 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
                         )
                       ),
                       attributesEnabled,
-                      Some(wsmContext.getCloudPlatform)
+                      wsmContext.getCloudPlatform
```
cahrens, line 1010: This could already handle None, but we will also filter out these workspaces below.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceSpec.scala
```diff
@@ -45,35 +58,35 @@ class AggregatedWorkspaceSpec extends AnyFlatSpec {
...
 
     val cp = ws.getCloudPlatform
-    cp shouldBe WorkspaceCloudPlatform.Gcp
+    cp shouldBe Some(WorkspaceCloudPlatform.Gcp)
   }
 
   it should "raise for an MC workspace that has cloud info for multiple clouds" in {
     val ws = AggregatedWorkspace(
-      mcWorkspace,
+      deleteFailedMcWorkspace,
```
cahrens, line 89: Checking that multiple cloud contexts throw an exception even for a non-ready workspace.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3251,21 +3307,24 @@ class WorkspaceServiceSpec
       ) should contain theSameElementsAs expected
   }
 
-  it should "not return a MC workspace that does not have a cloud context" in withTestDataServices { services =>
+  it should "not return MC workspaces that do not have a cloud context" in withTestDataServices { services =>
     val service = services.workspaceService
     val workspaceId1 = UUID.randomUUID().toString
     val workspaceId2 = UUID.randomUUID().toString
+    val workspaceId3 = UUID.randomUUID().toString
```
cahrens, line 3314: Add in another Azure workspace without cloud context so we can catch both ways that the filtering happens (Ready workspace vs. non-Ready workspace).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1006,23 +1007,29 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-                  Option(
-                    WorkspaceListResponse(
-                      accessLevel,
-                      workspaceDetails,
-                      submissionStats,
-                      workspaceSamResource.public.roles.nonEmpty || workspaceSamResource.public.actions.nonEmpty,
-                      Some(wsmContext.policies)
-                    )
-                  )
+                  // Remove workspaces that are non-ready with no cloud context (Ready workspaces with no
```
cahrens, line 1018: This is to maintain our previous behavior for the list endpoint… though I am starting to wonder if we could just allow workspaces with no cloud context to be returned. However, we would need to check that the list UI view can handle the cloud context being None, and think through if there are any other ramifications.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala
cahrens: FYI @jladieu ^

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
aherbst-broad: That'd be good, if only to clean up the special casing on this codepath.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3001,6 +3002,61 @@ class WorkspaceServiceSpec
...
+    runAndWait {
+      for {
+        _ <- slickDataSource.dataAccess.workspaceQuery.createOrUpdate(deletingAzureWorkspace)
+      } yield ()
+    }
+
+    when(service.workspaceManagerDAO.getWorkspace(deletingAzureWorkspace.workspaceIdAsUUID, services.ctx1))
+      .thenReturn(
+        new WorkspaceDescription().stage(WorkspaceStageModel.MC_WORKSPACE)
+      ) // no azureContext, should not be returned
```
aherbst-broad, line 3029: I don't understand this comment, are you saying the workspace should not be returned?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4134,7 +4134,7 @@ paths:
       tags:
         - workspaces_v2
       summary: Delete workspace
-      description: Starts a job to delete a workspace and all of its contents. Note that this is an in-progress API and currently unsupported.
+      description: Starts a job to delete a workspace and all of its contents.
```
aherbst-broad, line 4137: Good update

aherbst-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
cahrens: It was a comment that was already there… it's trying to point out that an azureContext is explicitly NOT being returned. If it were returned, it would look like this:

```
      when(service.workspaceManagerDAO.getWorkspace(azureWorkspace.workspaceIdAsUUID, services.ctx1))
        .thenReturn(
          new WorkspaceDescription()
            .stage(WorkspaceStageModel.MC_WORKSPACE)
            .azureContext(
              new AzureContext()
                .tenantId(UUID.randomUUID.toString)
                .subscriptionId(UUID.randomUUID.toString)
                .resourceGroupId(UUID.randomUUID.toString)
            )
        )

File: core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala (hunk 1 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala:33]: I'm having some trouble writing a test case for line 39. The test case that covers line 47 requires that a workspace actually exist (and it's in Ready state)-- it is using `testData.azureWorkspace`.

Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotReferenceCreationValidator.scala:33]: Ah never mind, I figured it out. I just pushed up a test case.

**Comment by cahrens** [comment]
jenkins retest

**Review by blakery — APPROVED** [review]

File: core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala (hunk 8 above)
Comment (cahrens) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala:1018]: Created https://broadworkbench.atlassian.net/browse/WOR-1314

**Comment by cahrens** [comment]
jenkins retest

File: core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala [hunk 11]
```diff
@@ -83,7 +83,7 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
         case _: AggregateWorkspaceNotFoundException =>
           // if a WSM workspace does not already exist, assume the platform is GCP, confirm platform compatibility,
           // and then create a stub workspace
-          snapshotValidator.validateWorkspacePlatformCompatibility(WorkspaceCloudPlatform.Gcp)
+          snapshotValidator.validateWorkspacePlatformCompatibility(Some(WorkspaceCloudPlatform.Gcp))
```
jladieu, line 86: 1) If I understand your comments correctly, it sounds like `wsmWorkspace` will _always_ have `getCloudPlatform` if the workspace state is `READY`; otherwise `fetchAggregatedWorkspace` will throw an error.  Do I have that right?

2) I also think maybe there's a business logic rule that isn't codified here explicitly (correct me if I'm wrong though as I'm not sure if my hunch is correct), which is that we should only create snapshots for READY workspaces.

If both of my assertions are correct, then I wonder if you can avoid having to make any changes within `validateWorkspacePlatformCompatibility` by just having this try block check the WSM state is READY and fail for that reason (rather than later on due to the omission of the platform).

As long as both (1) and (2) are actually true, I think this might be a bit more intention revealing.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala
```diff
@@ -43,14 +44,14 @@ class AggregatedWorkspaceService(workspaceManagerDAO: WorkspaceManagerDAO) exten
     val span = startSpanWithParent("getWorkspaceFromWorkspaceManager", ctx.tracingSpan.orNull)
     try {
       val wsmInfo = workspaceManagerDAO.getWorkspace(workspace.workspaceIdAsUUID, ctx)
-      (wsmInfo.getStage, Option(wsmInfo.getGcpContext), Option(wsmInfo.getAzureContext)) match {
-        case (WorkspaceStageModel.RAWLS_WORKSPACE, _, _) =>
+      (wsmInfo.getStage, Option(wsmInfo.getGcpContext), Option(wsmInfo.getAzureContext), workspace.state) match {
```
jladieu, line 47: Oof, this got pretty intense (scala newb here).  I defer to you on this as the code owner, but definitely whistled past this block on my way to the stuff I'm more familiar with.

jladieu (approved):
Approved with a comment/suggestion...I'm not actually sure if my suggestion is merited!  Please let me know either way so I can learn the logic.

Full disclosure, I honed in mainly on the stuff that's relevant to the place I was fiddling with last week, and defer judgment on the remainder to WSM folks!

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
cahrens: Yes, getCloudPlatform will throw an error if a workspace is in the Ready state and has no CloudPlatform.

I agree with your second point, but I didn't know what the UI might do (or not) to prevent the user from trying to use a non-Ready workspace. Would you object to me submitting this as-is and possibly either you or I do a follow-up PR specific to the snapshot service? I would love to get this change in ASAP because it alleviates workspace deletion pain.

jladieu: Sounds good, go for it!

PR #2600, opened 2023-10-23
[AJ-1326] Add timeout for query

nawatts:
https://broadworkbench.atlassian.net/browse/AJ-1326

I _think_ this adds a timeout for the query referenced in the ticket.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
@@ -812,7 +812,7 @@ trait AttributeComponent {
       def updateAction(insertIntoScratchFunction: String => WriteAction[Int]) = {
         val transactionId = UUID.randomUUID().toString
         insertIntoScratchFunction(transactionId) andThen
-          updateInMasterAction(transactionId) andThen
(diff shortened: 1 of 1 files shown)


davidangb (commented):
syntax looks good.

Is it worth moving the timeout value from a hardcoded 1200 to a config value, so it can easily be overridden in production without recompiling Rawls?

Have you been able to test this, at least manually, and see the effect? Does this bubble up a decent error message? Is it worth adding a unit test a la https://github.com/broadinstitute/rawls/blob/0c3e49b3005f288bdfa830c61d094043eef2171d/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala#L525-L551 ?

nawatts:
I did not test the timeout. It looked like there were [unit tests for the happy path](https://github.com/broadinstitute/rawls/blob/39ce46afaa7c1f013816b9b7a01431630a6ceb20/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponentSpec.scala#L1591-L1682), so I was trusting those to verify that adding the timeout didn't cause an issue.

And it didn't look like the [other timeout mentioned in the ticket](https://github.com/broadinstitute/rawls/blob/296cdfcfa49ab984e2133795a18e103cab4463cc/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L925-L939) did anything to change the error message.

PR #2601, opened 2023-10-23
[WOR-1320] Prevent FastPass interference with bucket migrations

marctalbott:
Ticket: [WOR-1320](https://broadworkbench.atlassian.net/browse/WOR-1320)
* Explicitly set the IAM policy version to 3 when restoring IAM at the end of the migration
    * When FastPass grants are present on a migrating workspace bucket, the call to restore IAM on the bucket fails due to an invalid policy number (see https://cloud.google.com/iam/help/allow-policies/versions for more details).
* Prevent FastPass grants on workspaces that have failed to be migrated
    * We were already preventing FastPass from working on migrating workspaces. Adding this change will prevent FastPass from granting users access to their workspace buckets after a migration has failed. Users still should not have access to their workspace buckets at this time as the workspace should be considered in progress
---

**PR checklist**

[WOR-1320]: https://broadworkbench.atlassian.net/browse/WOR-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
@@ -598,7 +598,8 @@ object MultiregionalBucketMigrationActor {
             .updateBucketIam(
               bucket,
               bucketPolicies,
-              Option.when(migration.requesterPaysEnabled)(GoogleProjectId(deps.googleProjectToBill.value))
+              Option.when(migration.requesterPaysEnabled)(GoogleProjectId(deps.googleProjectToBill.value)),
+              iamPolicyVersion = 3
             )
             .io *>
             Applicative[IO].whenA(migration.requesterPaysEnabled) {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/GoogleServicesDAO.scala
@@ -34,7 +34,8 @@ abstract class GoogleServicesDAO(groupsPrefix: String) extends ErrorReportable {
 
   def updateBucketIam(bucketName: GcsBucketName,
                       policyGroupsByAccessLevel: Map[WorkspaceAccessLevel, WorkbenchEmail],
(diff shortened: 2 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/migration/MultiregionalBucketMigrationActor.scala
```diff
@@ -598,7 +598,8 @@ object MultiregionalBucketMigrationActor {
             .updateBucketIam(
               bucket,
               bucketPolicies,
-              Option.when(migration.requesterPaysEnabled)(GoogleProjectId(deps.googleProjectToBill.value))
+              Option.when(migration.requesterPaysEnabled)(GoogleProjectId(deps.googleProjectToBill.value)),
+              iamPolicyVersion = 3
```
cahrens, line 602: What version would it have been at, in the case of a FastPass grant, if you didn't specify 3?

marctalbott: The policy we are setting will default to 1 since there are no conditionals. The existing policy on the bucket is 3 because of the FastPass grants.

cahrens: And there is no harm in setting it to 3 for all buckets?

marctalbott: Not that I can see. It's actually recommended by Google and it's the policy version we're using if a FastPass grant has been issued. 

> If you are integrating with IAM for the first time, we recommend using the most recent allow policy schema version [3] available at that time
https://cloud.google.com/iam/docs/policies#versions

blakery (approved):

cahrens (approved):
I do wonder if we should generally use Iam Policy Verison 3.

marctalbott:
> I do wonder if we should generally use Iam Policy Verison 3.

Agreed. I didn't want to blow up the scope of this ticket, but it seems like it'd be worth it to make version 3 the default

PR #2627, opened 2023-11-16
[WOR-1037] Set protected-data policy on billing profile for protected data Azure billing projects

marctalbott:
Ticket: [WOR-1037](https://broadworkbench.atlassian.net/browse/WOR-1037)
* When the protected data flag is set on an Azure billing project create request, set a `protected-data` policy on the billing profile in BPM
* Switch to the new BPM javax client library

---

**PR checklist**

[WOR-1037]: https://broadworkbench.atlassian.net/browse/WOR-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
@@ -133,7 +133,7 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar {
     val bpmDAO = new BillingProfileManagerDAOImpl(provider, config)
 
     intercept[NotImplementedError] {
-      bpmDAO.createBillingProfile("fake", Left(RawlsBillingAccountName("fake")), testContext)
+      bpmDAO.createBillingProfile("fake", Left(RawlsBillingAccountName("fake")), Map.empty, testContext)
     }
   }
 
@@ -147,7 +147,7 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar {
     val config = new MultiCloudWorkspaceConfig(true, None, None)
     val bpmDAO = new BillingProfileManagerDAOImpl(provider, config)
 
-    val profile = bpmDAO.createBillingProfile("fake", Right(coords), testContext)
+    val profile = bpmDAO.createBillingProfile("fake", Right(coords), Map.empty, testContext)
 
     assertResult(expectedProfile)(profile)
     verify(profileApi, times(1)).createProfile(ArgumentMatchers.any[CreateProfileRequest])
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
@@ -158,7 +158,7 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar with Mo
     when(provider.getProfileApi(ArgumentMatchers.eq(testContext))).thenReturn(profileApi)
     val config = new MultiCloudWorkspaceConfig(true, None, None)
     val coords = AzureManagedAppCoordinates(UUID.randomUUID(), UUID.randomUUID(), "fake_mrg")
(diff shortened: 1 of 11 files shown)


aherbst-broad (approved):
lgtm, looks there is some pact test breakage

cahrens:
@marctalbott can you please also return the `protectedData` boolean on the billing project response? We need that for the UI tickets where we want to enforce that the user chooses a protected data billing project.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -446,15 +457,17 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
     }
   }
 
-  it should "create a protected data landing zone if requested" in {
+  it should "create a protected data landing zone and attach a protected-data policy to the billing profile if requested" in {
```
marctalbott, line 460: Only functional test change in this file. Others are just updates for the changed function signature

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
```diff
@@ -147,12 +144,56 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoSugar {
...
+
+  it should "include any policies that are present" in {
+    val provider = mock[BillingProfileManagerClientProvider](RETURNS_SMART_NULLS)
+    val profileApi = mock[ProfileApi](RETURNS_SMART_NULLS)
+    val dummyProfile = new ProfileModel().id(UUID.randomUUID())
+    when(profileApi.createProfile(ArgumentMatchers.any[CreateProfileRequest])).thenReturn(dummyProfile)
+    when(provider.getProfileApi(ArgumentMatchers.eq(testContext))).thenReturn(profileApi)
+    val config = new MultiCloudWorkspaceConfig(true, None, None)
+    val coords = AzureManagedAppCoordinates(UUID.randomUUID(), UUID.randomUUID(), "fake_mrg")
+    val policies = Map("protected-data" -> Map[String, String]())
```
cahrens: Nit: it seems like it would be good to include a policy that has additional data, just to make sure that it gets passed through.

cahrens (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
marctalbott: Good call, I'll add that

marctalbott: Adding this to the test uncovered a bug that using a `Map[String, String]` means a policy can't have additional data with a duplicate key which I believe is something we want to support for cases like group constraints with multiple groups. Replaced the `Map[String, String]` with a `List[(String, String)]`

marctalbott: The bug is probably a non-issue for now since we're only setting the protected-data policy, but would need to be fixed if we ever wanted to support more policies

cahrens: Yes, we should be able to support policies like this:

```
"policies": [
        {
            "additionalData": [
                {
                    "group": "20231115_policy_test"
                },
                {
                    "group": "20231024_policy_test"
                }
            ],
            "name": "group-constraint",
            "namespace": "terra"
        }
    ]

Comment (cahrens) [reply core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala]: We've also seen this with regionality contraints--

```
const regionConstraintPolicy: WorkspacePolicy = {
  additionalData: [
    {
      'region-name': 'azure.eastus',
    },
    {
      'region-name': 'azure.westus2',
    },
    {
      'region-name': 'unknownRegion',
    },
  ],
  name: 'region-constraint',
  namespace: 'terra',
};

PR #2641, opened 2023-11-29
AJ-1438: disable most entity APIs for Azure workspaces

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1438

This PR disables all writable entity APIs and most read-only entity APIs for Azure workspaces. Azure workspaces use WDS instead of Rawls for data tables; the APIs in Rawls are misleading and should not be used for Azure workspaces.

This check is implemented by using `EntityManager` as a bottleneck. Inside `EntityManager.resolveProvider()`, we check the target workspace and throw an error if it is not a GCP (non-MC) workspace. As part of this, I needed to move the `copyEntities` implementation inside the `EntityManager`/`EntityProvider` structure, instead of its previous implementation directly in `EntityService`.

There are still a few read-only APIs that are implemented directly in `EntityService` and won't throw an error for Azure workspaces. However, all write APIs are now protected, so there's no way to get data IN to the system for those read APIs to return. I think this satisfies the intent of the Jira ticket, without blowing up scope too much.

_Update 8-Dec:_ I tested this in a BEE with the latest UI code, and all works as expected.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.rawls.entities
 
+import bio.terra.workspace.model.CloudPlatform
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.config.DataRepoEntityProviderConfig
 import org.broadinstitute.dsde.rawls.dataaccess.datarepo.DataRepoDAO
@@ -9,7 +10,7 @@ import org.broadinstitute.dsde.rawls.entities.base.{EntityProvider, EntityProvid
 import org.broadinstitute.dsde.rawls.entities.datarepo.{DataRepoEntityProvider, DataRepoEntityProviderBuilder}
 import org.broadinstitute.dsde.rawls.entities.exceptions.DataEntityException
 import org.broadinstitute.dsde.rawls.entities.local.{LocalEntityProvider, LocalEntityProviderBuilder}
-import org.broadinstitute.dsde.rawls.model.ErrorReport
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, WorkspaceType}
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.runtime.universe._
@@ -41,6 +42,10 @@ import scala.util.{Failure, Try}
 class EntityManager(providerBuilders: Set[EntityProviderBuilder[_ <: EntityProvider]]) {
 
   def resolveProvider(requestArguments: EntityRequestArguments): Try[EntityProvider] = {
+
+    if (!WorkspaceType.RawlsWorkspace.equals(requestArguments.workspace.workspaceType)) {
+      throw new DataEntityException(s"This functionality only available to ${CloudPlatform.GCP} workspaces.")
+    }
 
     // soon: look up the reference name to ensure it exists.
     // for now, this simplistic logic illustrates the approach: choose the right builder for the job.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
@@ -44,7 +44,9 @@ class EntityManager(providerBuilders: Set[EntityProviderBuilder[_ <: EntityProvi
   def resolveProvider(requestArguments: EntityRequestArguments): Try[EntityProvider] = {
 
     if (!WorkspaceType.RawlsWorkspace.equals(requestArguments.workspace.workspaceType)) {
-      throw new DataEntityException(s"This functionality only available to ${CloudPlatform.GCP} workspaces.")
+      throw new DataEntityException(
+        s"This API is disabled for ${CloudPlatform.AZURE} workspaces. Contact support for alternatives."
+      )
     }
 
     // soon: look up the reference name to ensure it exists.
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -540,44 +540,31 @@ class EntityService(protected val ctx: RawlsRequestContext,
     }
   }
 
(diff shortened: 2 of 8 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
```diff
@@ -42,6 +43,10 @@ class EntityManager(providerBuilders: Set[EntityProviderBuilder[_ <: EntityProvi
 
   def resolveProvider(requestArguments: EntityRequestArguments): Try[EntityProvider] = {
 
+    if (!WorkspaceType.RawlsWorkspace.equals(requestArguments.workspace.workspaceType)) {
+      throw new DataEntityException(s"This functionality only available to ${CloudPlatform.GCP} workspaces.")
+    }
+
```
davidangb, line 52: This is the heart of the PR: throw an error for non-`RawlsWorkspace` workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -540,44 +540,31 @@ class EntityService(protected val ctx: RawlsRequestContext,
...
+      entityCopyResponse <- entityProvider
+        .copyEntities(sourceWsCtx,
+                      destWsCtx,
+                      entityCopyDef.entityType,
+                      entityCopyDef.entityNames,
+                      linkExistingEntities,
+                      ctx
+        )
+        .recover(bigQueryRecover)
+    } yield entityCopyResponse
```
davidangb, line 567: While moving the important implementation inside the `LocalEntityProvider`, I took the opportunity to refactor and simplify this method. It is now one big for-yield instead of a mix of flatMaps and for-yields, AND it removes outbound REST calls from Rawls to Sam from being inside a database transaction.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -132,14 +132,14 @@ trait WorkspaceSupport {
...
     // if the source has any auth domains, the dest must also *at least* have those auth domains
-    if (sourceWorkspaceADs.subsetOf(destWorkspaceADs)) DBIO.successful(true)
+    if (sourceWorkspaceADs.subsetOf(destWorkspaceADs)) true
     else {
       val missingGroups = sourceWorkspaceADs -- destWorkspaceADs
       val errorMsg =
         s"Source workspace has an Authorization Domain containing the groups ${missingGroups.mkString(", ")}, which are missing on the destination workspace"
-      DBIO.failed(new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.UnprocessableEntity, errorMsg)))
+      throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.UnprocessableEntity, errorMsg))
     }
```
davidangb, line 143: This `authDomainCheck` method has only one usage - inside `copyEntities`, which I rewrote above. There was no need for `authDomainCheck` to return its result wrapped in a `ReadWriteAction`.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala
```diff
@@ -317,7 +317,7 @@ trait EntityApiService extends UserInfoDirectives {
                   entity(as[EntityCopyDefinition]) { copyDefinition =>
                     complete {
                       entityServiceConstructor(ctx)
-                        .copyEntities(copyDefinition, request.uri, linkExistingEntitiesBool)
+                        .copyEntities(copyDefinition, linkExistingEntitiesBool)
```
davidangb, line 320: I have no idea why the request URI was passed into `copyEntities`. It was never used, and I removed it from the argument list for `copyEntities`.

davidangb:
two service timeouts in integration tests. Jenkins retest.

jladieu (approved):

yuliadub (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -62,6 +63,9 @@ object RawlsApiService extends LazyLogging {
           Sentry.captureException(wsmApiException)
         }
         complete(wsmApiException.getCode -> ErrorReport(wsmApiException).copy(stackTrace = Seq()))
+      case dataEntityException: DataEntityException =>
+        // propagate only the message; don't include the stack trace
+        complete(dataEntityException.code -> ErrorReport(dataEntityException.getMessage))
```
davidangb, line 68: I thought about removing stack traces from the response in the catch-all case just below this … but that seemed like it could impact use cases I didn't want to think through. So, I'm adding a case just for `DataEntityException`, which is only thrown by data tables. There's no need to send the stack trace to the end user for these exceptions.

calypsomatic (approved):

jladieu (approved):

davidangb:
Jenkins retest, trying to get the TSPS fix in

davidangb:
a few 503s from jenkins. Jenkins retest.

davidangb:
One error on "Billing Project tmp-billing-project-1f5db88701 does not exist in Rawls". Jenkins retest.

davidangb:
well this is frustrating. More 503s. Jenkins retest.

davidangb:
503s from:
[Test Result](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/69086/testReport/) (2 failures / -3)
[org.broadinstitute.dsde.test.api.WorkspaceApiSpec.Rawls should allow writers to launch workflows if they have can-compute permission](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/69086/testReport/junit/org.broadinstitute.dsde.test.api/WorkspaceApiSpec/Rawls_should_allow_writers_to_launch_workflows_if_they_have_can_compute_permission/)
[org.broadinstitute.dsde.test.api.WorkspaceApiSpec.Rawls should not allow writers to launch workflows if they don't have can-compute permission](https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/69086/testReport/junit/org.broadinstitute.dsde.test.api/WorkspaceApiSpec/Rawls_should_not_allow_writers_to_launch_workflows_if_they_don_t_have_can_compute_permission/)

Jenkins retest.

PR #2644, opened 2023-12-01
[WOR-161] Stop trying to clone workspace files if it hasn't succeeded within a day

marctalbott:
Ticket: [WOR-161](https://broadworkbench.atlassian.net/browse/WOR-161)
In some cases, the asynchronous file transfer process never succeeds. This clogs up the logs and results in Rawls performing unnecessary work. This PR introduces some guardrails to prevent Rawls from endlessly retrying.

* Add columns to `CLONE_WORKSPACE_FILE_TRANSFER`
    * `created` tracks when a workspace is cloned and the file transfer attempts begin. If a file transfer is still pending after 1 day, it will be marked as failed.
    * `finished` tracks when a workspace file transfer has completed. 
    * `outcome` indicates whether a workspace file transfer succeeded or failed.
* Completed file transfers are no longer removed from the DB. This has the added benefit of allowing us to track a workspace's cloning lineage going forward.
* A couple of small changes to existing tests to clean up their logs and make it easier to identify real failures

---

**PR checklist**

[WOR-161]: https://broadworkbench.atlassian.net/browse/WOR-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
@@ -50,28 +50,39 @@ class CloneWorkspaceFileTransferMonitorActor(val dataSource: SlickDataSource,
       }
       _ <- pendingTransfers.toList
         .traverse { pendingTransfer =>
-          if (pendingTransfer.created.isBefore(DateTime.now().minusDays(1))) {
-            logger.warn(
-              s"File transfer from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName} did not succeed within allowed time and will no longer be retried. [workspaceId=${pendingTransfer.destWorkspaceId}]"
-            )
-            IO.fromFuture(IO(markTransferAsComplete(pendingTransfer, transferSucceeded = false)))
-          } else {
-            IO.fromFuture(IO(copyBucketFiles(pendingTransfer))).attempt.map {
-              case Left(e) =>
-                // We do not want to throw e here. traverse stops executing as soon as it encounters a Failure, but we
-                // want to continue traversing the list to transfer the rest of the buckets even if one of the
-                // copy operations fails.
-                logger.warn(
-                  s"Failed to copy files from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName}",
-                  e
-                )
-                List.empty
-              case Right(res) => res
-            }
+          IO.fromFuture(IO(attemptTransfer(pendingTransfer))).attempt.map {
+            case Left(e) =>
+              // We do not want to throw e here. traverse stops executing as soon as it encounters a Failure, but we
+              // want to continue traversing the list to transfer the rest of the buckets even if one of the
+              // copy operations fails.
+              logger.warn(
+                s"Failed to copy files from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName}",
+                e
+              )
+              List.empty
+            case Right(res) => res
           }
         }
         .unsafeToFuture()
     } yield ()
+
+  private def attemptTransfer(
+    pendingTransfer: PendingCloneWorkspaceFileTransfer
+  )(implicit executionContext: ExecutionContext): Future[List[Option[StorageObject]]] = {
+    val transferExpired = pendingTransfer.created.isBefore(DateTime.now().minusDays(1))
+    for {
+      copiedObjects <-
+        if (!transferExpired) copyBucketFiles(pendingTransfer)
+        else {
+          logger.warn(
+            s"File transfer from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName} did not succeed within allowed time and will no longer be retried. [workspaceId=${pendingTransfer.destWorkspaceId}]"
+          )
+          Future.successful(List.empty)
+        }
+
+      _ <- markTransferAsComplete(pendingTransfer, transferSucceeded = !transferExpired)
+    } yield copiedObjects
+  }
 
   private def copyBucketFiles(
     pendingCloneWorkspaceFileTransfer: PendingCloneWorkspaceFileTransfer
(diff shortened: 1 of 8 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
```diff
@@ -423,7 +427,7 @@ class CloneWorkspaceFileTransferMonitorSpec(_system: ActorSystem)
                             destinationBucketName,
                             goodObjectToCopy.getName,
                             Option(destWorkspace.googleProjectId)
-        )
+        )(system.dispatchers.defaultGlobalDispatcher)
```
marctalbott, line 430: This was causing a NPE since the call wasn't properly mocked. The test still succeeded because the `verify` only checks that the method was called 5+ times, not that the method succeeded on those calls.

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
```diff
@@ -213,7 +214,8 @@ class CloneWorkspaceFileTransferMonitorSpec(_system: ActorSystem)
 
       val mockGcsDAO = mock[GoogleServicesDAO](RETURNS_SMART_NULLS)
       val failureMessage = "because I feel like it"
-      val exception = new HttpResponseException.Builder(403, failureMessage, new HttpHeaders()).build
+      val exception =
+        new HttpResponseException.Builder(403, failureMessage, new HttpHeaders()).setMessage(failureMessage).build
```
marctalbott, line 218: Added `.setMessage` to this exceptions and others in the tests since they were showing up in the logs with null messages and I was initially concerned something was wrong. The message makes it more clear that these exceptions are expected.

Before:
```
com.google.api.client.http.HttpResponseException: null
	at com.google.api.client.http.HttpResponseException$Builder.build(HttpResponseException.java:293)
	...
```
After:
```
com.google.api.client.http.HttpResponseException: because I feel like it
	at com.google.api.client.http.HttpResponseException$Builder.build(HttpResponseException.java:293)
	...
```

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
```diff
@@ -66,6 +66,24 @@ class CloneWorkspaceFileTransferMonitorActor(val dataSource: SlickDataSource,
...
+      copiedObjects <-
+        if (!transferExpired) copyBucketFiles(pendingTransfer)
+        else {
+          logger.warn(
+            s"File transfer from ${pendingTransfer.sourceWorkspaceBucketName} to ${pendingTransfer.destWorkspaceBucketName} did not succeed within allowed time and will no longer be retried. [workspaceId=${pendingTransfer.destWorkspaceId}]"
+          )
+          Future.successful(List.empty)
+        }
+
+      _ <- markTransferAsComplete(pendingTransfer, transferSucceeded = !transferExpired)
```
cahrens, line 83: So `copyBucketFiles` doesn't complete until the transfer is complete? But somehow we get to the point where we can kill off a transfer that is taking too long….

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
cahrens: Perhaps make the exception text more meaningful, like "expected test exception"?

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitorSpec.scala
```diff
@@ -610,4 +615,117 @@ class CloneWorkspaceFileTransferMonitorSpec(_system: ActorSystem)
...
+      runAndWait(workspaceQuery.createOrUpdate(destWorkspace))
+      runAndWait(
+        cloneWorkspaceFileTransferQuery.save(destWorkspace.workspaceIdAsUUID,
+                                             sourceWorkspace.workspaceIdAsUUID,
+                                             copyFilesWithPrefix
+        )
+      )
+
+      val mockGcsDAO = mock[GoogleServicesDAO](RETURNS_SMART_NULLS)
+      val failureMessage = "because I feel like it"
```
cahrens: No, not again!!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/CloneWorkspaceFileTransferMonitor.scala
marctalbott: This addresses cases where a source workspace's files can't be copied to a destination workspace for longer than a day due to persistent errors. When that happens, the monitor currently tries to clone the files nonstop and fills up our logs unnecessarily.

It doesn't attempt to protect against long running copy operations while transferring large buckets. While it would be nice to have those protections, the solution is likely to switch to STS which would be more involved.

cahrens: Oh, so `copyBucketFIles` would throw an exception at some point, and then the monitor kicks off this code again?

marctalbott: Yes, that's correct. And if `copyBucketFiles` continues to throw exceptions for over a day for a given workspace, this will make it so the monitor stops trying to transfer the files for that workspace

blakery: I think the idea is to stop _retries_, not to interrupt an in-progress operation?
And then if it fails, this entire function will return a failed future (it won't even get to `markTransferAsComplete`).

blakery (approved):

cahrens (approved):

PR #2654, opened 2023-12-06
[WOR-1348] Fix GCP Enhanced Bucket Monitoring on Clone

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1348

Rawls needs to look at the source workspace bucket to determine a cloned workspace's enhanced bucket  monitoring, not the destination bucket, which can happen if the destination workspace has a bucket name.

---
**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
@@ -2718,6 +2718,32 @@ class WorkspaceServiceSpec
       workspace.bucketName should startWith(s"${services.workspaceServiceConfig.workspaceBucketNamePrefix}-secure")
   }
 
+  it should "clone a workspace with an enhanced bucket monitoring, resulting in the child workspace having enhanced logging even if the destination bucket location is defined" in withTestDataServices {
+    services =>
+      val baseWorkspaceName = "secure_space_for_workin"
+      val baseWorkspaceRequest = WorkspaceRequest(
+        testData.testProject1Name.value,
+        baseWorkspaceName,
+        Map.empty,
+        authorizationDomain = Option(Set(testData.dbGapAuthorizedUsersGroup))
+      )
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -2718,6 +2718,32 @@ class WorkspaceServiceSpec
...
   }
 
+  it should "clone a workspace with an enhanced bucket monitoring, resulting in the child workspace having enhanced logging even if the destination bucket location is defined" in withTestDataServices {
+    services =>
+      val baseWorkspaceName = "secure_space_for_workin"
+      val baseWorkspaceRequest = WorkspaceRequest(
+        testData.testProject1Name.value,
+        baseWorkspaceName,
+        Map.empty,
+        authorizationDomain = Option(Set(testData.dbGapAuthorizedUsersGroup))
```
cahrens: I think the auth domain case worked before, and what failed was when the original workspace had "enahncedBucketLogging" set to true but no auth domains.

tlangs: Intersting, I got a test failure when I removed my fix, and a pass when I included it.

cahrens: Hmmm, the bug we are seeing from terra-ui is specific to not checking "enhancedBucketLogging" but that's probably because we force "enhancedBucketLogging" to be true for workspaces that have an auth domain. So if we didn't send the option in the clone request I guess we would see the bug then also.

cahrens (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
tlangs: updated the test case to be even clearer

tlangs:
jenkins retest

aherbst-broad (approved):

tlangs:
jenkins retest

PR #2656, opened 2023-12-06
AJ-1485: Ignore redundant `AvroUpsert` messages.

jladieu:
[AJ-1485](https://broadworkbench.atlassian.net/browse/AJ-1485) Maintenance Ticket

Double deliveries can happen occasionally, per:
https://cloud.google.com/pubsub/docs/subscription-overview#default_properties

Rather than let such redundant messages get stuck in the queue, ack them to clear them out.

This is done as a response to a `Rawls import pub/sub message` [alert](https://broadinstitute.slack.com/archives/C031KGQUVB4/p1701822121418369) to prevent future such messages from getting stuck in the queue.

[AJ-1485]: https://broadworkbench.atlassian.net/browse/AJ-1485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
@@ -162,15 +162,6 @@ class AvroUpsertMonitorSpec(_system: ActorSystem)
     mockImportServiceDAO
   }
 
-  def setUpMockImportServiceToThrowAnError(services: TestApiService, exception: Exception): ImportServiceDAO = {
-    val mockImportServiceDAO = setUpMockImportService(services)
-
-    when(mockImportServiceDAO.getImportStatus(any[UUID], any[WorkspaceName], any[UserInfo]))
-      .thenReturn(Future.failed(exception))
-
-    mockImportServiceDAO
-  }
-
   def setUpMockImportService(services: TestApiService): ImportServiceDAO = {
     setUpPubSub(services)
 
@@ -190,18 +181,6 @@ class AvroUpsertMonitorSpec(_system: ActorSystem)
         slickDataSource
       )
     )
-
-    mockImportServiceDAO
-  }
-
-  def setUpMockImportServiceToReturnStatus(services: TestApiService,
-                                           status: ImportStatuses.ImportStatus
-  ): ImportServiceDAO = {
-    val mockImportServiceDAO = setUpMockImportService(services)
-    when(
-      mockImportServiceDAO.getImportStatus(any[UUID], any[WorkspaceName], any[UserInfo])
-    )
-      .thenReturn(Future.successful(Some(status)))
 
     mockImportServiceDAO
   }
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -830,6 +848,43 @@ class AvroUpsertMonitorSpec(_system: ActorSystem)
 
   }
 
+  it should "ack pubsub message if message is double delivered" in withTestDataApiServices { services =>
+    val timeout = 30000 milliseconds
+    val interval = 250 milliseconds
+    val importUuid = UUID.randomUUID()
+
+    setUpMockImportServiceToReturnStatus(services, ImportStatuses.Done)
```
jladieu: What I really wanted to do here was inline `setUpMockImportServiceToReturnStatus` (as well as the `setUpMockImportServiceToThrowAnError` in the other two tests) but the compiler wouldn't let me use the mock DSL from within this scope (it complained about me trying to use `when`).  I'm not sure I understand why, but open to advice if doing so is reasonable.

davidangb: hm - it works for me:
```suggestion
    val mockImportServiceDAO = setUpMockImportService(services)
    when(
      mockImportServiceDAO.getImportStatus(any[UUID], any[WorkspaceName], any[UserInfo])
    ).thenReturn(Future.successful(Some(ImportStatuses.Done)))
```

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
```diff
@@ -830,6 +848,43 @@ class AvroUpsertMonitorSpec(_system: ActorSystem)
...
+    // Publish message on the request topic
+    services.gpsDAO.publishMessages(
+      importReadPubSubTopic,
+      List(MessageRequest(importUuid.toString, testAttributes(importUuid)))
+    )
+
+    // check that a pubsub message was acked.
+    eventually(Timeout(scaled(timeout)), Interval(scaled(interval))) {
+      services.gpsDAO.acks should not be empty
+    }
```
davidangb, line 868: Thinking out loud … I poked at this a bit, and really wanted to find a way to assert that it is acking the _second_ message - the duplicate one - but given the structure of `MockGooglePubSubDAO` I couldn't find a way to do that. I don't think it's worth refactoring a whole bunch of stuff to make that test work.

davidangb (approved):
thank you!

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorSpec.scala
jladieu: Huh, I must have been doing something silly.  When I let the IDE take care of this for me it worked fine and was easy to modify to produce what you provided.  Fixed in 0bd38c139e83f5050c6c471b75df87313bcc6572

jladieu: Fixed in 0bd38c139e83f5050c6c471b75df87313bcc6572

calypsomatic (approved):
LGTM

PR #2669, opened 2023-12-14
WOR-1403: add error handling for missing billing profile and for missing landing zone

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1403

Continue deletion if either the landing zone or billing profile is missing.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -4,8 +4,10 @@ import akka.http.scaladsl.model.StatusCodes.InternalServerError
 import akka.http.scaladsl.model.{StatusCode, StatusCodes}
 import bio.terra.profile.model.ProfileModel
 import bio.terra.workspace.client.ApiException
+import bio.terra.profile.client.{ApiException => BpmApiException}
 import bio.terra.workspace.model.{CreateLandingZoneResult, DeleteAzureLandingZoneResult}
 import cats.implicits.{catsSyntaxFlatMapOps, toTraverseOps}
+import org.apache.http.HttpStatus
 import org.broadinstitute.dsde.rawls.billing.BillingProfileManagerDAO.ProfilePolicy
 import org.broadinstitute.dsde.rawls.config.MultiCloudWorkspaceConfig
 import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceManagerResourceMonitorRecord
@@ -214,7 +216,10 @@ class BpmBillingProjectLifecycle(
   private def cleanupLandingZone(
     landingZoneId: UUID,
     ctx: RawlsRequestContext
-  ): DeleteAzureLandingZoneResult = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+  ): Option[DeleteAzureLandingZoneResult] = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+    case Failure(e: ApiException) if e.getCode == HttpStatus.SC_FORBIDDEN =>
+      logger.warn(s"No landing zone available with ID $landingZoneId for BPM-backed billing project.")
+      None
     case Failure(e: ApiException) =>
       val msg = s"Unable to delete landing zone: ${e.getMessage}"
       throw new LandingZoneDeletionException(RawlsErrorReport(StatusCode.int2StatusCode(e.getCode), msg, e))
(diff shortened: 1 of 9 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycleSpec.scala
```diff
@@ -864,7 +862,29 @@ class BpmBillingProjectLifecycleSpec extends AnyFlatSpec {
                                      mock[WorkspaceManagerResourceMonitorRecordDao]
       )
 
-    Await.result(bp.initiateDelete(billingProjectName, testContext), Duration.Inf)
+    Await.result(bp.initiateDelete(billingProjectName, testContext), Duration.Inf) shouldBe (Some(jobReportId))
```
blakery, line 864: The result of this is more meaningful now, so we really should verify it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -214,7 +216,10 @@ class BpmBillingProjectLifecycle(
   private def cleanupLandingZone(
     landingZoneId: UUID,
     ctx: RawlsRequestContext
-  ): DeleteAzureLandingZoneResult = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+  ): Option[DeleteAzureLandingZoneResult] = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
```
blakery, line 219: Made the return an optional value, so we could indicate a non-failure instance where we don't need to create the monitor job to wait for the result. 
This matches up nicely with the result and return behavior of the initializeDelete method, and doesn't conflict with the usage in cleanup when creation fails.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -214,7 +216,10 @@ class BpmBillingProjectLifecycle(
...
+    case Failure(e: ApiException) if e.getCode == HttpStatus.SC_FORBIDDEN =>
```
blakery: The service returns a 403 instead of a 404 when there's no landing zone present.
It's fine to move on, because if this was a 403 for permission reasons, we won't be able to delete the billing profile anyway, and we don't have a valid instance of a user having access to the billing project but not the underlying resources.

cahrens: It would be good to have this comment in the code.

cahrens: And if we DO happen to get a 404 from the service (which seems more correct), I think we should also handle it and allow the deletion to proceed. Perhaps at some point the 403 will be corrected to a 404.

blakery: good point

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -214,7 +216,13 @@ class BpmBillingProjectLifecycle(
   private def cleanupLandingZone(
     landingZoneId: UUID,
     ctx: RawlsRequestContext
-  ): DeleteAzureLandingZoneResult = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+  ): Option[DeleteAzureLandingZoneResult] = Try(workspaceManagerDAO.deleteLandingZone(landingZoneId, ctx)) match {
+    // The service returns a 403 instead of a 404 when there's no landing zone present.
+    // It's fine to move on here, because if this was a 403 for permission reasons, we won't be able to delete the billing profile anyway,
+    // and we don't have a valid instance of a user having access to the billing project but not the underlying resources.
+    case Failure(e: ApiException) if e.getCode == HttpStatus.SC_FORBIDDEN || e.getCode == HttpStatus.SC_NOT_FOUND =>
```
aherbst-broad: I've started pushing this sort of logic down into the DAO layer in the similar BPM error handling [code](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala#L170). The motivation is to keep the HTTP level concerns in the DAO (i.e., StatusCodes.NOT_FOUND etc.) and just return Option[foo] and make the upper levels deal with that interface. What do you think about doing that here?

aherbst-broad (approved):

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
blakery: I think that is a better method in the long term, with how we're looking to structure our code, and separate our testing concerns.

I'll give it a try, and see what kind of issues it creates in the immediate term.

PR #2670, opened 2023-12-14
submission monitor 20 per pass

davidangb:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -57,6 +57,8 @@ submissionmonitor {
   trackDetailedSubmissionMetrics = true
   attributeUpdatesPerWorkflow = 20000
   enableEmailNotifications = true
+  # how many workflows should the submission monitor attempt to process and update at once?
+  workflowsPerMonitorPass = 20
 }
 
 workspaceManagerResourceMonitor {
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
@@ -400,9 +400,15 @@ trait WorkflowComponent {
       }
 
     def listWorkflowRecsForSubmissionAndStatuses(submissionId: UUID,
+                                                 limit: Int,
                                                  statuses: WorkflowStatuses.WorkflowStatus*
     ): ReadAction[Seq[WorkflowRecord]] =
-      findWorkflowsBySubmissionId(submissionId).filter(_.status inSetBind (statuses.map(_.toString))).result
+      findWorkflowsBySubmissionId(submissionId).filter(_.status inSetBind (statuses.map(_.toString))).take(limit).result
+
+    def countWorkflowRecsForSubmissionAndStatuses(submissionId: UUID,
+                                                  statuses: WorkflowStatuses.WorkflowStatus*
+    ): ReadAction[Int] =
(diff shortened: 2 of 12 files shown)


core/src/main/resources/reference.conf
```diff
@@ -57,6 +57,8 @@ submissionmonitor {
   trackDetailedSubmissionMetrics = true
   attributeUpdatesPerWorkflow = 20000
   enableEmailNotifications = true
+  # how many workflows should the submission monitor attempt to process and update at once?
+  workflowsPerMonitorPass = 20
```
davidangb, line 61: what is a good initial value for this setting?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -400,9 +400,15 @@ trait WorkflowComponent {
       }
 
     def listWorkflowRecsForSubmissionAndStatuses(submissionId: UUID,
+                                                 limit: Int,
                                                  statuses: WorkflowStatuses.WorkflowStatus*
     ): ReadAction[Seq[WorkflowRecord]] =
-      findWorkflowsBySubmissionId(submissionId).filter(_.status inSetBind (statuses.map(_.toString))).result
+      findWorkflowsBySubmissionId(submissionId).filter(_.status inSetBind (statuses.map(_.toString))).take(limit).result
```
davidangb: the `.take(limit)` here should result in only returning `${workflowsPerMonitorPass}` workflows to the submission monitor actor for any given pass.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
```diff
@@ -400,9 +400,15 @@ trait WorkflowComponent {
...
+
+    def countWorkflowRecsForSubmissionAndStatuses(submissionId: UUID,
+                                                  statuses: WorkflowStatuses.WorkflowStatus*
+    ): ReadAction[Int] =
+      findWorkflowsBySubmissionId(submissionId).filter(_.status inSetBind (statuses.map(_.toString))).length.result
```
davidangb, line 424: previously, `SubmissionMonitorActor.updateSubmissionStatus()` called `listWorkflowRecsForSubmissionAndStatuses()`, returning all workflow records as a Seq, and then checked if that Seq was empty or not. This moves the logic to the db, returning just a count. I needed to change *something* here, because by adding a limit to `listWorkflowRecsForSubmissionAndStatuses()`, it would have been invalid for `SubmissionMonitorActor.updateSubmissionStatus()` - two birds with one stone?

core/src/main/resources/reference.conf
cjllanwarne: Less than 600!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkflowComponent.scala
davidangb: The drawback to adding a limit here is that if the first `limit` rows from the db are still running, but limit+1, limit+2, limit+3 are done … the monitor won't complete the ones that are done. This isn't great.

davidangb: You could alleviate this by `select … order by rand() limit 20`; in mysql, to get a random selection of 20 rows that match the where clause. I'm not sure how to do that in Slick.

PR #2668, opened 2023-12-14
[WOR-1365] Expose the bee template version so we can trigger the workflow on promotions  to staging.

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1365

I have set up a trigger from Beehive, which I tested. I will update it to run from `develop` after this PR merges, and also add triggers for BPM and WSM promotions to staging: https://beehive.dsp-devops.broadinstitute.org/environments/staging/chart-releases/rawls/deploy-hooks/github-actions/22

Example of not building a Rawls custom image and just running the test against staging versions:

image

PR #2677, opened 2023-12-19
[WOR-1365] Make input argument handling consistent.

cahrens:
Minor cleanup to #2668.

When the test is run twice a day via the chron trigger, input arguments are not present. This ended up meaning that we ran against dev by default because `create-bee` defaults to dev if no version template is supplied (and we ended up not building a custom Rawls version, which isn't needed in this case anyway). So while this behavior was correct, it was a bit difficult to reason about and caused the notification message to not include the Rawls version tested against.

For consistency, I am making the argument default handling the same as the other input parameters. I'm also changing the "build custom image" default to false, since it's actually uncommon that we need to build a custom Rawls image (and not building it reduces total time by about 10 minutes). Previously we would build a custom image from `develop` on every scheduled run, even though we were telling the bee to use dev versions.

Also changed the boolean options to choices because booleans are not well supported.

---

**PR checklist**

### DIFF
--- a/.github/workflows/rawls-run-azure-e2e-tests.yaml
+++ b/.github/workflows/rawls-run-azure-e2e-tests.yaml
@@ -19,13 +19,13 @@ on:
   workflow_dispatch:
     inputs:
       build-branch:
-        description: 'Build a custom image of Rawls from the specified branch. Default is true; if false, the tests run against the version specified by bee-version-template.'
+        description: 'Build a custom image of Rawls from the specified branch. Default is false, meaning tests are run against the version specified by bee-version-template.'
         required: true
-        default: true
+        default: false
         type: boolean
       branch:
         description: 'Branch of Rawls to build a custom image from. Ignored if custom image option is false.'
-        required: true
+        required: false
         default: 'develop'
         type: string
       bee-version-template:
@@ -72,7 +72,9 @@ jobs:
   init-github-context:
     runs-on: ubuntu-latest
     outputs:
+      build-branch: ${{ steps.extract-inputs.outputs.build-branch }}
       branch: ${{ steps.extract-inputs.outputs.branch }}
+      bee-version-template: ${{ steps.extract-inputs.outputs.bee-version-template }}
       fully-specified-branch: refs/heads/${{ steps.extract-inputs.outputs.branch }}
       delete-bee: ${{ steps.extract-inputs.outputs.delete-bee }}
       student-subjects: ${{ steps.extract-inputs.outputs.student-subjects }}
@@ -82,7 +84,9 @@ jobs:
       - name: Get inputs or use defaults
         id: extract-inputs
         run: |
+          echo "build-branch=${{ inputs.build-branch                || false }}" >> "$GITHUB_OUTPUT"
           echo "branch=${{ inputs.branch                            || 'develop' }}" >> "$GITHUB_OUTPUT"
+          echo "bee-version-template=${{ inputs.bee-version-template || 'dev' }}" >> "$GITHUB_OUTPUT"
           echo "delete-bee=${{ inputs.delete-bee                    || false }}" >> "$GITHUB_OUTPUT"
           echo "owner-subject=${{ inputs.owner-subject              || 'hermione.owner@quality.firecloud.org' }}" >> "$GITHUB_OUTPUT"
           echo "student-subjects=${{ toJson(inputs.student-subjects || '["harry.potter@quality.firecloud.org","ron.weasley@quality.firecloud.org"]') }}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -96,16 +100,16 @@ jobs:
       contents: 'read'
       id-token: 'write'
     outputs:
-      custom-version-json: ${{ inputs.build-branch && steps.render-rawls-version.outputs.custom-version-json || ''}}
+      custom-version-json: ${{ (needs.init-github-context.outputs.build-branch == true || needs.init-github-context.outputs.build-branch == 'true') && steps.render-rawls-version.outputs.custom-version-json || ''}}
```
cahrens: This is really annoying, but even though these arguments are defined as booleans, we end up with strings. And then 'false' is truthy, so just checking `needs.init-github-context.outputs.build-branch` leads to an incorrect condition.

I probably can ONLY check for 'true', but just in case I decided to check for both (which some folks on github issues recommend).

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -275,7 +280,7 @@ jobs:
       - init-github-context
       - rawls-swat-e2e-test-job
       - delete-billing-project-v2-from-bee-workflow
-    if: ${{ needs.init-github-context.outputs.delete-bee && always() }} # always run to confirm bee is destroyed unless explicitly requested not to
+    if: ${{ needs.init-github-context.outputs.delete-bee == true || needs.init-github-context.outputs.delete-bee == 'true' }}
```
cahrens: I don't think the `delete-bee` option was working properly before given the stringification of booleans.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -82,8 +84,10 @@ jobs:
       - name: Get inputs or use defaults
         id: extract-inputs
         run: |
+          echo "build-branch=${{ inputs.build-branch                || false }}" >> "$GITHUB_OUTPUT"
           echo "branch=${{ inputs.branch                            || 'develop' }}" >> "$GITHUB_OUTPUT"
-          echo "delete-bee=${{ inputs.delete-bee                    || false }}" >> "$GITHUB_OUTPUT"
+          echo "bee-version-template=${{ inputs.bee-version-template || 'dev' }}" >> "$GITHUB_OUTPUT"
+          echo "delete-bee=${{ inputs.delete-bee                    || true }}" >> "$GITHUB_OUTPUT"
```
cahrens: Changing default value of delete-bee` to `true` for consistency with input documentation and usual use cases.

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: Decided to switch to an explicit choice because the booleans still weren't working properly.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -19,13 +19,16 @@ on:
   workflow_dispatch:
     inputs:
       build-branch:
-        description: 'Build a custom image of Rawls from the specified branch. Default is true; if false, the tests run against the version specified by bee-version-template.'
+        description: 'Build a custom image of Rawls from the specified branch. Default is false, meaning tests are run against the version specified by bee-version-template.'
         required: true
-        default: true
-        type: boolean
```
cahrens, line 25: Declaring a boolean type leads to misery because github actions don't support them well. They get turned into strings, making comparisons error-prone (since 'false' is truthy).

Going with a choice makes it clear what the values will be.

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -275,7 +286,7 @@ jobs:
       - init-github-context
       - rawls-swat-e2e-test-job
       - delete-billing-project-v2-from-bee-workflow
-    if: ${{ needs.init-github-context.outputs.delete-bee && always() }} # always run to confirm bee is destroyed unless explicitly requested not to
+    if: ${{ needs.init-github-context.outputs.delete-bee == 'true' }}
```
cahrens, line 289: Examples of the `delete-bee` option working:

image
image

vs.
image
image

.github/workflows/rawls-run-azure-e2e-tests.yaml
```diff
@@ -96,16 +106,16 @@ jobs:
       contents: 'read'
       id-token: 'write'
     outputs:
-      custom-version-json: ${{ inputs.build-branch && steps.render-rawls-version.outputs.custom-version-json || ''}}
+      custom-version-json: ${{ needs.init-github-context.outputs.build-branch == 'true' && steps.render-rawls-version.outputs.custom-version-json || ''}}
```
cahrens, line 109: Custom version building a branch (with dev as the bee template version):

image

image

vs. not (with staging as the bee template version)

image
image

aherbst-broad (approved):

.github/workflows/rawls-run-azure-e2e-tests.yaml
cahrens: Looks like this:

image

marctalbott (approved):

PR #2292, opened 2023-03-27
WX-820 Finite monitoring for submissions

aednichols:
Ticket: https://broadworkbench.atlassian.net/browse/WX-820

Exceptions in workflow/submission monitoring frequently cause infinite retry loops. It's not a perfect fix, but backing off from infinity to 90 days should take a lot of load off of the system.

Generated query looks like this:

```
select 
  x2.x3, 
  x4.`namespace`, 
  x4.`name` 
from 
  (
    select 
      `STATUS` as x5, 
      `DATE_SUBMITTED` as x6, 
      `ID` as x3, 
      `WORKSPACE_ID` as x7 
    from 
      `SUBMISSION` 
    where 
      (
        `STATUS` in (
          'Accepted', 'Evaluating', 'Submitting', 
          'Submitted', 'Aborting'
        )
      ) 
      and (
        `DATE_SUBMITTED` > '2023-02-26 21:36:54.762'
      )
  ) x2, 
  `WORKSPACE` x4 
where 
  x2.x7 = x4.`id`
```

---

**PR checklist**

- [x] Include the JIRA issue number in the PR description and title
- [x] Make sure Swagger is updated if API changes
  - [x] **...and Orchestration's Swagger too!**
- [x] If you changed anything in `model/`, then you should [publish a new official `rawls-model`](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model) and update `rawls-model` in [Orchestration's dependencies](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/project/Dependencies.scala).
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green, including CI tests
- [ ] **Squash commits and merge** to develop (branches are automatically deleted after merging)
- [ ] Inform other teams of any substantial changes via Slack and/or email

PR #2680, opened 2023-12-19
AJ-1526: process workflows individually inside submission monitor

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1526

Behavior of the submission monitor before this PR:
* write all outputs for all workflows in this submission, batching into transactions of up to 20,000 attributes
* once outputs are completely written, update statuses for all workflows in a single transaction
* once workflow statuses are updated, update the status of the submission if it is complete

Behavior after this PR:
* loop through all workflows in this submission. For each workflow, in a transaction just for that workflow:
    * write the outputs, if any
    * update the status for this workflow
* once this loop is complete, update the status of the submission if it is complete

Imagine a case where a submission has 600 running workflows, all with outputs, and each of those workflows has large outputs that require 2 minutes to write. In this case, we previously required 1200 minutes (20 hours) to write outputs before any workflow changed status. After this PR, we'll see one workflow move to `Succeeded` every 2 minutes.

Now also imagine a case in which back Rawls is restarted - or in which we hit an unrecoverable error - after processing 400 of those workflows. Before this PR, we'd start back from square one and try to write all 600 workflow's outputs again. After this PR, we'll notice that 400 workflows have completed and pick up where we left off at workflow 401.

This PR moves from a smaller number of large db transactions to a large number of smaller db transactions. I think that's the right move. DB transactions are now per-workflow instead of spanning workflows, which feels correct.

Still TODO: This optimizes `handleStatusResponses`. I also see a smaller opportunity to optimize `queryExecutionServiceForStatus`, but would prefer to save that for a separate PR.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -29,7 +29,7 @@ import org.broadinstitute.dsde.rawls.model.SubmissionStatuses.SubmissionStatus
 import org.broadinstitute.dsde.rawls.model.WorkflowStatuses.WorkflowStatus
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.{addJitter, AuthUtil, FutureSupport}
-import org.broadinstitute.dsde.rawls.util.TracingUtils.{addSpanAttribute, trace, traceDBIOWithParent, traceWithParent}
+import org.broadinstitute.dsde.rawls.util.TracingUtils.{trace, traceDBIOWithParent, traceWithParent}
 import org.broadinstitute.dsde.rawls.{RawlsException, RawlsFatalExceptionWithErrorReport}
 import org.broadinstitute.dsde.workbench.dataaccess.NotificationDAO
 import org.broadinstitute.dsde.workbench.model.{Notifications, WorkbenchUserId}
@@ -47,6 +47,8 @@ import scala.language.postfixOps
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 import spray.json._
+
+import scala.Option
 
 /**
  * Created by dvoet on 6/26/15.
@@ -242,6 +244,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           abortResults <- Future.traverse(workflowRecs) { workflowRec =>
             Future.successful(workflowRec.externalId).zip(executionServiceCluster.abort(workflowRec, petUserInfo))
           }
+          // TODO: should this also mark the workflows as Aborted in the Rawls db???
         } yield abortResults
       }
 
@@ -315,13 +318,15 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     }
 
   /**
-   * once all the execution service queries have completed this function is called to handle the responses
-    * the WorkflowRecords in ExecutionServiceStatus response have not been saved to the database but have been updated with their status from Cromwell.
+    * once all the execution service queries have completed this function is called to handle the responses
+    * the WorkflowRecords in ExecutionServiceStatus response have not been saved to the database but
+    * have been updated with their status from Cromwell.
     *
-    * @param response
-   * @param executionContext
-   * @return
-   */
+    * @param response all workflows in this submission whose statuses are different in Cromwell than in Rawls. If a
+    *                 workflow has outputs in Cromwell, those outputs are included here.
+    * @param executionContext
+    * @return
+    */
   def handleStatusResponses(
     response: ExecutionServiceStatusResponse
   )(implicit executionContext: ExecutionContext): Future[StatusCheckComplete] =
@@ -332,6 +337,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         s.putAttribute("workspaceName", OpenCensusAttributeValue.stringAttributeValue(workspaceName.name))
       }
 
+      // log any failures.
       response.statusResponse.collect { case Failure(t) => t }.foreach { t =>
         logger.error(s"Failure monitoring workflow in submission $submissionId", t)
       }
@@ -341,19 +347,45 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         aWorkflow
       }
 
-      // just the workflow records in this response list which have outputs
-      val workflowsWithOutputs = response.statusResponse.collect { case Success(Some((workflowRec, Some(outputs)))) =>
-        (workflowRec, outputs)
-      }
+      // partition into two variables: those workflows that do have
+      // outputs vs. those workflows that do not have outputs
+      val (yesOutputs, noOutputs) = response.statusResponse
+        .collect { case Success(Some((workflowRec, maybeOutputs))) =>
+          (workflowRec, maybeOutputs)
+        }
+        .partition(_._2.isDefined)
 
-      // Loop over each workflow that has outputs. For each workflow, persist the outputs, then update the workflow's
-      // status.
-      logger.info(s"will process ${workflowsWithOutputs.size} workflow(s) for submission $submissionId")
+      // collapse the option for the workflow records in this response list which have outputs
+      val workflowsWithOutputs: Seq[(WorkflowRecord, ExecutionServiceOutputs)] = yesOutputs.collect {
+        case (workflowRec, Some(outputs)) => (workflowRec, outputs)
+      }
+      logger.info(
+        s"will process ${noOutputs.size} workflow(s) without outputs and ${workflowsWithOutputs.size} workflow(s) with outputs for submission $submissionId"
+      )
+
       traceWithParent("workflows", rootSpan) { workflowsSpan =>
         workflowsSpan.tracingSpan.foreach { s =>
           s.putAttribute("numWorkflows", OpenCensusAttributeValue.longAttributeValue(workflowsWithOutputs.size))
         }
-        workflowsWithOutputs.zipWithIndex
+
+        // Loop over those workflows that do not have outputs, and update their status to match what is in Cromwell.
+        // This makes Rawls aware of any workflows that have failed or been aborted in Cromwell.
+        val workflowsWithoutOutputsFuture = noOutputs
+          .traverse { case (workflowRec, _) =>
+            IO.fromFuture(IO(datasource.inTransaction { dataAccess =>
+              for {
+                currentRec <- dataAccess.workflowQuery.findWorkflowById(workflowRec.id).result.head
+                numRowsUpdated <- dataAccess.workflowQuery.updateStatus(currentRec,
+                                                                        WorkflowStatuses.withName(workflowRec.status)
+                )
+              } yield numRowsUpdated
+            }))
+          }
+          .unsafeToFuture()
+
+        // Loop over each workflow that has outputs. For each workflow, persist the outputs, then update the workflow's
+        // status.
+        val workflowsWithOutputsFuture = workflowsWithOutputs.zipWithIndex
           .traverse { case ((workflowRec, execServiceOutputs), idx) =>
             logger.info(
               s"processing workflow ${idx + 1}/${workflowsWithOutputs.size} with id ${externalId(workflowRec)} for submission $submissionId"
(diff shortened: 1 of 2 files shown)


davidangb:
Jenkins retest

davidangb:
one error in swat tests on "group AuthDomain_7e6de40f-477a-4784-b711-b5721be78c46 cannot be deleted because it is a member of at least 1 other group". Jenkins retest.

davidangb:
a spate of 503s from swat tests. Jenkins retest.

davidangb:
11 more failures, all 503s/504s. Jenkins retest.

davidangb:
down to three 504 errors in swat tests. Jenkins retest.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -315,130 +316,137 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+                              tracingContext: RawlsTracingContext
+  )(implicit
+    executionContext: ExecutionContext
+  ) =
+    // In an independent transaction,
+    //  - attach the outputs for this workflow, if any outputs exist
+    //  - update the status of the workflow in Rawls to match what Cromwell says
+    // If attaching outputs fails for legit reasons (e.g. they're missing), it will mark the workflow as failed. This is correct.
+    // If attaching outputs throws an exception (because e.g. deadlock or ConcurrentModificationException), the status will remain un-updated
+    // and will be re-processed next time we call queryForWorkflowStatus().
```
mspector, line 423: These are great comments (all of them in this PR)

I have only a passing familiarity with Scala, but am trying to build context by reviewing: would you mind pointing out where these exceptions and subsequent re-processing are handled?

It sounds like deadlocks and `ConcurrentModificationException` are expected to happen occasionally, is that right? Would we be able to tell if a workflow were stuck in a reprocessing loop somehow? Is that a possibility?

davidangb: Note that this comment is repurposed from the old line 361!

> pointing out where these exceptions and subsequent re-processing are handled

This gets into the higher-level flow of this actor. The best place to start for that is the `receive` function at line 134, which shows all the things this actor is capable of. It starts with a `StartMonitorPass` message, which will find all workflows that are still running in Rawls but are completed in Cromwell. It will then pass those workflows as the payload in a `ExecutionServiceStatusResponse` message back to self. So, if a workflow hit a non-fatal exception here, it would not be marked as complete in Rawls, and the next time the actor received a `StartMonitorPass`, it would notice that workflow is running in Rawls but complete in Cromwell and would then send it off for re-processing.

This PR doesn't do anything to help with _noticing_ reprocessing loops (though I hope it helps _avoid_ a class of them!). See https://github.com/broadinstitute/rawls/pull/2292 for work on that front.

mspector: Thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -315,130 +316,137 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-        )
-        datasource.inTransaction { dataAccess =>
-          DBIO.sequence(workflows map { workflowRecord =>
-            dataAccess.workflowQuery.updateStatus(workflowRecord, WorkflowStatuses.Failed) andThen
-              dataAccess.workflowQuery.saveMessages(Seq(AttributeString(fatal.toString)), workflowRecord.id)
-          })
+      // partition workflows into two variables: those workflows that do have
+      // outputs vs. those workflows that do not have outputs. This allows us to process
+      // the easier workflows first (those without outputs)
+      val (yesOutputs, noOutputs) = response.statusResponse
```
aednichols, line 347: I'm not sure we have more than a trivial number of no-output workflows, I've thought about eliminating this partition in the past. Seems like the tracing would help answer that question.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -242,6 +242,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           abortResults <- Future.traverse(workflowRecs) { workflowRec =>
             Future.successful(workflowRec.externalId).zip(executionServiceCluster.abort(workflowRec, petUserInfo))
           }
+          // TODO: should this also mark the workflows as Aborted in the Rawls db???
```
aednichols, line 245: I don't think so, it could end up being a lie if Rawls tells Cromwell to abort, and Cromwell for some reason does not do so.

If everything works as expected the status will get picked up from Cromwell automatically later.

aednichols (approved):
LGTM and thank you 👍 

Interesting to see the trend of grouping for efficiency, go too far and hit a wall.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: I ran a trivial calculate-an-md5 workflow in my bee. This workflow generated an output for every workflow, and the logs showed:
```
[INFO] [18:28:23.917] [rawls-submission-monitor-dispatcher-54] o.b.d.r.j.SubmissionMonitorActor - will process 18 workflow(s) without outputs and 34 workflow(s) with outputs for submission 754c7205-2808-40f1-b3e5-30e504704072
[INFO] [18:29:30.214] [rawls-submission-monitor-dispatcher-49] o.b.d.r.j.SubmissionMonitorActor - will process 10 workflow(s) without outputs and 56 workflow(s) with outputs for submission 754c7205-2808-40f1-b3e5-30e504704072
[INFO] [18:30:34.914] [rawls-submission-monitor-dispatcher-54] o.b.d.r.j.SubmissionMonitorActor - will process 0 workflow(s) without outputs and 10 workflow(s) with outputs for submission 754c7205-2808-40f1-b3e5-30e504704072
```

I wonder if - outside of this PR - there's an enhancement where `queryExecutionServiceForStatus` doesn't return any workflows which are not in a terminal status. It may be that `handleStatusResponses` is processing workflows it doesn't need to.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -748,7 +765,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           val cnt = attributeCount(update.upserts.values)
           logger.debug(
             s"Updating $cnt attribute values for entity ${update.entityRef.entityName} of type ${update.entityRef.entityType} in ${submissionId.toString}/${workflowRecord.externalId
-                .getOrElse("MISSING_WORKFLOW")}. ${safePrint(workspace.attributes)}"
+                .getOrElse("MISSING_WORKFLOW")}. ${safePrint(update.upserts)}"
```
davidangb, line 768: This is not strictly part of this PR, but as I was poring over logs, I noticed this problem. This log statement correctly outputted the count of attribute updates, but then output the contents of _workspace_ updates.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -315,130 +316,148 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+            // Doing so would potentially overwrite them with the execution service status if they'd been marked as failed by handleOutputs.
+            doRecordUpdate = !WorkflowStatuses.terminalStatuses.contains(WorkflowStatuses.withName(currentRec.status))
+            numRowsUpdated <-
+              if (doRecordUpdate) {
+                for {
+                  updateResult <- dataAccess.workflowQuery.updateStatus(currentRec,
+                                                                        WorkflowStatuses.withName(workflowRec.status)
+                  )
+                  _ = logger.info(
+                    s"workflow ${externalId(currentRec)} status change ${currentRec.status} -> ${workflowRec.status} in submission ${submissionId}"
```
davidangb, line 448: this is a new log line as of this PR - will it log too many lines?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -709,10 +728,8 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     workflowsWithOutputs.map { case (workflowRecord, outputsResponse) =>
       val outputs = outputsResponse.outputs
       logger.debug(
-        s"attaching outputs for ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}: ${outputs.size} outputs"
-      )
-      logger.debug(
-        s"output expressions for ${submissionId.toString}/${workflowRecord.externalId.getOrElse("MISSING_WORKFLOW")}: ${outputExpressionMap.size} expressions"
+        s"attaching outputs for ${submissionId.toString}/${workflowRecord.externalId
+            .getOrElse("MISSING_WORKFLOW")}: ${outputExpressionMap.size} expressions, ${outputs.size} outputs"
```
davidangb: here, I combined two log lines into one to reduce overall logging volume. Will this have any repercussions?

davidangb:
I ran a submission on 100 rows and isolated the log entries from this PR for that submission, the output is here:
[SubmissionMonitorActor.log](https://github.com/broadinstitute/rawls/files/13846947/SubmissionMonitorActor.log)

Looking for any opinions on if this is too much logging or if it's helpful.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: See also the output captured in this comment: https://github.com/broadinstitute/rawls/pull/2680#issuecomment-1879303953. The most likely case for workflows without outputs are those that have just transitioned from Submitted -> Running … by using the `.partition` here, we prioritize marking those workflows as Running in rawls as soon as possible, before writing outputs for any workflows (in this batch) that have completed.

davidangb:
@aednichols I am resetting your approval (thank you!) because I made nontrivial changes to logging, would love your opinion on those.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -315,130 +316,148 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+        * and calling processWorkflow() for each workflow in the group
+        * @param workflowGroup the workflows to process
+        * @param tracingLabel human-readable label for tracing purposes
+        * @return nothing interesting
+        */
+      def processMultipleWorkflows(
+        workflowGroup: Seq[(WorkflowRecord, Option[ExecutionServiceOutputs])],
+        tracingLabel: String
+      ) =
+        traceWithParent(tracingLabel, rootSpan) { innerSpan =>
```
cjllanwarne, line 372: I think this trace applies to the multiple workflows. Is it an anti-pattern to spin up too many sub-traces (eg one per workflow)?

cjllanwarne (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
cjllanwarne: IMO this is incredibly helpful/reassuring

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
cjllanwarne: Can't speak to repercussions (are we watching that line specifically anywhere?) but this seems fine at first glance.

Is there any way to make it more obvious what an expression vs an output is? In your example they are all `1 expression , 1 output`

cjllanwarne:
Doesn't look like your changes here made any difference to it, my main take away from your logs example was how clunky our value printing is (eg ` Updating 1 attribute values for entity HCC1143_1012 of type sample in 0ad5cef6-451e-47ea-b6f4-0f720fbe82b7/b44036b7-3482-4fc8-94ad-5afcd8e2f8d4. [First 10 items] Map(AttributeName(default,run8) -> AttributeString(gs://fc-c8f1689c-feb5-4a84-b6ec-e446f8f5a7fa/submissions/0ad5cef6-451e-47ea-b6f4-0f720fbe82b7/ga4ghMd5/b44036b7-3482-4fc8-94ad-5afcd8e2f8d4/call-md5/cacheCopy/md5sum.txt))`). And... I do wonder why we include user data in our logs in the first place, even at debug level.

mspector (approved):

davidangb:
> Doesn't look like your changes here made any difference to it … I do wonder why we include user data in our logs in the first place, even at debug level.

@cjllanwarne I did actually fix a bug here. Before this PR it looks like we *intended* to log user data, but failed at it. This PR changed that: https://github.com/broadinstitute/rawls/pull/2680/files#r1443370255. I also worry about this though and wonder if we should turn it off.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: at the `handleOutputs` level, we are adding a trace per workflow. I fear it might be too much - if those traces are unwieldy in practice we'll have to tweak after the fact.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: I tweaked this to change "output" to "attribute values" in 4da55d5c8a620abe0dc7329105a96177bc64865c, hopefully that's a little more understandable

davidangb:
There was an unrelated problem with TSPS in swat tests; should now be fixed. Jenkins retest.

davidangb:
37 failures in swat tests. Jenkins retest.

PR #2683, opened 2023-12-20
[AJ-1525] Pass Cost Saving Params to LZ when creating Billing Project

ashanhol:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1525

This PR adds an extra optional argument when creating a billing project that enables cost saving parameters in Landing Zone (specifically, Vertical Pod Autoscaling and Spot Nodes). 
Note: This is a quick PR that unblocks our ability to create workspaces with these features. If we were ever to enable this in the UI we would want to reorchestrate to allow us to pass custom parameters to LZ, which would allow us to enable any number of features on LZ in the future. 

Orchestration's API update: https://github.com/broadinstitute/firecloud-orchestration/pull/1266

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -148,8 +148,8 @@ prometheus {
 multiCloudWorkspaces {
     azureConfig {
         costSavingLandingZoneParameters = {
-          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "false"
-          "AKS_COST_SAVING_VPA_ENABLED": "false"
+          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "true"
+          "AKS_COST_SAVING_VPA_ENABLED": "true"
         }
     }
 }
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -144,3 +144,12 @@ sam.timeout = 3 minutes
 prometheus {
   endpointPort = 9098
 }
+
+multiCloudWorkspaces {
+    azureConfig {
+        costSavingLandingZoneParameters = {
+          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "true"
+          "AKS_COST_SAVING_VPA_ENABLED": "true"
+        }
+    }
+}
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -148,8 +148,8 @@ prometheus {
 multiCloudWorkspaces {
     azureConfig {
         costSavingLandingZoneParameters = {
-          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "true"
-          "AKS_COST_SAVING_VPA_ENABLED": "true"
+          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "false"
+          "AKS_COST_SAVING_VPA_ENABLED": "false"
         }
     }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -119,7 +119,10 @@ class BpmBillingProjectLifecycle(
       }
 
       val maybeAttach = if (lzId.isDefined) Map("attach" -> "true") else Map.empty
-      val params = config.azureConfig.get.landingZoneParameters ++ maybeAttach
+      val costSavingParams =
+        if (createProjectRequest.costSavings.getOrElse(false)) config.azureConfig.get.costSavingLandingZoneParameters
+        else Map.empty
+      val params = config.azureConfig.get.landingZoneParameters ++ maybeAttach ++ costSavingParams
 
       val lzDefinition = createProjectRequest.protectedData match {
         case Some(true) => config.azureConfig.get.protectedDataLandingZoneDefinition
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
@@ -120,7 +120,7 @@ class BpmBillingProjectLifecycle(
 
       val maybeAttach = if (lzId.isDefined) Map("attach" -> "true") else Map.empty
       val costSavingParams =
-        if (createProjectRequest.costSavings.getOrElse(false)) config.azureConfig.get.costSavingLandingZoneParameters
+        if (createProjectRequest.costSavings.contains(true)) config.azureConfig.get.costSavingLandingZoneParameters
         else Map.empty
       val params = config.azureConfig.get.landingZoneParameters ++ maybeAttach ++ costSavingParams
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/config/MultiCloudWorkspaceConfig.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/config/MultiCloudWorkspaceConfig.scala
@@ -44,6 +44,14 @@ case object MultiCloudWorkspaceConfig {
                 entry.getKey -> entry.getValue.unwrapped().asInstanceOf[String]
               }
               .toMap,
+            azc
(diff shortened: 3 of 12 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -294,7 +295,7 @@ class UserAuthJsonSupport extends JsonSupport {
   )
 
   implicit val CreateRawlsV2BillingProjectFullRequestFormat: RootJsonFormat[CreateRawlsV2BillingProjectFullRequest] =
-    jsonFormat7(CreateRawlsV2BillingProjectFullRequest)
+    jsonFormat8(CreateRawlsV2BillingProjectFullRequest)
```
yuliadub, line 298: not sure if we want to change this?

ashanhol: we do, we added a param
so instead of 7 params theres 8

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
```diff
@@ -119,7 +119,10 @@ class BpmBillingProjectLifecycle(
       }
 
       val maybeAttach = if (lzId.isDefined) Map("attach" -> "true") else Map.empty
-      val params = config.azureConfig.get.landingZoneParameters ++ maybeAttach
+      val costSavingParams =
+        if (createProjectRequest.costSavings.getOrElse(false)) config.azureConfig.get.costSavingLandingZoneParameters
```
yuliadub: it seems like that if the costSaving bool is included but is set to false, this code will still use the cost savings params that are hardcoded to true, or am I reading this wrong?

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
yuliadub: ah, got it :) tricky scala!

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
ashanhol: the full line is 
`val costSavingParams = if (createProjectRequest.costSavings.getOrElse(false)) config.azureConfig.get.costSavingLandingZoneParameters else Map.empty`
so there's an else that sets it to empty like the `maybeAttach` val above it

davidangb: Scala syntax goodness: using `contains` might be easier to read because it mentions the word "true":
```scala
if (createProjectRequest.costSavings.contains(true))
```
no functional difference, and this is just style, so your call to change or not

core/src/main/scala/org/broadinstitute/dsde/rawls/config/MultiCloudWorkspaceConfig.scala
```diff
@@ -43,6 +44,14 @@ case object MultiCloudWorkspaceConfig {
                 entry.getKey -> entry.getValue.unwrapped().asInstanceOf[String]
               }
               .toMap,
+            azc
+              .getConfig("costSavingLandingZoneParameters")
+              .entrySet()
+              .asScala
+              .map { entry =>
+                entry.getKey -> entry.getValue.unwrapped().asInstanceOf[String]
```
davidangb: should these be `String`s? It looks like in rawls.conf the values are all `Boolean`s, though in the test-only reference.conf you have some Strings in there. If you think there's a future use case where we need to pass strings down to WSM/LZ or to Azure for these config settings, leaving this as `String` is fine.

local-dev/templates/rawls.conf.ctmpl
```diff
@@ -401,6 +401,10 @@ multiCloudWorkspaces {
       "AKS_AUTOSCALING_MAX": "100"
       "AKS_MACHINE_TYPE": "Standard_D4as_v5"
     }
+    costSavingLandingZoneParameters = {
+      "AKS_COST_SAVING_SPOT_NODES_ENABLED": "true"
+      "AKS_COST_SAVING_VPA_ENABLED": "true"
+    }
```
davidangb: If you store `costSavingLandingZoneParameters` in rawls.conf, you'll need to make a corresponding change to the config in terra-helmfile:
https://github.com/broadinstitute/terra-helmfile/blob/a024f3908c519139ee55dbc6695f62522f53af3c/charts/rawls/templates/config/_conf.tpl#L430. The `local-dev/templates/rawls.conf.ctmpl` file is only used when running Rawls locally.

Alternately, you could put this stanza in [core/src/main/resources/reference.conf](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/reference.conf) and leave it out of rawls.conf altogether.

**reference.conf**: the default values for any setting. Requires a full rebuild/redeploy of Rawls to change, since it's right in the codebase. But, these values are designed to be overridden by values in rawls.conf, so you can effectively change them by overriding. 

**rawls.conf**: the environment-specific values for any setting. Can be changed in k8s "live", with a simple rolling restart, no rebuild or redeploy needed. 

If the `costSavingLandingZoneParameters` config stanza doesn't change between environments (dev vs. prod), and doesn't change frequently, it's a pretty good candidate to store in reference.conf. 

See https://github.com/lightbend/config?tab=readme-ov-file#standard-behavior for understanding how the Typesafe Config library layers these config files. When reading that page, substitute "rawls.conf" anywhere you see "application.conf".

davidangb (commented):
fairly minor comments inline, thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BpmBillingProjectLifecycle.scala
yuliadub: I see, the value get set to false if we cant get it, which means the if will not evaluate to true. I was confusing the getorElse and if/else together. thanks for clearing it up :)

core/src/main/scala/org/broadinstitute/dsde/rawls/config/MultiCloudWorkspaceConfig.scala
ashanhol: This was a mistake, we will be doing future LZ params which might make more sense as strings but not under costSavingLandingZoneParameters. I'll fix it.

ashanhol: Oh no wait I'm wrong, the expected params for all LZ params are string so we should match

yuliadub (approved):
I'd suggest we test this in a BEE to make sure everything looks good before merging but LGTM

ashanhol:
jenkins retest

ashanhol:
jenkins retest

jladieu (approved):

ashanhol:
Note: Billing project deploy from UI works in BEE. Have not been able to test calling the POST billing create api- seems permissions restricted

core/src/main/resources/reference.conf
```diff
@@ -144,3 +144,12 @@ sam.timeout = 3 minutes
 prometheus {
   endpointPort = 9098
 }
+
+multiCloudWorkspaces {
+    azureConfig {
+        costSavingLandingZoneParameters = {
+          "AKS_COST_SAVING_SPOT_NODES_ENABLED": "true"
```
aherbst-broad, line 151: My understanding was that the cost savings VPA + spot nodes stuff was going to be opt-in, is that the case? If so these should parameters should default to false.

ashanhol: yes thank you, changed. i had modified it in a couple places and forgot to change this back.

ashanhol: Oh jeez vacation brain, nvm i actually have to reverse this. 
These actually are true by default. The check that enables and unenables them is actually in BpmBillingProjectLifecycle. It checks the user set flag and either pulls these true parameters, or doesn't even use them if the cost savings flag is false.

PR #2698, opened 2024-01-14
[WOR-849][WOR-850][WOR-852] Create BPM billing profile for GCP billing projects.

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-849

This is the first in a group of tickets that will need to merge together, so keep on a feature branch. 

Testing: in addition to the unit tests, I created a GCP billing project using my locally running Rawls. We don't return the billing profile ID when fetching billing project details, but I was able to check in the debugger that we have persisted the ID and thus can fetch the billing profile (which we do to [add policies and Azure MRG information](https://github.com/broadinstitute/rawls/blob/f8c0f1df40bc33301c041df36fce68ba78292fa6/core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala#L315)). 

image

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.rawls.billing
 
+import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import bio.terra.profile.client.ApiException
 import bio.terra.profile.model._
@@ -16,6 +17,7 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsBillingAccountName,
   RawlsRequestContext
 }
+import org.broadinstitute.dsde.rawls.util.Retry
 import spray.json.{DefaultJsonProtocol, RootJsonFormat}
 
 import java.util.{Date, UUID}
@@ -55,6 +57,15 @@ trait BillingProfileManagerDAO {
   def getBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Option[ProfileModel]
 
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]]
+
+  @throws[ApiException]
+  def updateBillingProfile(billingProfileId: UUID,
+                           rawlsBillingAccountName: RawlsBillingAccountName,
+                           ctx: RawlsRequestContext
+  ): Future[ProfileModel]
+
+  @throws[ApiException]
+  def removeBillingAccountFromBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Future[Unit]
 
   def addProfilePolicyMember(billingProfileId: UUID,
                              policy: ProfilePolicy,
@@ -113,8 +124,15 @@ object BillingProfileManagerDAO {
 class BillingProfileManagerDAOImpl(
   apiClientProvider: BillingProfileManagerClientProvider,
   config: MultiCloudWorkspaceConfig
-) extends BillingProfileManagerDAO
-    with LazyLogging {
+)(implicit val executionContext: ExecutionContext, val system: ActorSystem)
+    extends BillingProfileManagerDAO
+    with LazyLogging
+    with Retry {
+
+  private def when5xx: Throwable => Boolean = {
+    case e: ApiException => e.getCode / 100 == 5
+    case _               => false
+  }
 
   override def listManagedApps(subscriptionId: UUID,
                                includeAssignedApps: Boolean,
@@ -211,6 +229,30 @@ class BillingProfileManagerDAOImpl(
 
     Future.successful(callListProfiles())
   }
+
+  def updateBillingProfile(billingProfileId: UUID,
+                           rawlsBillingAccountName: RawlsBillingAccountName,
+                           ctx: RawlsRequestContext
+  ): Future[ProfileModel] =
+    retry(when5xx) { () =>
+      Future {
+        apiClientProvider
+          .getProfileApi(ctx)
+          .updateProfile(
+            new UpdateProfileRequest().billingAccountId(rawlsBillingAccountName.withoutPrefix()),
+            billingProfileId
+          )
+      }
+    }
+
+  override def removeBillingAccountFromBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Future[Unit] =
+    retry(when5xx) { () =>
+      Future {
+        apiClientProvider
+          .getProfileApi(ctx)
+          .removeBillingAccount(billingProfileId)
+      }
+    }
 
   def addProfilePolicyMember(billingProfileId: UUID,
                              policy: ProfilePolicy,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -131,10 +131,6 @@ class BillingProfileManagerDAOImpl(
     policies: Map[String, List[(String, String)]] = Map.empty,
     ctx: RawlsRequestContext
   ): ProfileModel = {
-    val azureManagedAppCoordinates = billingInfo match {
-      case Left(_)       => throw new NotImplementedError("Google billing accounts not supported in billing profiles")
-      case Right(coords) => coords
-    }
 
     val policyInputs = new BpmApiPolicyInputs().inputs(
       policies
@@ -154,15 +150,29 @@ class BillingProfileManagerDAOImpl(
 
     // create the profile
     val profileApi = apiClientProvider.getProfileApi(ctx)
-    val createProfileRequest = new CreateProfileRequest()
-      .tenantId(azureManagedAppCoordinates.tenantId)
-      .subscriptionId(azureManagedAppCoordinates.subscriptionId)
-      .managedResourceGroupId(azureManagedAppCoordinates.managedResourceGroupId)
-      .displayName(displayName)
-      .id(UUID.randomUUID())
-      .biller("direct") // community terra is always 'direct' (i.e., no reseller)
-      .cloudPlatform(CloudPlatform.AZURE)
-      .policies(policyInputs)
+    val createProfileRequest = billingInfo match {
+      case Left(billingAccountName) => {
+        val rawlsBillingAccountName = billingAccountName
+        new CreateProfileRequest()
+          .billingAccountId(rawlsBillingAccountName.value)
+          .displayName(displayName)
+          .id(UUID.randomUUID())
+          .biller("direct") // community terra is always 'direct' (i.e., no reseller)
+          .cloudPlatform(CloudPlatform.GCP)
+          .policies(policyInputs)
+      } // OH, BUT IT COULD BE!!!!
+      case Right(coords) =>
+        val azureManagedAppCoordinates = coords
+        new CreateProfileRequest()
+          .tenantId(azureManagedAppCoordinates.tenantId)
+          .subscriptionId(azureManagedAppCoordinates.subscriptionId)
+          .managedResourceGroupId(azureManagedAppCoordinates.managedResourceGroupId)
+          .displayName(displayName)
+          .id(UUID.randomUUID())
+          .biller("direct") // community terra is always 'direct' (i.e., no reseller)
+          .cloudPlatform(CloudPlatform.AZURE)
+          .policies(policyInputs)
+    }
 
     logger.info(s"Creating billing profile [id=${createProfileRequest.getId}]")
     profileApi.createProfile(createProfileRequest)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -154,7 +154,7 @@ class BillingProfileManagerDAOImpl(
       case Left(billingAccountName) =>
         val rawlsBillingAccountName = billingAccountName
         new CreateProfileRequest()
-          .billingAccountId(rawlsBillingAccountName.value)
+          .billingAccountId(rawlsBillingAccountName.withoutPrefix())
           .displayName(displayName)
           .id(UUID.randomUUID())
           .biller("direct") // community terra is always 'direct' (i.e., no reseller)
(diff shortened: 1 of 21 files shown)


cahrens:
jenkins retest

cahrens:
Oops, test failures appeared with my last change. I will investigate and put back into review when they are passing.

cahrens:
jenkins retest

cahrens:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -62,6 +63,10 @@ class GoogleBillingProjectLifecycle(
                                  ctx: RawlsRequestContext
   ): Future[CreationStatus] =
     for {
+      - <- createBillingProfile(createProjectRequest, ctx).flatMap { profileModel =>
+        addMembersToBillingProfile(profileModel, createProjectRequest, ctx)
+        billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
+      }
```
marctalbott: These should be individual lines in the for-comp
```suggestion
      profileModel <- createBillingProfile(createProjectRequest, ctx)
      _ <- addMembersToBillingProfile(profileModel, createProjectRequest, ctx)
      _ <- billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -154,15 +150,29 @@ class BillingProfileManagerDAOImpl(
...
+        new CreateProfileRequest()
+          .tenantId(azureManagedAppCoordinates.tenantId)
+          .subscriptionId(azureManagedAppCoordinates.subscriptionId)
+          .managedResourceGroupId(azureManagedAppCoordinates.managedResourceGroupId)
+          .displayName(displayName)
+          .id(UUID.randomUUID())
+          .biller("direct") // community terra is always 'direct' (i.e., no reseller)
+          .cloudPlatform(CloudPlatform.AZURE)
+          .policies(policyInputs)
+    }
```
okotsopoulos: Could you break out the common set-up and only branch on the cloud-specific set-up?

(Sorry about my rusty Scala, the syntax may be off!)
```suggestion
    val createProfileRequest =
      new CreateProfileRequest()
          .displayName(displayName)
          .id(UUID.randomUUID())
          .biller("direct") // community terra is always 'direct' (i.e., no reseller)
          .policies(policyInputs)
    billingInfo match {
      case Left(billingAccountName) =>
        val rawlsBillingAccountName = billingAccountName
        createProfileRequest
          .billingAccountId(rawlsBillingAccountName.withoutPrefix())
          .cloudPlatform(CloudPlatform.GCP)
      case Right(coords) =>
        val azureManagedAppCoordinates = coords
        createProfileRequest
          .tenantId(azureManagedAppCoordinates.tenantId)
          .subscriptionId(azureManagedAppCoordinates.subscriptionId)
          .managedResourceGroupId(azureManagedAppCoordinates.managedResourceGroupId)
          .cloudPlatform(CloudPlatform.AZURE)
    }
```

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAOSpec.scala
```diff
@@ -128,16 +128,25 @@ class BillingProfileManagerDAOSpec extends AnyFlatSpec with MockitoTestUtils {
 
   behavior of "createBillingProfile"
 
-  it should "fail when provided with Google billing account information" in {
+  it should "create a GCP profile in billing profile manager if given RawlsBillingAccountName" in {
```
okotsopoulos, line 133: Nice that this was already a test case, makes it clear what test needs to change!

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
```diff
@@ -132,4 +165,126 @@ class GoogleBillingProjectLifecycleSpec extends AnyFlatSpec {
...
+      )
+    )
+      .thenReturn(profileModel)
+
+    when(
+      samDAO.syncPolicyToGoogle(
+        ArgumentMatchers.eq(SamResourceTypeNames.billingProject),
+        ArgumentMatchers.eq(createRequest.projectName.value),
+        ArgumentMatchers.eq(SamBillingProjectPolicyNames.owner)
+      )
```
okotsopoulos: I think that if all of the argument matchers are `eq` you can simplify as follows:
```suggestion
      samDAO.syncPolicyToGoogle(
        SamResourceTypeNames.billingProject,
        createRequest.projectName.value,
        SamBillingProjectPolicyNames.owner
      )
```

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
```diff
@@ -132,4 +165,126 @@ class GoogleBillingProjectLifecycleSpec extends AnyFlatSpec {
...
+    doReturn(Future.successful())
+      .when(wsmResourceRecordDao)
+      .create(any)
+
+    Await.result(bp.postCreationSteps(
+                   createRequest,
+                   mock[MultiCloudWorkspaceConfig],
+                   testContext
+                 ),
+                 Duration.Inf
```
okotsopoulos, line 217: Do we want to allow an infinite wait here?

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
```diff
@@ -132,4 +165,126 @@ class GoogleBillingProjectLifecycleSpec extends AnyFlatSpec {
...
+    )
+
+    verify(repo, Mockito.times(1)).setBillingProfileId(
```
okotsopoulos: If you prefer, I think that `Mockito.times(1)` is the default argument, so it could be eliminated to streamline.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
```diff
@@ -119,7 +139,20 @@ class GoogleBillingProjectLifecycleSpec extends AnyFlatSpec {
...
     ).thenReturn(Future.successful(Map(WorkbenchEmail(userInfo.userEmail.value) -> Seq())))
-    val gbp = new GoogleBillingProjectLifecycle(mock[BillingRepository], samDAO, mock[GoogleServicesDAO])
+    val gbp = new GoogleBillingProjectLifecycle(repo, bpm, samDAO, mock[GoogleServicesDAO])
+
+    when(repo.setBillingProfileId(createRequest.projectName, profileModel.getId))
+      .thenReturn(Future.successful(1))
+
+    when(
+      bpm.createBillingProfile(
+        ArgumentMatchers.eq(createRequest.projectName.value),
```
okotsopoulos, line 151: Since we've directly imported `ArgumentMatchers.any` it would be nice to be consistent and directly import other `ArgumentMatchers` too (a bit easier to read, IMO).

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
```diff
@@ -132,4 +165,126 @@ class GoogleBillingProjectLifecycleSpec extends AnyFlatSpec {
...
+                 ),
+                 Duration.Inf
+    )
+
+    verify(bpm, Mockito.times(2)).addProfilePolicyMember(
+      ArgumentMatchers.eq(profileModel.getId),
+      ArgumentMatchers.eq(ProfilePolicy.Owner),
+      ArgumentMatchers.argThat(arg => Set(user1Email, user2Email).contains(arg)),
+      any[RawlsRequestContext]
+    )
```
okotsopoulos: I'm a bit confused by this verification -- the emails specifically.  This would pass if `user1Email` was added to the policy twice and `user2Email` was never added, is that intended?

okotsopoulos (commented):
Looks good!  I had a few questions.  Tomorrow I'll check out your feature branch and see about creating profiles locally, I think that would be a good exercise for me.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
cahrens: I will make the change as I definitely agree it is easier to read, but functionally is there a difference?

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
cahrens: I actually prefer the more explicit form, so will change the usage of `any` to be consistent.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
cahrens: Yes, I know it's the default and I will delete it in some of the places. We just do tend to be explicit.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
cahrens: We seem to do this all over the tests. I honestly don't know what the alternative would be.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
cahrens: I just copied this test case from the Azure spec file and didn't examine it too closely. I will see about improving it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
marctalbott: There is. Each `<-` in a for-comp is equivalent to a `flatMap`, so you would need an additional `flatMap` call on `addMembersToBillingProfile` for it to be functionally equivalent to my suggestion. 
```
- <- createBillingProfile(createProjectRequest, ctx).flatMap { profileModel =>
        addMembersToBillingProfile(profileModel, createProjectRequest, ctx).flatMap { _ =>
          billingRepository.setBillingProfileId(createProjectRequest.projectName, profileModel.getId)
        }
      }
```

As you had it, Rawls won't wait for `addMembersToBillingProfile` to complete before it moves on to `billingRepository.setBillingProfileId` -- it's essentially dropping the `addMembersToBillingProfile` `Future` on the floor instead of adding it to the chain of `Future`s that will run in order.

You can see it in action using the scala interpreter with something like this

```
scala> import scala.concurrent.Future
scala> implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global

scala> def slow(): Future[Unit] = Future {
     |   Thread.sleep(500)
     |   println("i'm slow")
     | }
scala> def fast(): Future[Unit] = Future {
     |   Thread.sleep(100)
     |   println("i'm fast")
     | }

scala> for {
     |   _ <- fast().flatMap { _ =>
     |     slow()
     |     fast()
     |   }
     |   _ <- fast()
     | } yield ()

output
i'm fast
i'm fast
i'm fast
i'm slow

scala> for {
     |   _ <- fast()
     |   _ <- slow()
     |   _ <- fast()
     |   _ <- fast()
     | } yield ()

output
i'm fast
i'm slow
i'm fast
i'm fast

cahrens: Ah, thank you for the detailed response! It really helped.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
okotsopoulos: That would be great, it would be good to understand the expected behavior here.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
okotsopoulos: That's fine!

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
cahrens: I updated the test and the place I copied it from in the last commit.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
cahrens: See my most recent commit.

okotsopoulos: Looks good, thanks!

okotsopoulos (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycleSpec.scala
okotsopoulos: (Not a blocker to merge) In places where we have to wait in a test, it's usually preferable to set some kind of timeout, considering our expectations on how long something will take plus a sizable buffer to avoid needlessly flaky tests.  That way a test will fail fast(er) if it's stuck, and go onto a retry if those are configured.

cahrens: @marctalbott do you happen to know the best practice for Scala unittests? I think `Duration.Inf` is what we commonly use.

marctalbott: In this case, everything is mocked out, so I have no problem with using `Duration.Inf` here. The `Await` is a means to get the value out of the `Future` -- the timeout won't have any impact as we're not really waiting for anything in this test since Mockito will return all `Future`s quickly. 

As for other uses of `Duration.Inf` in Rawls unit tests, I'm sure there are places we're misusing it. Still, I'm not _that_ worried about those places since they're all unit tests and in theory won't be waiting for long. I'd be very concerned if we had infinite waits in our integration tests.

marctalbott (approved):

cahrens:
Converting to draft so we don't get review reminders about this (or accidentally merge it).

PR #2704, opened 2024-01-16
AJ-1551: StartMonitorPass optimization

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1551

Followup from #2680.

More optimizations inside `SubmissionMonitorActor`, this time within `queryExecutionServiceForStatus`.

1. Previously, we called `loadSubmission`. This retrieves the submission record, its method config, all the inputs and outputs for that config, all of the workflows in the submission, all input resolutions for each workflow, and the entity on which the submission ran. After loading all this, the only property of the submission we actually used was its status. This PR changes `loadSubmission` to `findById`, which skips retrieving things we don't need.
2. After querying to retrieve the submission, we would previously re-retrieve that submission in `abortActiveWorkflows` and `queryForWorkflowStatuses` (via `getWorkspaceAndSubmitter`). Now, we pass arguments where needed so we don't retrieve the same info multiple times.

Finally, a question: should `abortActiveWorkflows` mark the workflows as Aborted in the Rawls db? See TODO [here](https://github.com/broadinstitute/rawls/blob/467b75ad2edabd51f6fa4964a7686e3ab72b502f/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala#L247) and Adam's opinion [here](https://github.com/broadinstitute/rawls/pull/2680#discussion_r1443175308). My belief is:

* if we do NOT mark these workflows as aborted in the Rawls db (which is the current state), any workflows which are still Running in Rawls but completed in Cromwell will be processed by `SubmissionMonitorActor.handleStatusResponses`. The actor will attempt to write outputs for these workflows and then toggle the workflows to complete. In the case of large/slow outputs, this will be a slow process and will leave workflows in Running for a while. But, this approach preserves the most Cromwell fidelity, as it should eventually complete these workflows and persist their outputs.
* if we DO mark these workflows as aborted in the Rawls db, the abort will appear to take place "immediately" to the end user, and `SubmissionMonitorActor.handleStatusResponses` will skip processing these workflows. This has the benefit of more immediate user feedback, but it comes at the cost of potentially ignoring the successful outputs/status of some of these workflows, and doesn't handle the corner case where Cromwell refused an abort and left a workflow running even though Rawls thinks it is aborted.

I'm questioning this behavior ^ because of our experience with a big submission with troublesome outputs. We manually aborted the submission, but `handleStatusResponses` kept trying to write the outputs and we had to manually abort the individual workflows as well. I am prone to leave the behavior as-is to avoid rocking the boat.

For more discussion of this question, see https://broadinstitute.slack.com/archives/C1EH66VCM/p1705501698664409

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -223,18 +223,20 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         )
       }
 
-    def getWorkspaceAndSubmitter(dataAccess: DataAccess): ReadWriteAction[(RawlsUserEmail, WorkspaceRecord)] =
+    def getWorkspace(dataAccess: DataAccess, workspaceId: UUID): ReadWriteAction[WorkspaceRecord] =
       for {
-        submissionRec <- dataAccess.submissionQuery.findById(submissionId).result.map(_.head)
-        workspaceRec <- dataAccess.workspaceQuery.findByIdQuery(submissionRec.workspaceId).result.map(_.head)
-      } yield (RawlsUserEmail(submissionRec.submitterEmail), workspaceRec)
+        workspaceRec <- dataAccess.workspaceQuery.findByIdQuery(workspaceId).result.map(_.head)
+      } yield workspaceRec
 
-    def abortActiveWorkflows(submissionId: UUID): Future[Seq[(Option[String], Try[ExecutionServiceStatus])]] =
+    def abortActiveWorkflows(
+      submissionRec: SubmissionRecord
+    ): Future[Seq[(Option[String], Try[ExecutionServiceStatus])]] =
       datasource.inTransaction { dataAccess =>
         for {
           // look up abortable WorkflowRecs for this submission
           wfRecs <- dataAccess.workflowQuery.findWorkflowsForAbort(submissionId).result
-          (submitter, workspaceRec) <- getWorkspaceAndSubmitter(dataAccess)
+          submitter = RawlsUserEmail(submissionRec.submitterEmail)
+          workspaceRec <- getWorkspace(dataAccess, submissionRec.workspaceId)
         } yield (wfRecs, submitter, workspaceRec)
       } flatMap { case (workflowRecs, submitter, workspaceRec) =>
         for {
@@ -255,14 +257,15 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
         })
       }
 
-    def queryForWorkflowStatuses() =
+    def queryForWorkflowStatuses(submissionRec: SubmissionRecord) =
       datasource.inTransaction { dataAccess =>
         for {
           wfRecs <- dataAccess.workflowQuery.listWorkflowRecsForSubmissionAndStatuses(
             submissionId,
             WorkflowStatuses.runningStatuses: _*
           )
-          (submitter, workspaceRec) <- getWorkspaceAndSubmitter(dataAccess)
+          submitter = RawlsUserEmail(submissionRec.submitterEmail)
+          workspaceRec <- getWorkspace(dataAccess, submissionRec.workspaceId)
         } yield (wfRecs, submitter, workspaceRec)
       } flatMap { case (externalWorkflowIds, submitter, workspaceRec) =>
         for {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -212,7 +212,7 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     executionContext: ExecutionContext
   ): Future[ExecutionServiceStatusResponse] = {
     val submissionFuture = datasource.inTransaction { dataAccess =>
-      dataAccess.submissionQuery.loadSubmission(submissionId)
+      dataAccess.uniqueResult[SubmissionRecord](dataAccess.submissionQuery.findById(submissionId))
```
davidangb, line 215: Note that `uniqueResult[SubmissionRecord](findById(submissionId))` is the first line of `loadSubmission`, so this is a good replacement.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -223,18 +223,20 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
       }
 
-    def getWorkspaceAndSubmitter(dataAccess: DataAccess): ReadWriteAction[(RawlsUserEmail, WorkspaceRecord)] =
+    def getWorkspace(dataAccess: DataAccess, workspaceId: UUID): ReadWriteAction[WorkspaceRecord] =
       for {
-        submissionRec <- dataAccess.submissionQuery.findById(submissionId).result.map(_.head)
-        workspaceRec <- dataAccess.workspaceQuery.findByIdQuery(submissionRec.workspaceId).result.map(_.head)
-      } yield (RawlsUserEmail(submissionRec.submitterEmail), workspaceRec)
+        workspaceRec <- dataAccess.workspaceQuery.findByIdQuery(workspaceId).result.map(_.head)
+      } yield workspaceRec
```
davidangb, line 229: changed to skip retrieval of the submitter, and therefore renamed the method.

cjllanwarne (approved):
LGTM based on your description. We appear to be replacing "get everything and then pick a field from it" with more specific "get just the field" logic.

I don't see any tests changed here which is either a great sign (of an optimization with no differences visible externally) or a sign that we should add one. Do you know which is the case?

davidangb:
> I don't see any tests changed here which is either a great sign (of an optimization with no differences visible externally) or a sign that we should add one. Do you know which is the case?

@cjllanwarne I think it's the former. During development of this PR I made some mistakes and saw a bunch of tests fail; once I fixed the issue they started passing again. This Pr shouldn't change any functionality at all, it just optimizes SQL queries - so I'm content with the existing tests.

jladieu (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -223,18 +223,20 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
 
-    def abortActiveWorkflows(submissionId: UUID): Future[Seq[(Option[String], Try[ExecutionServiceStatus])]] =
+    def abortActiveWorkflows(
+      submissionRec: SubmissionRecord
+    ): Future[Seq[(Option[String], Try[ExecutionServiceStatus])]] =
       datasource.inTransaction { dataAccess =>
         for {
           // look up abortable WorkflowRecs for this submission
           wfRecs <- dataAccess.workflowQuery.findWorkflowsForAbort(submissionId).result
```
calypsomatic, line 237: Just checking if I'm reading this right - looks like `submissionId` used to be passed in to the method, but now we have an entire record passed in, so wouldn't this need to be updated to `submissionRec.Id`?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -273,17 +276,17 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
       } map ExecutionServiceStatusResponse
 
     submissionFuture flatMap {
-      case Some(submission) =>
-        val abortFuture = if (submission.status == SubmissionStatuses.Aborting) {
+      case Some(submissionRec) =>
+        val abortFuture = if (SubmissionStatuses.withName(submissionRec.status) == SubmissionStatuses.Aborting) {
           // abort workflows if necessary
           for {
             _ <- abortQueuedWorkflows(submissionId)
```
calypsomatic, line 283: Same here - do we still have a `submissionId` variable?

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
davidangb: Good question - but I don't think this needs to change. `SubmissionMonitorActor` accepts a `submissionId` in its constructor, so this line of code changed from using the `submissionId` argument to using the `submissionId` class-level variable. Previously, the class-level variable IS what was passed as the argument.

calypsomatic: Huh.  So was the `submissionId` argument unnecessary before?  If the instance of `SubmissionMonitorActor` is dedicated to a particular submission, does that mean the whole `submissionRec` could be an instance variable instead of passing around?  Not proposing that, just wondering how it works.

davidangb: heh yes, the `submissionId` argument was unnecessary before. Turning `submissionRec` into an instance variable is less ideal … since this actor operates repeatedly over time for the same submission, the submission _id_ is constant, but other values in `submissionRec` can and do change.

calypsomatic (approved):

PR #2719, opened 2024-01-30
WOR-1480 use open telemetry to emit metrics

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1480

1. replace use of OpenTelemetryMetrics (which does not actually use open telemetry) with the real thing
2. use open telemetry to produce prometheus page
3. better way of figuring out standard http route (by grabbing it from swagger)

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/InstrumentationDirectives.scala
```diff
@@ -57,11 +59,15 @@ trait InstrumentationDirectives extends RawlsInstrumented with TracingDirectives
     ))
 
   private val redactWorkspaceNames =
-    (Slash ~ "api").? / "workspaces" / (!"entities" ~ Segment) / (Segment ~ SegmentIgnore.repeat(0,
-                                                                                                 Int.MaxValue,
-                                                                                                 separator = Slash
+    (Slash ~ "api").? / "workspaces" / (!("entities" | "id") ~ Segment) / (Segment ~ SegmentIgnore.repeat(0,
```
dvoet, line 59: this and the following change made me look for better and more future proof ways to figure out a standard name for an api "route".  We have this all in swagger, why not parse that?

tlangs (approved):

dvoet:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassMonitor.scala
```diff
@@ -18,27 +17,20 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
...
-)(implicit openTelemetry: OpenTelemetryMetrics[IO])
-    extends Actor
+) extends Actor
     with LazyLogging {
   import context.dispatcher
 
-  private val openTelemetryTags: Map[String, String] = Map("service" -> "FastPassMonitor")
   override def receive: Receive = { case DeleteExpiredGrants =>
-    for {
-      _ <- openTelemetry.incrementCounter("fastpass-monitor-sweeper-start", tags = openTelemetryTags).unsafeToFuture()
```
dvoet, line 39: note that I removed this metric because it has no value - this function is called on a schedule every 10s

aherbst-broad (approved):

PR #2736, opened 2024-02-14
[WOR-852] Update billing profile when updating billing account

marctalbott:
Ticket: [WOR-852](https://broadworkbench.atlassian.net/browse/WOR-852)
* When Rawls updates or removes the billing account on a billing project, perform the same operation on the billing project's billing profile if it exists
* Add workaround for PATCH requests to BPM API client

---

**PR checklist**

[WOR-852]: https://broadworkbench.atlassian.net/browse/WOR-852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
@@ -5,9 +5,11 @@ import bio.terra.profile.api.{AzureApi, ProfileApi, SpendReportingApi, Unauthent
 import bio.terra.profile.client.ApiClient
 import io.opencensus.trace.Tracing
 import io.opentelemetry.api.GlobalOpenTelemetry
+import jakarta.ws.rs.client.ClientBuilder
 import org.broadinstitute.dsde.rawls.model.RawlsRequestContext
 import org.broadinstitute.dsde.rawls.util.{TracingUtils, WithOtelContextFilter}
 import org.glassfish.jersey.client.ClientConfig
+import org.glassfish.jersey.jdk.connector.JdkConnectorProvider
 
 /**
  * Implementors of this trait know how to instantiate billing profile manager client
@@ -28,6 +30,11 @@ trait BillingProfileManagerClientProvider {
 class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extends BillingProfileManagerClientProvider {
   def getApiClient(ctx: RawlsRequestContext): ApiClient = {
     val client: ApiClient = new ApiClient()
+
+    val clientConfig = new ClientConfig()
+    clientConfig.connectorProvider(new JdkConnectorProvider)
+    client.setHttpClient(ClientBuilder.newClient(clientConfig))
+
     TracingUtils.enableCrossServiceTracing(client.getHttpClient, ctx)
     client.setBasePath(basePath)
     client.setAccessToken(ctx.userInfo.accessToken.token)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -222,15 +222,13 @@ class BillingProfileManagerDAOImpl(
   def updateBillingProfile(billingProfileId: UUID,
                            rawlsBillingAccountName: RawlsBillingAccountName,
                            ctx: RawlsRequestContext
-  ): ProfileModel = {
-    logger.info(s"updating profile $billingProfileId to billing account $rawlsBillingAccountName")
+  ): ProfileModel =
     apiClientProvider
       .getProfileApi(ctx)
       .updateProfile(
         new UpdateProfileRequest().billingAccountId(rawlsBillingAccountName.withoutPrefix()),
         billingProfileId
       )
-  }
 
   override def removeBillingAccountFromBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Unit =
     apiClientProvider
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
@@ -56,6 +56,13 @@ trait BillingProfileManagerDAO {
 
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]]
 
+  def updateBillingProfile(billingProfileId: UUID,
+                           rawlsBillingAccountName: RawlsBillingAccountName,
+                           ctx: RawlsRequestContext
+  ): ProfileModel
+
+  def removeBillingAccountFromBillingProfile(billingProfileId: UUID, ctx: RawlsRequestContext): Unit
+
   def addProfilePolicyMember(billingProfileId: UUID,
                              policy: ProfilePolicy,
                              memberEmail: String,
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -28,6 +30,11 @@ trait BillingProfileManagerClientProvider {
 class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extends BillingProfileManagerClientProvider {
   def getApiClient(ctx: RawlsRequestContext): ApiClient = {
     val client: ApiClient = new ApiClient()
+
+    val clientConfig = new ClientConfig()
```
marctalbott, line 34: See https://github.com/DataBiosphere/jade-data-repo/pull/1288/files#diff-608bbfbb0303e87fa7cb15128f663bd479532f68c996e27d031c68f241d628aeR90 and https://broadinstitute.slack.com/archives/C03B4RKTA5B/p1651527855852349 for more details. 

TL;DR: There's a limitation in the `HttpUrlConnection` library that the jersey `HttpUrlConnector` tries to workaround. The workaround doesn't work in Java 17 which prevents us from calling PATCH endpoints using our codegen clients in Rawls. Phil and Shelby were able to get it working using this fix which sets up a different connector provider. I have not stepped through this connector provider as closely, so I'm not sure exactly how it gets around the limitation, but I can confirm that it works in practice.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -851,13 +852,35 @@ class UserService(
...
+          case Some(newBillingAccount) =>
+            billingProfileManagerDAO.updateBillingProfile(UUID.fromString(billingProfileId), newBillingAccount, ctx)
+          case None =>
+            billingProfileManagerDAO.removeBillingAccountFromBillingProfile(UUID.fromString(billingProfileId), ctx)
+        }
+      }
+    catch {
+      // Until BPM is the system of record for Terra billing information, Rawls will not throw an exception if BPM fails to update
+      case e: ApiException =>
+        logger.warn(
```
aherbst-broad: Can this go to sentry too? I think even if we're not the SOR we want to know about it and a log entry is very likely to get missed.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -851,13 +852,35 @@ class UserService(
...
   }
 
+  private def updateBillingAccountInBillingProfile(project: RawlsBillingProject,
+                                                   billingAccount: Option[RawlsBillingAccountName]
+  ): Unit =
+    try
+      project.billingProfileId.map { billingProfileId =>
+        billingAccount match {
+          case Some(newBillingAccount) =>
+            billingProfileManagerDAO.updateBillingProfile(UUID.fromString(billingProfileId), newBillingAccount, ctx)
```
aherbst-broad: These should get wrapped in a retryOn5xx predicate in the not-unlikely event there is an issue with an intermediary (gateway timeout etc) , either here or in the DAO.

aherbst-broad (commented):
Looking good, had a comment re: retries.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: Good call, I'll make sure they go to sentry

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -56,6 +56,15 @@ trait BillingProfileManagerDAO {
 
   def getAllBillingProfiles(ctx: RawlsRequestContext)(implicit ec: ExecutionContext): Future[Seq[ProfileModel]]
 
+  @throws[ApiException]
```
cahrens, line 61: I don't think I've seen this annotation before. Should all of our DAO methods be including this?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
cahrens: What things specifically go to Sentry?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
marctalbott: It was needed to mock out throwing an exception. It seems like there are other places where it's not needed to do the same thing, but I assume that the other DAO methods don't include it because there's no test where we mock out an exception for those methods. I'm not inclined to add it to the other methods preemptively, but I could be convinced

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: In general or just in this case?

In this case, we want to emit an event to Sentry when a billing profile fails to update so we have a sense for how often that's happening and will know when a billing project and its billing profile are out of sync.

cahrens: What I am asking is what controls whether or not an error goes to Sentry-- does the exception need to be rethrown with specific error codes, or we need to call logger.error, or a different method call…?

marctalbott: Oh I see, 5xx errors from Rawls will automatically be sent to Sentry, and I think error-level log messages as well? You can also manually send events to Sentry which is what I'm planning on doing here

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -851,13 +853,41 @@ class UserService(
...
+    (billingAccount match {
+      case Some(newBillingAccount) =>
+        billingProfileManagerDAO
+          .updateBillingProfile(UUID.fromString(billingProfileId), newBillingAccount, ctx)
+          .flatMap(_ => Future.unit)
+      case None =>
+        billingProfileManagerDAO.removeBillingAccountFromBillingProfile(UUID.fromString(billingProfileId), ctx)
+    }).recover {
+      // Until BPM is the system of record for Terra billing information, Rawls will not throw an exception if BPM fails to update
+      case e: ApiException =>
```
aherbst-broad: Nit: Should this cast a wider net, like Exception? I am thinking about ProcessingExceptions, SocketExceptions etc. that can crop up that are not in the ApiException hierarchy.

aherbst-broad (approved):
lgtm

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
marctalbott: Yeah, that's a good call. I was mostly concerned about preventing bugs in BPM from causing issues, but this should really cover any issues related to the call to BPM

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
```diff
@@ -212,6 +230,30 @@ class BillingProfileManagerDAOImpl(
     Future.successful(callListProfiles())
   }
 
+  def updateBillingProfile(billingProfileId: UUID,
+                           rawlsBillingAccountName: RawlsBillingAccountName,
+                           ctx: RawlsRequestContext
+  ): Future[ProfileModel] =
+    retry(when5xx) { () =>
```
sergiygetlin, line 237: I wonder where we define number of attempts for this retry?

sergiygetlin (approved):
LGTM; Have one question.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerDAO.scala
marctalbott: It's defined in the `Retry` trait. We retry 3 times after the initial failure -- after 100ms, 1s, and 3s. There are other options as well, but the default seemed fine to me

PR #2745, opened 2024-02-20
[WOR-1452] Call Leo during Google workspace deletion to cleanup dangling resources

aherbst-broad:
[**Ticket**](https://broadworkbench.atlassian.net/browse/WOR-1452)
[Design doc](https://docs.google.com/document/d/13cfxxjTGXdn48m6kXSN5AL-D9usk5yFmBQ0nQaV0L5Y/edit?ouid=104924582591628367950&usp=docs_home&ths=true)
[Related Leo PR](https://github.com/DataBiosphere/leonardo/pull/4170)

## Context
We want to notify Leonardo at the time of **Google** workspace deletion to cleanup any remaining internal Leo DB records and their linked Sam resources. The premise is that Rawls will delete the workspace google cloud project as part of workspace deletion, cleaning up the actual cloud resources, so Leo does not have anything further to do in GCP and only needs to reconcile it's state.

## This PR
* Adds the needed API to the Leo DAO
* Refactors the current `LeonardoDeletionAction` class to a `LeonardoService` class. The intention is to keep the Leo DAO "dumb" and let the service class contain the requisite retry and error handling logic. Rawls' current pattern for DAOs is inconsistent. For example, HttpGoogleServicesDAO does all of the above, while the WSM DAO is much lighter weight and pushes the error handling and retries to the callsites. 
   * I'd like to centralize on a pattern of "bare, dumb DAO" + "smarter service that knows how to retry, handle errors and propagate a sane exception hierarchy to upper layers". This is a baby step in that direction.
* Removes support for deleting MC azure workspaces via the current WorkspaceService deleteWorkspace codepath. Nothing calls this codepath for MC workspaces anymore and so we shouldn't need the added complexity of the `isAzureMcWorkspace` checks that are peppered throughout that code.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/leonardo/LeonardoService.scala
```diff
@@ -14,12 +15,18 @@ import org.broadinstitute.dsde.workbench.client.leonardo.model.{
   ListAppResponse,
   ListRuntimeResponse
 }
+import org.broadinstitute.dsde.workbench.model.Notifications.WorkspaceName
 
 import java.util.UUID
 import scala.concurrent.{blocking, ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
-class LeonardoResourceDeletionAction(leonardoDAO: LeonardoDAO)(implicit
```
aherbst-broad, line 22: Much of the logic in this class, although oriented around deletion, seemed reusable enough that I wanted to capture it in a dedicated service class so others who are reaching for leonardo APIs can use this and have sane retry and error handling. 

There are still a few instances of using a "bare" Leo DAO in the rawls codebase that should get moved over eventually.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/leonardo/LeonardoService.scala
```diff
@@ -121,7 +128,38 @@ class LeonardoResourceDeletionAction(leonardoDAO: LeonardoDAO)(implicit
...
+            s"Unexpected failure cleaning up leonardo workspace resources for workspaceId=$workspaceId . Received ${t.getCode}: [${t.getResponseBody}]"
+          )
+          Future.failed(t)
+        } else {
+          Future.successful()
+        }
+      }
+    }
 
-class LeonardoOperationFailureException(message: String, val workspaceId: UUID)
```
aherbst-broad, line 126: This wasn't being used anywhere outside of a test. I'd like to rethink the overall approach in this class to returning the bare Leo client ApiException rather than a higher level exception like this as I think there's real utility there (at the very least, it'd prevent mixing up wsm.ApiException, bpm.ApiException etc. as I often do.)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -655,12 +660,14 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
   private def deleteWorkspaceInternal(workspaceContext: Workspace,
                                       maybeMcWorkspace: Option[WorkspaceDescription],
                                       parentContext: RawlsRequestContext
-  ): Future[WorkspaceDeletionResult] =
-    for {
-      _ <- Applicative[Future].unlessA(isAzureMcWorkspace(maybeMcWorkspace))(
-        assertNoGoogleChildrenBlockingWorkspaceDeletion(workspaceContext)
+  ): Future[WorkspaceDeletionResult] = {
+    if (isAzureMcWorkspace(maybeMcWorkspace)) {
```
aherbst-broad, line 664: This code path is not supported anymore for MC workspaces, it's all flowing through the MC service [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala#L114) which prevents this code from being called w/mc workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -701,20 +708,36 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-                                   ctx.userInfo
-          ) recoverWith { case t: Throwable =>
-            logger.error(
-              s"Unexpected failure deleting workspace (while deleting google project) for workspace `${workspaceContext.toWorkspaceName}`",
-              t
-            )
-            Future.failed(t)
-          }
-        } else Future.successful()
+        maybeDeleteGoogleProject(workspaceContext.googleProjectId,
```
aherbst-broad, line 713: This is the crux of the change. We are deleting the google cloud project (but not the sam resource as we still need the permissions so leo's sam assertions still work). Once leo has cleaned up it's resource records, we can clear out the google project sam resource.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1308,7 +1300,7 @@ class WorkspaceServiceSpec
     }
   }
 
-  it should "delete an Azure workspace" in withTestDataServices { services =>
```
aherbst-broad, line 1311: This is all not applicable anymore.

project/Dependencies.scala
```diff
@@ -130,7 +130,7 @@ object Dependencies {
   val billingProfileManager = clientLibExclusions("bio.terra" % "billing-profile-manager-client" % "0.1.508-SNAPSHOT")
   val terraCommonLib = tclExclusions(clientLibExclusions("bio.terra" % "terra-common-lib" % "0.1.11-SNAPSHOT" classifier "plain"))
   val sam: ModuleID = clientLibExclusions("org.broadinstitute.dsde.workbench" %% "sam-client" % "0.1-70fda75")
-  val leonardo: ModuleID = "org.broadinstitute.dsde.workbench" % "leonardo-client_2.13" % "1.3.6-acbd45d"
+  val leonardo: ModuleID = "org.broadinstitute.dsde.workbench" % "leonardo-client_2.13" % "1.3.6-45d1ea4-SNAP"
```
aherbst-broad: TODO: Update w/the real dependency version when [this](https://github.com/DataBiosphere/leonardo/pulls) PR merges.

marctalbott (approved):

LizBaldo (approved):
LGTM on the logic of deleting the cloud resource -> call Leonardo cleanupAll -> then delete the project resource in Sam logic 👍 

The improvements on the LeonardoService make sense to me as well, but I am not super familiar with Rawls

blakery (approved):

yonghaoy (approved):
LG from AoU. Thanks for doing this. To confirm, what I throught before: forceDelete would be a query parameter in `DELETE` 
But now seems no api interface change which means all workspace deletion will be force delete.  AoU would like this because no code change on our end. But would Terra UI be okay with this as well?

aherbst-broad:
@yonghaoy Not sure I understand. This is invoking the `deleteAllResources` endpoint in Leo after the parent google project has been deleted, so we're not really forcing deletion so much as we are informing leo to clean up it's references to already deleted cloud resources.

yonghaoy:
> @yonghaoy Not sure I understand. This is invoking the `deleteAllResources` endpoint in Leo after the parent google project has been deleted, so we're not really forcing deletion so much as we are informing leo to clean up it's references to already deleted cloud resources.

My original thought: 
`deleteWorkspace` API has a parameter `deleteAllResource`. 
When `deleteAllResource=true`, call Leo, otherwise, do not call Leo.

Always calling Leo is a better choice, thanks for clarification!

PR #2757, opened 2024-02-28
[WOR-850] Fix deletion workflow for Azure billing project

cahrens:
This PR fixes the problem I introduced in WOR-850 related to Azure billing project deletion.

Previous to WOR-850, we decided which billing project lifecycle (GCP or Azure) to call based on whether not the billing project had a billing profile ID. Since GCP projects will begin having billing profile IDs, I needed to key off of something else. I noticed that the BillingProject model object that is passed around has an `AzureManagedAppCoordinates` variable, so I decided to use that out of expediency.

The problem is that the `AzureManagedAppCoordinates` don't actually exist in the database record, and unless explicitly populated will be None. This broke Azure billing project deletion because we ended up calling the Google billing project lifecycle for deletion (leaving behind the landing zone).

This PR take the following approach:
1) Moves billing project and billing profile deletion code from `BillingProjectLifecycle` (the shared base class) to a new class, `BillingProjectDeletion`.
2) On all billing projects,, call the GCP method which will attempt to delete the parent Google project. This will never error.
3) On all billing projects, attempt to cleanup any landing zone that exists.

A nice side effect of this is that `BPMBillingProjectDeleteRunner` can now depend on `BillingProjectDeletion` instead of the bigger API of `BillingProjectLifecycle`.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -557,6 +559,7 @@ object Boot extends IOApp with LazyLogging {
                                            workspaceManagerDAO,
                                            workspaceManagerResourceMonitorRecordDao
           ),
+          billingProjectDeletion,
```
cahrens, line 562: Billing project and billing profile deletion methods are moved out of `BillingProjectLifecycle` and into `BillingProjectDeletion`.

`BillingProjectOrchestrator` uses `billingProfileManagerDAO` to get hints about cloud platform, though arguably we could remove that.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectDeletion.scala
```diff
@@ -0,0 +1,92 @@
...
+    * Delete the billing project and associated billing profile.
+    */
+  def finalizeDelete(projectName: RawlsBillingProjectName,
+                     ctx: RawlsRequestContext,
+                     billingProfileExpected: Boolean = true
+  )(implicit
+    executionContext: ExecutionContext
+  ): Future[Unit] = deleteBillingProfileAndUnregisterBillingProject(projectName, billingProfileExpected, ctx)
+
+  def deleteBillingProfileAndUnregisterBillingProject(projectName: RawlsBillingProjectName,
```
cahrens: Code all taken from `BillingProjectLifecycle`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -198,7 +197,21 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
             )
           )
       }
-      jobControlId <- projectLifecycle.initiateDelete(projectName, ctx)
+      billingProfileId <- billingRepository.getBillingProfileId(projectName)
+
+      maybeGoogleProject = billingProfileId match {
```
cahrens: I am getting the CloudPlatform if possible to avoid making the GCP call to delete the parent Google project when we KNOW that the project is an Azure one.  I also use it to filter out a warning log message in the AzureBillingProjectLifecycle method.

Arguably, I could remove the optimization/filtering and simplify the code.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectDeletionSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+    billingProjectName,
+    Some(RawlsBillingAccountName("fake_billing_account_name")),
+    None,
+    None,
+    None,
+    None
+  )
+  val profileModel: ProfileModel = new ProfileModel().id(UUID.randomUUID())
+
+  behavior of "finalizeDelete"
```
cahrens, line 48: Test cases moved from Azure/GCP BillingProjectLifecycleSpec.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectDeletionSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+    intercept[BpmApiException] {
+      Await.result(billingProjectDeletion.finalizeDelete(billingProjectName, testContext), Duration.Inf)
+    }
+    verify(bpm).deleteBillingProfile(ArgumentMatchers.eq(profileModel.getId), ArgumentMatchers.eq(testContext))
+    verify(repo, Mockito.never).deleteBillingProject(ArgumentMatchers.eq(billingProjectName))
+  }
+
+  behavior of "unregisterBillingProject"
+
+  it should "delete the project even if Sam deleteResource fails" in {
```
cahrens, line 147: Test case taken from BillingProjectLifecycleSpec.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -198,7 +197,21 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+        case Some(profileId) =>
+          val billingProfile = billingProfileManagerDAO.getBillingProfile(UUID.fromString(profileId), ctx)
+          val googleOrUnknownPlatform = billingProfile match {
+            case Some(profile) => profile.getCloudPlatform.equals(CloudPlatform.GCP)
+            case None          => true
+          }
+          googleOrUnknownPlatform
+        case None => true
+      }
+      _ = googleBillingProjectLifecycle.maybeCleanupResources(projectName, maybeGoogleProject, ctx)
```
cahrens: This will attempt to delete the parent Google project. I left the method in the `GoogleBillingProjectLifecycle` because it requires the gcsDAO.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -198,7 +197,21 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
+
+      jobControlId <- azureBillingProjectLifecycle.maybeCleanupResources(projectName, maybeGoogleProject, ctx)
```
cahrens: This will attempt to delete the landing zone if the billing project record contains a `landingZoneId`. I left this in `AzureBillingProjectLifecycle` because it is an Azure-specific resource.

cahrens:
Passed Azure e2e tests: https://github.com/broadinstitute/rawls/actions/runs/8101781076

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
marctalbott: We will only make a GCP call when there is a `google-project` Sam resource that is the child resource of the `billing-project` Sam resource. That will only be the case for older billing projects.

Since this is really substituting a BPM call for a Sam call, I'd vote for simplification

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
marctalbott: I'd prefer to see the contents of `maybeCleanupResources` here in the orchestrator i.e. make a direct call to `deleteGoogleProjectIfChild` and check for LZ id existence. We can leave the actual Google project deletion and LZ deletion to other services.

I'm a bit confused as to why the `gcsDao` (and `samDao`) is passed into `deleteGoogleProjectIfChild` as a parameter in the first place. The `UserService` is constructed with these DAOs.

Leaving the specifics of landing zone deletion to the Azure lifecycle makes sense, but checking for a LZ id isn't cloud-specific IMO.

cahrens: OK, I will work on these simplifications.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/AzureBillingProjectLifecycleSpec.scala
```diff
@@ -831,30 +841,7 @@ class AzureBillingProjectLifecycleSpec extends AnyFlatSpec {
     }
   }
 
-  behavior of "initiateDelete"
-
-  it should "return None for the jobId if the landing zone does not exist" in {
```
cahrens, line 836: `BillingProjectOrchestrator` now gets the `landingZoneId` and only calls `cleanupLandingZone` if one exists.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/AzureBillingProjectLifecycle.scala
```diff
@@ -211,23 +210,4 @@ class AzureBillingProjectLifecycle(
         case None => Some(landingZoneResponse)
       }
   }
-
-  override def maybeCleanupResources(projectName: RawlsBillingProjectName,
```
cahrens: Moved to `BillingProjectOrchestrator` as `maybeDeleteLandingZone`.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
cahrens: Done

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/AzureBillingProjectLifecycle.scala
```diff
@@ -184,7 +183,7 @@ class AzureBillingProjectLifecycle(
     *  starts a landing zone deletion job
     *  does not ensure that the landing zone deletion completes successfully.
     */
-  private def cleanupLandingZone(
+  def cleanupLandingZone(
```
cahrens, line 186: Called by `BillingProjectOrchestrator` if the billing project has a `landingZoneId` (called one other place within `AzureBillingProjectLifecycle` as well).

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -49,10 +49,9 @@ import scala.concurrent.{ExecutionContext, Future}
 class BillingProjectOrchestrator(ctx: RawlsRequestContext,
                                  val samDAO: SamDAO,
                                  notificationDAO: NotificationDAO,
-                                 billingProfileManagerDAO: BillingProfileManagerDAO,
                                  billingRepository: BillingRepository,
-                                 googleBillingProjectLifecycle: BillingProjectLifecycle,
-                                 azureBillingProjectLifecycle: BillingProjectLifecycle,
+                                 googleBillingProjectLifecycle: GoogleBillingProjectLifecycle,
```
cahrens, line 53: We now access non-shared methods/properties.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/GoogleBillingProjectLifecycle.scala
```diff
@@ -69,20 +64,4 @@ class GoogleBillingProjectLifecycle(
...
-                                     ctx: RawlsRequestContext
-  )(implicit
-    executionContext: ExecutionContext
-  ): Future[Option[UUID]] =
-    // Note: GoogleBillingProjectLifecycleSpec does not test that this method is called because the method
-    // lives in a companion object (which makes straight mocking impossible), and the method will be removed
-    // once workspace migration is complete. Note also that the more "integration" level test BillingApiServiceV2Spec
-    // does verify that code in this method is executed when a Google-based project is deleted.
-    if (maybeGoogleProject) {
-      deleteGoogleProjectIfChild(projectName, ctx.userInfo, gcsDAO, samDAO, ctx).map(_ => None)
```
cahrens: Moved to `BillingProjectOrchestrator`.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ApiServiceSpec.scala
```diff
@@ -67,7 +67,6 @@ import org.mockito.ArgumentMatcher
 import org.scalatest.concurrent.Eventually
 import spray.json._
 
-import java.time.temporal.ChronoUnit
```
cahrens, line 70: Unused import… just happened to notice it.

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -466,140 +443,109 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
-    Await.result(bpo.deleteBillingProjectV2(billingProjectName), Duration.Inf)
+    val azureBillingProjectLifecycle = mock[AzureBillingProjectLifecycle]
+    when(azureBillingProjectLifecycle.cleanupLandingZone(landingZoneId, testContext)).thenReturn(None)
 
-    // Will only call Azure maybeCleanupResources because we know cloud context is Azure
-    verify(billingProjectLifecycle).maybeCleanupResources(billingProjectName, maybeGoogleProject = false, testContext)
-    verify(billingProjectDeletion).finalizeDelete(billingProjectName, testContext, true)
-  }
-
-  it should "call the Azure lifecycle to initiate delete of an Azure project when maybeCleanupResources returns a jobId" in {
```
cahrens: Folded into "create a job to delete the Azure project and set the status of the billing project to Deleting if cleanupLandingZone returns returns a job report".

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -610,107 +556,53 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
+      )
     )
+    doReturn(Future.successful()).when(bpo).maybeDeleteGoogleProject(billingProjectName, testContext)
 
     Await.result(bpo.deleteBillingProjectV2(billingProjectName), Duration.Inf)
 
     verify(monitorRecordDao).create(ArgumentMatchers.argThat(matchedExpectedEvent))
-  }
-
-  it should "not create a job to delete the Azure project after calling maybeCleanupResources fails" in {
```
cahrens: Folded into "fail if cleanupLandingZone throws an exception and not create a job to delete the Azure project".

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -610,107 +556,53 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
...
-      mock[MultiCloudWorkspaceConfig],
-      monitorRecordDao
-    )
-
-    intercept[Exception](Await.result(bpo.deleteBillingProjectV2(billingProjectName), Duration.Inf))
-
-    verify(monitorRecordDao, never).create(ArgumentMatchers.any())
-  }
-
-  it should "set the status of the billing project to Deleting after successful delete initializing" in {
```
cahrens, line 579: Folded into "create a job to delete the Azure project and set the status of the billing project to Deleting if cleanupLandingZone returns returns a job report".

core/src/test/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestratorSpec.scala
```diff
@@ -721,52 +613,6 @@ class BillingProjectOrchestratorSpec extends AnyFlatSpec {
     )
   }
 
-  it should "delete the billing project if a billing profile ID exists, but the profile cannot be obtained" in {
```
cahrens: We no longer check the billing profile.

cahrens:
@marctalbott @blakery I have updated the implementation per Marc's feedback-- see the last commit.

The Azure e2e tests passed, demonstrating that deleting an Azure billing project works. 
https://github.com/broadinstitute/rawls/actions/runs/8126637028

blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -207,14 +209,47 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
     } yield ()
 
+  /**
+    * In a method for unit test mocking.
+    */
+  def maybeDeleteGoogleProject(projectName: RawlsBillingProjectName, ctx: RawlsRequestContext)(implicit
+    executionContext: ExecutionContext
+  ): Future[Unit] =
+    deleteGoogleProjectIfChild(projectName, ctx.userInfo, googleBillingProjectLifecycle.gcsDAO, samDAO, ctx).map(_ =>
+      None
```
marctalbott: Why does this map to `None`?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -198,7 +196,11 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
             )
           )
       }
-      jobControlId <- projectLifecycle.initiateDelete(projectName, ctx)
+      // If backed by a Google project (v1 workspaces/billing projects), delete it.
+      _ = maybeDeleteGoogleProject(projectName, ctx)
```
marctalbott: This needs to be an arrow since `maybeDeleteGoogleProject` is a `Future`

marctalbott (approved):
2 comments but generally looks good

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
cahrens: It was just leftover from when it needed to conform to a common API with the Azure implementation. I have removed it.

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
cahrens: Thanks, done!

PR #2763, opened 2024-03-01
[WOR-1543] Switch connector providers

marctalbott:
Ticket: [WOR-1543](https://broadworkbench.atlassian.net/browse/WOR-1543)
* Switch to `JavaNetHttpConnectorProvider`. Created 50+ Azure billing projects without NPE issues
    * Also tried `GrizzlyConnectorProvider`, but ran into timeout issues for requests that took >10 seconds and wasn't able to figure out how to bump the timeout

---

**PR checklist**

[WOR-1543]: https://broadworkbench.atlassian.net/browse/WOR-1543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
@@ -31,6 +31,9 @@ class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extend
   def getApiClient(ctx: RawlsRequestContext): ApiClient = {
     val client: ApiClient = new ApiClient()
 
+    // By default, the client uses the `HttpUrlConnectorProvider` which relies on a workaround for
+    // PATCH endpoints that is incompatible with Java 17. Specifying a different ConnectorProvider
+    // allows us to call PATCH endpoints in BPM.
     val clientConfig = new ClientConfig()
     clientConfig.connectorProvider(new JavaNetHttpConnectorProvider())
     client.setHttpClient(ClientBuilder.newClient(clientConfig))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -32,7 +32,7 @@ class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extend
     val client: ApiClient = new ApiClient()
 
     val clientConfig = new ClientConfig()
-    clientConfig.connectorProvider(new JdkConnectorProvider)
```
aherbst-broad, line 35: I suggest a comment here explaining the nature of the workaround for future devs.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
```diff
@@ -32,7 +32,7 @@ class HttpBillingProfileManagerClientProvider(baseBpmUrl: Option[String]) extend
...
+    clientConfig.connectorProvider(new JavaNetHttpConnectorProvider())
```
cahrens, line 38: Can we remove `jersey-jdk-connector` as a dependency? Though I'm not finding it, so perhaps it is inherited from somewhere else.

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProfileManagerClientProvider.scala
marctalbott: Yeah, it's pulled in via WSM so there's no direct dependency to remove

PR #2769, opened 2024-03-04
[WOR-1452] Call Leo during Google workspace deletion to cleanup dangling resources (take 2)

aherbst-broad:
Note this is a re-introduction of the code that was reverted on 3/1. There was concern that deleting the google project _prior_ to notifying leo to delete its own sam resources would cause issues. Leo was making sam calls to delete resources using the pet SA from the deleted google project. While deleting the google project takes some time (up to 24 hours I think), Leo is being [modified](https://github.com/DataBiosphere/leonardo/pull/4283) to use the user's token instead of the pet.

This PR has also modified the deletion order of operations to notify leo to clean up resources, then delete the google project. The previous order of operations was to delete the google project, notify leo, then delete the google project sam resource. It should be ok to notify leo to cleanup, then proceed onward with gcp project deletion and gcp project sam resource cleanup. 

**Original PR description**

[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1452)
[Design doc](https://docs.google.com/document/d/13cfxxjTGXdn48m6kXSN5AL-D9usk5yFmBQ0nQaV0L5Y/edit?ouid=104924582591628367950&usp=docs_home&ths=true)
https://github.com/DataBiosphere/leonardo/pull/4170

Context

We want to notify Leonardo at the time of Google workspace deletion to cleanup any remaining internal Leo DB records and their linked Sam resources. The premise is that Rawls will delete the workspace google cloud project as part of workspace deletion, cleaning up the actual cloud resources, so Leo does not have anything further to do in GCP and only needs to reconcile it's state.

This PR

Adds the needed API to the Leo DAO
Refactors the current LeonardoDeletionAction class to a LeonardoService class. The intention is to keep the Leo DAO "dumb" and let the service class contain the requisite retry and error handling logic. Rawls' current pattern for DAOs is inconsistent. For example, HttpGoogleServicesDAO does all of the above, while the WSM DAO is much lighter weight and pushes the error handling and retries to the callsites.
I'd like to centralize on a pattern of "bare, dumb DAO" + "smarter service that knows how to retry, handle errors and propagate a sane exception hierarchy to upper layers". This is a baby step in that direction.
Removes support for deleting MC azure workspaces via the current WorkspaceService deleteWorkspace codepath. Nothing calls this codepath for MC workspaces anymore and so we shouldn't need the added complexity of the isAzureMcWorkspace checks that are peppered throughout that code.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -701,20 +708,23 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
-                                   ctx.userInfo
-          ) recoverWith { case t: Throwable =>
-            logger.error(
-              s"Unexpected failure deleting workspace (while deleting google project) for workspace `${workspaceContext.toWorkspaceName}`",
-              t
-            )
-            Future.failed(t)
-          }
-        } else Future.successful()
+        maybeDeleteGoogleProject(workspaceContext.googleProjectId,
```
aherbst-broad, line 718: This is the only difference from the original PR. We're notifying leo to cleanup resources first, then deleting the gcp project and related sam gcp project resource after, rather than splitting the operations.

yonghaoy (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
LizBaldo: I did not realize that it could take up to 24h to delete a project, in this case making the delete project call before is not that important

LizBaldo (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
yonghaoy: And I see `deletePet` in that method... https://github.com/broadinstitute/rawls/blob/456fbf19de5407006cdeba6f966dfd8d1298ab18/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L837

aherbst-broad:
jenkins retest

PR #2771, opened 2024-03-06
AJ-1664: AvroUpsertMonitor can accept workspaceId instead of namespace/name

davidangb:
Ticket: AJ-1664

Rawls accepts a pub/sub message from Import Service to trigger imports into data tables. This message contains the workspace namespace/name into which to import.

As we are replacing Import Service with cWDS, we'd prefer to send a workspaceId instead of namespace/name, and have Rawls do the right thing internally.

This PR implements that.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
@@ -290,90 +290,89 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
       self ! None
   }
 
-  private def importEntities(message: PubSubMessage) = {
-    val attributes = new AvroUpsertMonitorMessageParser(message).parse
-
-    val importFuture = for {
-      workspace <- dataSource.inTransaction { dataAccess =>
-        dataAccess.workspaceQuery.findByName(attributes.workspace).map {
-          case Some(workspace) => workspace
-          case None =>
-            throw new RawlsException(s"Workspace ${attributes.workspace} not found while importing entities")
+  private def importEntities(message: PubSubMessage) =
+    new AvroUpsertMonitorMessageParser(message, dataSource).parse flatMap { attributes =>
+      val importFuture = for {
+        workspace <- dataSource.inTransaction { dataAccess =>
+          dataAccess.workspaceQuery.findByName(attributes.workspace).map {
+            case Some(workspace) => workspace
+            case None =>
+              throw new RawlsException(s"Workspace ${attributes.workspace} not found while importing entities")
+          }
         }
-      }
-      petUserInfo <- getPetServiceAccountUserInfo(workspace.googleProjectId, attributes.userEmail)
-      importStatus <- importServiceDAO.getImportStatus(attributes.importId, attributes.workspace, petUserInfo)
-      _ <- importStatus match {
-        // Currently, there is only one upsert monitor thread - but if we decide to make more threads, we might
-        // end up with a race condition where multiple threads are attempting the same import / updating the status
-        // of the same import.
-        case Some(status) if status == ImportStatuses.ReadyForUpsert =>
-          publishMessageToUpdateImportStatus(attributes.importId, Option(status), ImportStatuses.Upserting, None)
-          toFutureTry(
-            initUpsert(attributes.upsertFile,
-                       attributes.importId,
-                       message.ackId,
-                       workspace,
-                       attributes.userEmail,
-                       attributes.isUpsert
-            )
-          ) map {
-            case Success(importUpsertResults) =>
-              val failureMessages = stringMessageFromFailures(importUpsertResults.failures, 100)
-              val baseMsg =
-                s"Successfully updated ${importUpsertResults.successes} entities; ${importUpsertResults.failures.size} updates failed."
-              if (importUpsertResults.failures.isEmpty)
-                publishMessageToUpdateImportStatus(attributes.importId,
-                                                   Option(status),
-                                                   ImportStatuses.Done,
-                                                   Option(baseMsg)
-                )
-              else {
-                val msg = baseMsg + s" First 100 failures are: $failureMessages"
+        petUserInfo <- getPetServiceAccountUserInfo(workspace.googleProjectId, attributes.userEmail)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -99,6 +99,12 @@ case class UpdateImportStatus(importId: String,
 
 case class ImportUpsertResults(successes: Int, failures: List[RawlsErrorReport])
 
+case class AvroUpsertAttributes(workspace: WorkspaceName,
+                                userEmail: RawlsUserEmail,
+                                importId: UUID,
+                                upsertFile: String,
+                                isUpsert: Boolean
+)
```
davidangb, line 107: moved up from line 553 to make it more easily visible to other classes.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -550,37 +555,6 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
...
-        message.attributes.getOrElse(workspaceNamespace, attributeNotFoundException(workspaceNamespace)),
-        message.attributes.getOrElse(workspaceName, attributeNotFoundException(workspaceName))
-      ),
-      RawlsUserEmail(message.attributes.getOrElse(userEmail, attributeNotFoundException(userEmail))),
-      UUID.fromString(message.attributes.getOrElse(jobId, attributeNotFoundException(jobId))),
-      message.attributes.getOrElse(upsertFile, attributeNotFoundException(upsertFile)),
-      java.lang.Boolean.parseBoolean(message.attributes.getOrElse(isUpsert, "true"))
-    )
-  }
-
```
davidangb, line 583: The `parseMessage` function has moved to `AvroUpsertMonitorMessageParser`, just makes it a bit more encapsulated and easy to test

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitor.scala
```diff
@@ -284,90 +290,89 @@ class AvroUpsertMonitorActor(val pollInterval: FiniteDuration,
...
-    val attributes = parseMessage(message)
-
-    val importFuture = for {
-      workspace <- dataSource.inTransaction { dataAccess =>
-        dataAccess.workspaceQuery.findByName(attributes.workspace).map {
-          case Some(workspace) => workspace
-          case None =>
-            throw new RawlsException(s"Workspace ${attributes.workspace} not found while importing entities")
+  private def importEntities(message: PubSubMessage) =
+    new AvroUpsertMonitorMessageParser(message, dataSource).parse flatMap { attributes =>
```
davidangb, line 294: message parsing has changed from a synchronous call to a Future (because it may need to make a db query), so I had to add this extra flatMap into the chain.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorMessageParser.scala
```diff
@@ -0,0 +1,60 @@
...
+    workspaceNameFuture map { workspaceName =>
+      AvroUpsertAttributes(
+        workspaceName,
+        RawlsUserEmail(message.attributes.getOrElse(userEmail, attributeNotFoundException(userEmail))),
+        UUID.fromString(message.attributes.getOrElse(jobId, attributeNotFoundException(jobId))),
+        message.attributes.getOrElse(upsertFile, attributeNotFoundException(upsertFile)),
+        java.lang.Boolean.parseBoolean(message.attributes.getOrElse(isUpsert, "true"))
+      )
+    }
+  }
```
davidangb, line 58: this `parse` function is the evolution of the previous `parseMessage`, but adds a potential db query.

jladieu (approved):
Nice!

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorMessageParser.scala
```diff
@@ -0,0 +1,60 @@
...
+    val workspaceName = "workspaceName"
+    val userEmail = "userEmail"
+    val jobId = "jobId"
+    val upsertFile = "upsertFile"
+    val isUpsert = "isUpsert"
+
+    val workspaceId = "workspaceId"
+
+    def attributeNotFoundException(attribute: String): Nothing = throw new Exception(
+      s"unable to parse message - attribute $attribute not found in ${message.attributes}"
```
calypsomatic, line 25: `$attribute` doesn't also need {} around it?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorMessageParser.scala
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    // does this message contain a workspaceId? This indicates a message from cWDS instead of a message
+    // from Import Service.
+    val isWdsMessage = message.attributes.contains(workspaceId)
```
calypsomatic, line 30: Don't we also know this from the `isCWDS` attribute?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorMessageParser.scala
davidangb: heh, nope - if it's just a single variable, no {} needed; if it's an expression (like including the .) then they are needed.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/AvroUpsertMonitorMessageParser.scala
davidangb: yes we could! do you have a preference one way or the other?

calypsomatic: I guess checking the `workspaceId` directly is more reliable that it actually has one to search for, although so far the `isCWDS` attribute is looking pretty useless.

calypsomatic (approved):

PR #2779, opened 2024-03-11
[WOR-1428] Support for policies on GCP workspaces

marctalbott:
Ticket: [WOR-1428](https://broadworkbench.atlassian.net/browse/WOR-1428)
* When importing a snapshot into a workspace, fetch the workspace's auth domain. If the auth domain exists, set up a `group-constraint` policy on the WSM stub workspace containing the auth domain groups.
* When cloning a Rawls workspace, attempt to clone the WSM stub workspace so that policies will propagate to the newly cloned workspace. Ignore 404s when cloning as not all Rawls workspaces have a WSM stub workspace.

---

**PR checklist**

[WOR-1428]: https://broadworkbench.atlassian.net/browse/WOR-1428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -74,21 +81,42 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+
       try {
         // if there's an existing WSM workspace, make sure its platform is compatible
         // with that of the snapshot's dataset.
         val wsmWorkspace = aggregatedWorkspaceService.fetchAggregatedWorkspace(rawlsWorkspace, ctx)
         snapshotValidator.validateWorkspacePlatformCompatibility(wsmWorkspace.getCloudPlatform)
+
+        // if the existing WSM workspace doesn't have a group-constraint policy, but the Rawls workspace
+        // has an auth domain, backfill the group-constraint policy on the existing WSM workspace
+        if (!wsmWorkspace.policies.exists(policy => policy.namespace.equals("terra") && policy.name.equals("group-constraint")) && wsmPolicyInputs.isDefined) {
```
marctalbott: In theory this check isn't strictly necessary. If the workspace has a `group-constraint` policy, then it should be reflected in the auth domain that we fetch earlier. Attempting to update the policy would be a no-op since the groups in the `group-constraint` policy will be the same as the groups in the workspace's auth domain. 

Still, checking here prevents us from an unnecessary WSM call.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1166,6 +1166,27 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+          workspaceId = destWorkspaceContext.workspaceIdAsUUID,
+          displayName = destWorkspaceContext.name,
+          spendProfile = None,
+          billingProjectNamespace = destWorkspaceContext.namespace,
+          ctx
+        )
+      }.recoverWith { case e: ApiException =>
+        if (e.getCode != StatusCodes.NotFound.intValue) {
+          logger.warn(
+            s"Unexpected failure cloning workspace (while cloning in Workspace Manager) [sourceWorkspaceId=${sourceWorkspaceContext.workspaceId}, destWorkspaceId=${destWorkspaceContext.workspaceId}]. Received ${e.getCode}: [${e.getResponseBody}]"
```
cahrens: Nit: is it worth including in the message that it is a Rawls stage workspace?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1166,6 +1166,27 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
         fastPassServiceConstructor(childContext, dataSource)
           .syncFastPassesForUserInWorkspace(destWorkspaceContext)
       )
+
+      _ <- Future {
```
cahrens: Why is the Future needed here?

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -239,10 +242,198 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+
+        // assert that the service DID NOT call WSM's createWorkspace
+        verify(mockWorkspaceManagerDAO, times(0)).createWorkspace(
+          any[UUID],
+          ArgumentMatchers.eq(WorkspaceType.RawlsWorkspace),
+          any[Option[WsmPolicyInputs]],
+          any[RawlsRequestContext]
+        )
+
+        // assert that the workspace's existing policies were not updated
```
cahrens, line 438: I'm trying to understand why we don't want them to be updated.. shouldn't the new policy just merge in?

If the answer is no because it would be an error case, then it seems like we should error if we encounter this, not just silently ignore the new policy.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -2700,6 +2700,102 @@ class WorkspaceServiceSpec
       workspace.bucketName should startWith(s"${services.workspaceServiceConfig.workspaceBucketNamePrefix}-secure")
   }
 
+  it should "clone the WSM stub workspace if it exists" in withTestDataServices { services =>
+    val baseWorkspace = testData.workspace
+    val newWorkspaceName = "cloned_space"
+    val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
```
cahrens, line 2703: How do we know in this test case that the stub workspace exists?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
cahrens: I wonder if we should error if the `group-constraint` policy is not reflected in the auth domain.

marctalbott: We discussed during standup, but putting a summary here as well for posterity.

The idea with policy for the stub Rawls-stage workspaces in WSM is that Rawls will set up `group-constraint` policies at creation time, and then WSM will manage the policies from that point on. 

In the backfill case, we are accounting for the fact that there are existing stub Rawls-stage workspaces that predate Rawls setting any policy on them. 

If an existing Rawls-stage workspace has an auth domain, but it doesn't have a `group-constraint` policy, we interpret that to mean it's an older Rawls-stage workspace and backfill the policy. 

If there is a `group-constraint` policy, then Rawls assumes that WSM knows best and doesn't attempt to compare the policy or otherwise change it.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
marctalbott: We don't want it to be updated because it's not Rawls's place to modify existing policies on a workspace in WSM (rawls-stage or MC).

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
marctalbott: The WSM dao used in this test returns a successful `CloneWorkspaceResult` instead of throwing a 404 which is interpreted as the workspace existing. The verify at the end of the test ensures that the cloneWorkspace request is issued.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: Cloning the workspace will take some time, so wrapping it in a Future prevents it from blocking the thread

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: Yes, added

cahrens (approved):

blakery (approved):

PR #2780, opened 2024-03-11
AJ-1680 add snapshot APIs by workspaceId

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1680
Adds two APIs that are duplicates of existing APIs, except that they take in `workspaceId` as a parameter instead of `workspaceNamespace/workspaceName`.
cWDS will be using these two Rawls APIs, but does not know the workspace namespace or name, only the workspaceId.  Since Rawls knows how to connect the two, it makes more sense to have Rawls take in workspaceId as a parameter and find the workspace as part of the API, rather than cWDS needing to make an additional call to Rawls to find information about the workspace in order to ask Rawls to find that workspace again for the needed APIs.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -56,50 +57,62 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     with WorkspaceSupport
     with LazyLogging {
 
-  def createSnapshot(workspaceName: WorkspaceName,
-                     snapshotIdentifiers: NamedDataRepoSnapshot
+  // Finds a workspace using the workspaceId then calls the createSnapshot method
+  def createSnapshotByWorkspaceId(workspaceId: String,
```
calypsomatic, line 61: Scala apparently can't handle method signatures that are differ by parameter type (and name of the parameter) so I had to give the methods specific names.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -56,50 +57,62 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     with WorkspaceSupport
     with LazyLogging {
```
calypsomatic, line 58: The code added is primarily copy-paste with minor changes to use `workspaceId` instead of `namespace/name`.  Please check carefully to ensure I didn't inadvertently make additional changes.  (Or to suggest a better/cleaner way to factor out shared code paths.)

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -993,6 +1008,44 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
       }
     }
 
+    "create a new snapshot reference when called with workspaceId" in withMinimalTestDatabase { _ =>
```
calypsomatic, line 1011: I'm happy to add more/different tests; most of my new code should use the same code paths so I've only thrown in a minimal happy-path test just like an existing one to verify the new methods work at all.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4065,6 +4065,80 @@ paths:
...
           $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/{workspaceId}/snapshots/v2:
+    parameters:
+      - $ref: '#/components/parameters/workspaceIdPathParam'
+    get:
+      tags:
+        - snapshots_v2
+      summary: List snapshots
+      description: Enumerate references to snapshots in the Terra Data Repo, or return only those that reference a given TDR snapshot id
+      operationId: enumerateDataRepoSnapshot_v2
```
davidangb: nit: each endpoint needs a unique `operationId`, since this is copy/pasted it conflicts with the endpoint that accepts the namespace/name. Same goes for the `createDataRepoSnapshot_v2` endpoint

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -56,50 +57,62 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
-      val wsid = rawlsWorkspace.workspaceIdAsUUID // to avoid UUID parsing multiple times
-      val snapshot =
-        new WrappedSnapshot(dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken))
-      val snapshotValidator = new SnapshotReferenceCreationValidator(rawlsWorkspace, snapshot)
-
-      // prevent snapshots from disallowed platforms
-      snapshotValidator.validateSnapshotPlatform()
+    ).flatMap(rawlsWorkspace => createSnapshot(rawlsWorkspace, snapshotIdentifiers))
+//Given a rawls workspace, creates a snapshot reference in workspace manager
+  def createSnapshot(rawlsWorkspace: Workspace,
```
davidangb: nit: make it private?

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -152,30 +165,37 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
     *
-    * @param workspaceName the workspace owning the snapshot references
+    * @param workspaceId the id of the workspace owning the snapshot references
     * @param offset pagination offset for the list
     * @param limit pagination limit for the list
     * @param referencedSnapshotId the TDR snapshotId for which to return matching references
     * @return the list of snapshot references
     */
-  def enumerateSnapshots(workspaceName: WorkspaceName,
+  def enumerateSnapshots(workspaceId: UUID,
```
davidangb: this needs a call to `getV2WorkspaceContextAndPermissions`; that's where permissions are checked against Sam.

I'm guessing if the user doesn't have the right permissions, some underlying call will fail, but `getV2WorkspaceContextAndPermissions` prevents even making those external calls and ensures the error messages are consistent.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -182,6 +182,19 @@ trait WorkspaceSupport {
...
+      workspaceContext <- getV2WorkspaceContext(workspaceId)
+      _ <- accessCheck(workspaceContext, requiredAction, ignoreLock = false) // throws if user does not have permission
+    } yield workspaceContext
+
+  def getV2WorkspaceContext(workspaceId: String): Future[Workspace] =
+    userEnabledCheck.flatMap { _ =>
+      dataSource.inTransaction { dataAccess =>
+        withV2WorkspaceContext(workspaceId, dataAccess)(DBIO.successful)
+      }
+    }
```
davidangb, line 201: for all these new methods in `WorkspaceSupport`, could you have them accept/pass on the `attributeSpecs` argument? If you can pass that all the way down to `WorkspaceComponent.loadWorkspace`, you can prevent retrieving potentially multiple paragraphs worth of text from the db.

`loadWorkspace` will by default populate the entire workspace, including its description and any workspace attributes such as reference data. Descriptions can be multiple paragraphs of markdown, and reference data can also be relatively big. `attributeSpecs` allows you to control which attributes you want to retrieve; you'll see it gets set to `Some(WorkspaceAttributeSpecs(all = false))` when you want zero extra attributes.

davidangb (commented):
nice job on this! Relatively minor comments inline, but one important one about a permissions check.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
jladieu: Nit: can you rename this method `enumerateSnapshotsByWorkspaceId` for consistency with `enumerateSnapshotsByWorkspaceName`

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/SnapshotApiService.scala
```diff
@@ -39,10 +48,12 @@ trait SnapshotApiService extends UserInfoDirectives {
             parameters("offset".as[Int], "limit".as[Int], "referencedSnapshotId".as[UUID].optional) {
               (offset, limit, referencedSnapshotId) =>
                 complete {
-                  snapshotServiceConstructor(ctx).enumerateSnapshots(WorkspaceName(workspaceNamespace, workspaceName),
-                                                                     offset,
-                                                                     limit,
-                                                                     referencedSnapshotId
+                  snapshotServiceConstructor(ctx).enumerateSnapshotsByWorkspaceName(WorkspaceName(workspaceNamespace,
```
jladieu, line 51: Consider extracting `WorkspaceName` to a local var and sending that down.  I'm hoping it would let you fit the arg on one line so it doesn't wrap and split the statement.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
```diff
@@ -993,6 +1008,44 @@ class SnapshotServiceSpec extends AnyWordSpecLike with Matchers with MockitoSuga
...
+      Await.result(
+        snapshotService.createSnapshotByWorkspaceId(workspace.workspaceIdAsUUID.toString,
+                                                    NamedDataRepoSnapshot(snapRefName, snapRefDescription, snapshotUuid)
+        ),
+        Duration.Inf
+      )
+
+      // assert that the service called WSM's createDataRepoSnapshotReference
+      verify(mockWorkspaceManagerDAO).createDataRepoSnapshotReference(
+        ArgumentMatchers.eq(workspace.workspaceIdAsUUID),
```
jladieu, line 1039: nit: Any chance you can import `eq` the same way `any` is imported, and eliminate the extra-verbose `ArgumentMatchers` prefix?

jladieu (approved):
Awesome job!  A few minor comments, nothing blocking.

core/src/test/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotServiceSpec.scala
calypsomatic: Apparently not, there's some other function `eq` and the compiler gets very angry if it doesn't know for sure which one it is

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
calypsomatic: Ok I hope I did it right, it kicked off some more renaming necessities.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -56,50 +57,62 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
     with WorkspaceSupport
     with LazyLogging {
 
-  def createSnapshot(workspaceName: WorkspaceName,
-                     snapshotIdentifiers: NamedDataRepoSnapshot
+  // Finds a workspace using the workspaceId then calls the createSnapshot method
+  def createSnapshotByWorkspaceId(workspaceId: String,
+                                  snapshotIdentifiers: NamedDataRepoSnapshot
+  ): Future[DataRepoSnapshotResource] =
+    getV2WorkspaceContextAndPermissionsById(workspaceId, SamWorkspaceActions.write).flatMap { rawlsWorkspace =>
```
davidangb: last step on the `attributeSpecs` saga: specify here that you don't care to retrieve any attributes. Note this is parallel to the behavior in `createSnapshotByWorkspaceName`:
```suggestion
    getV2WorkspaceContextAndPermissionsById(workspaceId,
                                            SamWorkspaceActions.write,
                                            Some(WorkspaceAttributeSpecs(all = false))
    ).flatMap { rawlsWorkspace =>
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
davidangb: I think you should take the same approach here you did with the `createSnapshotBy*` APIs, instead of incorporating the check inside the "by id" method:

```scala
def enumerateSnapshotsByWorkspaceName(workspaceName: WorkspaceName, …): Future[SnapshotListResponse] = {
    getV2WorkspaceContextAndPermissions(…).map( …enumerateSnapshots… )
}
def enumerateSnapshotsByWorkspaceId(workspaceId: UUID, …): Future[SnapshotListResponse] = {
    getV2WorkspaceContextAndPermissionsById(…).map( …enumerateSnapshots… )
}

private def def enumerateSnapshots(workspaceId: UUID, …): SnapshotListResponse = { … }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -152,28 +165,37 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+  def enumerateSnapshots(workspaceId: UUID,
                          offset: Int,
                          limit: Int,
                          referencedSnapshotId: Option[UUID] = None
   ): Future[SnapshotListResponse] =
-    getV2WorkspaceContextAndPermissions(workspaceName,
-                                        SamWorkspaceActions.read,
-                                        Some(WorkspaceAttributeSpecs(all = false))
-    ).map { workspaceContext =>
+    getV2WorkspaceContextAndPermissionsById(workspaceId.toString, SamWorkspaceActions.read).map { workspaceContext =>
```
davidangb: this can/should also specify `attributeSpecs = Some(WorkspaceAttributeSpecs(all = false)`

davidangb (commented):
very close, one more restructuring suggestion

calypsomatic:
gateway timeouts
jenkins retest

calypsomatic:
More 503s.
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -152,28 +168,49 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
-                                        Some(WorkspaceAttributeSpecs(all = false))
-    ).map { workspaceContext =>
+    Future.successful {
       referencedSnapshotId match {
-        case None     => retrieveSnapshotReferences(workspaceContext.workspaceIdAsUUID, offset, limit)
-        case Some(id) => findBySnapshotId(workspaceContext.workspaceIdAsUUID, id, offset, limit)
+        case None     => retrieveSnapshotReferences(workspaceId, offset, limit)
+        case Some(id) => findBySnapshotId(workspaceId, id, offset, limit)
       }
     }
```
davidangb: this unnecessarily returns a `Future`. The return value is set to be `Future[SnapshotListResponse]`. Internally to the method, it only calls synchronous operations but then wraps the return value inside `Future.successful(…)`. This could be:
```scala
  private def enumerateSnapshots(workspaceId: UUID,
                                 offset: Int,
                                 limit: Int,
                                 referencedSnapshotId: Option[UUID] = None
  ): SnapshotListResponse =
    referencedSnapshotId match {
      case None     => retrieveSnapshotReferences(workspaceId, offset, limit)
      case Some(id) => findBySnapshotId(workspaceId, id, offset, limit)
    }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -152,28 +168,49 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+                             offset: Int,
+                             limit: Int,
+                             referencedSnapshotId: Option[UUID] = None
+  ): Future[SnapshotListResponse] =
+    getV2WorkspaceContextAndPermissionsById(workspaceId,
+                                            SamWorkspaceActions.read,
+                                            Some(WorkspaceAttributeSpecs(all = false))
+    ).flatMap(workspaceContext =>
+      enumerateSnapshots(workspaceContext.workspaceIdAsUUID, offset, limit, referencedSnapshotId)
+    )
```
davidangb, line 193: in `enumerateSnapshotsByWorkspaceName` and `enumerateSnapshotsById`, there's extraneous `Future`-handling. See my comment below on having `enumerateSnapshots` NOT return a future. In that case, you can use `map` here instead of `flatMap`.

The difference between map and flatMap:
* `map`: chains `Future`s together, using the inner value of the last Future
* `flatMap`: chains `Future`s together and collapses nesting; e.g. `Future[Future[SnapshotListResponse]]` collapses into `Future[SnapshotListResponse]`

in `enumerateSnapshotsByWorkspaceName` and `enumerateSnapshotsById` you are calling two methods:
* `getV2WorkspaceContextAndPermissionsBy*` which returns `Future[Workspace]`
* `enumerateSnapshots ` which _currently_ returns `Future[SnapshotListResponse]`

with the current method signatures in this PR:
* `Future[Workspace] map Future[SnapshotListResponse]` chains the futures and returns `Future[Future[SnapshotListResponse]]` (which is the incorrect return type)
* `Future[Workspace] flatMap Future[SnapshotListResponse]` chains the futures and collapses the nesting, therefore returns `Future[SnapshotListResponse]` (which is correct)

_but_, if you change `enumerateSnapshots` to return `SnapshotListResponse` instead of `Future[SnapshotListResponse]`:
* `Future[Workspace] map SnapshotListResponse` chains the futures and returns `Future[SnapshotListResponse]` (which is correct)
* `Future[Workspace] flatMap SnapshotListResponse` does not compile, because flatMap expects another `Future` on its right-hand side

does this help?

jladieu (approved):

davidangb (commented):
sorry, one last comment on `Future`-handling, which is one of the trickiest things

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
calypsomatic: Ah yes that was an earlier iteration when I didn't have a separate `enumerateSnapshotsById` method and was trying to wrangle correct return values... But do I need to do a similar thing in the calling methods or just propagate the lack of a Future all the way up?

calypsomatic: Nm I see you answer that immediately down below.

davidangb (approved):
nice, thank you!

PR #2790, opened 2024-03-20
[WOR-1567] Add support for specifying properties and cloning semantics

aherbst-broad:
[Ticket](https://broadworkbench.atlassian.net/browse/WOR-1567)

### Context
Rawls must allow the specification of cloning instructions on a snapshot reference as well as properties.

### This PR
* Adds the following fields to the snapshot V2 creation endpoint request:
  * [cloningInstructions](https://github.com/broadinstitute/rawls/pull/2790/files#diff-6e80a735b01ebcbdc253d755a6078cc1af5a585a6bd04b92b6455a9469988dffR4418) -- these are a subset of the possible cloning instructions for a WSM referenced datarepo snapshot, specifically those that are applicable to referenced resources (i.e., not controlled). **This will default to COPY_NOTHING if no value is provided for this field, which is the current behavior**
  * [properties](https://github.com/broadinstitute/rawls/pull/2790/files#diff-6e80a735b01ebcbdc253d755a6078cc1af5a585a6bd04b92b6455a9469988dffR4414), which is a map of KVPs that thread into the snapshot reference resource creation request. 
  * Updates the response format to expose the properties 

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4411,6 +4411,22 @@ components:
           type: string
           description: A description for the snapshot
           default: ""
+        properties:
+          type: object
+          description: A map of key-value pairs to associate with the snapshot. Optional, defaults to an empty map if not specified.
+          default: {}
+        cloningInstructions:
+          type: string
+          default: COPY_NOTHING
+          description: |
+            Cloning semantics for this snapshot. Optional, defaults to COPY_NOTHING if not specified.
+            COPY_NOTHING: Don't clone the snapshot.
+            COPY_REFERENCE: Create a new referenced snapshot that points to same cloud resource as the source snapshot.
+            COPY_LINK_REFERENCE: Create a new referenced snapshot that points to the same cloud resource as the source resource, AND link the source workspace policy to the destination workspace policy; changes in the source will propagate to the destination.
+          enum:
+            - COPY_NOTHING
+            - COPY_REFERENCE
+            - COPY_LINK_REFERENCE
       description: A Data Repo snapshot
     UpdateDataReferenceRequestBody:
       type: object
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4455,6 +4455,8 @@ components:
           type: string
         cloningInstructions:
           type: string
+        properties:
+          type: object
     DataRepoSnapshotAttributes:
       type: object
       required: [ instanceName, snapshot ]
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -187,16 +187,20 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
                                                description: Option[DataReferenceDescriptionField],
                                                instanceName: String,
                                                cloningInstructions: CloningInstructionsEnum,
+                                               properties: Option[Map[String, String]],
                                                ctx: RawlsRequestContext
   ): DataRepoSnapshotResource = {
     val snapshot = new DataRepoSnapshotAttributes().instanceName(instanceName).snapshot(snapshotId.toString)
     val commonFields =
       new ReferenceResourceCommonFields()
         .name(name.value)
-        // Note: that we're ignoring the passed in `cloningInstructions` is a known issue, addressed as part of
-        // https://github.com/broadinstitute/rawls/pull/2081
-        .cloningInstructions(CloningInstructionsEnum.NOTHING)
+        .cloningInstructions(cloningInstructions)
     description.map(d => commonFields.description(d.value))
+
+    val requestProperties = new Properties
+    properties.foreach(p => p.foreach(kv => requestProperties.add(new Property().key(kv._1).value(kv._2))))
+    commonFields.properties(requestProperties)
+
     val request = new CreateDataRepoSnapshotReferenceRequestBody().snapshot(snapshot).metadata(commonFields)
     getReferencedGcpResourceApi(ctx).createDataRepoSnapshotReference(request, workspaceId)
   }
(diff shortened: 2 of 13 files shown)


aherbst-broad:
jenkins retest

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4411,6 +4411,22 @@ components:
           type: string
           description: A description for the snapshot
           default: ""
+        properties:
```
aherbst-broad, line 4414: @davidangb let me know if there any issues with the structure of this API modification. I'm also returning the properties and cloning instructions in the response.

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -140,12 +149,27 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
-          CloningInstructionsEnum.NOTHING,
+          cloningInstructions,
+          snapshotIdentifiers.properties,
           ctx
         )
     }
 
-  def getSnapshot(workspaceName: WorkspaceName, referenceId: String): Future[DataRepoSnapshotResource] = {
+  private def getSnapshotFromDataRepo(snapshotIdentifiers: NamedDataRepoSnapshot) =
+    Try(dataRepoDAO.getSnapshot(snapshotIdentifiers.snapshotId, ctx.userInfo.accessToken)) match {
```
aherbst-broad, line 159: This was 500ing for me in my testing when I mistakenly provided a bogus uuid

nawatts (approved):

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4411,6 +4411,22 @@ components:
...
+          description: A map of key-value pairs to associate with the snapshot. Optional, defaults to an empty map if not specified.
+          default: {}
+        cloningInstructions:
+          type: string
+          default: COPY_NOTHING
+          description: |
+            Cloning semantics for this snapshot. Optional, defaults to COPY_NOTHING if not specified.
+            COPY_NOTHING: Don't clone the snapshot.
+            COPY_REFERENCE: Create a new referenced snapshot that points to same cloud resource as the source snapshot.
+            COPY_LINK_REFERENCE: Create a new referenced snapshot that points to the same cloud resource as the source resource, AND link the source workspace policy to the destination workspace policy; changes in the source will propagate to the destination.
```
davidangb, line 4425: does this infer that `COPY_REFERENCE` does NOT do anything with policy?

I have workspace A.
I link snapshot 1 to workspace A. This links snapshot 1's policies to workspace A.
I clone workspace A to workspace B. After cloning:

if the A:1 link is `COPY_REFERENCE`:
* workspace B contains a reference to snapshot 1
* workspace B does _not_ have snapshot 1's policies

if the A:1 link is `COPY_LINK_REFERENCE `:
* workspace B contains a reference to snapshot 1
* workspace B _does_ have snapshot 1's policies

I'm unclear on why this would be different. When I add the reference from workspace A to snapshot 1, the snapshot policies get copied onto the workspace. At that point, they're the workspace's policies; even if I remove the reference, the policies remain on workspace A.

Wouldn't cloning workspace A copy all of its policies, regardless of where they came from, to workspace B?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -187,16 +187,20 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
...
       new ReferenceResourceCommonFields()
         .name(name.value)
-        // Note: that we're ignoring the passed in `cloningInstructions` is a known issue, addressed as part of
-        // https://github.com/broadinstitute/rawls/pull/2081
-        .cloningInstructions(CloningInstructionsEnum.NOTHING)
+        .cloningInstructions(cloningInstructions)
     description.map(d => commonFields.description(d.value))
+
+    val requestProperties = new Properties
+    properties.foreach(p => p.foreach(kv => requestProperties.add(new Property().key(kv._1).value(kv._2))))
```
marctalbott, line 201: FWIW you could also do this to avoid the `._1` and `._2` with the tuple. Totally up to you on this since what you have here is perfectly clear IMO.
```suggestion
    properties.foreach(p =>
      p.foreach { case (key, value) => requestProperties.add(new Property().key(key).value(value)) }
    )```

**Review by marctalbott — APPROVED** [review]

File: core/src/main/resources/swagger/api-docs.yaml (hunk 1 above)
Comment (davidangb) [reply core/src/main/resources/swagger/api-docs.yaml:4414]: I don't see any models changed for responses here in the OpenApi spec, is that worth updating?

**Review by davidangb — APPROVED** [review]
one question about functional behavior; syntax lgtm.

File: core/src/main/resources/swagger/api-docs.yaml (hunk 3 above)
Comment (aherbst-broad) [reply core/src/main/resources/swagger/api-docs.yaml:4425]: Full disclosure, this is [pulled](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/ReferencedGcpResource/createDataRepoSnapshotReference) from the WSM API docs and is a straight passthrough. 

The distinction here is whether the policies are linked or merged. The COPY_REFERENCE semantics merge the policies on clone, whereas the LINK_REFERENCE _links_ the policies (the TPS [design doc](https://docs.google.com/document/d/1dRtE_nZf8e231DZmk_-9CDbjYTiIN2ZZwI95uF1w2pg/edit#heading=h.1uc5ubqdd419) gets into the details). Here is a useful snippet: 

```
TPS supports merging the policies from one policy object to another without creating a dependency between the source of the merge and the target. We expect this to be used to model the policy effects when workspaces or resources in workspaces are copied (cloned).

The idea is that when a merge is done, the destination PAO must enforce all of the policies that are on the originating PAO. However, later on new policies might be added to the originating PAO. Those should be of no consequence to the destination PAO.
```
In both examples above, the destination workspace should receive snapshot 1's policies. My understanding from the doc and the above snippet is that merging is present to prevent the policy modifications from the source PAO propagating to the cloned one. 

If this is confusing, maybe we limit this enum to solely NOTHING or COPY_REFERENCE?

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: Good call, I think the only schema that requires modification is the `DataRepoSnapshotResource`, added [here](https://github.com/broadinstitute/rawls/pull/2790/commits/354d3ac480b08af95dcb3cb71ae334f69c61a251).

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

core/src/main/resources/swagger/api-docs.yaml
davidangb: got it, thank you, I appreciate the explanation! I understand that `COPY_REFERENCE` is suggested for cloning, so baseline we'll go with that. 

I also feel like `COPY_LINK_REFERENCE` is what we *should* suggest for cloning, but that's a business decision and I will check on that internally with Product; no changes requested for this PR.

aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

PR #2806, opened 2024-03-29
AJ-1697 Add Rawls provider contract test

calypsomatic:
Ticket: [AJ-1697](https://broadworkbench.atlassian.net/browse/AJ-1697)

Sets up Rawls as a provider for contract tests.  Adds `RawlsProviderSpec` and `verify_consumer_pacts` GHA.

---

**PR checklist**

[AJ-1697]: https://broadworkbench.atlassian.net/browse/AJ-1697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+    _ => mock[GenomicsService]
+  val mockSnapshotServiceConstructor: RawlsRequestContext => SnapshotService =
+    _ => mock[SnapshotService]
+  val mockSpendReportingConstructor: RawlsRequestContext => SpendReportingService =
+    _ => mock[SpendReportingService]
+  val mockBillingProjectOrchestratorConstructor: RawlsRequestContext => BillingProjectOrchestrator =
+    _ => mock[BillingProjectOrchestrator]
+  val mockBucketMigrationServiceConstructor: RawlsRequestContext => BucketMigrationService =
+    _ => mock[BucketMigrationService]
+  val mockStatusServiceConstructor: () => StatusService = mock[() => StatusService]
```
ichengchang: It is important to use the singleton pattern here for rendering`StatusService` so that the constructor returns the same service object every time being called. Below is a suggestion on how to do this. You also need to examine all other service constructors to follow the same pattern.
```suggestion
lazy val mockStatusServiceConstructor: () => StatusService = {
    // Create and return a singleton mockStatusService
    lazy val mockStatusService: StatusService = mock[StatusService]
    () => mockStatusService
  }
```

calypsomatic: Does it make a difference whether the test is actually using or relying on these services?  Most of these are being constructed just to have something to initialize the RawlsApi with but not actually used in the test in any way.  Can I get away with only doing this for service constructors under test?

ichengchang:
@calypsomatic I have a few recommendations.

- You can run provider test locally with `wds-rawls` pact without spinning up a local pactbroker. 
- Change all service constructors to return singletons (see comments above).
- First start with a simple endpoint as sanity check before moving on to a scenario that requires more complex setup (query parameters etc). My favorite go to is the status endpoint.

Here's the step-by-step for `/status`endpoint sanity check. 

**Add the `rawlsOK` state**

```
object States {
  val oneSnapshot = "one snapshot in the given workspace"
  val rawlsOK = "Rawls is ok" 
}
```

**Add mocked status check response**

```
import akka.http.scaladsl.model.{StatusCode, StatusCodes}
import org.broadinstitute.dsde.rawls.model.Subsystems.Subsystem
import org.broadinstitute.dsde.rawls.model.{Subsystems, SubsystemStatus}

  val subsystemsStatusMap: Map[Subsystem, SubsystemStatus] =
    Subsystems.CriticalSubsystems.map { subsystem =>
      subsystem -> SubsystemStatus(ok = true, None)
    }.toMap ++
      Subsystems.GoogleSubsystems.map { subsystem =>
        subsystem -> SubsystemStatus(ok = true, None)
      }.toMap

  private def mockSubsystemsStatus(healthy: Boolean): OngoingStubbing[Future[(StatusCode, StatusCheckResponse)]] =
    when {
      mockStatusServiceConstructor().getStatus
    } thenReturn {
      Future.successful(
        StatusCodes.OK,
        StatusCheckResponse(
          ok = healthy,
          systems = subsystemsStatusMap
        )
      )
    }
```

**Add state handler for rawlsOK**

```
  private val providerStatesHandler: StateManagementFunction = StateManagementFunction {
    **case ProviderState(States.rawlsOK, _) =>
      mockSubsystemsStatus(true)**
    case ProviderState(States.oneSnapshot, _) =>
      mockEnumerateSnapshots(
      mockSnapshotServiceConstructor(RawlsRequestContext(userInfo = mockUserInfo, otelContext = mockOtelContext)),
      mockedEnumerateSnapshotsResponse
      )
    case _ =>
      loggerIO.debug("State not found")
  }
```

**Copy the following pact and save the `wds-rawls.json` to `pact4s/src/test/resources` directory**

**wds-rawls.json**
```
{
  "consumer": {
    "name": "wds"
  },
  "interactions": [
    {
      "description": "a status request",
      "providerStates": [
        {
          "name": "Rawls is ok"
        }
      ],
      "request": {
        "method": "GET",
        "path": "/status"
      },
      "response": {
        "body": {
          "ok": true,
          "systems": {

          }
        },
        "status": 200
      }
    },
    {
      "description": "a request for the workspace's snapshots",
      "providerStates": [
        {
          "name": "one snapshot in the given workspace"
        }
      ],
      "request": {
        "matchingRules": {
          "query": {
            "limit": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "regex",
                  "regex": "10"
                }
              ]
            },
            "offset": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "regex",
                  "regex": "0"
                }
              ]
            }
          }
        },
        "method": "GET",
        "path": "/api/workspaces/facade00-0000-4000-a000-000000000000/snapshots/v2",
        "query": {
          "limit": [
            "10"
          ],
          "offset": [
            "0"
          ]
        }
      },
      "response": {
        "body": {
          "gcpDataRepoSnapshots": [
            {
              "attributes": {
                "instanceName": "string",
                "snapshot": "string"
              },
              "metadata": {
                "properties": [

                ],
                "resourceId": "5ca1ab1e-0000-4000-a000-000000000000",
                "workspaceId": "facade00-0000-4000-a000-000000000000"
              }
            }
          ]
        },
        "generators": {
          "body": {
            "$.gcpDataRepoSnapshots[0].attributes.instanceName": {
              "size": 20,
              "type": "RandomString"
            },
            "$.gcpDataRepoSnapshots[0].attributes.snapshot": {
              "size": 20,
              "type": "RandomString"
            }
          }
        },
        "headers": {
          "Content-Type": "application/json; charset=UTF-8"
        },
        "matchingRules": {
          "body": {
            "$.gcpDataRepoSnapshots[0].attributes.instanceName": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "type"
                }
              ]
            },
            "$.gcpDataRepoSnapshots[0].attributes.snapshot": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "type"
                }
              ]
            },
            "$.gcpDataRepoSnapshots[0].metadata.resourceId": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "regex",
                  "regex": "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
                }
              ]
            },
            "$.gcpDataRepoSnapshots[0].metadata.workspaceId": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "regex",
                  "regex": "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
                }
              ]
            }
          },
          "header": {
            "Content-Type": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "regex",
                  "regex": "application/json(;\\s?charset=[\\w\\-]+)?"
                }
              ]
            }
          }
        },
        "status": 200
      }
    }
  ],
  "metadata": {
    "pact-jvm": {
      "version": "4.6.7"
    },
    "pactSpecification": {
      "version": "3.0.0"
    }
  },
  "provider": {
    "name": "rawls"
  }
}
```

**To test locally. Update the `ProviderInfoBuilder` to match following.**

```
  val provider: ProviderInfoBuilder =
    ProviderInfoBuilder(
      name = "rawls",
      // pactSource = PactSource
      //  .PactBrokerWithSelectors(pactBrokerUrl)
      //  .withAuth(BasicAuth(pactBrokerUser, pactBrokerPass))
      //  .withPendingPactsEnabled(ProviderTags(providerBranch)) // TODO providerBranch or providerVer?
      //  .withConsumerVersionSelectors(consumerVersionSelectors)
      pactSource = PactSource.FileSource(
        Map("wds" -> new File("src/test/resources/wds-rawls.json"))
      )
    )
      .withStateManagementFunction(
        providerStatesHandler
//          .withBeforeEach(() => resetMocks())
      )
      .withHost("localhost")
      .withPort(8080)
```

**Run provider test locally. 
Position your working directory to be the root directory of rawls.
Run following docker command.**

Note: You may need to disconnect from vpn to run the docker successfully or try to use split vs non-split vpn.

```
docker run --rm -w /working/rawls \
   -v "$PWD":/working/rawls \
   -e PACT_PUBLISH_RESULTS=false \
   sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
   sbt -J-Xmx2g -J-XX:+UseG1GC \
   sbt "set scalafmtOnCompile := false" "project pact4s" "testOnly *RawlsProviderSpec"
```

ichengchang:
@calypsomatic Once you confirm the /status is working. You can go ahead and try figure out why the enumerate snapshots by id is not working properly. Don't forget to change all service constructors to return singletons.

ichengchang:
@calypsomatic Questions about the level of mocking.

There's no one size fits all but I can give an example to illustrate some of the considerations to think about. Again my go to example is /status and I want to caution this applies to rawls repo, may not others.

You see the /status is being mocked using `mockStatusServiceConstructor().getStatus` and I am just passing the `healthy` parameter that represents the state (Rawls OK or not OK). At this level, I use mock[StatusService] and that's sufficient.

If you look under the hood one-level deep, and that is at the microservice level, on how the route is specified, you see that in `StatusApiService.scala`.  You discover that down the wire there is something extra in the `StatusService` implementation layer which refers to `(healthMonitor ? GetCurrentStatus)`. So now you see a dependency of the Actor system healthMonitor: ActorRef. Question is, should I mock `GetCurrentStatus` message being send to healthMonitor? Unfortunately that's beyond capability of mockito. But let's say if we have a more powerful mocking framework, should we mock `GetCurrentStatus`. The answer should still be no. Why? We run the risk of over mocking. Today we are using akka as our concurrency framework, down-the-wire it's message passing architecture. Tomorrow what about we swap out this architecture and replace by something else.If we mock `(healthMonitor ? GetCurrentStatus)` and its underlying `getCurrentStatus` method, everything that relies on the mock will break one day if akka goes away. In fact, if we want to test akka, it's best to do so with the test kit that comes with akka. So we are fine with mocking `StatusService` at the level we did and not go deeper. This is just one way to consider what level of mocking is too brittle.

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
ichengchang: You need to use singleton object. Think this way, when you spin up the service on the mock server using the constructor it has all the global states initiated with this copy of object.

When you use mockito to mock the behavior of endpoint, you want to mock the method referenced by the same object. When calling the constructor the second time without singleton, it will generate a new mock object and so you end up not mocking the intended copy deployed to the mock server.

ichengchang: So to answer the question, you can get away with singleton for those services not under test. In my opinion though, if someone else in the future picks up some new service to be tested, they won't be puzzled.

ichengchang: Another thought is how do you guarantee that a service is a standalone service and other services don't depend on, and that happen to be the service with a non-singleton constructor.

ichengchang:
I saw some chatters in #dsp-contract-testing this morning. Again this is related to the level of mocking you need to think through (see above notes on level of mocking). In your code, you are deriving the http routes from `RawlsApiServiceImpl` and looking at its implementation you discover it has a mixin `StandardUserInfoDirectives`. Within `StandardUserinfoDirectives` you see the following block of code. In particular `headerValueByName`, `optionalHeaderValueByName` are part of the akka http routing ecosystem.  The `SnapshotApiService` depends on the function `requireUserInfo`.

```
def requireUserInfo(otelContext: Option[Context]): Directive1[UserInfo] = {
    (
      headerValueByName("OIDC_access_token") &
        headerValueByName("OIDC_CLAIM_user_id") &
        headerValueByName("OIDC_CLAIM_expires_in") &
        headerValueByName("OIDC_CLAIM_email") &
        optionalHeaderValueByName("OAUTH2_CLAIM_idp_access_token")
    ) 
```

We should avoid mocking the internal functions `headerValueByName`, `optionalHeaderValueByName` directly. Instead we should mock the high level function `requireUserInfo` to return a useful Directive1[UserInfo]. You already have one in `org.broadinstitute.dsde.rawls.openam.MockUserInfoDirectives` to bypass akka http routing.

In your `RawlsProviderSpec`, just add `MockUserInfoDirectives` as mixin 

```
 val rawlsApiService = new RawlsApiServiceImpl(
    mockMultiCloudWorkspaceServiceConstructor,
    mockWorkspaceServiceConstructor,
    mockEntityServiceConstructor,
    mockUserServiceConstructor,
    mockGenomicsServiceConstructor,
    mockSnapshotServiceConstructor,
    mockSpendReportingConstructor,
    mockBillingProjectOrchestratorConstructor,
    mockBucketMigrationServiceConstructor,
    mockStatusServiceConstructor,
    mockExecutionServiceCluster,
    mockAppVersion,
    FiniteDuration(1, TimeUnit.MINUTES),
    1000000L,
    "test",
    mockSamDAO,
    mockOpenIDConnectConfiguration
  ) with MockUserInfoDirectives
```

and also please add `override` to qualify the function `requireUserInfo` in `MockUserInfoDirectives`

```
trait MockUserInfoDirectives extends UserInfoDirectives {
  protected def userInfo = UserInfo(RawlsUserEmail("owner-access"),
                                    OAuth2BearerToken("token"),
                                    123,
                                    RawlsUserSubjectId("123456789876543212345")
  )

  override def requireUserInfo(otelContext: Option[Context]): Directive1[UserInfo] = {
    println("In MockUserInfoDirectives requireUserInfo")
    provide(userInfo)
  }
}
```

This should get rid of the 405 and unblock you and enable you to run provider test and debug other potential issues.

.github/workflows/consumer_contract_tests.yml
```diff
@@ -109,7 +109,7 @@ jobs:
                           -v jar-cache:/root/.ivy2 \
                           -w /working \
                           sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
-                          sbt "project pact4s" clean test
+                          sbt "project pact4s" clean "testOnly org.broadinstitute.dsde.rawls.consumer.*"
```
calypsomatic: Do I need to update line 111 too since the rest of rawls is updating its base image?  I got "could not retrieve sbt 1.9.9" when trying this locally but it seems like the GHAs are working.

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
```diff
@@ -0,0 +1,311 @@
...
+import scala.collection.immutable.Map
+
+object States {
+  val oneSnapshot = "one snapshot in the given workspace"
+  val rawlsOK = "Rawls is ok"
+}
+
+class RawlsProviderSpec extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
+
+  private val WORKSPACE_UUID = "facade00-0000-4000-a000-000000000000"
```
calypsomatic, line 75: TODO: Can I pass these as state params or something?

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
```diff
@@ -0,0 +1,321 @@
...
+  lazy val consumerVer: Option[String] = sys.env.get("CONSUMER_VERSION")
+
+  var consumerVersionSelectors: ConsumerVersionSelectors = ConsumerVersionSelectors()
+  // consumerVersionSelectors = consumerVersionSelectors.mainBranch
+  // The following match condition basically says
+  // 1. If verification is triggered by consumer pact change, verify only the changed pact.
+  // 2. For normal Sam PR, verify all consumer pacts in Pact Broker labelled with a deployed environment (alpha, dev, prod, staging).
+  consumerBranch match {
+    case Some(s) if !s.isBlank => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
+    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
```
ichengchang, line 251: @calypsomatic At this point, the new `wds-rawls` pact still belongs to the `aj-1697-rawls-contract` branch of the `terra-workspace-data-service` repository and has not yet been merged or deployed. As a result, `consumerVersionSelectors.deployedOrReleased.mainBranch` does not yield any results. Try directly referencing the `wds` branch `aj-1697-rawls-contract` for the provider test to locate the pact from the pact broker. This may resolve the error you saw during the provider test.
```suggestion
    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch.branch("aj-1697-rawls-contract", Some("wds"))
```

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
```diff
@@ -0,0 +1,322 @@
...
+      name = "rawls",
+      pactSource = PactSource
+        .PactBrokerWithSelectors(pactBrokerUrl)
+        .withAuth(BasicAuth(pactBrokerUser, pactBrokerPass))
+        .withPendingPactsEnabled(ProviderTags(providerBranch))
+        .withConsumerVersionSelectors(consumerVersionSelectors)
+    )
+      .withStateManagementFunction(
+        providerStatesHandler
+//          .withBeforeEach(() => resetMocks())
```
calypsomatic: `resetMocks` was copied from the LeoProvider, I think, and I'm not sure if I need it - I did have some errors from it at some point and it seems to work without it.  Thoughts?

.github/workflows/consumer_contract_tests.yml
ichengchang: I notice the **publish-contracts** job in `.github/workflows/consumer_contract_tests.yml#L138-L144` is missing `release-tag` input parameter. Here's the suggested change.

```
inputs: '{
            "run-name": "${{ env.PUBLISH_CONTRACTS_RUN_NAME }}",
            "pact-b64": "${{ needs.rawls-consumer-contract-tests.outputs.pact-b64 }}",
            "repo-owner": "${{ github.repository_owner }}",
            "repo-name": "${{ github.event.repository.name }}",
            "repo-branch": "${{ needs.init-github-context.outputs.repo-branch }}",
            "release-tag": "${{ needs.init-github-context.outputs.repo-version }}"
          }'
```

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
jladieu: I'm fine to leave it out until something happens that makes us add it back in!

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
jladieu: [ref](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/test/java/bio/terra/workspace/pact/WdsContractVerificationTest.java#L200-L212) This is how I did provider state stuff in WSM; maybe something similar is usable here?

calypsomatic: I've been mussing with it a bit, it's different for scala and I'm not confident I'll be able to get it to work.

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,369 @@
...
+                          -e POSTGRES_READ_URL=localhost:5432 \
+                          -e POSTGRES_USERNAME=rawls-test \
+                          -e POSTGRES_PASSWORD=rawls-test \
+                          sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
+                          sbt -J-Xmx2g -J-XX:+UseG1GC \
+                          "project pact4s" clean coverage "testOnly org.broadinstitute.dsde.rawls.provider.*"  coverageReport
+
+  can-i-deploy: # The can-i-deploy job will run as a result of a Rawls PR. It reports the pact verification statuses on all deployed environments.
+    runs-on: ubuntu-latest
+    if: ${{ inputs.pb-event-type == '' }}
```
nawatts, line 337: How do the various `if` conditions for jobs interact?

`can-i-deploy` needs `verify-consumer-pact`, which needs `regulated-tag-job`, which only runs if `is-bump` is "no"?

Curious in what situation would we run `verify-consumer-pact` but not `can-i-deploy`?

jladieu (approved):

.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,27 @@
+# This action must be done after the checkout action
+name: 'bump-skip'
+description: 'Set skip-out when we are doing a version bump'
```
nawatts: Is this an old name copied from somewhere? I don't see any other reference to `skip-out` in this PR.

```suggestion
description: 'Set is-bump when we are doing a version bump'
```

nawatts (approved):

ichengchang:
@calypsomatic Great job in adding `release-tag` to `consumer_contract_tests.yml`. The `can-i-deploy` is now working for consumer workflow. However, the `can-i-deploy` in `verify_consumer_pacts.yml` still fails.

This is where we need to think about the formal tagging procedure. First, we need to pivot to `semver` instead of Git hash for both `consumer_contract_tests.yml` and `verify_consumer_pacts.yml`. You can take a look at how [terra-billing-profile-manager](https://github.com/DataBiosphere/terra-billing-profile-manager/) manage version tag for these workflows.

.github/workflows/verify_consumer_pacts.yml
ichengchang: You can refer to [terra-billing-profile-manager](https://github.com/DataBiosphere/terra-billing-profile-manager/) to see how the formal tagging procedure works.

.github/actions/bump-skip/action.yml
ichengchang: Yes, likely a copy paste thing, should be updated to `is-bump`.

calypsomatic:
> @calypsomatic Great job in adding `release-tag` to `consumer_contract_tests.yml`. The `can-i-deploy` is now working for consumer workflow. However, the `can-i-deploy` in `verify_consumer_pacts.yml` still fails.
> 
> This is where we need to think about the formal tagging procedure. First, we need to pivot to `semver` instead of Git hash for both `consumer_contract_tests.yml` and `verify_consumer_pacts.yml`. You can take a look at how [terra-billing-profile-manager](https://github.com/DataBiosphere/terra-billing-profile-manager/) manage version tag for these workflows.

@ichengchang It looks to me like when Rawls bumps the version, it's [doing](https://github.com/broadinstitute/rawls/blob/60225d5739417f4ccd6442bc2353fe7b01081940/.github/workflows/verify_consumer_pacts.yml#L105) the same thing that [bpm](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/46483be2d13dc331f16f52ddabadf02f85a804b5/.github/workflows/tag.yml#L55) does.  The only difference is that BPM uses the `new-tag` output from the `tag` workflow, instead of plain `tag`.  As far as I can tell, this isn't a different type of versioning, just whether the tag is guaranteed to be an updated one.  I remain very confused.

.github/workflows/verify_consumer_pacts.yml
nawatts: I can see from the BPM Actions that this workflow is occasionally "Run manually by BroadBot". I'm guessing that's originating from Pact Broker.

> Curious in what situation would we run verify-consumer-pact but not can-i-deploy?

I still don't understand the distinction here. I thought the answer to "can I deploy?" is based on verifying consumer Pacts?

ichengchang:
> > @calypsomatic Great job in adding `release-tag` to `consumer_contract_tests.yml`. The `can-i-deploy` is now working for consumer workflow. However, the `can-i-deploy` in `verify_consumer_pacts.yml` still fails.
> > This is where we need to think about the formal tagging procedure. First, we need to pivot to `semver` instead of Git hash for both `consumer_contract_tests.yml` and `verify_consumer_pacts.yml`. You can take a look at how [terra-billing-profile-manager](https://github.com/DataBiosphere/terra-billing-profile-manager/) manage version tag for these workflows.
> 
> @ichengchang It looks to me like when Rawls bumps the version, it's [doing](https://github.com/broadinstitute/rawls/blob/60225d5739417f4ccd6442bc2353fe7b01081940/.github/workflows/verify_consumer_pacts.yml#L105) the same thing that [bpm](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/46483be2d13dc331f16f52ddabadf02f85a804b5/.github/workflows/tag.yml#L55) does. The only difference is that BPM uses the `new-tag` output from the `tag` workflow, instead of plain `tag`. As far as I can tell, this isn't a different type of versioning, just whether the tag is guaranteed to be an updated one. I remain very confused.

There's a subtle difference between `new-tag` and `tag`. The bottom line is that the `semver` tag should only be dispatched ONCE during a PR merge. The contract tests run in separate workflows, they also require the `semver` tag to publish pacts and verification results using the same bumper. There is a distinction, however, we don't want the bumper to dispatch the tag to the repository so we use the `dry run` flag. When dry run is used, the `new-tag` contains the next version tag to be used.

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,369 @@
...
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+          
+  regulated-tag-job:
+    runs-on: ubuntu-latest
+    needs: [ bump-check ]
+    if: ${{ needs.bump-check.outputs.is-bump == 'no' }}
+    outputs:
+      tag: '${{ steps.tag.outputs.tag }}'
```
ichengchang:
```suggestion
      tag: '${{ steps.tag.outputs.tag }}'
      new-tag: '${{ steps.tag.outputs.new-tag }}'
```

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,369 @@
...
+        id: tag
+        env:
+          DEFAULT_BUMP: patch
+          GITHUB_TOKEN: '${{ secrets.BROADBOT_TOKEN }}'
+          DRY_RUN: true
+          RELEASE_BRANCHES: develop
+          WITH_V: true
+      - name: Echo generated tag to console
+        run: |
+          echo "Newly created version tag: '${{ steps.tag.outputs.tag }}'"
```
ichengchang:
```suggestion
          echo "Newly created version tag: '${{ steps.tag.outputs.new-tag }}'"
```

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,369 @@
...
+    runs-on: ubuntu-latest
+    needs: [ bump-check ]
+    if: ${{ needs.bump-check.outputs.is-bump == 'no' }}
+    outputs:
+      tag: '${{ steps.tag.outputs.tag }}'
+    steps:
+      - name: Checkout current code
+        uses: actions/checkout@v4
+        with:
+          ref: '${{ inputs.ref }}'
```
ichengchang:
```suggestion
          ref: ${{ github.head_ref || '' }}
```

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,369 @@
...
+          echo "CONSUMER_NAME=${{ inputs.consumer-name }}" >> $GITHUB_ENV
+          echo "CONSUMER_VERSION=${{ inputs.consumer-version-number }}" >> $GITHUB_ENV
+
+      - name: Set PROVIDER_VERSION envvar
+        run: |
+          # The PROVIDER_VERSION envvar is used to identify the provider version
+          # for publishing the results of provider verification.
+          if [[ -z "${{ inputs.pb-event-type }}" ]]; then
+            echo "PROVIDER_BRANCH=${{ env.CURRENT_BRANCH }}" >> $GITHUB_ENV
+            echo "PROVIDER_VERSION=${{ needs.regulated-tag-job.outputs.tag }}" >> $GITHUB_ENV
```
ichengchang:
```suggestion
            echo "PROVIDER_VERSION=${{ needs.regulated-tag-job.outputs.new-tag }}" >> $GITHUB_ENV
```

ichengchang:
There is a type in https://github.com/broadinstitute/rawls/blob/develop/.github/workflows/tag.yml#L66

Please change it to:

echo "Newly created version tag: '${{ steps.tag.outputs.new_tag }}'"

pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala
```diff
@@ -0,0 +1,322 @@
...
+  lazy val consumerVer: Option[String] = sys.env.get("CONSUMER_VERSION")
+
+  var consumerVersionSelectors: ConsumerVersionSelectors = ConsumerVersionSelectors()
+  // consumerVersionSelectors = consumerVersionSelectors.mainBranch
+  // The following match condition basically says
+  // 1. If verification is triggered by consumer pact change, verify only the changed pact.
+  // 2. For normal Rawls PR, verify all consumer pacts in Pact Broker labelled with a deployed environment (alpha, dev, prod, staging).
+  consumerBranch match {
+    case Some(s) if !s.isBlank => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
+    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
```
ichengchang, line 251: Could you add back the branch selector for the time being because the wds pact still belongs to `aj-1697-rawls-contract`? We can remove that after we got a success verification and when wds pact has merged to main.
```suggestion
    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch.branch("aj-1697-rawls-contract", Some("wds"))
```

ichengchang:
@calypsomatic I think we are very close to wrapping this up. If you could add versioning to the `consumer_contract_tests.yml` similar to `terra-billing-profile-manager` that'll be great!

.github/workflows/consumer_contract_tests.yml
```diff
@@ -50,6 +50,31 @@ env:
...
+        # The 'ref' parameter ensures that the consumer version is postfixed with the HEAD commit of the PR branch,
+        # facilitating cross-referencing of a pact between Pact Broker and GitHub.
+        ref: ${{ github.head_ref || '' }}
+        # The 'dry-run' parameter prevents the new tag from being dispatched.
+        dry-run: true
+        release-branches: develop
+      secrets: inherit
+      
   init-github-context:
     runs-on: ubuntu-latest
```
ichengchang, line 79:
```suggestion
    runs-on: ubuntu-latest
    needs: [ bump-check ]
    if: ${{ needs.bump-check.outputs.is-bump == 'no' }}
```

.github/workflows/consumer_contract_tests.yml
ichengchang: Please update 
```
  rawls-consumer-contract-tests:
    runs-on: ubuntu-latest
    needs: [ init-github-context ]
```

to 
```
  rawls-consumer-contract-tests:
    runs-on: ubuntu-latest
    needs: [ bump-check, init-github-context ]
    if: ${{ needs.bump-check.outputs.is-bump == 'no' }}
```

ichengchang: Also, please update
```
  publish-contracts:
    runs-on: ubuntu-latest
    if: ${{ needs.init-github-context.outputs.fork == 'false' }}
    needs:
      - init-github-context
      - rawls-consumer-contract-tests
```

to
```
  publish-contracts:
    runs-on: ubuntu-latest
    if: ${{ (needs.init-github-context.outputs.fork == 'false') && (needs.bump-check.outputs.is-bump == 'no') }}
    needs:
      - bump-check
      - regulated-tag-job
      - init-github-context
      - rawls-consumer-contract-tests
```

.github/workflows/consumer_contract_tests.yml
```diff
@@ -140,9 +167,9 @@ jobs:
             "pact-b64": "${{ needs.rawls-consumer-contract-tests.outputs.pact-b64 }}",
             "repo-owner": "${{ github.repository_owner }}",
             "repo-name": "${{ github.event.repository.name }}",
-            "repo-branch": "${{ needs.init-github-context.outputs.repo-branch }}"
+            "repo-branch": "${{ needs.init-github-context.outputs.repo-branch }}",
+            "release-tag": "${{ needs.init-github-context.outputs.repo-version }}"
```
ichengchang:
```suggestion
            "release-tag": "${{ needs.regulated-tag-job.outputs.new-tag }}"
```

.github/workflows/consumer_contract_tests.yml
```diff
@@ -140,9 +167,9 @@ jobs:
...
           }'
-
   can-i-deploy:
     runs-on: ubuntu-latest
     if: ${{ needs.init-github-context.outputs.fork == 'false' }}
```
ichengchang:
```suggestion
  can-i-deploy:
    runs-on: ubuntu-latest
    needs: [ bump-check, init-github-context, publish-contracts ]
    if: ${{ (needs.init-github-context.outputs.fork == 'false') && (needs.bump-check.outputs.is-bump == 'no') }}
```

.github/workflows/consumer_contract_tests.yml
ichengchang: In `can-i-deploy`, update

```
"version": "${{ needs.init-github-context.outputs.repo-version }}"
```
to
```
"version": "${{ needs.regulated-tag-job.outputs.new-tag }}"
```

.github/workflows/consumer_contract_tests.yml
```diff
@@ -140,15 +170,13 @@ jobs:
...
+            "release-tag": "${{ needs.regulated-tag-job.outputs.new-tag }}"
           }'
-
   can-i-deploy:
     runs-on: ubuntu-latest
-    if: ${{ needs.init-github-context.outputs.fork == 'false' }}
-    needs:
-      - init-github-context
-      - publish-contracts
+    needs: [ bump-check, init-github-context, publish-contracts ]
```
ichengchang:
```suggestion
    needs: [ bump-check, regulated-tag-job, init-github-context, publish-contracts ]
```

.github/workflows/consumer_contract_tests.yml
```diff
@@ -96,20 +123,21 @@ jobs:
...
+      - uses: actions/checkout@v4
       - name: Run consumer tests
         run: |
           docker run --rm -v $PWD:/working \
                           -v jar-cache:/root/.ivy \
                           -v jar-cache:/root/.ivy2 \
                           -w /working \
                           sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
-                          sbt "project pact4s" clean test
+                          sbt "project pact4s" clean "testOnly org.broadinstitute.dsde.rawls.consumer.*"
```
ichengchang:
```suggestion
                          sbt -J-Xmx2g -J-XX:+UseG1GC "project pact4s" clean "testOnly org.broadinstitute.dsde.rawls.consumer.*"
```

ichengchang (approved):
Great work!

PR #2825, opened 2024-04-10
Rawls attempts to fetch "rawls" workspaces from WSM in addition to "MC"

ashanhol:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1750

Currently, Rawls only returns policies for Azure workspaces, and an empty list for GCP ("rawls") workspaces. This is because GCP workspaces usually don't have a corresponding entry in WSM. However, GCP workspaces with linked snapshots do, and we need to try and retrieve that to display policies in the UI. 

This PR attempts to make the WSM fetch call for all workspaces, and returns a default entry if it doesn't exist. It's a similar method to what we're doing in the [snapshot code here](https://github.com/broadinstitute/rawls/pull/2570/files#diff-3fb111d317a44913c11c6b540e6a2d05fdf5715c28ae0a0bb87921bb8e1057eeR80).

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -316,13 +316,18 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
               try
                 wsmService.fetchAggregatedWorkspace(workspaceContext, ctx)
               catch {
-                case _: Exception =>
-                  // return workspace with no WSM information
-                  AggregatedWorkspace(workspaceContext,
-                                      Some(workspaceContext.googleProjectId),
-                                      azureCloudContext = None,
-                                      policies = List.empty
-                  )
+                case e: AggregateWorkspaceNotFoundException =>
+                  // return workspace with no WSM information for gcp workspace
+                  if (workspaceContext.workspaceType == WorkspaceType.RawlsWorkspace) {
+                    AggregatedWorkspace(workspaceContext,
+                                        Some(workspaceContext.googleProjectId),
+                                        azureCloudContext = None,
+                                        policies = List.empty
+                    )
+                  } else {
+                    // bubble up an MC workspace exception
+                    throw e
+                  }
               }
 
             // maximum access level is required to calculate canCompute and canShare. Therefore, if any of
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -311,7 +311,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       getV2WorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Option(attrSpecs)) flatMap {
         workspaceContext =>
           dataSource.inTransaction { dataAccess =>
-            val wsmContext = wsmService.optimizedFetchAggregatedWorkspace(workspaceContext, ctx)
+            // some GCP workspaces, like those with linked snapshots, have a stub WSM workspace
+            val wsmContext =
+              try
+                wsmService.fetchAggregatedWorkspace(workspaceContext, ctx)
+              catch {
+                case _: Exception =>
+                  // return workspace with no WSM information
+                  AggregatedWorkspace(workspaceContext,
+                                      Some(workspaceContext.googleProjectId),
+                                      azureCloudContext = None,
+                                      policies = List.empty
+                  )
+              }
 
             // maximum access level is required to calculate canCompute and canShare. Therefore, if any of
             // accessLevel, canCompute, canShare is specified, we have to get it.
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
@@ -3042,6 +3042,27 @@ class WorkspaceServiceSpec
     )
   }
 
+  private def createGcpWorkspaceStub(services: TestApiService,
+                                     workspaceName: String,
+                                     policies: List[WsmPolicyInput] = List()
+  ): Workspace = {
+    val workspaceRequest = WorkspaceRequest(
+      testData.testProject1Name.value,
+      workspaceName,
+      Map.empty
+    )
+    when(services.workspaceManagerDAO.getWorkspace(any[UUID], any[RawlsRequestContext])).thenReturn(
+      new WorkspaceDescription()
+        .stage(WorkspaceStageModel.RAWLS_WORKSPACE)
+        .policies(policies.asJava)
+    )
+
+    Await.result(
+      services.mcWorkspaceService.createMultiCloudWorkspace(workspaceRequest, new ProfileModel().id(UUID.randomUUID())),
+      Duration.Inf
+    )
+  }
+
   it should "get the details of an Azure workspace" in withTestDataServices { services =>
     val managedAppCoordinates = AzureManagedAppCoordinates(UUID.randomUUID(), UUID.randomUUID(), "fake_mrg_id")
     val workspace = createAzureWorkspace(services, managedAppCoordinates)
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -311,7 +311,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       getV2WorkspaceContextAndPermissions(workspaceName, SamWorkspaceActions.read, Option(attrSpecs)) flatMap {
         workspaceContext =>
           dataSource.inTransaction { dataAccess =>
-            val wsmContext = wsmService.optimizedFetchAggregatedWorkspace(workspaceContext, ctx)
```
ashanhol, line 314: This is the only place the optimized function was used

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -311,7 +311,19 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+            // some GCP workspaces, like those with linked snapshots, have a stub WSM workspace
```
ashanhol, line 314: Perhaps this is oversimplified now... I was trying to get rid of the "optimized" function and repurpose that logic here so that it'd check `fetchAggregatedWorkspace` for the rawls workspace too, but `fetchAggregatedWorkspace` could also fail on an MC workspace. Is there a way to modify the `catch` to be `if AggregateWorkspaceNotFoundException && WorkspaceType.RawlsWorkspace`?

nawatts: > Is there a way to modify the catch to be if AggregateWorkspaceNotFoundException && WorkspaceType.RawlsWorkspace?

It looks like there's something similar in SnapshotService that might be helpful:

https://github.com/broadinstitute/rawls/blob/19c84937b0d48342577f51b4f229b4be12a02e48/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala#L122-L143

ashanhol: ya know, I linked that exact line of code in the PR description and yet I missed how it was checking. thank you

nawatts: Oops, sorry, I missed that in the description.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -311,7 +311,24 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+            val wsmContext =
+              try
+                wsmService.fetchAggregatedWorkspace(workspaceContext, ctx)
```
yuliadub, line 317: do we need a test for this specific call or does that exist somewhere else already? (I'd assume so but wanted to check)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -311,7 +311,24 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+                  // return workspace with no WSM information for gcp workspace
+                  if (workspaceContext.workspaceType == WorkspaceType.RawlsWorkspace) {
+                    AggregatedWorkspace(workspaceContext,
+                                        Some(workspaceContext.googleProjectId),
+                                        azureCloudContext = None,
+                                        policies = List.empty
+                    )
+                  } else {
+                    // bubble up an MC workspace exception
+                    throw e
```
yuliadub, line 329: do we want to log any kind of message that explains why throwing here is okay?

yuliadub (approved):

nawatts (approved):

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
nawatts: I don't think so. This branch behaves the same as before this PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
ashanhol: im still trying to figure that out

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3042,6 +3042,27 @@ class WorkspaceServiceSpec
...
+      Map.empty
+    )
+    when(services.workspaceManagerDAO.getWorkspace(any[UUID], any[RawlsRequestContext])).thenReturn(
+      new WorkspaceDescription()
+        .stage(WorkspaceStageModel.RAWLS_WORKSPACE)
+        .policies(policies.asJava)
+    )
+
+    Await.result(
+      services.mcWorkspaceService.createMultiCloudWorkspace(workspaceRequest, new ProfileModel().id(UUID.randomUUID())),
```
ashanhol: I think this is the right way to go about it. From what I saw, looks like I either have to create a workspace this way _or_ call `services.workspaceService.createWorkspace`, otherwise calling both throws a "workspace' already exists" error.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -311,7 +311,24 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
...
+                  }
+              }
```
davidangb, line 331: This code is correct and in good shape, and I would not block the PR on changes here. However, the use of `try/catch` is Java-idiomatic, where the use of `Try/Success/Failure` is Scala-idiomatic. To be a good Scala citizen, you'd have something like this:

```suggestion
val wsmContext =
              Try(wsmService.fetchAggregatedWorkspace(workspaceContext, ctx)) match {
                case Success(found) => found
                case Failure(notFound: AggregateWorkspaceNotFoundException)
                    if workspaceContext.workspaceType == WorkspaceType.RawlsWorkspace =>
                  // return workspace with no WSM information for gcp workspace
                  AggregatedWorkspace(workspaceContext,
                                      Some(workspaceContext.googleProjectId),
                                      azureCloudContext = None,
                                      policies = List.empty
                  )
                case Failure(x) => throw x // bubble up an MC workspace exception
              }
```

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3148,6 +3170,47 @@ class WorkspaceServiceSpec
...
+    response.policies should not be empty
+    val policies: List[WorkspacePolicy] = response.policies.get
+    policies should not be empty
+    val policy: WorkspacePolicy = policies.head
+    policy.name shouldBe wsmPolicyInput.getName
+    policy.namespace shouldBe wsmPolicyInput.getNamespace
+    val additionalData = policy.additionalData
+    additionalData.length shouldEqual 2
+    additionalData.head.getOrElse("pair1Key", "fail") shouldEqual "pair1Val"
+    additionalData.tail.head.getOrElse("pair2Key", "fail") shouldEqual "pair2Val"
```
davidangb, line 3211: I also won't block the PR on any changes here; this code is fine. You could streamline a lot of the assertions by defining the expected result and checking for that, though:
```suggestion
    val response = readWorkspace.convertTo[WorkspaceResponse]

    response.workspace.name shouldBe workspaceName
    response.azureContext shouldEqual None
    response.workspace.cloudPlatform shouldBe Some(WorkspaceCloudPlatform.Gcp)

    val expectedPolicy: WorkspacePolicy = new WorkspacePolicy("test_name",
                                                              "test_namespace",
                                                              List(
                                                                Map("pair1Key" -> "pair1Val"),
                                                                Map("pair2Key" -> "pair2Val")
                                                              )
    )

    response.policies should contain(List(expectedPolicy))
```

davidangb (approved):
:+1:, with optional syntax suggestions - thank you!

PR #2826, opened 2024-04-11
[WOR-1627] Switch to V2 workspace creation API

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1627

The WSM v2 Create Workspace API allows the creation of the cloud context as part of the workspace creation. To somewhat simplify or workspace creation code, I have switched us to the newer API (which was are also using for the cloning path).

This wasn't quite as satisfying as I hoped, in that [SnapshotService](https://github.com/broadinstitute/rawls/blob/0f4b788b257f324d4df684d80ff36a6ee0bfcd27/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala#L142) is still calling `createWorkspace` (which uses the synchronous V1 API) to create a RAWLS stage workspace. I can't really justify making the SnapshotService code poll for workspace creation completion, so we will continue to use both the V1 and V2 APIs from Rawls.

Successful run of workspace e2e tests using this branch: https://github.com/broadinstitute/rawls/actions/runs/8652411416

---

**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAOSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAOSpec.scala
@@ -294,26 +294,31 @@ class HttpWorkspaceManagerDAOSpec
 
     val billingProjectId = "billing-project-namespace";
 
-    val expectedRequest = new CreateWorkspaceRequestBody()
-      .id(testData.azureWorkspace.workspaceIdAsUUID)
-      .displayName(testData.azureWorkspace.name)
-      .spendProfile(testData.azureBillingProfile.getId.toString)
-      .stage(WorkspaceStageModel.MC_WORKSPACE)
-      .applicationIds(Seq("exampleApp").asJava)
-      .policies(policyInputs)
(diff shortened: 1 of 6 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAOSpec.scala
```diff
@@ -294,26 +294,31 @@ class HttpWorkspaceManagerDAOSpec
...
+      CloudPlatform.AZURE,
       Some(policyInputs),
       testContext
     )
 
-    verify(workspaceApi).createWorkspace(expectedRequest)
+    val createWorkspaceArgumentCaptor = captor[CreateWorkspaceV2Request]
+    verify(workspaceApi).createWorkspaceV2(createWorkspaceArgumentCaptor.capture)
+    // V2-specific values
+    createWorkspaceArgumentCaptor.getValue.getJobControl shouldNot be(null)
```
cahrens, line 311: Since the jobId is randomly created when the method is called, we don't have a way to compare it.

aherbst-broad (approved):
Nice change, glad we're migrating over to the newer endpoint.

marctalbott (approved):

PR #2834, opened 2024-04-19
WOR-1545: Async workspace cloning

blakery:
Ticket: [WOR-1545](https://broadworkbench.atlassian.net/browse/WOR-1545)

* new field to store arguments as a simple `Map[String, String]` in monitor records
* new workspace states for cloning
* new `JobType`s for cloning
* new runner for cloning jobs and add to boot monitors
* moving logic for each cloning phase to separate steps called by the cloning runner

In addition to the unit tests, I tested both the old and new paths in a bee multiple times, and tracked the steps and state changes.

---

**PR checklist**

[WOR-1545]: https://broadworkbench.atlassian.net/browse/WOR-1545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
```diff
@@ -31,6 +37,16 @@ object WorkspaceManagerResourceMonitorRecord {
       JobType.WSMWorkspaceDeletionPoll
     )
 
+    val CloneWorkspaceInit: Value = Value("CloneWorkspaceInit")
+    val CreateWdsAppInClonedWorkspace: Value = Value("CreateWdsAppInClonedWorkspace")
+    val CloneWorkspaceContainerInit: Value = Value("CloneWorkspaceContainerInit")
+    val CloneWorkspaceAwaitContainerResult: Value = Value("CloneWorkspaceAwaitContainerResult")
+    val cloneJobTypes: List[WorkspaceManagerResourceMonitorRecord.JobType.Value] = List(
```
blakery, line 44: Making CloneWorkspaceAwaitContainerResult separate from the existing type for container cloning, to make it easier to ensure the code paths are kept completely separate for now.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceManagerResourceMonitorRecordComponent.scala
```diff
@@ -118,6 +157,12 @@ trait WorkspaceManagerResourceMonitorRecordComponent {
 
   class WorkspaceManagerResourceMonitorRecordTable(tag: Tag)
       extends Table[WorkspaceManagerResourceMonitorRecord](tag, "WORKSPACE_MANAGER_RESOURCE_MONITOR_RECORD") {
+
+    implicit val argsMapper: TypedType[Map[String, String]] = MappedColumnType.base[Map[String, String], String](
```
blakery, line 161: Mapper to serialize the new field.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BootMonitors.scala
```diff
@@ -492,7 +500,7 @@ object BootMonitors extends LazyLogging {
             billingRepo,
             new BillingProjectDeletion(samDAO, billingRepo, billingProfileManagerDAO)
           )
-        )
+        ) ++ JobType.cloneJobTypes.map(jobType => jobType -> workspaceCloneRunner).toMap
```
blakery, line 503: Map all clone job types to the clone runner, and concatenate it with existing mappings.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceInitStep.scala
```diff
@@ -0,0 +1,47 @@
...
+  override val jobType: JobType = JobType.CloneWorkspaceInit
+
+  override def runStep(userCtx: RawlsRequestContext): Future[JobStatus] = {
+    val jobId = WorkspaceCloningRunner.getInitialWSMJobId(job.args) match {
+      case None =>
+        return fail("Initial Workspace Clone", "No JobId for WSM cloning operation provided")
+          .map(_ => Complete)
+      case Some(id) => id
+    }
+    val result = workspaceManagerDAO.getCloneWorkspaceResult(workspaceId, jobId, userCtx)
```
blakery, line 37: I didn't map potential exceptions from WSM here, because we'd just want to retry any exceptions, since there isn't any scenario where we should get an exception from WSM (unlike deleting where a 404 or 403 is perfectly valid). 
The top-level monitor is set up to catch any unknown exceptions, and re-run the job on the next cycle.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -307,6 +309,103 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       } yield profileModel
     }
 
+  def cloneMultiCloudWorkspaceAsync(
```
blakery, line 312: This is all just straight up duplicated, for now; up to the call to `cloneAzureWorkspaceAsync` (instead of `cloneAzureWorkspace`).
When we're ready to kill the old path we can just change the call in `cloneMultiCloudWorkspace`, and remove this altogether.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -768,7 +867,8 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
 
   private def createNewWorkspaceRecord(workspaceId: UUID,
                                        request: WorkspaceRequest,
-                                       parentContext: RawlsRequestContext
+                                       parentContext: RawlsRequestContext,
+                                       state: WorkspaceState = WorkspaceState.Ready
```
blakery, line 871: I had to add the option to specify a different state, so we could create it as `Cloning`.
Defaults to `Ready` so I didn't have to fuss with changing existing calls.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -307,6 +309,103 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+          }
+
+        case (wsType, profileOpt) =>
+          Future.failed(
+            RawlsExceptionWithErrorReport(
+              ErrorReport(
+                StatusCodes.BadRequest,
+                s"Cloud platform mismatch: Cannot clone $wsType workspace '$sourceWorkspaceName' " +
+                  s"into billing project '${billingProject.projectName}' " +
+                  s"(hosted on ${profileOpt.map(_.getCloudPlatform).getOrElse(CloudPlatform.GCP)})."
```
aherbst-broad: Nit:The hardcoding of the fallback to GCP is a bit awkward here; I wonder if we should just throw rather than potentially returning misinformation.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -307,6 +309,103 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+        ) match {
+          case Success(cloneResult) =>
+            clonedMultiCloudWorkspaceCounter.inc()
+            logger.info(
+              "Created azure workspace " +
+                s"[ workspaceId='${newWorkspace.workspaceId}'" +
+                s", workspaceName='${newWorkspace.toWorkspaceName}'" +
+                s"]"
+            )
+            // The jobId returned by WSM is the default job id generated by stairway
```
aherbst-broad, line 382: Boy that's annoying. WSM should accept a job ID as an argument to it's async APIs, can you drop a ticket for this so we don't forget about it?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -307,6 +309,103 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+            // Currently, stairway generates this via a 'ShortUUID' which doesn't parse correctly
+            // So for the time being, we need to add it as a job argument instead
+            val base64EncodedJobId = cloneResult.getJobReport.getId
```
aherbst-broad, line 385: I'm confused, where does base64 encoding enter into things here?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/WorkspaceCloningRunner.scala
```diff
@@ -0,0 +1,160 @@
...
+        logFailure(msg)
+        return cloneFail(workspaceId, msg).map(_ => Complete)
+    }
+
+    getUserCtx(userEmail).transformWith {
+      case Failure(t) =>
+        val msg =
+          s"Unable to retrieve clone workspace results for workspace $workspaceId: unable to retrieve request context for $userEmail"
+        logFailure(msg, Some(t))
+        cloneFail(workspaceId, msg).map(_ => Incomplete)
```
aherbst-broad, line 128: Why does this go to Incomplete? Is this job in a retry-able state if we can't get a user context for the email?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  override def runStep(userCtx: RawlsRequestContext): Future[JobStatus] = {
+    val operationName = "Await Storage Container Clone"
+    Try(workspaceManagerDAO.getJob(job.jobControlId.toString, userCtx)) match {
+      case Success(result) => handleCloneResult(workspaceId, result)
+      case Failure(e: ApiException) =>
+        e.getMessage
+        e.getCode match {
+          case 500 =>
+            fail(operationName, e.getMessage).map(_ => Complete)
```
aherbst-broad: It'd be good to retry in this case; 500s are not impossible nor should they be interpreted as a immediately terminal state.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
```diff
@@ -0,0 +1,83 @@
...
+          case 404 =>
+            val msg = s"Unable to find job in WSM for clone container operation"
+            fail(operationName, msg).map(_ => Complete)
+          case 401 =>
+            val msg = s"Unable to get job result, user is unauthed with jobId ${job.jobControlId}: ${e.getMessage}"
+            fail(operationName, msg).map(_ => Complete)
+          case code =>
+            logger.error(s"API call to get clone result failed with status code $code: ${e.getMessage}")
+            Future.successful(Incomplete)
```
aherbst-broad: 🤔 I am unsure about the defaulting to Incomplete here. If we get a 403 for example should the job keep trying? I worry about non terminal jobs stacking up in the backend.  Do we have an eventual timeout for these sorts of jobs?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceCreateWDSAppStep.scala
```diff
@@ -0,0 +1,46 @@
...
+  workspaceId: UUID,
+  job: WorkspaceManagerResourceMonitorRecord
+)(implicit executionContext: ExecutionContext)
+    extends WorkspaceCloningStep(workspaceRepository, monitorRecordDao, workspaceId, job) {
+
+  override val jobType: JobType = JobType.CreateWdsAppInClonedWorkspace
+
+  override def runStep(userCtx: RawlsRequestContext): Future[JobStatus] =
+    if (WorkspaceCloningRunner.isAutomaticAppCreationDisabled(job.args)) {
+      scheduleNextJob(UUID.randomUUID()).map(_ => Complete)
```
aherbst-broad, line 32: I suggest adding a log breadcrumb here that app creation is being explicitly skipped. I can see this leading to hard to debug problems if we unexpectedly hit this branch.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/WorkspaceCloningStep.scala
```diff
@@ -0,0 +1,74 @@
...
+      .map(report => report.getMessage)
+      .getOrElse("Error not specified in job")
+    val jobId = result.getJobReport.getId
+    val message = s"Workspace Clone Operation [$operationName] failed for jobId [$jobId]: $errorMessage"
+    logger.error(s"${job.jobType} failure: $message")
+    workspaceRepository.setFailedState(workspaceId, CloningFailed, message)
+  }
+
+  def fail(operationName: String, errorMessage: String): Future[Int] = {
+    val message = s"Workspace Clone Operation [$operationName] failed for jobId [${job.jobControlId}]: $errorMessage"
```
aherbst-broad: Can this include a source + dest workspace ID breadcrumb? If the job record is deleted from the table it'll be a bit tricky to determine those data points for debugging. It'd be useful in the other overload for this method too.

aherbst-broad (commented):
Looking good. Is the workspace state being updated anywhere? I couldn't see it.  That is, we want to reflect that the workspace is being cloned, when the container is being cloned etc.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
blakery: Yeah, I should just go back and rework these cases.
I took all this pretty much directly from the original `CloneWorkspaceContainerRunner`, under the theory that this step was pretty much the same logic.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/WorkspaceCloningRunner.scala
blakery: Yes, this leaves the job in a retryable state, but the error shows up on the workspace. 

It's the same pattern we use for this type of failure in all our other runners. 
The rational being that these are probably transient errors, like permission propagation delays, or connection issues, but if it happens, we want to see it on the workspace, in case it isn't transient.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
blakery: The `ShortUUID` from stairway is base64 encoded, in addition to not being the full UUID. I tried to parse it, briefly.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
blakery: Done: https://broadworkbench.atlassian.net/browse/WOR-1649

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
blakery: It's an error message, so I can just make it default to a string. I'll change it in the original location as well.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
blakery: I reworked this to return `Complete` for 4xx and `Incomplete` for others.

blakery:
> Looking good. Is the workspace state being updated anywhere? I couldn't see it. That is, we want to reflect that the workspace is being cloned, when the container is being cloned etc.

I forgot about the `CloningContainer` state, so I added that just now. 
`Cloning` is set on the original record creation. 
`Ready` is in `CloneWorkspaceAwaitStorageContainerStep.cloneSuccess()`.

aherbst-broad (approved):

marctalbott (approved):

PR #2841, opened 2024-04-22
AJ-1788: no stack trace in response for create-snapshot-reference on TDR API exception

davidangb:
Ticket: AJ-1788

For the API create-snapshot-reference API: if we encounter an ApiException retrieving the snapshot from TDR, don't include a stack trace in the error response.

This should address a low-priority security finding about verbose error messages in CWDS. CWDS is just propagating the stack trace up the chain from Rawls.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
@@ -170,8 +170,8 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
         )
       // else, propagate by wrapping in an error report
       case Failure(other) =>
-        logger.warn(s"Unexpected error when retrieving snapshot: ${other.getMessage}")
-        throw new RawlsExceptionWithErrorReport(ErrorReport(other))
+        logger.warn(s"Unexpected error when retrieving snapshot: ${other.getMessage}", other)
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, other.getMessage))
     }
 
   def getSnapshotResourceFromWsm(workspaceName: WorkspaceName,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -158,10 +158,17 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+      // if snapshot not found in TDR, this is a bad request
       case Failure(ex: ApiException) if ex.getCode == StatusCodes.NotFound.intValue =>
         throw new RawlsExceptionWithErrorReport(
           ErrorReport(StatusCodes.BadRequest, s"Snapshot ${snapshotIdentifiers.snapshotId} not found.")
         )
+      // on some other TDR API exception, strip the stack trace and propagate
+      case Failure(ex: ApiException) =>
+        throw new RawlsExceptionWithErrorReport(
+          ErrorReport(ex.getCode, ex.getMessage)
+        )
```
davidangb, line 170: this `ErrorReport(StatusCode, String)` constructor _does not_ propagate a stack trace

core/src/main/scala/org/broadinstitute/dsde/rawls/snapshot/SnapshotService.scala
```diff
@@ -158,13 +158,20 @@ class SnapshotService(protected val ctx: RawlsRequestContext,
...
+      // else, propagate by wrapping in an error report
       case Failure(other) =>
-        logger.warn(s"Unexpected error when retrieving snapshot: ${other.getMessage}")
-        throw new RawlsExceptionWithErrorReport(ErrorReport(other))
+        logger.warn(s"Unexpected error when retrieving snapshot: ${other.getMessage}", other)
+        throw new RawlsExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, other.getMessage))
```
davidangb, line 174: same `ErrorReport(StatusCode, String)` constructor here - does not propagate a stack trace. We now log the stack trace instead of including it in the response.

nawatts (approved):

calypsomatic (approved):

PR #2853, opened 2024-05-01
[WOR-1631] Fix deletion bug when creating workspace with a name already in use

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1631

Fixes a bug introduced by #2224.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -648,6 +648,9 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       _ <- createWdsAppInWorkspace(workspaceId, parentContext, None, workspaceRequest.attributes)
 
     } yield savedWorkspace).recoverWith {
+      case r: RawlsExceptionWithErrorReport if r.errorReport.statusCode.contains(StatusCodes.Conflict) =>
+        // Workspace already exists with this name/namespace, and we don't want to delete it.
+        Future.failed(r)
       case e @ (_: ApiException | _: WorkspaceManagerOperationFailureException | _: RawlsExceptionWithErrorReport) =>
         logger.info(s"Error creating workspace ${workspaceRequest.toWorkspaceName} [workspaceId = ${workspaceId}]", e)
         for {
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
@@ -303,7 +303,12 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
 
     // Make sure that the pre-existing workspace was not deleted.
     verify(workspaceManagerDAO, times(0)).deleteWorkspaceV2(any(), anyString(), any())
-    Await.result(slickDataSource.inTransaction(_.workspaceQuery.findByName(WorkspaceName(namespace, workspaceName))), Duration.Inf).get.name shouldBe workspaceName
+    Await
+      .result(slickDataSource.inTransaction(_.workspaceQuery.findByName(WorkspaceName(namespace, workspaceName))),
+              Duration.Inf
+      )
+      .get
+      .name shouldBe workspaceName
   }
 
   it should "throw an exception if the billing profile was created before 9/12/2023" in {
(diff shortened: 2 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -298,6 +300,10 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
     }
 
     thrown.errorReport.statusCode shouldBe Some(StatusCodes.Conflict)
+
+    // Make sure that the pre-existing workspace was not deleted.
+    verify(workspaceManagerDAO, times(0)).deleteWorkspaceV2(any(), anyString(), any())
```
cahrens, line 305: I verified that both of these statements fail without the bug fix.

I considered also adding a test that ensures that a `RawlsExceptionWithErrorReport` with a different StatusCode still deletes the workspace, but looking through the code I didn't see any method we are calling in the creation block that would throw such an error. Do you think I should remove `RawlsExceptionWithErrorReport` from the deletion error handling all together?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -648,6 +648,9 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       _ <- createWdsAppInWorkspace(workspaceId, parentContext, None, workspaceRequest.attributes)
 
     } yield savedWorkspace).recoverWith {
+      case r: RawlsExceptionWithErrorReport if r.errorReport.statusCode.contains(StatusCodes.Conflict) =>
```
cahrens, line 651: Thrown by https://github.com/broadinstitute/rawls/blob/bae3a702f5f9e321b25b607f95abe783a62a9d49/core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala#L279

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
marctalbott: We need to keep `RawlsExceptionWithErrorReport` since that is thrown by `buildPolicyInputs` for invalid policies ([link](https://github.com/broadinstitute/rawls/blob/1824ef0ebf0834483542970b626f2841397bbdc0/model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala#L788)). It would be worth adding that other test to cover this IMO

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: Thanks for the pointer! I will add a test case for that.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
blakery: I don't know how removing `RawlsExceptionWithErrorReport` from the deletion error handling would work. What would you be matching on?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
blakery: I'd like this to be closer/more tightly scoped to the call site. But I don't see a way to do that without refactoring the entire function (which I'm sure is way more than what you want to do right now).

Which is to say: LGTM.

blakery (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
cahrens: I meant removing it from the exceptions that we handled and then deleted the workspace on (the following case statement). But Marc pointed out a reason why I should leave it there, so I will add a test case for that.

cahrens: @marctalbott actually, you already wrote a test for that case. :) https://github.com/broadinstitute/rawls/blob/1824ef0ebf0834483542970b626f2841397bbdc0/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala#L646

I verified that the test case fails if I remove the handling for `RawlsExceptionWithErrorReport`.

PR #2854, opened 2024-05-01
[WOR-1644] Pass additional policies on clone

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1644

Pass additional policies supplied during the clone request to WSM. This fixes the case of cloning an Azure workspace without PHI tracking enabled -> a destination workspace with PHI tracking enabled (destination billing project must be protected and enterprise).

Tested locally with locally running terra-ui (support has already been merged there).

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
@@ -120,15 +120,16 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
                               spendProfile: Option[ProfileModel],
                               billingProjectNamespace: String,
                               ctx: RawlsRequestContext,
-                              location: Option[String]
+                              additionalPolicyInputs: Option[WsmPolicyInputs]
   ): CloneWorkspaceResult = {
     val request = new CloneWorkspaceRequest()
       .destinationWorkspaceId(workspaceId)
       .displayName(displayName)
-      .location(location.orNull)
       .projectOwnerGroupId(billingProjectNamespace)
 
     spendProfile.map(_.getId.toString).map(request.spendProfile)
+
+    additionalPolicyInputs.map(request.additionalPolicies)
 
     getWorkspaceApi(ctx).cloneWorkspace(
       request,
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -120,16 +120,17 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
                               spendProfile: Option[ProfileModel],
                               billingProjectNamespace: String,
                               ctx: RawlsRequestContext,
-                              location: Option[String]
```
cahrens, line 123: Location isn't used anywhere (I think it was before location was tied to the billing project), so I went ahead and removed it.

cahrens: Cloning an Azure workspace through `MultiCloudWorkspaceSerivce` now always passes a policies argument. But I did not add policies to the clone call for a Rawls stage workspace in `WorkspaceService`. Can anyone think of a reason why we should pass it there too?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -368,7 +368,8 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
             displayName = request.name,
             spendProfile = Option(profile),
             billingProjectNamespace = request.namespace,
-            context
+            context,
+            buildPolicyInputs(request)
```
cahrens, line 372: I don't see that there's any test coverage for this method-- added a note in https://broadinstitute.slack.com/archives/C02L9L96YB1/p1714575548800209

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
aherbst-broad: The only use case I can think of is a GCP workspace with a snapshot getting cloned and the user clicking the PHI button -- but I don't think that's UI we support atm. So, no, not right now.

aherbst-broad (approved):

blakery (approved):

PR #2863, opened 2024-05-07
AJ-1827: all_attribute_values max length

davidangb:
Ticket: <Link to Jira ticket>

The `ENTITY.all_attribute_values` column in the db is a `TEXT` column. This has "a maximum length of 65,535 (216 − 1) characters. The effective maximum length is less if the value contains multibyte characters." ([link](https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html#:~:text=A%20TEXT%20column%20with%20a%20maximum%20length%20of%2065%2C535%20(2,be%20given%20for%20this%20type)))

In Rawls code, we truncate strings before inserting into this column. However, we truncated by character count, not byte count - so if the string contained multibyte characters we could overflow the column.

In this PR, we truncate to a byte limit, not to a character count limit.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
@@ -18,6 +18,7 @@ import slick.dbio.Effect.Read
 import slick.jdbc.{GetResult, JdbcProfile, ResultSetConcurrency, ResultSetType, SQLActionBuilder, TransactionIsolation}
 import slick.sql.SqlStreamingAction
 
+import java.nio.charset.StandardCharsets
 import java.sql.Timestamp
 import java.util.{Date, UUID}
 import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala
```diff
@@ -72,8 +73,8 @@ case class EntityAndAttributesResult(entityRecord: EntityRecord,
 )
 
 object EntityComponent {
-  // the length of the all_attribute_values column, which is TEXT, -1 becaue i'm nervous
-  val allAttributeValuesColumnSize = 65534
+  // the length of the all_attribute_values column, which is TEXT, minus a few bytes because i'm nervous
+  val allAttributeValuesColumnSize = 65532
```
davidangb, line 77: 65532 is divisible by 4 so it makes the unit test work better, and only drops 2 bytes.

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponentSpec.scala
```diff
@@ -583,8 +601,9 @@ class EntityComponentSpec extends TestDriverComponentWithFlatSpecAndMatchers wit
               .result
           )
         )
+        // measure the byte size of the actual value is exactly what we specified
         assertResult(EntityComponent.allAttributeValuesColumnSize) {
-          entityRec.get.allAttributeValues.get.length
+          entityRec.get.allAttributeValues.get.getBytes(StandardCharsets.UTF_8).length
         }
       }
```
davidangb, line 608: I verified that this test fails as `java.sql.BatchUpdateException: Data truncation: Data too long for column 'all_attribute_values' at row 1` without the runtime changes in `EntityComponent`

calypsomatic (approved):

nawatts (approved):

PR #2885, opened 2024-05-22
AJ-1836 Expose bpm's limits

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1836
Includes the new `limits` field in a BPM organization.  Verified on my BEE that it is returned and everything seems to function correctly.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
@@ -90,7 +90,7 @@ case class WorkspaceBillingAccount(
   currentBillingAccountOnGoogleProject: Option[RawlsBillingAccountName]
 )
 
-case class RawlsBillingProjectOrganization(enterprise: Boolean)
+case class RawlsBillingProjectOrganization(enterprise: Boolean, limits: Map[String, String])
 
 case class RawlsBillingProjectResponse(
   projectName: RawlsBillingProjectName,
@@ -327,7 +327,7 @@ class UserAuthJsonSupport extends JsonSupport {
     WorkspaceBillingAccount
   )
 
-  implicit val BillingProjectOrganizationFormat: RootJsonFormat[RawlsBillingProjectOrganization] = jsonFormat1(
+  implicit val BillingProjectOrganizationFormat: RootJsonFormat[RawlsBillingProjectOrganization] = jsonFormat2(
     RawlsBillingProjectOrganization.apply
   )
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -348,7 +348,9 @@ class UserService(
         platform,
         protectedData,
         region = maybeRegion,
-        organization = Option(p.getOrganization).map(org => RawlsBillingProjectOrganization(org.isEnterprise))
+        organization = Option(p.getOrganization).map(org =>
+          RawlsBillingProjectOrganization(org.isEnterprise, org.getLimits.asScala.toMap)
```
calypsomatic, line 352: Since the imported `Organization` is java, there's a lot of compatibility issues between java map and scala map.  I have a lot of `asJava` and `asScala` scattered around and would love input on any better ways to deal with this.

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -90,7 +90,7 @@ case class WorkspaceBillingAccount(
   currentBillingAccountOnGoogleProject: Option[RawlsBillingAccountName]
 )
 
-case class RawlsBillingProjectOrganization(enterprise: Boolean)
+case class RawlsBillingProjectOrganization(enterprise: Boolean, limits: Map[String, String])
```
davidangb, line 93: will the `limits` key always be present in the BPM response, even if its value is `{}`? If not, you may need to make this an `Option[Map[String,String]]` to deal with the missingness

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
davidangb: I don't think the `asJava`/`asScala` proliferation is too bad - I like having them inline in the code because then when I'm reading the code it's easy to see that a `Map` is not the same as a `Map`

davidangb (commented):
one question inline, otherwise looks good

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
calypsomatic: Yes, I managed to have it default to an empty map in BPM, so if there's no limits present it will be {}

davidangb (approved):

PR #2886, opened 2024-05-22
[WOR-1679] Don't update workspace or billing project state until async job polling completes

marctalbott:
Ticket: [WOR-1679](https://broadworkbench.atlassian.net/browse/WOR-1679)
* When polling on an async workspace or billing project operation (LZ create, workspace clone, etc.), don't update the state on the resource until the job is complete. 

---

**PR checklist**

[WOR-1679]: https://broadworkbench.atlassian.net/browse/WOR-1679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStepSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStepSpec.scala
@@ -33,7 +33,7 @@ class CloneWorkspaceAwaitStorageContainerStepSpec
 
   behavior of "retrieving the report for the container cloning job"
 
-  it should "report errors from api response and return Incomplete for jobs failed with a 500" in {
+  it should "return Incomplete for jobs failed with a 500" in {
     val ctx = mock[RawlsRequestContext]
     val wsmDao = mock[WorkspaceManagerDAO]
     val apiMessage = "some failure message"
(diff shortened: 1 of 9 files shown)


blakery (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStepSpec.scala
```diff
@@ -69,7 +69,7 @@ class CloneWorkspaceAwaitStorageContainerStepSpec
       monitorRecord
     )
     whenReady(runner.runStep(ctx))(_ shouldBe WorkspaceManagerResourceMonitorRecord.Incomplete)
-    verify(workspaceRepository).setFailedState(
+    verify(workspaceRepository, never).setFailedState(
```
cahrens, line 72: Nit: need to update the spec name as we don't report the errors here.

cahrens:
Is there already existing test coverage that we DO report the error ultimately if the operations can never succeed? And where in the code does that happen?

I'm also wondering if there would be value in storing the last error that was hit, even if we don't set the status to something terminal. Though I don't know if there is a place to do that.

marctalbott:
> Is there already existing test coverage that we DO report the error ultimately if the operations can never succeed? And where in the code does that happen?
> 
> I'm also wondering if there would be value in storing the last error that was hit, even if we don't set the status to something terminal. Though I don't know if there is a place to do that.

We never stop retrying as long as it returns `Incomplete`. There is a ticket to stop retrying after 24 hours - https://broadworkbench.atlassian.net/browse/WOR-1682

As for storing the last error, there isn't really a place to do that. Repurposing the error message field on the workspace and billing project could have unintended consequences if any place assumes that an error message equates to a failed state. We could be more consistent about logging the errors, but I didn't want to blow up our log volume either.

cahrens (approved):

PR #2898, opened 2024-05-30
[WOR-1695] Return policy information for GCP workspaces from list method

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1695

This is related to https://github.com/broadinstitute/rawls/pull/2825, which added the functionality to the details endpoint.

Tested manually with [this workspace](https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/import-test) in a locally running Rawls.

---

**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
@@ -3541,7 +3541,7 @@ class WorkspaceServiceSpec
         new WorkspaceDescription()
           .id(googleWorkspace.workspaceIdAsUUID)
           .gcpContext(new GcpContext())
-          .stage(WorkspaceStageModel.MC_WORKSPACE)
+          .stage(WorkspaceStageModel.RAWLS_WORKSPACE)
       )
     )
 
@@ -3626,7 +3626,7 @@ class WorkspaceServiceSpec
           new WorkspaceDescription()
             .id(googleWorkspace.workspaceIdAsUUID)
             .gcpContext(new GcpContext())
-            .stage(WorkspaceStageModel.MC_WORKSPACE)
+            .stage(WorkspaceStageModel.RAWLS_WORKSPACE)
         )
       )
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3512,7 +3538,10 @@ class WorkspaceServiceSpec
         new WorkspaceDescription().id(deletingAzureWorkspace.workspaceIdAsUUID).stage(WorkspaceStageModel.MC_WORKSPACE),
         // no azureContext, should not be returned
         new WorkspaceDescription().id(readyAzureWorkspace.workspaceIdAsUUID).stage(WorkspaceStageModel.MC_WORKSPACE),
-        new WorkspaceDescription().id(googleWorkspace.workspaceIdAsUUID).gcpContext(new GcpContext())
+        new WorkspaceDescription()
+          .id(googleWorkspace.workspaceIdAsUUID)
+          .gcpContext(new GcpContext())
+          .stage(WorkspaceStageModel.RAWLS_WORKSPACE)
```
cahrens, line 3544: Without a stage set, there is no matching case in [this method](https://github.com/broadinstitute/rawls/blob/f85710459f67412b52516960707c4e685c6b7f83/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/AggregatedWorkspaceService.scala#L106), which is called from the `listWorkspaces` method when WSM returns a matching workspace for the Rawls workspace (before this code change, Rawls workspaces would short-circuit and not check if there was a matching WSM workspace).

I believe we should always expect a WSM workspace to have the stage set.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -3045,29 +3045,44 @@ class WorkspaceServiceSpec
...
+
+    when(
+      services.workspaceManagerDAO.getWorkspace(ArgumentMatchers.eq(createdWorkspace.workspaceIdAsUUID),
+                                                any[RawlsRequestContext]
+      )
+    ).thenReturn(
+      workspaceDescription
+    )
+
+    when(services.workspaceManagerDAO.listWorkspaces(any, any)).thenReturn(
```
cahrens, line 3080: The list endpoints use the `listWorkspaces` method instead of the `getWorkspace` method. It makes sense to mock both calls for consistency, and then I can use the same method for `getDetails` and `list` tests.

nawatts (approved):

aherbst-broad (approved):

blakery (approved):

PR #2905, opened 2024-06-03
[WOR-1680] Mark Rawls workspace state as CloningFailed if Azure workspace resource failed to clone

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1680

This ticket improves our error visibility/handling for Azure workspace cloning. If one of the workspace resources that is meant to clone fails to clone, store the error message for that and mark the Rawls workspace as `CloningFailed`. This handling is necessary because WSM resource cloning subflights do not fail the parent clone flight.

Example display for a Workspace that has the CBAS database issue (which is now fixed for new workspaces): 

image

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
@@ -29,7 +29,7 @@ class CloneWorkspaceAwaitStorageContainerStep(
 )(implicit executionContext: ExecutionContext)
     extends WorkspaceCloningStep(workspaceRepository, monitorRecordDao, workspaceId, job) {
 
-  override val jobType: JobType = JobType.CloneWorkspaceContainerInit
+  override val jobType: JobType = JobType.CloneWorkspaceAwaitContainerResult
 
   override def runStep(userCtx: RawlsRequestContext): Future[JobStatus] = {
     val operationName = "Await Storage Container Clone"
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceAwaitStorageContainerStep.scala
```diff
@@ -29,7 +29,7 @@ class CloneWorkspaceAwaitStorageContainerStep(
 )(implicit executionContext: ExecutionContext)
     extends WorkspaceCloningStep(workspaceRepository, monitorRecordDao, workspaceId, job) {
 
-  override val jobType: JobType = JobType.CloneWorkspaceContainerInit
+  override val jobType: JobType = JobType.CloneWorkspaceAwaitContainerResult
```
cahrens, line 32: This was a copy/paste error, but as far as I can tell, it had no effect because we never call `scheduleNextJob` from this step (since it is the last one).

aherbst-broad: Oof, good catch.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/clone/CloneWorkspaceInitStep.scala
```diff
@@ -37,7 +40,23 @@ class CloneWorkspaceInitStep(
...
     result.getJobReport.getStatus match {
       case StatusEnum.SUCCEEDED =>
-        scheduleNextJob(UUID.randomUUID()).map(_ => Complete)
+        val workspaceResult = result.getWorkspace();
+        val errorMessages = new StringBuilder
+        if (workspaceResult != null) {
+          for (resource <- workspaceResult.getResources().asScala)
+            if (resource.getResult() == CloneResourceResult.FAILED) {
+              if (errorMessages.nonEmpty) {
+                errorMessages.append(", ");
```
aherbst-broad, line 49: String has a [join](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-) method for this sort of concatenation

aherbst-broad (approved):

marctalbott (approved):

PR #2919, opened 2024-06-17
Return cloudPlatform in response from createWorkspace (WOR-723).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-723

Populates `cloudPlatform` in the `WorkspaceDetails` as returned from the create and clone methods. Previously it was just None, as the argument is optional.

Tested with a locally running Rawls and locally running terra-ui (though that only hits the v2 clone endpoint).
Dependent terra-ui PR: https://github.com/DataBiosphere/terra-ui/pull/4897

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -38,7 +38,9 @@ import org.broadinstitute.dsde.rawls.model.{
   RawlsRequestContext,
   SamWorkspaceActions,
   Workspace,
+  WorkspaceCloudPlatform,
   WorkspaceDeletionResult,
+  WorkspaceDetails,
   WorkspaceName,
   WorkspaceRequest,
   WorkspaceState,
@@ -226,7 +228,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
   def createMultiCloudOrRawlsWorkspace(workspaceRequest: WorkspaceRequest,
                                        workspaceService: WorkspaceService,
                                        parentContext: RawlsRequestContext = ctx
-  ): Future[Workspace] =
+  ): Future[WorkspaceDetails] =
     for {
       billingProject <- traceFutureWithParent("getBillingProjectContext", parentContext) { s =>
         getBillingProjectContext(RawlsBillingProjectName(workspaceRequest.namespace), s)
@@ -606,7 +608,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
   def createMultiCloudWorkspace(workspaceRequest: WorkspaceRequest,
                                 profile: ProfileModel,
                                 parentContext: RawlsRequestContext = ctx
-  ): Future[Workspace] = {
+  ): Future[WorkspaceDetails] = {
     assertBillingProfileCreationDate(profile)
 
     traceFutureWithParent("createMultiCloudWorkspace", parentContext)(s1 =>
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -262,14 +264,20 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
-            workspaceService.createWorkspace(workspaceRequest, s)
+            workspaceService
+              .createWorkspace(workspaceRequest, s)
+              .map(workspace => (workspace, WorkspaceCloudPlatform.Gcp))
           }
         )
-    } yield workspace
+    } yield WorkspaceDetails.fromWorkspaceAndOptions(workspace,
+                                                     Some(workspaceRequest.authorizationDomain.getOrElse(Set.empty)),
+                                                     useAttributes = true,
```
cahrens, line 278: The `WorkspaceDetails` constructor we used to call set `useAttributes` to true.

`fromWorkspaceAndOptions(workspace, Option(authorizationDomain), useAttributes = true)`

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceUnitTestsSpec.scala
```diff
@@ -122,8 +143,125 @@ class MultiCloudWorkspaceServiceUnitTestsSpec
...
+      None
+    )
+
+    doReturn(Future(billingProject))
+      .when(service)
+      .getBillingProjectContext(RawlsBillingProjectName(destWorkspaceRequest.namespace), requestContext)
+
+    doReturn(Future()).when(service).requireCreateWorkspaceAction(billingProject.projectName, requestContext)
+
+    // Because there is no billing profile, we will call the legacy GCP clone method.
```
marctalbott: Just to clarify, this calls the legacy GCP clone method because it's cloning a Rawls workspace, not because there's a missing billing profile right? We have Rawls workspaces with billing profiles now

marctalbott (approved):
Wanted to clarify one comment, but otherwise it all looks good to me

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -279,6 +279,28 @@ object Workspace {
...
+                     workspaceId: String,
+                     createdDate: DateTime,
+                     lastModified: DateTime,
+                     createdBy: String,
+                     attributes: AttributeMap,
+                     state: WorkspaceState,
+                     workspaceType: WorkspaceType
+  ) = {
+    val googleProjectId =
+      if (workspaceType == WorkspaceType.RawlsWorkspace) GoogleProjectId("google-id") else GoogleProjectId("")
```
blakery, line 303: I thought this was fishy, until I realized this was the test constructor. 
I still think its fishy, but only for existing reasons.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceUnitTestsSpec.scala
cahrens: Good point. We call the legacy method if either there is no billing profile or there is one, but its cloudPlatform is GCP. I will change the mocking to return a GCP billing profile so it is clearer and testing our new expected path.

```
        case (RawlsWorkspace, profileOpt)
            if profileOpt.isEmpty ||
              profileOpt.map(_.getCloudPlatform).contains(CloudPlatform.GCP) =>

blakery (approved):

PR #2921, opened 2024-06-18
ID-1308 Rawls logs Submission Metrics

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1308

CBAS is going to start emitting events to Bard to track submission metrics in Azure, but we need to do the same in GCP. ~The simplest way to do this is to have Rawls log the info and have that get shoved into BigQuery via a log sink.~

Having Rawls log in the correct format proved to be more difficult that expected. Instead of re-doing everything about how Rawls logs and possibly breaking something, it's easier to just call out to Bard.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
```diff
@@ -420,6 +421,7 @@ trait WorkflowSubmission extends FutureSupport with LazyLogging with MethodWiths
                                                          RawlsUserEmail(submissionRec.submitterEmail)
       )
       petUserInfo <- googleServicesDAO.getUserInfoUsingJson(petSAJson)
+      samUserInfo <- samDAO.getUserStatus(RawlsRequestContext(petUserInfo))
```
Ghost-in-a-Jar: This is a new api call out to sam on workflow submissions right? I think it'll be a drop in the ocean on requests to sam but figured it was worth calling out.

Ghost-in-a-Jar (approved):

samanehsan (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/WorkflowSubmissionActor.scala
tlangs: Yeah, this is a new call, but one that Bard would have had to make if Rawls was sending to Bard. Just the price of tying events to a user.

aherbst-broad (approved):

PR #2926, opened 2024-06-25
WOR-1664 Refactor MulticloudWorkspaceService to use WorkspaceRepository

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1664

Remove remaining instances of raw slick datasource usage in 'MultiCloudWorkspaceService`.
Tests were changed as little possible, to illustrate the unchanged functionality.

* Move additional functionality to `WorkspaceRepository`
* Use `BillingRepository` for appropriate datasource calls
* Split billing functionality from `WorkspaceSupport` to minimize disruption to unrelated classes

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
@@ -5,7 +5,18 @@ import cats.{Applicative, ApplicativeThrow}
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.billing.BillingRepository
 import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
-import org.broadinstitute.dsde.rawls.model.{CreationStatuses, ErrorReport, RawlsBillingProject, RawlsBillingProjectName, RawlsRequestContext, SamBillingProjectActions, SamBillingProjectRoles, SamResourceAction, SamResourceTypeName, SamResourceTypeNames}
+import org.broadinstitute.dsde.rawls.model.{
+  CreationStatuses,
+  ErrorReport,
+  RawlsBillingProject,
+  RawlsBillingProjectName,
+  RawlsRequestContext,
+  SamBillingProjectActions,
+  SamBillingProjectRoles,
+  SamResourceAction,
+  SamResourceTypeName,
+  SamResourceTypeNames
+}
 import org.broadinstitute.dsde.rawls.util.TracingUtils.traceFutureWithParent
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -16,12 +27,11 @@ trait BillingProjectSupport {
   protected val ctx: RawlsRequestContext
   implicit protected val executionContext: ExecutionContext
 
-
   def requireCreateWorkspaceAction(project: RawlsBillingProjectName, context: RawlsRequestContext = ctx): Future[Unit] =
     raiseUnlessUserHasAction(SamBillingProjectActions.createWorkspace,
-      SamResourceTypeNames.billingProject,
-      project.value,
-      context
+                             SamResourceTypeNames.billingProject,
+                             project.value,
+                             context
     ) {
       RawlsExceptionWithErrorReport(
         ErrorReport(
@@ -35,7 +45,7 @@ trait BillingProjectSupport {
   // granted on the Workspace's Billing Project
   def requireBillingProjectOwnerAccess(projectName: RawlsBillingProjectName,
                                        parentContext: RawlsRequestContext
-                                      ): Future[Unit] =
+  ): Future[Unit] =
     for {
       billingProjectRoles <- traceFutureWithParent("listUserRolesForResource", parentContext)(context =>
         samDAO.listUserRolesForResource(SamResourceTypeNames.billingProject, projectName.value, context)
@@ -50,13 +60,12 @@ trait BillingProjectSupport {
       }
     } yield ()
 
-
   /**
     * Load the specified billing project, throwing if the billing project is not ready.
     */
   def getBillingProjectContext(projectName: RawlsBillingProjectName,
                                context: RawlsRequestContext = ctx
-                              ): Future[RawlsBillingProject] =
+  ): Future[RawlsBillingProject] =
     for {
 
       maybeBillingProject <- traceFutureWithParent("loadBillingProject", context) { _ =>
@@ -84,9 +93,9 @@ trait BillingProjectSupport {
                                        resType: SamResourceTypeName,
                                        resId: String,
                                        context: RawlsRequestContext = ctx
-                                      )(
-                                        throwable: Throwable
-                                      ): Future[Unit] =
+  )(
+    throwable: Throwable
+  ): Future[Unit] =
     samDAO
       .userHasAction(resType, resId, action, context)
       .flatMap(ApplicativeThrow[Future].raiseUnless(_)(throwable))
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
@@ -21,12 +21,19 @@ import org.broadinstitute.dsde.rawls.util.TracingUtils.traceFutureWithParent
 
 import scala.concurrent.{ExecutionContext, Future}
 
+/**
+  * Trait for common functionality for billing projects.
+  * Split from WorkspaceSupport, to better fit scope
+  */
 trait BillingProjectSupport {
   val samDAO: SamDAO
   val billingRepository: BillingRepository
   protected val ctx: RawlsRequestContext
   implicit protected val executionContext: ExecutionContext
 
+  /**
+    * Throws an exception of the user does not have the "create_workspace" permission on the specified billing project
+    */
   def requireCreateWorkspaceAction(project: RawlsBillingProjectName, context: RawlsRequestContext = ctx): Future[Unit] =
     raiseUnlessUserHasAction(SamBillingProjectActions.createWorkspace,
                              SamResourceTypeNames.billingProject,
@@ -80,6 +87,9 @@ trait BillingProjectSupport {
       _ <- failUnlessBillingProjectReady(billingProject)
     } yield billingProject
 
+  /**
+    * Throws an exception if the specified billing project status is not in a Ready state.
+    */
   def failUnlessBillingProjectReady(billingProject: RawlsBillingProject): Future[Unit] =
     Applicative[Future].unlessA(billingProject.status == CreationStatuses.Ready) {
       Future.failed(
@@ -89,6 +99,9 @@ trait BillingProjectSupport {
       )
     }
 
+  /**
+    * Throws the passed exception if the user does not have the specified action on the specified resource
+    */
   private def raiseUnlessUserHasAction(action: SamResourceAction,
                                        resType: SamResourceTypeName,
                                        resId: String,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
@@ -1,0 +1,94 @@
+package org.broadinstitute.dsde.rawls.util
+
+import akka.http.scaladsl.model.StatusCodes
+import cats.{Applicative, ApplicativeThrow}
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
+import org.broadinstitute.dsde.rawls.billing.BillingRepository
+import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
+import org.broadinstitute.dsde.rawls.model.{CreationStatuses, ErrorReport, RawlsBillingProject, RawlsBillingProjectName, RawlsRequestContext, SamBillingProjectActions, SamBillingProjectRoles, SamResourceAction, SamResourceTypeName, SamResourceTypeNames}
+import org.broadinstitute.dsde.rawls.util.TracingUtils.traceFutureWithParent
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait BillingProjectSupport {
+  val samDAO: SamDAO
+  val billingRepository: BillingRepository
+  protected val ctx: RawlsRequestContext
+  implicit protected val executionContext: ExecutionContext
+
+
+  def requireCreateWorkspaceAction(project: RawlsBillingProjectName, context: RawlsRequestContext = ctx): Future[Unit] =
+    raiseUnlessUserHasAction(SamBillingProjectActions.createWorkspace,
+      SamResourceTypeNames.billingProject,
+      project.value,
+      context
+    ) {
+      RawlsExceptionWithErrorReport(
+        ErrorReport(
+          StatusCodes.Forbidden,
+          s"You are not authorized to create a workspace in billing project $project"
+        )
+      )
+    }
+
+  // Creating a Workspace without an Owner policy is allowed only if the requesting User has the `owner` role
+  // granted on the Workspace's Billing Project
+  def requireBillingProjectOwnerAccess(projectName: RawlsBillingProjectName,
+                                       parentContext: RawlsRequestContext
+                                      ): Future[Unit] =
+    for {
+      billingProjectRoles <- traceFutureWithParent("listUserRolesForResource", parentContext)(context =>
+        samDAO.listUserRolesForResource(SamResourceTypeNames.billingProject, projectName.value, context)
+      )
+      _ <- ApplicativeThrow[Future].raiseUnless(billingProjectRoles.contains(SamBillingProjectRoles.owner)) {
+        RawlsExceptionWithErrorReport(
+          ErrorReport(
+            StatusCodes.Forbidden,
+            s"Missing ${SamBillingProjectRoles.owner} role on billing project '$projectName'."
+          )
+        )
+      }
+    } yield ()
+
+
+  /**
+    * Load the specified billing project, throwing if the billing project is not ready.
+    */
+  def getBillingProjectContext(projectName: RawlsBillingProjectName,
+                               context: RawlsRequestContext = ctx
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
blakery, line 1: This changes in this file are significantly easier to understand if you hide whitespace.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -112,25 +91,6 @@ trait WorkspaceSupport {
       .userHasAction(resType, resId, action, context)
       .flatMap(ApplicativeThrow[Future].raiseUnless(_)(throwable))
 
-  // Creating a Workspace without an Owner policy is allowed only if the requesting User has the `owner` role
-  // granted on the Workspace's Billing Project
-  def requireBillingProjectOwnerAccess(projectName: RawlsBillingProjectName,
```
blakery, line 117: moved to `BillingSupport`

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -246,42 +182,4 @@ trait WorkspaceSupport {
       }
     }
 
-  /**
-    * Load the specified billing project, throwing if the billing project is not ready.
-    */
-  def getBillingProjectContext(projectName: RawlsBillingProjectName,
```
blakery, line 252: moved to `BillingSupport`

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -246,42 +182,4 @@ trait WorkspaceSupport {
...
-
-      billingProject = maybeBillingProject.getOrElse(
-        throw RawlsExceptionWithErrorReport(
-          ErrorReport(StatusCodes.BadRequest, s"Billing Project $projectName does not exist")
-        )
-      )
-      _ <- failUnlessBillingProjectReady(billingProject)
-    } yield billingProject
-
-  def failUnlessBillingProjectReady(billingProject: RawlsBillingProject): Future[Unit] =
```
blakery, line 270: moved to `BillingSupport`

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -246,42 +182,4 @@ trait WorkspaceSupport {
...
-    Applicative[Future].unlessA(billingProject.status == CreationStatuses.Ready) {
-      Future.failed(
-        RawlsExceptionWithErrorReport(
-          ErrorReport(StatusCodes.BadRequest, s"Billing Project ${billingProject.projectName} is not ready")
-        )
-      )
-    }
-
-  def failIfWorkspaceExists(name: WorkspaceName): ReadWriteAction[Unit] =
```
blakery, line 279: Moved to `WorkspaceService`, since it is no longer used anywhere else.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
```diff
@@ -0,0 +1,103 @@
...
+  SamBillingProjectRoles,
+  SamResourceAction,
+  SamResourceTypeName,
+  SamResourceTypeNames
+}
+import org.broadinstitute.dsde.rawls.util.TracingUtils.traceFutureWithParent
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait BillingProjectSupport {
```
aherbst-broad, line 28: Add a comment here explaining the need for this trait/interface.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/BillingProjectSupport.scala
```diff
@@ -0,0 +1,103 @@
...
+  val samDAO: SamDAO
+  val billingRepository: BillingRepository
+  protected val ctx: RawlsRequestContext
+  implicit protected val executionContext: ExecutionContext
+
+  def requireCreateWorkspaceAction(project: RawlsBillingProjectName, context: RawlsRequestContext = ctx): Future[Unit] =
```
aherbst-broad, line 37: Add comments for these methods.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -21,6 +32,26 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
       access.workspaceQuery.findById(workspaceId.toString)
     }
 
+  def getWorkspaceContext(workspaceName: WorkspaceName,
```
aherbst-broad: 🤔 I know this was nomenclature from `WorkspaceService` but: I'm wondering if we should make this an overload of `getWorkspace()` -- they're both returning `Future[Option[Workspace]]`. There doesn't seem to be anything context-ful about this method.  Thoughts?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -21,6 +32,26 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
...
+                          attributeSpecs: Option[WorkspaceAttributeSpecs] = None
+  ): Future[Option[Workspace]] = dataSource.inTransaction { dataAccess =>
+    dataAccess.workspaceQuery.findByName(workspaceName, attributeSpecs)
+  }
+
+  def getV2WorkspaceContext(workspaceName: WorkspaceName,
```
aherbst-broad: Is there a use-case for retaining access to non-V2 workspaces? If not I suggest either removing the non-v2 code paths or marking them deprecated for future removal so they don't continue getting ported.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -41,7 +72,43 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
       access.workspaceQuery.delete(workspace.toWorkspaceName)
     }
 
+  def deleteWorkspaceRecord(workspaceName: WorkspaceName): Future[Boolean] =
```
aherbst-broad: Nit: Call this `deleteWorkspace` to remain consistent with the naming scheme in the rest of this class (and to retain the repository abstraction over the data layer that we're dealing with objects and not records).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -41,7 +72,43 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
...
+    dataSource.inTransaction { access =>
+      access.workspaceQuery.delete(workspaceName)
+    }
+
   def updateCompletedCloneWorkspaceFileTransfer(wsId: UUID, finishTime: DateTime): Future[Int] =
     dataSource.inTransaction(_.workspaceQuery.updateCompletedCloneWorkspaceFileTransfer(wsId, finishTime.toDate))
 
+  def createNewMCWorkspaceRecord(workspaceId: UUID,
```
aherbst-broad: Two things:
* Same naming nit: `createMCWorkspace` (`new` goes without saying I think). 
* I'm wondering if it we should remove the dependency on `WorkspaceRequest`, as that feels like an API layer model leaking into the repo layer, especially since the only two things we care about on the request are teh name and the attributes.

aherbst-broad (commented):
Looking good!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: I didn't do that initially, because Scala doesn't allow multiple overloads with optional parameters.

Then I realized I could just use plain pass-through overloads for all of the combinations.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: I'm all for consolidating code paths.

I don't see much point in deprecating internal service functionality, if it's possible to just change it instead.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: It does seem a bit cleaner without the `WorkspaceRequest` there.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
```diff
@@ -75,31 +68,18 @@ trait WorkspaceSupport {
...
-        // verify the user has `read` on the workspace to avoid exposing its existence
-        require(SamWorkspaceActions.read, NoSuchWorkspaceException.apply) *> Future.failed(t)
+    for {
+      _ <- userEnabledCheck
+      workspace <- getWorkspaceContext(workspaceName)
+      workspaceId = workspace.workspaceId
+      _ <- raiseUnlessUserHasAction(SamWorkspaceActions.compute, SamResourceTypeNames.workspace, workspaceId) {
+        WorkspaceAccessDeniedException(workspaceName)
+      }.recoverWith { case t: Throwable =>
+        raiseUnlessUserHasAction(SamWorkspaceActions.read, SamResourceTypeNames.workspace, workspaceId) {
```
cahrens, line 79: Can you add the comment below back in? I was confused about this raise until I read the comment in the previous version.

```
 // verify the user has `read` on the workspace to avoid exposing its existence

File: core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala [hunk 12]
```diff
@@ -151,40 +112,34 @@ trait WorkspaceSupport {
...
-        withWorkspaceContext(workspaceName, dataAccess, attributeSpecs)(DBIO.successful)
-      }
-    }
-
-  def withWorkspaceContext[T](workspaceName: WorkspaceName,
-                              dataAccess: DataAccess,
-                              attributeSpecs: Option[WorkspaceAttributeSpecs] = None
-  )(op: (Workspace) => ReadWriteAction[T]) =
-    dataAccess.workspaceQuery.findByName(workspaceName, attributeSpecs) flatMap {
+    workspaceRepository.getWorkspace(workspaceName, attributeSpecs).map {
```
cahrens, line 125: Why is the `userEnabledCheck` is not needed?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -379,12 +375,16 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
     assertBillingProfileCreationDate(profile)
     val workspaceId = UUID.randomUUID()
     // Merge together source workspace and destination request attributes
-    val mergedRequest = request.copy(attributes = sourceWorkspace.attributes ++ request.attributes)
```
cahrens, line 382: If `mergedRequest` is unnecessary, please also remove it below at line 492.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -16,9 +27,20 @@ import scala.concurrent.Future
...
+      access.workspaceQuery.findV2WorkspaceById(workspaceId, attributeSpecs)
+    }
+
+  def getWorkspace(workspaceName: WorkspaceName): Future[Option[Workspace]] = getWorkspace(workspaceName, None)
+
+  def getWorkspace(workspaceName: WorkspaceName,
+                   attributeSpecs: Option[WorkspaceAttributeSpecs]
+  ): Future[Option[Workspace]] =
+    dataSource.inTransaction { dataAccess =>
+      dataAccess.workspaceQuery.findV2WorkspaceByName(workspaceName, attributeSpecs)
```
cahrens, line 43: I thought that transactions are used to combine multiple operations into a single transaction so, for example, they could be rolled back together. I am confused why a single query would need to be wrapped in a transaction.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -16,9 +27,20 @@ import scala.concurrent.Future
   */
 class WorkspaceRepository(dataSource: SlickDataSource) {
 
-  def getWorkspace(workspaceId: UUID): Future[Option[Workspace]] =
+  def getWorkspace(workspaceId: UUID): Future[Option[Workspace]] = getWorkspace(workspaceId, None)
+
+  def getWorkspace(workspaceId: UUID, attributeSpecs: Option[WorkspaceAttributeSpecs]): Future[Option[Workspace]] =
     dataSource.inTransaction { access =>
-      access.workspaceQuery.findById(workspaceId.toString)
+      access.workspaceQuery.findV2WorkspaceById(workspaceId, attributeSpecs)
```
cahrens, line 34: Why is this changing to the V2WorkspaceById method? I am getting confused about where we expose v1 workspaces and where we do not.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -36,12 +58,48 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
       access.workspaceQuery.updateStateWithErrorMessage(workspaceId, state, message)
     }
 
-  def deleteWorkspaceRecord(workspace: Workspace): Future[Boolean] =
+  def deleteWorkspace(workspace: Workspace): Future[Boolean] =
     dataSource.inTransaction { access =>
```
cahrens: You could just all the other signature of `deleteWorkspace` here.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: TBH, I never really considered it; I was just iterating on what was already there. I don't think I've seen any usage in the code base that wasn't wrapped in a transaction. 

Now it's kind of bugging me, and I want to go through and change all of the single call operations to `dataSource.dataAccess`.

cahrens: Perhaps it has to be wrapped in a transaction??? I just don't know.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: The method name previously _said_ it was for v2 workspaces, but wasn't using the v2 specific query.

Per @aherbst-broad's suggestion above, there isn't really a use case for continuing to provide access for v1 workspaces; the intent here was to be more more consistent about it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: I don't understand.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
cahrens: I don't understand… the previous method did not say it was for v2 workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
cahrens: You can call the method defined immediately below at line 66, getting the WorkspaceName from the workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/WorkspaceSupport.scala
blakery: Both call sites of this are methods that were already doing access checks. So I thought it was more appropriate to do the  `userEnabledCheck` in the same location.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: It doesn't _have_ to be: I checked, and you can get the `dataAccess` object outside of the transaction block.

That said, since there isn't anywhere else in the code base using it directly (at least that I'm aware of), I'd want to be a bit more intentional about trying it, and make sure I fully understood any implications first.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: There two parts to the refactor:
1. Eliminating the user-facing queries that return v2 workspaces. Included in this was removing the 'v2' component of the name, since it is now redundant.
2. The `getV2WorkspaceContext` (using the id) wasn't actually using the v2 specific query.

cahrens: OK, I'm still having trouble understanding the diff, but I agree that we only want to return v2 workspaces now.

cahrens (approved):

PR #2930, opened 2024-06-26
AJ-1905: dont log inTransactionWithAttrTempTable exceptions at ERROR level

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1905

We were logging and rethrowing errors in `DataSource.inTransactionWithAttrTempTable`. This led to many cases of double-logging, which led to many cases of double-alerting via Sentry.

This PR keeps the [log-and-rethrow anti-pattern](https://softwareengineering.stackexchange.com/questions/365427/try-catch-log-rethrow-is-anti-pattern), but changes the log to DEBUG level; debug-level logs are currently filtered out by our logback.xml.

Because I haven't audited EVERY place that calls `DataSource.inTransactionWithAttrTempTable`, I left the logging in; it would be trivial to turn on debug logs for `DataSource` if we needed them.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
@@ -2,13 +2,11 @@ package org.broadinstitute.dsde.rawls.dataaccess
 
 import _root_.slick.basic.DatabaseConfig
 import _root_.slick.jdbc.{JdbcProfile, TransactionIsolation}
-import akka.http.scaladsl.model.StatusCodes.ClientError
 import com.google.common.base.Throwables
 import com.typesafe.scalalogging.LazyLogging
 import liquibase.database.jvm.JdbcConnection
 import liquibase.resource.{ClassLoaderResourceAccessor, ResourceAccessor}
 import liquibase.{Contexts, Liquibase}
-import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, DataAccessComponent, ReadWriteAction}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
```diff
@@ -77,33 +75,15 @@ class SlickDataSource(val databaseConfig: DatabaseConfig[JdbcProfile])(implicit
     }
 
     database.run(callerActionWithTempTables.withPinnedSession).recover { case t: Throwable =>
-      logTransactionError(t, tempTableTypes)
+      // log the transaction error at DEBUG level. Since we rethrow the error, calling code is responsible for handling
+      // it and, if appropriate, logging at ERROR level. Only ERROR-level logs are picked up as Sentry alerts.
+      logger.debug(
+        s"Transaction with temporary tables failed for (${tempTableTypes.mkString(",")}). ${t.getClass.getName}: ${t.getMessage}"
+      )
       throw t
```
pshapiro4broad, line 83: Did you consider throwing a new exception that has the message with `tempTableTypes` in it? Or is the worry there that we don't know if some caller is depending on the current exception being thrown, and changing that could break that code?

pshapiro4broad (approved):
LGTM, but I wonder if someone would like to see the info that used to be logged in the exception message.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/DataSource.scala
davidangb: > we don't know if some caller is depending on the current exception being thrown, and changing that could break that code?

exactly this!

nawatts (approved):

calypsomatic (approved):

PR #2933, opened 2024-06-27
AJ-1910: ignore non-5xx errors logged by workbench-libs when talking to Google

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1910

Rawls sends a lot of alerts to Sentry that are actually 4xx errors from Google.

Turns out that workbench-libs, which Rawls relies on, logs all errors from Google at ERROR level. This happens even if the response from Google is a 4xx status code:

https://github.com/broadinstitute/workbench-libs/blob/5d1d38da768d1c1779877362a6fbb64b30c6aba0/util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/package.scala#L87

The code referenced above has a "warnOnError" flag, but that flag is mostly unused within workbench-libs, much less Rawls.

This PR adds filtering of Sentry events. It looks for specific Google exception classes, extracts their status codes, and prevents sending to Sentry for non-5xx errors. It also sends exceptions of code 0 to Sentry; 0 indicates an unknown status code.

Tested by running locally, setting a Sentry DSN in my local env, and monitoring results.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/util/SentryEventFilter.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/util/SentryEventFilter.scala
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.rawls.util
 
+import com.google.api.client.http.HttpResponseException
+import com.google.cloud.http.BaseHttpServiceException
 import io.sentry.SentryEvent
 
 object SentryEventFilter {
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/util/SentryEventFilterSpec.scala
```diff
@@ -87,11 +90,6 @@ class SentryEventFilterSpec extends AnyFlatSpec with Matchers {
     val messageEvent = evt(Some("other event"))
     messageEvent.setLogger(httpGoogleServicesDao)
     SentryEventFilter.filterEvent(messageEvent) shouldBe messageEvent
-
-    val throwable = new StorageException(403, "some event")
-    val throwableEvent = new SentryEvent(throwable)
-    throwableEvent.setLogger(httpGoogleServicesDao)
-    SentryEventFilter.filterEvent(throwableEvent) shouldBe throwableEvent
```
davidangb, line 94: StorageException 403s are now ignored, this part of the test is invalid

core/src/main/scala/org/broadinstitute/dsde/rawls/util/SentryEventFilter.scala
```diff
@@ -22,10 +24,24 @@ object SentryEventFilter {
...
+                null
               case _ =>
                 event
             }
           case _ => event
         }
     }
+
+  private def isStatusCodeIgnorable(code: Int): Boolean =
+    Option(code).nonEmpty && code > 0 && code / 100 < 5
```
calypsomatic: `code / 100 < 5` seems kinda weird... can't you just do like `code < 500` or something?  Or is there a good reason for this in particular?

calypsomatic (approved):

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/util/SentryEventFilter.scala
davidangb: lol you're absolutely right. I [copy-pasted](https://github.com/broadinstitute/rawls/blob/fd092c70b4e228bee4e90a7b04d9267a1edfd4dc/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala#L174) that and didn't stop to think about it - good catch!

PR #2939, opened 2024-07-01
WOR-1665: Refactor MultiCloudWorkspaceServiceSpec to use WorkspaceRepository for mocking

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1665

Goals of this refactor
* remove the requirement for a real database in MultiCloudWorkspaceServiceSpec (primary goal)
* remove usage of shared mocks 
  - it's easy to end up testing the behavior of the mocks
  - interactions can be hidden behind default setup
  - they add an unnecessary additional maintence burden to the class being mocked
* ensure tests are explicitly evaluating behavior, and improve specificity

I eventually split the tests in `MultiCloudWorkspaceServiceSpec` into separate test classes, both to ensure I wasn't still using resources in `TestDriverComponent` accidentally, and to make it easier to track which tests I had refactored.

---

**PR checklist**


core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceSpec.scala
```diff
@@ -74,176 +84,71 @@ class MultiCloudWorkspaceServiceSpec extends AnyFlatSpec with Matchers with Opti
...
+      mock[MultiCloudWorkspaceConfig],
+      mock[LeonardoDAO],
+      "MultiCloudWorkspaceService-test",
+      mock[WorkspaceManagerResourceMonitorRecordDao],
+      mock[WorkspaceRepository],
+      billingRepository
     )
     val workspaceRequest = WorkspaceRequest(
-      "fake_mc_billing_project_name",
+      testData.azureBillingProject.projectName.value,
```
blakery: This is a more realistic test, since now we're checking permissions for the same project we're returning.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceUnitTestsSpec.scala
blakery, line 1: This class contained tests for async cloning, that were already well-formed unit tests. 
They were moved to `MultiCloudWorkspaceServiceCloneSpec` with very little modification.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -78,6 +78,35 @@ object MultiCloudWorkspaceService {
 
   def getStorageContainerName(workspaceId: UUID): String = s"sc-${workspaceId}"
 
+  def buildPolicyInputs(workspaceRequest: WorkspaceRequest): Option[WsmPolicyInputs] = {
```
blakery, line 81: Moved to the companion object to make it clear it was a pure function, and make it easier to test directly.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -183,21 +212,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
 
   private def deleteMultiCloudWorkspace(workspace: Workspace): Future[WorkspaceDeletionResult] =
     for {
-      _ <- deleteWorkspaceInWSM(workspace.workspaceIdAsUUID).recover { case e: ApiException =>
```
blakery, line 186: The api exception is already being handled in `deleteWorkspaceInWSM`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -682,18 +709,6 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
     } yield {}
   }
 
-  private def getWorkspaceFromWsm(workspaceId: UUID, ctx: RawlsRequestContext): Option[WorkspaceDescription] =
```
blakery, line 685: Moved to an inner function since it is private and only used in one spot

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -906,28 +891,27 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
         )
       case Right(_) => ()
     }
+  }
 
   private def createWdsAppInWorkspace(workspaceId: UUID,
                                       parentContext: RawlsRequestContext,
                                       sourceWorkspaceId: Option[UUID],
                                       workspaceAttributeMap: AttributeMap
-  ): Future[Unit] =
```
blakery, line 914: This was wrapped in a future for no reason

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -541,10 +556,10 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
         }
 
         // create a WDS application in Leo
-        _ <- createWdsAppInWorkspace(workspaceId,
-                                     parentContext,
-                                     Some(sourceWorkspace.workspaceIdAsUUID),
-                                     request.attributes
+        _ = createWdsAppInWorkspace(workspaceId,
```
cahrens, line 559: What difference does this change make?

cahrens: Oh, I see--https://github.com/broadinstitute/rawls/pull/2939/files#r1670926092

blakery: The call to leo in synchronous, but everything was being unnecessarily wrapped in `Future`s. 
I removed the extraneous `Future`s in `createWdsAppInWorkspace`, so this changed from `flatmap`, to a simple assignment.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
```diff
@@ -0,0 +1,1398 @@
...
+        ),
+        Duration.Inf
+      )
+    }
+
+    result.errorReport.statusCode.value shouldBe StatusCodes.Forbidden
+  }
+
+  it should "fail if the destination workspace already exists" in {
+    // This is covered in the tests for the workspace repository
```
marctalbott: Is this included just for completion then? Or is there anything else that we need to verify here? Maybe error messages or cleanup or something?

marctalbott (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
cahrens: I imagine this was accidentally left behind.

If you wish to leave it for documentation purposes, I would suggest disabling the test or commenting it out-- seeing a test case that passed in the logs but actually did nothing would be very confusing if something related to this functionality actually broke.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
```diff
@@ -0,0 +1,1398 @@
...
+
+    result shouldBe WorkspaceDetails.fromWorkspaceAndOptions(destWorkspace,
+                                                             Some(Set.empty),
+                                                             useAttributes = true,
+                                                             Some(WorkspaceCloudPlatform.Azure)
+    )
+    verify(service).cloneAzureWorkspaceAsync(sourceWorkspace, billingProfile, destWorkspaceRequest, testContext)
+  }
+
+  it should "pass a request to clone a GCP workspace to cloneWorkspace in workspaceService" in {
```
cahrens, line 1092: The description of this seems the same as https://github.com/broadinstitute/rawls/pull/2939/files#diff-6a2f47cec9c431e15a9b8800cc081a8a558d119d8c0f952ad655263c6f6af621R338, but this one does have an auth domain. I'm wondering if the tests should be combined, or if the test description should be modified to make it clear what is different about this one.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
```diff
@@ -0,0 +1,849 @@
...
+      )
+    }
+
+    assert(actual.errorReport.message.contains("Unable to find billing profile"))
+  }
+
+  behavior of "createMultiCloudWorkspace"
+
+  it should "not delete the original workspace if a workspace with the same name already exists" in {
+    val workspaceManagerDAO = mock[WorkspaceManagerDAO] // spy(new MockWorkspaceManagerDAO())
```
cahrens: Stray comment at end.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
```diff
@@ -0,0 +1,849 @@
...
+    verify(workspaceManagerDAO)
+      .createAzureStorageContainer(
+        UUID.fromString(result.workspaceId),
+        MultiCloudWorkspaceService.getStorageContainerName(UUID.fromString(result.workspaceId)),
+        testContext
+      )
+    verify(leonardoDAO, never()).createWDSInstance(any, any, any)
+
+  }
+
```
cahrens, line 459: What happened to "create the workspace even if WDS instance creation fails"? 

https://github.com/broadinstitute/rawls/pull/2939/files#diff-b8196c4af84188faa6eb1f65595077eb0b25f6c06f5a8e64b5c60e4832b93118L481

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
```diff
@@ -0,0 +1,849 @@
...
+          WorkspaceRequest(namespace, name, Map.empty),
+          workspaceId,
+          new ProfileModel().id(UUID.randomUUID()),
+          testContext
+        ),
+        Duration.Inf
+      )
+    }
+
+    e.errorReport.statusCode shouldBe Some(StatusCodes.InternalServerError)
```
cahrens, line 569: Nit / style difference: when there are several things that all have to be checked as a block, in multiple test cases, I think having a helper method like `verifyWorkspaceCreationRollback` is actually quite nice.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
```diff
@@ -0,0 +1,849 @@
...
+            new WsmPolicyInput()
+              .name("protected-data")
+              .namespace("terra")
+              .additionalData(List().asJava)
+          ).asJava
+        )
+    )
+  }
+
+  it should "merge the protected data policy with other inputs in the request" in {
```
cahrens, line 812: What happened to "fail with an improperly structured additional fields element on a policy and rollback workspace creation"?

https://github.com/broadinstitute/rawls/pull/2939/files#diff-b8196c4af84188faa6eb1f65595077eb0b25f6c06f5a8e64b5c60e4832b93118L660

cahrens:
@blakery I'm done reviewing-- just have a few questions/minor comments.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
blakery: It wasn't left intentionally.
As I was going through tests, I would sometimes leave some that needed additional verification or something, but all the rest had a `todo` in the comments.
I just missed this one when cleaning them up, because I left out the `todo`.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
blakery: That test was effectively just testing that `createWdsAppInWorkspace` doesn't throw an exception when it fails, which was already covered by a test in [`MultiCloudWorkspaceServiceCloneSpec`](https://github.com/broadinstitute/rawls/blob/03f096a99c42784f3865d15fa4ae2e0f5ac84a3d/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala#L639C23-L639C25).

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
blakery: The difference is that one is for the sync version (`cloneMultiCloudWorkspace`), and one is for the async version (`cloneMultiCloudWorkspaceAsync`), and that part of the functionality should be the same for both versions.

The output on the test report is probably a bit more clear, because the grouping is more obvious.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
blakery: I thought there was a test in `WorkspaceModel`, but looks like I was mistaken. I'll add a test there.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCreateSpec.scala
blakery: I find that those tend to get hard to maintain quickly, and often end up lying about what they're doing.
I agree that it _can_ be worth it if you have a lot of the same things that have the same verifications many times; I think I probably just have a higher threshold.

cahrens (approved):

PR #2948, opened 2024-07-10
[WOR-1717] 400 when auth domain requested for Azure workspace instead of ignoring it

marctalbott:
Ticket: [WOR-1717](https://broadworkbench.atlassian.net/browse/WOR-1717)
* Return a 400 if any groups are passed in the `authorizationDomain` field of a create or clone workspace request. Rawls used to just ignore these groups which isn't a very clear user experience. Users can set an auth domain on the workspace in Sam by using a `group-constraint` policy

---

**PR checklist**

[WOR-1717]: https://broadworkbench.atlassian.net/browse/WOR-1717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
@@ -388,6 +388,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
   ): Future[Workspace] = {
 
     assertBillingProfileCreationDate(profile)
+    validateWorkspaceRequest(request)
     val workspaceId = UUID.randomUUID()
     // Merge together source workspace and destination request attributes
     val mergedAttributes = sourceWorkspace.attributes ++ request.attributes
@@ -499,6 +500,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
   ): Future[Workspace] = {
 
     assertBillingProfileCreationDate(profile)
+    validateWorkspaceRequest(request)
 
     val wsmConfig = multiCloudWorkspaceConfig.workspaceManager
     val workspaceId = UUID.randomUUID()
@@ -649,6 +651,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
                                 parentContext: RawlsRequestContext = ctx
   ): Future[Workspace] = {
     assertBillingProfileCreationDate(profile)
+    validateWorkspaceRequest(workspaceRequest)
 
     traceFutureWithParent("createMultiCloudWorkspace", parentContext)(s1 =>
       createMultiCloudWorkspaceInt(workspaceRequest, UUID.randomUUID(), profile, s1) andThen { case Success(_) =>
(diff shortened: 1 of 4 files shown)


blakery (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -649,6 +651,7 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
                                 parentContext: RawlsRequestContext = ctx
   ): Future[Workspace] = {
     assertBillingProfileCreationDate(profile)
+    validateWorkspaceRequest(workspaceRequest)
```
cahrens, line 654: I wish `createMultiCloudWorkspace` was called `createAzureWorkspace`, since we do plan to move to a world where a "multi-cloud"/WSM workspace could be a GCP one. It feels weird that this method assumes the workspace is Azure.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
```diff
@@ -733,6 +733,38 @@ class MultiCloudWorkspaceServiceCloneSpec
     result.errorReport.statusCode.value shouldBe StatusCodes.Forbidden
   }
 
+  it should "throw an exception for an Azure workspace if the workspaceRequest contains a nonempty authorizationDomain" in {
```
cahrens, line 736: Do we have test coverage of cloning and creating GCP workspaces with an authorizationDomain?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
marctalbott: We may not support the `authorizationDomain` field in GCP workspaces once they're moved over to WSM either as it would be best for auth domains on all WSM workspaces to be implemented with policy, so one could argue that this actually isn't Azure specific. Still, I'll look at renaming it to `createAzureWorkspace` for clarity and we can always change it back once GCP workspaces are in WSM

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala
marctalbott: Yeah, there are a few tests

- [Checking that the auth domain protected workspace bucket has secure logging after creating](https://github.com/broadinstitute/rawls/blob/f507389df6eefc8711f3394ce71ba57d086b8963/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala#L2251)
- [Checking that the auth domain protected workspace bucket has secure logging after cloning](https://github.com/broadinstitute/rawls/blob/f507389df6eefc8711f3394ce71ba57d086b8963/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala#L2654)
- [Checks that MultiCloudWorkspaceService passes GCP clone requests to WorkspaceService](https://github.com/broadinstitute/rawls/blob/f507389df6eefc8711f3394ce71ba57d086b8963/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceServiceCloneSpec.scala#L1092)
- [API unit test that includes an auth domain for a GCP workspace](https://github.com/broadinstitute/rawls/blob/f507389df6eefc8711f3394ce71ba57d086b8963/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala#L406)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
cahrens: I don't feel that strongly about it… `createMultiCloudWorkspace` has always assumed Azure. But every time I look at the method I have to think about it, as both the method name and the documentation lead one to believe that it is not cloud-specific.

cahrens (approved):

PR #2955, opened 2024-07-16
WOR-1776: don't fail on 404s from leo when deleting workspace resources

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1776

If all runtimes/apps for a workspace are already gone, or they are all in an inactive state, leo will throw a 404 when deleting all of that resource for a workspace.

If this happens, we should continue deleting the workspace.

---

**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunnerSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunnerSpec.scala
@@ -202,8 +202,8 @@ class WorkspaceDeletionRunnerSpec extends AnyFlatSpec with MockitoSugar with Mat
     )
     whenReady(
       runner.runStep(monitorRecord.copy(jobType = JobType.WorkspaceDeleteInit),
-        azureWorkspace,
-        RawlsRequestContext(null, null)
+                     azureWorkspace,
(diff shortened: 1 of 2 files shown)


aherbst-broad (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunnerSpec.scala
```diff
@@ -184,6 +184,32 @@ class WorkspaceDeletionRunnerSpec extends AnyFlatSpec with MockitoSugar with Mat
     }
   }
 
+  it should "return complete when leo returns a 404 for the workspace when deleting apps" in {
```
marctalbott: Am I misunderstanding this test? It appears to be verifying that the step returns `Incomplete` not `Complete`

marctalbott (approved):
Change looks good, but I had a clarifying question about the tests

core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/deletion/WorkspaceDeletionRunnerSpec.scala
blakery: I just named the test poorly (and somewhat robotically).
The actually important part is that it moves on to the next step. 
I'll fix the wording.

PR #2956, opened 2024-07-18
WOR-1797 background processes using rawls SA

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1797

Change background processes to use rawls service account context instead of a user specific context

see ticket for design doc

---

**PR checklist**


marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/workspace/runners/RawlsSAContextCreator.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.broadinstitute.dsde.rawls.dataaccess.SamDAO
+import org.broadinstitute.dsde.rawls.model.RawlsRequestContext
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait RawlsSAContextCreator {
+
+  val samDAO: SamDAO
+
+  def getRawlsSAContext()(implicit executionContext: ExecutionContext): Future[RawlsRequestContext] = Future(
```
blakery: Only noting this here, because you aren't afraid of making wider changes in favor of consistency - feel free to ignore this idea.

The error handling in the individual runners when calling `getUserCtx` was there because a failure there could represent a failure related to the job.
A failure to get the rawls SA context likely has nothing to do with an individual job.
It would probably make sense to remove all the error handling at the call sites (unhandled exceptions will simply result in the job being re-tried), and add some generic logging here.

blakery (approved):

PR #2962, opened 2024-07-23
WOR-1787 - break up WorkspaceService functionality

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1787

* move methods to new service
* move submissions to new service

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -107,17 +109,7 @@ trait RawlsApiService
     with VersionApiService
     with ServicePerimeterApiService {
 
-  val multiCloudWorkspaceServiceConstructor: RawlsRequestContext => MultiCloudWorkspaceService
```
blakery, line 110: These are not needed here - they're all inherited from the included traits.
If they were to be included here, they should have an override qualifier.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ApiServiceSpec.scala
```diff
@@ -414,16 +445,19 @@ trait ApiServiceSpec
     val sealedInstrumentedRoutes: Route = captureRequestMetrics {
       traceRequests { otelContext =>
         sealRoute(
-          adminRoutes(otelContext) ~
```
blakery, line 417: The route order is potentially significant.
These were rearranged so they were the same order as in `RawlsApiService`.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/MethodConfigApiServiceSpec.scala
```diff
@@ -263,14 +263,14 @@ class MethodConfigApiServiceSpec extends ApiServiceSpec with TestDriverComponent
                                                 Map(),
                                                 AgoraMethod("dsde", "method_doesnt_exist", 1)
       )
-
       Post(s"${testData.workspace.path}/methodconfigs", httpJson(newMethodConfig)) ~>
         sealRoute(services.methodConfigRoutes()) ~>
-        check {
+        check { testResult: RouteTestResult =>
```
blakery, line 268: Without `testResult: RouteTestResult =>`, this throws a 500, instead of the exception the actual service throws.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
```diff
@@ -329,7 +331,7 @@ object Boot extends IOApp with LazyLogging {
         HealthMonitor.CheckAll
       )
 
-      val statusServiceConstructor: () => StatusService = () => StatusService.constructor(healthMonitor)
+      val statusServiceConstructor: () => StatusService = () => StatusService.constructor(healthMonitor)()
```
cahrens, line 334: What is the impact of this change? It seems like it would change the type.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ApiServiceSpec.scala
```diff
@@ -355,25 +360,18 @@ trait ApiServiceSpec
...
-      execServiceBatchSize,
       workspaceManagerDAO,
       leonardoService,
-      methodConfigResolver,
       gcsDAO,
       samDAO,
       notificationDAO,
       userServiceConstructor,
-      genomicsServiceConstructor,
       maxActiveWorkflowsTotal,
```
cahrens: I don't think that `maxActiveWorkflowsTotal` and `maxActiveWorkflowsPerUser` are needed anymore, as the only usage (as far as I can tell) was moved to the SubmissionsService.

core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala
blakery: The parenthesis were implied before. The compiler was able to resolve it, but it's ambiguous and fragile.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ApiServiceSpec.scala
blakery: TY

core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
```diff
@@ -33,6 +33,7 @@ import org.broadinstitute.dsde.rawls.workspace.{
   MultiCloudWorkspaceAclManager,
   MultiCloudWorkspaceService,
   RawlsWorkspaceAclManager,
+  WorkspaceRepository,
```
cahrens: This import addition seems suspect. Perhaps it gets deleted in a later commit?

core/src/test/scala/org/broadinstitute/dsde/rawls/submissions/SubmissionsServiceSpec.scala
```diff
@@ -0,0 +1,471 @@
...
+  override def afterAll(): Unit = {
+    mockServer.stopServer
+    super.afterAll()
+  }
+
+  // noinspection TypeAnnotation,NameBooleanParameters,ConvertibleToMethodValue,UnitMethodIsParameterless
+  class TestApiService(dataSource: SlickDataSource, val user: RawlsUser)(implicit
+                                                                        override val executionContext: ExecutionContext
+  ) extends WorkspaceApiService
+    with MethodConfigApiService
```
cahrens: This file is a lot bigger than what was removed. Did you add more test coverage, or is it all setup code (that can't be removed from the original test)?

blakery: It's setup code.
Right now, there's some places where some of the state setup for individual tests is using the real code paths in other services, so it wasn't really safe to just go ripping things out (yet).

cahrens (approved):
It's great to see WorkspaceService get smaller!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -161,7 +137,7 @@ object WorkspaceService {
     } yield jobId
   }
 
-  private[workspace] def getTerminalStatusDate(submission: Submission, workflowID: Option[String]): Option[DateTime] = {
+  def getTerminalStatusDate(submission: Submission, workflowID: Option[String]): Option[DateTime] = {
```
marctalbott: Can this and `extractOperationIdsFromCromwellMetadata` be moved to `SubmissionsService` where they're used?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1324,34 +1283,39 @@ class WorkspaceService(protected val ctx: RawlsRequestContext,
       workspaceId <- loadV2WorkspaceId(workspaceName)
       results <- Future.traverse(input) {
         case WorkspaceCatalog(email, true) =>
-          toFutureTry(
-            samDAO.addUserToPolicy(SamResourceTypeNames.workspace,
-                                   workspaceId,
-                                   SamWorkspacePolicyNames.canCatalog,
-                                   email,
-                                   ctx
+          samDAO
```
marctalbott, line 1254: Why the changes here? Won't this not get to the first `map` if the initial call fails now?

core/src/main/scala/org/broadinstitute/dsde/rawls/methods/MethodConfigurationService.scala
```diff
@@ -0,0 +1,513 @@
+package org.broadinstitute.dsde.rawls.methods
```
marctalbott, line 1: Reviewed this under the assumption it was moved wholesale from `WorkspaceService`

core/src/main/scala/org/broadinstitute/dsde/rawls/submissions/SubmissionsService.scala
```diff
@@ -0,0 +1,957 @@
+package org.broadinstitute.dsde.rawls.submissions
```
marctalbott, line 1: Reviewed this under the assumption it was moved wholesale from `WorkspaceService`

marctalbott (approved):
Generally looks good, but a couple minor comments

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
blakery: You can chain as many `.map` calls onto a future as you like, and it won't actually throw the exception until you try to extract the value.
So as long as the `recover` is before the `<-` it's fine.
I think you might even be able to do it at the `yield`, since the `for` comprehension is just syntactic sugar for `.map`.

The actual motivation for this was just to get rid of the `toFutureTry` to eliminate an extra trait that was needlessly complicating things.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
cahrens: I think `getTerminalStatusDate` also was supposed to move.

PR #2980, opened 2024-08-05
add handler for SQL exceptions

blakery:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-654

Adds a top level exception handler to handle `SQLException`s:
 * Don't return the raw exception or include a stacktrace
 * Log original exception
 * Include an ID to look up the error in the logs

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
@@ -35,7 +35,7 @@ import org.broadinstitute.dsde.rawls.user.UserService
 import org.broadinstitute.dsde.rawls.workspace.{MultiCloudWorkspaceService, WorkspaceService}
 import org.broadinstitute.dsde.workbench.oauth2.OpenIDConnectConfiguration
 
-import java.sql.SQLTransactionRollbackException
+import java.sql.{SQLException, SQLTransactionRollbackException}
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -61,6 +61,14 @@ object RawlsApiService extends LazyLogging {
         )
         Sentry.captureException(rollback)
         complete(StatusCodes.InternalServerError -> ErrorReport(rollback))
+      case sql: SQLException =>
+        val sentryId = Sentry.captureException(sql)
+        logger.error(
+          s"Unhandled SQL exception with sentry id [$sentryId]: ${sql.getMessage} [${sql.getErrorCode} ${sql.getSQLState}] ${sql.getNextException}",
+          sql
+        )
+        val message = s"Internal server exception with sentry id: [${sentryId.toString}]"
+        complete(StatusCodes.InternalServerError -> ErrorReport(message))
       case wsmApiException: ApiException =>
         if (wsmApiException.getCode >= 500) {
           Sentry.captureException(wsmApiException)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
@@ -67,7 +67,7 @@ object RawlsApiService extends LazyLogging {
           s"Unhandled SQL exception with sentry id [$sentryId]: ${sql.getMessage} [${sql.getErrorCode} ${sql.getSQLState}] ${sql.getNextException}",
           sql
         )
-        val message = s"Internal server exception with sentry id: [${sentryId.toString}]"
+        val message = s"Internal server exception [sentryId=${sentryId.toString}]"
         complete(StatusCodes.InternalServerError -> ErrorReport(message))
       case wsmApiException: ApiException =>
         if (wsmApiException.getCode >= 500) {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -61,6 +61,14 @@ object RawlsApiService extends LazyLogging {
         )
         Sentry.captureException(rollback)
         complete(StatusCodes.InternalServerError -> ErrorReport(rollback))
+      case sql: SQLException =>
+        val sentryId = Sentry.captureException(sql)
+        logger.error(
+          s"Unhandled SQL exception with sentry id [$sentryId]: ${sql.getMessage} [${sql.getErrorCode} ${sql.getSQLState}] ${sql.getNextException}",
+          sql
+        )
+        val message = s"Internal server exception with sentry id: [${sentryId.toString}]"
```
blakery: If anyone has better wording for this, I'd love to hear it.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -61,6 +61,14 @@ object RawlsApiService extends LazyLogging {
         )
         Sentry.captureException(rollback)
         complete(StatusCodes.InternalServerError -> ErrorReport(rollback))
+      case sql: SQLException =>
+        val sentryId = Sentry.captureException(sql)
```
blakery, line 65: Running locally or in a BEE, this returns all zeros. Presumably in a real environment it will return the value with an actual UUID.

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiServiceSpec.scala
```diff
@@ -39,4 +57,72 @@ class RawlsApiServiceSpec extends ApiServiceSpec with VersionApiService {
...
+      routes ~>
+      check {
+        status shouldBe StatusCodes.InternalServerError
+        val report = responseAs[ErrorReport]
+        report.message should include("some message")
+      }
+  }
+
+  it should "expose a generic error message with an errorId for an SQLException" in
+    Using(mockStatic(classOf[Sentry])) { sentry =>
```
blakery, line 105: I finally figured out how to mock static methods in scala!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
```diff
@@ -61,6 +61,14 @@ object RawlsApiService extends LazyLogging {
         )
         Sentry.captureException(rollback)
         complete(StatusCodes.InternalServerError -> ErrorReport(rollback))
+      case sql: SQLException =>
```
blakery, line 64: I left the handling for `SQLTransactionRollbackException`s intact, but we should probably consider if we want this type of handling to happen for that as well, or even as a default for exception types we haven't explicitly enumerated here.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
cahrens: Will you be able to test this once it is merged to dev?

blakery: 🤦 
I was thinking I would, because it would have sentry hooked up properly.
But when testing in a BEE, I added specific code to throw the exception directly.

So now that I think about it, I don't know how, without intentionally breaking the database or something. I'm definetly open to ideas.

cahrens: I wonder if it would be possible to muck with a locally running Rawls (which uses  dev databases) to do something so that dev Rawls throws a SQLException.

Or maybe there is a way to just test what `Sentry.captureException()` returns in dev independently.

blakery: I guess I could just change something in the DB to an invalid value. I'm not sure that's an `SQLException` though - it might be some sort of slick exception.
FWIW, I did verify that when sentry is _not_ hooked up, `Sentry.captureException` still returns a non-null `SentryId`, which converts to a valid string. It isn't useful, but it at least isn't going to break anything.

cahrens: It seems like it might be worth attempting to change a DB value, since it would be a quick experiment.

marctalbott: You could try testing it locally with the correct DSN for dev and see if a real connection to Sentry yields a nonzero UUID. IIRC dev Rawls isn't actually hooked up to Sentry and may not yield useful information.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/RawlsApiService.scala
marctalbott:
```suggestion
        val message = s"Internal server exception [sentryId=${sentryId.toString}]"
```

This would be a bit more consistent with other exceptions we throw during various workspace operations, but I think what you have works perfectly well too

marctalbott (approved):

cahrens (approved):

PR #2981, opened 2024-08-07
Experimental GCP MC workspace creation support

aherbst-broad:
## Context
This is a proof-of-concept demonstrating the ability for Rawls to handle the creation and fetching of GCP MC workspaces (i.e., GCP workspaces managed by WSM).

## This PR
* Adds code in the MultiCloudWorkspaceService to create GCP workspaces in WSM. This behavior is flagged and requires that a) the parent billing project has a billing profile and b) the `system:mc_gcp` flag is passed in the attributes map in the creation request
  * Deletion seems to "just work", cloning does not. 
  * I've removed a check that the entity service was making to allow storage of tabular data in Rawls'.
  * I make no claims about workflows
  * There is no affordance in the UI to make these requests so this is an API only PoC at the moment
* Cleans up and adds documentation in related areas


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
```diff
@@ -247,6 +250,24 @@ class HttpWorkspaceManagerDAO(apiClientProvider: WorkspaceManagerApiClientProvid
       .createAzureStorageContainer(requestBody, workspaceId)
   }
 
+  override def createGcpStorageBucket(workspaceId: UUID,
+                                      storageBucketName: String,
+                                      location: String,
+                                      ctx: RawlsRequestContext
+  ): CreatedControlledGcpGcsBucket = {
+    val creationParams =
+      new GcpGcsBucketCreationParameters().name(storageBucketName).location(location)
```
rtitle, line 259: Will WSM buckets need to support any additional features that we use today, like autoclass or CORS?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -249,32 +256,36 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
-                createMultiCloudWorkspace(
-                  workspaceRequest,
-                  profileModel,
-                  s
-                ).map(workspace => (workspace, WorkspaceCloudPlatform.Azure))
-              }
+      workspaceOpt <-
+        if (isMcRequest(workspaceRequest, billingProfileOpt)) {
+          billingProfileOpt
+            .traverse { profileModel =>
```
rtitle, line 270: I haven't compiled but -- does this give you an `Option[Option[..]]`? This may be a use case for `.flatTraverse`

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
```diff
@@ -49,7 +49,12 @@ class EntityManager(providerBuilders: Set[EntityProviderBuilder[_ <: EntityProvi
 
   def resolveProvider(requestArguments: EntityRequestArguments): Try[EntityProvider] = {
 
-    if (!WorkspaceType.RawlsWorkspace.equals(requestArguments.workspace.workspaceType)) {
+    // Azure workspaces use a WDS application in the data plane, Rawls entity service requests are not supported.
+    if (
+      WorkspaceType.McWorkspace.eq(
```
rtitle, line 54: Thinking out loud, wondering what might be a good way to represent the "flavor" of workspace in the Rawls model. I guess we have 3 cases that we're supporting, right?
1. Rawls workspace - GCP
2. MC workspace - GCP
3. MC workspace - Azure

cahrens: It's seems like there's a 4th one, the Rawls w/"Rawls Stage" MC Workspace.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -331,6 +359,20 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
       } yield profileModel
     }
 
+  /**
```
rtitle, line 357: Thanks for adding scaladoc comments, I want to do the same in many places in Leo

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -844,6 +909,61 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+    for {
+      maybeWorkspace <- workspaceRepository.getWorkspace(workspaceId)
+      workspace <- Future {
+        maybeWorkspace.getOrElse(throw new RuntimeException("Workspace not found"))
+      }
+      wsmWorkspace <- Future {
+        workspaceManagerDAO.getWorkspace(workspaceId, ctx)
+      }
+      _ <- Future {
+        workspaceRepository.updateBucketName(workspace.workspaceIdAsUUID,
```
rtitle, line 922: It looks like these 2 `workspaceRepository` methods already return `Future` so don't need to be wrapped in `Future`.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
```diff
@@ -844,6 +909,61 @@ class MultiCloudWorkspaceService(override val ctx: RawlsRequestContext,
...
+          workspaceManagerDAO
+            .createAzureStorageContainer(
+              workspaceId,
+              MultiCloudWorkspaceService.getStorageContainerName(workspaceId),
+              ctx
+            )
+            .getResourceId
+        )
+      case CloudPlatform.GCP =>
+        Future.successful(
```
rtitle: Shouldn't use `Future.successful` for actual non-pure things that may fail.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
rtitle: *nod *

I wonder if it's worth explicitly enumerating all the cases, instead of relying on combinations of certain fields to distinguish them.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/HttpWorkspaceManagerDAO.scala
aherbst-broad: Yes, see this [code](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L192). I would prefer to factor that into WSM if possible rather than have Rawls do that out of band.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityManager.scala
aherbst-broad: Yeah, it's admittedly awkward. I'll see about evolving the data model.

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/MultiCloudWorkspaceService.scala
aherbst-broad: Check out https://github.com/broadinstitute/rawls/pull/2981/commits/5a104207b93b51dfbf2d97dd95e094c954f29f1b

PR #2983, opened 2024-08-08
[WOR-1781][WOR-1782] APIs to manage workspace settings

marctalbott:
Ticket: [WOR-1781](https://broadworkbench.atlassian.net/browse/WOR-1781), [WOR-1782](https://broadworkbench.atlassian.net/browse/WOR-1782)
* Add workspace settings APIs with support for setting lifecycle rules on GCP workspace buckets

---

**PR checklist**

[WOR-1781]: https://broadworkbench.atlassian.net/browse/WOR-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WOR-1782]: https://broadworkbench.atlassian.net/browse/WOR-1782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aednichols (commented):
Nice, this looks awesome.

I'm currently working on making Rawls instruct Cromwell to segregate workflow files into `submissions/intermediates` and `submissions/final-outputs`.

Do you think it makes sense to enable that behavior with a setting instead of forcing it on everyone?

```
{
    "type": "SegregateFinalOutputs",
    "config": {
        "enable_segregation": true
    }
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -99,4 +102,55 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
       } yield newWorkspace
     }
 
+  def getWorkspaceSettings(workspaceId: UUID): Future[List[WorkspaceSetting]] =
```
aherbst-broad: All of these methods could use comments, esp. to explain the semantics around only returning Applied, the expected calling pattern etc.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,115 @@ class WorkspaceService(
       }
     } yield {}
 
+  def getWorkspaceSettings(workspaceName: WorkspaceName): Future[List[WorkspaceSetting]] =
```
aherbst-broad: What do you think about placing these methods in a more focused WorkspaceSettingsService class or some such thing -- this service is already very large.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,120 @@ class WorkspaceService(
...
+    def computeNewSettings(workspace: Workspace,
+                           requestedSettings: List[WorkspaceSetting],
+                           existingSettings: List[WorkspaceSetting]
+    ): List[WorkspaceSetting] =
+      if (existingSettings.isEmpty) {
+        requestedSettings
+      } else {
+        val requestedSettingsMap = requestedSettings.map(s => s.`type` -> s.config).toMap
+        existingSettings.map { case WorkspaceSetting(settingType, _) =>
+          WorkspaceSetting(settingType, requestedSettingsMap.getOrElse(settingType, settingType.defaultConfig()))
```
aherbst-broad: Does this mean a previously configured setting will assume the default on any new settings request, regardless if it's included or not? I think it should retain the existing setting, if that's the case.

marctalbott:
@aednichols You certainly could do that with settings! You would know better than me whether it _should_ be forced on everyone, but if you expect that users may prefer different segregation behavior then adding a setting type may be the best way to go about that.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,120 @@ class WorkspaceService(
...
+          }
+
+          for {
+            _ <- gcsDAO.setBucketLifecycle(workspace.bucketName, googleRules)
+            _ <- workspaceRepository.markWorkspaceSettingApplied(workspace.workspaceIdAsUUID, setting.`type`)
+          } yield None
+        case _ => throw new RawlsException("unsupported workspace setting")
+      }).recoverWith { case e =>
+        workspaceRepository
+          .removePendingSetting(workspace.workspaceIdAsUUID, setting.`type`)
```
aherbst-broad: Can `e` be logged somewhere? It looks like it's being swallowed which will make diagnosing issues here difficult. Agreed it should not be surfaced in the error report.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,120 @@ class WorkspaceService(
...
+                  "unsupported lifecycle action"
+                ) // validated earlier but needed for completeness
+            }
+
+            new LifecycleRule(action, conditionBuilder.build())
+          }
+
+          for {
+            _ <- gcsDAO.setBucketLifecycle(workspace.bucketName, googleRules)
+            _ <- workspaceRepository.markWorkspaceSettingApplied(workspace.workspaceIdAsUUID, setting.`type`)
```
aherbst-broad: Making sure I understand: not only will this shift the newly inserted setting record from PENDING -> APPLIED, it will mark any previous settings of this type for the workspace from APPLIED -> DELETED. Is that right?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20240723_workspace_settings.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
aherbst-broad, line 1: I think this should also capture who initiated the change, ideally the sam/terra UID and not an email.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,120 @@ class WorkspaceService(
...
+      * number of setting types and it's easier to reason about this way.
+      */
+    def applySetting(workspace: Workspace,
+                     setting: WorkspaceSetting
+    ): Future[Option[(WorkspaceSettingType, ErrorReport)]] =
+      (setting match {
+        case WorkspaceSetting(WorkspaceSettingTypes.GcpBucketLifecycle, GcpBucketLifecycleConfig(rules)) =>
+          val googleRules = rules.map { rule =>
+            val conditionBuilder =
+              LifecycleCondition.newBuilder().setMatchesPrefix(rule.conditions.matchesPrefix.toList.asJava)
```
aherbst-broad: Is it possible rule, or rule.conditions or rule.conditions.matchesPrefix are null here?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2001,6 +2005,120 @@ class WorkspaceService(
...
+        case WorkspaceSetting(settingType @ WorkspaceSettingTypes.GcpBucketLifecycle,
+                              GcpBucketLifecycleConfig(rules)
+            ) =>
+          rules.flatMap { rule =>
+            val actionValidation = rule.action.`type` match {
+              case actionType if actionType.equals("Delete") => None
+              case actionType => Some(validationErrorReport(settingType, s"unsupported lifecycle action $actionType"))
+            }
+            val ageValidation = rule.conditions.age.collect {
+              case age if age < 0 =>
```
aherbst-broad: Should this be <= 0?

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -561,6 +568,36 @@ object WorkspaceState {
...
+  }
+}
+
+sealed trait WorkspaceSettingConfig
+object WorkspaceSettingConfig {
+  case class GcpBucketLifecycleConfig(rules: List[GcpBucketLifecycleRule]) extends WorkspaceSettingConfig
+
+  case class GcpBucketLifecycleRule(action: GcpBucketLifecycleAction, conditions: GcpBucketLifecycleCondition)
+  case class GcpBucketLifecycleAction(`type`: String)
+  case class GcpBucketLifecycleCondition(matchesPrefix: Set[String], age: Option[Int])
```
aherbst-broad: Can age be more strongly typed instead of an Int? I worry about mixing up seconds/hours/days etc.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -561,6 +568,36 @@ object WorkspaceState {
...
+    override def defaultConfig(): WorkspaceSettingConfig = GcpBucketLifecycleConfig(List.empty)
+  }
+}
+
+sealed trait WorkspaceSettingConfig
+object WorkspaceSettingConfig {
+  case class GcpBucketLifecycleConfig(rules: List[GcpBucketLifecycleRule]) extends WorkspaceSettingConfig
+
+  case class GcpBucketLifecycleRule(action: GcpBucketLifecycleAction, conditions: GcpBucketLifecycleCondition)
+  case class GcpBucketLifecycleAction(`type`: String)
```
aherbst-broad: Nit: I would prefer not to have to backtick every usage of this param, but I am lazy.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5818,6 +5896,62 @@ components:
...
+        - type
+      type: object
+      properties:
+        "type":
+          type: string
+          description: The type of the lifecycle action
+          enum:
+            - Delete
+    GcpBucketLifecycleCondition:
+      type: object
```
aherbst-broad, line 5963: Are any of these properties required?

aherbst-broad (commented):
Looking good! Can you describe your thinking around handling concurrent requests? I know we talked a bit about a constraint etc. but I don't see evidence of that here -- it'd be good to go in knowing how this will behave in the face of simultaneous requests to mutate settings.

marctalbott:
> Looking good! Can you describe your thinking around handling concurrent requests? I know we talked a bit about a constraint etc. but I don't see evidence of that here -- it'd be good to go in knowing how this will behave in the face of simultaneous requests to mutate settings.

@aherbst-broad Certainly. `WorkspaceRepository` checks for existing pending settings before saving any new settings at the beginning of the request ([source](https://github.com/broadinstitute/rawls/pull/2983/files#diff-c09b3f79a063ae9bb4a5006d4dbaad90ff59072e31af673a3faffb9ba7a118aeR117)). If there are any pending settings, it will throw an exception. Your question did get me thinking that we could still be at risk of phantom reads and end up with multiple `Pending` settings, so perhaps we should change the isolation level to Serializable. I'm not super familiar with isolation levels, but does that sound reasonable to you?

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Either can be omitted, but we _do_ need at least one now that I think about it which isn't accounted for. I'll change that.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: I considered it, but I ultimately didn't as I felt that managing settings belonged in the WorkspaceService as much as anything else. Still, I'm not opposed to it and it would certainly make changing it easier given how much Intellij struggles with the WorkspaceService

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: No, you can set age to 0 to delete immediately. We actually do it when deleting a bucket - [deleteBucket](https://github.com/broadinstitute/rawls/blob/7c38ae3a7b478df692127a76c9ca06f6a604f5f1/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala#L313)

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: Yeah, it does. My thinking was that this would be a simple overwrite -- any settings provided become the new settings on the workspace and old settings are reverted to the default if not respecified. It's pretty heavy-handed and requires the caller to do more work, so I'm definitely not attached to it.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: No, we constructed `setting` here by parsing the json and none of those fields are optional. We'd throw a `NoSuchElementException` if any of them were null. I'll add some tests that assert that similar to [this test](https://github.com/broadinstitute/rawls/pull/2983/files/5178b1e84d12e115f6bbb3c5cd80922ffc6f67da#diff-1f0ac22ea82efa99f74f486a4f2b602cf2b5109a86d4aa8f3748c2e7ddb9b1bdR784) that checks for the presence of `config`

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: That's correct

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: Absolutely, I'll log it

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
marctalbott: That's fair. I avoided it initially with `WorkspaceSetting` because I'm also not a fan, but then I wanted this to mimic the Google lifecycle rules format as closely as possible so I ended up caving and using backticks for both since they use `type`.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
marctalbott: I could add `type Days = Int` to the object so this would read as:

```
case class GcpBucketLifecycleCondition(matchesPrefix: Set[String], age: Option[Days])
```

Everything else would remain the same as it's functionally unchanged, but it would make it clearer to future devs

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -47,7 +48,10 @@ import spray.json.DefaultJsonProtocol._
 import spray.json._
 import org.broadinstitute.dsde.rawls.metrics.MetricsHelper
 import cats.effect.unsafe.implicits.global
+import com.google.cloud.storage.BucketInfo.LifecycleRule
+import com.google.cloud.storage.BucketInfo.LifecycleRule.{LifecycleAction, LifecycleCondition}
 import org.broadinstitute.dsde.rawls.billing.BillingRepository
+import org.broadinstitute.dsde.rawls.model.WorkspaceSettingConfig._
```
blakery: Are these imports actually needed?
I don't see any code changes, so I'm wondering of maybe they're left over from changes that were moved elsewhere?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -3,18 +3,23 @@ package org.broadinstitute.dsde.rawls.workspace
 import akka.http.scaladsl.model.StatusCodes
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 import org.broadinstitute.dsde.rawls.dataaccess.SlickDataSource
+import org.broadinstitute.dsde.rawls.dataaccess.slick.WorkspaceSettingRecord
```
blakery: Nit: unused imports from when the settings were a part of this class

blakery (approved):
Looks good with the repo split out

aherbst-broad (approved):

PR #2985, opened 2024-08-10
AJ-1027: submission monitor uses same methods as APIs to save entities

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-1027

### Description

The `saveEntityPatch()` method, and its helper `applyEntityPatch`, were only used to write submission outputs via `SubmissionMonitorActor`. This means the code path for updating entities during workflow outputs was quite different than the code path for updating entities from a user's API call.

This PR aligns those code paths. It removes `saveEntityPatch()` entirely, replacing it with a call to `entityQuery.save()`, which is the same path used by the user-facing API.

Associated to https://broadworkbench.atlassian.net/browse/AJ-1027.  A highlight of the Slack thread mentioned in that ticket is that the bug in question appears when writing workflow outputs, but does not appear when using the API. The bug has been impossible to reproduce (but users keep reporting it, infrequently). There were tests for exactly this use case (which are deleted in this PR for being obsolete). So, this PR takes a different route - instead of trying to find and resolve the bug, it aligns the code paths.

### Testing

All test suites are passing.

My hesitation in merging such a major refactor is changing the performance characteristics of submissions. Will this change add overhead when saving outputs? Conversely, will it improve performance? Is the code convergence worth a minor performance hit?

We can certainly keep an eye on the `handleOutputs` [trace](https://console.cloud.google.com/traces/list?referrer=search&project=broad-dsde-prod&pageState=(%22traceFilter%22:(%22chips%22:%22%255B%257B_22k_22_3A_22SpanName_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22handleOutputs_5C_22_22_2C_22i_22_3A_22span_22%257D%255D%22),%22traceIntervalPicker%22:(%22groupValue%22:%22P1D%22,%22customValue%22:null))) for any performance changes. For the most targeted analysis, we can watch the `saveEntityPatchSequence` trace underneath the `handleOutputs` trace; `saveEntityPatchSequence` contains all the changes in this PR.

---

**PR checklist**


mspector (approved):
If the performance hit is minor, it does still feel worth it to me. Merging this and keeping an eye on traces seems like a reasonable way forward. 

I would imagine that converging these code paths would make later performance improvements easier (or at least more broadly applicable)

dvoet (approved):

salonishah11 (approved):
I haven't tested this but it makes sense to use the same path to update entities that is used by the API.

PR #2993, opened 2024-08-13
WX-1711 local Docker build fixes, streamlining 

aednichols:
Rawls CI migrated exclusively to GCR in April 2024, but the situation for local builds was confusing and not fully updated.

CI uses `rawls-build-tag-publish-and-run-tests` now I believe there is minimal concern about changes intended for local impacting CI.

- Remove extra image tag `DOCKERTAG_SAFE_NAME` based on branch name that we were assigning in addition to git hash
  - Its `sed` invocation is not portable to BSD
- Remove all references to Docker Hub and provided our production Rawls GCR as default
- Remove support for running tests before building an image
  - People run tests from SBT; and the `SKIP_TESTS` environment variable was non-discoverable, causing a lot of waiting around
- Remove hardcoded reliance on `origin/<branch>` that prevented the build from working if the branch was not pushed
- Remove `rawls-tests` image that we stopped updating in April
- In accordance with deprecation warning, switched from `gcloud docker push` to direct `docker push`
- Updated README
  - But the basic command to build an image locally is unchanged at `./docker/build.sh jar -d build`
- Fix Bash `if [ $DOCKER_CMD="push" ]` that always pushed the image regardless of `DOCKER_CMD` value
- Delete`docker/install.sh` which became redundant with the removal of test support
- Add suggestion to Dockerfile on how to further nuke confusing bash

Tested on macOS with Docker Desktop and Fedora Silverblue 40 with Podman aliased to `docker`


salonishah11 (approved):

davidangb (approved):
did not test it myself, thank you for the cleanup!

PR #2999, opened 2024-08-19
TSPS-276 update openapi yaml

jsotobroad:
Ticket: https://broadworkbench.atlassian.net/browse/TSPS-276

* update default value of `NamedDataRepoSnapshot` to be null so it is compatible with openapi generator for java code. otherwise you get something like Object = {} as the default and im not sure thats what it should be anyways.  Happy to chat about a better way of doing it, this just worked for us.
* update `SubmissionValidationValue` value field to be object instead of String because it can be a string or an array so this will just not parse correctly for array values.  realistically we should be using [oneOf](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/) for this but again doesnt seem [supported yet](https://github.com/OpenAPITools/openapi-generator/issues?q=is%3Aissue+is%3Aopen+java+oneof) for java client generation.

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4548,8 +4548,7 @@ components:
           default: ""
         properties:
           type: object
-          description: A map of key-value pairs to associate with the snapshot. Optional, defaults to an empty map if not specified.
-          default: {}
+          description: A map of key-value pairs to associate with the snapshot.
         cloningInstructions:
           type: string
           default: COPY_NOTHING
@@ -5035,7 +5034,7 @@ components:
       type: object
       properties:
         value:
-          type: string
+          type: object
           description: the value of the input expression for the specified entity
         error:
           type: string
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -4548,7 +4548,7 @@ components:
           default: ""
         properties:
           type: object
-          description: A map of key-value pairs to associate with the snapshot.
+          description: A map of key-value pairs to associate with the snapshot. Optional, defaults to an empty map if not specified.
         cloningInstructions:
           type: string
           default: COPY_NOTHING


davidangb (approved):
lgtm, would love to get a thumb from Workflows w/r/t the changes to `SubmissionValidationValue`

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4549,7 +4549,6 @@ components:
         properties:
           type: object
           description: A map of key-value pairs to associate with the snapshot. Optional, defaults to an empty map if not specified.
-          default: {}
```
jsotobroad, line 4552: @aherbst-broad  i believe you added this [here](https://github.com/broadinstitute/rawls/commit/6f95120cc5cabc3cdeba1e2bba5c90d54acbaafc#diff-6e80a735b01ebcbdc253d755a6078cc1af5a585a6bd04b92b6455a9469988dffR4417). Does this always need to exist in the payload no matter what or can it be `null` by default?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5035,7 +5034,7 @@ components:
       type: object
       properties:
         value:
-          type: string
+          type: object
```
jsotobroad, line 5037: @jgainerdewar you may be the best person to answer this even though this was in here [3 name changes + 9 years ago](https://github.com/broadinstitute/rawls/commit/cff3543cc983e89136126aa90a270f9cd6f7160c) 😂.  Maybe at the time the only expression was in fact a string and at some point arrays were added and this wasnt updated to match?

core/src/main/resources/swagger/api-docs.yaml
aherbst-broad: Should be fine, it's an option in the [model class](https://github.com/broadinstitute/rawls/blob/develop/model/src/main/scala/org/broadinstitute/dsde/rawls/model/DataReferenceModel.scala#L17)

core/src/main/resources/swagger/api-docs.yaml
jgainerdewar: LGTM assuming you can see in API responses that we aren't string-ifying more complex types.

jgainerdewar (approved):

PR #3001, opened 2024-08-20
[WOR-1197] Remove dupes and upgrade liquibase

aherbst-broad:
This upgrades liquibase to latest. A previous attempt failed out due to checksum validation failures at runtime. I pulled the changeset checksums from the `DATABASECHANGELOG` table in dev to a Bee. Running an upgraded liquibase reproduced the checksum errors. There are two issues:
  a) string interpolation within the changeset causing the checksum to be different than when the changeset originally ran or
  b) re-used id values between changesets

I've added a 'validCheckSum' element of '1:any' to the failing changesets, assuming they will not be changed in the future--all of the offending changesets are quite old and unlikely to change. This will prevent us from having to re-compute checksums any time there is a new algorithm version.

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
@@ -88,10 +88,6 @@
     <include file="changesets/20201106_workspace_google_project_columns.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20201113_add_billing_account_workspace.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20210430_add_workspace_billing_account_error_message.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20210524_add_memory_retry_multiplier.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20210609_add_workspace_attr_temp_table_procedure.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20210609_change_workspace_attr_temp_owner_id.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20210510_add_billing_project_export_columns.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20210915_create_entity_cache_table.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20210916_clone_workspace_file_transfer.xml" relativeToChangelogFile="true"/>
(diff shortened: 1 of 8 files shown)


project/Dependencies.scala
```diff
@@ -81,7 +81,7 @@ object Dependencies {
   // protobuf is only need to use the MySQL X DevAPI which we don't. exclude it to avoid interference with Google client libraries
   val mysqlConnector: ModuleID =  "com.mysql"                         % "mysql-connector-j"  % "9.0.0" exclude("com.google.protobuf", "protobuf-java")
   // Update warning for liquibase-core: Here be dragons! See https://broadworkbench.atlassian.net/browse/WOR-1197
-  val liquibaseCore: ModuleID =   "org.liquibase"                 % "liquibase-core"        % "4.17.2" // scala-steward:off
+  val liquibaseCore: ModuleID =   "org.liquibase"                 % "liquibase-core"        % "4.29.1"
```
davidangb, line 83: nit: remove the comment above this line

edited to add: I know this PR is still in draft, I was curious about the changes!

aherbst-broad: We attempted to update liquibase last year and the checksum computation failed out. This is reintroducing that change. Against the checksums stored in the DEV DB we get the same error as last year (against a BEE), so I'm formulating a plan of attack to get this out the door.

aherbst-broad: @davidangb see the PR description for my conclusions.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changelog.xml
```diff
@@ -88,10 +88,6 @@
     <include file="changesets/20201106_workspace_google_project_columns.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20201113_add_billing_account_workspace.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20210430_add_workspace_billing_account_error_message.xml" relativeToChangelogFile="true"/>
-    <include file="changesets/20210524_add_memory_retry_multiplier.xml" relativeToChangelogFile="true"/>
```
aherbst-broad, line 91: These were duplicates.

cahrens: Yeah, I recall the duplicates.

cahrens (approved):

blakery (approved):

PR #3009, opened 2024-08-28
[WOR-1808] Add soft delete policy to workspace settings

trholdridge:
Ticket: [https://broadworkbench.atlassian.net/browse/WOR-1808](https://broadworkbench.atlassian.net/browse/WOR-1808)

Would recommend looking through this by commit--the first commit essentially accomplishes soft delete policy setting, following the template of the lifecycle rules implementation. See comments for additional info about the changes in subsequent commits.

- Added support for setting soft delete policy in workspace settings
- Changed some types around workspace settings to reflect the correspondence between setting type and config type
- Removed the `defaultConfig`s specified in `WorkspaceModel`

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
@@ -21,13 +21,13 @@ import org.broadinstitute.dsde.rawls.model.{
   Workspace,
   WorkspaceName,
   WorkspaceSetting,
-  WorkspaceSettingResponse,
-  WorkspaceSettingTypes
+  WorkspaceSettingResponse
 }
 import org.broadinstitute.dsde.rawls.model.WorkspaceSettingTypes.WorkspaceSettingType
 import org.broadinstitute.dsde.rawls.util.WorkspaceSupport
 
 import java.time.Duration
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 
@@ -84,12 +84,16 @@ class WorkspaceSettingService(protected val ctx: RawlsRequestContext,
               actionValidation ++ ageValidation ++ atLeastOneConditionValidation
             }
           case GcpBucketSoftDeleteSetting(GcpBucketSoftDeleteConfig(retentionDuration)) =>
-            val retentionDurationValidation = retentionDuration match {
-              case retentionDuration if retentionDuration < 0 || retentionDuration > 7_776_000 =>
-                Some(validationErrorReport(setting.settingType, "retention duration must be from 0 to 90 days"))
+            retentionDuration match {
+              case duration if (duration < 7.days.toSeconds || duration > 90.days.toSeconds) && duration != 0 =>
+                Some(
+                  validationErrorReport(
+                    setting.settingType,
+                    "retention duration must be from 7 to 90 days, or 0 to disable soft delete retention"
+                  )
+                )
               case _ => None
             }
-            retentionDurationValidation
         }
       }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
@@ -85,8 +85,8 @@ class WorkspaceSettingService(protected val ctx: RawlsRequestContext,
             }
           case GcpBucketSoftDeleteSetting(GcpBucketSoftDeleteConfig(retentionDuration)) =>
             val retentionDurationValidation = retentionDuration match {
-              case retentionDuration if retentionDuration < 0 || retentionDuration > 7776000 =>
-                Some(validationErrorReport(setting.settingType, "retention duration must be between 0 and 90 days"))
+              case retentionDuration if retentionDuration < 0 || retentionDuration > 7_776_000 =>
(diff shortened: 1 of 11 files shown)


model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -573,7 +573,13 @@ object WorkspaceState {
...
 }
 
-case class WorkspaceSetting(settingType: WorkspaceSettingType, config: WorkspaceSettingConfig)
+sealed abstract class WorkspaceSetting(val settingType: WorkspaceSettingType, val config: WorkspaceSettingConfig)
+
+case class GcpBucketLifecycleSetting(override val config: GcpBucketLifecycleConfig)
+    extends WorkspaceSetting(settingType = WorkspaceSettingTypes.GcpBucketLifecycle, config)
+
+case class GcpBucketSoftDeleteSetting(override val config: GcpBucketSoftDeleteConfig)
+    extends WorkspaceSetting(settingType = WorkspaceSettingTypes.GcpBucketSoftDelete, config)
```
trholdridge, line 582: I made this change because previously, the way the types were set up allowed for a `WorkspaceSetting` with a `settingType` of `GcpBucketLifecycle` and a `config` of `GcpBucketSoftDeleteConfig` or vice versa. This wasn't reflective of how we want the types to match up. The new sealed class with two subclasses means that any `WorkspaceSetting` will have to either deal with lifecycles or soft delete.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -585,7 +585,6 @@ object WorkspaceSettingTypes {
   sealed trait WorkspaceSettingType extends RawlsEnumeration[WorkspaceSettingType] {
     override def toString: String = getClass.getSimpleName.stripSuffix("$")
     override def withName(name: String): WorkspaceSettingType = WorkspaceSettingTypes.withName(name)
-    def defaultConfig(): WorkspaceSettingConfig
```
trholdridge, line 577: Marc mentioned that this was a holdover from a previous refactor of the workspace settings, and since it's only used in tests now I took it out. Having it there added some confusion about what it was being used for/where this default config comes into play.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
```diff
@@ -53,32 +57,40 @@ class WorkspaceSettingService(protected val ctx: RawlsRequestContext,
...
+                    )
                   )
-                )
+                case _ => None
+              }
+              actionValidation ++ ageValidation ++ atLeastOneConditionValidation
+            }
+          case GcpBucketSoftDeleteSetting(GcpBucketSoftDeleteConfig(retentionDuration)) =>
+            val retentionDurationValidation = retentionDuration match {
+              case retentionDuration if retentionDuration < 0 || retentionDuration > 7_776_000 =>
```
aherbst-broad: [Magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) alert, consider extracting this to a named constant.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingServiceUnitTests.scala
```diff
@@ -548,4 +580,38 @@ class WorkspaceSettingServiceUnitTests extends AnyFlatSpec with MockitoTestUtils
...
+    )
+
+    val service = workspaceSettingServiceConstructor()
+
+    val exception = intercept[RawlsExceptionWithErrorReport] {
+      Await.result(service.setWorkspaceSettings(workspace.toWorkspaceName, List(longDurationSetting)), Duration.Inf)
+    }
+    exception.errorReport.statusCode shouldBe Some(StatusCodes.BadRequest)
+    exception.errorReport.message should include("Invalid settings requested.")
+    exception.errorReport.causes should contain theSameElementsAs List(
```
aherbst-broad: Food for thought: It's good to avoid matching on strings in tests when determining if a particular error was thrown since this leads to fragility down the line (i.e, what happens if someone fixes a typo in an error message?). It's usually better to craft a custom, semantically clear exception type that exposes the needed properties against which a unit test may make assertions.

Feel free to investigate that approach, or leave this as-is, I don't think it's a deal breaker here.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -595,6 +606,9 @@ object WorkspaceSettingConfig {
   case class GcpBucketLifecycleRule(action: GcpBucketLifecycleAction, conditions: GcpBucketLifecycleCondition)
   case class GcpBucketLifecycleAction(actionType: String)
   case class GcpBucketLifecycleCondition(matchesPrefix: Option[Set[String]], age: Option[Days])
+
+  type Seconds = Long
```
aherbst-broad, line 610: It's a little odd to have a broad type alias like this in a narrow class. That said, I like the expressiveness.

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceSettingService.scala
marctalbott: You could also use scala durations to write `90.days.toSeconds` (in addition to extracting to a named constant)

marctalbott (approved):

marctalbott (approved):

PR #3017, opened 2024-09-05
WX-1840 Pause intermediates/final separation

aednichols:
- Primarly due to breaking side effects on Theiagen, who pass strings to `File` final outputs to snapshot objects from public buckets into their workflow outputs
- Also was not great that we showed no-context links that people followed to soft-deleted files


jgainerdewar (approved):
LGTM if it looks good to the tests.

salonishah11 (approved):

PR #3018, opened 2024-09-08
TSPS-303 update methodRepoMethod swagger element 

jsotobroad:
Ticket: related to https://broadworkbench.atlassian.net/browse/TSPS-303 where we are trying to query for the method config of one of our methods.

Im not sure if this is the best solution but mostly just wanted to get this in front of some 👀 .  The docs even state that the value of `methodVersion` can be a string or an int but the type is an int so this fails when trying to get a docstore method config.
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.lang.Integer` from String "0.0.94": not a valid `java.lang.Integer` value
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 678] (through reference chain: bio.terra.rawls.model.MethodConfiguration["methodRepoMethod"]->bio.terra.rawls.model.MethodRepoMethod["methodVersion"])
```
---

**PR checklist**

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -5637,7 +5637,7 @@ components:
           type: string
           description: The name of the method in Agora
         methodVersion:
-          type: integer
+          type: object
           description: The integer method version in Agora (or a string method version
             for Dockstore)
         methodPath:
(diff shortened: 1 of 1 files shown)


jgainerdewar (approved):
LGTM!

PR #3034, opened 2024-09-16
[WOR-1794] Enforce submission cost cap

marctalbott:
Ticket: [WOR-1794](https://broadworkbench.atlassian.net/browse/WOR-1794)
* Add column to `WORKFLOW` table that stores the last fetched cost for the workflow if the workflow's submission has a cost cap set.
* When updating submission status, check that the current cost of all workflows is under the cost cap. Abort the submission if the cost cap has been exceeded.
* Submission monitor is functionally unchanged if a cost cap has not been set.
---

**PR checklist**

[WOR-1794]: https://broadworkbench.atlassian.net/browse/WOR-1794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceDAO.scala
@@ -25,10 +25,7 @@ trait ExecutionServiceDAO extends ErrorReportable {
   def abort(id: String, userInfo: UserInfo): Future[Try[ExecutionServiceStatus]]
   def getLabels(id: String, userInfo: UserInfo): Future[ExecutionServiceLabelResponse]
   def patchLabels(id: String, userInfo: UserInfo, labels: Map[String, String]): Future[ExecutionServiceLabelResponse]
-  def getCost(id: String,
-              workflowCostBreakdownParams: Option[WorkflowCostBreakdownParams],
-              userInfo: UserInfo
-  ): Future[WorkflowCostBreakdown]
+  def getCost(id: String, userInfo: UserInfo): Future[WorkflowCostBreakdown]
 
   // get the version of the execution service itself
   def version(): Future[ExecutionServiceVersion]
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceDAO.scala
@@ -25,6 +25,10 @@ trait ExecutionServiceDAO extends ErrorReportable {
   def abort(id: String, userInfo: UserInfo): Future[Try[ExecutionServiceStatus]]
   def getLabels(id: String, userInfo: UserInfo): Future[ExecutionServiceLabelResponse]
   def patchLabels(id: String, userInfo: UserInfo, labels: Map[String, String]): Future[ExecutionServiceLabelResponse]
+  def getCost(id: String,
+              workflowCostBreakdownParams: Option[WorkflowCostBreakdownParams],
+              userInfo: UserInfo
+  ): Future[WorkflowCostBreakdown]
 
   // get the version of the execution service itself
   def version(): Future[ExecutionServiceVersion]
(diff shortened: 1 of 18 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -577,10 +606,14 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
           logger.debug(s"submission $submissionId terminating to status $newStatus")
           dataAccess.submissionQuery.updateStatus(submissionId, newStatus)
         } map (_ => true)
+      } else if (costCapThreshold.isDefined && costCapThreshold.get <= workflowRecs.flatMap(_.cost).sum) {
+        logger.info(s"Submission $submissionId exceeded its cost cap and will be aborted.")
```
aherbst-broad: What do you think about including the cost cap and current cost for diagnostic purposes?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
```diff
@@ -131,6 +131,22 @@ class HttpExecutionServiceDAO(executionServiceURL: String, override val workbenc
...
+    val uri = Uri(executionServiceURL + s"/api/workflows/v1/$id/cost")
+    workflowCostBreakdownParams
+      .map(p =>
+        Map("includeSubworkflowBreakdown" -> p.includeSubworkflowBreakdown.toString,
+            "includeTaskBreakdown" -> p.includeTaskBreakdown.toString
+        )
+      )
+      .map(p => uri.withQuery(Query(p)))
+
+    retry(when5xx)(() => pipeline[WorkflowCostBreakdown](userInfo) apply Get(uri))
```
aherbst-broad: Will we get logging/sentries or some other notification when these calls fail?

aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceDAO.scala
```diff
@@ -25,6 +25,10 @@ trait ExecutionServiceDAO extends ErrorReportable {
   def abort(id: String, userInfo: UserInfo): Future[Try[ExecutionServiceStatus]]
   def getLabels(id: String, userInfo: UserInfo): Future[ExecutionServiceLabelResponse]
   def patchLabels(id: String, userInfo: UserInfo, labels: Map[String, String]): Future[ExecutionServiceLabelResponse]
+  def getCost(id: String,
+              workflowCostBreakdownParams: Option[WorkflowCostBreakdownParams],
```
jgainerdewar: Apologies for not letting you know earlier, we decided to pull out these params for now since we have no immediate use for them.

model/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala
```diff
@@ -144,6 +144,16 @@ case class WorkflowCost(
   cost: Option[Float]
 )
 
+case class WorkflowCostBreakdownParams(includeTaskBreakdown: Boolean, includeSubworkflowBreakdown: Boolean)
+
+case class WorkflowCostBreakdown(id: String,
+                                 cost: BigDecimal,
+                                 currency: String,
+                                 status: String,
+                                 subworkflowBreakdown: Option[Map[String, BigDecimal]],
```
jgainerdewar: These two breakdown items will not be included after all, but there will be a list of strings with the key `errors` that describes any issues we ran into while computing the cost. Might be worth having a separate conversation about the various error conditions and how we want to handle them, but if `errors` is nonempty we know that the `cost` is NOT the full current cost of the workflow.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpExecutionServiceDAO.scala
marctalbott: This will get logged, but won't otherwise alert via Sentry or GCP. This follows the [current behavior](https://github.com/broadinstitute/rawls/blob/54010c7378acaff88da1ae6276307701708a5ad0/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala#L341-L344) when Rawls has trouble monitoring a submission. It does pose a problem because Rawls won't be able to enforce a cost cap if it's not able to check the status of a submission's workflows for an extended period of time. 

Error handling in general probably needs some thought since it's also possible that Cromwell's cost estimation won't be complete as Janet mentioned in [another comment](https://github.com/broadinstitute/rawls/pull/3034#discussion_r1763411443). Not sure how conservative we need to be, but I'm guessing pretty conservative since user spend is at stake.

blakery (approved):

PR #3044, opened 2024-09-24
AJ-2002 Render local-dev config with gsm instead of vault

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-2002

---
This PR updates the `render` script used for setting up config for local development to use gsm instead of vault.  Rather than relying on an external repo (first, `firecloud-develop`, later, `firecloud-develop-shim`), everything is done locally in one script.

For testing, I recommend checking out the repo in a separate directory from any rawls you may have previously checked out, to simulate starting development anew.
Then, from the root of the repo:
1. Make sure you are on Broad internet or VPN.
2. Run `./local-dev/bin/render`
3. Run `./config/docker-rysnc-local-rawls.sh`
4. Once everything seems to have built, navigate to `https://local.dsde-dev.broadinstitute.org:20443/`.  You should see the swagger page and be able to authenticate and call the APIs.

**PR checklist**


local-dev/templates/rawls.conf
```diff
@@ -53,15 +45,15 @@ gcs {
...
+  requesterPaysRole=organizations/400176686919/roles/RequesterPays
+  terraBucketReaderRole=organizations/400176686919/roles/terraBucketReader
+  terraBucketWriterRole=organizations/400176686919/roles/terraBucketWriter
+  terraBillingProjectOwnerRole=organizations/400176686919/roles/terra_billing_project_owner
+  terraWorkspaceCanComputeRole=organizations/400176686919/roles/terra_workspace_can_compute
+  terraWorkspaceNextflowRole=organizations/400176686919/roles/terra_workspace_nextflow_role
 
   servicePerimeters {
-    staticProjects = {{$staticPerimeterProjects | toJSON}}
+    staticProjects = {}
```
calypsomatic, line 49: I think we don't need anything here for local development, see https://broadinstitute.slack.com/archives/C03BCARP494/p1727097190516919?thread_ts=1727096967.429529&cid=C03BCARP494

local-dev/templates/sqlproxy.env
```diff
@@ -0,0 +1,4 @@
+GOOGLE_PROJECT=broad-dsde-dev
```
calypsomatic, line 1: The name of the google project was part of the secret before, but I think it's ok for this to not be secret

local-dev/templates/rawls.conf
```diff
@@ -1,34 +1,26 @@
-{{- $appsDomain := "test.firecloud.org" -}}
-{{- $appsSubDomain := printf "dev.%s" $appsDomain -}}
```
calypsomatic, line 2: These are all just to prevent typing the same thing over and over again - and presumably to make it easy to replace if needed for some reason - but I figured it wasn't necessary and just put the values in directly

local-dev/bin/render
```diff
@@ -2,26 +2,85 @@
...
+RAWLS_CONF_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret rawls-conf)
+RAWLS_REFRESH_TOKEN_CRED_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret refresh-token-oauth-credential)
+RAWLS_SAKEY_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret rawls-sa)
+CROMWELL_SAKEY_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret cromwell-sa)
+B2C_APPID_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret b2c-application-id)
+LANDING_ZONE_POSTGRES_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret wsm-azure-postgres-creds)
+WEB_OBJECT=$(echo "$RAWLS_REFRESH_TOKEN_CRED_JSON" | jq -r '.value | fromjson | .web')
+
+#Define some reusable variables
+APPS_DOMAIN="test.firecloud.org"
```
calypsomatic: I pulled these out early on in the theory that they might want to be replaced with a different value, but I've come to think there's no need for these to be variables and replacement, so maybe I can remove this part of the script and just hard code the values in the conf file.  Thoughts?

local-dev/bin/render
```diff
@@ -2,26 +2,88 @@
...
+cp local-dev/templates/site.conf config/site.conf
 
-Don't forget to add the local.broadinstitute.org alias to your hosts file!
+##Files to download from secrets##
+gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-sa > config/rawls-account.json
+gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-sa > config/sqlproxy-service-account.json
+gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-sa-pem | jq -r '."key.pem"' > config/rawls-account.pem
+gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-bigquery-sa > config/bigquery-account.json
+gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-billing-sa | jq -r '."key.pem"' > config/billing-account.pem
+gcloud secrets versions access latest --project broad-dsde-dev --secret buffer-client-sa-b64 | jq -r '.key' | base64 --decode > config/buffer-account.json
```
davidangb: might wanna create a helper function for `gcloud secrets versions access latest --project broad-dsde-dev --secret %1` to avoid all the copy/paste … this would also make a central place to change `broad-dsde-dev` in case anyone ever wanted to read secrets from elsewhere

davidangb (commented):
I like the approach overall!

This removes the .ctmpl extension from the config files (yay!), but keeps the consul/go template syntax for variables, e.g. `{{ $rawlsOauthCredential }}`. I'm wondering if we should use a go-template CLI tool to do the variable substitution instead of doing it ourselves with `sed` … what do you think?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceAdminService.scala
```diff
@@ -6,28 +6,41 @@ import org.broadinstitute.dsde.rawls.NoSuchWorkspaceException
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.slick._
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
-import org.broadinstitute.dsde.rawls.model.{AttributeName, AttributeValue,WorkspaceAttributeSpecs, ErrorReportSource, ManagedGroupRef, RawlsGroupName, RawlsRequestContext, SamResourceTypeName, SamResourceTypeNames, Workspace, WorkspaceDetails, WorkspaceFeatureFlag, WorkspaceName}
+import org.broadinstitute.dsde.rawls.model.{
```
calypsomatic, line 9: These scala files were failing the `scalafmt` check after I pulled in the latest changes, so I ran `scalafmt`, but otherwise have not touched them.  Not sure how they ended up on main in the incorrect format!

local-dev/bin/render
```diff
@@ -2,26 +2,93 @@
...
-IMAGE_TAG="${IMAGE_TAG:-latest}"
-IMAGE="${IMAGE_NAME}:${IMAGE_TAG}"
+# Helper function to fetch secrets
+fetch_secret() {
+  gcloud secrets versions access latest --project "$PROJECT_ID" --secret "$1"
+}
 
-VAULT_TOKEN=$( cat ~/.vault-token )
+##Pull mysql instance info for sqlproxy##
+RAWLS_MYSQL_INSTANCE_JSON=$(gcloud secrets versions access latest  --project broad-dsde-dev --secret rawls-mysql-instance)
```
davidangb: nit: can this use `fetch_secret`?

local-dev/bin/render
```diff
@@ -2,26 +2,93 @@
...
+cp local-dev/templates/docker-rsync-local-rawls.sh config/docker-rsync-local-rawls.sh
+cp local-dev/templates/liquibase.properties config/liquibase.properties
+cp local-dev/templates/local-mysql.conf config/local-mysql.conf
+cp local-dev/templates/local-sam.conf config/local-sam.conf
+cp local-dev/templates/mod_security_ignore.conf config/mod_security_ignore.conf
+cp local-dev/templates/oauth2.conf config/oauth2.conf
+cp local-dev/templates/site.conf config/site.conf
+
+##Files to download from secrets##
+#gcloud secrets versions access latest --project broad-dsde-dev --secret rawls-sa > config/rawls-account.json
```
davidangb: nit: remove commented-out line

davidangb (commented):
this is awesome, and unblocks development, thank you! A few things I noticed:
* If the `config` directory is missing entirely, such as on a clean checkout of the rawls repo, the script fails
* The script required VPN for me. Is that expected, perhaps due to the kubectl stuff? If so, probably worth adding a code comment or an `echo` statement to that effect
* my preference would be to have a few `echo` status lines through the script, so a user knows what it's doing and could help debug if it fails partway through
* I bet this script will fail on Windows :) I'm ok with that

local-dev/templates/docker-rsync-local-rawls.sh
```diff
@@ -126,9 +126,9 @@ start_server () {
     -e PROXY_URL='http://rawls-sbt:8080/' \
     -e PROXY_URL2='http://rawls-sbt:8080/api' \
     -e PROXY_URL3='http://rawls-sbt:8080/register' \
-    -e CALLBACK_URI='https://local.broadinstitute.org/oauth2callback' \
+    -e CALLBACK_URI='https://local.dsde-dev.broadinstitute.org/oauth2callback' \
```
calypsomatic, line 129: I've replaced `local.broadinstitute.org` with `local.dsde-dev.broadinstitute.org` everywhere in the code - please tell me if any of these need to stay `local.broadinstitute.org`!

local-dev/templates/docker-rsync-local-rawls.sh
```diff
@@ -150,6 +150,11 @@ start_server () {
     docker start sqlproxy
     echo "Starting proxy..."
     docker start rawls-proxy
+     # Check if rawls-proxy started successfully
```
calypsomatic, line 158: rawls-proxy was failing quietly and I didn't notice - I've added this in so it wouldn't continue to run if the proxy failed, but haven't been able to reproduce a failure to see if it even works.  Can remove, or welcome comments on a better check

davidangb: I'm not opposed, but all of the containers (rawls-proxy, rawls-sbt, sqlproxy, rawls-rsync-container) are necessary. I think we should check all of them or none of them to be consistent.

local-dev/bin/render
```diff
@@ -2,26 +2,96 @@
...
-  --templates /code/local-dev/templates \
-  --output-dir /code/config
+# Extract the name and zone values using jq
+CLOUDSQL_ZONE=$(echo "$RAWLS_MYSQL_INSTANCE_JSON" | jq -r '.zone')
+CLOUDSQL_INSTANCE=$(echo "$RAWLS_MYSQL_INSTANCE_JSON" | jq -r '.name')
 
-cat <<EOF
+# Define the template and output file paths
+MYSQL_TEMPLATE_FILE="local-dev/templates/sqlproxy.env"
+MYSQL_OUTPUT_FILE="config/sqlproxy.env"
```
davidangb: add a `mkdir -p config` here, to ensure the `config` directory exists before we try to write to it?

davidangb (commented):
looking good, two comments inline!

Before/just after this merges, make sure to announce to anyone doing rawls development about the move from `local.broadinstitute.org` to `local.dsde-dev.broadinstitute.org`

local-dev/templates/docker-rsync-local-rawls.sh
calypsomatic: Ok I added one for sqlproxy - I think for the last one you'll see in the output any failures

davidangb (approved):
it's all working for me, and I can auth successfully to swagger-ui. Thank you!

ungwudik:
Looks like rawls repo is not configured to require a minimum of 2 reviews/approvals. Making a note of this for now.

PR #3045, opened 2024-09-24
WOR-1823: use original comparisons to resolve canShare

blakery:

Fixes a bug introduced earlier, which matches accessLevel exactly, instead of using comparisons when resolving canShare on a workspace.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -283,12 +283,11 @@ class WorkspaceService(
       }
       owners = ownersPolicy.map(policy => policy.memberEmails.map(_.value))
       canShare <- options.anyPresentFuture("canShare") {
-        accessLevel match {
-          case WorkspaceAccessLevels.Read  => Future.successful(false)
-          case WorkspaceAccessLevels.Owner => Future.successful(true)
-          case _ =>
-            val sharePolicy = SamWorkspaceActions.sharePolicy(accessLevel.toString.toLowerCase())
-            samDAO.userHasAction(SamResourceTypeNames.workspace, workspaceId, sharePolicy, ctx)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -283,12 +283,11 @@ class WorkspaceService(
       }
       owners = ownersPolicy.map(policy => policy.memberEmails.map(_.value))
       canShare <- options.anyPresentFuture("canShare") {
-        accessLevel match {
-          case WorkspaceAccessLevels.Read  => Future.successful(false)
-          case WorkspaceAccessLevels.Owner => Future.successful(true)
-          case _ =>
-            val sharePolicy = SamWorkspaceActions.sharePolicy(accessLevel.toString.toLowerCase())
-            samDAO.userHasAction(SamResourceTypeNames.workspace, workspaceId, sharePolicy, ctx)
+        if (accessLevel < WorkspaceAccessLevels.Read) Future.successful(false)
```
blakery, line 286: These are the original comparisons used to determine canShare.

Note: the only permission that is `< WorkspaceAccessLevels.Read` is `NoAccess`, which would have thrown an exception when getting the workspace earlier.

cahrens: Should there be a test case for PROJECT_OWNER?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2230,7 +2229,6 @@ class WorkspaceService(
         )
     }
 
-
```
cahrens, line 2233: You might want to delete the blank lines.

cahrens (approved):

marctalbott (approved):

PR #3046, opened 2024-09-25
[WOR-1698] Workspace admin support API

marctalbott:
Ticket: [WOR-1698](https://broadworkbench.atlassian.net/browse/WOR-1698)
* Add new API `/api/admin/workspaces/{workspaceId}` which returns the workspace details and its settings.
  * I opted not to return the same response as the non-admin `getWorkspace` as that response includes a lot of information about specific user access to the workspace. 
  * One notable omission is the Azure managed app coordinates. We'll need to add a WSM admin API that returns workspace information to fetch the managed app coordinates. Given our current stance on Azure, I felt this wasn't worth implementing right now.

---

**PR checklist**

[WOR-1698]: https://broadworkbench.atlassian.net/browse/WOR-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aherbst-broad (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -683,6 +683,17 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
 
         callback.future.map(_ => ())
       }
+
+    override def userHasAction(resourceTypeName: SamResourceTypeName,
```
blakery: The name of this should probably indicate that this is for an admin permission (something like: `userHasAdminAction`).

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceAdminService.scala
```diff
@@ -98,23 +118,41 @@ extends LazyLogging
       }
     }
 
+  def getWorkspaceById(workspaceId: UUID): Future[WorkspaceAdminResponse] =
+    samDAO.admin
```
blakery: Why not just put this as part of the `for` comprehension?
```suggestion
    for {
      userIsAdmin <- samDAO.admin.userHasAction(SamResourceTypeNames.workspace, SamResourceTypeAdminActions.readSummaryInformation, ctx)
      _ = if (!userIsAdmin) throw new RawlsExceptionWithErrorReport(
        ErrorReport(StatusCodes.Forbidden, "You must be an admin to call this API.")
      )
      workspaceOpt <- workspaceRepository.getWorkspace(workspaceId)
      workspace = workspaceOpt.getOrElse(throw NoSuchWorkspaceException(workspaceId.toString))
      settings <- workspaceSettingRepository.getWorkspaceSettings(workspaceId)
    } yield WorkspaceAdminResponse(
      WorkspaceDetails.fromWorkspaceAndOptions(workspace, None, useAttributes = false),
       settings
    )
```

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/AdminApiServiceSpec.scala
```diff
@@ -315,4 +318,26 @@ class AdminApiServiceSpec extends ApiServiceSpec {
     }
   }
 
+  it should "get a workspace with its current settings" in {
```
blakery, line 321: 🥳

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceAdminServiceUnitTests.scala
blakery, line 1: Super Nit/FYI/note for posterity: `___Spec` is the normal convention for naming test files with scalatest. I just started naming things `___UnitTests` because I needed to put things in separate classes when migrating conventions.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -5797,6 +5814,18 @@ components:
           description: True if the user can launch compute in this workspace, false
             otherwise
       description: ""
+    WorkspaceDetailsWithSettings:
```
blakery: Should this be `WorkspaceAdminResponse`, or `WorkspaceAdminDetailsWithSettings`, or something similar, to match up with the name in `WorkspaceModel`?

blakery (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
marctalbott: Yeah, I left it out because it's in the `AdminDao`, but probably doesn't hurt to be extra clear

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Yes, good catch, just missed it when I renamed the model

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceAdminService.scala
marctalbott: No compelling reason, I can switch it

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceAdminServiceUnitTests.scala
marctalbott: Yeah, I've just been sticking with `UnitTests` to distinguish the two styles of testing. I don't feel particularly strongly about continuing to do that vs following the normal convention

blakery (approved):

PR #3048, opened 2024-09-26
WOR-1823 part 2: listWorkspaces refactors

blakery:

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -2263,6 +2201,19 @@ class WorkspaceService(
       case (None, None) => Future(Some(config.defaultLocation))
     }
 
+  private def withLibraryAttributeNamespaceCheck[T](attributeNames: Iterable[AttributeName])(op: => T): T = {
```
blakery, line 2204: This used to be defined in a trait, but it's only used in `WorkspaceService`.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -43,19 +46,6 @@ trait LibraryPermissionsSupport extends RoleSupport {
       case _                                        => changeMetadataChecker(workspaceId) _
     }
 
-  def withLibraryAttributeNamespaceCheck[T](attributeNames: Iterable[AttributeName])(op: => T): T = {
```
blakery, line 46: Moved to its only call site.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/RoleSupport.scala
```diff
@@ -28,20 +28,4 @@ trait RoleSupport {
           new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, "You must be an admin."))
         )
     }
-
-  def tryIsCurator(userEmail: RawlsUserEmail): Future[Boolean] =
```
blakery, line 32: This was easily eliminated by taking its one usage, and using `gcsDAO.isLibraryCurator` in the same `for` comprehension as the rest of the operations.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
blakery, line 1: In this file, I just added some missing `behavior of "XXX"`, and moved some things around so they were next to each other.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -283,8 +290,7 @@ class WorkspaceService(
       }
       owners = ownersPolicy.map(policy => policy.memberEmails.map(_.value))
       canShare <- options.anyPresentFuture("canShare") {
-        if (accessLevel < WorkspaceAccessLevels.Read) Future.successful(false)
-        else if (accessLevel >= WorkspaceAccessLevels.Owner) Future.successful(true)
+        if (accessLevel >= WorkspaceAccessLevels.Owner) Future.successful(true)
```
blakery, line 293: `No Access` is only the permission below reader. If the reader calls either `getWorkspace` or `getWorkspaceById` without at least read access, the entire endpoint will throw an exception.

Even if that were not the case, we'll always get the correct answer if we call Sam anyway.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -51,8 +51,8 @@ class WorkspaceRepository(dataSource: SlickDataSource) {
     _.workspaceQuery.getV2WorkspaceId(workspaceName)
   }
 
-  def listWorkspacesByIds(workspaceIds: Seq[UUID]): Future[Seq[Workspace]] = dataSource.inTransaction {
-    _.workspaceQuery.listV2WorkspacesByIds(workspaceIds)
+  def listWorkspacesByIds(workspaceIds: Seq[UUID], attributeSpecs: Option[WorkspaceAttributeSpecs] = None): Future[Seq[Workspace]] = dataSource.inTransaction {
+    _.workspaceQuery.listV2WorkspacesByIds(workspaceIds,attributeSpecs)
```
cahrens: I'm surprised that a space isn't required here.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -319,14 +325,97 @@ class WorkspaceService(
...
+        workspaceRepository.listSubmissionSummaryStats(accessLevelWorkspaceUUIDs)
+      }
+      workspaces <- workspaceRepository.listWorkspacesByIds(accessLevelWorkspaceUUIDs, Option(options.attrSpecs))
+      highestAccessLevelByWorkspaceId = accessLevelWorkspaceResources.map { resource =>
+        resource.resourceId -> resource.allRoles.flatMap(role => WorkspaceAccessLevels.withRoleName(role.value)).max
+      }.toMap
+      workspaceSamResourceByWorkspaceId = accessLevelWorkspaceResources.map(r => r.resourceId -> r).toMap
+      aggregatedWorkspaces = new AggregatedWorkspaceService(workspaceManagerDAO)
+        .fetchAggregatedWorkspaces(workspaces, ctx)
+        // Filter out workspaces with no cloud contexts, logging cloud context exceptions
```
cahrens: We no longer log the cloud context exceptions (which is probably fine, but the comment should be changed).

cahrens (approved):
A couple of questions/notes, but otherwise LGTM.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -549,7 +560,416 @@ class WorkspaceServiceUnitTests extends AnyFlatSpec with OptionValues with Mocki
...
+        (WorkspaceCloudPlatform.Azure,
+         SamWorkspaceRoles.reader,
+         Set[SamResourceRole](SamWorkspaceRoles.canCompute),
+         false
+        ),
+        (WorkspaceCloudPlatform.Gcp, SamWorkspaceRoles.owner, Set[SamResourceRole](), true),
+        (WorkspaceCloudPlatform.Gcp, SamWorkspaceRoles.projectOwner, Set[SamResourceRole](), true),
+        (WorkspaceCloudPlatform.Gcp, SamWorkspaceRoles.writer, Set(SamWorkspaceRoles.canCompute), true),
+        (WorkspaceCloudPlatform.Gcp, SamWorkspaceRoles.writer, Set[SamResourceRole](), false),
+        (WorkspaceCloudPlatform.Gcp, SamWorkspaceRoles.reader, Set(SamWorkspaceRoles.canCompute), true),
```
marctalbott, line 742: Technically this situation shouldn't be allowed, although I know we just looked at how someone can get into this situation when a writer attempts to modify other users' access to a workspace. I'm not sure if it makes sense to include it as a test case if it's an unsupported situation

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
blakery: I also would have thought the formatter would catch that.

marctalbott (approved):

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
blakery: I included it because being exhaustive on a table driven test is trivial, and the code supports the situation.

PR #3049, opened 2024-09-27
WOR-1823 part 3: workspace deletion refactoring

blakery:

Refactors workspace deletion and adds unit tests.

---

**PR checklist**


core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -43,19 +46,6 @@ trait LibraryPermissionsSupport extends RoleSupport {
       case _                                        => changeMetadataChecker(workspaceId) _
     }
 
-  def withLibraryAttributeNamespaceCheck[T](attributeNames: Iterable[AttributeName])(op: => T): T = {
```
blakery: This was moved to `WorkspaceService`, since it is only called there.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/LibraryPermissionsSupport.scala
```diff
@@ -4,16 +4,19 @@ import akka.http.scaladsl.model.StatusCodes
...
+import scala.concurrent.{ExecutionContext, Future}
 
 /**
   * Created by ahaessly on 3/31/17.
   */
-trait LibraryPermissionsSupport extends RoleSupport {
+trait LibraryPermissionsSupport  {
   val samDAO: SamDAO
+  val gcsDAO: GoogleServicesDAO
+  val ctx: RawlsRequestContext
```
blakery: These were being inherited by `RoleSupport`, but there's no real reason for this trait to be connected to it.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/RoleSupport.scala
```diff
@@ -28,20 +28,4 @@ trait RoleSupport {
           new RawlsExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, "You must be an admin."))
         )
     }
-
-  def tryIsCurator(userEmail: RawlsUserEmail): Future[Boolean] =
-    gcsDAO.isLibraryCurator(userEmail.value) recoverWith { case t =>
```
blakery: This was only used once, So I just moved it inline.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/TracingUtils.scala
```diff
@@ -96,6 +96,33 @@ object TracingUtils {
...
 
+  /**
+    * Trace a sync operation with the RawlsRequestContext
+    *
+    * @param name The name that will be used in the trace
+    * @param parentContext the RawlsRequestContext to use for tracing, if `parentContext.otelContext` is defined
+    * @param f the operation to execute within the trace
+    * @return the result of the operation `f`
+    */
+  def trace[T](name: String, parentContext: RawlsRequestContext)(f: RawlsRequestContext => T): T =
```
blakery, line 107: Not having this was pushing a lot of different places to use a `Future` where it wasn't really necessary.

core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -17,19 +17,15 @@ import org.broadinstitute.dsde.rawls.entities.exceptions.{
 import org.broadinstitute.dsde.rawls.expressions.ExpressionEvaluator
 import org.broadinstitute.dsde.rawls.metrics.RawlsInstrumented
 import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{AttributeUpdateOperation, EntityUpdateDefinition}
-import org.broadinstitute.dsde.rawls.model.{
```
blakery, line 20: The changes in this file are just from moving an exception definition to the place where it's thrown.

core/src/main/scala/org/broadinstitute/dsde/rawls/metrics/MetricsHelper.scala
```diff
@@ -40,11 +40,12 @@ object MetricsHelper {
   def incrementFastPassRevokedCounter(memberType: IamMemberType): IO[Unit] =
     incrementFastPassUpdatedCounter(memberType, "revoke")
 
-  def incrementCounter(name: String,
-                       count: Int = 1,
-                       labels: Map[String, String] = Map.empty,
-                       description: Option[String] = None
-  ): IO[Unit] = {
```
blakery, line 47: Unneccesary IO wrapping - only called in one place anyway.

core/src/main/scala/org/broadinstitute/dsde/rawls/submissions/SubmissionsRepository.scala
```diff
@@ -0,0 +1,53 @@
...
+  */
+class SubmissionsRepository(
+  dataSource: SlickDataSource,
+  trackDetailedSubmissionMetrics: Boolean = true,
+  override val workbenchMetricBaseName: String
+) extends RawlsInstrumented {
+
+  import dataSource.dataAccess.driver.api._
+
+  def getActiveWorkflowsAndSetStatusToAborted(
```
blakery, line 24: This was moved in wholesale from `WorkspaceService`.

core/src/main/scala/org/broadinstitute/dsde/rawls/util/AttributeSupport.scala
```diff
@@ -172,10 +171,13 @@ trait AttributeSupport {
...
-   * @throws org.broadinstitute.dsde.rawls.workspace.AttributeNotFoundException when removing from a list attribute that does not exist
+   * @throws AttributeNotFoundException when removing from a list attribute that does not exist
    * @throws AttributeUpdateOperationException when adding or removing from an attribute that is not a list
    * @return the updated entity
    */
   def applyOperationsToEntity(entity: Entity, operations: Seq[AttributeUpdateOperation]): Entity =
     entity.copy(attributes = applyAttributeUpdateOperations(entity, operations))
 }
+
+class AttributeUpdateOperationException(message: String) extends RawlsException(message)
```
blakery, line 182: This is the only place these are thrown, so it makes sense for them to be defined here.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -61,10 +67,15 @@ class WorkspaceServiceUnitTests
 
   implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
 
-  val defaultRequestContext: RawlsRequestContext =
-    RawlsRequestContext(
-      UserInfo(RawlsUserEmail("test"), OAuth2BearerToken("Bearer 123"), 123, RawlsUserSubjectId("abc"))
+  val ctx: RawlsRequestContext = RawlsRequestContext(
```
blakery, line 70: I got tired of having `defaultRequestContext` making practically every mock require a line break. Sorry about the resulting diff size.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -79,7 +84,6 @@ import scala.jdk.DurationConverters.JavaDurationOps
 import scala.language.postfixOps
 import scala.util.Try
 
-//noinspection NameBooleanParameters,TypeAnnotation,EmptyParenMethodAccessedAsParameterless,ScalaUnnecessaryParentheses,RedundantNewCaseClass,ScalaUnusedSymbol
```
blakery, line 82: Nearly all of the changes in this file are fixing ide warnings from removing this.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -465,286 +467,116 @@ class WorkspaceService(
...
+      fastPassServiceConstructor(childContext).removeFastPassGrantsForWorkspace(workspace)
+    )
+    // notify leonardo so it can cleanup any dangling sam resources and other non-cloud state
+    _ <- traceFutureWithParent("notifyLeonardo", ctx)(_ =>
+      leonardoService.cleanupResources(workspace.googleProjectId, workspace.workspaceIdAsUUID, ctx)
+    )
+    // Delete Google Project
+    _ <- traceFutureWithParent("deleteGoogleProject", ctx)(_ => deleteGoogleProject(workspace.googleProjectId))
+    // attempt to delete workspace in WSM, in case thsi is a TDR snapshot - but don't fail on it
+    _ = Try(workspaceManagerDAO.deleteWorkspace(workspace.workspaceIdAsUUID, ctx)).recover {
```
blakery, line 498: This is the same behavior we had before, because we would only fail if it was a `McWorkspace` we were trying to delete, which would never happen, since those are handled by `MultiCloudWorkspaceService`.

cahrens (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/util/TracingUtils.scala
marctalbott: This appears to be equivalent to `traceNakedWithParent` below

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -465,286 +467,116 @@ class WorkspaceService(
...
+    )
+    _ <- traceFutureWithParent("deleteFastPassGrantsTransaction", ctx)(childContext =>
+      fastPassServiceConstructor(childContext).removeFastPassGrantsForWorkspace(workspace)
+    )
+    // notify leonardo so it can cleanup any dangling sam resources and other non-cloud state
+    _ <- traceFutureWithParent("notifyLeonardo", ctx)(_ =>
+      leonardoService.cleanupResources(workspace.googleProjectId, workspace.workspaceIdAsUUID, ctx)
+    )
+    // Delete Google Project
+    _ <- traceFutureWithParent("deleteGoogleProject", ctx)(_ => deleteGoogleProject(workspace.googleProjectId))
```
marctalbott, line 496: There are technically still v1 workspaces in Rawls and this would bork all other v1 workspaces in a billing project by deleting the billing project's google project. Painful to have to leave it in given our stance on v1 workspaces, but as long as they're still around we probably need to keep the check

core/src/main/scala/org/broadinstitute/dsde/rawls/util/TracingUtils.scala
blakery: That's using` RawlsTracingContext` not `RawlsRequestContext`.

It's arguable that the better approach would be an overload or something, but we have similar divergence between others, such as between `traceFutureWithParent` and `traceFuture`. 
Also, this is the most common/basic/unembellished instance of this, and therefore worth giving the most basic name. But I didn't feel like renaming usages of `traceNakedWithParent` in what is already a substantial PR.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
blakery: If it was a v1 workspace, we'd never have gotten this far. 
The first operation is `getWorkspaceContextAndPermissions`, which only returns v2 workspaces. So calling this endpoint on a v1 workspace would already return a `NoSuchWorkspace` exception.

Note: IIRC, v1 workspaces are not returned or treated as valid targets for _any_ user-facing operation at this point.

marctalbott: Ah, `getWorkspaceContextAndPermissions` was supposed to return v1 workspaces in contrast to `getV2WorkspaceContextAndPermissions`, but I see that changed a few months ago when we moved `MultiCloudWorkspaceService` over to the `WorkspaceRepository`. The idea was to allow users to _delete_ v1 workspaces, but not do anything else with them. Since the decision's been made for us, this is fine!

core/src/main/scala/org/broadinstitute/dsde/rawls/util/TracingUtils.scala
marctalbott: Oh, right. Reading is hard. I prefer this to overloading 👍

marctalbott (approved):

PR #3058, opened 2024-10-01
[WOR-1700] Billing project support API

marctalbott:
Ticket: [WOR-1700](https://broadworkbench.atlassian.net/browse/WOR-1700)
* Add new API to return billing project information for admins. Note that this does not include any information about the billing project's billing profile or landing zone, so this API will be of limited use for Azure billing projects. New admin APIs in BPM and LZS are required to improve this API for Azure billing projects.

---

**PR checklist**

[WOR-1700]: https://broadworkbench.atlassian.net/browse/WOR-1700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -6786,32 +6786,19 @@ components:
           description: the name of the project
         billingAccount:
           type: string
-          description: the billing account to use in google projects
+          description: the billing account to use in google projects (cloudPlatform GCP only)
         invalidBillingAccount:
           type: boolean
           description: whether or not the billing account is usable by Terra
         status:
-          type: string
-          enum:
-            - Creating
-            - Ready
-            - Error
-            - Deleting
-            - DeletionFailed
-            - AddingToPerimeter
-            - CreatingLandingZone
-          description: the status of allocating the billing project's resources.
+          $ref: '#/components/schemas/BillingProjectStatus'
         message:
           type: string
           description: informational message about the project
         azureManagedAppCoordinates:
           $ref: '#/components/schemas/AzureManagedAppCoordinates'
         cloudPlatform:
-          type: string
-          enum:
-            - GCP
-            - AZURE
-            - UNKNOWN
+          $ref: '#/components/schemas/BillingProjectCloudPlatform'
         landingZoneId:
           type: string
           format: uuid
@@ -6824,19 +6811,19 @@ components:
           description: the cromwell backend to use for this billing project
         servicePerimeter:
           type: string
-          description: the name of the service perimeter for this billing project
+          description: the name of the service perimeter for this billing project (cloudPlatform GCP only)
         googleProjectNumber:
           type: string
-          description: the google project number for this billing project
+          description: the google project number for this billing project (cloudPlatform GCP only)
         spendReportDataset:
           type: string
-          description: the name of the BigQuery dataset containing project spend data
+          description: the name of the BigQuery dataset containing project spend data (cloudPlatform GCP only)
         spendReportTable:
           type: string
-          description: the name of the BigQuery table containing project spend data
+          description: the name of the BigQuery table containing project spend data (cloudPlatform GCP only)
         spendReportDatasetGoogleProject:
           type: string
-          description: the name of the Google Project where the BigQuery dataset resides
+          description: the name of the Google Project where the BigQuery dataset resides (cloudPlatform GCP only)
     RawlsBillingProjectResponse:
       required:
         - projectName
@@ -6862,27 +6849,14 @@ components:
             $ref: '#/components/schemas/ProjectRole'
           description: the roles the caller has on the project
         status:
-          type: string
-          enum:
-            - Creating
-            - Ready
-            - Error
-            - Deleting
-            - DeletionFailed
-            - AddingToPerimeter
-            - CreatingLandingZone
-          description: the status of allocating the billing project's resources.
+          $ref: '#/components/schemas/BillingProjectStatus'
         message:
           type: string
           description: informational message about the project
         azureManagedAppCoordinates:
           $ref: '#/components/schemas/AzureManagedAppCoordinates'
         cloudPlatform:
-          type: string
-          enum:
-            - GCP
-            - AZURE
-            - UNKNOWN
+          $ref: '#/components/schemas/BillingProjectCloudPlatform'
         landingZoneId:
           type: string
           format: uuid
@@ -6895,6 +6869,24 @@ components:
           description: whether this billing project supports protected data (cloudPlatform AZURE only)
         organization:
           $ref: '#/components/schemas/RawlsBillingProjectOrganization'
+    BillingProjectStatus:
+      type: string
+      enum:
+        - Creating
+        - Ready
+        - Error
+        - Deleting
+        - DeletionFailed
+        - AddingToPerimeter
+        - CreatingLandingZone
+      description: the status of allocating the billing project's resources.
+    BillingProjectCloudPlatform:
+      type: string
+      enum:
+        - GCP
+        - AZURE
+        - UNKNOWN
+      description: the cloud platform of the billing project
     AzureManagedAppCoordinates:
       required:
         - tenantId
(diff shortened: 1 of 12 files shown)


core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,16 @@ components:
           $ref: '#/components/schemas/ProjectRole'
           description: the role of the user in the project
       description: an element of a list of project users and their role
+    BillingProjectAdminResponse:
```
marctalbott, line 6769: I stubbed this out a while ago and forgot to update it, fixing

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,87 @@ components:
...
+        projectName:
+          type: string
+          description: the name of the project
+        billingAccount:
+          type: string
+          description: the billing account to use in google projects
+        invalidBillingAccount:
+          type: boolean
+          description: whether or not the billing account is usable by Terra
+        status:
```
cahrens, line 6803: Is there a way to create a shared definition of status (the possible enum values)? With the duplication, it would be easy for them to get out of sync.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,87 @@ components:
...
+        landingZoneId:
+          type: string
+          format: uuid
+          description: the UUID of the landing zone associated with the project (cloudPlatform AZURE only)
+        billingProfileId:
+          type: string
+          description: the billing profile ID associated with the project
+        cromwellBackend:
+          type: string
+          description: the cromwell backend to use for this billing project
```
cahrens, line 6821: Can there be a more useful description? I didn't know we had more than one Cromwell backend.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,87 @@ components:
...
+        servicePerimeter:
+          type: string
+          description: the name of the service perimeter for this billing project
+        googleProjectNumber:
+          type: string
+          description: the google project number for this billing project
```
cahrens: GCP only

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,87 @@ components:
...
+        spendReportDataset:
+          type: string
+          description: the name of the BigQuery dataset containing project spend data
```
cahrens: GCP only

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6756,87 @@ components:
...
+        spendReportTable:
+          type: string
+          description: the name of the BigQuery table containing project spend data
```
cahrens: GCP only

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingAdminService.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class BillingAdminService(samDAO: SamDAO,
+                          billingRepository: BillingRepository,
+                          workspaceRepository: WorkspaceRepository,
+                          ctx: RawlsRequestContext
+)(implicit protected val ec: ExecutionContext)
+    extends LazyLogging {
+
+  def getBillingProject(billingProjectName: RawlsBillingProjectName): Future[BillingProjectAdminResponse] =
```
cahrens: Maybe `getBillingProjectAndWorkspaces`?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
```diff
@@ -5,17 +5,7 @@ import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
...
-  PendingCloneWorkspaceFileTransfer,
-  RawlsRequestContext,
-  Workspace,
-  WorkspaceAttributeSpecs,
-  WorkspaceName,
-  WorkspaceState,
-  WorkspaceSubmissionStats,
-  WorkspaceTag
-}
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, PendingCloneWorkspaceFileTransfer, RawlsBillingProjectName, RawlsRequestContext, Workspace, WorkspaceAttributeSpecs, WorkspaceName, WorkspaceState, WorkspaceSubmissionStats, WorkspaceTag}
```
cahrens, line 8: I'm surprised that scalaFmt would prefer this. Can you make sure you ran `sbt scalafmtAll`?

core/src/main/resources/swagger/api-docs.yaml
marctalbott: I gave this vague description because I don't have a good understanding of what it's for. It's not set by default. I found a comment that says that possible values are "JES" and "PAPIv2", but I'm not sure if that's still accurate.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceRepository.scala
marctalbott: Agreed, but I did run it and the GHA check also passed.

core/src/main/resources/swagger/api-docs.yaml
cahrens: Ah, it's very old then!

Why can't you use the existing `RawlsBillingProjectResponse`? Are they too different?

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingAdminService.scala
marctalbott: Happy to change it, but I do want it to be clear that the full workspaces aren't included which I'm not sure it is with `getBillingProjectAndWorkspaces`.

Could do something like `getBillingProjectSupportSummary` which is how it's worded in Sam for managed groups

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Done! Also did it for cloud platform

core/src/main/resources/swagger/api-docs.yaml
marctalbott: Yeah, they're just a bit different so I thought it was better to document what can actually get returned

core/src/main/resources/swagger/api-docs.yaml
cahrens: Thanks!

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingAdminService.scala
cahrens: `getBillingProjectSupportSummary` seems reasonable.

cahrens (approved):

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6736,6 +6766,74 @@ components:
...
+        workspaces:
+          type: object
+          additionalProperties:
+            type: string
+            format: uuid
+          example:
+            "workspace1": "00000000-0000-0000-0000-000000000000"
+            "workspace2": "11111111-1111-1111-1111-111111111111"
+          description: "workspaces in the billing project with their IDs"
+    RawlsBillingProject:
```
cahrens, line 6786: @marctalbott it seems like (based on your UI PR) that far fewer fields are actually returned in this response. Is it just that case that theoretically this many _could_ be returned if configured for the billing project?

marctalbott: Yeah, that's correct. There are a lot of possible fields which is what leads this to be so long

trholdridge (approved):
Looks good to me. The only thing I could think of was including a BillingAdminService test for a billing project with no workspaces in it. But that isn't a necessity.

PR #3064, opened 2024-10-03
[WOR-432] Add additional validation for updateAcl

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-432

The terra-ui portion of this ticket has already been completed, so we don't present options in the Share Workspace modal that will fail. This PR adds additional validation logic to the Rawls `updateAcl` method so that we don't end up in a confusing "partially updated" state (if called via the Rawls Swagger).

Previously, the Rawls method did not check the permissions that the caller has on the workspace, and instead relied on Sam to throw exceptions. That was problematic because changing the workspace access level or permissions of a user is a 2-step process: we first add the new access levels/permissions, and then remove the old ones.

So, for instance, if a Writer attempted to change a Writer w/canShare -> Reader wout/canShare, the following would occur:

1. Add the user as Reader without canShare (would succeed)
2. Attempt to remove the user as Writer with canShare (would fail because a Writer can't modify the canShare permission)

Now we do validation before calling Sam.

---

**PR checklist**

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
@@ -369,12 +369,25 @@ class FastPassServiceSpec
     val newWorkspaceName = "space_for_workin"
     val workspaceRequest = WorkspaceRequest(testData.testProject1Name.value, newWorkspaceName, Map.empty)
     val workspace = Await.result(services.workspaceService.createWorkspace(workspaceRequest), Duration.Inf)
+    // Mock the caller being an owner on the workspace.
+    when(services.samDAO.listPoliciesForResource(ArgumentMatchers.eq(SamResourceTypeNames.workspace), any(), any()))
+      .thenReturn(
+        Future(
+          Set(
+            SamPolicyWithNameAndEmail(
+              SamWorkspacePolicyNames.owner,
+              SamPolicy(Set(WorkbenchEmail(services.ctx1.userInfo.userEmail.value)), Set.empty, Set.empty),
+              WorkbenchEmail("ownerPolicy@example.com")
+            )
+          )
+        )
+      )
 
     val aclAdd = Set(
       WorkspaceACLUpdate(testData.userWriter.userEmail.value, WorkspaceAccessLevels.Write, canCompute = Option(true)),
       WorkspaceACLUpdate(testData.userReader.userEmail.value, WorkspaceAccessLevels.Read, canShare = Option(true))
     )
-    Await.ready(services.workspaceService.updateACL(workspace.toWorkspaceName, aclAdd, false), Duration.Inf)
+    Await.result(services.workspaceService.updateACL(workspace.toWorkspaceName, aclAdd, false), Duration.Inf)
 
     verify(services.mockFastPassService)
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/fastpass/FastPassServiceSpec.scala
```diff
@@ -369,12 +369,25 @@ class FastPassServiceSpec
...
+          )
+        )
+      )
 
     val aclAdd = Set(
       WorkspaceACLUpdate(testData.userWriter.userEmail.value, WorkspaceAccessLevels.Write, canCompute = Option(true)),
       WorkspaceACLUpdate(testData.userReader.userEmail.value, WorkspaceAccessLevels.Read, canShare = Option(true))
     )
-    Await.ready(services.workspaceService.updateACL(workspace.toWorkspaceName, aclAdd, false), Duration.Inf)
+    Await.result(services.workspaceService.updateACL(workspace.toWorkspaceName, aclAdd, false), Duration.Inf)
```
cahrens, line 390: `Await.ready` doesn't verify that the operation did not throw an exception. This test case DID fail without the mock because after this line, we verify that FastPass was called. However, the other test case related to updateACL (below) did not fail even though an exception was thrown, and checking that the method did not error makes debugging failures easier.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -717,6 +717,8 @@ class WorkspaceServiceSpec
   }
 
   it should "invite a user to a workspace" in withTestDataServicesCustomSam { services =>
+    populateWorkspacePolicies(services)
```
cahrens, line 720: We didn't error before even though the user didn't have access to the workspace.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -1573,7 +1573,7 @@ class WorkspaceServiceUnitTests
   behavior of "updateAcl"
   it should "call Sam for Rawls workspaces" in {
     val projectOwnerEmail = "projectOwner@example.com"
-    val ownerEmail = "owner@example.com"
+    val ownerEmail = ctx.userInfo.userEmail.value // user making the request is an owner
```
cahrens, line 1576: In all of these test cases, the user calling `updateACL` didn't have access to the workspace.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -1811,6 +1809,136 @@ class WorkspaceServiceUnitTests
     Await.result(service.updateACL(WorkspaceName("fake_namespace", "fake_name"), aclUpdate, true), Duration.Inf)
   }
 
+  it should "not allow users to change their own ACL" in {
```
cahrens, line 1812: This was pre-existing functionality, but it had no test.

aherbst-broad (approved):

trholdridge (approved):

PR #3072, opened 2024-10-07
[WOR-432][CORE-49] Add additional validation for updateAcl (take 2)

cahrens:
Ticket: [CORE-49](https://broadworkbench.atlassian.net/browse/CORE-49)

* This reverts broadinstitute/rawls#3065 and modifies the code to list the actions that the user has on the workspace. Rawls checks that the user has the necessary `share_policy::_` permissions to make the requested ACL change.
* The change to list the user's workspace actions broke a lot of older unit tests that relied on the database and a test implementation of Sam `MockSamDAO`. We are trying to reduce unit test reliance on the database in Rawls as these tests are slow and modifying the test implementation of Sam can be difficult. Given this, I converted existing `updateAcl` tests into unit tests and moved them into a contained test file called `WorkspaceServiceUpdateAclSpec`.

[CORE-49]: https://broadworkbench.atlassian.net/browse/CORE-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -976,7 +975,11 @@ class WorkspaceService(
 
           // figure out which of the incoming aclUpdates are actually changes by removing all the existingAcls
           aclChanges = normalize(aclUpdates) -- existingAcls
-          _ = validateAclChanges(aclChanges, existingAcls, workspace)
+          callingUserRoles <- samDAO.listUserRolesForResource(SamResourceTypeNames.workspace,
```
cahrens: I wanted to just rely on what is in `existingAcls`, but if the calling user has the `projectOwner` role, they don't appear directly in `existingAcls` with their email address… instead there is a group for `projectOwner`, and presumably the calling user is a member of that group, but I don't know how to get that information.

This solution isn't ideal though because it does introduce another Sam call (though this method isn't called often, so that's a minor issue) and it breaks the pattern of going through the `workspaceAclManager` (though the Sam call could be hidden by adding a pass-through method to that interface, if folks felt like it was worth it).

marctalbott: The additional sam call isn't ideal, but this does have the benefit of working in cases where the user has access to the workspace via group membership more generally than just with the project-owner case (e.g. a group was added as a writer on the workspace).

marctalbott: We could add a new Sam API that returned a flattened policy membership for a resource if we wanted to, but we would still need the direct policy membership to calculate the ACL changes

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1077,9 +1080,35 @@ class WorkspaceService(
...
+      throw new InvalidWorkspaceAclUpdateException(
+        ErrorReport(StatusCodes.BadRequest, "you do not have access to change permissions for this workspace")
+      )
+    }
+    val callingUserMaxRole = callingUserRoles.flatMap(role => WorkspaceAccessLevels.withRoleName(role.value)).max
+    // Add the existingAcl entries that are being modified so we can check what we will
+    // be removing as well as what we are adding.
+    val allRolePermissionChanges =
+      aclChanges ++ existingAcls.filter(existingAcl => emailsBeingChanged.contains(existingAcl.email.toLowerCase))
+    if (callingUserMaxRole < WorkspaceAccessLevels.Owner) {
```
marctalbott: This makes assumptions about the actions contained in the `workspace` roles defined in Sam's resource type configuration, which isn't ideal as the name of a role in Sam shouldn't be used for authz control. Given that, this could be modified to check actions directly to make sure the user has the necessary `share_policy::_` actions for the requested ACL changes which would be the _right_ thing. 

Still, I don't know if that's strictly necessary here, but happy to discuss. This check isn't intended to be true authorization as that is still handled by Sam -- the goal here is just to provide some guard rails and prevent ACL changes we believe are unlikely to succeed.

marctalbott (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1075,53 +1143,6 @@ class WorkspaceService(
       .void
   }
 
-  private def validateAclChanges(aclChanges: Set[WorkspaceACLUpdate],
```
marctalbott, line 1078: Moved inside of `updateACL` since it's only called from there

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -956,6 +955,71 @@ class WorkspaceService(
...
+      // be removing as well as what we are adding.
+      val changingPolicies =
+        (aclChanges ++ existingAcls.filter(existingAcl => emailsBeingChanged.contains(existingAcl.email.toLowerCase)))
+          .flatMap(aclUpdateToPolicies)
+
+      if (changingPolicies.exists(policy => !callingUserActions.contains(SamWorkspaceActions.sharePolicy(policy.value)))) {
+        throw new InvalidWorkspaceAclUpdateException(
+          ErrorReport(StatusCodes.BadRequest, "you do not have sufficient permissions to make these changes")
+        )
+      }
```
marctalbott, line 979: These changes and the new call to Sam to list user actions on the workspace are the functional changes this PR is making

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -1522,222 +1284,6 @@ class WorkspaceServiceSpec
...
-         testData.workspace.namespace,
-         SamBillingProjectPolicyNames.canComputeUser,
-         aclTestUser.userEmail.value
-        )
-      )
-      services.samDAO.callsToRemoveFromPolicy should contain theSameElementsAs Set.empty
-    }
-  }
-
-  it should "not clobber catalog permission" in withTestDataServicesCustomSam { services =>
```
marctalbott, line 1605: This wasn't moved because `WorkspaceService` doesn't actually do anything to prevent the clobbering of catalog permissions. The `RawlsWorkspaceAclManager` filters out the catalog policy when listing policies, so a new unit test was added to `RawlsWorkspaceAclManagerUnitTests`.

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scala
```diff
@@ -497,244 +497,6 @@ class WorkspaceServiceSpec
     Await.result(populateAcl, Duration.Inf)
   }
```
marctalbott, line 498: In general, these tests were moved to `WorkspaceServiceUpdateAclSpec` and adjusted to remove their reliance on database and `CustomizableSamDao`. The one exception is called out below

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -1570,247 +1572,6 @@ class WorkspaceServiceUnitTests
...
-                                              fastPassServiceConstructor = _ => mockFastPassService
-    )(ctx)
-
-    val aclUpdate = Set(
-      WorkspaceACLUpdate(readerEmail, WorkspaceAccessLevels.Read, Option(true), Option(false)),
-      WorkspaceACLUpdate("foo@bar.com", WorkspaceAccessLevels.Read, Option(false), Option(false))
-    )
-
-    Await.result(service.updateACL(WorkspaceName("fake_namespace", "fake_name"), aclUpdate, true), Duration.Inf)
-  }
```
marctalbott, line 1812: These tests are covered by the various permutation tests, so they weren't moved over to `WorkspaceServiceUpdateAclSpec`

core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceUnitTests.scala
```diff
@@ -1570,247 +1572,6 @@ class WorkspaceServiceUnitTests
     verify(wsmDAO).getRoles(any(), any())
   }
 
-  behavior of "updateAcl"
```
marctalbott, line 1573: Other than the exceptions at the bottom, these tests were moved over to `WorkspaceServiceUpdateAclSpec` and modified to mock the `AclManager`s instead of Sam, WSM, and the database.

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -956,6 +955,71 @@ class WorkspaceService(
...
+          ErrorReport(StatusCodes.BadRequest, "you do not have access to change permissions for this workspace")
+        )
+      }
+      // Add the existingAcl entries that are being modified so we can check what we will
+      // be removing as well as what we are adding.
+      val changingPolicies =
+        (aclChanges ++ existingAcls.filter(existingAcl => emailsBeingChanged.contains(existingAcl.email.toLowerCase)))
+          .flatMap(aclUpdateToPolicies)
+
+      if (changingPolicies.exists(policy => !callingUserActions.contains(SamWorkspaceActions.sharePolicy(policy.value)))) {
```
dvoet, line 975: We don't grant `alter_policies` to workspace users so it should not be necessary to check that here. Just calling it out here in case anyone disagrees.

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
marctalbott: ➕ I also don't foresee us changing that model anytime soon since it'd let users modify the project owners policy which would not be great.

PR #3073, opened 2024-10-07
[CORE-50] Update workspace states after billing account change

trholdridge:
Ticket: [https://broadworkbench.atlassian.net/browse/CORE-50](https://broadworkbench.atlassian.net/browse/CORE-50)

This change accomplishes the AC of setting workspaces to the `Ready` state after their billing account is successfully updated. If a workspace is in the `UpdateFailed` state, a user can take its billing project through the billing account change process to attempt to fix this.

There are some things about the `BillingAccountChangeSynchronizer` that could be improved in service of readability and testing (for example, using `WorkspaceRepository` instead of `inTransaction` for database queries). I would like to do those things at some point, but I am not wrapping them up in this ticket.

Demo:

https://github.com/user-attachments/assets/6c7f714f-3618-4e85-88e7-d41b5d170332

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
@@ -659,6 +659,9 @@ trait WorkspaceComponent {
 
     def setIsLocked(isLocked: Boolean): WriteAction[Boolean] =
       query.map(_.isLocked).filter(_ =!= isLocked).update(isLocked).map(_ > 0)
+
+    def setState(state: WorkspaceState): WriteAction[Int] =
+      query.map(_.state).update(state.toString)
   }
 
   private def groupByWorkspaceId(runningSubmissions: Seq[(UUID, Int)]): Map[UUID, Int] =
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
@@ -254,9 +254,14 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
       billingAccount <- R.reader(_.newBillingAccount)
       _ <- inTransaction {
         workspacesToUpdate.setErrorMessage(errorMessage) *>
-          Applicative[WriteAction].whenA(errorMessage.isEmpty) {
-            workspacesToUpdate.setCurrentBillingAccountOnGoogleProject(billingAccount)
-          }
+          (errorMessage match {
+            case Some(_) => workspacesToUpdate.setState(WorkspaceState.UpdateFailed)
+            case None =>
+              DBIO.seq(
+                workspacesToUpdate.setCurrentBillingAccountOnGoogleProject(billingAccount),
+                workspacesToUpdate.setState(WorkspaceState.Ready)
+              )
+          })
       }
     } yield ()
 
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -659,6 +659,9 @@ trait WorkspaceComponent {
 
     def setIsLocked(isLocked: Boolean): WriteAction[Boolean] =
       query.map(_.isLocked).filter(_ =!= isLocked).update(isLocked).map(_ > 0)
+
+    def setState(state: WorkspaceState): WriteAction[Int] =
```
cahrens, line 663: @trholdridge there's already an `updateState` method up at line 404. Did you try using that?

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -202,6 +202,9 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
 
       outcome <- workspaces.foldMapA { workspace =>
         for {
+          _ <- inTransaction {
+            workspaceQuery.withWorkspaceId(workspace.workspaceIdAsUUID).setState(WorkspaceState.Updating)
```
cahrens: Have you been able to test if this is working correctly?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
trholdridge: I created the new one because the other DB updates in that part of `BillingAccountChangeSynchronizer` extend an existing query. I wasn't sure how to back the `workspaceId` out of that query to pass to the existing `updateState` function. I will look into that though. Would be good to rule it out as a source of error.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
trholdridge: I will check by running locally/on dev once Sam is back up, but if I just change the later code to never set it to `UpdateFailed`/`Ready`, the test does appear to confirm that it gets set to `Updating`.

trholdridge: Hm... interestingly, I can't find evidence of it moving to `Updating` when changing the billing account through Terra UI and querying the dev database directly. I don't have a better methodology than just querying repeatedly after adding/removing the billing account, though.

cahrens: I was asking if you had test coverage for it to make sure this bit worked.

trholdridge: No, I haven't figured out yet how to go about testing it directly.

calypsomatic:
> 
> . I would like to do those things at some point, but I am not wrapping them up in this ticket.
> 
Unrelated to this PR, but is there a ticket somewhere in the backlog for this?

calypsomatic (approved):

dvoet (approved):

dvoet:
> There are some things about the `BillingAccountChangeSynchronizer` that could be improved in service of readability and testing (for example, using `WorkspaceRepository` instead of `inTransaction` for database queries). I would like to do those things at some point, but I am not wrapping them up in this ticket.

I would be careful about fiddling with `inTransaction`. The transaction semantics were probably chosen with care. In other words, the code is probably precise about when updates are committed to the database. If a refactor would make this code more maintainable, great, do it, just be careful to understand what is going on with the transactions.

PR #3097, opened 2024-10-28
[CORE-53] Fix storage request errors on requester pays workspaces

trholdridge:
Ticket: [https://broadworkbench.atlassian.net/browse/CORE-53](https://broadworkbench.atlassian.net/browse/CORE-53)

While working through this ticket, Adam and I found that the current requester pays behavior for storage cost + region are inconsistent in the UI. For both features, the only information needed from Google storage service is the bucket location (which is used in the cost calculation to identify rates). However, the current implementation has the logic for estimated cost in Orchestration, and the logic for displaying bucket region in Terra UI. Both do the Google storage service request and error handling independently.

This change adds bucket location info to the `bucketOptions` returned by the workspace details API. It also adds an optional `userProject` parameter which can be used when getting `bucketOptions`. There is some new validation, too: if a reader is trying to get `bucketOptions` on a requester pays workspace without providing a project to bill, this is considered to be a bad request. In all other cases, a `userProject` is not required (the request will be billed to the workspace's google project, as it was previously).

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3194,6 +3194,13 @@ paths:
       parameters:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
         - $ref: '#/components/parameters/workspaceNamePathParam'
+        - name: userProject
+          in: query
+          description: |
+            When specified, bill google storage requests to this project
+          required: false
+          schema:
+            type: string
       responses:
         200:
           description: Successful Request
@@ -3386,6 +3393,13 @@ paths:
             type: array
             items:
               type: string
+        - name: userProject
+          in: query
+          description: |
+            When specified, bill google storage requests to this project
+          required: false
+          schema:
+            type: string
       responses:
         200:
           description: Successful Request
@@ -3428,6 +3442,13 @@ paths:
             type: array
             items:
               type: string
+        - name: userProject
+          in: query
+          description: |
+            When specified, bill google storage requests to this project
+          required: false
+          schema:
+            type: string
       responses:
         200:
           description: Successful Request
@@ -6325,11 +6346,15 @@ components:
     WorkspaceBucketOptions:
       required:
         - requesterPays
+        - location
       type: object
       properties:
         requesterPays:
           type: boolean
           description: Whether the bucket is requester pays
+        location:
+          type: string
+          description: The region the bucket is in
       description: Extra information about a GCS bucket attached to a workspace
     WorkspaceAccessLevel:
       type: string
--- a/core/src/main/resources/swagger/api-docs.yaml
+++ b/core/src/main/resources/swagger/api-docs.yaml
@@ -3194,13 +3194,7 @@ paths:
       parameters:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
         - $ref: '#/components/parameters/workspaceNamePathParam'
-        - name: userProject
-          in: query
-          description: |
-            When specified, bill google storage requests to this project
-          required: false
-          schema:
-            type: string
+        - $ref: '#/components/parameters/userProjectQueryParam'
       responses:
         200:
           description: Successful Request
@@ -3393,13 +3387,7 @@ paths:
             type: array
             items:
               type: string
-        - name: userProject
-          in: query
-          description: |
-            When specified, bill google storage requests to this project
-          required: false
-          schema:
-            type: string
+        - $ref: '#/components/parameters/userProjectQueryParam'
       responses:
         200:
           description: Successful Request
@@ -3442,13 +3430,7 @@ paths:
             type: array
             items:
               type: string
-        - name: userProject
-          in: query
-          description: |
-            When specified, bill google storage requests to this project
-          required: false
-          schema:
-            type: string
+        - $ref: '#/components/parameters/userProjectQueryParam'
       responses:
         200:
           description: Successful Request
@@ -7236,6 +7218,14 @@ components:
       in: path
       description: Submission Id
       required: true
+      schema:
+        type: string
+    userProjectQueryParam:
+      name: userProject
+      in: query
+      description: |
+        When specified, bill google storage requests to this project
+      required: false
       schema:
         type: string
     workbenchRolePathParam:
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -1100,11 +1100,11 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
         billing <- Option(bucketDetails.getBilling)
         rp <- Option(billing.getRequesterPays)
       } yield rp.booleanValue()
-      val location = bucketDetails.getLocation
+      val location = Option(bucketDetails.getLocation)
 
       WorkspaceBucketOptions(
         requesterPays = requesterPays.getOrElse(false),
-        location = location
+        location = location.getOrElse("")
       )
     }
   }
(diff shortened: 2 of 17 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1100,9 +1100,11 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
         billing <- Option(bucketDetails.getBilling)
         rp <- Option(billing.getRequesterPays)
```
trholdridge, line 1101: The requester pays info in `bucketOptions` is still requested directly from Google. However, in the process of checking whether a user project is needed to make the request, I use the workspace settings to check whether a workspace is requester pays. This makes the Google request a bit redundant...

My other idea was to make the storage service request without checking whether the workspace is requester pays, and then handle a requester pays error by prompting the user to provide a project to bill. But part of the existing problem is that it never actually does error---currently, readers can just make storage requests and it will bill to the workspace's project (that's not the desired behavior). I think it's because of how the service accounts are used, but I am not certain. So I went with the pre-Google-request validation.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6325,11 +6328,15 @@ components:
     WorkspaceBucketOptions:
       required:
         - requesterPays
+        - location
```
kevinmarete, line 6331: Now that location is a required response in BucketOptions, we need to check/update rawls producer contract tests so as not to break them.

kevinmarete (approved):
I am not a scala expert but everything I have reviewed seems in order.

core/src/main/resources/swagger/api-docs.yaml
trholdridge: Do you mean that for other services where Rawls is the provider, their contracts will have to be updated once the new Rawls version is incorporated? Or do I have to make a manual change in Rawls somewhere?

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3194,6 +3194,7 @@ paths:
       parameters:
         - $ref: '#/components/parameters/workspaceNamespacePathParam'
         - $ref: '#/components/parameters/workspaceNamePathParam'
+        - $ref: '#/components/parameters/userProjectQueryParam'
```
davidangb, line 3197: as experienced in https://broadworkbench.atlassian.net/browse/CORE-95, this addition will be a breaking change for any generated Rawls clients. I believe some of them are out there, even if they are not supported products. Is this ok?

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
```diff
@@ -67,13 +67,16 @@ trait WorkspaceApiService extends UserInfoDirectives {
...
         path("workspaces" / "id" / Segment) { workspaceId =>
           get {
-            parameterSeq { allParams =>
-              complete {
-                workspaceServiceConstructor(ctx).getWorkspaceById(workspaceId,
-                                                                  WorkspaceFieldSpecs.fromQueryParams(allParams,
-                                                                                                      "fields"
-                                                                  )
-                )
+            parameters("userProject".optional) { userProject =>
```
davidangb, line 70: I believe you can do:
```scala
parameters("userProject".as[GoogleProjectId].optional)
```
and skip the code later down which maps the String into GoogleProjectId.

I haven't tried this myself so it might not actually work!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
davidangb: This makes sense to me as-is. I would be more inclined to store the bucket's location in some Rawls table and eliminate the call to Google entirely … but this would be a roadmap item outside of this PR.

_Maybe_ since we already have the requester-pays setting from workspace settings, and we also have the live requester-pays value from Google, we could use this opportunity to verify them against each other and make sure our workspace settings reflect reality. But we'd only need this if we have worries about data integrity.

davidangb (commented):
one question about changing signatures for any generated clients out there in the wild - that might be a question for Product. Otherwise looks good to me, one minor and entirely optional comment inline!

core/src/main/resources/swagger/api-docs.yaml
kevinmarete: I meant the provider state should be updated if the responses are changing. 
On quick review of [Rawls provider tests](https://github.com/broadinstitute/rawls/blob/develop/pact4s/src/test/scala/org/broadinstitute/dsde/rawls/provider/RawlsProviderSpec.scala), there doesn't seem to be a provider test for Bucket Options. Therefore there may be no need for you to do anything!

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiService.scala
trholdridge: I did actually try that originally! It resulted in some sort of JSON parse error, I think because `GoogleProjectId` *has* a `value: String` as opposed to *being* a String... though I'm not quite proficient enough in Scala to know if that explanation makes sense.

davidangb: quite possible; not a big deal here. Feel free to ignore this comment.

davidangb (approved):

davidangb (approved):
nice addition, thank you!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1100,9 +1100,11 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
...
       } yield rp.booleanValue()
+      val location = bucketDetails.getLocation
```
marctalbott: Consider wrapping `bucketDetails.getLocation` in an `Option`. I expect that buckets will always have a location, but it's good practice to wrap values we get from Google client libraries (and Java libraries more generally) to protect against NPEs

kevinmarete (approved):
Nice work, just fix those failing tests!

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1407,9 +1422,43 @@ class WorkspaceService(
...
+                                     workspaceContext.workspaceId,
+                                     SamWorkspaceActions.write,
+                                     ctx
+    )
+    requesterPays <- workspaceSettingsRepository
+      .getWorkspaceSettings(workspaceContext.workspaceIdAsUUID)
+      .map(_.exists {
+        case GcpBucketRequesterPaysSetting(config) => config.enabled
+        case _                                     => false
+      })
```
marctalbott: Totally fine to keep this out of scope, but it might be worth adding a new method to `WorkspaceSettingRepository` to load a specific workspace setting if it's present on the workspace.

```def getSettingForWorkspace(workspaceId: UUID, settingType: WorkspaceSettingType): Future[Option[WorkspaceSetting]]```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
marctalbott: I'd prefer to pass the user's credentials on this call to Google so we can let Google decide whether they have the necessary access to use the Google project for requester pays. This will guarantee that Terra stays in sync with Google and does not allow users to bill Google projects that they should not be able to. It has the added benefit of simplifying the `getBucketOptions` logic -- we no longer have to connect the Google project to a workspace and check the user's permissions on the workspace, we can just pass the requested Google project along and surface the Google error if needed.

trholdridge: After talking about this with Marc and Doug, we decided to keep it as-is.

There are a few wrinkles to using user accounts/pet accounts for credentials, but the main one is that only writers with compute would be able to utilize requester pays workspaces. This would probably make for a confusing user experience; even writers without compute are normally able to incur other types of spend, and it would be weird to cut them off from doing so on requester pays buckets.

marctalbott (approved):

PR #3107, opened 2024-10-31
[CORE-40] Add cross-billing project spend report API

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-40

Adds a new spend reporting API that gets a spend report for each workspace the user has owner access to.  Returns a `SpendReportingResult`, with each workspace broken down by category, and a summary for the total of all workspaces.  Includes offset and pageSize parameters for paginating.
In the case of workspaces with non-UUID resourceIds or billing projects without billing accounts, they are simply dismissed without error and not included in the report.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -519,9 +519,10 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       }
     }
 
-  override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
-                                        action: SamResourceAction,
-                                        ctx: RawlsRequestContext
+  override def listResourcesWithRolesOrActions(resourceTypeName: SamResourceTypeName,
+                                               actions: Seq[SamResourceAction],
+                                               roles: Seq[SamResourceRole],
+                                               ctx: RawlsRequestContext
   ): Future[Seq[SamUserResource]] =
     retry(when401or5xx) { () =>
       val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
@@ -530,8 +531,8 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
         /* format = */ "hierarchical",
         /* resourceTypes = */ util.List.of(resourceTypeName.value),
         /* policies = */ util.List.of(),
-        /* roles = */ util.List.of,
-        /* actions = */ util.List.of(action.value),
+        /* roles = */ roles.map(_.toString).asJava,
+        /* actions = */ actions.map(_.toString).asJava,
         /* includePublic = */ false,
         callback
       )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -16,7 +16,9 @@ import org.broadinstitute.dsde.workbench.client.sam
 import org.broadinstitute.dsde.workbench.client.sam.api._
 import org.broadinstitute.dsde.workbench.client.sam.model.{
   FilteredFlatResourcePolicy,
+  FilteredHierarchicalResource,
   FilteredHierarchicalResourcePolicy,
+  FilteredResourcesHierarchicalResponse,
   ListResourcesV2200Response
 }
 import org.broadinstitute.dsde.workbench.client.sam.{ApiCallback, ApiClient, ApiException}
@@ -522,7 +524,7 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
   override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
                                         action: SamResourceAction,
                                         ctx: RawlsRequestContext
-  ): Future[Seq[SamUserResource]] =
+  ): Future[Seq[FilteredHierarchicalResource]] =
     retry(when401or5xx) { () =>
       val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
 
@@ -540,20 +542,6 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
         resourcesResponse.getFilteredResourcesHierarchicalResponse
           .getResources()
           .asScala
-          .map { resource =>
-            val resourcePolicies = resource.getPolicies.asScala.toList
-            val publicPolicies = resourcePolicies.filter(_.getIsPublic)
-            // might need to filter public policies out of this next statement:
-            val (inheritedPolicies, directPolicies) = resourcePolicies.partition(_.getInherited)
-            SamUserResource(
-              resource.getResourceId,
-              toSamRolesAndActions(directPolicies),
-              toSamRolesAndActions(inheritedPolicies),
-              toSamRolesAndActions(publicPolicies),
-              resource.getAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet,
-              resource.getMissingAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet
-            )
-          }
           .toSeq
       }
     }
@@ -563,12 +551,6 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       rolesAndActions.getRoles.asScala.map(SamResourceRole).toSet,
       rolesAndActions.getActions.asScala.map(SamResourceAction).toSet
     )
-
-  private def toSamRolesAndActions(policies: scala.collection.immutable.List[FilteredHierarchicalResourcePolicy]) = {
-    val roles = policies.flatMap(_.getRoles.asScala)
-    val actions = policies.flatMap(_.getActions.asScala)
-    SamRolesAndActions(roles.map(role => SamResourceRole(role.toString)).toSet, actions.map(SamResourceAction).toSet)
-  }
 
   override def getPetServiceAccountKeyForUser(googleProject: GoogleProjectId,
                                               userEmail: RawlsUserEmail
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -526,7 +526,7 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
 
       resourcesApi(ctx).listResourcesV2Async(
-        /* format = */ "hierarchical", // Todo what to pass in here
+        /* format = */ "hierarchical",
         /* resourceTypes = */ util.List.of(resourceTypeName.value),
         /* policies = */ util.List.of(),
         /* roles = */ util.List.of,
@@ -536,7 +536,6 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       )
 
       callback.future.map { resourcesResponse =>
-        println(resourcesResponse)
         resourcesResponse.getFilteredResourcesHierarchicalResponse
           .getResources()
           .asScala
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.rawls.util.{FutureSupport, Retry}
 import org.broadinstitute.dsde.workbench.client.sam
 import org.broadinstitute.dsde.workbench.client.sam.api._
 import org.broadinstitute.dsde.workbench.client.sam.model.{
+  FilteredFlatResourcePolicy,
   FilteredHierarchicalResourcePolicy,
   ListResourcesV2200Response
 }
@@ -540,11 +541,15 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
           .getResources()
           .asScala
           .map { resource =>
+            val resourcePolicies = resource.getPolicies.asScala.toList
+            val publicPolicies = resourcePolicies.filter(_.getIsPublic)
+            // might need to filter public policies out of this next statement:
+            val (inheritedPolicies, directPolicies) = resourcePolicies.partition(_.getInherited)
             SamUserResource(
               resource.getResourceId,
-              toSamRolesAndActions(resource.getPolicies()), // TODO what to use here?
-              toSamRolesAndActions(resource.getPolicies()), // What are all these three things?
-              toSamRolesAndActions(resource.getPolicies()),
+              toSamRolesAndActions(directPolicies),
+              toSamRolesAndActions(inheritedPolicies),
+              toSamRolesAndActions(publicPolicies),
               resource.getAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet,
               resource.getMissingAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet
             )
@@ -559,10 +564,9 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       rolesAndActions.getActions.asScala.map(SamResourceAction).toSet
     )
 
-  private def toSamRolesAndActions(policies: util.List[FilteredHierarchicalResourcePolicy]) = {
-    val scalaPolicies = policies.asScala.toList
-    val roles = scalaPolicies.flatMap(_.getRoles.asScala)
-    val actions = scalaPolicies.flatMap(_.getActions.asScala)
+  private def toSamRolesAndActions(policies: scala.collection.immutable.List[FilteredHierarchicalResourcePolicy]) = {
+    val roles = policies.flatMap(_.getRoles.asScala)
+    val actions = policies.flatMap(_.getActions.asScala)
     SamRolesAndActions(roles.map(role => SamResourceRole(role.toString)).toSet, actions.map(SamResourceAction).toSet)
   }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -534,7 +534,7 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
         /* policies = */ util.List.of(),
         /* roles = */ util.List.of(),
         /* actions = */ util.List.of(action.value),
-        /* includePublic = */ false,
+        /* includePublic = */ true,
         callback
       )
 
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.rawls.util.{FutureSupport, Retry}
 import org.broadinstitute.dsde.workbench.client.sam
 import org.broadinstitute.dsde.workbench.client.sam.api._
 import org.broadinstitute.dsde.workbench.client.sam.model.{
+  FilteredFlatResource,
   FilteredFlatResourcePolicy,
   FilteredHierarchicalResource,
   FilteredHierarchicalResourcePolicy,
@@ -524,12 +525,12 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
   override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
                                         action: SamResourceAction,
                                         ctx: RawlsRequestContext
-  ): Future[Seq[FilteredHierarchicalResource]] =
+  ): Future[Seq[FilteredFlatResource]] =
     retry(when401or5xx) { () =>
       val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
 
       resourcesApi(ctx).listResourcesV2Async(
-        /* format = */ "hierarchical",
+        /* format = */ "flat",
         /* resourceTypes = */ util.List.of(resourceTypeName.value),
         /* policies = */ util.List.of(),
         /* roles = */ util.List.of(),
@@ -539,7 +540,7 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       )
 
       callback.future.map { resourcesResponse =>
-        resourcesResponse.getFilteredResourcesHierarchicalResponse
+        resourcesResponse.getFilteredResourcesFlatResponse
           .getResources()
           .asScala
           .toSeq
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -519,10 +519,9 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       }
     }
 
-  override def listResourcesWithRolesOrActions(resourceTypeName: SamResourceTypeName,
-                                               actions: Seq[SamResourceAction],
-                                               roles: Seq[SamResourceRole],
-                                               ctx: RawlsRequestContext
+  override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
+                                        action: SamResourceAction,
+                                        ctx: RawlsRequestContext
   ): Future[Seq[SamUserResource]] =
     retry(when401or5xx) { () =>
       val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
@@ -531,8 +530,8 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
         /* format = */ "hierarchical",
         /* resourceTypes = */ util.List.of(resourceTypeName.value),
         /* policies = */ util.List.of(),
-        /* roles = */ roles.map(_.toString).asJava,
-        /* actions = */ actions.map(_.toString).asJava,
+        /* roles = */ util.List.of(),
+        /* actions = */ util.List.of(action.value),
         /* includePublic = */ false,
         callback
       )
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
@@ -14,12 +14,17 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.rawls.util.{FutureSupport, Retry}
 import org.broadinstitute.dsde.workbench.client.sam
 import org.broadinstitute.dsde.workbench.client.sam.api._
+import org.broadinstitute.dsde.workbench.client.sam.model.{
+  FilteredHierarchicalResourcePolicy,
+  ListResourcesV2200Response
+}
 import org.broadinstitute.dsde.workbench.client.sam.{ApiCallback, ApiClient, ApiException}
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroupName}
 
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 import java.util
+import java.util.List
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.jdk.CollectionConverters._
@@ -513,11 +518,54 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
       }
     }
 
+  override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
+                                        action: SamResourceAction,
+                                        ctx: RawlsRequestContext
+  ): Future[Seq[SamUserResource]] =
+    retry(when401or5xx) { () =>
+      val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
+
+      resourcesApi(ctx).listResourcesV2Async(
+        /* format = */ "hierarchical", // Todo what to pass in here
+        /* resourceTypes = */ util.List.of(resourceTypeName.value),
+        /* policies = */ util.List.of(),
+        /* roles = */ util.List.of,
+        /* actions = */ util.List.of(action.value),
+        /* includePublic = */ false,
+        callback
+      )
+
+      callback.future.map { resourcesResponse =>
+        println(resourcesResponse)
+        resourcesResponse.getFilteredResourcesHierarchicalResponse
+          .getResources()
+          .asScala
+          .map { resource =>
+            SamUserResource(
+              resource.getResourceId,
+              toSamRolesAndActions(resource.getPolicies()), // TODO what to use here?
+              toSamRolesAndActions(resource.getPolicies()), // What are all these three things?
+              toSamRolesAndActions(resource.getPolicies()),
+              resource.getAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet,
+              resource.getMissingAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet
+            )
+          }
+          .toSeq
+      }
+    }
+
   private def toSamRolesAndActions(rolesAndActions: sam.model.RolesAndActions) =
     SamRolesAndActions(
       rolesAndActions.getRoles.asScala.map(SamResourceRole).toSet,
       rolesAndActions.getActions.asScala.map(SamResourceAction).toSet
     )
+
+  private def toSamRolesAndActions(policies: util.List[FilteredHierarchicalResourcePolicy]) = {
+    val scalaPolicies = policies.asScala.toList
+    val roles = scalaPolicies.flatMap(_.getRoles.asScala)
+    val actions = scalaPolicies.flatMap(_.getActions.asScala)
+    SamRolesAndActions(roles.map(role => SamResourceRole(role.toString)).toSet, actions.map(SamResourceAction).toSet)
+  }
 
   override def getPetServiceAccountKeyForUser(googleProject: GoogleProjectId,
                                               userEmail: RawlsUserEmail
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
@@ -350,6 +350,16 @@ trait RawlsBillingProjectComponent {
         .result
         .map(_.headOption.map(RawlsBillingProjectRecord.toBillingProjectSpendExport))
 
+    def getBillingProjectsSpendConfiguration(
+      billingProjectNames: Seq[RawlsBillingProjectName]
+    ): ReadAction[Seq[Option[BillingProjectSpendExport]]] =
+      rawlsBillingProjectQuery
+        .withProjectNames(billingProjectNames)
+        .result
+        .map(projectRecords =>
+          projectRecords.map(record => Some(RawlsBillingProjectRecord.toBillingProjectSpendExport(record)))
+        )
+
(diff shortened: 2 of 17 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -237,8 +242,79 @@ class SpendReportingService(
...
+                       |    spend_category""".stripMargin.trim
+
+    val bpSubQuery = billingProjects
+      .map { bp =>
+        val tableName = bp._1.spendExportTable.getOrElse(spendReportingServiceConfig.defaultTableName)
+        val timePartitionColumn: String = getTimePartitionColumn(tableName)
+        baseQuery
+          .replace("_PARTITIONTIME", timePartitionColumn)
+          .replace("_BILLING_ACCOUNT_TABLE", tableName)
+          .replace("_PROJECT_ID_LIST", "(" + bp._2.mkString(", ") + ")") + "\nUNION ALL"
```
davidangb: do you need any quotes around the individual billing project ids here?
```
(foo, bar, baz)
```
vs
```
('foo', 'bar', 'baz')
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +435,69 @@ class SpendReportingService(
...
+    }
+  }
+
+  def getOwnerWorkspaces(): Future[Seq[WorkspaceListResponse]] =
+    workspaceServiceConstructor(ctx).listWorkspaces(WorkspaceFieldSpecs(), -1) map {
+      case JsArray(jsArray) =>
+        val workspaces = jsArray.map(_.convertTo[WorkspaceListResponse])
+        workspaces.filter(_.accessLevel == Owner)
+      case _ => throw new IllegalArgumentException("Expected a JsArray")
+    }
```
davidangb: Can this avoid using `listWorkspaces`? A couple reasons for this:
1. For many users, it will query for and calculate a lot more workspaces than you actually need; you're throwing away a lot of data
2. The `accessLevel` check is not ideal: https://github.com/broadinstitute/rawls/blob/995cd5571ca87abe6a25e23b9148be725e71912b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L291-L293

edited to add: is this even used? I only see it called from a unit test.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +435,69 @@ class SpendReportingService(
...
+      groupedWorkspaces.map { case (namespace, wsList) =>
+        getSpendExportConfiguration(RawlsBillingProjectName(namespace)).map { exportConfig =>
+          Some(
+            exportConfig -> wsList
+              .map(_.workspace.googleProject)
+          )
+        }
+      }
+    Future.sequence(billingFutures).map(_.flatten.toMap)
+  }
```
davidangb: I also see this method only called from a unit test. Is it used? If so: this method loops over each namespace and gets its spend config via `getSpendExportConfiguration`. Each invocation of this opens a new db transaction, queries for the config for a single namespace, and completes that transaction.

Could you instead write a new method that accepts a list of namespaces and makes one SQL query for all those namespaces and returns their result?

davidangb (commented):
initial feedback!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: It will be used!  I'm just constructing helper methods piece by piece and making sure they work before I get to the part that uses it!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: Same answer re: it being used.  I will try to rewrite.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: @davidangb ....Any advice on what to use instead of `listWorkspaces`?  I could go directly to Sam, but that's what `listWorkspaces` does, so I'd have to re-use a lot of code.

davidangb: I am not aware of a single function to do this. I think you'd have to call Sam to list the resources on which you have the appropriate action, and then use something like `WorkspaceRepository.listWorkspacesByIds` to get the details of those workspaces.

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -59,177 +59,195 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
     requireUserInfo(Option(otelContext)) { userInfo =>
       val ctx = RawlsRequestContext(userInfo, Option(otelContext))
       pathPrefix("billing" / "v2") {
-
-        pathPrefix(Segment) { projectId =>
-          pathEnd {
+        pathPrefix("spendReport") {
```
calypsomatic, line 62: This file must've been hit by `scalafmt` for some reason; adding in this new `spendReport` route is all I did

core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/ApiServiceSpec.scala
```diff
@@ -417,6 +405,19 @@ trait ApiServiceSpec
                                   samDAO
       )
 
+    val spendReportingBigQueryService = bigQueryServiceFactory.getServiceFromJson("json", GoogleProject("test-project"))
```
calypsomatic, line 408: This is just moved after the creation of `workspaceServiceConstructor` so I can pass it in as a parameter

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: Looking into this more closely, I can call `listWOrkspaces` with a param asking for only those the user owns.  That will save some re-calculation.  If I then check permissions correctly when getting the spend data, will I be safe?  Or would there be a situation where the user has permission to see the spend report for a billing project, but there's a workspace in that project they don't own and shouldn't see name and billing info on?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -350,6 +350,16 @@ trait RawlsBillingProjectComponent {
...
+    def getBillingProjectsSpendConfiguration(
+      billingProjectNames: Seq[RawlsBillingProjectName]
+    ): ReadAction[Seq[Option[BillingProjectSpendExport]]] =
+      rawlsBillingProjectQuery
+        .withProjectNames(billingProjectNames)
+        .result
+        .map(projectRecords =>
+          projectRecords.map(record => Some(RawlsBillingProjectRecord.toBillingProjectSpendExport(record)))
+        )
+
```
davidangb, line 384: is it important for this to return `BillingProjectSpendExport` wrapped inside an `Option`? It doesn't look like there's anything optional; it always uses `Some`. You might be able to simplify this to:
```scala
    def getBillingProjectsSpendConfiguration(
      billingProjectNames: Seq[RawlsBillingProjectName]
    ): ReadAction[Seq[BillingProjectSpendExport]] =
      rawlsBillingProjectQuery
        .withProjectNames(billingProjectNames)
        .result
        .map(projectRecords => projectRecords.map(RawlsBillingProjectRecord.toBillingProjectSpendExport))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +140,65 @@ object SpendReportingService {
...
+          storageCost,
+          "0",
+          currency,
+          Option(start),
+          Option(end),
+          workspace = Some(workspaceName),
+          googleProjectId = Some(GoogleProject(projectId.value)),
+          category = Some(TerraSpendCategories.Storage)
+        )
+      )
```
davidangb: You might be able to clean this up a bit, like:
```scala
val baseSpend = SpendReportingForDateRange(
        "0",
        "0", // Ignoring credits for now; do we want to include them?
        currency,
        Option(start),
        Option(end),
        workspace = Some(workspaceName),
        googleProjectId = Some(GoogleProject(projectId.value))
      )

      // Each row gets a summary of compute, storage, and total
      List(
        baseSpend.copy(cost = totalCost),
        baseSpend.copy(cost = computeCost, category = Some(TerraSpendCategories.Compute)),
        baseSpend.copy(cost = storageCost, category = Some(TerraSpendCategories.Storage)),
      )
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +516,58 @@ class SpendReportingService(
...
+    end: DateTime
+  ): Future[List[SpendReportingResults]] = {
+    validateReportParameters(start, end)
+    for {
+      workspaces <- getOwnerWorkspaces()
+      projectNames = workspaces
+        .map(wsResp =>
+          wsResp.workspace.googleProject -> WorkspaceName(wsResp.workspace.namespace, wsResp.workspace.name)
+        )
+        .toMap // Map[GoogleProjectId, WorkspaceName]
```
davidangb: For others watching the PR: this is where Bria and I discussed changing the business logic. Is it feasible to:
1. Query Sam's `listResourcesV2` for all `billing-project` resources on which the user has the `read_spend_report` action
2. Query the Rawls db to retrieve the workspace names/ids associated with each of those billing projects
3. (maybe) Query Sam's `listResourcesV2` for all `workspace` resources on which the user has the `read` action, to verify that the user should be able to see all the workspace names from step 2

Is there ever a case in which a user will have `read_spend_report` on a `billing-project` but cannot read the workspace associated with that project?

core/src/test/scala/org/broadinstitute/dsde/rawls/mock/MockSamDAO.scala
```diff
@@ -239,6 +239,46 @@ class MockSamDAO(dataSource: SlickDataSource)(implicit executionContext: Executi
       case _ => Future.successful(Seq.empty)
     }
 
+  override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
+                                        action: SamResourceAction,
+                                        ctx: RawlsRequestContext
+  ): Future[Seq[SamUserResource]] =
+    resourceTypeName match {
```
calypsomatic, line 247: The content of this mock method is mostly copied from `listUserResources`, with just the action included.  Not sure if we need anything better

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -513,12 +518,54 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
...
+      callback.future.map { resourcesResponse =>
+        resourcesResponse.getFilteredResourcesHierarchicalResponse
+          .getResources()
+          .asScala
+          .map { resource =>
+            SamUserResource(
+              resource.getResourceId,
+              toSamRolesAndActions(resource.getPolicies()), // TODO what to use here?
+              toSamRolesAndActions(resource.getPolicies()), // What are all these three things?
+              toSamRolesAndActions(resource.getPolicies()),
```
davidangb: Each policy in `getPolicies()` has info about whether it is public, direct, or inherited. I think you can use that info to fill in the details of `SamUserResource`, like:
```scala
          .map { resource =>
            val resourcePolicies = resource.getPolicies.asScala.toList
            val publicPolicies = resourcePolicies.filter(_.getIsPublic)
            // might need to filter public policies out of this next statement:
            val (inheritedPolicies, directPolicies) = resourcePolicies.partition(_.getInherited)

            SamUserResource(
              resource.getResourceId,
              toSamRolesAndActions(directPolicies.asJava), // TODO what to use here?
              toSamRolesAndActions(inheritedPolicies.asJava), // What are all these three things?
              toSamRolesAndActions(publicPolicies.asJava),
```
if you do this you probably also want to change `toSamRolesAndActions` to accept a Scala list instead of the repeated `.asJava` and `.asScala`

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +539,49 @@ class SpendReportingService(
...
+      gcpOnlyGroupedWorkspaces = groupedWorkspaces
+        .map { case (key, workspaces) =>
+          key -> workspaces.filter(_.workspaceType == WorkspaceType.RawlsWorkspace)
+        }
+        .filter { case (_, workspaces) => workspaces.nonEmpty }
+      // Only use the BPs we know exist in the DB and are GCP
+      spendConfigs <- getSpendExportConfigurations(gcpOnlyGroupedWorkspaces.keys.toList)
+    } yield spendConfigs.map { config =>
+      config -> gcpOnlyGroupedWorkspaces(config.billingProjectName).map(ws => (ws.googleProjectId, ws.toWorkspaceName))
+    }.toMap
```
davidangb: This implementation makes sense to me, and matches what we talked about. I don't know, though, if it aligns with the letter of the jira ticket "all workspaces where the user has owner access permissions or higher". Is that an important Product requirement? Or does this implementation make even more sense than the requirement?

Also … this should be paginated, according to the jira ticket

marctalbott: IIRC it was an important product requirement to expand the access to spend reports to workspace owners, but I'll let @admull confirm that. The `readSpendReport` action is only granted to billing project owners, so this implementation definitely does make sense and is more in line with the access controls on the existing spend report API, but I think we wanted more folks to be able to see how much money they were spending

admull: Yes, the purpose of this feature is to give someone access to be able to see spend for any workspace where they are an Owner or above.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -513,12 +518,54 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
...
 
+  override def listResourcesWithActions(resourceTypeName: SamResourceTypeName,
+                                        action: SamResourceAction,
+                                        ctx: RawlsRequestContext
+  ): Future[Seq[SamUserResource]] =
+    retry(when401or5xx) { () =>
+      val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
+
+      resourcesApi(ctx).listResourcesV2Async(
+        /* format = */ "hierarchical",
```
dvoet:
```suggestion
        /* format = */ "flat",
```
`hierarchical` will not return actions that are included in roles

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +539,49 @@ class SpendReportingService(
...
+    for {
+      billingProjectResources <- samDAO.listResourcesWithActions(SamResourceTypeNames.billingProject,
+                                                                 SamBillingProjectActions.readSpendReport,
+                                                                 ctx
+      )
+      billingProjectIds = billingProjectResources.map(resource => RawlsBillingProjectName(resource.resourceId)).toList
+      groupedWorkspaces <- workspaceServiceConstructor(ctx).getWorkspacesByBillingProjects(billingProjectIds)
+      gcpOnlyGroupedWorkspaces = groupedWorkspaces
+        .map { case (key, workspaces) =>
+          key -> workspaces.filter(_.workspaceType == WorkspaceType.RawlsWorkspace)
```
dvoet: does it make sense to do this filtering in the db query?

calypsomatic: I almost did that, then convinced myself that it was better to do it outside the db query for reasons I can't remember.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
dvoet: > see spend for any workspace where they are an Owner or above

The current code grants access for all billing projects where the caller has the readSpendReport action. This seems appropriate and we should make sure that correct roles have this action. But this is not quite @admull's comment which talks about _workspace_ owners.

calypsomatic: So a user could have `readSpendReport` on a billing project they don't own, correct?  So I think it makes sense to start with only the billing projects they're able to read spend report on, and then I suppose I could filter the resulting workspaces for owner and above.  So I think I'm just missing that second step

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
calypsomatic: Hm, with hierarchical I get a `policies` object like so:
```
     "policies": [
        {
          "actions": [],
          "inherited": false,
          "isPublic": false,
          "policy": "owner",
          "roles": [
            {
              "actions": [
                "read_spend_report"
              ],
              "role": "owner"
            }
          ]
        }
      ]
```
but with `flat` it's just:
```
      "policies": [
        {
          "inherited": false,
          "isPublic": false,
          "policy": "owner"
        }
      ]
```
Looks like I want hierarchical?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: Ok I've updated it to compare workspace ownership to `readSpendReport` permission on the billing project - by making a call to sam for BPs and another one for workspaces and intersecting the results.  Not sure if this is what we want or if it's efficient, so please let me know!

marctalbott: Sorry, missed your earlier comment! Currently, users can only have the `readSpendReport` action on billing projects that they own. Since we want to let anyone with Owner access or above on a workspace see the spendReport, we should either add a new `readSpendReport` action to the `owner` role on the Sam `workspace` resource type or check the `own` action on the Sam resource

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -11,7 +11,10 @@ import org.broadinstitute.dsde.rawls.coordination.{DataSourceAccess, Uncoordinat
 import org.broadinstitute.dsde.rawls.dataaccess._
 import org.broadinstitute.dsde.rawls.dataaccess.slick.{TestDriverComponent, WorkflowRecord}
 import org.broadinstitute.dsde.rawls.expressions.{BoundOutputExpression, OutputExpression}
-import org.broadinstitute.dsde.rawls.jobexec.SubmissionMonitorActor.{ExecutionServiceStatusResponse, StatusCheckComplete}
+import org.broadinstitute.dsde.rawls.jobexec.SubmissionMonitorActor.{
```
calypsomatic, line 14: I didn't make any changes to this file, it just somehow got mangled by scalafmt

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -513,12 +519,57 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
...
+            val (inheritedPolicies, directPolicies) = resourcePolicies.partition(_.getInherited)
+            SamUserResource(
+              resource.getResourceId,
+              toSamRolesAndActions(directPolicies),
+              toSamRolesAndActions(inheritedPolicies),
+              toSamRolesAndActions(publicPolicies),
+              resource.getAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet,
+              resource.getMissingAuthDomainGroups.asScala.map(WorkbenchGroupName).toSet
+            )
+          }
```
davidangb: after figuring out if this should specify hierarchical or flat on the request to Sam … should this method do all the translation of that response into `SamUserResource`? If this method returned `Future[Seq[FilteredFlatResource]]` or `Future[Seq[FilteredHierarchicalResource]]` as-is, it would require less logic. In your use case I don't think the structure matters - you only use the resource id, right?

If some other caller in the future needed the result of this method as `SamUserResource`, we could write a translator at that point.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
```diff
@@ -260,6 +261,21 @@ trait WorkspaceComponent {
     def listWithBillingProject(billingProject: RawlsBillingProjectName): ReadAction[Seq[Workspace]] =
       workspaceQuery.withBillingProject(billingProject).read
 
+    def groupByBillingProjectOfType(workspaceIds: List[UUID],
+                                    workspaceType: WorkspaceType
+    ): ReadWriteAction[Map[RawlsBillingProjectName, Seq[Workspace]]] = {
+      val query = for {
+        workspace <- workspaceQuery if workspace.id inSet workspaceIds.toSet
+        if workspace.workspaceType === workspaceType.toString
```
davidangb, line 269: 1. always prefer `inSetBind` over `inSet`; `inSetBind` uses bind parameters and is safer. `inSet` inlines values - but takes steps to escape dangerous characters. So they're probably both safe in this case, especially since we know the values are UUIDs, but `inSetBind` is the best-practice.
2. I've never seen the `queryObject … if … if` syntax before. I can only find examples in our code with a single `if`. Do you know what SQL this actually generates? You might be able to use [Slick logging config](https://scala-slick.org/doc/3.3.3/config.html#logging) to output the query, or you could write some debugging code to look at `workspace.result.statements`

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +145,90 @@ object SpendReportingService {
...
+    names: Map[GoogleProjectId, WorkspaceName]
+  ): SpendReportingResults = {
+
+    var total = BigDecimal(0.0)
+    var total_credits = BigDecimal(0.0)
+
+    val currency = allRows.map(_.get("currency").getStringValue).distinct match {
+      case head :: _ => Currency.getInstance(head)
+      case _         => throw RawlsExceptionWithErrorReport(StatusCodes.NotFound, "No currencies found for spend data")
+    }
```
davidangb: Does this need to handle the case where it found multiple currency values? You can match on a single-element list using `head :: Nil` like:
```scala
    val currency = allRows.map(_.get("currency").getStringValue).distinct match {
      case head :: Nil => Currency.getInstance(head)
      ...
    }
```
( but it might be easier to save `allRows.map(_.get("currency").getStringValue).distinct` to a `val` and check its size )

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +145,90 @@ object SpendReportingService {
...
 
+  def extractCrossBillingProjectSpendReportingResults(
+    allRows: List[FieldValueList],
+    start: DateTime,
+    end: DateTime,
+    names: Map[GoogleProjectId, WorkspaceName]
+  ): SpendReportingResults = {
+
+    var total = BigDecimal(0.0)
+    var total_credits = BigDecimal(0.0)
```
davidangb, line 153: nit: add a code comment calling out that these are `var`s, since we try to avoid those

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +145,90 @@ object SpendReportingService {
...
+          getRoundedNumericValue("credits").toString,
+          currencyCode.toString,
+          category = Option(TerraSpendCategories.Storage)
+        ),
+        SpendReportingForDateRange(
+          getRoundedNumericValue("compute_cost").toString,
+          getRoundedNumericValue("credits").toString,
+          currencyCode.toString,
+          category = Option(TerraSpendCategories.Compute)
+        )
```
davidangb, line 200: Each of these `SpendReportingForDateRange` objects include the same value for `getRoundedNumericValue("credits")`. Is this triple-counting credits?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +145,90 @@ object SpendReportingService {
...
+          currencyCode.toString,
+          Option(start),
+          Option(end),
+          category = Option(TerraSpendCategories.Other)
+        ),
+        SpendReportingForDateRange(
+          getRoundedNumericValue("storage_cost").toString,
+          getRoundedNumericValue("credits").toString,
+          currencyCode.toString,
+          category = Option(TerraSpendCategories.Storage)
```
davidangb, line 193: Why does the "other_cost" object have a startTime and endTime, but the "storage_cost" and "compute_cost" ones don't?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -133,6 +145,90 @@ object SpendReportingService {
...
+      )
+
+      SpendReportingAggregation(SpendReportingAggregationKeys.Workspace, List(workspaceTotal))
+
+    }
+
+    val summary = SpendReportingForDateRange(
+      total.toString,
+      total_credits.toString,
+      currency.toString, // TODO: what to do about combined summary for currencies?
```
davidangb: I have the same question! Maybe for this version of the report we need to throw an error if we find multiple currencies.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -237,8 +345,76 @@ class SpendReportingService(
...
+    billingProjects: Map[BillingProjectSpendExport, Seq[(GoogleProjectId, WorkspaceName)]],
+    pageSize: Int,
+    offset: Int
+  ): String = {
+    val baseQuery = s"""
+                       |  SELECT
+                       |    project.id AS project_id,
+                       |    project.name AS project_name,
+                       |    currency,
+                       |    SUM(IFNULL((SELECT SUM(c.amount) FROM UNNEST(credits) c), 0)) as credits,
```
davidangb, line 371: do `credits` also have a description like "Cloud Storage" or "Compute Engine" we could use to categorize them?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -359,4 +552,64 @@ class SpendReportingService(
...
+        _ = if (billingMap.isEmpty) {
+          return Future.successful(None)
+        }
+        projectNames: Map[GoogleProjectId, WorkspaceName] = billingMap.values.flatten.toMap
+        query = getAllUserWorkspaceQuery(billingMap, pageSize, offset)
+        queryJob = setUpAllUserWorkspaceQuery(query, start, end)
+
+        job: Job <- bigQueryService.use(_.runJob(queryJob)).unsafeToFuture().map(_.waitFor())
+        _ = logSpendQueryStats(job.getStatistics[JobStatistics.QueryStatistics])
+        result = job.getQueryResults()
```
davidangb: this would be a great place to add tracing: let's get timing and visibility into how the BigQuery query is performing. Might be easiest to pull this logic out into a separate, traced, method

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -59,177 +59,199 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
+              parameters(
+                "startDate".as[DateTime],
+                "endDate".as[DateTime],
+                "pageSize".as[Int],
+                "offset".as[Int]
+              ) { (startDate, endDate, pageSize, offset) =>
+                complete {
+                  spendReportingConstructor(ctx).getSpendForAllWorkspaces(
+                    startDate,
+                    endDate.plusDays(1).minusMillis(1),
```
davidangb, line 74: fancy!

davidangb (commented):
thoughts inline!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: Probably just a mistake on my part!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: Good point.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: I feel like it shouldn't throw an error - for a single billing project it should be a single currency, but since this covers multiple billing projects it seems plausible that there might be different currencies.  So I think ideally we would either not have a summary, or have a summary per currency?  Not really sure what the best thing to do here is but I don't think it should be error throwing.  We could potentially do that for a first pass and deal with it later on?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -237,8 +345,76 @@ class SpendReportingService(
...
+       |)
+       |SELECT
+       |  project_id,
+       |  project_name,
+       |  SUM(category_cost) AS total_cost,
+       |  SUM(CASE WHEN spend_category = 'Storage' THEN category_cost ELSE 0 END) AS storage_cost,
+       |  SUM(CASE WHEN spend_category = 'Compute' THEN category_cost ELSE 0 END) AS compute_cost,
+       |  SUM(CASE WHEN spend_category = 'Other' THEN category_cost ELSE 0 END) AS other_cost,
+       |  currency,
+       |  SUM(credits) as credits
```
kevinmarete: I guess to David's suggestion above we can split the credits by category similar to the category_costs
```suggestion
    |  SUM(CASE WHEN spend_category = 'Storage' THEN credits ELSE 0 END) AS storage_credits,
    |  SUM(CASE WHEN spend_category = 'Compute' THEN credits ELSE 0 END) AS compute_credits,
    |  SUM(CASE WHEN spend_category = 'Other' THEN credits ELSE 0 END) AS other_credits,
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: I support throwing an error for now and in the future, we can choose provided multiple aggregated reports for each currency [{UsdAggregate}, {EuroAggregate},...etc}]

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceComponent.scala
calypsomatic: Hm looks like it's just `select .... from WORKSPACE where id in (...) and workspace_type = rawls)`,

kevinmarete (approved):
I am not a Scala expert, but I have added a few optional comments for consideration. I tested the API locally, and everything seems to work well. Nice job!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: On closer inspection, I think it shows the same credits three times, one for each category, but only counts it once in the total.  Still less than ideal

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
calypsomatic: I don't know and I don't have the permissions to figure it out

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
```diff
@@ -513,6 +521,31 @@ class HttpSamDAO(baseSamServiceURL: String, rawlsCredential: RawlsCredential, ti
...
+    retry(when401or5xx) { () =>
+      val callback = new SamApiCallback[ListResourcesV2200Response]("listResourcesV2")
+
+      resourcesApi(ctx).listResourcesV2Async(
+        /* format = */ "hierarchical",
+        /* resourceTypes = */ util.List.of(resourceTypeName.value),
+        /* policies = */ util.List.of(),
+        /* roles = */ util.List.of(),
+        /* actions = */ util.List.of(action.value),
+        /* includePublic = */ false,
```
davidangb: will the `includePublic = false` prevent seeing spend reports for public workspaces, even if the calling user is an owner? As in, will our support/comms team be unable to see spend for our public workspaces?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
davidangb: I do see a `type` column defined for the `credits` record. If this `CASE … WHEN` doesn't work out, we could look at the `type` column. I think we'll really just need to test this empirically and see what results it gives.

davidangb (approved):
one last question about public workspaces and a note about accuracy of credits.

Since I'll be out, I'm giving this a proactive thumb. Since it's an API, we can release it and iterate on it before actually releasing user-facing features.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
calypsomatic: Good point, I'll change it to true.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpSamDAO.scala
dvoet: hmmm, this does not seem right, I will look at what sam is doing

dvoet: a full resource in the response for `flat` looks like
```
{
      "actions": [
        "read_spend_report"
      ],
      "authDomainGroups": [],
      "missingAuthDomainGroups": [],
      "policies": [
        {
          "inherited": false,
          "isPublic": false,
          "policy": "owner"
        }
      ],
      "resourceId": "eaacdc66-9ac5-4dd9-aeae-1a2c98f1eda8",
      "resourceType": "spend-profile",
      "roles": [
        "owner"
      ]
    }
```
The actions are not included in the policies. The is probably what it means to be flat.

The full resource for a hierarchical response looks like 
```
{
      "authDomainGroups": [],
      "missingAuthDomainGroups": [],
      "policies": [
        {
          "actions": [],
          "inherited": false,
          "isPublic": false,
          "policy": "owner",
          "roles": [
            {
              "actions": [
                "read_spend_report"
              ],
              "role": "owner"
            }
          ]
        }
      ],
      "resourceId": "7900ff92-a1f0-4c3e-b051-e549e2211e32",
      "resourceType": "spend-profile"
    }
```
The actions are both at policy and role level. Also note there could be multiple policies that contain the action. Flat is really what you want because it hides all the details about how user has the action on the resource.

dvoet: Looking at this further, flat vs hierarchical does not matter because nothing is used of the response other than the resource id. But hierarchical is much harder to use correctly. So I would either have this `listResourcesWithActions` function return only a list of resource ids or switch to flat

dvoet (approved):

PR #3118, opened 2024-11-06
CORE-69: update scala-sbt image

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-69

Update the `scala-sbt` [docker image](https://github.com/sbt/docker-sbt) tag in use to match our current versions of java/sbt/scala.

This should have no functional impact other than speeding up sbt operations. The docker image comes with these versions of java/sbt/scala preinstalled, so it can use them immediately. If our docker image tag and sbt/scala versions are mismatched, it has to download sbt/scala before it can continue.

---

**PR checklist**

### DIFF
--- a/.github/workflows/consumer_contract_tests.yml
+++ b/.github/workflows/consumer_contract_tests.yml
@@ -136,7 +136,7 @@ jobs:
                           -v jar-cache:/root/.ivy \
                           -v jar-cache:/root/.ivy2 \
                           -w /working \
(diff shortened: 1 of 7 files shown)


calypsomatic (approved):
One non-blocking thought - this image appears in enough files that it seems like if we ever want to update it again it would benefit from being a variable set only once somewhere?

davidangb:
>  if we ever want to update it again it would benefit from being a variable set only once somewhere?

I agree - though I didn't want to take on a refactoring like that :(

dvoet (approved):

PR #3119, opened 2024-11-06
CORE-189: streamline ENTITY_ATTRIBUTE_TEMP temp table

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-189

## Incremental performance improvements for batch-upsert operations

The `ENTITY_ATTRIBUTE_TEMP` temp table is used during batch upsert/batch update operations. These operations can include tens or hundreds of thousands of updates to entity attributes. `UPDATE` sql statements cannot be batched, and issuing individual update statements would incur lots and lots of overhead to process each individual statement.

Instead, each "update" is implemented as an `INSERT` statement into the `ENTITY_ATTRIBUTE_TEMP` temp table. The MySQL driver, via `rewriteBatchedStatements=true` [on the connection string](
https://github.com/broadinstitute/terra-helmfile/blob/d70c8da26bcdeb475da33972715f798fee64cb83/charts/rawls/templates/config/_conf.tpl#L201), batches multiple inserts to avoid overhead. Finally, after executing all the inserts into the temp table, we perform a single update of the persistent `ENTITY_ATTRIBUTE_xx_xx` table using values from the temp table. This final update is the [`updateInMaster` function](https://github.com/broadinstitute/rawls/blob/995cd5571ca87abe6a25e23b9148be725e71912b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala#L797-L816).

This PR attempts to optimize the `updateInMaster` function by optimizing its SQL and the schema of the temp table:
* remove `transaction_id` from the `ENTITY_ATTRIBUTE_TEMP` temp table. Since each connection gets its own temp table, this is extraneous and irrelevant and only adds to the complexity of the `where` clause used by this table.
* align the `namespace` and `name` columns of `ENTITY_ATTRIBUTE_TEMP` to what is actually in use by the persistent `ENTITY_ATTRIBUTE_*` shard tables. Specifically, change from `text` to `varchar` and set the same explicit character set and collation. The change from `text` to `varchar` should help with performance.
* remove the extraneous `entity_tmp_owner_id_idx` index from the temp table. This index doesn't get used for anything and therefore only adds overhead on inserts.

the visual explain-plan for the `updateInMaster` function:
explain
I interpret this as:
* for every row in the temp table (the full table scan on `ta`)
* find the matching row in the persistent table (the nested loop and lookup against the `UNQ_ENTITY_ATTRIBUTE` index on `a`)
* and update that row (`query block #1`)

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241106_streamline_entity_attr_temp_table_procedure.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241106_streamline_entity_attr_temp_table_procedure.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
+
+    <changeSet logicalFilePath="dummy" author="davidan" id="streamline_entity_attr_temp_table">
+        <createProcedure replaceIfExists="true" procedureName="createEntityAttributeTempTable">
+            CREATE PROCEDURE createEntityAttributeTempTable ()
+            BEGIN
+            create temporary table ENTITY_ATTRIBUTE_TEMP (
+                id bigint(20) unsigned NOT NULL AUTO_INCREMENT primary key,
+                namespace varchar(32) CHARACTER SET utf8mb3 COLLATE utf8mb3_bin NOT NULL,
+                name varchar(200) CHARACTER SET utf8mb3 COLLATE utf8mb3_bin NOT NULL,
+                value_string text,
+                value_json longtext,
+                value_number double DEFAULT NULL,
+                value_boolean bit(1) DEFAULT NULL,
+                value_entity_ref bigint(20) unsigned DEFAULT NULL,
+                list_index int(11) DEFAULT NULL,
+                list_length int(11) DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241106_streamline_entity_attr_temp_table_procedure.xml
```diff
@@ -0,0 +1,33 @@
...
+                value_entity_ref bigint(20) unsigned DEFAULT NULL,
+                list_index int(11) DEFAULT NULL,
+                list_length int(11) DEFAULT NULL,
+                owner_id bigint(20) unsigned NOT NULL,
+                deleted bit(1) DEFAULT false,
+                deleted_date timestamp NULL DEFAULT NULL,
+                INDEX entity_tmp_owner_id_idx (owner_id),
+                INDEX UNQ_ENTITY_ATTRIBUTE(owner_id, namespace, name, list_index)
+            );
+            END
```
davidangb, line 27: compare to https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20230504_update_entity_attr_temp_table_procedure.xml; this:
* removes transaction_id
* changes the namespace and name columns from `text` to `varchar`
* removes the `entity_tmp_owner_id_idx` index

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/AttributeComponent.scala
```diff
@@ -33,9 +33,7 @@ trait AttributeRecord[OWNER_ID] {
   val deletedDate: Option[Timestamp]
 }
 
-trait AttributeScratchRecord[OWNER_ID] extends AttributeRecord[OWNER_ID] {
-  val transactionId: String
-}
+trait AttributeScratchRecord[OWNER_ID] extends AttributeRecord[OWNER_ID]
```
davidangb, line 36: the transactionId is a leftover from when we had a persistent `ENTITY_ATTRIBUTE_SCRATCH` table which was used for multiple transactions. In that setup, different rows had different transactionIds.

Back in 2021 we switched from a persistent table to temp tables, but we retained the transactionId column. However, the transactionId value is the same for every row in the temp table, so it is completely extraneous.

marctalbott (approved):
I have no experience with the temporary attribute table, but these changes seem reasonable based on the PR description

samanehsan (approved):

PR #3139, opened 2024-11-21
CORE-133: add status column for billing account change synchronizer

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-133

This is setup/pre-work for improving the SQL query used by the billing account change synchronizer.

In this PR:
* add a new STATUS column to the BILLING_ACCOUNT_CHANGES table. The column is an enum of `Ignored`, `Outstanding`, `Synchronized`. It reflects whether this change has been synchronized to Google.
* add a db index for the STATUS column
* as new rows are added to the table, they will have status=`Outstanding`
* as these rows are processed by the synchronizer, the synchronizer will update their status to `Synchronized`.
* any time a new `Outstanding` row is added, any pre-existing `Outstanding` rows _for the same billing project_ will be changed to `Ignored`.
* when the current implementation of the `BillingAccountChangeSynchronizer` finds a change it wants to process, also perform a "preview" lookup using the new STATUS column, to validate that the STATUS implementation is correct. This "preview" query is unused for anything other than validation/logging.

This status column works in conjunction with the pre-existing "Outcome" column. When STATUS='Synchronized`, OUTCOME will be 'Success' or 'Failure'; when STATUS is anything else, OUTCOME will be null. I considered changing the OUTCOME column for this PR to be non-null and include "Ignored" and "Outstanding" values, but I went with a separate column instead.

As of this PR, we are only writing the statuses. We never read them or rely on them. My intention is to let this run in production for a bit and manually validate that the status-tracking is correct before switching over the logic in the synchronizer to use this column. When that time comes, the synchronizer can query for the "Outstanding" status instead of its current logic.

---

**PR checklist**

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241121_billing_account_changes_status.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241121_billing_account_changes_status.xml
@@ -7,7 +7,7 @@
         <!-- add BILLING_ACCOUNT_CHANGES.STATUS column, defaulting to 'outstanding' -->
         <addColumn tableName="BILLING_ACCOUNT_CHANGES">
             <column name="STATUS"
-                    type="ENUM('ignored','outstanding','synchronized')"
+                    type="ENUM('failed', 'ignored','outstanding','synchronized')"
                     defaultValue="outstanding"
                     afterColumn="GOOGLE_SYNC_TIME">
                 <constraints nullable="false"/>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241121_billing_account_changes_status.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241121_billing_account_changes_status.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
+    <changeSet id="add_billing_account_changes_status" author="davidan" logicalFilePath="dummy">
+        <!-- add BILLING_ACCOUNT_CHANGES.STATUS column, defaulting to 'outstanding' -->
+        <addColumn tableName="BILLING_ACCOUNT_CHANGES">
+            <column name="STATUS"
+                    type="ENUM('ignored','outstanding','synchronized')"
+                    defaultValue="outstanding"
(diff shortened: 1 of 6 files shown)


calypsomatic (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -577,5 +604,20 @@ trait RawlsBillingProjectComponent {
...
+    def ignoreAllOutstanding(billingProjectName: RawlsBillingProjectName): WriteAction[Int] =
+      query
+        .filter(c =>
+          c.billingProjectName === billingProjectName.value && c.status === BillingAccountChangeStatus.Outstanding.toString
+        )
+        .map(_.status)
+        .update(BillingAccountChangeStatus.Ignored.toString)
+
+    def nextOutstanding(): BillingAccountChangeQuery =
+      query.filter(_.status === BillingAccountChangeStatus.Outstanding.toString).sortBy(_.id).take(1)
```
davidangb, line 620: `nextOutstanding()` is the query we'll eventually switch over to for polling, instead of the expensive query we have now. This results in a very light SQL statement:
```sql
select
  `ID`, `BILLING_PROJECT_NAME`, `USER_ID`, `PREVIOUS_BILLING_ACCOUNT`, `NEW_BILLING_ACCOUNT`, `CREATED`, `GOOGLE_SYNC_TIME`, `STATUS`, `OUTCOME`, `MESSAGE`
from `BILLING_ACCOUNT_CHANGES`
where `STATUS` = 'Outstanding'
order by `ID`
limit 1
```

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241121_billing_account_changes_status.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
```
davidangb, line 5: of note: I'm using the latest xsd here - `dbchangelog-4.30.xsd` - instead of the ancient `dbchangelog-3.4.xsd` in other changesets. No particular reason other than we should probably stay current.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -106,6 +117,44 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
...
+      }
+      if (changeStatus != BillingAccountChangeStatus.Outstanding) {
+        sb.append(s" Legacy lookup had status $changeStatus.")
+      }
+      if (changeId != byStatus.head.id) {
+        sb.append(s" Legacy found id $changeId, but by-status head had id ${byStatus.head.id}")
+      }
+      logger.warn(sb.toString())
+    }
+
```
davidangb, line 157: this will all get removed when we switch over

dvoet (approved):

calypsomatic (approved):

PR #3140, opened 2024-11-22
[CORE-172] Change submission cost cap to per-workflow instead of aggregate submission cost

marctalbott:
Ticket: [CORE-172](https://broadworkbench.atlassian.net/browse/CORE-172)
* It is more useful for our users to cap each workflow rather than cap the submission in its entirety. This PR repurposes the existing column on the submission table that was added in our initial implementation of submission cost capping to use it as a per-workflow cost cap. Rawls will now attempt to abort any workflow that exceeds the cost cap if there is one set on the submission.

---

**PR checklist**

[CORE-172]: https://broadworkbench.atlassian.net/browse/CORE-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -582,24 +599,12 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
-
-    workflowRecsAction.flatMap { workflowRecs =>
-      val nonTerminalWorkflows =
-        if (costCapThreshold.isDefined)
-          workflowRecs
-            .filterNot(wf => WorkflowStatuses.terminalStatuses.contains(WorkflowStatuses.withName(wf.status)))
-        else workflowRecs
-
-      if (nonTerminalWorkflows.isEmpty) {
+  )(implicit executionContext: ExecutionContext): ReadWriteAction[Boolean] =
```
marctalbott, line 602: I've reverted `updateSubmissionStatus` to its implementation prior to the submission cost cap. This method checks on the workflows in a submission and updates the submission status to a terminal status if appropriate. The cost capped implementation would also update the submission status to `Aborting` if the submission's workflows had exceeded the cost cap, but that is no longer necessary as each workflow will be aborted independently.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -312,10 +312,27 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
   ): Future[Option[WorkflowRecord]] =
     workflowRec.externalId match {
       // fetch cost information for the workflow if submission has a cost cap threshold defined
-      case Some(externalId) if costCapThreshold.isDefined =>
-        executionServiceCluster.getCost(workflowRec, petUser).map { costBreakdown =>
-          Option(workflowRec.copy(status = costBreakdown.status, cost = costBreakdown.cost.some))
-        }
+      case Some(externalId) if perWorkflowCostCap.isDefined =>
```
marctalbott: This is where we do the per-workflow cost cap enforcement. If there is no per-workflow cost cap set, then it only queries Cromwell for the workflow status.

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -312,10 +312,27 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
...
+        for {
+          costBreakdown <- executionServiceCluster.getCost(workflowRec, petUser)
+          updatedWorkflowRec <-
+            if (costBreakdown.cost > perWorkflowCostCap.get) {
```
dvoet: nit: you can include `perWorkflowCostCap` in the match then get rid of `if perWorkflowCostCap.isDefined` and this `.get` call

dvoet (approved):

trholdridge (approved):
looks good!

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
jgainerdewar: Nice that we won't have to continue asking Cromwell about the costs of finished workflows while the submission is running!

jgainerdewar (approved):

PR #3145, opened 2024-12-03
CORE-133: Optimize SQL in billing account change synchronizer

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-133

Followup to #3139. In that other PR, we added a new `STATUS` column to the `BILLING_ACCOUNT_CHANGES` table, with logic to populate that column's values. We also added validation into the `BillingAccountChangeSynchronizer` which compares the polling query it currently uses to a prospective new query using the `STATUS` column. After running for a week or so, we verified that the business logic around `STATUS` is correct.

In this PR, we switch the `BillingAccountChangeSynchronizer` away from the current expensive polling query to the new optimized query using the `STATUS` column.

As a bonus, this PR also changes the `BillingAccountChangeSynchronizer` from running at a fixed rate via [startTimerAtFixedRate](https://doc.akka.io/japi/akka/snapshot/akka/actor/TimerScheduler.html#startTimerAtFixedRate(java.lang.Object,java.lang.Object,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration)) to running at a fixed delay via [startTimerWithFixedDelay](https://doc.akka.io/japi/akka/snapshot/akka/actor/TimerScheduler.html#startTimerWithFixedDelay(java.lang.Object,java.lang.Object,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration)). Instead of running every N seconds, we'll now wait N seconds between runs. The difference is that the former does not take into account how long the task runs, while the latter does.

---

**PR checklist**

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import cats.implicits.catsSyntaxOptionId
+import com.google.common.annotations.VisibleForTesting
 import org.broadinstitute.dsde.rawls.RawlsException
 import org.broadinstitute.dsde.rawls.dataaccess.GoogleApiTypes.GoogleApiType
 import org.broadinstitute.dsde.rawls.dataaccess.GoogleOperationNames.GoogleOperationName
@@ -540,6 +541,7 @@ trait RawlsBillingProjectComponent {
       )
     }
 
+    @VisibleForTesting
     def getLastChange(billingProject: RawlsBillingProjectName): ReadAction[Option[BillingAccountChange]] =
       BillingAccountChanges
         .withProjectName(billingProject)
@@ -573,27 +575,6 @@ trait RawlsBillingProjectComponent {
 
     def withProjectName(billingProjectName: RawlsBillingProjectName): BillingAccountChangeQuery =
       query.filter(_.billingProjectName === billingProjectName.value)
-
-    /* SELECT *
-     * FROM BILLING_ACCOUNT_CHANGES BAC,
-     * (  SELECT BILLING_PROJECT_NAME, MAX(ID) AS MAXID
-     *    FROM BILLING_ACCOUNT_CHANGES
-     *    GROUP BY BILLING_PROJECT_NAME
-     * ) AS SUBTABLE
-     * WHERE SUBTABLE.MAXID = BAC.ID
-     */
-    /**
-      * Selects the latest changes for all billing projects in query.
-      */
-    def latestChanges: BillingAccountChangeQuery = {
-      val latestChangeIds = query
-        .groupBy(_.billingProjectName)
-        .map { case (_, group) => group.map(_.id).max }
-
-      query
-        .filter(_.id.in(latestChangeIds))
-        .sortBy(_.id.asc)
-    }
 
     def unsynced: BillingAccountChangeQuery =
       query.filter(_.googleSyncTime.isEmpty)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
@@ -88,9 +88,7 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
 
   def readABillingProjectChange: IO[Option[BillingAccountChange]] =
     inTransaction {
-      BillingAccountChanges.latestChanges.unsynced
-        .take(1)
-        .result
+      BillingAccountChanges.nextOutstanding().result
     }.map(_.headOption)
 
   private def syncBillingAccountChange[F[_]](tracingContext: RawlsTracingContext)(implicit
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -540,6 +541,7 @@ trait RawlsBillingProjectComponent {
       )
     }
 
+    @VisibleForTesting
     def getLastChange(billingProject: RawlsBillingProjectName): ReadAction[Option[BillingAccountChange]] =
       BillingAccountChanges
         .withProjectName(billingProject)
```
davidangb, line 547: this is now only used in tests, so I annotated it

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponent.scala
```diff
@@ -574,27 +576,6 @@ trait RawlsBillingProjectComponent {
...
-    def latestChanges: BillingAccountChangeQuery = {
-      val latestChangeIds = query
-        .groupBy(_.billingProjectName)
-        .map { case (_, group) => group.map(_.id).max }
-
-      query
-        .filter(_.id.in(latestChangeIds))
-        .sortBy(_.id.asc)
-    }
-
```
davidangb, line 597: this is the old unoptimized query, now unused.

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -47,7 +46,7 @@ object BillingAccountChangeSynchronizer {
     Behaviors.setup { context =>
       val actor = BillingAccountChangeSynchronizer(dataSource, gcsDAO, samDAO)
       Behaviors.withTimers { scheduler =>
-        scheduler.startTimerAtFixedRate(UpdateBillingAccounts, initialDelay, pollInterval)
+        scheduler.startTimerWithFixedDelay(UpdateBillingAccounts, initialDelay, pollInterval)
```
davidangb, line 49: see https://doc.akka.io/japi/akka/snapshot/akka/actor/TimerScheduler.html for detailed explanation of the difference

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -88,9 +87,7 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
 
   def readABillingProjectChange: IO[Option[BillingAccountChange]] =
     inTransaction {
-      BillingAccountChanges.latestChanges.unsynced
-        .take(1)
-        .result
+      BillingAccountChanges.nextOutstanding().result
```
davidangb, line 90: this is the core of the PR - changing from the old query to the new one

core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/BillingAccountChangeSynchronizer.scala
```diff
@@ -101,12 +98,6 @@ final case class BillingAccountChangeSynchronizer(dataSource: SlickDataSource,
     for {
       billingProject <- loadBillingProject
 
-      // Try out the new query using the STATUS column and compare results. If validation fails, don't
-      // fail the entire process.
-      _ <- validateStatusColumn.recover { case e: Exception =>
-        logger.warn(s"BillingAccountChangeStatus logic failed with ${e.getMessage}", e)
-      }
-
```
davidangb, line 109: we no longer need to validate the old query and new query return the same results, since we're getting rid of the old query

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponentSpec.scala
```diff
@@ -361,35 +361,6 @@ class RawlsBillingProjectComponentSpec
...
-        // billing project billing account. Right now, we're getting the latest changes
-        // in order of ID. We *COULD* get the changes in order of when the first skipped
-        // change was made. That's more complicated, so we'll do this for now to keep
-        // things simple.
-        change1 <- BillingAccountChanges.getLastChange(testData.testProject2Name)
-        change2 <- BillingAccountChanges.getLastChange(testData.testProject1Name)
-      } yield changes shouldBe List(change1, change2).map(_.value)
-    }
-  }
-
```
davidangb, line 392: the old query is gone, so this test can go too

samanehsan (approved):

trholdridge (commented):
Looks good! Why is it better to take the task running time into account vs. just having them be "evenly spaced"?

davidangb:
> Why is it better to take the task running time into account vs. just having them be "evenly spaced"?

@trholdridge imagine a task that runs for 7 seconds. If that task is scheduled via `startTimerAtFixedRate` at a fixed rate of 5 seconds, the system will try to _start_ that task every 5 seconds. Since the first run of the task takes 7 seconds, the next run will start immediately, since it is already past the second run's start time. The third run will also start immediately, and so forth … resulting in a lot of load on the system because the task runs continuously.

However, if that task is scheduled via `startTimerWithFixedDelay`, it will run for 7 seconds, wait for 5 seconds, run for 7 seconds, run for 5, and so on. `startTimerWithFixedDelay` guarantees some downtime between tasks.

In this PR's use case of polling for billing account changes, we aren't worried about delays of a few seconds; we'll get to the change when we get to it. So, let's choose the scheduling approach that is lighter on the system. If we had strict latency requirements we might choose otherwise.

trholdridge (approved):

PR #3146, opened 2024-12-05
AN-297 Fix local back Rawls, update README

aednichols:
- No changes to production code
- For local dev only, flipped values for `useWorkflowCollectionLabel` and `useWorkflowCollectionField`
  - Local is still different than Prod but at least it actually works now
  - Document that Prod should become more like local by adopting CromIAM
- Update README with clear front vs. back Rawls requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ And when you're done, spin down mysql (it is also fine to leave it running for y
 127.0.0.1	local.dsde-dev.broadinstitute.org
 ```
 
-### Running:
+### Running Front Rawls (default)
 
 After satisfying the above requirements, execute the following command from the root of the Rawls repo:
 
@@ -81,13 +81,18 @@ After satisfying the above requirements, execute the following command from the 
 By default, this will set up an instance of rawls pointing to the database and Sam in dev. 
 It will also set up a process that will watch the local files for changes, and restart the service when source files change.
 
-See docker-rsync-local-rawls.sh for more configuration options.
+See `docker-rsync-local-rawls.sh` for more configuration options.
 
-If Rawls successfully starts up, you can now access the Rawls Swagger page: https://local.dsde-dev.broadinstitute.org:20443/
+When Rawls starts up, access the Rawls Swagger page: https://local.dsde-dev.broadinstitute.org:20443/
 
-### Useful Tricks:
+### Running Back Rawls
 
-#### Front & Back Rawls
+#### Additional requirements for Back Rawls
+
+1. Broad campus network or NonSplit VPN
+2. Personal clone of the [Rawls Dev database](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/overview?project=broad-dsde-dev)
+3. Edit `local-dev/templates/sqlproxy.env` to set your clone instance name 
+4. Re-render local configuration after editing the template: `./local-dev/bin/render`
 
 By default, a locally run Rawls will boot as a "front" instance of Rawls. A front Rawls will serve all HTTP requests and can modify the database, but it will not do monitoring tasks such as submission monitoring, PFB imports, or Google billing project creation.
 
--- a/README.md
+++ b/README.md
@@ -58,21 +58,25 @@ And when you're done, spin down mysql (it is also fine to leave it running for y
 
 ### Requirements
 
+* Broad campus network or NonSplit VPN
 * [Docker Desktop](https://www.docker.com/products/docker-desktop) (8GB+ recommended)
-* Make sure you have `kubectl` and `gcloud` installed.
-* You will then need to authenticate in gcloud; if you are not already then running the script will ask you to.
-* Render the local configuration files. From the root of the Rawls repo, run:
+* `kubectl` and `gcloud` installed.
+
+### Initial setup
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -82,13 +85,17 @@ After satisfying the above requirements, execute the following command from the
...
+When Rawls starts up, access the Rawls Swagger page: https://local.dsde-dev.broadinstitute.org:20443/
 
-If Rawls successfully starts up, you can now access the Rawls Swagger page: https://local.dsde-dev.broadinstitute.org:20443/
+### Back Rawls
 
-### Useful Tricks:
+#### Additional setup for Back Rawls
 
-#### Front & Back Rawls
+1. Personal clone of the [Rawls Dev database](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/overview?project=broad-dsde-dev)
```
dvoet, line 97: why are we recommending a personal clone?

LizBaldo: You can interfere with Dev when testing if not using a clone. We had this issue with Leo and more recently with Rawls, it is safer to create a clone and then stop it after testing to avoid cost

README.md
```diff
@@ -98,21 +105,15 @@ If you are developing a ticket that deals with any sort of monitoring or asynchr
...
 
-**Important**: It is highly recommended that use your own Cloud SQL instance when running an instance of back Rawls by cloning the [dev Rawls DB](https://console.cloud.google.com/sql/instances/terraform-qfarbdq3lrexxck5htofjs5z6m/overview?project=broad-dsde-dev). See note below on database work.
-
-#### Developing Database Schema Changes
-
-If you are writing Liquibase migrations or doing database work, it is mandatory that you create your own Cloud SQL instance and develop against that. To point to your own CloudSQL instance, modify `/config/sqlproxy.env` accordingly.
-
+### Developing Database Schema Changes
 
+If you are writing Liquibase migrations or doing database work, set up a database clone by following the steps for [running Back Rawls](#additional-requirements-for-back-rawls).
```
LizBaldo: This might be a bit confusing if we already tell the dev to use a clone. It might make sense to move the justification of using a clone up a bit

README.md
aednichols: More than 1 back Rawls sharing a database act as consumers that compete over grabbing & submitting workflows.

This caused a ton of confusion where we believed local back Rawls was working; but it was actually Dev back Rawls stealing its workflows.

README.md
```diff
@@ -98,21 +105,15 @@ If you are developing a ticket that deals with any sort of monitoring or asynchr
...
 
 ## Developer quick links:
 * Swagger UI: https://rawls.dsde-dev.broadinstitute.org
 
 ## Build Rawls docker image
 
-Note: this may use more than 4.5 GB of Docker memory. If your Docker is not configured with enough memory (Docker for Mac defaults to 2GB), you may see a cryptic error messages saying `Killed`.
```
lucymcnatt, line 114: Is this just never an issue with docker set to 8+ gb?

README.md
```diff
@@ -164,16 +165,8 @@ After publishing:
 If you get the error message `release version 17 not supported`:
 * Run `java -version` and verify that you're running 17. If not, you will need to install / update your PATH.
 
-If you have trouble submitting workflows and see errors like `HTTP error calling URI https://cromiam-priv.dsde-dev.broadinstitute.org`:
-* Connect to the NonSplit VPN and try again
-* CromIAM doesn't accept requests from outside the Broad trusted IP space 
-
-When running back Rawls with a DB clone, the app may crash on launch with an error related to `OpenTelemetry`.
-* Work around by setting `entityStatisticsCache.enabled = false`.
```
lucymcnatt, line 172: is this not a problem anymore?

aednichols: I originally added this comment last year and I didn't run into the issue re-testing the same scenario recently.

README.md
aednichols: I overhauled the Docker build process a couple months ago and ran dozens of builds that worked fine with 8 GB.

I also think that "you may see a cryptic error message" is not that helpful without specifics, there are *lots* of ways to get cryptic errors with Docker.

aednichols:
@jgainerdewar TIL Job Manager has a labels field, confirmed accurate for my test workspace `e0027d6e-0a77-4425-b90d-3f3cfdaa2884`.

Screenshot 2024-12-05 at 12 56 36

jgainerdewar (approved):

README.md
lucymcnatt: lol fair, there is more than one way to be cryptic

lucymcnatt (approved):

sam-schu:
GitHub won't let me comment on the line because you didn't modify it, but I just noticed the asterisk at the end of this line:

> If you are developing a ticket that deals with any sort of monitoring or asynchronous features, you will likely want to boot your Rawls as a "back" instance, which will run a fully-featured instance of Rawls with monitoring tasks enabled. To boot your local instance as a "back" instance, run*:

Did there used to be a footnote corresponding to this?

aednichols:
@sam-schu I do not know, it certainly doesn't have one now. Removed.

sam-schu (approved):
Thank you for doing this! May the future of Rawls development be better for all!

LizBaldo (approved):

PR #3166, opened 2025-01-17
[CORE-267] Query BQ once per BP for spend report, handle errors

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-267
For the consolidated spend report, instead of unioning all billing projects together into one BQ query, this PR does a separate query for each billing project.  If BQ returns an error for any one of them, it is left out of the results.

As currently implemented, BPs that generate errors are essentially ignored, resulting in all these workspaces simply not showing up in the UI.  Since the form of the result returned from the Rawls API has a list of [SpendReportingForDateRange](https://github.com/broadinstitute/rawls/blob/9c4e0f908012989fafa7dc670dd9830198d48940/core/src/main/scala/org/broadinstitute/dsde/rawls/model/SpendReportingModel.scala#L88), it might not be very straightforward to return a result for all the workspaces in a problematic BP that indicates that there was a problem.  So that would suggest that the UI would have to figure out which, if any, workspaces are missing and display an 'N/A' for them, but that leaves out some information that we do have.  Reviewers, is it worth it to try to shove some indication of the problem into a `SpendReportingResult` or is it ok just to leave it to the UI?

Testing:
I compared my results from the consolidated spend report API both before and after I split the query up and I believe the substance did not change.  I then created a BP and gave it a false BQ dataset in its spend report configuration.  After checking that this resulted in errors on the main branch, I verified that I was able to successfully get the report on this branch.

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -359,7 +359,7 @@ class SpendReportingService(
 
   def getAllUserWorkspaceQuery(
     spendExportTable: String,
-    workspaces: Seq[(GoogleProjectId, WorkspaceName)],
+    workspaces: Seq[GoogleProjectId],
     pageSize: Int,
     offset: Int
   ): String = {
@@ -389,7 +389,11 @@ class SpendReportingService(
       baseQuery
         .replace("_PARTITIONTIME", timePartitionColumn)
         .replace("_BILLING_ACCOUNT_TABLE", spendExportTable)
-        .replace("_PROJECT_ID_LIST", "(" + workspaces.map(tuple => s""""${tuple._1.value}"""").mkString(", ") + ")")
+        .replace("_PROJECT_ID_LIST",
+                 workspaces
+                   .map(projectId => s""""${projectId}"""")
+                   .mkString("(", ", ", ")")
+        )
 
     s"""WITH spend_categories AS (
        |$bpSubQuery
@@ -565,8 +569,11 @@ class SpendReportingService(
           return Future.successful(None)
         }
         projectNames: Map[GoogleProjectId, WorkspaceName] = billingMap.values.flatten.toMap
-        results <- Future.sequence(billingMap.map { case (spendExportTable, workspaces) =>
-          val query = getAllUserWorkspaceQuery(spendExportTable, workspaces, pageSize, offset)
+        tableToProjectIdsMap: Map[String, Seq[GoogleProjectId]] = billingMap.map { case (table, workspaces) =>
+          table -> workspaces.map(_._1)
+        }
+        results <- Future.sequence(tableToProjectIdsMap.map { case (spendExportTable, projects) =>
+          val query = getAllUserWorkspaceQuery(spendExportTable, projects, pageSize, offset)
           val queryJob = setUpAllUserWorkspaceQuery(query, start, end)
           runBigQueryJob(queryJob, childContext)
             .map { result =>
@@ -575,7 +582,8 @@ class SpendReportingService(
                 case rows => Some(extractCrossBillingProjectSpendReportingResults(rows, start, end, projectNames))
               }
             }
-            .recoverWith { case ex =>
+            .recoverWith { case ex: Throwable =>
+              logger.warn(s"Error fetching results from BigQuery: ${ex.getMessage}")
               Future.successful(None)
             }
         })
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -358,7 +358,8 @@ class SpendReportingService(
   }
 
   def getAllUserWorkspaceQuery(
-    billingProjects: Map[BillingProjectSpendExport, Seq[(GoogleProjectId, WorkspaceName)]],
+    billingProject: BillingProjectSpendExport,
+    workspaces: Seq[(GoogleProjectId, WorkspaceName)],
     pageSize: Int,
     offset: Int
   ): String = {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -418,6 +420,67 @@ class SpendReportingService(
        |""".stripMargin.trim
   }
 
+//  def getAllUserWorkspaceQuery(
+//                                billingProjects: Map[BillingProjectSpendExport, Seq[(GoogleProjectId, WorkspaceName)]],
```
calypsomatic: Yes I know!  I'll delete it!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -358,7 +358,8 @@ class SpendReportingService(
   }
 
   def getAllUserWorkspaceQuery(
-    billingProjects: Map[BillingProjectSpendExport, Seq[(GoogleProjectId, WorkspaceName)]],
+    billingProject: BillingProjectSpendExport,
+    workspaces: Seq[(GoogleProjectId, WorkspaceName)],
```
davidangb: the `WorkspaceName`s aren't used anywhere inside this method. If you changed this method to accept a `Seq[GoogleProjectId]` instead then some interior code gets a little simpler

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -383,16 +384,17 @@ class SpendReportingService(
...
-      .mkString("\nUNION ALL\n")
+//    val bpSubQuery = billingProjects
+//      .map { bp =>
+    val tableName = billingProject.spendExportTable.getOrElse(spendReportingServiceConfig.defaultTableName)
+    val timePartitionColumn: String = getTimePartitionColumn(tableName)
+    val bpSubQuery =
+      baseQuery
+        .replace("_PARTITIONTIME", timePartitionColumn)
+        .replace("_BILLING_ACCOUNT_TABLE", tableName)
+        .replace("_PROJECT_ID_LIST", "(" + workspaces.map(tuple => s""""${tuple._1.value}"""").mkString(", ") + ")")
```
dvoet:
```suggestion
        .replace("_PROJECT_ID_LIST", "(" + workspaces.map { case (project, _) => s""""${project.value}"""" }.mkString(", ") + ")")
```
the `_1` syntax is tough, using the curlies and case make it look nicer

davidangb: nit/optional: you can use the 3-argument form of `mkString` here instead; your call on which is more readable
```suggestion
        .replace("_PROJECT_ID_LIST", workspaces.map(tuple => s""""${tuple._1.value}"""").mkString("(", ", ", ")"))
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -568,15 +631,34 @@ class SpendReportingService(
...
+          val queryJob = setUpAllUserWorkspaceQuery(query, start, end)
+          runBigQueryJob(queryJob, childContext)
+            .map { result =>
+              result.getValues.asScala.toList match {
+                case Nil  => None
+                case rows => Some(extractCrossBillingProjectSpendReportingResults(rows, start, end, projectNames))
+              }
+            }
+            .recoverWith { case ex =>
+              Future.successful(None)
```
davidangb, line 587: log something here to show that we ran into a problem?

davidangb (commented):
some syntax suggestions inline, not a full review:

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -568,15 +631,34 @@ class SpendReportingService(
...
+          SpendReportingResults(
+            acc.spendDetails ++ res.spendDetails,
+            SpendReportingForDateRange(
+              (BigDecimal(acc.spendSummary.cost) + BigDecimal(res.spendSummary.cost)).toString,
+              (BigDecimal(acc.spendSummary.credits) + BigDecimal(res.spendSummary.credits)).toString,
+              acc.spendSummary.currency,
+              acc.spendSummary.startTime,
+              acc.spendSummary.endTime
+            )
+          )
```
davidangb: Consider writing a `+` method in the `SpendReportingResults` case class, then calling it here.

if you had:
```scala
case class SpendReportingResults(spendDetails: Seq[SpendReportingAggregation],
                                 spendSummary: SpendReportingForDateRange
) {
  def +(other: SpendReportingResults): SpendReportingResults =
    new SpendReportingResults(
      this.spendDetails ++ other.spendDetails,
      SpendReportingForDateRange(
        (BigDecimal(this.spendSummary.cost) + BigDecimal(other.spendSummary.cost)).toString,
        (BigDecimal(this.spendSummary.credits) + BigDecimal(other.spendSummary.credits)).toString,
        this.spendSummary.currency,
        this.spendSummary.startTime,
        this.spendSummary.endTime
      )
    )
}
```

then this code could simply be:
```scala
combinedResults = results.flatten.reduceOption((acc, res) => acc + res)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
dvoet, line 587: should at least log this exception

dvoet:
@davidangb jinx

marctalbott (approved):

davidangb (approved):

PR #3172, opened 2025-01-24
CORE-310: workflows include Cromwell-estimated costs when actual cost is not available

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-310

The `getSubmissionStatus` API did not return workflow costs unless/until actual costs were populated into BigQuery.

This API is called from the submission-detail page of Terra UI while a submission is running. Because there can be a significant delay in populating actual costs, it meant that in many cases costs were not available for currently-running workflows.

This PR exposes the estimated cost from Cromwell until the actual cost is available. The estimated cost is available much earlier.

This also adds a new `costType: Actual | Estimated` property to the `Workflow` response object. Callers such as Terra UI can inspect this to know whether the reported cost is estimated by Cromwell or the actual cost as calculated from billing export.

See also https://github.com/broadinstitute/firecloud-orchestration/pull/1560

---

**PR checklist**

  - [ ] in workbench-libs


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/SubmissionComponent.scala
```diff
@@ -433,7 +433,9 @@ trait SubmissionComponent {
                new DateTime(wr.statusLastChangedDate.getTime),
                entityRef,
                workflowResolutions.sortBy(_.inputName), // enforce consistent sorting
-               messages
+               messages,
+               wr.cost.map(_.floatValue),
+               wr.cost.map(_ => WorkflowCostTypes.Estimated)
```
davidangb, line 438: when retrieving a workflow from the db, if that workflow has a value for `cost`, the value is an estimated value.

core/src/main/scala/org/broadinstitute/dsde/rawls/submissions/SubmissionsService.scala
```diff
@@ -450,25 +451,34 @@ class SubmissionsService(
...
+              if (costMap.contains(wfId))
+                workflow.copy(cost = costMap.get(wfId), costType = Option(WorkflowCostTypes.Actual))
+              else
+                workflow
+            case None => workflow
+          }
+        }
+        val costedSubmission = submission.copy(cost = Some(costMap.values.sum), workflows = costedWorkflows)
+        costedSubmission
+    }
```
davidangb, line 496: Extracted this into a standalone method to support unit testing.

This preserves the current business logic: every time the `getSubmissionStatus` API is called, execute a live request to BigQuery, then overlay the BQ results on top of the workflows.

I can picture an optimization here - out of scope for this PR - in which we persist the BQ results so we are not constantly making requests to it.

dvoet (approved):

kevinmarete (approved):

dvoet (approved):

kevinmarete (approved):

kevinmarete (approved):

dvoet (approved):

PR #3179, opened 2025-02-06
[CORE-175] Get bucket bytes usage from GCP monitoring

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-175
Adds a new `getBucketUsageV2` API that gathers more accurate data using the [monitoring API](https://cloud.google.com/monitoring/api/metrics_gcp#gcp-storage).  
This will return a list of bytes in the bucket separated by storage class and type; e.g. COLDLINE live-objects, COLDLINE soft-deleted-objects, and STANDARD live-objects will each have a separate byte count.  The object type is left out of this response, as all we care about for cost is the storage class.  

---

**PR checklist**

  - [ ] in workbench-libs


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1155,6 +1202,13 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
       .setServiceAccountPrivateKeyFromPemFile(new java.io.File(pemFile))
       .build()
 
+  def getBucketServiceAccountCredentials: GoogleCredentials =
```
calypsomatic, line 1205: The `Credential` that is used elsewhere is deprecated; in order to use the current metrics API I need to use `Credentials` instead, which needs the service account key in the form of a JSON.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1125,6 +1132,46 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
...
+  ): BucketMetricsResponse = {
+    val metricServiceClient = MetricServiceClient.create(
+      MetricServiceSettings
+        .newBuilder()
+        .setCredentialsProvider(FixedCredentialsProvider.create(getBucketServiceAccountCredentials))
+        .build()
+    )
+    val projectName = projectId.value
+    val now = System.currentTimeMillis
+    val fiveMinutesAgo = now - 5 * 60 * 1000
```
calypsomatic, line 1146: The metric is sampled every 300 seconds, so going back 5 minutes should give us the most recent value (although it's only measured once a day).

model/src/main/scala/org/broadinstitute/dsde/rawls/model/WorkspaceModel.scala
```diff
@@ -1086,6 +1086,16 @@ case class WorkspaceStatus(workspaceName: WorkspaceName, statuses: Map[String, S
...
 
+case class BucketMetricsResponse(
+  metrics: Seq[BucketMetric]
+)
+
+//See https://cloud.google.com/storage/docs/storage-classes
+//I believe the options for storageclass are STANDARD, NEARLINE, COLDLINE, ARCHIVE, REGIONAL, MULTI-REGIONAL and DRA.
+//I've seen REGIONAL and MULTI_REGIONAL in practice; I have not seen DRA so I don't know exactly how it would be coded
+//Hence I'm afraid to make an enumeration until I can verify what's possible to see.
+case class BucketMetric(storageClass: String, valueInBytes: Double)
```
calypsomatic, line 1097: I believe all we care about is `storageClass` and `valueInBytes`; orch can then calculate cost based on storage class.  There might be duplicates of storage class, as soft-deleted objects should be separated out, and REGIONAL and MULTI-REGIONAL both count as STANDARD, but I don't think we should see any actual double-counting as long as the interval we look over isn't more than 5 minutes.

core/src/main/resources/swagger/api-docs.yaml
```diff
@@ -6240,6 +6256,16 @@ components:
           format: timestamp
           description: timestamp (UTC) marking the date that the bucket usage was last updated (YYYY-MM-DDThh:mm:ss.fffZ)
       description: ""
+    BucketMetricsResponse:
+      type: object
```
dvoet: this should be an array

dvoet (approved):
swagger fix then 👍

kevinpalis (approved):

PR #3185, opened 2025-02-11
[CORE-175] Add aggregation to bucket usage metric query

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-175
While working on core-175, I occasionally noticed that a call to the new `getBucketUsageV2` API would return nothing.  At first I thought it was a fluke, then maybe an artifact of development, but as I saw it again and again, it became more alarming.   I eventually managed to reproduce it directly in the [google api explorer](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/list).  I think it might be that since the time interval of 5 minutes is exactly the sampling rate, there may be a certain period of time every N minutes where 5 minutes back doesn't catch anything; the [description](https://cloud.google.com/monitoring/api/metrics_gcp#gcp-storage) of the metric says "After sampling, data is not visible for up to 690 seconds."  So this probably means we need to go back 11.5 minutes, but then we might get duplicate data points, so instead I went with @davidangb 's suggestion from a slack [discussion](https://broadinstitute.slack.com/archives/C07QQ3L0AHX/p1739285251911529?thread_ts=1739213985.429469&cid=C07QQ3L0AHX) to use aggregation.  In practice, this never produced the empty metrics response while the [current implementation](https://rawls.dsde-dev.broadinstitute.org/#/workspaces_v2/getBucketUsageV2) did.
 
---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
@@ -39,7 +39,7 @@ import com.google.api.services.storage.model.Bucket.Lifecycle.Rule.{Action, Cond
 import com.google.api.services.storage.model._
 import com.google.api.services.storage.{Storage, StorageScopes}
 import com.google.cloud.monitoring.v3.{MetricServiceClient, MetricServiceSettings}
-import com.google.monitoring.v3.{ListTimeSeriesRequest, TimeInterval}
+import com.google.monitoring.v3.{Aggregation, ListTimeSeriesRequest, TimeInterval}
 import com.google.api.services.monitoring.v3.MonitoringScopes
 import com.google.protobuf.util.Timestamps
 import com.google.auth.oauth2.ServiceAccountCredentials
@@ -47,6 +47,8 @@ import com.google.auth.oauth2.GoogleCredentials
 import com.google.cloud.Identity
 import com.google.cloud.storage.Storage.{BucketSourceOption, BucketTargetOption}
 import com.google.cloud.storage.{BucketInfo, Cors, HttpMethod, StorageClass, StorageException}
+import com.google.monitoring.v3.Aggregation.Aligner
+import com.google.protobuf.Duration
 import io.opentelemetry.api.common.AttributeKey
 import org.apache.commons.lang3.StringUtils
 import org.broadinstitute.dsde.rawls.dataaccess.CloudResourceManagerV2Model.{Folder, FolderSearchResponse}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -1155,6 +1157,14 @@ class HttpGoogleServicesDAO(val clientSecrets: GoogleClientSecrets,
       .setName("projects/" + projectName)
       .setFilter("metric.type=\"storage.googleapis.com/storage/v2/total_bytes\"")
       .setInterval(interval)
```
davidangb, line 1159: if you're aggregating to a full 24-hour period, would it make sense to increase the interval above 5 minutes to ensure data? I'm not sure if it is necessary; your PR description sounds like it is not.

calypsomatic: Yeah I was a little confused about combining the interval with the aggregation period.  Does the aggregation just disregard the interval or what?

davidangb: My understanding - which could be wrong - is that interval is processed first, and Google finds all the time series within that interval, which is every 5 minutes. Then, it takes those every-5-minute data points and aggregates them into every-`aggregation.alignmentPeriod` data points.

davidangb: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.alertPolicies#Aggregation is what I was looking at

calypsomatic: That sounds reasonable, but not sure it explains why the aggregation would work when it doesn't without it?  Since there wouldn't be any data points to aggregate.  Overall I'm still very confused

calypsomatic: Ok, further experimentation suggests that it takes only the END time of the given interval, and then aggregates the previous alignmentPeriod up to that time.  So if you give it the interval 2:30-2:35, it'll take everything in the 24 hours from 2:30 previous day to 2:30 day of the interval.  If you give it an interval greater than the alignmentPeriod, say M2:00-T2:30, then you'll get one point for M2:30-T2:30 and one for M2:00-M2:30.  So I think the way I have it written now is FINE

marctalbott (approved):

davidangb (approved):

PR #3186, opened 2025-02-11
[CORE-289] Add workspace spend report table

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-289

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
@@ -1,31 +1,32 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
     <changeSet logicalFilePath="dummy" author="sehsan" id="create_WORKSPACE_SPEND_REPORT_TABLE">
         <createTable tableName="WORKSPACE_SPEND_REPORT">
+            <column name="ID" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
             <column name="GOOGLE_PROJECT_ID" type="VARCHAR(254)">
-                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+                <constraints nullable="false"/>
             </column>
             <column name="REPORT_START_DATE" type="DATETIME">
-                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+                <constraints nullable="false"/>
             </column>
             <column name="REPORT_END_DATE" type="DATETIME">
-                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
-            </column>
-            <column name="TOTAL_SPEND" type="NUMBER(10,2)">
                 <constraints nullable="false"/>
             </column>
             <column name="TOTAL_COMPUTE" type="NUMBER(10,2)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="TOTAL_STORAGE" type="NUMBER(10,2)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="OTHER_SPEND" type="NUMBER(10,2)">
-                <constraints nullable="false"/>
+                <constraints nullable="true"/>
             </column>
             <column name="IS_DATA_AVAILABLE" type="BOOLEAN">
                 <constraints nullable="false"/>
             </column>
         </createTable>
+        <addUniqueConstraint columnNames="GOOGLE_PROJECT_ID, REPORT_START_DATE, REPORT_END_DATE" constraintName="IDX_WORKSPACE_SPEND_REPORT" tableName="WORKSPACE_SPEND_REPORT"/>
     </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="GOOGLE_PROJECT_ID" type="VARCHAR(254)">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+            </column>
+            <column name="REPORT_START_DATE" type="DATETIME">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+            </column>
+            <column name="REPORT_END_DATE" type="DATETIME">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+            </column>
+            <column name="TOTAL_SPEND" type="NUMBER(10,2)">
```
calypsomatic: If the data is unavailable, then shouldn't the spend columns be null, so shouldn't they be nullable?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="sehsan" id="create_WORKSPACE_SPEND_REPORT_TABLE">
+        <createTable tableName="WORKSPACE_SPEND_REPORT">
+            <column name="GOOGLE_PROJECT_ID" type="VARCHAR(254)">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="PK_WORKSPACE_SPEND_REPORT"/>
+            </column>
```
davidangb, line 10: an alternative to the composite primary key is to add an autoincrement bigint `id` column as the primary key, then add a unique constraint/index for the GOOGLE_PROJECT_ID/REPORT_START_DATE/REPORT_END_DATE combination.

Will we ever need some other table to have a foreign key to this WORKSPACE_SPEND_REPORT table? If so, the single `id` column seems better. And of course we can always add the id column later.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="TOTAL_COMPUTE" type="NUMBER(10,2)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="TOTAL_STORAGE" type="NUMBER(10,2)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="OTHER_SPEND" type="NUMBER(10,2)">
+                <constraints nullable="false"/>
+            </column>
```
davidangb, line 25: Does `TOTAL_SPEND` need its own column? Is that always going to be the sum of TOTAL_COMPUTE+ TOTAL_STORAGE+ OTHER_SPEND ?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
```
davidangb: this is using `dbchangelog-3.4.xsd`, i.e. only supports Liquibase 3.4 syntax. We should be able to go all the way to [dbchangelog-4.31.xsd](https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.31.xsd); the [most recent changeset](https://github.com/broadinstitute/rawls/blob/f246f1f651231c873329b1b7d15c68192acd13e0/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20241218_alter_audit_workflow_status_trigger_frequency.xml#L5) uses 4.30.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
samanehsan: Ah I think I got confused with the inconsistency we found in the UI but I forgot that we just don't show the other spend category. What do you think about making it a computed column?

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
samanehsan: Thanks for catching that! I copied one of the older changesets so I'll update that

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
samanehsan: I thought we talked about making it zero during sprint planning? But maybe a `null` value would make the intention even clearer 🤔

calypsomatic: I think it's important to distinguish between actually $0.00 and "we just don't know".  maybe the data_available column would cover that, but having these null seems safer

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
davidangb: If the `TOTAL_SPEND` is really just the sum of the other three columns - @calypsomatic perhaps you can chime in? - I'd delete the db column altogether. Callers can do their own summing, or can write SQL to sum it; no need to persist anything.

davidangb: ah - I see you did fully delete the column!

davidangb (approved):
:+1:, unless Bria or others chimes in and says we actually do need to keep the `TOTAL_SPEND` column

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250211_add_workspace_spend_report_table.xml
calypsomatic: Yeah it's just the sum of the other categories, which the query [calculates and returns](https://github.com/broadinstitute/rawls/blob/88bcc22220beeb9aa328d09c46cc660c10ebdb9d/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala#L403) for you

samanehsan: Yeah, I just decided to delete it

calypsomatic (approved):

PR #3192, opened 2025-02-18
[CORE-319] Allow adding users upon workspace creation

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-319
This adds an `addUsers` parameter to a `WorkspaceRequest` that allows collaborators to be added to a workspace simultaneously with workspace creation.  I've used the same object, `WorkspaceACLUpdate` that is defined for adding users after workspace creation, as that seemed logical, but will listen to arguments that some new or different argument makes more sense.

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
@@ -1790,14 +1790,61 @@ class WorkspaceService(
     } else {
       Set(WorkbenchEmail(ctx.userInfo.userEmail.value))
     }
+    val groupedAclUpdates: Map[WorkspaceAccessLevel, List[WorkspaceACLUpdate]] = workspaceRequest.addUsers
+      .map { aclUpdates =>
+        aclUpdates.groupBy(_.accessLevel)
+      }
+      .getOrElse(Map.empty)
+
     val ownerPolicy =
-      SamWorkspacePolicyNames.owner -> SamPolicy(ownerPolicyMembership, Set.empty, Set(SamWorkspaceRoles.owner))
-    val writerPolicy = SamWorkspacePolicyNames.writer -> SamPolicy(Set.empty, Set.empty, Set(SamWorkspaceRoles.writer))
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1790,16 +1789,71 @@ class WorkspaceService(
...
+            aclUpdates
+              .filter(_.canCompute.get)
+              .map(update => WorkbenchEmail(update.email))
+              .toSet
+          }
+          .getOrElse(Set.empty),
+        Set.empty,
+        Set(SamWorkspaceRoles.canCompute)
+      )
     val canCatalogPolicy =
```
calypsomatic, line 1857: Looking at how to add users to a workspace now, there is no option for cataloging.  I'm guessing this isn't relevant for workspaces and should always be blank, or do I need to account for the possibility?

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
```diff
@@ -1790,16 +1789,71 @@ class WorkspaceService(
...
       Set(WorkbenchEmail(ctx.userInfo.userEmail.value))
     }
+
+    val groupedAclUpdates: Map[WorkspaceAccessLevel, List[WorkspaceACLUpdate]] = workspaceRequest.addUsers
+      .map { aclUpdates =>
+        aclUpdates.groupBy(_.accessLevel)
+      }
+      .getOrElse(Map.empty)
+
+    def getEmailsByAccessLevel(access: WorkspaceAccessLevel): Set[WorkbenchEmail] =
```
calypsomatic, line 1799: I tried to clean up the repetition a bit by adding these helper functions, but will welcome suggestions for making them even smarter.

kevinmarete (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala
dvoet: You are correct to leave it as is. Cataloging was a feature for the old firecloud data library. I don't think it is in use anymore.

dvoet (approved):

PR #3193, opened 2025-02-21
[CORE-307] Fix Spend Report Visibility for Billing Project Users (Workspace Owners)

kevinmarete:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-307

This PR ensures that Billing Project Users can view spend reports for workspaces they own on the billing project workspaces list page. Previously, these costs appeared as N/A, and users would receive a `403 Forbidden error` when attempting to access the billing project spend report. This was due to the requireProjectAction check for `readSpendReport`, which was restricted to project owners. I have updated this to `createWorkspace`, allowing workspace owners to access spend reports across all billing projects.

**Before**
Before

**After**
After

---

**PR checklist**

- [X] Include the JIRA issue number in the PR description and title
  - [ ] in workbench-libs
- [X] Verify all tests go green, including CI tests

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -467,41 +467,6 @@ class SpendReportingService(
         workspaceServiceConstructor(childContext).getGCPWorkspacesByBillingProjects(validWorkspaceIds)
       }
 
-  def getSpendReportableWorkspaceGoogleProjectsInBillingProject(
-    projectName: RawlsBillingProjectName,
-    childContext: RawlsRequestContext
-  ): Future[Map[GoogleProjectId, WorkspaceName]] = {
-
-    def filterProjects(projects: Map[RawlsBillingProjectName, Seq[Workspace]]): Map[GoogleProjectId, WorkspaceName] =
-      projects.flatMap { case (_, value) =>
-        value.collect {
-          case workspace if workspace.namespace == projectName.value =>
-            workspace.googleProjectId -> workspace.toWorkspaceName
-        }
-      }
-
-    for {
-      projects <- getSpendReportableWorkspaceGoogleProjects(childContext)
-      filteredProjects = filterProjects(projects)
-      hasAction <- samDAO.userHasAction(SamResourceTypeNames.billingProject,
-                                        projectName.value,
-                                        SamBillingProjectActions.readSpendReport,
-                                        childContext
-      )
-      result <-
-        if (hasAction || filteredProjects.nonEmpty) {
-          Future.successful(filteredProjects)
-        } else {
-          Future.failed(
-            RawlsExceptionWithErrorReport(
-              StatusCodes.Forbidden,
-              s"${childContext.userInfo.userEmail.value} cannot perform ${SamBillingProjectActions.readSpendReport.value} on project ${projectName.value}"
-            )
-          )
-        }
-    } yield result
-  }
-
   def getSpendForGCPBillingProject(
     project: RawlsBillingProjectName,
     start: DateTime,
@@ -511,7 +476,14 @@ class SpendReportingService(
     validateReportParameters(start, end)
     for {
       spendExportConf <- getSpendExportConfiguration(project)
-      projectNames <- getSpendReportableWorkspaceGoogleProjectsInBillingProject(project, childContext)
+      projectNames <- getSpendReportableWorkspaceGoogleProjects(ctx).map { workspacesByProject =>
+        workspacesByProject
+          .getOrElse(project, Seq.empty)
+          .map { workspace =>
+            workspace.googleProjectId -> workspace.toWorkspaceName
+          }
+          .toMap
+      }
 
       query = getQuery(aggregations, spendExportConf)
       queryJob = setUpQuery(query, spendExportConf, start, end, projectNames)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -311,13 +311,6 @@ class SpendReportingService(
 
       }
 
-  def getWorkspaceGoogleProjects(projectName: RawlsBillingProjectName): Future[Map[GoogleProjectId, WorkspaceName]] =
-    dataSource.inTransaction(_.workspaceQuery.listWithBillingProject(projectName)).map {
-      _.collect {
-        case w if w.workspaceVersion == WorkspaceVersions.V2 => w.googleProjectId -> w.toWorkspaceName
-      }.toMap
-    }
-
   def validateReportParameters(startDate: DateTime, endDate: DateTime): Unit = if (startDate.isAfter(endDate)) {
     throw RawlsExceptionWithErrorReport(
       StatusCodes.BadRequest,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -269,18 +269,6 @@ class SpendReportingService(
   def bytesProcessedCounter: Histogram =
     spendReportingMetrics.expand(BigQueryKey, BigQueryBytesProcessedMetric).asHistogram("bytes")
 
-  private def requireProjectAction[T](projectName: RawlsBillingProjectName, action: SamResourceAction)(
-    op: => Future[T]
-  ): Future[T] =
-    samDAO.userHasAction(SamResourceTypeNames.billingProject, projectName.value, action, ctx).flatMap {
-      case true => op
-      case false =>
-        throw RawlsExceptionWithErrorReport(
-          StatusCodes.Forbidden,
-          s"${ctx.userInfo.userEmail.value} cannot perform ${action.value} on project ${projectName.value}"
-        )
-    }
-
   private def toISODateString(dt: DateTime): String = dt.toString(ISODateTimeFormat.date())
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -477,7 +477,9 @@ class SpendReportingService(
     aggregations: Set[SpendReportingAggregationKeyWithSub]
   ): Future[SpendReportingResults] = {
     validateReportParameters(start, end)
-    requireProjectAction(project, SamBillingProjectActions.readSpendReport) {
+    requireProjectAction(project,
+                         SamBillingProjectActions.createWorkspace
```
dvoet: `readSpendReport` sounds like the right action to be checking. This probably should be a sam config change to grant `readSpendReport` to the right workspace role.

dvoet (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
dvoet: reading the jira more closely, it sounds like we have some inconsistent permission checking. Here, we are checking a project action but in the consolidated spend report it is perhaps checking a workspace action? We should make the permission checks consistent

kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/bd3cd9f1f92da9525e4e262a0778a4b893fb73a2

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -470,32 +469,68 @@ class SpendReportingService(
...
-  ): Future[SpendReportingResults] = {
+  ): Future[SpendReportingResults] = traceFutureWithParent("getSpendForGCPBillingProject", ctx) { childContext =>
     validateReportParameters(start, end)
-    requireProjectAction(project, SamBillingProjectActions.readSpendReport) {
-      for {
-        spendExportConf <- getSpendExportConfiguration(project)
-        projectNames <- getWorkspaceGoogleProjects(project)
+    for {
+      spendExportConf <- getSpendExportConfiguration(project)
+      projectNames <- getOwnedWorkspaceGoogleProjectsInProject(project, childContext)
```
davidangb: `getWorkspaceGoogleProjects` is now unused except in a test - clean it up?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -470,32 +469,68 @@ class SpendReportingService(
...
+            .filter(resourceId => Try(UUID.fromString(resourceId)).isSuccess) // filter out non-UUIDs
+            .toList
+          workspaceServiceConstructor(childContext).getGCPWorkspacesByBillingProjects(validWorkspaceIds)
+      }
+
+  def getOwnedWorkspaceGoogleProjectsInProject(
+    projectName: RawlsBillingProjectName,
+    childContext: RawlsRequestContext
+  ): Future[Map[GoogleProjectId, WorkspaceName]] =
+    requireProjectAction(projectName, SamBillingProjectActions.createWorkspace) {
```
davidangb: this is still checking `create_workspace` on the project instead of [read_spend_report](https://github.com/broadinstitute/sam/blob/d88b19fae8fdc160d1f42120bce61de85dcaef2d/src/main/resources/reference.conf#L756-L758); this mixes up two separate concepts (creating a workspace vs. reading a spend report)

1. Do we need a permission check on the billing project at all? The code is already checking permissions on each workspace in the project.
2. If we do need to check permissions on the project itself, could we modify Sam's config to give the `read_spend_report` action to the appropriate users, and check for that action?

davidangb (changes requested):

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
dvoet: I think this will make more sense if we change our thinking from "show spend report for all workspaces in a billing project" to "show spend report for all workspaces _filtered by_ billing project". If you do that, then you see that the `getOwnedWorkspaceGoogleProjects` function above does everything you need since it is already grouped by billing project.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -470,32 +469,68 @@ class SpendReportingService(
     bytesProcessedCounter += stats.getEstimatedBytesProcessed
   }
 
+  def getOwnedWorkspaceGoogleProjects(
```
dvoet: the naming is off here, these are not owned workspaces, they are workspaces where the caller can `readSpendReport`

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/802413ba4a91edce2ca73e664deee052fad7bd79

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/290b38d49a0b1c46cb4d53ee8660ad94487bf369

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/290b38d49a0b1c46cb4d53ee8660ad94487bf369

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -470,32 +450,85 @@ class SpendReportingService(
...
+        } else {
+          Future.failed(
+            RawlsExceptionWithErrorReport(
+              StatusCodes.Forbidden,
+              s"${childContext.userInfo.userEmail.value} cannot perform ${SamBillingProjectActions.readSpendReport.value} on project ${projectName.value}"
+            )
+          )
+        }
+    } yield result
+  }
```
dvoet: This is far too complicated. Checking the billing project action is not required, access at the workspace level has already been checked and that is sufficient. The filter function does not need to cycle through all the workspaces because the map is already keyed by billing project.

This function should be 
```
getSpendReportableWorkspaceGoogleProjects(childContext).map { workspacesByProject =>
  workspacesByProject.get(projectName).getOrElse(Map.empty).map { workspace =>
    workspace.googleProjectId -> workspace.toWorkspaceName
  }.toMap
}
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -470,32 +450,85 @@ class SpendReportingService(
...
+  ): Future[Map[RawlsBillingProjectName, Seq[Workspace]]] =
+    samDAO
+      .listResourcesWithActions(
+        SamResourceTypeNames.workspace,
+        SamWorkspaceActions.readSpendReport,
+        childContext
+      )
+      .flatMap {
+        case ownerWorkspaces if ownerWorkspaces.isEmpty =>
+          Future.successful(Map.empty[RawlsBillingProjectName, Seq[Workspace]])
```
dvoet: Why is this case necessary? Will `getGCPWorkspacesByBillingProjects` fail if `validWorkspaceIds` is empty?

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/6ca541dbe6e14fd161620ceb45f298cda6aabc57

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/6ca541dbe6e14fd161620ceb45f298cda6aabc57

dvoet (approved):

davidangb (approved):
Thanks for seeing this one through!

I like the unification of how the per-billing-project spend report and the consolidated spend report identify their workspaces; nice job on that.

Do all the deleted unit tests deserve deletion? Are any of those still valid?

kevinmarete:
> 98345d0



> Thanks for seeing this one through!
> 
> I like the unification of how the per-billing-project spend report and the consolidated spend report identify their workspaces; nice job on that.
> 
> Do all the deleted unit tests deserve deletion? Are any of those still valid?

Fixed: https://github.com/broadinstitute/rawls/pull/3193/commits/98345d04a44bee6da37f9cba9312abad8814968d I reviewed the five tests I had removed. I fixed and restored three of them, added one new test, and confirmed that the remaining two are no longer valid cases.

PR #3198, opened 2025-02-25
[CORE-290] Use/Store consolidated spend report query results in workspace spend report table

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-290

Changes:
- Query the workspace spend report table in Rawls for rows that match the given google project ids, start date and end date
- If there is a result for all project ids (a full cache hit), use those results and filter out any records where `isDataAvailable = false`
- Otherwise:
  - Run the bigQuery job and store those results in the workspace spend report table
  - For every google project id that does not have any query results, store a "dummy record" for that project id, start and end date in the database with `isDataAvailable = false` and no cost data. Otherwise subsequent calls will get a cache miss.

To-Do:
- [x] Handle the case where workspace spend is unavailable
- [x] Handle credits (need to add credits per category to the db table first) 
- [ ] Make matching by start/end more flexible (just check M/D/Y and not the exact timestamp)?
- [ ] Delete records older than X days (could make this a follow-on PR)
- [x] Tests!

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
@@ -1,8 +1,5 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
-import akka.http.scaladsl.model.StatusCodes
-import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
-
 import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
@@ -109,33 +106,26 @@ object WorkspaceSpendReportRecord {
       val currencyCode = Currency.getInstance(currencyString)
       val projectId = record.googleProjectId
 
-      def formatCategoryCost(cost: Option[Float]): String = {
-        cost match {
-          case Some(cost) => toBigDecimal(cost).toString()
-          case None => "N/A"
-        }
-      }
-
-      def toBigDecimal(cost: Float): BigDecimal = {
-        BigDecimal(cost).setScale(currencyCode.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
+      def toBigDecimal(cost: Option[Float]): BigDecimal = {
+        BigDecimal(cost.getOrElse(0.0f)).setScale(currencyCode.getDefaultFractionDigits, RoundingMode.HALF_EVEN)
       }
 
       val credits = BigDecimal(0.0).toString() // TODO: Store credits per category
       val subAggregation = List(
         SpendReportingForDateRange(
-          formatCategoryCost(record.otherSpend),
+          toBigDecimal(record.otherSpend).toString(),
           credits,
           currencyCode.toString,
           category = Option(TerraSpendCategories.Other)
         ),
         SpendReportingForDateRange(
-          formatCategoryCost(record.totalStorage),
+          toBigDecimal(record.totalStorage).toString(),
           credits,
           currencyCode.toString,
           category = Option(TerraSpendCategories.Storage)
         ),
         SpendReportingForDateRange(
-          formatCategoryCost(record.totalCompute),
+          toBigDecimal(record.totalCompute).toString(),
           credits,
           currencyCode.toString,
           category = Option(TerraSpendCategories.Compute)
@@ -155,7 +145,7 @@ object WorkspaceSpendReportRecord {
 //        getRoundedNumericValue("other_credits") + getRoundedNumericValue("storage_credits") + getRoundedNumericValue(
 //          "compute_credits"
 //        )
-      total = total + toBigDecimal(total_cost.getOrElse(0.0f))
+      total = total + toBigDecimal(total_cost)
 //      total_credits = total_credits + credits
       start = convertLocalDateTimeToJodaDateTime(record.reportStartDate)
       end = convertLocalDateTimeToJodaDateTime(record.reportEndDate)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
@@ -3,11 +3,12 @@ package org.broadinstitute.dsde.rawls.dataaccess.slick
 import akka.http.scaladsl.model.StatusCodes
 import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
 
-import java.time.{LocalDateTime, ZoneId, ZonedDateTime}
+import java.time.{Instant, LocalDateTime, ZoneId, ZoneOffset, ZonedDateTime}
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.joda.time.DateTime
 
+import java.sql.Timestamp
 import java.util.Currency
 import scala.language.{postfixOps, reflectiveCalls}
 import scala.math.BigDecimal.RoundingMode
@@ -15,8 +16,8 @@ import scala.math.BigDecimal.RoundingMode
 case class WorkspaceSpendReportRecord(
   id: Long,
   googleProjectId: String,
-  reportStartDate: LocalDateTime,
-  reportEndDate: LocalDateTime,
+  reportStartDate: Timestamp,
+  reportEndDate: Timestamp,
   currency: String,
   isDataAvailable: Boolean,
   totalCompute: Option[Float],
@@ -33,8 +34,8 @@ object WorkspaceSpendReportRecord {
     WorkspaceSpendReportRecord(
       workspaceSpendReport.id,
       workspaceSpendReport.googleProjectId,
-      workspaceSpendReport.reportStartDate,
-      workspaceSpendReport.reportEndDate,
+      new Timestamp(workspaceSpendReport.reportStartDate.toInstant(ZoneOffset.UTC).toEpochMilli),
+      new Timestamp(workspaceSpendReport.reportEndDate.toInstant(ZoneOffset.UTC).toEpochMilli),
       workspaceSpendReport.currency,
       workspaceSpendReport.isDataAvailable,
       workspaceSpendReport.totalCompute,
@@ -49,8 +50,8 @@ object WorkspaceSpendReportRecord {
     WorkspaceSpendReport(
       record.id,
       record.googleProjectId,
-      record.reportStartDate,
-      record.reportEndDate,
+      LocalDateTime.ofInstant(record.reportStartDate.toInstant, ZoneOffset.UTC),
+      LocalDateTime.ofInstant(record.reportEndDate.toInstant, ZoneOffset.UTC),
       record.currency,
       record.isDataAvailable,
       record.totalCompute,
@@ -231,9 +232,9 @@ trait WorkspaceSpendReportComponent {
 
     def googleProjectId = column[String]("GOOGLE_PROJECT_ID", O.Length(254))
 
-    def reportStartDate = column[LocalDateTime]("REPORT_START_DATE", O.SqlType("DATETIME"))
+    def reportStartDate = column[Timestamp]("REPORT_START_DATE", O.SqlType("DATETIME"))
 
-    def reportEndDate = column[LocalDateTime]("REPORT_END_DATE", O.SqlType("DATETIME"))
+    def reportEndDate = column[Timestamp]("REPORT_END_DATE", O.SqlType("DATETIME"))
 
     def currency = column[String]("CURRENCY", O.Length(100))
 
@@ -284,7 +285,8 @@ trait WorkspaceSpendReportComponent {
         .filter(x =>
           x.googleProjectId.inSetBind(
             projectIds.map(_.value)
-          ) && x.reportStartDate === startDate && x.reportEndDate === endDate
+          ) && x.reportStartDate === new Timestamp(startDate.toInstant(ZoneOffset.UTC).toEpochMilli)
+            && x.reportEndDate === new Timestamp(endDate.toInstant(ZoneOffset.UTC).toEpochMilli)
         )
 
     private def loadWorkspaceSpendReport(lookup: WorkspaceSpendReportQueryType): ReadAction[Seq[WorkspaceSpendReport]] =
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
@@ -1,7 +1,9 @@
 package org.broadinstitute.dsde.rawls.dataaccess.slick
 
 import java.time.LocalDateTime
+import java.time.ZoneId
 import org.broadinstitute.dsde.rawls.model._
+import org.joda.time.DateTime
 
 import scala.language.postfixOps
 
@@ -42,6 +44,45 @@ object WorkspaceSpendReportRecord {
       record.otherSpend,
       record.isDataAvailable
     )
+
+  def convertJodaToJava(dateTimeOpt: Option[DateTime]): LocalDateTime = {
+    dateTimeOpt match {
+      case Some(dateTime) =>
+        val instant = dateTime.toInstant
+        LocalDateTime.ofInstant(java.time.Instant.ofEpochMilli(dateTime.getMillis), ZoneId.systemDefault())
+      case None =>
+        throw new IllegalArgumentException("DateTime value is missing")
+    }
+  }
+
+  def fromSpendReportingResults(spendReportingResults: SpendReportingResults): Seq[WorkspaceSpendReport] = {
+    val summary = spendReportingResults.spendSummary
+    spendReportingResults.spendDetails.flatMap {
+      spendDetail => spendDetail.spendData.map {
+        var totalStorage: Option[Float] = None
+        var totalCompute: Option[Float] = None
+        var otherSpend: Option[Float] = None
+        spendData => spendData.subAggregation.get.spendData.foreach({
+          categorySpend => {
+            val categoryCost = Some(categorySpend.cost.toFloat)
+            categorySpend.category match {
+              case Some(TerraSpendCategories.Storage) => totalStorage = categoryCost
+              case Some(TerraSpendCategories.Compute) => totalCompute = categoryCost
+              case Some(TerraSpendCategories.Other) => otherSpend = categoryCost
+              case None => throw new IllegalArgumentException("Spend data has no category") // Does this ever happen?
+            }
+          }
+        })
+        // TODO: Store credits and currency (what about aggregation key)?
+        WorkspaceSpendReport.newWorkspaceSpendReport(
+          spendData.googleProjectId.get.value,
+          convertJodaToJava(summary.startTime),
+          convertJodaToJava(summary.endTime),
+          totalCompute,
+          totalStorage,
+          otherSpend,
+          isDataAvailable = true) // TODO: Handle N/A case
+      }}}
 }
 
 trait WorkspaceSpendReportComponent {
@@ -85,20 +126,19 @@ trait WorkspaceSpendReportComponent {
 
   object WorkspaceSpendReportQuery extends TableQuery(new WorkspaceSpendReportTable(_)) {
 
-    def getWorkspaceSpendReportByProjectIdsAndReportDate(projectIds: Seq[String],
+    def getWorkspaceSpendReportByProjectIdsAndReportDate(projectIds: Set[String],
                                                          startDate: LocalDateTime,
                                                          endDate: LocalDateTime
                                             ): ReadAction[Seq[WorkspaceSpendReport]] =
       loadWorkspaceSpendReport(filterByProjectIdsAndReportDate(projectIds, startDate, endDate))
 
 
-    def filterByProjectIdsAndReportDate(projectIds: Seq[String],
+    def filterByProjectIdsAndReportDate(projectIds: Set[String],
                                 startDate: LocalDateTime,
                                 endDate: LocalDateTime
-                               ): ReadAction[Seq[WorkspaceSpendReportRecord]] =
+                               ) =
       workspaceSpendReportQuery
         .filter(x => x.googleProjectId.inSetBind(projectIds.map(_.value)) && x.reportStartDate === startDate && x.reportEndDate === endDate)
-        .result
 
     private def loadWorkspaceSpendReport(lookup: WorkspaceSpendReportQueryType): ReadAction[Seq[WorkspaceSpendReport]] =
       for {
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+)
+
+object WorkspaceSpendReportRecord {
+
+  def fromWorkspaceSpendReport(workspaceSpendReport: WorkspaceSpendReport): WorkspaceSpendReportRecord =
+    WorkspaceSpendReportRecord(
+      workspaceSpendReport.id,
+      workspaceSpendReport.googleProjectId,
+      new Timestamp(workspaceSpendReport.reportStartDate.toInstant(ZoneOffset.UTC).toEpochMilli),
+      new Timestamp(workspaceSpendReport.reportEndDate.toInstant(ZoneOffset.UTC).toEpochMilli),
```
davidangb: can you simplify this to `Timestamp.valueOf(workspaceSpendReport. reportEndDate)`? Is there a complication with the timezone?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+      workspaceSpendReport.storageCredits,
+      workspaceSpendReport.otherCredits
+    )
+
+  def toWorkspaceSpendReport(record: WorkspaceSpendReportRecord): WorkspaceSpendReport =
+    WorkspaceSpendReport(
+      record.id,
+      record.googleProjectId,
+      LocalDateTime.ofInstant(record.reportStartDate.toInstant, ZoneOffset.UTC),
+      LocalDateTime.ofInstant(record.reportEndDate.toInstant, ZoneOffset.UTC),
```
davidangb: Similar comment; can you simplify this to `record.reportEndDate.toLocalDateTime`?

Edit: there's a bunch of other conversions elsewhere - this comment would apply generally to all of them

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+      record.computeCredits,
+      record.storageCredits,
+      record.otherCredits
+    )
+
+  def convertJodaToJava(dateTimeOpt: Option[DateTime]): LocalDateTime =
+    dateTimeOpt match {
+      case Some(dateTime) =>
+        val instant = dateTime.toInstant
+        LocalDateTime.ofInstant(java.time.Instant.ofEpochMilli(dateTime.getMillis), ZoneId.systemDefault())
```
davidangb: can this be just `dateTime.toLocalDateTime`?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+      )
+
+      val totalCompute: Option[Float] = record.totalCompute
+      val totalStorage: Option[Float] = record.totalStorage
+      val otherSpend: Option[Float] = record.otherSpend
+      val cost: Option[Float] = for {
+        cost1 <- totalCompute
+        cost2 <- totalStorage
+        cost3 <- otherSpend
+      } yield cost1 + cost2 + cost3
```
davidangb: You could use a one-liner here:
```scala
val cost: Option[Float] = Option(Seq(totalCompute, totalStorage, otherSpend).flatten.sum)
```
your call on whichever is more readable

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -571,28 +574,87 @@ class SpendReportingService(
...
+                                        projectNames: Map[GoogleProjectId, WorkspaceName],
+                                        start: DateTime,
+                                        end: DateTime
+  ): Unit = {
+    val reports = results.collect { case Some(report) => report }
+    val googleProjectsWithReports =
+      reports.map(r => r.spendSummary.googleProjectId).collect { case Some(project) => project.toString() }.toSeq
+    val filteredProjectNames: Map[GoogleProjectId, WorkspaceName] = projectNames.filter { case (id, _) =>
+      !googleProjectsWithReports.contains(id.toString())
     }
```
davidangb: This code creates `googleProjectsWithReports ` as a `Seq`, then checks a bunch of other values to see if they exist in that `Seq`. When doing lookups like that, a `Set` is much more performant. At this scale it might not matter, but it's a good practice. Additionally, once you have a `Set` you can use its utility methods to help with diff/intersection/union/etc.

I might rewrite this code as:
```scala
    // extract all google project ids from the supplied SpendReportingResults
    val googleProjectsWithReports: Set[GoogleProjectId] =
      results                                     // Iterable[Option[SpendReportingResults]]
        .flatten                                  // Iterable[SpendReportingResults]
        .flatMap(_.spendSummary.googleProjectId)  // Iterable[GoogleProject]
        .map(p => GoogleProjectId(p.value))       // Iterable[GoogleProjectId]
        .toSet                                    // Set[GoogleProjectId]
    // find all supplied project names which are not in the SpendReportingResults projects
    val missingProjectIds: Set[GoogleProjectId] = projectNames.keySet diff googleProjectsWithReports
```

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+      case head :: List() => Currency.getInstance(head)
+      case head :: tail =>
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.InternalServerError,
+          s"Inconsistent currencies found while aggregating spend data: $head and ${tail.head} cannot be combined"
+        )
+      case List() => throw RawlsExceptionWithErrorReport(StatusCodes.NotFound, "No currencies found for spend data")
+    }
+
+    val all = records.map { record =>
```
calypsomatic, line 174: This chunk of code seems pretty similar to what's in `SpendReportingService`.  Is there a way it can be reused?

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,304 @@
...
+            summary.currency,
+            isDataAvailable = true,
+            totalCompute,
+            totalStorage,
+            otherSpend,
+            computeCredits,
+            storageCredits,
+            otherCredits
+          )
+      }
```
davidangb: Use of vars in Scala is generally a code smell. You can avoid that by arranging the incoming data into a more friendly format, then extracting from that data (think about the ARRANGE/ACT/ASSERT pattern for tests, without ASSERT). I would do it something like this:
```scala
      spendDetail.spendData.map { spendData =>
        // create a map of category->spend
        val spendCategoryMap: Map[Option[TerraSpendCategories.TerraSpendCategory], SpendReportingForDateRange] =
          spendData.subAggregation
            .flatMap(_.spendData)
            .map(datedReport => datedReport.category -> datedReport)
            .toMap
        // helper to look up the cost and credits for a given category
        def getCategoryCost(category: TerraSpendCategories.TerraSpendCategory): (Option[Float], Option[Float]) =
          spendCategoryMap
            .get(Some(category))
            .map { categorySpend =>
              (Option(categorySpend.cost.toFloat), Option(categorySpend.credits.toFloat))
            }
            .getOrElse((None, None))

        // get cost and credits for the categories we care about
        val (totalStorage, storageCredits) = getCategoryCost(TerraSpendCategories.Storage)
        val (totalCompute, computeCredits) = getCategoryCost(TerraSpendCategories.Compute)
        val (otherSpend, otherCredits) = getCategoryCost(TerraSpendCategories.Other)

        WorkspaceSpendReport.newWorkspaceSpendReport(
          spendData.googleProjectId.get.value,
          convertJodaToJava(summary.startTime),
          convertJodaToJava(summary.endTime),
          summary.currency,
          isDataAvailable = true,
          totalCompute,
          totalStorage,
          otherSpend,
          computeCredits,
          storageCredits,
          otherCredits
        )
      }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -571,28 +574,87 @@ class SpendReportingService(
           return Future.successful(None)
         }
         projectNames: Map[GoogleProjectId, WorkspaceName] = billingMap.values.flatten.toMap
+        projectIds = projectNames.keySet.map(x => x.value)
```
davidangb: nit: you don't need an `x` variable here:
```suggestion
        projectIds = projectNames.keySet.map(_.value)
```

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -571,28 +574,87 @@ class SpendReportingService(
...
         tableToProjectIdsMap: Map[String, Seq[GoogleProjectId]] = billingMap.map { case (table, workspaces) =>
           table -> workspaces.map(_._1)
         }
+
+        cacheResults <- checkCachedWorkspaceSpend(projectIds, start, end)
+        _ = if (cacheResults.size == projectIds.size) {
+          val hasDataAvailable = cacheResults.filter(cached => cached.isDataAvailable)
+          return Future.successful(
+            Some(workspaceSpendReportRepository.toSpendReportResults(hasDataAvailable, projectNames))
+          )
```
davidangb, line 546: early `return`s are non-idiomatic Scala … fun with Scala! Because this for-comprehension is already pretty big, I'd look to solve this by breaking out everything after the `return` into a separate method. Then, you could do something like:
```scala
        combinedResults <-
          if (cacheResults.size == projectIds.size) {
            val hasDataAvailable = cacheResults.filter(cached => cached.isDataAvailable)
            Future.successful(
              Some(workspaceSpendReportRepository.toSpendReportResults(hasDataAvailable, projectNames))
            )
          } else {
            getUncachedSpendForAllWorkspaces(tableToProjectIdsMap,
                                             projectNames,
                                             start,
                                             end,
                                             pageSize,
                                             offset,
                                             childContext
            )
          }
      } yield combinedResults

      private def getUncachedSpendForAllWorkspaces( …
```

davidangb (commented):
Some early feedback for you to think about, mostly about Scala idioms

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
samanehsan: That looks much better!!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
samanehsan: I'm getting an error on this line `.flatMap(_.spendData)` that I can't quite figure out how to fix 🤔 
```
type mismatch;
 found   : Seq[org.broadinstitute.dsde.rawls.model.SpendReportingForDateRange]
 required: Option[?]

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala (hunk 1 above)
Comment (samanehsan) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala]: Yeah I noticed a timezone difference, but let me try this again...maybe the issue was with my test inputs?

File: core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala (hunk 7 above)
Comment (davidangb) [reply core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala]: Yup, there was definitely an error in there … does this work better?
```scala
        val spendCategoryMap: Map[Option[TerraSpendCategories.TerraSpendCategory], SpendReportingForDateRange] =
          spendData.subAggregation
            .map(_.spendData)
            .getOrElse(Seq())
            .map(datedReport => datedReport.category -> datedReport)
            .toMap
```

samanehsan: That fixed it, thanks!!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
samanehsan: I was doing some manual testing and noticed an issue in how I'm handling these timestamps. For instance if the end datetime is `2025-02-10T23:59:59.999Z`, it gets converted to the timestamp before inserting it into the db `2025-02-10 23:59:59.999` but it gets stored as `2025-02-11 00:00:00` in mysql 🤔

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
davidangb: +1, would love to see alignment/reuse here

samanehsan:
The test that's failing is flaky `SpendReportingServiceSpec.getSpendForAllWorkspaces should get the spend report from multiple billing projects` and I'm trying to figure out why the results are sometimes none 🤔

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,255 @@
...
+      val currencyCode = Currency.getInstance(currencyString)
+      val projectId = record.googleProjectId
+
+      val subAggregation = List(
+        SpendReportingForDateRange(
+          SpendReportUtils.toBigDecimal(record.otherSpend, currencyCode).toString(),
+          SpendReportUtils.toBigDecimal(record.otherCredits, currencyCode).toString(),
+          currencyCode.toString,
+          category = Option(TerraSpendCategories.Other)
+        ),
```
davidangb: the `toBigDecimal` will turn `None` into `0.00` … is that correct here? That means we will never report `N/A`, we'll report `0`. This may be what we want; I am not sure of the requirements

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+  def getCurrency(currencies: Seq[String]): Currency =
+    currencies.distinct match {
+      case head :: List() => Currency.getInstance(head)
+      case head :: tail =>
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.InternalServerError,
+          s"Inconsistent currencies found while aggregating spend data: $head and ${tail.head} cannot be combined"
+        )
+      case List() => throw RawlsExceptionWithErrorReport(StatusCodes.NotFound, "No currencies found for spend data")
+    }
```
davidangb, line 23: this also duplicates code in `SpendReportingService.extractSpendReportingResults`, would be good to consolidate

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -617,26 +615,88 @@ class SpendReportingService(
...
+        endLocalDateTime,
+        "USD",
+        isDataAvailable = false,
+        Option.empty,
+        Option.empty,
+        Option.empty,
+        Option.empty,
+        Option.empty,
+        Option.empty
+      )
```
davidangb: You could make a new method `WorkspaceSpendReport.newEmptyReport` or whatever you want to name it, and include all the defaults so you don't have to write them all out here. Optional; your call

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -617,26 +615,88 @@ class SpendReportingService(
...
+                      Some(crossBillingResults)
+                  }
+                }
+                .recoverWith { case ex: Throwable =>
+                  logger.warn(s"Error fetching results from BigQuery: ${ex.getMessage}")
+                  Future.successful(None)
+                }
+              queryResults.flatMap { res =>
+                Future.successful(insertRecordsWithMissingSpendData(res, workspaceNamesByProjectId, start, end))
+              }
```
davidangb: Two suggestions here:

1. you can simplify this with `.map` instead of `.flatMap`:
```suggestion
              queryResults.map { res =>
                insertRecordsWithMissingSpendData(res, workspaceNamesByProjectId, start, end)
              }

```

2. I think this should catch and ignore any errors that happen when saving records to the cache. If we hit an error here, we can still deliver results to the end user; we'll just fail to cache those results. There is a race condition which makes errors here likely: since the calls to `workspaceSpendReportRepository.getWorkspaceSpendReports` and `workspaceSpendReportRepository.insertWorkspaceSpendReport` are each in their own separate transaction, if multiple users hit an uncached consolidated spend report at the same time, both could see no records in cache, and then both would try to write those records to cache. One of those writes will fail due to the unique constraints on the cache table.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -1761,6 +1793,186 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with Matchers with Mocki
...
+
+    spendSummary.credits shouldBe toBigDecimal(zero.toString).toString()
+    spendSummary.cost shouldBe toBigDecimal(total.toString).toString()
+    spendSummary.currency shouldBe "USD"
+    spendSummary.startTime.get.toString(ISODateTimeFormat.date()) shouldBe from.toString(
+      ISODateTimeFormat.date()
+    )
+    spendSummary.endTime.get.toString(ISODateTimeFormat.date()) shouldBe to.toString(ISODateTimeFormat.date())
+    verifyNoInteractions(bigQueryService)
+  }
```
davidangb, line 1995: Suggestions for additional unit tests: 
1. What happens when cached results exist for the requested dates, but only cover a subset of the requested project?
2. What happens when cached results exist for all the requested projects, but for the wrong dates?
3. A test that verifies that after we get results from BigQuery, we persist those results to the cache. This may need to use an `eventually` since the cache-writes are fire-and-forget.

davidangb (commented):
more comments inline, this is getting close!

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
samanehsan: I think that's okay (assuming I'm understanding the behavior correctly)? It looks like the UI displays `N/A` on the consolidated spend report if a workspace is not included in the API response from Rawls. When we initially query the data, this should insert an empty record for the workspace with isDataAvailable=False. When the cached results are returned, only records with isDataAvailable=True are included in the API response.

samanehsan:
@davidangb do you have any idea why SpendReportingServiceSpec.getSpendForAllWorkspaces `should get the spend report from multiple billing projects` and `should get the spend report from multiple billing projects when cached reports are from different dates` are flaky? 🤔 Sometimes the result of `getSpendForAllWorkspaces` in those tests is None

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -1440,92 +1574,75 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with Matchers with Mocki
...
+    val bigQueryService = mock[GoogleBigQueryService[IO]](RETURNS_SMART_NULLS)
+    val job1 = mock[Job]
+    when(job1.getQueryResults(any())).thenReturn(createTableResult(table1))
+    when(job1.getStatistics).thenReturn(mock[JobStatistics.QueryStatistics](RETURNS_SMART_NULLS))
+    when(job1.waitFor()).thenReturn(job1)
+    val job2 = mock[Job]
+    when(job2.getQueryResults(any())).thenReturn(createTableResult(table2))
+    when(job2.getStatistics).thenReturn(mock[JobStatistics.QueryStatistics](RETURNS_SMART_NULLS))
+    when(job2.waitFor()).thenReturn(job2)
+    when(bigQueryService.runJob(any(), any())).thenReturn(IO(job1), IO(job2))
```
davidangb: @samanehsan I'm pretty sure this is the cause of flakiness. this always returns job1 and job2 in that order, but they are not always executed in that order, so sometimes execution fails.

samanehsan:
@calypsomatic @davidangb let me know if there are other changes you'd like to see, or if I can answer any questions!

calypsomatic (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/WorkspaceSpendReportComponent.scala
```diff
@@ -0,0 +1,291 @@
...
+                               totalStorage: Float,
+                               storageCredits: Float,
+                               totalCompute: Float,
+                               computeCredits: Float,
+                               otherSpend: Float,
+                               otherCredits: Float,
+                               projectNames: Map[GoogleProjectId, WorkspaceName]
+  ): SpendReportingForDateRange = {
+    val currencyString = currency
+    val currencyCode = Currency.getInstance(currencyString)
```
davidangb: nit:
```suggestion
    val currencyCode = Currency.getInstance(currency)
```

davidangb (approved):
one tiny nitpick inline. Nice job getting this through! I am very curious to see how it performs in practice.

samanehsan:
@davidangb I just did a final round of manual testing as a sanity check and found that adding a jobId for testing is causing this error:

```Error fetching results from BigQuery: Invalid job ID "broad-gcp-billing.gcp_billing_export_views.gcp_billing_export_v1_001AC2_2B914D_822931". Job IDs must be alphanumeric (plus underscores and dashes) and must be at most 1024 characters long```

What do you think about using `jobId.hashCode.toString` to create a hash of the job id?

davidangb:
> What do you think about using jobId.hashCode.toString to create a hash of the job id?

@samanehsan I worry just a little bit about hash collisions when using .hashCode, even though it's probably just fine. Maybe base64-encode the value via `Base64.getEncoder.encodeToString(jobId.getBytes).take(1023)`? If a hash collision is only a problem for unit tests it's no big deal … I wonder if BigQuery would have any kind of problem if it saw duplicate job ids … probably not?

samanehsan:
What about making the jobId param an optional and only setting the job id if it exists? Then we can just change this behavior for the tests.

samanehsan:
@davidangb I tested the behavior, and re-using the same job id (which would be the spend export table name) in BigQuery returns the same query results 🤯: That is not what I expected, based on https://cloud.google.com/bigquery/docs/managing-jobs#repeat_jobs

Specifically I ran a query to get workspace spend reports from this year, then when I changed the report dates to last November it re-queried BigQuery but returned the same results (including a workspace that didn't even exist until January)

PR #3201, opened 2025-02-26
CORE-349: request estimated costs from Cromwell for all workflows

davidangb:
Ticket: CORE-349

All workflows now request their estimated cost from Cromwell and display that cost in the UI. Before this PR, we only requested estimated cost for workflows if a per-workflow cost threshold was set for the submission.

This behavior can be disabled via the `submissionmonitor.enableCostEstimatesForAllWorkflows` config flag, which is set to true in reference.conf. If that flag is set to false, behavior will revert to the prior functionality wherein we only request&persist cost estimates if the submission has a cost threshold. This config flag will allow us to easily turn off the feature if Cromwell performance cannot keep up to demand.

---

**PR checklist**

- [x] If you changed anything in `model/`, then you should publish a new official `rawls-model` and perform the corresponding dependency updates as specified in the [README](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model):
  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
@@ -26,6 +26,7 @@ trait ExecutionServiceCluster extends ErrorReportable {
                       userInfo: UserInfo
   ): Future[(ExecutionServiceId, Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]])]
 
+  // currently unused
   def status(workflowRec: WorkflowRecord, userInfo: UserInfo): Future[ExecutionServiceStatus]
 
   def outputs(workflowRec: WorkflowRecord, userInfo: UserInfo): Future[ExecutionServiceOutputs]
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
@@ -26,7 +26,6 @@ trait ExecutionServiceCluster extends ErrorReportable {
                       userInfo: UserInfo
   ): Future[(ExecutionServiceId, Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]])]
 
-  // currently unused
   def status(workflowRec: WorkflowRecord, userInfo: UserInfo): Future[ExecutionServiceStatus]
 
   def outputs(workflowRec: WorkflowRecord, userInfo: UserInfo): Future[ExecutionServiceOutputs]
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
@@ -313,8 +313,9 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
     executionContext: ExecutionContext
   ): Future[Option[WorkflowRecord]] =
     (workflowRec.externalId, perWorkflowCostCap) match {
-      // fetch cost information for the workflow if submission has a cost cap threshold defined
-      case (Some(externalId), _) =>
+      // fetch cost information for the workflow if submission has a cost cap threshold defined or if the
+      // enableCostEstimatesForAllWorkflows config flag is set to true
+      case (Some(externalId), _) if config.enableCostEstimatesForAllWorkflows || perWorkflowCostCap.isDefined =>
         for {
           costBreakdown <- executionServiceCluster.getCost(workflowRec, petUser)
           updatedWorkflowRec <-
@@ -363,6 +364,13 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
               }
             }
         } yield updatedWorkflowRec
+      // if config enableCostEstimatesForAllWorkflows is set to false, and no cost cap is defined,
+      // fetch workflow status only
+      case (Some(externalId), _) =>
+        executionServiceCluster.status(workflowRec, petUser).map { newStatus =>
+          if (newStatus.status != workflowRec.status) Option(workflowRec.copy(status = newStatus.status))
+          else None
+        }
       case _ => Future.successful(None)
     }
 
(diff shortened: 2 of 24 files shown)


core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -2339,7 +2347,9 @@ class SubmissionTestExecutionServiceDAO(workflowStatus: => String, workflowCost:
   }
 
   override def getCost(id: String, userInfo: UserInfo): Future[WorkflowCostBreakdown] =
-    Future.successful(WorkflowCostBreakdown(id, workflowCost, "USD", workflowStatus, Seq.empty))
+    if (abortedMap.keySet.contains(id))
+      Future(WorkflowCostBreakdown(id, workflowCost, "USD", WorkflowStatuses.Aborted.toString, Seq.empty))
+    else Future(WorkflowCostBreakdown(id, workflowCost, "USD", workflowStatus, Seq.empty))
```
davidangb, line 2432: the submission monitor now calls `getCost()` for all workflows instead of calling `status()` for non-capped workflows and  `getCost()` for capped workflows. Thus, the abort-related logic in this mock's `status()` needed to be copied into this mock's `getCost()`.

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -100,7 +100,7 @@ class SubmissionMonitorSpec(_system: ActorSystem)
             workflowsRecs.map { workflowRec =>
               scala.util.Success(
                 Option(
-                  (workflowRec.copy(status = WorkflowStatuses.Succeeded.toString),
+                  (workflowRec.copy(status = WorkflowStatuses.Succeeded.toString, cost = Option(BigDecimal.valueOf(0))),
                    Some(ExecutionServiceOutputs(workflowRec.externalId.get, Map("o1" -> Left(AttributeString("foo")))))
```
davidangb, line 104: all workflows now return an estimated cost, so test expectations needed updating

core/src/test/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorSpec.scala
```diff
@@ -129,7 +129,13 @@ class SubmissionMonitorSpec(_system: ActorSystem)
...
       ignoreStatusLastChangedDate(
         ExecutionServiceStatusResponse(
-          workflowsRecs.map(workflowRec => scala.util.Success(None))
+          workflowsRecs.map { workflowRec =>
+            scala.util.Success(
+              Option(
+                (workflowRec.copy(cost = Option(BigDecimal.valueOf(0))), None)
+              )
+            )
+          }
```
davidangb, line 138: previously, the monitor would return None because the workflow's status hadn't changed. Now, since the monitor compares both status and cost, and the cost _has_ changed, it returns Some instead of None.

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
```diff
@@ -26,6 +26,7 @@ trait ExecutionServiceCluster extends ErrorReportable {
                       userInfo: UserInfo
   ): Future[(ExecutionServiceId, Seq[Either[ExecutionServiceStatus, ExecutionServiceFailure]])]
 
+  // currently unused
   def status(workflowRec: WorkflowRecord, userInfo: UserInfo): Future[ExecutionServiceStatus]
```
davidangb: I chose not to delete the now-unused `status()` method, but I could be convinced it needs cleaning up

core/src/main/scala/org/broadinstitute/dsde/rawls/jobexec/SubmissionMonitorActor.scala
```diff
@@ -362,12 +363,6 @@ trait SubmissionMonitor extends FutureSupport with LazyLogging with RawlsInstrum
               }
             }
         } yield updatedWorkflowRec
-      // fetch workflow status only if cost cap threshold is not defined
-      case (Some(externalId), None) =>
-        executionServiceCluster.status(workflowRec, petUser).map { newStatus =>
-          if (newStatus.status != workflowRec.status) Option(workflowRec.copy(status = newStatus.status))
-          else None
-        }
```
davidangb: collapsing these two cases into one is the core of this PR

core/src/main/scala/org/broadinstitute/dsde/rawls/submissions/SubmissionsService.scala
```diff
@@ -439,20 +439,28 @@ class SubmissionsService(
...
+                                                       submission.submissionDate,
+                                                       submissionDoneDate,
+                                                       tableName
+              )
             )
-          ) map { costMapTry => annotateSubmissionWithActualCost(submission, costMapTry) }
+          }
         }
+
+        costMapTryFuture map { costMapTry => annotateSubmissionWithActualCost(submission, costMapTry) }
```
davidangb, line 463: this diff is easier to read imo if you ignore whitespace.

This isn't strictly necessary to complete this PR, but it skips a useless SQL query for all workflows which are 1) younger than 24 hours or 2) are still running

dvoet:
Is getting the cost estimate an expensive thing for cromwell? Will this change put a lot more load on cromwell?

dvoet (approved):
I am ok with this so long as the analysis team is ok with getting estimates for all workflows instead of only those with a cost threshold

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
dvoet: shrug, why do you want to keep it?

kevinmarete (approved):

jgainerdewar (approved):

kevinmarete (approved):

dvoet (approved):

core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/ExecutionServiceCluster.scala
davidangb: this is now used again, if we turn the feature off via config

PR #3205, opened 2025-03-02
[CORE-340] Add ID Column to BILLING_PROJECT Table & New getBillingProjectById API

kevinmarete:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-340

This PR has the following changes:

- Adds a new non-nullable `ID` column (UUID) to the `BILLING_PROJECT` table in Rawls. While the `NAME` column remains the primary identifier, this UUID will support TDR’s transition to using UUID-based references.
- Adds a new API `/api/billing/v2/id/{projectUuid}` to fetch a billing project by id, mirroring the existing getBillingProject API.

Updated 
---

**PR checklist**

- [X] Include the JIRA issue number in the PR description and title
- [X] If you changed anything in `model/`, then you should publish a new official `rawls-model` and perform the corresponding dependency updates as specified in the [README](https://github.com/broadinstitute/rawls/blob/develop/README.md#publish-rawls-model):
  - [ ] in workbench-libs
- [X] Verify all tests go green, including CI tests

### DIFF
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
@@ -2,14 +2,9 @@
 <databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
     <changeSet id="add_billing_project_id" author="kmarete" logicalFilePath="dummy">
         <addColumn tableName="BILLING_PROJECT">
-            <column name="ID" type="BINARY(16)">
+            <column name="ID" type="BINARY(16)" valueComputed="UNHEX(MD5(NAME))">
                 <constraints nullable="false" />
             </column>
         </addColumn>
-
-        <!-- Populate ID column with UUID derived from NAME (Unique PK) -->
-        <update tableName="BILLING_PROJECT">
-            <column name="ID" valueComputed="UNHEX(MD5(NAME))"/>
-        </update>
     </changeSet>
 </databaseChangeLog>
--- a/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
+++ b/core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd">
+    <changeSet id="add_billing_project_id" author="kmarete" logicalFilePath="dummy">
+        <addColumn tableName="BILLING_PROJECT">
+            <column name="ID" type="BINARY(16)">
+                <constraints nullable="false" />
+            </column>
+        </addColumn>
+
+        <!-- Populate ID column with UUID derived from NAME (Unique PK) -->
+        <update tableName="BILLING_PROJECT">
+            <column name="ID" valueComputed="UNHEX(MD5(NAME))"/>
+        </update>
+    </changeSet>
+</databaseChangeLog>
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -32,7 +32,6 @@ import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport, StringValid
 import org.broadinstitute.dsde.workbench.dataaccess.NotificationDAO
 import org.broadinstitute.dsde.workbench.model.{Notifications, WorkbenchEmail, WorkbenchUserId}
 
-import java.security.MessageDigest
 import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -161,11 +160,7 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
       )
       _ <- billingRepository.createBillingProject(
         RawlsBillingProject(
-          createProjectRequest.id.getOrElse { // if no id is provided, generate one from the project name
-            UUID.nameUUIDFromBytes(
-              MessageDigest.getInstance("MD5").digest(createProjectRequest.projectName.value.getBytes)
-            )
-          },
+          UUID.randomUUID(), // New ID[UUID] not PK of the billing project
           createProjectRequest.projectName,
           CreationStatuses.Creating,
           createProjectRequest.billingAccount,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -159,12 +159,14 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
         None
       )
       _ <- billingRepository.createBillingProject(
-        RawlsBillingProject(createProjectRequest.projectName,
-                            CreationStatuses.Creating,
-                            createProjectRequest.billingAccount,
-                            None,
-                            None,
-                            createProjectRequest.servicePerimeter
+        RawlsBillingProject(
+          UUID.randomUUID(), // New ID[UUID] not PK of the billing project
+          createProjectRequest.projectName,
+          CreationStatuses.Creating,
+          createProjectRequest.billingAccount,
+          None,
+          None,
+          createProjectRequest.servicePerimeter
         )
       )
     } yield notificationDAO.fireAndForgetNotifications(invites)
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
@@ -32,6 +32,7 @@ import org.broadinstitute.dsde.rawls.{RawlsExceptionWithErrorReport, StringValid
 import org.broadinstitute.dsde.workbench.dataaccess.NotificationDAO
 import org.broadinstitute.dsde.workbench.model.{Notifications, WorkbenchEmail, WorkbenchUserId}
 
+import java.security.MessageDigest
 import java.util.UUID
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -160,7 +161,11 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
       )
       _ <- billingRepository.createBillingProject(
         RawlsBillingProject(
-          UUID.randomUUID(), // New ID[UUID] not PK of the billing project
+          createProjectRequest.id.getOrElse { // if no id is provided, generate one from the project name
+            UUID.nameUUIDFromBytes(
+              MessageDigest.getInstance("MD5").digest(createProjectRequest.projectName.value.getBytes)
+            )
+          },
           createProjectRequest.projectName,
           CreationStatuses.Creating,
           createProjectRequest.billingAccount,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
@@ -95,7 +95,7 @@ case class WorkspaceBillingAccount(
 case class RawlsBillingProjectOrganization(enterprise: Boolean, limits: Map[String, String])
 
 case class RawlsBillingProjectResponse(
-  id: String,
+  id: UUID,
   projectName: RawlsBillingProjectName,
   billingAccount: Option[RawlsBillingAccountName],
   servicePerimeter: Option[ServicePerimeterName],
@@ -120,7 +120,7 @@ object RawlsBillingProjectResponse {
     region: Option[String] = None,
     organization: Option[RawlsBillingProjectOrganization] = None
   ): RawlsBillingProjectResponse = this(
-    project.id.toString,
+    project.id,
     project.projectName,
     project.billingAccount,
     project.servicePerimeter,
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
@@ -67,6 +67,7 @@ case class ManagedGroup(membersGroup: RawlsGroup, adminsGroup: RawlsGroup) exten
 
 case class RawlsBillingAccount(accountName: RawlsBillingAccountName, firecloudHasAccess: Boolean, displayName: String)
 case class RawlsBillingProject(
+  id: UUID,
   projectName: RawlsBillingProjectName,
   status: CreationStatuses.CreationStatus,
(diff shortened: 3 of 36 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -85,7 +87,14 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
               get {
                 complete {
                   import spray.json._
-                  userServiceConstructor(ctx).getBillingProject(RawlsBillingProjectName(projectId)).map {
+
+                  // Check if the projectId is a UUID or a billingProjectName
```
marctalbott: I didn’t think about the overlap between the existing `getBillingProject` and new `getBillingProjectById` APIs. My concern with this approach is that I could create a new billing project with a UUID for a name that overlapped with an existing billing project's id which would result in my new billing project being inaccessible to me. 

Instead, how does `GET /api/billing/v2/id/{billingProjectUUID}` sound?

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -281,6 +281,12 @@ class UserService(
     if (roles.nonEmpty) Some(mapCloudPlatformAndPolicies(p, billingProfile, roles, workspaceManagerDAO)) else None
   )
 
+  def getBillingProjectById(id: UUID): Future[Option[RawlsBillingProjectResponse]] =
+    billingRepository.getBillingProjectById(id).flatMap {
+      case Some(billingProject) => getBillingProject(billingProject.projectName)
```
marctalbott: I would recommend extracting the logic in `getBillingProject` that's required to construct the full `RawlsBillingProjectResponse` into a shared helper method rather than calling the full method here and adding a second round trip to the database.

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
```diff
@@ -0,0 +1,15 @@
...
+    <changeSet id="add_billing_project_id" author="kmarete" logicalFilePath="dummy">
+        <addColumn tableName="BILLING_PROJECT">
+            <column name="ID" type="BINARY(16)">
+                <constraints nullable="false" />
+            </column>
+        </addColumn>
+
+        <!-- Populate ID column with UUID derived from NAME (Unique PK) -->
+        <update tableName="BILLING_PROJECT">
+            <column name="ID" valueComputed="UNHEX(MD5(NAME))"/>
```
marctalbott: This seems like a fine way to seed the ids, but there is also a `UUID()` function in mysql that you can use to generate a random UUID if you'd like.

kevinmarete: I was getting an error message that UUIDs may not match in other replicas, so I decided to take a more deterministic approach.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -281,6 +281,12 @@ class UserService(
...
+      case None                 => Future.successful(None)
```
calypsomatic: Shouldn't this be a 404?  Or is that handled further up the chain?  Or is this how it's done elsewhere in the code?

kevinmarete: That is handled further up the chain

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
```diff
@@ -0,0 +1,15 @@
...
+        </update>
+    </changeSet>
```
davidangb, line 9: 1. Do you need separate `<addColumn>` and `<update>` operations? Can you just add the `valueComputed="UNHEX(MD5(NAME))"` attribute to the column inside the `<addColumn>` operation and do away with the `<update>` entirely?
2. Will this allow us (TDR) to insert rows and specify an exact UUID? We will need that functionality to copy over TDR spend profiles. I'm wondering if you need to use `defaultValueComputed` instead of `valueComputed`

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
```diff
@@ -94,6 +95,7 @@ case class WorkspaceBillingAccount(
 case class RawlsBillingProjectOrganization(enterprise: Boolean, limits: Map[String, String])
 
 case class RawlsBillingProjectResponse(
+  id: String,
```
davidangb: `UUID` instead of `String`?

davidangb (commented):
+1 to Marc's comments about not querying the db twice and about creating a unique API signature for the get-by-id API. See comments inline, too:

core/src/main/scala/org/broadinstitute/dsde/rawls/model/UserAuth.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/7d070f460201a426a5fdb8320f42e6293e82d8cc

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
kevinmarete: Fixed 1: https://github.com/broadinstitute/rawls/pull/3205/commits/f2554e9b7323ef1d111bba3cef44fa77dfade619 and 2: https://github.com/broadinstitute/rawls/pull/3205/commits/40daaf845287660732d21f9e2cf33adaaca1bcd0

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/a67f8bf57cbcdcc828805559a4f3c2da04e87ce0

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/3205dfcbb24ab052c6166be8b3845bcdbc6728ee

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
```diff
@@ -159,12 +160,18 @@ class BillingProjectOrchestrator(ctx: RawlsRequestContext,
...
-                            None,
-                            None,
-                            createProjectRequest.servicePerimeter
+        RawlsBillingProject(
+          createProjectRequest.id.getOrElse { // if no id is provided, generate one from the project name
+            UUID.nameUUIDFromBytes(
+              MessageDigest.getInstance("MD5").digest(createProjectRequest.projectName.value.getBytes)
+            )
+          },
+          createProjectRequest.projectName,
```
davidangb, line 164: I suggest removing the id from `CreateRawlsV2BillingProjectFullRequest`, and always generating a `UUID.randomUUID()` for the id here - in other words, don't allow the user to request a specific id.

As a security best practice, we don't allow users to specify their own ids because it allows phishing. A malicious user could start requesting various UUIDs and determine the existence/non-existence of a billing project - which they otherwise have no permission to - based on whether the creation was successful or had an id conflict. This is also why we'd want to use a random UUID instead of a UUID deterministically generated from the name.

In the case of billing projects specifically, this security best practice is somewhat moot, since we allow users to specify a name and we require names to be unique and we allow project to be retrieved by name. Nonetheless, we could follow the best practice for ids.

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
```diff
@@ -269,17 +269,34 @@ class UserService(
...
+
+        for {
+          roles <- rolesFuture
+          billingProfile <- billingProfileFuture
+        } yield
+          if (roles.nonEmpty) Some(mapCloudPlatformAndPolicies(project, billingProfile, roles, workspaceManagerDAO))
+          else None
+
+      case None => Future.successful(None)
     }
```
davidangb: There is no need to wrap the billing profile into a Future just to unwrap that Future inside the for-comprehension. You can change it to be a simple variable, like this:
```scala
    billingProjectFuture.flatMap {
      case Some(project) =>
        val rolesFuture = samDAO
          .listUserRolesForResource(SamResourceTypeNames.billingProject, project.projectName.value, ctx)
          .map(samRolesToProjectRoles)

        val billingProfile =
          project.billingProfileId.flatMap(id => billingProfileManagerDAO.getBillingProfile(UUID.fromString(id), ctx))

        for {
          roles <- rolesFuture
        } yield
          if (roles.nonEmpty) Some(mapCloudPlatformAndPolicies(project, billingProfile, roles, workspaceManagerDAO))
          else None

      case None => Future.successful(None)
    }
```

The other thing I notice in this code - and I realize you are preserving existing functionality - is that we make the call to BPM even in cases where the user has no roles on the project, and then we ignore the result from BPM. Could you make the BPM call conditional on `roles` being non-empty? This could look something like this; feel free to rewrite for readability:

```scala
    billingProjectFuture.flatMap {
      case Some(project) =>
        samDAO
          .listUserRolesForResource(SamResourceTypeNames.billingProject, project.projectName.value, ctx)
          .map(samRolesToProjectRoles) map { roles =>
          if (roles.nonEmpty) {
            val billingProfile =
              project.billingProfileId.flatMap(id =>
                billingProfileManagerDAO.getBillingProfile(UUID.fromString(id), ctx)
              )
            Some(mapCloudPlatformAndPolicies(project, billingProfile, roles, workspaceManagerDAO))
          } else {
            None
          }
        }
      case None => Future.successful(None)
    }
```

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -59,27 +60,42 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
-            get {
-              parameters(
-                "startDate".as[DateTime],
-                "endDate".as[DateTime],
-                "pageSize".as[Int],
-                "offset".as[Int]
-              ) { (startDate, endDate, pageSize, offset) =>
+        pathPrefix("id") {
+          pathPrefix(Segment) { id =>
+            pathEnd {
```
davidangb: nit: `pathPrefix(…) { pathEnd {}}` is redundant; you can just do:

```suggestion
          path(Segment) { id =>
```

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
```diff
@@ -59,27 +60,42 @@ trait BillingApiServiceV2 extends UserInfoDirectives {
...
-                    endDate.plusDays(1).minusMillis(1),
-                    pageSize,
-                    offset
-                  )
+                  import spray.json._
+                  userServiceConstructor(ctx).getBillingProjectById(UUID.fromString(id)).map {
+                    case Some(projectResponse) => StatusCodes.OK -> Option(projectResponse).toJson
+                    case None => StatusCodes.NotFound -> Option(StatusCodes.NotFound.defaultMessage).toJson
+                  }
                 }
```
davidangb: As a rule, we want to return objects from APIs and let the spray-json implicit formatters handle all the JSON serialization, instead of explicitly calling `.toJson`. The reason you needed to `import spray.json._` and handle your own serialization is that the two cases return different payloads; the `Some` case returns `Option[RawlsBillingProjectResponse]` while the `None` case returns `Option[String]`. If you align the types, you can do away with the .toJson:

```scala
              complete {
                userServiceConstructor(ctx).getBillingProjectById(UUID.fromString(id)).map {
                  case Some(projectResponse) => StatusCodes.OK -> Option(projectResponse)
                  case None                  => StatusCodes.NotFound -> None
                }
              }
```

davidangb (commented):
see Scala esoteric comments inline, this is close~

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/efbbc6a033de0efa715aae06d8ee79469a0144c4

core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/BillingApiServiceV2.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/a150b008292b4d4bc5aa972d9fe8b0e238c552a6

core/src/main/scala/org/broadinstitute/dsde/rawls/user/UserService.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/6efc708f40e9e33608084c727e60a501bb9e0bc3

core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/a3f2d04f283d9f061ccbcc5720d704db3d2caeef

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
marctalbott: Does this need to be reworked now that https://github.com/broadinstitute/rawls/commit/40daaf845287660732d21f9e2cf33adaaca1bcd0 was undone?

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponentSpec.scala
```diff
@@ -31,6 +31,10 @@ class RawlsBillingProjectComponentSpec
     assertResult(None) {
       runAndWait(rawlsBillingProjectQuery.load(project.projectName))
     }
+
+    assertResult(None) {
+      runAndWait(rawlsBillingProjectQuery.loadById(project.id))
+    }
```
marctalbott, line 41: Small thing, but I'd either move this up or add a second `.loadById` check above the deletion that checks that the project is loaded correctly.

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
```diff
@@ -1217,9 +1224,44 @@ class UserServiceSpec
 
   behavior of "getBillingProject"
 
+  it should "return the project when it exists" in {
+    val projectId = UUID.randomUUID()
+    val projectName = RawlsBillingProjectName(UUID.randomUUID().toString)
+    val project = RawlsBillingProject(projectId, projectName, CreationStatuses.Ready, None, None)
+    val repository = mock[BillingRepository]
+    when(repository.getBillingProjectById(projectId)).thenReturn(Future.successful(Some(project)))
+    when(repository.getBillingProject(projectName)).thenReturn(Future.successful(Some(project)))
```
marctalbott: Nit: this second mock isn't needed

davidangb (approved):
nice, thank you for all the revisions! :+1: assuming tests pass

marctalbott (approved):
A couple of minor comments, but otherwise this looks good!

core/src/main/resources/org/broadinstitute/dsde/rawls/liquibase/changesets/20250228_add_billing_project_id.xml
kevinmarete: No need because we add a random UUID when creating billing projects: https://github.com/broadinstitute/rawls/blob/a3f2d04f283d9f061ccbcc5720d704db3d2caeef/core/src/main/scala/org/broadinstitute/dsde/rawls/billing/BillingProjectOrchestrator.scala#L163

core/src/test/scala/org/broadinstitute/dsde/rawls/user/UserServiceSpec.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/d5bfaad39e058c8e98bcd7c1f49ec543c0ed6f5a

core/src/test/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/RawlsBillingProjectComponentSpec.scala
kevinmarete: Fixed: https://github.com/broadinstitute/rawls/pull/3205/commits/65fecb1e876ce23d3bb4e8deae18ca2f9cd5958e

calypsomatic (approved):

PR #3212, opened 2025-03-05
[CORE-368] Validate entitytype rename

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-368
Adds in name validation when renaming an entity.  https://github.com/DataBiosphere/terra-ui/pull/5272 also does this from the front end.

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
@@ -245,7 +245,7 @@ class EntityService(protected val ctx: RawlsRequestContext,
   def renameEntityType(workspaceName: WorkspaceName, oldName: String, renameInfo: EntityTypeRename): Future[Int] = {
     import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction}
 
-    validateEntityName(renameInfo.newName)
+    validateEntityType(renameInfo.newName)
     def validateExistingType(dataAccess: DataAccess,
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/entities/EntityService.scala
```diff
@@ -243,6 +245,7 @@ class EntityService(protected val ctx: RawlsRequestContext,
   def renameEntityType(workspaceName: WorkspaceName, oldName: String, renameInfo: EntityTypeRename): Future[Int] = {
     import org.broadinstitute.dsde.rawls.dataaccess.slick.{DataAccess, ReadAction}
 
+    validateEntityName(renameInfo.newName)
```
davidangb: `validateEntityName` checks syntax for an entity name, not an entity type. See https://github.com/broadinstitute/rawls/blob/4143698146d5505466c9d2648503d985f9c8c153/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L1408-L1423 for the logic around validating entity types. You might wanna break out the two validations inside EntityComponent.validateEntity that look at entityType into a new method, then call that new method from here

calypsomatic: Ah dang I looked very carefully at the different methods and still picked the wrong one 😵

kevinmarete (approved):

davidangb (approved):

PR #3216, opened 2025-03-10
CORE-350: spend report excludes previous accounts

davidangb:
Ticket: CORE-350

Objective of this ticket is to only show spend for the _current_ billing account for each workspace; do not include spend for any previous billing accounts. End result is that workspaces will not show the spend incurred while their projects were in a RBS pool.

Implementation-wise, this changes the where clause in the consolidated spend report.

_previous where clause:_
```sql
where …
project.id in ("project1", "project2", "project3", "project4", "project5")
```

_new where clause:_
```sql
where …
(
  (billing_account_id = 'billingAccount1' and project.id in ('project1', 'project2'))
  or
  (billing_account_id = 'billingAccount2' and project.id in ('project3'))
  or
  (billing_account_id = 'billingAccount3' and project.id in ('project4', 'project5'))
)
```

Here's the before/after of the consolidated spend report for me in the dev env. Note the difference in the fourth row for the `nonbroad-billing-dev/TestMissingSpendInfo` workspace.

_before:_
Screenshot 10-03-2025 at 09 17

_after:_
Screenshot 10-03-2025 at 09 17 (1)

---

**PR checklist**

  - [ ] in workbench-libs

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
@@ -311,6 +311,7 @@ class SpendReportingService(
     )
   }
 
+  // TODO CORE-350: single-project report: query
   def getQuery(aggregations: Set[SpendReportingAggregationKeyWithSub], config: BillingProjectSpendExport): String = {
     // Unbox potentially many SpendReportingAggregationKeyWithSubs for query,
     // all of which have optional subAggregationKeys and convert to Set[SpendReportingAggregationKey]
@@ -337,9 +338,10 @@ class SpendReportingService(
     if (isBroadTable) spendReportingServiceConfig.defaultTimePartitionColumn else "_PARTITIONTIME"
   }
 
+  // TODO CORE-350: consolidated report: query
   def getAllUserWorkspaceQuery(
     spendExportTable: String,
-    workspaces: Seq[GoogleProjectId],
+    billingProjectsByAccount: Map[RawlsBillingAccountName, Seq[GoogleProjectId]],
     pageSize: Int,
     offset: Int
   ): String = {
@@ -357,8 +359,8 @@ class SpendReportingService(
                        |  FROM
                        |    _BILLING_ACCOUNT_TABLE
                        |  where
-                       |    project.id in _PROJECT_ID_LIST AND
-                       |    _PARTITIONTIME BETWEEN @startDate AND @endDate
+                       |    _PARTITIONTIME BETWEEN @startDate AND @endDate and
+                       |    _PROJECT_CLAUSE
                        |  GROUP BY
                        |    project_id,
                        |    spend_category,
@@ -369,11 +371,7 @@ class SpendReportingService(
       baseQuery
         .replace("_PARTITIONTIME", timePartitionColumn)
         .replace("_BILLING_ACCOUNT_TABLE", spendExportTable)
-        .replace("_PROJECT_ID_LIST",
-                 workspaces
-                   .map(projectId => s""""${projectId}"""")
-                   .mkString("(", ", ", ")")
-        )
+        .replace("_PROJECT_CLAUSE", billingProjectsByAccountClause(billingProjectsByAccount))
 
     s"""WITH spend_categories AS (
        |$bpSubQuery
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -605,28 +641,33 @@ class SpendReportingService(
...
+            config -> workspacesByBillingProjectName.getOrElse(config.billingProjectName, Seq.empty[Workspace])
           }
-          table -> combinedWorkspaces
-        }
-      } yield combinedResults
+          .toMap
+          .filter(_._2.nonEmpty)
+
+//        groupedByTable = spendConfigs.groupBy(
+//          _.spendExportTable.getOrElse(spendReportingServiceConfig.defaultTableName)
```
calypsomatic: Nit: remove commented-out code

calypsomatic (approved):
600 thumbs for adding all the helpful comments!

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
kevinmarete: +1

kevinmarete (approved):
Thanks for adding the comments, great implementation.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -485,6 +496,13 @@ class SpendReportingService(
           .toMap
       }
 
+      _ = if (projectNames.isEmpty) {
+        throw RawlsExceptionWithErrorReport(
+          StatusCodes.NotFound,
+          s"no spend data found for billing project ${project.value} between dates ${toISODateString(start)} and ${toISODateString(end)}"
+        )
+      }
+
```
davidangb, line 505: this prevents us from asking BigQuery for results in cases where there would be no results. The need for this was uncovered by fixing up the unit tests.

core/src/main/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingService.scala
```diff
@@ -566,22 +587,45 @@ class SpendReportingService(
...
         billingMap <- getBillingWithSpendPermission(childContext)
         _ = if (billingMap.isEmpty) {
-          return Future.successful(None)
-        }
-        projectNames: Map[GoogleProjectId, WorkspaceName] = billingMap.values.flatten.toMap
-        tableToProjectIdsMap: Map[String, Seq[GoogleProjectId]] = billingMap.map { case (table, workspaces) =>
-          table -> workspaces.map(_._1)
+          throw new RawlsExceptionWithErrorReport(
+            ErrorReport(StatusCodes.NotFound, "No workspaces eligible for spend report")
+          )
```
davidangb, line 596: this now throws instead of early-returning `Future.successful(None)`. Akka was hanging trying to send `None` in the response payload; now the API returns properly in cases where the user has no workspaces on which to report. The UI should avoid this case but it was possible via the API.

core/src/test/scala/org/broadinstitute/dsde/rawls/spendreporting/SpendReportingServiceSpec.scala
```diff
@@ -1420,16 +1429,17 @@ class SpendReportingServiceSpec extends AnyFlatSpecLike with Matchers with Mocki
...
-      e.errorReport.message.contains("no workspaces") shouldBe true
-    }
+    val exception = Await.result(recoverToExceptionIf[RawlsExceptionWithErrorReport] {
+                                   service.getSpendForAllWorkspaces(from, to, 100, 0)
+                                 },
+                                 Duration.Inf
+    )
+
+    exception.errorReport.statusCode shouldBe Option(StatusCodes.NotFound)
+    exception.errorReport.message.contains("No workspaces eligible for spend report") shouldBe true
```
davidangb, line 1439: this unit test, and the `handle errors from bigquery gracefully` test just below, were always passing; in effect, they weren't running. The tests assert on a `Future` but didn't wait for that `Future` to complete. So, the test finished - successfully - while the Future continued to run in the background. I've added the proper `Await`s to them … which also required fixing them up a bit to pass.

kevinmarete (approved):
Good catch on this edge case!

calypsomatic (approved):

